PR #1273, opened 2015-12-28
New inputs and textareas

KimberlyMunoz:
Moved our old input and textarea macros into the new format. I also applied the new styles from @sonnakim and @cfarm's [work on forms here](https://docs.google.com/document/d/1SvgJ2hIyUVo2BzCvoV95NOKtr8ylv2TIdkkYzAXexj0/edit#).

image

I am currently working on the form-l layout grid and submit buttons

If I missed anything that needs to be converted let me know.

## Changes
- Changed styling of microtext to Design Manual specs
- Converted more forms to global_dict
- Converted common-form-fields.html into atoms
- Fixed various spacing issues

## Review
- @anselmbradford 
- @sebworks 
- @cfarm

PR #1276, opened 2015-12-29
Adding context method to return unique id

sebworks:
Adding context method to return unique id

## Additions

- Adding context method to return unique id.

## Testing

- Visit  `http://localhost:8000/the-bureau/leadership-calendar/` and click on labels on the two filters. Should check/uncheck boxes on both filters.

## Review

- @kave 
- @rosskarchner 
- @anselmbradford 
- @KimberlyMunoz 

## Notes

I am using a closure but it may make more sense to use a generator or iterator.

### DIFF
--- a/cfgov/jinja2/v1/_includes/macros/filters.html
+++ b/cfgov/jinja2/v1/_includes/macros/filters.html
@@ -357,17 +357,17 @@
    ========================================================================== #}
 {% macro _filter_checkbox(field, value, label) %}
     {% from 'macros/util/format/url.html' import slugify as slugify %}
-    {% set id_slug = slugify( value ) %}
+    {% set id = get_unique_id('filter_', '_') ~ slugify( value ) %}
     <input class="cf-input"
            type="checkbox"
            name="filter_{{ field }}"
            value="{{ value }}"
-           id="filter_{{ id_slug }}"
+           id="{{ id }}"
            {%- if is_filter_selected(field, value) %}
            checked
            {%- endif -%}
            >
-    <label class="form-group_item" for="filter_{{ id_slug }}">
+    <label class="form-group_item" for="{{ id }}">
         {{ label if label else value }}
     </label>
 {% endmacro %}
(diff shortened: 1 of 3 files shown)


cfgov/v1/util/util.py
```diff
@@ -13,4 +13,12 @@ def id_validator(id_string, search=re.compile(r'[^a-zA-Z0-9-_]').search):
...
-    return components[0] + "".join(x.title() for x in components[1:])
\ No newline at end of file
+    return components[0] + "".join(x.title() for x in components[1:])
+
+class UniqueId(object):
+	index = 0
+
+	@classmethod
+	def get_unique_id(cls, prefix='', suffix=''):
+		cls.index = cls.index + 1
```
anselmbradford: Another approach is to get the current system unix timestamp and add that in place of index.

cfgov/jinja2/v1/_includes/macros/filters.html
```diff
@@ -357,17 +357,17 @@
...
 {% macro _filter_checkbox(field, value, label) %}
     {% from 'macros/util/format/url.html' import slugify as slugify %}
-    {% set id_slug = slugify( value ) %}
+    {% set id = get_unique_id('filter_', '_') ~ slugify( value ) %}
     <input class="cf-input"
            type="checkbox"
            name="filter_{{ field }}"
            value="{{ value }}"
-           id="filter_{{ id_slug }}"
+           id="{{id}}"
```
anselmbradford: Trivial: space out `{{ id }}` here and throughout.

anselmbradford:
Minor comments, LGTM overall though :+1: 

Seems like helper method could be used in https://github.com/cfpb/cfgov-refresh/pull/1273#discussion_r48543903 too.

cfgov/v1/util/util.py
sebworks: So they have `uuid.uuid4()` but I wasn't sure if that was too long. 

https://docs.python.org/2/library/uuid.html

kave:
Can we also use this method for the FormFieldWithButton id so content no longer needs to manually input it?

KimberlyMunoz:
> Can we also use this method for the FormFieldWithButton id so content no longer needs to manually input it?

This seems like a good idea. :bulb:

KimberlyMunoz:
I like it. We'll have to add it into some of the existing form elements, but it'll make it much easier to make sure our forms don't have duplicate ids in the long run. :+1:

kave:
:+1:

Issue #1281, opened 2015-12-29
Wagtail & other issues: Education & Outreach Landing page


Schlachtmeyer:
![educationoutreach_landing](https://cloud.githubusercontent.com/assets/14129916/12040156/7e008d98-ae34-11e5-8a56-67fbcec2ca76.png)

- Spacing okay between Hero & 2/7 molecule?
- Font & spacing issues in sidebar
- What's happening with the images in the 50/50 molecule?
- Two columns needed in the gray well (discussed this ENH with the team, not sure if it has been spec'd)

kave:
@KimberlyMunoz

KimberlyMunoz:
According to @duelj, the spacing is off on top. I'll remove the "block" class and adjust the CSS for .content_main to get it to 30px. This might require some tweaking of the breadcrumbs.

The images in the 50-50 molecule should have a toggle for "is circle or is widescreen" without it they're displaying with the widescreen ratio. I'll look into this.

KimberlyMunoz:
Two column issue is being addressed in #1218, so let's continue following the progress in that issue so we don't track it in too many places. 

Spacing between the hero and molecule has been addressed in #1283.

The sidebar is having weird issues because everything is rendering as a block regardless of what content enters it as. @kave, can you look into this?

The last issue depends on a few assumptions. When I created it, I assumed round images would be uploaded with padding and enforced some rules so it wouldn't break the layout. The upside of this method is that other images can be uploaded in that 4:3 format. I'm checking with Justin to make sure all the images in the 4:3 format will be round circles before I change the styles to enforce that.

KimberlyMunoz:
Talked it over with Justin and #1288 should fix the last issue.

Schlachtmeyer:
This page looks much better now! But I think the spacing is still off in the sidebar.

![educationoutreach_landing_sidebar_0106](https://cloud.githubusercontent.com/assets/14129916/12122063/d522e856-b3a6-11e5-95fe-0b1598d0dca5.png)

kave:
@KimberlyMunoz each side-bar element is surrounded by a block class div. Do we still want to do that?

KimberlyMunoz:
Slug, hyperlink and paragraph shouldn't be in block class divs because that spacing is a too much for this usecase.

jimmynotjim:
Issue opened in Jira as DFJR-678

schaferjh:
[...]

PR #1283, opened 2015-12-30
Fixing spacing on top element in cases when page-intro isn't on the page

KimberlyMunoz:
In #1281, @Schlachtmeyer pointed out the spacing on one of the landing pages is off when there is no page intro. I moved over some code from the sublanding to fix this.

# Before

image

# After

image

## Review
- @kave 
- @anselmbradford 
- @sebworks


sebworks:
:+1:

kave:
:+1:

Issue #1277, opened 2015-12-29
Wagtail & other issues: Credit Card Data sublanding page


Schlachtmeyer:
[Not sure how to post this without stretching the image - help!]

![creditcarddata_sublanding](https://cloud.githubusercontent.com/assets/14129916/12039829/cd5a0fac-ae31-11e5-8190-c60c07f6cf57.png)

Is spacing correct between Hero and Links Blob?
Is spacing correct between the Links Blob paragraph and link?
Is spacing correct in the E-Mail Signup between field and disclaimer text?

For the E-Mail Signup, what should the heading be (@sarahsimpson09 @tnferrell) and if the heading becomes "Email Signup" or something short, where should that sentence go?

KimberlyMunoz:
Tagging @duelj to weigh in on the spacing questions. For the first one, my fix in #1283 should take care of it.

KimberlyMunoz:
Actually, nevermind. The spacing underneath the hero is a new issue. Looking into it now.

schaferjh:
@Schlachtmeyer does the page look stretched on your computer/in the review mode? Or is it just when you post the image in the comment?

Schlachtmeyer:
They're just stretched when I attach the images, sorry about that.

schaferjh:
No, it's fine -- if it's just an image attachment thing then I'm both not sure how to fix it (never run into it before) and not worried about it. :neutral_face:  A stretched image works fine for our purposes here, as long as it's rendering fine in Wagtail!

KimberlyMunoz:
Looking into fixing the link blob and email signup spacing issue now.

duelj:
**Hero:**  looks good and should match spacing found [here](https://huetingj.github.io/design-manual/ui-toolkit/heroes.html)

**Space after hero:** 30px. This would be treated as if the bottom of the hero were a stroke dividing an introduction section from the rest of the page (even though there is no stroke). The sidebar background should come flush to the bottom of the hero background and just the content inside should be 30px below the bottom of the hero. 

**Space between text blob and link:** 15px to match [updated spacing](https://github.com/cfpb/design-manual/issues/269#issuecomment-157544901) between paragraphs of text 

**Email sign up label:** The label should be adjusted to "stay informed" or whatever content specifies and the current label should look like paragraph text following the style of the [email sign up molecule](https://github.cfpb.gov/flapjack/Modules-V1/wiki/Email-Sign-Up)

[...]

PR #1284, opened 2015-12-30
Removes empty pre-content block when no breakout sidebar is present

KimberlyMunoz:
In #1277, @Schlachtmeyer pointed out a weird gap between the hero and content area. I've removed that by checking for a sidebar breakout before rendering since the introduction text in the pre-content area should always be paired with one. This might make debugging a little tricky for content people, so I'm open to other solutions.

# Before
image

# After
image

# Review
- @kave 
- @sebworks 
- @anselmbradford


kave:
:+1:

Issue #1225, opened 2015-12-15
Fix the table organism to include prototype data


KimberlyMunoz:
It was/will be removed in #1224 since it's causing an error on the learn-page.

PR #1253, opened 2015-12-21
Page Specifc JS based on atomic elements

kave:
In this concept, atomic elements will define their associated JS which will be rendered on page load.

### Testing
- create a demo page with Expandable/Expandable Group elements.

## Reviews
@rosskarchner 
@anselmbradford

PR #1286, opened 2015-12-30
Clean up prototype pages

anselmbradford:
Clean up prototype pages

## Changes

- Add missing atomic elements to doc detail and learn pages that are already built.
- Minor updating of global_dict format of image and text 25/75 and well.

## Testing

- `/document-detail/` and `/learn-page/`

## Review

- @sebworks 
- @KimberlyMunoz


cfgov/jinja2/v1/browse-basic/index.html
```diff
@@ -78,10 +78,10 @@
              %}
             {% set value = global_dict.expandable %}
             {% include 'molecules/expandable.html' with context %}
+            {% set value = None if prototype %}
```
sebworks: Does this need and `else` condition?

sebworks:
:+1:

KimberlyMunoz:
Fixes #1225

KimberlyMunoz:
Also, I notice expandables aren't working after a gulp build. That's fine if it was like that before and out of scope for this PR, but we should make a note if it's an issue.

anselmbradford:
> Also, I notice expandables aren't working after a gulp build. That's fine if it was like that before and out of scope for this PR, but we should make a note if it's an issue.

Yeah, that's broken after https://github.com/cfpb/cfgov-refresh/pull/1253 I believe. @kave would know whether that matters or not for the prototype pages.

cfgov/jinja2/v1/browse-basic/index.html
anselmbradford: I'm actually unclear when the prototype pages would be viewed without the `prototype` value being present. The purpose of this is when setting the mock data in the `value` for the prototype pages, if you don't null it out the next atomic element gets the date from the previous atomic element. So like the text intro gets expandable data, etc.

PR #1289, opened 2015-12-31
Fix for get_unique_id method

sebworks:
Fixing get_unique_id method to return hex based time stamp

## Additions

- Fixing get_unique_id method to return hex based time stamp

## Testing

- View filter checkbox ids on the `http://localhost:8000/the-bureau/leadership-calendar/` page.

## Review
- @anselmbradford 
- @KimberlyMunoz 
- @kave

### DIFF
--- a/cfgov/v1/util/util.py
+++ b/cfgov/v1/util/util.py
@@ -1,5 +1,7 @@
 import re
 from django.conf import settings
+from time import time
+
 
 
 def id_validator(id_string, search=re.compile(r'[^a-zA-Z0-9-_]').search):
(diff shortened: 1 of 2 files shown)


cfgov/v1/util/util.py
```diff
@@ -15,14 +17,20 @@ def to_camel_case(snake_str):
     components = snake_str.split('_')
     return components[0] + "".join(x.title() for x in components[1:])
 
+
+def get_unique_id(prefix='', suffix=''):
+    index = hex(int(time()*10000000))[2:]
```
anselmbradford: Why does time need to be multiplied by `10000000`?

sebworks: It's a floating point number so you want to include the milliseconds before casting to integer. That is what will help make it unique if you have concurrent calls to `get_unique_id`.

anselmbradford: Why 10 million though? Isn't it a float out only two decimal places?

sebworks: So I think it depends on float point precision.  `time.time()` returns `1451584127.121017` at this moment.

anselmbradford: Hm, interesting. I set `index = time()` and it was two spaces in the output HTML for me. :+1:

anselmbradford:
:+1:

cfgov/v1/util/util.py
sebworks: Yep, same for me when the floating point number is converted into a string. 

`str(time.time())` produces `1451585483.05`

PR #1292, opened 2016-01-02
fixed JS loading for prototype pages

kave:
This will fix the rendering of JS for prototype pages i.e the browse-page

@anselmbradford 
@KimberlyMunoz


anselmbradford:
Tested and works :+1:

PR #1296, opened 2016-01-04
Fixing initial load state of Filter

sebworks:
Fixing initial load state of Filter

## Changes

-  Modified `cfgov/unprocessed/css/molecules/expandable.less` to target `.m-expandable_content[aria-expanded='true']`

## Testing

- Visit `http://localhost:8000/browse-filterable/` and turn your js off. 

## Review
- @anselmbradford 
- @KimberlyMunoz

### DIFF
--- a/cfgov/unprocessed/css/molecules/expandable.less
+++ b/cfgov/unprocessed/css/molecules/expandable.less
@@ -164,6 +164,17 @@
     }
 }
 
+// Setting height to prevent flash on load.
+.m-expandable_content:not([aria-expanded='true']) {
+    height: 0;
+}
+
+.no-js {
+    .m-expandable_content:not([aria-expanded='true']) {
+        height: 100%;
+    }
+}
+
 .m-expandable {
     // Modifiers
     &__borders {


cfgov/jinja2/v1/_includes/molecules/expandable.html
```diff
@@ -40,12 +40,11 @@
     %}
 {% endif %}
 {% set value = value or global_dict.expandable %}
-
 <div class="m-expandable
-            m-expandable__expanded
```
anselmbradford: This is intentional, as with this change the expandable state where it is expanded on page load will not work. See fourth expandable on `/browse-basic/`, which should be expanded on page load -->

screen shot 2016-01-04 at 9 56 26 am

sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/commit/4b62b60e5fbeb77cf6c5cc4f29c2bc12f09a272c

anselmbradford: So this'll make it so when JS is off they don't have the `m-expandable__expanded` modifier, yet they'll be expanded. So for example, they'll all be expanded, but the markup won't be equal across all of them, which shouldn't be the case:

screen shot 2016-01-04 at 10 28 29 am

sebworks: @anselmbradford, I've removed that code and have opted to used an animation. This will only work on IE10+ however. Not sure if this is the best approach.

anselmbradford:
Why is the fade in needed? It doesn't look very smooth to me because the whole thing doesn't fade in so the border is broken momentarily:

fadein

anselmbradford:
Looks good :+1: This likely will be ignored in IE8, but the expandables still work in IE8 so that should be fine. Possibly include a code comment above these changes that says why they're necessary.

PR #1274, opened 2015-12-29
Integrate notification into filter

anselmbradford:
Integrate notification into filter.

## Additions
- Adds JS file for Notification instance to change content, show, and hide at runtime.

## Changes

- Integrates notification into filter. 
- Moves validation library from bower to npm.
- Wraps notification content in a div so that it can be overwritten at once.

## Removes
- Padding below the expandables. I don't think this is required and it messes things up for, e.g. the spacing of an expandable inside a filter.

## Testing

- Submit filter on `/browse-filterable/`

## Review

- @KimberlyMunoz 
- @sebworks 

## Screenshots

notify

## Todos

- Validation needs to be added. This PR adds some of the skeleton for it, but the existing validation JS needs to be integrated.

PR #1299, opened 2016-01-04
Backend: Notifications

kave:
*Note* Integration into Filterable List will happen in a Future PR

screen shot 2016-01-04 at 11 53 02 am
@anselmbradford 
@kurtw


cfgov/v1/models/molecules.py
```diff
@@ -226,3 +226,20 @@ class Meta:
...
+                                                         "depending on notification type.")
+    explanation = blocks.TextBlock(required=False,
+                                   help_text="Text to display as an explanation within the notification markup.")
+
+    class Meta:
+        icon = 'view'
+        template = '_includes/molecules/notification.html'
+
+    class Media:
+        js = ("notification.js",)
```
kurtw: God I love this.

kave: :dancer:

kurtw:
I'm a bit unclear on how we're going to be using this. With this, we have a way of creating 3 different types of notifications but how do we identify certain `error` or `warning` notifications for different situations?

Would it make sense to have a notification on a "per field" basis? e.g. The date range has it's own set of notifications.

kave:
@kurtw take a look at this PR https://github.com/cfpb/cfgov-refresh/pull/1274 pretty much these would be used in conjunction with other organisms (probably as streamBlocks within them) if you wanted to customize the messages presented by the notifications.

cfgov/unprocessed/js/routes/on-demand/expandable.js
```diff
@@ -18,12 +17,4 @@ var expandable;
...
-}
-
-selector = '.o-expandable-group';
-var expandableGroup = document.querySelectorAll( selector );
-var regularExpandableGroup = new ExpandableGroup( expandableGroup[0] );
-var accordionExpandableGroup = new ExpandableGroup( expandableGroup[1] );
-
-regularExpandableGroup.init();
-accordionExpandableGroup.init();
+}
```
jimmynotjim: Minor, add newline at end of file back in

cfgov/unprocessed/js/routes/on-demand/notification.js
```diff
@@ -0,0 +1,14 @@
...
+   ========================================================================== */
+
+'use strict';
+
+var Notification = require( '../../molecules/Notification' );
+
+var notifications = document.querySelectorAll( '.m-notification' );
+var notification;
+for ( var i = 0, len = notifications.length; i < len; i++ ) {
+	notification = new Notification( notifications[i] ).init();
```
jimmynotjim: Extra indentation, JS is 2-space

kurtw:
Right, but that doesn't address when specific messages are used. I'm assuming that we want specific error and warning messages to be shown in a scenario, like when the user inputs the wrong thing or something.

I think what I'm saying is the we should have an identifying field for a notification so we know which one we're showing.

kave:
@kurtw eh, i mean sure but do we really want to have big notifications for on field errors? We can just use form errors to highlight incorrect fields. i think these notifications are primarily for submission notifications @anselmbradford  what do you think?

anselmbradford:
If I recall correctly, the field errors would show in cases where a browser supports them, but not all browsers do, so the main notification would need to be capable of displaying them also in those cases.

kave:
@anselmbradford really? thats a first. now i'm just curious for my own knowledge what browsers did you see this happen? IE ?

anselmbradford:
@kave Yeah less than IE10 and Safari (surprisingly)—which supports the form validation attributes but doesn't display a notice. http://caniuse.com/#search=required

kave:
@kurtw so after speaking it over with ans, we decided the most consistent approach is to just have the bulleted list of forms errors added to the notification `explanation` section.

having a `field name` box may over complicate things as content would have to know the id/name of the html fields their associating the notifications too and these would potentially be auto-generated.

PR #1229, opened 2015-12-16
New Select Box

KimberlyMunoz:
TIL I learned some new things about styling select boxes in Firefox. But we should be able to move to a JS-solution here as well so long as we're okay with the placeholder being unstyled on macs.

## Changes

- Removes Jquery plugin for select boxes
- Adds styles from design manual for select boxes
- Moves over to the include/global dict style over macros for select boxes and state selectors.

## Review

- @jimmynotjim 
- @anselmbradford 
- @sebworks 

## Preview

image

Hover

image

Disabled

image

PR #1301, opened 2016-01-04
Moves modernizr to build process

anselmbradford:
Moves modernizr to build process. 

## Changes

- Moves modernizr added in https://github.com/cfpb/cfgov-refresh/pull/1229 to the gulp scripts task.
- Handles html5shiv with modernizr.
- Handles function.bind with modernizr.

## Testing

- `csspointerevents` class should be present on root element, and `no-csspointerevents` should appear on <IE10.

## Review

- @KimberlyMunoz 
- @sebworks 
- @jimmynotjim


gulp/tasks/scripts.js
```diff
@@ -3,19 +3,31 @@
...
 var webpackConfig = require( '../../config/webpack-config.js' );
 var webpackStream = require( 'webpack-stream' );
-var paths = require( '../../config/environment' ).paths;
 
 gulp.task( 'scripts', function() {
   return gulp.src( paths.unprocessed + '/js/routes/common.js' )
+    .pipe( gulpModernizr( {
+      tests:   ['csspointerevents'],
+      options: ['setClasses',
+                'html5printshiv',
```
KimberlyMunoz: Nice.

package.json
```diff
@@ -20,11 +20,13 @@
     "gulp-less": "^3.0.3",
     "gulp-load-plugins": "^1.1.0",
     "gulp-imagemin": "^2.3.0",
+    "gulp-modernizr": "^1.0.0-alpha",
     "gulp-mq-remove": "0.0.2",
     "gulp-notify": "^2.2.0",
     "gulp-sitespeedio": "^0.0.8",
     "gulp-sourcemaps": "^1.5.2",
     "gulp-replace": "^0.5.3",
+    "gulp-uglify": "^1.5.1",
```
KimberlyMunoz: Why is `gulp-uglify` a dependency when `gulp-rename` is a dev-dependency? Are we all the modules in gulp build moving to dependencies since they're required to build the site?

KimberlyMunoz:
Checked this out in IE10 and IE8 and it looks great! One minor question, but :+1:

package.json
anselmbradford: So I built up `dependencies` by moving `dev-dependencies` over one at a time and running `setup.sh production` and `setup.sh development` and `setup.sh test` and making sure it still built. `gulp-rename` is making it over because its a dependency of `gulp-cssmin`.

jimmynotjim:
:+1:

package.json
KimberlyMunoz: Sounds good to me.

PR #1309, opened 2016-01-06
Updates gulp-load-plugins to 1.2.0

anselmbradford:
## Changes

- Updates gulp-load-plugins to 1.2.0. I don't get why the caret isn't autoupdating this, but it isn't.

## Review

- @jimmynotjim 
- @sebworks 
- @KimberlyMunoz


sebworks:
:+1:

PR #1312, opened 2016-01-06
Adds FCM to sublanding prototype

anselmbradford:
## Changes

- Adds FCM to sublanding prototype


anselmbradford:
@jimmynotjim The CSS will have to be revisited as the map is getting cropped by the right-hand sidebar, but :+1: for this PR to add FCM to the page. Squash and merge at will. Thanks for the help!

jimmynotjim:
:+1:

PR #1316, opened 2016-01-06
Fixes to spacing on 50-50 and 25-75 text-image molecules

### LINKED ISSUE #1235: Image Text 50/50 has too much spacing below
It looks like padding that's added for when they stack on small screens needs to be scoped to just small screens and conditioned out of the last block to avoid too much distance between them and the next block.

screen shot 2015-12-16 at 4 19 29 pm

KimberlyMunoz:
Fixes #1235 by removing the spacing from the molecule to have it handled by `.block` and fixing the spacing on the header for the 50-50 molecule.

image

## Review
- @jimmynotjim 
- @anselmbradford 
- @sebworks

### DIFF
--- a/cfgov/unprocessed/css/molecules/image-text-50-50.less
+++ b/cfgov/unprocessed/css/molecules/image-text-50-50.less
@@ -82,7 +82,8 @@
 */
 
 .m-image-text-50-50 {
-    padding-bottom: unit( @grid_gutter-width * 2 / @base-font-size-px, em );
+
+    margin-bottom: unit( @grid_gutter-width * 2 / @base-font-size-px, em );
 
     &_container {
         height: 150px;
@@ -119,10 +120,6 @@
         });
     }
 
-    &__bottom-rule {
-        border-bottom: 1px solid @m-image-text-50-50_rule;
-        margin-bottom: unit( @grid_gutter-width / @base-font-size-px, em );
-    }
     .respond-to-min(@bp-sm-min, {
         .grid_column(6);
     });
--- a/cfgov/unprocessed/css/organisms/image-text-50-50-group.less
+++ b/cfgov/unprocessed/css/organisms/image-text-50-50-group.less
@@ -16,10 +16,18 @@
     - cfgov-molecules
 */
 .o-image-text-50-50-group {
+    // TODO: Find another way to handle spacing
+    margin-bottom: unit( @grid_gutter-width * -2 / @base-font-size-px, em );
+
     .respond-to-min(@bp-sm-min, {
         margin-left: -@grid_gutter-width / 2;
         margin-right: -@grid_gutter-width / 2;
     });
+
+    > h2 {
+        margin-left: @grid_gutter-width / 2;
+        margin-right: @grid_gutter-width / 2;
+    }
 }
 
 /* topdoc
--- a/cfgov/jinja2/v1/_includes/molecules/image-text-50-50.html
+++ b/cfgov/jinja2/v1/_includes/molecules/image-text-50-50.html
@@ -56,8 +56,11 @@
     {% set value = global_dict.image_text_50_50 %}
     {% set prototype = true %}
 {% endif %}
-<div class="m-image-text-50-50
-            {{ 'm-image-text-50-50__bottom-rule' if value.has_rule else '' }}">
+<div class="m-image-text-50-50">
+    {# TODO: Change to HR #}
+    {% if value.has_rule %}
+        <div class="a-rule-break"></div>
+    {% endif %}
 
     <div class="{{ 'm-image-text-50-50_widescreen-container' if value.is_widescreen
             else 'm-image-text-50-50_container'}}">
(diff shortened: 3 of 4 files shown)


cfgov/unprocessed/css/molecules/image-text-50-50.less
```diff
@@ -119,10 +118,10 @@
         });
     }
 
-    &__bottom-rule {
+    &.m-image-text-50-50__bottom-rule {
```
anselmbradford: Looks like this should probably follow the 25/75 format by including a div with the rule atom: https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/jinja2/v1/_includes/molecules/image-text-25-75.html#L59-L61

jimmynotjim: It should be at the top of the following block, the spacing kinda relies on it.

KimberlyMunoz: Fixed in 7f1b3d7. Although I'll have to open a new issue because my admin view doesn't have a way to test it.

KimberlyMunoz:
Can I get a review: @anselmbradford, @sebworks and @jimmynotjim?

sebworks:
There is some overlap on mobile:

screen shot 2016-01-14 at 11 44 54 am

cfgov/unprocessed/css/molecules/image-text-50-50.less
```diff
@@ -82,7 +82,6 @@
 */
 
 .m-image-text-50-50 {
-    padding-bottom: unit( @grid_gutter-width * 2 / @base-font-size-px, em );
```
jimmynotjim: This won't work if there are more than two 50/50s

jimmynotjim: An option could be doing 
```
&+& {
  padding-top: unit( @grid_gutter-width * 2 / @base-font-size-px, em );
}
```
so that it only gets added when there's more than one and it won't affect the flow following them.


Side thought, why aren't we using margin?

KimberlyMunoz: I removed that line and added the spacing on the `.m-image-text-50-50-group` element to avoid the need to use nth-of-type or not.

KimberlyMunoz:
@sebworks thanks for the catch. Can you take a look now?

sebworks:
:+1:

cfgov/unprocessed/css/molecules/image-text-50-50.less
jimmynotjim: I don't think that's going to work on a page like `/consumer-help/` where there's six of them within the group.

KimberlyMunoz: Let me try this in the CMS. I'm pretty sure that it makes you create it two modules at a time, but I'll fix if it allows this and it'll just be something that older IE users will just suffer through.

anselmbradford: The individual molecules shouldn't know if they're around other molecules, so shouldn't know about their margin. That should be set one level above them where it's controlled where they appear. For example,

https://github.com/cfpb/cfgov-refresh/blob/spacingfix/cfgov/unprocessed/css/molecules/image-text-50-50.less#L87

Should actually be defined in `.o-image-text-50-50-group`

KimberlyMunoz: Drat. The CMS does let you add six to one group. If we define it at the group level we'll need to fix that, but it's easier to do what Jimmy's suggesting.

sebworks: Can we do something like this: https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/unprocessed/css/organisms/sidebar-content.less?

jimmynotjim: Yeah, you have to be able to add as many as you want to the group because of the `h2` heading at the top of the group. You wouldn't want to break up the doc outline to fit a layout restriction.

jimmynotjim: Also, this will all be solved with a `half-width` organism that controls layout vs the molecules.

KimberlyMunoz: Fixed it by setting spacing after each `.m-image-text-50-50` and removing the block spacing on the bottom around the group.

cfgov/unprocessed/css/organisms/image-text-50-50-group.less
```diff
@@ -16,10 +16,18 @@
     - cfgov-molecules
 */
 .o-image-text-50-50-group {
+
+    margin-bottom: unit( @grid_gutter-width * -2 / @base-font-size-px, em );
```
jimmynotjim: We can go with this for now, but this seems like it's bound to give us trouble later. Can you add a TODO to explore other layout options just so we know to come back to this?

KimberlyMunoz: Squashed and added!

jimmynotjim:
:+1:

PR #1314, opened 2016-01-06
Create BrowseFilterablePage and FilterForm

kurtw:
- Add BrowseFilterablePage model
- Add FilterForm form for filterable list
- Add BaseExpandable class for expandable controls
- Add FilterControls organism using BaseExpandable
- Add url_parameters macro to handle adding existing get URL parameters into links
- Change jinja2 templates to handle Wagtail page
 - browse-filterable.html
 - filterable-list-controls.html
 - post-preview.html
 - notification.html
 - pagination.html

## TEST
Test 1:
- Create a BrowseFilterablePage
- Create Learn, Doc Detail, or Event pages as descendants (child pages) of the BrowseFilterablePage
- Play around with the controls to filter the list

Test 2:
- Create Create Learn, Doc Detail pages with Blog type categories (not as descendants) of BrowsefilterablePage
- Select `Newsroom` as the page type in the filter controls
- Go to filter page to check if they are there (they should be)
 - Additionally check that they have the Blog category (not their specific one)

Test 3
- Test Preview functionality works

Test 4
- Make sure that sharing works by only "sharing" the pages then going to the filter page at the `content.localhost` site and finding them. They should show as well, but not if going to the `localhost` site.

@KimberlyMunoz 
@sebworks 
@jimmynotjim 
@anselmbradford 
@kave 
@rosskarchner

PR #1322, opened 2016-01-08
Added CFGOVImage and CFGOVRendition

kurtw:
- Added data migration for old images saved under previous type.

@kave 
@rosskarchner 

## To Test
- **Before you run migrate** make sure you have images in your DB to see if they convert properly
- run migrate
- check images
- check to see if alt is there
- check to see if you can save alt data
- check that deletion deletes the instances from BOTH CFGOVImage and CFGOVRendition (just go into shell and query)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -148,7 +148,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Added a new nav-link molecule macro and styles.
 - Added Related Links to Sidebar/Footer.
 - Added Related Metadata molecule.
-
+- Added custom image and rendition models CFGOVImage and CFGOVRendition
 
 ### Changed
 - Updated the primary nav to move focus as user enters and leaves nav levels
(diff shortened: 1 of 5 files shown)


kave:
works fine, just need to remove the ImageBasic Model, so its no longer required, and probably either delete `ImageBasicUrl` or change it from `ImageBasicUrl` into `ImageUrlAlt` in case we ever need a block that just takes an image href instead of uploading it like so:
```
class ImageUrlAlt(blocks.StructBlock):
    url = blocks.CharBlock(required=False)
    alt = blocks.CharBlock(required=False)
```

cfgov/v1/migrations/0031_cfgovimage_cfgovrendition.py
```diff
@@ -0,0 +1,78 @@
...
+import wagtail.wagtailadmin.taggable
+import wagtail.wagtailimages.models
+import django.db.models.deletion
+from django.conf import settings
+import taggit.managers
+
+
+def migrate_images_data(apps, schema_editor):
+    CFGOVImageModel = apps.get_model('v1.CFGOVImage')
+    UserModel = apps.get_model(settings.AUTH_USER_MODEL)
```
vccabral: I didn't know you could do that. Cool.

kurtw:
Hey can we get this merged and  #1324 merged so we can do a final review of Browse Filterable Page #1314 and pull that in?

kave:
can you create an issue to eventually remove the `ImageBasic` model from atoms then :+1: after conflicts have been resolved

PR #1330, opened 2016-01-12
Adds input-contains-label class

anselmbradford:
## Additions

- ~~`input-contains-icon`~~ `input-contains-label` for placing an icon in front and at the end of an input field, plus modifiers for specific icons.

## Testing

- Demo at bottom of `/browse-basic/`.

## Review

- @jimmynotjim 
- @sebworks 
- @KimberlyMunoz 

## Screenshots

Mobile:
screen shot 2016-01-12 at 10 31 48 am

Tablet:
screen shot 2016-01-12 at 10 06 05 am

Desktop:
screen shot 2016-01-12 at 10 05 57 am

## Todos

-  `btn-inside-input` in cf-forms is deprecated by `input-contains-icon`, so that the classes can follow the format of `input-with-btn`. Also, the modifiers are added so that the spacing is controlled. `btn-inside-input` allows a button and a label but doesn't adjust the input padding based on the content of button.

### DIFF
--- a/cfgov/jinja2/v1/browse-basic/index.html
+++ b/cfgov/jinja2/v1/browse-basic/index.html
@@ -185,6 +185,89 @@
         {% endfor %}
     {% endif %}
 
+
+    {# TODO: Remove the below when input-contains-label
+             class is added cf-enhancements. #}
+    <div class="block">
+        <div class="block">
+            <div class="input-contains-label">
+                <label for="search-input1"
+                       class="input-contains-label_before
+                              input-contains-label_before__search">
+                </label>
+                <label for="search-input1"
+                       class="input-contains-label_after
+                              input-contains-label_after__clear">
+                </label>
+                <input id="search-input1"
+                       type="text"
+                       value="This is some really long text to make sure that the button doesn't overlap the content in such a way that this input becomes unusable.">
+            </div>
+        </div>
+
+        <div class="block">
+            <div class="input-contains-label
+                        input__super">
+                <label for="search-input2"
+                       class="input-contains-label_before
+                              input-contains-label_before__search">
+                </label>
+                <label for="search-input2"
+                       class="input-contains-label_after
+                              input-contains-label_after__clear">
+                </label>
+                <input class="input__super"
+                       id="search-input2"
+                       type="text"
+                       value="This is some really long text to make sure that the button doesn't overlap the content in such a way that this input becomes unusable.">
+            </div>
+        </div>
+
+
+        <div class="input-with-btn">
+            <div class="input-with-btn_input
+                        input-contains-label">
+                <label for="search-input3"
+                       class="input-contains-label_before
+                              input-contains-label_before__search">
+                </label>
+                <label for="search-input3"
+                       class="input-contains-label_after
+                              input-contains-label_after__clear">
+                </label>
+                <input type="text"
+                       id="search-input3"
+                       value="This is some really long text to make sure that the button doesn't overlap the content in such a way that this input becomes unusable.">
+            </div>
+            <div class="input-with-btn_btn">
+                <button class="btn">Search</button>
+            </div>
+        </div>
+
+        <div class="input-with-btn">
+            <div class="input-with-btn_input
+                        input-contains-label
+                        input__super">
+                <label for="search-input4"
+                       class="input-contains-label_before
+                              input-contains-label_before__search">
+                </label>
+                <label for="search-input4"
+                       class="input-contains-label_after
+                              input-contains-label_after__clear">
+                </label>
+                <input class="input__super"
+                       type="text"
+                       id="search-input4"
+                       value="This is some really long text to make sure that the button doesn't overlap the content in such a way that this input becomes unusable.">
+            </div>
+            <div class="input-with-btn_btn">
+                <button class="btn btn__super">Search</button>
+            </div>
+        </div>
+
+    </div>
+
 {% endblock %}
 
 {% block content_sidebar scoped -%}
--- a/cfgov/unprocessed/css/cf-enhancements.less
+++ b/cfgov/unprocessed/css/cf-enhancements.less
@@ -1127,6 +1127,199 @@ textarea {
 }
 
 /* topdoc
+  name: Icon inside input
+  family: cf-forms
+  patterns:
+    - name: Default icons before and after a default input.
+      markup: |
+        <div class="input-contains-label">
+            <label for="search-input"
+                   class="input-contains-label_before
+                          input-contains-label_before__search">
+            </label>
+            <label for="search-input"
+                   class="input-contains-label_after
+                          input-contains-label_after__clear">
+            </label>
+            <input type="text"
+                   value="This is some really long text to make sure that the button doesn't overlap the content in such a way that this input becomes unusable.">
+        </div>
+    - name: Super-sized icons before and after a super input.
+      markup: |
+        <div class="input-contains-label
+                    input__super">
+            <label for="search-input"
+                   class="input-contains-label_before
+                          input-contains-label_before__search">
+            </label>
+            <label for="search-input"
+                   class="input-contains-label_after
+                          input-contains-label_after__clear">
+            </label>
+            <input class="input__super"
+                   type="text"
+                   value="This is some really long text to make sure that the button doesn't overlap the content in such a way that this input becomes unusable.">
+        </div>
+    - name: Default icons before and after a default input, with a button.
+      markup: |
+        <div class="input-with-btn">
+            <div class="input-with-btn_input
+                        input-contains-label">
+                <label for="search-input"
+                       class="input-contains-label_before
+                              input-contains-label_before__search">
+                </label>
+                <label for="search-input"
+                       class="input-contains-label_after
+                              input-contains-label_after__clear">
+                </label>
+                <input type="text"
+                       id="search-input"
+                       value="This is some really long text to make sure that the button doesn't overlap the content in such a way that this input becomes unusable.">
+            </div>
+            <div class="input-with-btn_btn">
+                <button class="btn">Search</button>
+            </div>
+        </div>
+    - name: Super-sized icons before and after a super input, with a button.
+      markup: |
+        <div class="input-with-btn">
+            <div class="input-with-btn_input
+                        input-contains-label
+                        input__super">
+                <label for="search-input"
+                       class="input-contains-label_before
+                              input-contains-label_before__search">
+                </label>
+                <label for="search-input"
+                       class="input-contains-label_after
+                              input-contains-label_after__clear">
+                </label>
+                <input class="input__super"
+                       type="text"
+                       id="search-input"
+                       value="This is some really long text to make sure that the button doesn't overlap the content in such a way that this input becomes unusable.">
+            </div>
+            <div class="input-with-btn_btn">
+                <button class="btn btn__super">Search</button>
+            </div>
+        </div>
+  tags:
+    - cf-forms
+*/
+
+.input-contains-label {
+    position: relative;
+
+    input[type="text"],
+    input[type="search"],
+    input[type="email"],
+    input[type="url"],
+    input[type="tel"],
+    input[type="number"] {
+        box-sizing: border-box;
+        width: 100%;
+    }
+
+    &_before ~ input[type="text"],
+    &_before ~ input[type="search"],
+    &_before ~ input[type="email"],
+    &_before ~ input[type="url"],
+    &_before ~ input[type="tel"],
+    &_before ~ input[type="number"] {
+        padding-left: unit( 25px / @btn-font-size, em );
+
+        &.input__super {
+            padding-left: unit( 35px / @super-btn-font-size, em );
+        }
+    }
+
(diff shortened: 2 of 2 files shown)


anselmbradford:
cc @Scotchester

Scotchester:
If the thing the input is containing might be a button with a label, I'm not wild about the name `input-contains-icon`. Perhaps we could genericize it to something like `input-containing-stuff`? That seems only marginally better... this is a tough one.

It'd be nice to be able to do away with the wrapper completely, but I'm not sure we could get the positioning to work without `absolute`.

Here's a wacky idea. What if we had a `u-abs-pos` utility class? Then we could just replace with `input-contains-icon` with that generic class, e.g.:

```html
<div class="u-abs-pos">
```

Scotchester:
I think I would prefer to place the `_after` item after the `input` in the markup. Not a dealbreaker for me, though.

cfgov/unprocessed/css/cf-enhancements.less
```diff
@@ -1127,6 +1127,218 @@ textarea {
...
+        </div>
+  tags:
+    - cf-forms
+*/
+
+.input-contains-icon {
+    position: relative;
+
+    &_before {
+        .webfont-regular;
```
Scotchester: Trivial: Add `()` after the name when calling mixins.

cfgov/unprocessed/css/cf-enhancements.less
```diff
@@ -1127,6 +1127,218 @@ textarea {
...
+        position: absolute;
+        top: 0;
+        left: unit( 10px / @btn-font-size, em );
+        margin: unit( @btn-v-padding / @btn-font-size, em ) 0;
+        line-height: normal;
+        font-size: unit( 14px / @base-font-size-px, em );
+        pointer-events: none;
+    }
```
Scotchester: Property order needs updating here.
1. Box model (from inside out)
2. Positioning
3. Everything else alphabetical

cfgov/jinja2/v1/browse-basic/index.html
```diff
@@ -185,6 +185,95 @@ <h2 class="header-slug">
...
 
+
+    {# TODO: Remove the below when input-contains-icon
+             class is added cf-enhancements. #}
+    <div class="block">
+        <div class="block">
+            <div class="input-contains-icon">
+                <div class="input-contains-icon_before
+                            input-contains-icon_before__search">
+                    <span class="input-contains-icon_icon"></span>
```
jimmynotjim: Do we really need the `span`s? Seems you could use the icon font mixins on the divs to create the icons in pseudo elems and reduce the markup.

sebworks: Is there anyway to reduce the markup? Seems like it's a little heavy.

cfgov/unprocessed/css/cf-enhancements.less
```diff
@@ -1127,6 +1127,218 @@ textarea {
...
+            .cf-icon();
+            content: @cf-icon-search;
+        }
+    }
+
+    // Add "x Clear" button to input field.
+    &_after__clear {
+        .input-contains-icon_icon:before {
+            .cf-icon();
+            content: @cf-icon-delete;
```
Scotchester: Has your team discussed the possibility of using the delete icon with the circular background, to avoid any possible confusion with the text of the input?

cfgov/unprocessed/css/cf-enhancements.less
```diff
@@ -1127,6 +1127,218 @@ textarea {
...
+        &_after__clear ~ input[type="text"],
+        &_after__clear ~ input[type="search"],
+        &_after__clear ~ input[type="email"],
+        &_after__clear ~ input[type="url"],
+        &_after__clear ~ input[type="tel"],
+        &_after__clear ~ input[type="number"] {
+            padding-right: unit( 65px / @btn-font-size, em );
+
+            &.input__super {
+                //padding-right: unit(90px / @super-btn-font-size, em);
```
sebworks: extra comment?

cfgov/unprocessed/css/cf-enhancements.less
```diff
@@ -1127,6 +1127,218 @@ textarea {
...
+
+.input-contains-icon {
+    position: relative;
+
+    &_before {
+        .webfont-regular;
+        position: absolute;
+        top: 0;
+        left: unit( 10px / @btn-font-size, em );
+        margin: unit( @btn-v-padding / @btn-font-size, em ) 0;
```
jimmynotjim: Do you need to set left/right margin to `0` here?

cfgov/unprocessed/css/cf-enhancements.less
```diff
@@ -1127,6 +1127,218 @@ textarea {
...
+        line-height: normal;
```
jimmynotjim: Why set this? Is there something cf-icons is forcing that we should reconsider?

anselmbradford:
@Scotchester Thanks for looking! Some answers to your Qs…

> If the thing the input is containing might be a button with a label, I'm not wild about the name input-contains-icon. Perhaps we could genericize it to something like input-containing-stuff? That seems only marginally better... this is a tough one.

How about `input-contains-label` or `input-contains-annotation`? The added "thing" in the input for both the before and after can be any of the following:
 - An icon
 - A label
 - An icon and a label
 - Any of the above wrapped in a button*

_* Actually, I need to add a fix for a button in the before._ 

> Here's a wacky idea. What if we had a u-abs-pos utility class? Then we could just replace with input-contains-icon with that generic class, e.g.:

Doesn't `position: absolute` rarely appear without `left`, `top`, `bottom`, or `right`, in which case a utility class would just split the position across two locations?

> I think I would prefer to place the _after item after the input in the markup. Not a dealbreaker for me, though.

I couldn't get it to work without the input last because the `&_after ~ input` selector is order sensitive.

cfgov/unprocessed/css/cf-enhancements.less
anselmbradford: We haven't that I can recall. I'll ask the designers.

cfgov/jinja2/v1/browse-basic/index.html
anselmbradford: It's needed if the spans are wrapped in a button (which I need to push a fix for). I like the idea of removing the spans. Do you see a way to support both these scenarios with a reduced number of spans?:
```
.div
   .cf-icon
   .label
```

And:

```
div
   .btn
      .cf-icon
      .label
```

Scotchester:
> > Here's a wacky idea. What if we had a u-abs-pos utility class? Then we could just replace with input-contains-icon with that generic class, e.g.:

> Doesn't `position: absolute` rarely appear without `left`, `top`, `bottom`, or `right`, in which case a utility class would just split the position across two locations?

Oops, I misstated the idea. I meant `u-pos-rel`, for setting the context for the absolutely-positioned children.

Scotchester:
> > I think I would prefer to place the _after item after the input in the markup. Not a dealbreaker for me, though.

> I couldn't get it to work without the input last because the `&_after ~ input` selector is order sensitive.

Why not just reorder the selector?

anselmbradford:
> Why not just reorder the selector?

Wouldn't that apply the style to the `&_after` instead of the `input` if it was reordered? The padding needs to be applied to the input.

anselmbradford:
> Oops, I misstated the idea. I meant u-pos-rel, for setting the context for the absolutely-positioned children.

Wouldn't that be kind of counter to BEM, like `input-contains-icon_before` and `input-contains-icon_after` and named such because they're within the context of `input-contains-icon`?

Scotchester:
> Wouldn't that apply the style to the &_after instead of the input if it was reordered? The padding needs to be applied to the input.

Oh... right :flushed:

Scotchester:
> Wouldn't that be kind of counter to BEM, like input-contains-icon_before and input-contains-icon_after are named such because they're within the context of input-contains-icon?

Well, if we went through with that idea, I'd rename those accordingly. But I don't have good ideas for that, either.

This thing is freakin' hard to name well...

> How about `input-contains-label` or `input-contains-annotation`? The added "thing" in the input for both the before and after can be any of the following:
> - An icon
> - A label
> - An icon and a label
> - Any of the above wrapped in a button*

Both of those don't really capture the button use case, though. How about `input-contains-thing`?

anselmbradford:
@Scotchester @sebworks @jimmynotjim 
Maybe I should make it `input-contains-btn` and make everything a btn. By default a click of the button would highlight the input (faked through `pointer-events: none;` on the button by default), but JS could be added to provide another function (such as clearing the input value). That means I could get rid of the spans in https://github.com/cfpb/cfgov-refresh/pull/1330#discussion_r49515606 and wouldn't need to handle having spans and/or divs in the before/after bits. What do you think?

anselmbradford:
Or, another idea: `input-contains-label` and the structure is:

```
.input-contains-btn
    label.input-contains-btn_before
        &:before - icon
        :content - label
    label.input-contains-btn_after
        &:before - icon
        :content - label
    input
```

Labels act kind of like buttons because of their `for` attribute can be used to make a label click interact with an input. I kind of like this direction vs button, because that's what this PR is about. Having an icon that either when clicked focuses the input or does something else with the input (clears it). By default the labels would have a `for` attribute assigned to the input, but JS can override the behavior to do something other than focusing. This would remove all the need to include and adjust inclusion of cf-buttons.

anselmbradford:
Pushed up a version that uses labels instead. Let me know what you think!

screen shot 2016-01-12 at 7 24 04 pm

screen shot 2016-01-12 at 7 24 12 pm

screen shot 2016-01-12 at 7 23 51 pm

sebworks:
@anselmbradford, I like the label approach as long as it doesn't decrease the accessibility.

sebworks:
I'm also on board with `input-contains-icon` which I feel provides a better description of the purpose of the markup.

anselmbradford:
> I'm also on board with input-contains-icon which I feel provides a better description of the purpose of the markup.

@sebworks, do you mean `input-contains-label`? `-icon` is the original name. And oops, just realized I had updated the name wrong. Fixed!

sebworks:
Yes, I meant `input-contains-label`. Updated the comment.

anselmbradford:
>  I like the label approach as long as it doesn't decrease the accessibility.

Hm, so if it's just a label what happens during tabbing through fields is that if a label gets focus it passes it on to its associated input. This is fine if the labels only focus the input. If they're meant to clear the field (e.g.) then the accessibility will need to be handled on the JS-side. Perhaps a hidden button is added to the DOM purely for accessibility access (like the skip to content button). Perhaps focus can be forced on the label when the tab hits it. I'll have to do some testing later.

sebworks:
@anselmbradford , :100:  looks good; you might choose to address the accessibility in a separate PR.

Scotchester:
I'm not a huge fan of `-label`, either, since the clear button isn't really a label :\

But I don't think there's really a good answer out there. So I guess `-label` works as well as anything. :+1:

cfgov/unprocessed/css/cf-enhancements.less
anselmbradford: Removed in https://github.com/cfpb/cfgov-refresh/commit/91fbae6147b56362888b13d8c47d35d5d3284761

cfgov/unprocessed/css/cf-enhancements.less
anselmbradford: Removed in https://github.com/cfpb/cfgov-refresh/commit/91fbae6147b56362888b13d8c47d35d5d3284761 

Updated order elsewhere in https://github.com/cfpb/cfgov-refresh/commit/c533e5a706c270da2a48a2eec1388576b7067aef

cfgov/jinja2/v1/browse-basic/index.html
anselmbradford: Updated in https://github.com/cfpb/cfgov-refresh/commit/91fbae6147b56362888b13d8c47d35d5d3284761

cfgov/unprocessed/css/cf-enhancements.less
anselmbradford: Sounds like we're keeping it x without the circular background.

cfgov/unprocessed/css/cf-enhancements.less
anselmbradford: Removed. Thanks!

cfgov/unprocessed/css/cf-enhancements.less
anselmbradford: Removed in https://github.com/cfpb/cfgov-refresh/commit/91fbae6147b56362888b13d8c47d35d5d3284761

cfgov/unprocessed/css/cf-enhancements.less
anselmbradford: Removed in https://github.com/cfpb/cfgov-refresh/commit/91fbae6147b56362888b13d8c47d35d5d3284761

anselmbradford:
Squashing and merging

jimmynotjim:
:+1:

PR #1333, opened 2016-01-12
fixed prototype data for full width text

kave:
Forgot to update full width text prototype data.

@jimmynotjim


jimmynotjim:
LTGM

PR #1338, opened 2016-01-13
Create AbstractLearnPage for Learn and Doc Detail pages to inherit from

kurtw:
- Adds relevant date fields for AbstractLearnPage
 - `Date Published`, `Date Filed`, `Comments close by`
- Adds preview fields

screen shot 2016-01-14 at 11 13 02 am

@rosskarchner 
@kave


kave:
When users click create a page, does `AbstractLearnPage` show up in that list? If so we should hide that, if not  :+1:

kurtw:
Ready for another review

cfgov/v1/models/learn_page.py
```diff
@@ -1,47 +1,89 @@
...
+    preview_link_text = models.CharField(max_length=255, blank=True)
+    preview_image = models.ForeignKey(
+        'v1.CFGOVImage',
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name='+'
+    )
+    date_published = models.DateField()
+    date_filed = models.DateField(blank=True)
```
kave: need to include `null=True` if we're allowing blanks

cfgov/v1/models/doc_detail_page.py
```diff
@@ -1,22 +1,11 @@
...
+from .learn_page import AbstractLearnPage
 
 
-class DocumentDetailPage(CFGOVPage):
-    header = StreamField([
-        ('text_introduction', molecules.TextIntroduction()),
-        ('item_introduction', organisms.ItemIntroduction()),
-    ], blank=True)
-
+class DocumentDetailPage(AbstractLearnPage):
```
kave: lets actually move this model to the learn_page.py seems unnecessary to have them separated now since their referring to doc details as learn pages

kurtw: Cool.

kurtw:
Updated.

kave:
:+1:

PR #1341, opened 2016-01-14
Bugfix: Updates env name

anselmbradford:
Name was updated in environment, but not in initial-data.sh.
@KimberlyMunoz 
@rosskarchner


rosskarchner:
:+1:

PR #1342, opened 2016-01-14
Adds Global Search, initial Submit a complaint CTA, & Header

anselmbradford:
Adds Global Search, initial Submit a complaint CTA, & Header

## Additions

- Atomic Global Search molecule.
- Initial atomic Submit a Complaint CTA.
- Initial atomic Header.

## Changes

- Moves initial styling from non-atomic to atomic header.

## Testing

- Visit the site, click on search, enter text, resize window.

## Review

- @sebworks 
- @jimmynotjim 
- @KimberlyMunoz 

## Screenshots

Desktop size, contracted:
screen shot 2016-01-14 at 3 56 29 pm

Desktop size, expanded:
screen shot 2016-01-14 at 3 56 04 pm

Tablet size, contracted:
screen shot 2016-01-14 at 3 56 40 pm

Tablet size, shows button text:
screen shot 2016-01-14 at 3 55 45 pm

Tablet size, full suggestions width:
screen shot 2016-01-14 at 3 55 53 pm

Mobile size, contracted:
screen shot 2016-01-14 at 3 56 47 pm

Mobile size, smallest size, expanded:
screen shot 2016-01-14 at 3 55 37 pm

Mobile size, expanded:
screen shot 2016-01-14 at 3 55 28 pm

Entering text:
screen shot 2016-01-14 at 3 56 21 pm

## Todos

- **There's a bug in cf-forms where the button and input heights are different (https://github.com/cfpb/capital-framework/issues/261), which shows up in this PR.**
- Submit a complaint CTA needs to be incorporated into the Mega Menu mobile nav.
- It will likely make sense to make the button and input attached by default and a gap should be added to separate them through a modifier. This is currently set directly in the Global Search molecule, but should be set in cf-enhancements/cf-forms.

### DIFF
--- a/cfgov/jinja2/v1/_includes/templates/header.html
+++ b/cfgov/jinja2/v1/_includes/templates/header.html
@@ -13,7 +13,8 @@
     {% set value = global_dict.global_eyebrow %}
     {% include 'molecules/global-eyebrow.html' with context %}
 
-    <div class="wrapper">
+    <div class="o-header_content
+                wrapper">
         <button class="primary-nav_trigger js-primary-nav_trigger">
             <span class="u-visually-hidden">Menu</span>
         </button>
@@ -23,16 +24,23 @@
                 'global_header_cta': {
                     'is_in_list': false
                 }
-
             })
         %}
         {% set value = global_dict.global_header_cta %}
         {% include 'molecules/global-header-cta.html' with context %}
 
-        <a href="/">
-            <img class="header_logo u-js-only"
-                 src="/static/img/logo_sm-exec.svg"
-                 onerror="this.onerror=null;this.src='/static/img/logo_sm-exec.png';"
+    {% include 'molecules/global-search.html' with context %}
+
+    <a href="/">
+        <img class="o-header_logo header_logo u-js-only"
+             src="/static/img/logo_sm-exec.svg"
+             onerror="this.onerror=null;this.src='/static/img/logo_sm-exec.png';"
+             alt="Consumer Financial Protection Bureau"
+             width="237"
+             height="50">
+        <noscript>
+            <img class="header_logo logo__no-js"
+                 src="/static/img/logo_sm-exec.png"
                  alt="Consumer Financial Protection Bureau"
                  width="237"
                  height="50">
@@ -50,3 +58,4 @@
         {% endblock primary_nav %}
     </div>
 </header>
+<div class="content_bar"></div>
--- a/cfgov/unprocessed/css/header.less
+++ b/cfgov/unprocessed/css/header.less
@@ -42,7 +42,7 @@
     vertical-align: middle; // removes typical inline vertical whitespace
 
     .respond-to-min(@bp-med-min, {
-        margin: @grid_gutter-width 0 unit(20px / @base-font-size-px, em) 0;
+        margin: 0 0 unit(20px / @base-font-size-px, em) 0;
         height: 50px;
         width: 237px;
     });
--- a/cfgov/unprocessed/css/molecules/global-search.less
+++ b/cfgov/unprocessed/css/molecules/global-search.less
@@ -1,0 +1,287 @@
+
+// TODO: Move the theme variables to cf-enhancements.
+/* topdoc
+  name: Theme variables
+  family: cf-core
+  notes:
+    - "The following color and sizing variables are exposed, allowing you to
+       easily override them before compiling."
+  patterns:
+    - name: Sizing
+      codenotes:
+        - |
+          @margin__em
+          @margin_half__em
+  tags:
+    - cf-core
+*/
+
+@margin__em: unit( @grid_gutter-width / @base-font-size-px, em );
+@margin_half__em: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
+
+/* topdoc
+  name: Global Search molecule
+  family: cfgov-molecules
+  patterns:
+    - name: Default example
+      markup: |
+          <div class="m-global-search">
+              <button class="m-global-search_trigger">
+                  <span class="cf-icon"></span>
+                  <span class="m-global-search_trigger-label">
+                      <span class="u-visually-hidden">Search</span>
+                  </span>
+              </button>
+              <div class="m-global-search_content" aria-expanded="false">
+                  <form class="m-global-search_content-form">
+                      <div class="input-with-btn">
+                          <div class="input-with-btn_input
+                                      input-contains-label">
+                              <label for="search-input"
+                                     class="input-contains-label_before
+                                            input-contains-label_before__search">
+                              </label>
+                              <label for="search-input"
+                                     class="input-contains-label_after
+                                            input-contains-label_after__clear
+                                            u-hidden">
+                              </label>
+                              <input type="text"
+                                     id="search-input"
+                                     value=""
+                                     placeholder="Search the CFPB">
+                          </div>
+                          <div class="input-with-btn_btn">
+                              <button class="btn">Search</button>
+                          </div>
+                      </div>
+
+                      <div class="m-global-search_content-suggestions">
+                          <p class="h5">Suggested search terms:</p>
+                          <ul class="list list__horizontal">
+                              <li class="list_item"><a class="list_link" href="#">Regulations</a></li>
+                              <li class="list_item"><a class="list_link" href="#">Compliance guides</a></li>
+                              <li class="list_item"><a class="list_link" href="#">Mortgage</a></li>
+                              <li class="list_item"><a class="list_link" href="#">College loans</a></li>
+                          </ul>
+                      </div>
+                  </form>
+              </div>
+          </div>
+      codenotes:
+        - |
+          Structural cheat sheet:
+          -----------------------
+          .m-global-search
+            button.m-global-search_trigger
+              span.cf-icon
+              span.m-global-search_trigger-label
+                span.u-visually-hidden
+            .m-global-search_content
+              form.m-global-search_content-form
+                […
+                .input-with-btn
+                .input-contains-label
+                from cf-forms
+                …]
+                .m-global-search_content-suggestions
+                  .h5
+                  […
+                  .list.list__horizontal
+                  from cf-typography
+                  …]
+  tags:
+    - cfgov-molecules
+*/
+
+.m-global-search {
+
+    // Absolute position when there is a flyout at mobile/tablet sizes.
+    .respond-to-max( @bp-sm-max, {
+        // Make the height large so that the overflowing flyout can be hidden,
+        // but not get cropped when it's visible.
+        height: 500px;
+        width: 100%;
+        position: absolute;
+        top: 0;
+        overflow-x: hidden;
+        // TODO: This will be ignored in IE8-10,
+        //       leading to the overflow preventing clicks on the links.
+        pointer-events: none;
+    } );
+
+    &_trigger {
+        height: unit( 45px / 18px, em );
+        min-width: unit( 45px / 18px, em );
+        padding-left: unit( @grid_gutter-width / 2 / 18px, em );
+        padding-right: unit( @grid_gutter-width / 2 / 18px, em );
+
+        float: right;
+        position: relative;
+        z-index: 20;
+
+        background: @white;
+        border: none;
+        color: @black;
+        font-size: unit( 18px / @base-font-size-px, em );
+        pointer-events: auto;
+
+        &:focus {
+            outline: 1px dotted @black;
+        }
+
+        .cf-icon:before {
+            content: @cf-icon-search;
+            font-size: unit( 20px / 18px, em );
+        }
+
+        // Show "Search" text in trigger at tablet sizes.
+        .respond-to-min( @bp-sm-min {
+            &-label:before {
+                .webfont-medium();
+                content: 'Search';
+            }
+        } );
+
+        // Only show flyout and close trigger at tablet/mobile sizes.
+        .respond-to-max( @bp-sm-max, {
+            height: unit( 60px / 18px, em );
+            min-width: unit( 60px / 18px, em );
+            padding-top: unit( @grid_gutter-width / 2 / 18px, em );
+            padding-bottom: unit( @grid_gutter-width / 2 / 18px, em );
+
+            &[aria-expanded="true"] {
+                background: @gray-10;
+                border: 1px solid @gray-50;
+                border-bottom: none;
+                color: @black;
+
+                .cf-icon:before {
+                    content: @cf-icon-delete;
+                }
+
+                // Cover up the border edge of the flyout so that it is attached
+                // visually to the trigger button.
+                &:after {
+                    display: block;
+                    height: 1px;
+                    width: 100%;
+
+                    position: absolute;
+                    left: 0;
+                    bottom: -1px;
+
+                    content: '\a0';
+                    background: @gray-10;
+                }
+            }
+        } );
+
+        // Show "Close" text in trigger at tablet sizes.
+        .respond-to-range( @bp-sm-min, @bp-med-min, {
+            &[aria-expanded="true"] {
+                .m-global-search_trigger-label:before {
+                    content: 'Close';
+                }
+            }
+        } );
+
+        .respond-to-min( @bp-med-min {
+            &[aria-expanded="true"] {
+                display: none;
+            }
+        } );
+    }
+
+    &_content {
+        .respond-to-min( @bp-med-min {
+            display: none;
+        } );
+
+        padding-top: unit( @grid_gutter-width / 4 / @base-font-size-px, em );
+
+        &[aria-expanded="true"] {
+            display: block;
+
+            transform: translateX(0);
+        }
+
+        // Create flyout at table/mobile size.
+        .respond-to-max( @bp-sm-max, {
+            box-sizing: border-box;
+            width: 100%;
+
+            position: absolute;
+            left: 0;
+            // Offset for height of trigger button.
+            top: 60px;
+            z-index: 10;
+
+            background-color: @gray-10;
+            border: 1px solid @gray-50;
+            pointer-events: auto;
+
+            transform: translateX(100%);
+            transition: transform 0.25s ease-out;
+
+            &-form {
+                padding-top: @margin__em;
+                padding-left: @margin_half__em;
+                padding-right: @margin_half__em;
+                padding-bottom: @margin_half__em;
+            }
+
+            // Drop-shadow under the search flyout menu at mobile/tablet size.
+            &:after {
+                position: relative;
+                top: 5px;
+                content: '\a0';
+                display: block;
+                height: 5px;
+                width: 100%;
+                opacity: 0.2;
+                background: @gray;
+            }
+        } );
+
+        &-suggestions {
+            display: none;
+
+            // Only show search suggestions at tablet size.
+            .respond-to-range( @bp-sm-min, @bp-sm-max, {
+                & {
+                    display: block;
+                }
+            } );
+
+            .h5,
+            .list {
+                display: inline-block;
+            }
+        }
+    }
+
+    // TODO: Move these styles to cf-enhancements/cf-forms.
+    .respond-to-min( 480px, {
+        &_content-form {
+            // Attach button to input.
+            .input-with-btn_input {
+                width: 70%;
+            }
+            .input-with-btn_btn {
+                width: 30%;
+                border-left: 0;
+
+                button {
+                    border-top-left-radius: 0;
+                    border-bottom-left-radius: 0;
+                }
+            }
+        }
+    } );
+}
+
+/* topdoc
+    name: EOF
+    eof: true
+*/
--- a/cfgov/unprocessed/js/molecules/GlobalSearch.js
+++ b/cfgov/unprocessed/js/molecules/GlobalSearch.js
@@ -1,0 +1,219 @@
+'use strict';
+
+// Required polyfills for <IE9.
+require( '../modules/polyfill/query-selector' );
+require( '../modules/polyfill/event-listener' );
+require( '../modules/polyfill/class-list' );
+
+// Required modules.
+var atomicCheckers = require( '../modules/util/atomic-checkers' );
+var breakpointState = require( '../modules/util/breakpoint-state' );
+
+/**
+ * GlobalSearch
+ * @class
+ *
+ * @classdesc Initializes a new GlobalSearch molecule.
+ *
+ * @param {HTMLNode} element
+ *   The DOM element within which to search for the molecule.
+ * @returns {Object} An GlobalSearch instance.
+ */
+function GlobalSearch( element ) {
+
+  var BASE_CLASS = 'm-global-search';
+
+  var _dom =
+    atomicCheckers.validateDomElement( element, BASE_CLASS, 'GlobalSearch' );
+  var _triggerDom = _dom.querySelector( '.' + BASE_CLASS + '_trigger' );
+  var _contentDom = _dom.querySelector( '.' + BASE_CLASS + '_content' );
+  var _searchInputDom;
+  var _clearBtnDom;
+
+  var _isExpanded = false;
+  var _tabClicked = false;
+
+  var KEY_TAB = 9;
+
+  /**
+   * @returns {Object} The GlobalSearch instance.
+   */
+  function init() {
+    var inputSelector = '.' + BASE_CLASS + '_content-form input';
+    var clearBtnSelector =
+      '.' + BASE_CLASS + ' .input-contains-label_after__clear';
+    var searchIconSelector =
+      '.' + BASE_CLASS + ' .input-contains-label_before__search';
+    var searchBtnSelector = '.' + BASE_CLASS + ' .input-with-btn_btn button';
+
+    _searchInputDom = _contentDom.querySelector( inputSelector );
+    var searchIconDom = _contentDom.querySelector( searchIconSelector );
+    var searchBtnDom = _contentDom.querySelector( searchBtnSelector );
+    _clearBtnDom = _contentDom.querySelector( clearBtnSelector );
+
+    _triggerDom.addEventListener( 'click', _triggerClicked );
+    _clearBtnDom.addEventListener( 'mousedown', _clearClicked );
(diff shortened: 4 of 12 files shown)


cfgov/unprocessed/css/molecules/submit-complaint-cta.less
```diff
@@ -0,0 +1,46 @@
...
+          Structural cheat sheet:
+          -----------------------
+          .m-submit-complaint-cta
+            a
+  tags:
+    - cfgov-molecules
+*/
+
+.m-submit-complaint-cta {
+    float: right;
```
anselmbradford: Moved to header in https://github.com/cfpb/cfgov-refresh/pull/1347

cfgov/jinja2/v1/_includes/molecules/submit-complaint-cta.html
```diff
@@ -0,0 +1,15 @@
+{# ==========================================================================
+
+   Submit A Complaint Call to Action (CTA)
```
jimmynotjim: Should this be "Heading CTA" rather than Submit a Complaint specifically? It'd make it better for reuse in the future.

cfgov/unprocessed/css/header.less
```diff
@@ -51,10 +51,11 @@
     width: 190px;
     height: 40px;
     margin: 10px;
+    margin-top: 0;
     vertical-align: middle; // removes typical inline vertical whitespace
 
     .respond-to-min(@bp-med-min, {
-        margin: @grid_gutter-width 0 unit(20px / @base-font-size-px, em) 0;
+        margin: 0 0 unit(20px / @base-font-size-px, em) 0;
```
jimmynotjim: Do we need to reset top and left/right margin?

cfgov/unprocessed/css/header.less
```diff
@@ -51,10 +51,11 @@
     width: 190px;
     height: 40px;
     margin: 10px;
+    margin-top: 0;
```
jimmynotjim: This could be updated to `margin: 0 10px 10px;` to avoid resetting the top margin.

cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -0,0 +1,226 @@
+/* topdoc
+  name: Global Search molecule
+  family: cfgov-molecules
+  patterns:
+    - name: Default example
+      markup: |
+          <div class="m-global-search">
+              <p class="short-desc">
```
jimmynotjim: This markup doesn't match the markup that's used.

cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -0,0 +1,226 @@
...
+    // TODO: Move these positioning styles to the header organism.
+    .respond-to-min( @bp-med-min, {
+        float: right;
+        &_content {
+            padding-right: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
+        }
+    } );
+
+    &_trigger {
+        height: unit(60px / 24px, em);
```
jimmynotjim: Denominator should be updated to 18 because to match the `font-size` set on L:58

Here, L:50, L:51, and L:52

cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -0,0 +1,226 @@
...
+        padding-left: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
+        padding-right: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
+        position: relative;
+        background: @white;
+        border: none;
+        color: @black;
+        float: right;
+        font-size: unit( 18px / @base-font-size-px, em );
+        pointer-events: auto;
+        z-index: 20;
```
jimmynotjim: Update property order to match FEWD standards here and throughout.

cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -0,0 +1,226 @@
...
+
+        &:focus {
+            outline: 1px dotted @black;
+        }
+
+        .cf-icon:before {
+            content: @cf-icon-search;
+            font-size: unit( 20px / @base-font-size-px, em );
```
jimmynotjim: Because the parent `font-size` has been set to `18px`, the denominator here should be updated as well.

cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -0,0 +1,226 @@
...
+
+            // Drop-shadow under the search flyout menu at mobile/tablet size.
+            &:after {
+                position: relative;
+                top: 5px;
+                content: '\a0';
+                display: block;
+                height: 5px;
+                width: 100%;
+                background: rgba( 90, 93, 97, 0.2 );
```
jimmynotjim: Should this have an alpha channel? I thought the Design Manual specifically calls out not using transparent shadows. cc @ajbush @duelj

cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -0,0 +1,226 @@
...
+                }
+            } );
+
+            .h5,
+            .list {
+                display: inline-block;
+            }
+        }
+
+        &[aria-expanded="true"] {
```
jimmynotjim: Can this be combined with L:128?

cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -0,0 +1,226 @@
...
+            transform: translateY(0);
+            transition: transform 0.25s ease-out;
+            visibility: visible;
```
jimmynotjim: Visibility doesn't look like it's been changed, is this still needed?

cfgov/unprocessed/css/molecules/submit-complaint-cta.less
```diff
@@ -0,0 +1,46 @@
...
+    padding-left: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
+    padding-top: unit( @grid_gutter-width / 3 / @base-font-size-px, em );
+    padding-bottom: unit( @grid_gutter-width / 3 / @base-font-size-px, em );
+
+    a {
+        .webfont-medium();
+        color: @pacific;
+    }
+
+    a:before {
```
jimmynotjim: Move inside the `a` and use `&:before`

cfgov/jinja2/v1/_includes/molecules/submit-complaint-cta.html
anselmbradford: I could get on board with that. The docs would need an update https://github.cfpb.gov/flapjack/Modules-V1/wiki/Submit-a-complaint-CTA
@duelj @schaferjh ?

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: These values come from @duelj , but yes it would be good to confirm that this can't be a semi-transparent black to simplify the values, for example.

cfgov/unprocessed/css/header.less
anselmbradford: Let's come back to this. These header styles need to migrate over to `/organisms/header.less`, which will happen in another PR. This was updated here to remove the spacing that is added by the atomic header.

cfgov/unprocessed/css/header.less
anselmbradford: I'll circle back to this in another PR for the atomic header. Remind me if I forget!

cfgov/unprocessed/css/molecules/submit-complaint-cta.less
anselmbradford: Updated in Submit a Complaint CTA PR https://github.com/cfpb/cfgov-refresh/commit/a5b2cc149dca43fa44e78ca9c5ffe101445a6dc1

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/commit/c21458b122409b41d50e3198afeff28f849777ae

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: Updated in https://github.com/cfpb/cfgov-refresh/commit/7a707698e434c7faf32863e090ccb9078efd7000

cfgov/unprocessed/js/molecules/GlobalSearch.js
```diff
@@ -0,0 +1,171 @@
...
+    _setClearBtnState( _searchInputDom.value );
+
+    return this;
+  }
+
+  /**
+   * Event handler for when the search icon is clicked in the
+   * expanded state at desktop sizes. Closes the search box.
+   */
+  function _searchLabelClicked() {
```
jimmynotjim: Suggest making this trigger during a blur event on the input instead of specific clicks.

cfgov/unprocessed/js/molecules/GlobalSearch.js
```diff
@@ -0,0 +1,171 @@
...
+      collapse();
+    }
+  }
+
+  /**
+   * Event handler for when the search input trigger is clicked,
+   * which opens/closes the search input.
+   */
+  function _triggerClicked() {
+    _isExpanded = !_isExpanded;
```
jimmynotjim: This seems counter intuitive. Why not just do 

```
if ( !_isExpanded ) {
  expand();
} else {
  collapse();
}
```

cfgov/unprocessed/js/molecules/GlobalSearch.js
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * Open the search box.
+   * @returns {Object} An GlobalSearch instance.
+   */
+  function expand() {
+    _isExpanded = true;
+    _triggerDom.setAttribute( 'aria-expanded', 'true' );
+    _contentDom.setAttribute( 'aria-expanded', 'true' );
+    _searchInputDom.focus();
+    _searchInputDom.select();
```
jimmynotjim: `.select()` focuses an elem, so you shouldn't need both.

cfgov/unprocessed/js/molecules/GlobalSearch.js
```diff
@@ -0,0 +1,171 @@
...
+   */
+  function _hideClearBtn() {
+    _clearBtnDom.classList.add( 'u-hidden' );
+  }
+
+  /**
+   * Remove a hidden class from the input search label.
+   * Used when there is text input.
+   */
+  function _showClearBtn() {
```
jimmynotjim: Probably out of scope, but we could abstract these two as a `toggleVisibility` utility so that we're not repeating how we hide/show content throughout the project.

cfgov/unprocessed/js/molecules/GlobalSearch.js
anselmbradford: This selects the text in the input after focusing.

cfgov/unprocessed/js/molecules/GlobalSearch.js
anselmbradford: Yup, you're right! Fixed in https://github.com/cfpb/cfgov-refresh/commit/488849f5914deb228b74d9b6bc7274d905349571

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: It's set on https://github.com/cfpb/cfgov-refresh/pull/1342/files#diff-1e0592ad7470228d80cfa46d6aff3c62R207

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: Done! https://github.com/cfpb/cfgov-refresh/commit/27e925f06baf167f0f5ad0c6ab1e9f2b5310fdc1

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/commit/bb78767c0de58635727a2bba849086c88aa055f0

Sorry, what were L:68 and L:82? My line numbers have moved around.

anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/commit/62c72a886ebebaf94336e95873387f7b90c77a8a

cfgov/jinja2/v1/_includes/molecules/submit-complaint-cta.html
schaferjh: How about "Global header CTA"? (Since "heading" may seem too inexact?) Happy to make the change if that sounds good.

cfgov/unprocessed/js/molecules/GlobalSearch.js
jimmynotjim: According to MDN you don't need to focus first https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/select

anselmbradford: Ahh, ha. I misunderstood and thought you were saying the `focus()` would select the text. Fixed in https://github.com/cfpb/cfgov-refresh/commit/4ff9279128485d9d686ed1b9d4456295cced6c62

cfgov/unprocessed/js/molecules/GlobalSearch.js
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/commit/4ff9279128485d9d686ed1b9d4456295cced6c62

cfgov/jinja2/v1/_includes/molecules/submit-complaint-cta.html
anselmbradford: That works for me. @jimmynotjim ?

cfgov/unprocessed/js/molecules/GlobalSearch.js
anselmbradford: Yeah that's come back to this. I definitely see abstracting out parts of expand/collapse and showing/hiding to share with e.g. expandables and notifications.

anselmbradford:
Ready for a review again. The CTA name can be handled in https://github.com/cfpb/cfgov-refresh/pull/1347

sebworks:
The button text is `24px` at tablet sizes. Is this correct?
screen shot 2016-01-20 at 7 50 20 am

sebworks:
Tabbing through the elements with nav open causes a page reload.

sebworks:
Should the search box be shown when JS is disabled?

cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -0,0 +1,272 @@
...
+            background-color: @gray-10;
+            border: 1px solid @gray-50;
+            pointer-events: auto;
+            transform: translateX(100%);
+            transition: transform 0.2s ease-in,
+                        visibility 0s 0.5s;
+            visibility: hidden;
+
+            &-form {
+                padding-top: unit( @grid_gutter-width / @base-font-size-px, em );
```
sebworks: Is it worth it to make  `unit( @grid_gutter-width / @base-font-size-px, em )` a variable?

cfgov/unprocessed/js/molecules/GlobalSearch.js
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  /**
+   * @returns {boolean} Whether the search box is expanded or not.
+   *   This can be used to determine if the box is open
+   *   and needs to be closed if an overlapping dialog box is opened.
+   */
+  function isExpanded() {
+    return _isExpanded;
+  }
```
sebworks: If you need the box to be collapsed wouldn't you just call `collaspe`? Is the check necessary?

sebworks:
Should we use a css transition to show/hide the box?

cfgov/jinja2/v1/_includes/molecules/submit-complaint-cta.html
jimmynotjim: Sounds great! 
cg01025

anselmbradford: Updated in https://github.com/cfpb/cfgov-refresh/pull/1347

duelj:
@anselmbradford - in regard to the [
screen shot 2016-01-20 at 1 21 26 pm
search button text size](https://github.com/cfpb/cfgov-refresh/pull/1342#issuecomment-173195094) it should be 16px Avenir Next medium and the minicon should match the text size.

duelj:
@jimmynotjim and @anselmbradford - I think for the menu we want to keep the transparent stroke shadow. As long as it doesn't have feathered edges and is a transparent rectangle, I think we will be OK on it's use. As for the color, it would be nice to keep it our grey swatch but if it simplifies things to go black I don't see that being an issue since most people probably wouldn't be able to tell the difference. 

Wanted to see if @ajbush was on board with my thoughts here.

anselmbradford:
> The button text is 24px at tablet sizes. Is this correct?

Nope. Fixed in https://github.com/cfpb/cfgov-refresh/commit/be96ecafbf58b3923f033f8a2b65e88de82b43b7

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: Updated to use `@gray` and `opacity` rule in https://github.com/cfpb/cfgov-refresh/commit/6a126448e3207a204c075bc77fb1948ac2943b83

cfgov/unprocessed/js/molecules/GlobalSearch.js
anselmbradford: That's a good point. I added this in anticipation that the mega menu would need to close the search menu, but you're right it doesn't need to check first. Those checks can and should happen in the expand/collapse methods. Fixed in https://github.com/cfpb/cfgov-refresh/commit/8663c30f65b433a2ccd0703c1f05fc4e9e5796f1

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: Added in https://github.com/cfpb/cfgov-refresh/commit/dfc1db14aa7a825b0b5b39f8cba24c06d761a2a5

anselmbradford:
> Tabbing through the elements with nav open causes a page reload.

Great find! This is kind of complex to fix, but ended up tracking when the tab key is pressed and not collapsing the input when it is pressed and also forcing a click on the search button on mousedown so it fires before the focus event on the input fires. Fixed in https://github.com/cfpb/cfgov-refresh/commit/3157fd75fb9b397c8d4b22f2ae79827e4510414d

sebworks:
:+1:

cfgov/unprocessed/css/molecules/global-search.less
jimmynotjim: highfive

cfgov/jinja2/v1/_includes/templates/header.html
```diff
@@ -18,8 +18,10 @@
     {% set value = global_dict.global_header_cta %}
     {% include 'molecules/global-header-cta.html' with context %}
 
+    {% include 'molecules/global-search.html' with context %}
+
     <a href="/">
-        <img class="header_logo u-js-only"
+        <img class="o-header_logo header_logo u-js-only"
```
jimmynotjim: Can the old `header_logo` be removed now?

cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -0,0 +1,288 @@
...
+      codenotes:
+        - |
+          @margin__em
+          @margin_half__em
+  tags:
+    - cf-core
+*/
+
+@margin__em: unit( @grid_gutter-width / @base-font-size-px, em );
+@margin_half__em: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
```
jimmynotjim: When these get moved, we'll have to be careful using them because if the font size isn't 16px you're not going to get the expected output.

cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -0,0 +1,288 @@
...
+                .webfont-medium();
+                content: 'Search';
+            }
+        } );
+
+        // Only show flyout and close trigger at tablet/mobile sizes.
+        .respond-to-max( @bp-sm-max, {
+            height: unit( 60px / 18px, em );
+            min-width: unit( 60px / 18px, em );
+            padding-top: @margin_half__em;
```
jimmynotjim: This won't work here and L:151 because the font-size is 18

cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -0,0 +1,288 @@
...
+    &_content {
+        display: none;
+
+        padding-top: unit( @grid_gutter-width / 4 / @base-font-size-px, em );
+
+        &[aria-expanded="true"] {
+            display: block;
+
+            transform: translateY(0);
+            transition: transform 0.25s ease-out;
```
jimmynotjim: `transition` needs to live in the main `_content` block of styles or it won't animate when closing

cfgov/jinja2/v1/_includes/templates/header.html
anselmbradford: I'll remove it in a separate PR for the header organism.

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: Maybe the name should be something like `@base-space__em`, `@base-space_half__em`? That way it would also not imply it's for use only in a `margin` style.

@sebworks what do you think?

cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -0,0 +1,288 @@
...
+
+        // Create flyout at table/mobile size.
+        .respond-to-max( @bp-sm-max, {
+            box-sizing: border-box;
+            width: 100%;
+
+            position: absolute;
+            left: 0;
+            // Offset for height of trigger button.
+            top: 60px;
```
jimmynotjim: If you don't set any top value, shouldn't it align w/ the bottom of the button?

cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -0,0 +1,288 @@
...
+
+            .h5,
+            .list {
+                display: inline-block;
+            }
+        }
+    }
+
+    // TODO: Move these styles to cf-enhancements/cf-forms.
+    .respond-to-min( 480px, {
```
jimmynotjim: Can this be a breakpoint var instead of a px value?

cfgov/unprocessed/css/molecules/global-search.less
jimmynotjim: Even if it's em, if the font-size of the elem it's used in isn't 16px, the computed value will change from the expected 30 or 15px. ex.

```
.thing {
  font-size: unit( 18px / @base-font-size, em );
  margin-bottom: @base-space__em; // This will be computed as 1.875em * 18px = 33.75px
}

.parent-thing {
  @heading-5();

  .child-thing {
    margin-bottom: @base-space_half__em; // This will be computed as 0.9375em * 14px = 13.125px
}
```

sebworks: @jimmynotjim , what do you think about using:

`@margin__em: unit( @grid_gutter-width / 16px, em );`

jimmynotjim: That won't change anything. The output form unit will still be `1.875em`. If the font-size of the elem or a parent is set to something other than 16, the final calc in the browser will be incorrect. Just the world of EMs and why a lot of people prefer REMs.

If we want to be able to use these anywhere, we could convert them to `rem` instead of `em`, but we'd have to [do something else for IE](http://caniuse.com/#feat=rem). Possibly convert them to PX in the IE styles task?

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: It didn't appear we had a breakpoint at 480px, should one be added to CF?

jimmynotjim: Oh right, xs-max is 600, not 480, apologies.

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/commit/92570e377562e0b76c846ad23be7b7b18aea5b59 `unit( @grid_gutter-width / 2 / 18px, em );`

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/commit/913348a5ec0dd5bd35eaf0717bf76ab032067b50

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: Since it's absolutely positioned it aligns with the button's container:

screen shot 2016-01-22 at 12 24 57 pm

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: How should I leave this for this PR?

cfgov/unprocessed/css/molecules/global-search.less
jimmynotjim: huh, thanks for the screen shot

cfgov/unprocessed/css/molecules/global-search.less
jimmynotjim: Yep, I don't think there's anything to change, my only point was to be careful of situations like L:151 (that's now fixed)

jimmynotjim:
:+1:

anselmbradford:
> Should the search box be shown when JS is disabled?

Yes, I'll come back to this as the mega menu should be able to handle this too.

sebworks:
:+1:

PR #1343, opened 2016-01-14
Fixing the spacing around o-well module

KimberlyMunoz:
When the well is in a block, the spacing is off. This PR fixes the spacing without removing the block every time the well is being used. Fixes DF-677.

## Testing
- Check on landing page.

## Preview
image

### DIFF
--- a/cfgov/unprocessed/css/organisms/well.less
+++ b/cfgov/unprocessed/css/organisms/well.less
@@ -42,6 +42,7 @@
     padding-bottom: unit( ( @grid_gutter-width * 2 )  / @base-font-size-px, em );
     padding-left: unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em );
     border: 1px solid @o-well_border;
+    border-right: 0;
     background-color: @o-well_bg;
 
     .respond-to-min( @bp-sm-min, {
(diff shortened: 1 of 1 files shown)


anselmbradford:
Can this go in cf-enhancements for the block?

KimberlyMunoz:
I don't think it should, it's something related to the molecule. If anything we should stop wrapping everything in blocks and handle the spacing there instead of leaning on custom CF stuff that only we will use.

anselmbradford:
Hmm, really? I'm a bit confused on where we're going with this. I thought the atomic elements would eventually integrate into CF, and that to manage complexity we were encapsulating the UI in one organism/molecule per less file. `block` is analogous to an organism, so it seems like it would go in the block's CSS or in the page template's CSS. That way if/when block is updated its internal behavior can be handled in one place.

KimberlyMunoz:
I don't think it's been decided what the future of CF and atomic elements are and it makes sense to keep them somewhat separate for the time being.

Each molecule has workarounds and adjustments to account for the block spacing in the molecule itself. This is just like that but I separated it out for a scenario where maybe the well wouldn't be wrapped inside a block. I really don't like how we're tangling the molecules up with the block spacing and think it leads to extra code like this, so I'm trying to separate it out the best I can. I can delete the .block scoping and keep it so it assumes it's in a block at all times to bring it in line with all the other molecules, but I think that's going to cause headaches down the road.

sebworks:
Can we add a modifier on the well itself?

KimberlyMunoz:
I don't see how. This is only needed when it's wrapped in a `.block`.

KimberlyMunoz:
Maybe my attempt to detangle the workarounds for the `block` was silly. I'll just include them together like every other molecule and update our prototypes to match.

jimmynotjim:
Is there any way we can make wells use the `block__flush` like the Featured Content Module (see '/landing-page/` demo for ref). It seems that'd be a lot simpler, especially since these styles are duplicating the `__flush` styles.

KimberlyMunoz:
Talked with @jimmynotjim to talk with BEWDs about a potential backend solution for this, but otherwise we'll go with nesting `.block.block__flush` inside a `.block` like in the FCM.

jimmynotjim:
Discussed it with @kave and he thinks it might be possible to move the logic out of the template and styles. He'll update when he's sure. I'm going to close this PR for now, if his idea doesn't work, we'll reconvene.

PR #1349, opened 2016-01-15
error checking in the init data migration

kave:
@rosskarchner 
@jimmynotjim

### DIFF
--- a/cfgov/v1/migrations/9999_init_data.py
+++ b/cfgov/v1/migrations/9999_init_data.py
@@ -3,7 +3,6 @@ from __future__ import unicode_literals
 import os
 
 from django.contrib.auth.hashers import make_password
-from django.contrib.auth.models import User
 from django.db import migrations, models
 from django.conf import settings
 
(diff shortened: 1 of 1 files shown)


jimmynotjim:
It looks good, but I'd like @rosskarchner to sign off.

rosskarchner:
The logic looks fine-- but I'm wondering if high-numbered migrations work the way we're assuming they do?

If the numbers define simple order, then that's great. But if they are expected to be a *sequence*, then we might be shooting ourselves in the foot. For example, on a server where migration 9999 has already been run, and we deploy a new release that include hypothetical migration "1234", will it be ignored because it's less than 9999?

rosskarchner:
I guess we can test that by creating a migration smaller than 9999, running `migrate`, and seeing if it gets picked up?

kave:
@rosskarchner hmm yes i tested this scenario and your right... hmm i'll try to think of a solution

cfgov/v1/migrations/9999_init_data.py
```diff
@@ -15,24 +14,28 @@
...
 
         # Events Landing Page required for event `import-data` command
-        events = EventLandingPage(title='Events', slug='events')
-        site_root.add_child(instance=events)
-        revision = events.save_revision(
-            user=admin_user,
-            submitted_for_moderation=False,
-        )
-        revision.publish()
+        if not EventLandingPage.objects.filter(title='Events').exists():
```
kurtw: The `EventLandingPage` type is deprecated in the BrowseFilterablePage PR. It's a normal `BrowsePage` now.

kave:
Closing this PR in favor of https://github.com/cfpb/cfgov-refresh/pull/1357

PR #1351, opened 2016-01-19
Running legacy django and sheer apps within cfgov-refresh

rosskarchner:
By itself, this should leave your environment unharmed.

The interesting stuff is:

- some code for treating nemo (the wordpress theme) like a source of static files in Django (whist ignoring PHP files-- they won't get served while in DEBUG mode, and won't be gathered up by collectstatic)
- per the above, a middleware class for transforming references to the old nemo URL's, into /static/nemo
- some machinery for supporting "optional" django apps-- based on [this gist](https://gist.github.com/msabramo/945406)
- a function for conditionally enabling url's based on if an app is installed, and an entry in settings.LEGACY_APP_URLS for the app is true. This allows for situations where we want to continue using an apps data models or other functionality, but want those URL's to be handled by wagtail 

### Sheer sites now work!

For URL's and templates: 
- Once the 'sheerlike' app is ready, it goes through the entries in settings.SHEER_SITES, and [creates a distinct template engine](https://github.com/rosskarchner/cfgov-refresh/blob/one-with-everything/cfgov/sheerlike/apps.py#L10) .
- in urls.py, you can [instantiate a SheerSite object](https://github.com/rosskarchner/cfgov-refresh/blob/one-with-everything/cfgov/cfgov/urls.py#L50), which primarily exists to expose a 'urls' property, which can be passed to django's include() directive.
- a SheerSite also has a [urls_with_prefix](https://github.com/rosskarchner/cfgov-refresh/blob/one-with-everything/cfgov/cfgov/urls.py#L51) function, that lets you mount a subdirectory of a sheer site, to a URL. To achieve this today on www, we have to use URL rewriting-- this eliminates that administrative pain
- in either case. the URL's are built by walking the site directory, and building a SheerTemplateView for each HTML file (while turning /index.html URL's into just /'s). These views are bound to the template engine created earlier.

For static files:
- We add a new "[finder](https://github.com/rosskarchner/cfgov-refresh/blob/one-with-everything/cfgov/cfgov/settings/base.py#L159)" for static files
- The finder (and it's associated [storage](https://github.com/rosskarchner/cfgov-refresh/blob/one-with-everything/cfgov/sheerlike/storage.py#L6) class) expose all of the static files in a sheer site at static/site-slug, for example /static/owning-a-home
- similarly, that's how they are collected when 'collectstatic' is run

## random bits

I added a ./collectstatic directory, mostly as a place to test out `manage.py collectstatic` 

Reviewers: @kave @kurtw @richaagarwal @sebworks @anselmbradford, and really anyone who wants to give it a try would be appreciated.


rosskarchner:
I promise to re-learn how to use rebase soon!

cfgov/cfgov/settings/base.py
```diff
@@ -274,3 +328,26 @@
...
+
+GOVDELIVERY_USER = os.environ.get('GOVDELIVERY_USER')
+GOVDELIVERY_PASSWORD = os.environ.get('GOVDELIVERY_PASSWORD')
+GOVDELIVERY_ACCOUNT_CODE = os.environ.get('GOVDELIVERY_ACCOUNT_CODE')
+
+# LOAD OPTIONAL APPS
+# code from https://gist.github.com/msabramo/945406
+
+for app in OPTIONAL_APPS:
+    if app.get("condition", True):
```
kave: where does `condition` get set?

cfgov/cfgov/settings/base.py
```diff
@@ -34,14 +35,30 @@
...
 )
 
+OPTIONAL_APPS=[
+    {'import':'noticeandcomment','apps':('noticeandcomment','cfpb_common')},
+    {'import':'cfpb_common','apps':('cfpb_common','cfpb_common')},
+    {'import':'jobmanager','apps':('jobmanager','cfpb_common')},
+    {'import':'cal','apps':('cal','cfpb_common')},
+    {'import':'comparisontool','apps':('comparisontool','haystack','cfpb_common')},
+    {'import':'agreements','apps':('agreements','haystack', 'cfpb_common')},
+    {'import':'knowledgebase','apps':('knowledgebase','haystack', 'cfpb_common')},
```
kave: does this syntax mean, include `knowledgebase` that depends on `haystack` & `cfpb_common`

cfgov/cfgov/settings/base.py
```diff
@@ -274,3 +328,26 @@
...
+
+for app in OPTIONAL_APPS:
+    if app.get("condition", True):
+	try:
+	    __import__(app["import"])
+	except ImportError:
+	    pass
+	else:
+	    for name in app.get("apps",()):
+		if name not in INSTALLED_APPS:
```
kave: in what instances would the import fail?

kave:
a few comments to help my understanding, but looks very nice. i'll try it out locally later today

cfgov/cfgov/settings/base.py
rosskarchner: hmm-- whatever that exists for in the gist I borrowed it from, I don't think we're using it.

cfgov/cfgov/settings/base.py
rosskarchner: If an app is not installed.

kave: So it works like this:
it tries to import `{'import':'noticeandcomment','apps':('noticeandcomment','cfpb_common')}` using pythons `__import__` function if that fails then it installs it the django way?

rosskarchner: @kave  I'm not sure what you mean--  the "if" on 342 is always going to be true, unless the app in question defines a "condition" that is false.

It then tries to import whatever is in the "import" value in the dictionary.  If there is an ImportError, skip to the next app. `else` (which, in a try block, fires if there are no exceptions), apply the settings to INSTALLED_APPS, etc

rosskarchner: In other words: line 344 checks to see if an app is importable (aka, installed in the pip sense). If it isn't, nothing else happens. If it is, it gets added to INSTALLED_APPS.

kave: @rosskarchner ok now i understand. the main issue was i hadn't seen an `else` used after a `try` before had to look it up https://docs.python.org/2/tutorial/errors.html to understand what was going on.

jimmynotjim:
@rosskarchner `git rebase -i 35f7ac6` will do you

cfgov/cfgov/settings/base.py
kurtw: I can see this being set by a flag in the future.

kurtw:
_slow clap_

Will test after planning.

rosskarchner:
@kurtw give me a day to add some notes for testing, and get some uncommitted changes to the various apps, in.

rosskarchner:
# Testing with Owning a Home

Assumptions: you have an otherwise functional cfgov-refresh, with working database (but it doesn't need to be populated with anything in particular)

## Check out the following repo's as **siblings** of cfgov-refresh:
- [owning-a-home](https://github.com/cfpb/owning-a-home) (a sheer project)
- [owning-a-home API](https://github.com/cfpb/owning-a-home-api) (a django app)
- our wordpress theme, CFPB_NEMO (from our internal git)

## Download and unzip this somewhere:
[oah.sql.zip](https://github.com/cfpb/cfgov-refresh/files/138098/oah.sql.zip)
(make note of where you put it)

## Build the owning-a-home sheer site, using it's instructions

Before you run `grunt`, edit config/config.json to read:

```json
{                                                                                
    "rateCheckerAPI": "/oah-api/rates/rate-checker",                               
    "countyAPI": "/oah-api/county/",                                               
    "mortgageInsuranceAPI": ""                                                     
 }
```

## in cfgov-refresh...

```
pip install djangorestframework==2.4.8
pip install ../owning-a-home-api

cfgov/manage.py migrate

cfgov/manage.py dbshell
> source /path/to/wherever/you/put/oah.sql
# CTRL-D to close the dbshell
```

## Stop and review:

- We've build the owning a home sheer site, now in ../owning-a-home/dist . Our settings file is [configured](https://github.com/rosskarchner/cfgov-refresh/blob/one-with-everything/cfgov/cfgov/settings/base.py#L388) to look for it there, (unless overridden with OAH_SHEER_PATH)
- We've installed the django apps that power the JSON API used in the mortgage rate explorer, as well as another package it depends on. The apps needed to make this work are described in the [OPTIONAL_APPS](https://github.com/rosskarchner/cfgov-refresh/blob/one-with-everything/cfgov/cfgov/settings/base.py#L47) list
- We've loaded a snapshot of the mortgage rate data (in reality, it changes every day), into our SQL database.
- We've put the static assets from our legacy wordpress theme in reach of Django's staticfiles system

## make it go!

```sh
export WORDPRESS=http://www.consumerfinance.gov 
cfogv/manage.py sheer_index
# (observe that it now picks up the OAH  "journey" pages)
cfgov/manage.py collectstatic
# take a look at the 'collectstatic directory, 
# observe that it picks up nemo and owning-a-home assets
cfgov/manage.py runserver
# give it a try at http://localhost:8000/owning-a-home/ 
# and especially http://localhost:8000/owning-a-home/explore-rates/
```

kurtw:
This is fantastic work Ross! One question though. After we merge this in, should we focus on replacing the old site header on the sheer sites with our new one?

kave:
nice, just pulled this down and tried it out. everything works as described

PR #1352, opened 2016-01-19
Password Policy & Lockout

kave:
@rosskarchner @kurtw 

### User flows effected
- Login
- Account password change flow
- forgot my password flow

### Test
- Type in multiple wrong passwords to be locked out
- Go through "Forgot my password" flow
- goto Account Page, and try to change your password
 - should see password policy is enforced

### DIFF
--- a/cfgov/cfgov/settings/production.py
+++ b/cfgov/cfgov/settings/production.py
@@ -1,6 +1,6 @@
 from .base import *
 
-# Sends an email to developers in the email list if Debug=False for errors
+# Sends an email to developers in the ADMIN_EMAILS list if Debug=False for errors
 LOGGING = {
     'version': 1,
     'disable_existing_loggers': False,
--- a/cfgov/cfgov/settings/production.py
+++ b/cfgov/cfgov/settings/production.py
@@ -1,1 +1,27 @@
 from .base import *
+
+# Sends an email to developers in the email list if Debug=False for errors
+LOGGING = {
+    'version': 1,
+    'disable_existing_loggers': False,
+    'filters': {
+        'require_debug_false': {
+            '()': 'django.utils.log.RequireDebugFalse'
+        }
+    },
+    'handlers': {
+        'mail_admins': {
+            'level': 'ERROR',
+            'filters': ['require_debug_false'],
+            'class': 'django.utils.log.AdminEmailHandler',
+            'include_html': True,
+        }
+    },
+    'loggers': {
+        'django.request': {
+            'handlers': ['mail_admins'],
+            'level': 'ERROR',
+            'propagate': True,
+        },
+    }
+}
--- a/cfgov/cfgov/urls.py
+++ b/cfgov/cfgov/urls.py
@@ -7,7 +7,7 @@ from sheerlike.views.generic import SheerTemplateView
 from sheerlike.feeds import SheerlikeFeed
 
 from v1.views import LeadershipCalendarPDFView, EventICSView, unshare, renderDirectoryPDF, \
-    change_password, password_reset_confirm
+    change_password, password_reset_confirm, cfpb_login
 
 from wagtail.wagtailadmin import urls as wagtailadmin_urls
 from wagtail.wagtaildocs import urls as wagtaildocs_urls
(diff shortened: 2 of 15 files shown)


rosskarchner:
I'll test this out later-- but it *looks* good.

cfgov/cfgov/urls.py
```diff
@@ -6,17 +6,32 @@
 from sheerlike.views.generic import SheerTemplateView
 from sheerlike.feeds import SheerlikeFeed
 
-from v1.views import LeadershipCalendarPDFView, EventICSView, unshare, renderDirectoryPDF
+from v1.views import LeadershipCalendarPDFView, EventICSView, unshare, renderDirectoryPDF, \
```
sebworks: Is `EventICSView` being used?

cfgov/cfgov/util.py
```diff
@@ -0,0 +1,9 @@
+def admin_emails(delimited_list):
+    emails = []
+
+    if delimited_list:
```
sebworks: Is the check necessary if you use `def admin_emails(delimited_list='')`?

cfgov/cfgov/settings/production.py
```diff
@@ -1 +1,27 @@
 from .base import *
+
+# Sends an email to developers in the ADMIN_EMAILS list if Debug=False for errors
+LOGGING = {
+    'version': 1,
+    'disable_existing_loggers': False,
+    'filters': {
+        'require_debug_false': {
+            '()': 'django.utils.log.RequireDebugFalse'
```
sebworks: what does the `()` key indicate?

cfgov/templates/wagtailadmin/login.html
```diff
@@ -0,0 +1,84 @@
...
+                </li>
+            </ul>
+        </form>
+    </div>
+{% endblock %}
+
+{% block extra_js %}
+{{ block.super }}
+<script>
+    $(function(){
```
sebworks: Probably would be better if we didn't use jquery here.

kave: @sebworks so this is a wagtail template which i'm overriding, i mainly edited the `fields` section everything else i left as is

cfgov/cfgov/urls.py
kave: hmm doesn't look like it. i'll remove it

cfgov/cfgov/util.py
kave: yea the if only allows execution if its anything but `None` or `''`

cfgov/templates/wagtailadmin/account/password_reset/confirm.html
```diff
@@ -0,0 +1,53 @@
...
+
+{% block extra_css %}
+    {{ block.super }}
+
+    {% compress css %}
+        <link rel="stylesheet" href="{% static 'wagtailadmin/css/layouts/login.css' %}" type="text/css" />
+    {% endcompress %}
+{% endblock %}
+
+{% block furniture %}
```
kurtw: > block furniture
lol wut

cfgov/cfgov/settings/production.py
kave: not exactly sure, but its part of the syntax they use to instantiated filters https://docs.djangoproject.com/en/1.9/topics/logging/#django.utils.log.RequireDebugFalse

cfgov/templates/wagtailadmin/account/password_reset/confirm.html
kave: yea some of these blocks names are questionable lol

rosskarchner:
This is pretty bad-ass, @kave.

One weird bit, is that the "passwords don't match" error appears differently than the password quality errors. Can we get that one to appear in the same error box?

sebworks:
@kave, thanks for he explanations.

kurtw:
I'll +1 what Ross said. :facepunch: :anger: :tada:

rosskarchner:
Just noticed one thing-- a user can, once logged in, navigate to /django-admin, which has it's own password change screen that doesn't obey the new rules. Maybe we need to capture that URL, and redirect to our new form? Does wagtail have a similar page for already logged-in users to change their password?


screen shot 2016-01-20 at 9 38 12 am

kave:
@rosskarchner ahh i see. yea it does at the `/account/change_password/` url. i can reroute the django admin url to that

rosskarchner:
Similarly, we probably need to intercept people who end up at http://localhost:8000/django-admin/login/ , maybe involves setting these?

https://docs.djangoproject.com/en/1.9/ref/settings/#login-redirect-url

kave:
@rosskarchner so creating those setting fields didn't seem to work as expected, so i just overridden the admin url paths https://github.com/cfpb/cfgov-refresh/commit/c33030ceb5f5422dbe623d534c831fe15030caaf#diff-e45111e649c8f0371d5367b7351f1dd8R21

rosskarchner:
:+1:

PR #1358, opened 2016-01-21
addressed,security concerns with redirect_field_name

kave:
Set settings.LOGIN_REDIRECT_URL field to get proper usage of Django is_safe_url which returns True when the URL is safe to be used as a redirect; in the case of a False return, you'd simply use a default path to redirect to.

Also edited request to get redirect_field_name from GET request only.

@kurtw


kurtw:
:+1:

kurtw:
Inadvertent close.

PR #1361, opened 2016-01-22
Language dropdown for pages, which defaults to english

kave:
Included a Language selector for Pages, which defaults to English.
This was included because the use case for multiple languages on 1 page is not valid.

@KimberlyMunoz 
@kurtw 

### Testing
- Template tag: {{ page.language }}

### Screenshots
screen shot 2016-01-22 at 9 41 50 am

screen shot 2016-01-22 at 9 42 26 am

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -158,6 +158,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Added Frontend: Mega Menu.
 - Added Frontend: Global Eyebrow.
 - Added Frontend: Global Search molecule.
+- Added language dropdown for pages, which defaults to english
 
 ### Changed
 - Updated the primary nav to move focus as user enters and leaves nav levels
--- a/cfgov/jinja2/v1/wagtail/demo/index.html
+++ b/cfgov/jinja2/v1/wagtail/demo/index.html
@@ -12,7 +12,7 @@
 {% block content_main %}
     <div class="block
                 block__flush-top">
-        {{ page.title }}
+        {{ page.title }} ({{ page.language }})
     </div>
 
     {% for block in page.molecules %}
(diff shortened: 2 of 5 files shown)


sebworks:
What it make sense to include the direction of the language as well?

Schlachtmeyer:
Is there a workaround for the sidefoot, where we offer "Help in more languages" and each language links to the other languages?

kave:
@sebworks what do you mean by the direction of the language?

KimberlyMunoz:
> What it make sense to include the direction of the language as well?

Arabic is the only language where it's rtl. We'll include "if Arabic" then rtl logic on the front-end, but it shouldn't be another option for the content folks to enter since Arabic will always be RTL

@sebworks @kave

KimberlyMunoz:
@Schlachtmeyer We'll have to fix that in v2 with a special module for that circumstance. I think that needs to be its own module and not just a generic sidefoot panel.

kurtw:
:+1:

Schlachtmeyer:
So wait -- what do we do from a content standpoint for the sidefoot that contains all languages? I think this is most useful to the service providers that help out consumers in many languages so it's helpful to cross-reference all of them. Do we need to delete this for now? Is there a way to hard-code or use images or a nonstandard font?

kave:
@KimberlyMunoz

KimberlyMunoz:
Let's talk it over when I'm in the office this week, I think we need to do custom development to make that module accessible. There's a couple different ways to tackle it, including hard coding, but I don't want to single-handledly make a new module without talking it over with @schaferjh. For now this PR solves the issue of getting non-English content to be screenreader-accessible.

Schlachtmeyer:
@KimberlyMunoz Got it -- see you in the office hopefully soon!

schaferjh:
Sounds good, thanks!

PR #1362, opened 2016-01-22
Adding Expandable Observe

sebworks:
Adding code to refresh the height of the Expandable on DOM insertion/deletion. Also fixed some minor bugs so I could test on IE9/Firefox. 

Having written this I believe I took the wrong approach. @ans, how would you feel about exploring using css to handle showing/hiding the `m-expandable_content` instead of Javascript? This would probably involve using `margin-top ` instead of setting the height. 

## Testing

- Visit  `http://localhost:8000/browse-filterable/`.

## Notes

Firefox just recently added `innerText` and its better to use `textContent` per https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent. Of course this won't work in IE8 which I'm happy about. 

We should also consider throttling or debouncing `_refreshHeight`  to prevent costly reflows caused by accessing `offsetHeight`.

- @anselmbradford 
- @KimberlyMunoz 
- @jimmynotjim 

## Screenshots
screen shot 2016-01-22 at 11 54 16 am

### DIFF
--- a/cfgov/unprocessed/js/molecules/Expandable.js
+++ b/cfgov/unprocessed/js/molecules/Expandable.js
@@ -94,26 +94,19 @@ function Expandable( element ) { // eslint-disable-line max-statements, inline-c
     if ( MutationObserver ) {
       observeDOM = function() {
         var observer = new MutationObserver( function( mutations ) {
-          mutations.forEach( function( ) {
-            _refreshHeight();
-          } );
+          mutations.forEach( _refreshHeight );
         } );
 
         observer.observe( _content, { childList: true, subtree: true } );
       };
     } else {
       observeDOM = function() {
-          _content.addEventListener( "DOMNodeInserted", function( event ){
-          _refreshHeight();
-        }, false );
-
-        _content.addEventListener( "DOMNodeRemoved", _refreshHeight, false );
+        _content.addEventListener( 'DOMNodeInserted', _refreshHeight, false );
+        _content.addEventListener( 'DOMNodeRemoved', _refreshHeight, false );
       };
     }
 
-    window.setTimeout( function(){
-      observeDOM();
-    }, 0 )
+    window.setTimeout( observeDOM, 0 );
 
     return _that;
   }
--- a/cfgov/unprocessed/js/molecules/Expandable.js
+++ b/cfgov/unprocessed/js/molecules/Expandable.js
@@ -75,7 +75,47 @@ function Expandable( element ) { // eslint-disable-line max-statements, inline-c
     if ( window.addEventListener ) {
       window.addEventListener( 'resize', _refreshHeight );
     }
+
+    _initObserver();
+
     return this;
+  }
+
+  /**
+   * Watch for the insertion/removal of DOM nodes.
+   * @returns {Object} The Expandable instance.
+   */
+  function _initObserver() {
+    var MutationObserver = window.MutationObserver ||
+                           window.WebKitMutationObserver ||
+                           window.MozMutationObserver;
+    var observeDOM;
+
+    if ( MutationObserver ) {
+      observeDOM = function() {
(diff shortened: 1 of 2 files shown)


anselmbradford:
> Having written this I believe I took the wrong approach. @ans, how would you feel about exploring using css to handle showing/hiding the m-expandable_content instead of Javascript?

Sounds like a good idea. What change do you see for this? Currently almost all of the showing/hiding is handled by CSS. The JS adds/removes a CSS class that expands/collapses the content, though the height is calculated in JS and applied directly via inline styles.

sebworks:
So I think it would work by using `margin-top: -100%` to hide and `margin-top: 0` to show the content.   This way the height would always be calculated automatically by it's contents and you could still use an animation to hide/show the content.

anselmbradford:
> So I think it would work by using margin-top: -100% to hide and margin-top: 0 to show the content. This way the height would always be calculated automatically by it's contents and you could still use an animation to hide/show the content.

Sounds great if that'll do it! It could be added to the `__expanded` modifier class.

cfgov/unprocessed/js/molecules/Expandable.js
```diff
@@ -75,10 +75,50 @@ function Expandable( element ) { // eslint-disable-line max-statements, inline-c
...
+  function _initObserver() {
+    var MutationObserver = window.MutationObserver ||
+                           window.WebKitMutationObserver ||
+                           window.MozMutationObserver;
+    var observeDOM;
+
+    if ( MutationObserver ) {
+      observeDOM = function() {
+        var observer = new MutationObserver( function( mutations ) {
+          mutations.forEach( function( ) {
```
anselmbradford: Can you eliminate the nested function with `mutations.forEach( _refreshHeight );`?

cfgov/unprocessed/js/molecules/Expandable.js
```diff
@@ -75,10 +75,50 @@ function Expandable( element ) { // eslint-disable-line max-statements, inline-c
...
+            _refreshHeight();
+          } );
+        } );
+
+        observer.observe( _content, { childList: true, subtree: true } );
+      };
+    } else {
+      observeDOM = function() {
+          _content.addEventListener( "DOMNodeInserted", function( event ){
```
anselmbradford: Same as above, can you eliminate the nested function with `_content.addEventListener( "DOMNodeInserted", _refreshHeight, false );`?

sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/commit/b6653e5a83e700ed30c9310c39458555ea7ce857

cfgov/unprocessed/js/molecules/Expandable.js
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/commit/b6653e5a83e700ed30c9310c39458555ea7ce857

sebworks:
@anselmbradford , so `margin-top` will work but you still need to calculate the height to determine the duration. I'm not sure how to get around that so I think I'm going to go with that approach to now (unless you see a way to avoid doing that).

anselmbradford:
Sounds good :+1: Maybe something with animating `max-height` could be tried. Cursory look turned up this http://stackoverflow.com/questions/3508605/how-can-i-transition-height-0-to-height-auto-using-css

anselmbradford:
Hey what about using `translateY` within an `overflow: hidden;` area?

jimmynotjim:
I tried the `translate` with `overflow` route with the original refactor of the menu, but ran into the same issues you're discussing. Using `height` and calculating/writing that and the speed in JS, but letting the CSS transition handle the actual animation is most likely going to give us the best of both rendering performance and the extensibility.

PR #1364, opened 2016-01-22
Removes color variable

anselmbradford:
## Testing

- Visit the homepage in an incognito window. You should see the link color change on rollover:

screen shot 2016-01-22 at 6 26 57 pm

## Review
@jimmynotjim You were right, the link color set on the Global Header CTA can be removed. The color change is due to the global header CTA being a visited link color.


sebworks:
:+1:

PR #1367, opened 2016-01-26
Updates global search var names

anselmbradford:
Updates global search var names for tab key presses. See https://github.com/cfpb/cfgov-refresh/commit/3157fd75fb9b397c8d4b22f2ae79827e4510414d#commitcomment-15588805

## Changes

- Updates names in global search script.

## Review

- @jimmynotjim


jimmynotjim:
:+1:

PR #1368, opened 2016-01-26
Fixed issue with error messages not being displayed during incorrect Login with false Username

kave:
Fixed issue with error messages not being displayed during incorrect logins with false Usernames

- @kurtw


kurtw:
:+1: Nice!

PR #1369, opened 2016-01-26
Creating 50-50 module and using CF two-column css

KimberlyMunoz:
## Additions
- Creates two-column module for @kave and @kurtw to add to well organism

## Changes
- Changes all the two-columns modules to use `content-l_col-1-2` and removes unnecessary LESS
- Less LESS!

## Testing
- Check out the 50-50 text image module and the half link blob on the landing page

## Review
- @jimmynotjim 
- @anselmbradford 
- @sebworks

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/image-text-50-50-group.html
+++ b/cfgov/jinja2/v1/_includes/organisms/image-text-50-50-group.html
@@ -6,12 +6,14 @@
         'heading': 'Image Text 50 50 Group Header',
         'image_texts': [
             _image_text_50_50_markup(),
+            _image_text_50_50_markup(),
+            _image_text_50_50_markup(),
             _image_text_50_50_markup()
         ]
     } %}
 {% endif %}
 
-<div class="o-image-text-50-50-group">
+<div class="o-image-text-50-50-group content-l">
     <h2>{{ value.heading }}</h2>
     {% for block in value.image_texts %}
         {{ block }}
--- a/cfgov/unprocessed/css/organisms/image-text-50-50-group.less
+++ b/cfgov/unprocessed/css/organisms/image-text-50-50-group.less
@@ -4,7 +4,7 @@
   patterns:
     - name: Default example
       markup: |
-        <div class="o-image-text-50-50-group">
+        <div class="o-image-text-50-50-group content-l">
             Link blobs live inside the wrapper
         </div>
       codenotes:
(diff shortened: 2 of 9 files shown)


cfgov/jinja2/v1/_includes/organisms/image-text-50-50-group.html
```diff
@@ -6,12 +6,14 @@
         'heading': 'Image Text 50 50 Group Header',
         'image_texts': [
             _image_text_50_50_markup(),
+            _image_text_50_50_markup(),
```
KimberlyMunoz: I added these to make it easier to test spacing with more than two molecules.

cfgov/unprocessed/css/organisms/link-blob-group.less
```diff
@@ -20,6 +20,11 @@
     margin-left: -@grid_gutter-width / 2;
     margin-right: -@grid_gutter-width / 2;
   });
+
+  > h2 {
+        margin-left: @grid_gutter-width / 2;
+        margin-right: @grid_gutter-width / 2;
```
anselmbradford: Do these need unit conversion?

KimberlyMunoz: Converted and realized this and the other `> h2` on the 50-50 text need to have the `@font-face` exposed to have the margin be sized correctly.

kurtw:
:+1:

cfgov/unprocessed/css/organisms/image-text-50-50-group.less
```diff
@@ -19,14 +19,10 @@
...
 
-    .respond-to-min(@bp-sm-min, {
-        margin-left: -@grid_gutter-width / 2;
-        margin-right: -@grid_gutter-width / 2;
-    });
-
     > h2 {
-        margin-left: @grid_gutter-width / 2;
-        margin-right: @grid_gutter-width / 2;
+          .heading-2();
```
jimmynotjim: Why does an h2 need the `.heading-2()` mixin?

Same question in the link-blob group.

KimberlyMunoz: To expose the `@font-size` variable to get the spacing right when I convert to ems below. I don't necessarily like using the the `.heading-2()` mixin, but I'd prefer not to hardcode the values in case they change.

sebworks:
:+1:

cfgov/unprocessed/css/organisms/image-text-50-50-group.less
KimberlyMunoz: Talked in person, this is okay, but I'll make an issue to have exposed `@h2-font-size` variables in cf-enhancements for cases like this.

PR #1371, opened 2016-01-26
Password policy phase2 

rosskarchner:
A grab-bag of security-tastic treats to warm your risk mitigating heart.

- A user may not use any of their previous 10 passwords
- A user may not change their password more then once in a 24 hour period
- Passwords expire after 90 days
- a new model for temporary lockouts (currently only get created after some number of login failures, but we may find other uses for it)

I relocated some of the lockout-handling code into form methods, and similarly, moved some things out of 'messages' and into ValidationError's.

@kave @kurtw @richaagarwal


cfgov/v1/views.py
```diff
@@ -218,12 +213,28 @@ def custom_password_reset_confirm(request, uidb64=None, token=None,
...
+                            encrypted_password=user.password,
+                            locked_until = locked_until,
+                            expires_at = expires_at)
+
+                    password_history.save()
+		    user.temporarylockout_set.all().delete()
+		    try: 
+			user.failedloginattempt.delete()
+		    except ObjectDoesNotExist:
+		    	pass
```
kurtw: line 232 to 236 spacing issues

cfgov/v1/views.py
```diff
@@ -140,40 +150,25 @@ def login_with_lockout(request, template_name='wagtailadmin/login.html'):
...
             # Ensure the user-originating redirection url is safe.
             if not is_safe_url(url=redirect_to, host=request.get_host()):
                 redirect_to = resolve_url(settings.LOGIN_REDIRECT_URL)
 
+
+	    user = form.get_user()
+	    try:
+		user.failedloginattempt.delete()
+	    except ObjectDoesNotExist:
+	    	pass
```
kurtw: same spacing issues as below

cfgov/v1/views.py
rosskarchner: interesting-- I wonder why it didn't break?

kurtw:
Hmm I seem to be getting errors for every model class 

`/Users/wallk/v1/cfgov-refresh/cfgov/v1/models/demo.py:17: RemovedInDjango19Warning: Model class v1.models.demo.DemoPage doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.
  class DemoPage(CFGOVPage):`

cfgov/v1/forms.py
```diff
@@ -1,12 +1,86 @@
...
+	def clean(self):
+	    username = self.cleaned_data.get('username')
+	    password = self.cleaned_data.get('password')
+
+	    if username and password:
+		self.user_cache = authenticate(username=username,
+					       password=password)
+		if (self.user_cache is None and username is not None):
+		    UserModel = get_user_model()
+		    user = UserModel._default_manager.get(username=username)
```
kurtw: This needs to be protected for when there's not user for the username given. Getting this: `User matching query does not exist.`

cfgov/v1/forms.py
```diff
@@ -1,12 +1,86 @@
...
+			raise ValidationError("This account is temporarily locked; please try later or reset your password")
+		    else:
+			fa.save()
+			raise ValidationError('Login failed. %s more attempts until your account will be temporarily locked.' % (attempts_allowed-attempts_used))
+
+		else:
+		    self.confirm_login_allowed(self.user_cache)
+
+		    dt_now = timezone.now()
+		    current_password_data = self.user_cache.passwordhistoryitem_set.latest()
```
kurtw: Just like above, it needs to be protected. Getting `PasswordHistoryItem matching query does not exist.`

rosskarchner:
I saw that-- but it's not an error, it's a warning ;)

rosskarchner:
committed fixes for the spacing issues and the unknown user case

rosskarchner:
That warning is interesting-- apparently something is causing those models to be imported before the app is officially initialized-- which isn't a problem now, but will break in 1.9

rosskarchner:
The code that creates new PasswordHistoryItem entries now lives in a signal handler, so it'll catch password changes whereever they happen-- even an administrator changing a password in django-admi.

Another thing I had to learn about: [AppConfig's](https://docs.djangoproject.com/en/1.8/ref/applications/#configuring-applications).

kurtw:
After a user's password expires, there's should probably be something that directs them to a `Reset Password` view or an automatic redirect.

image

cfgov/v1/apps.py
```diff
@@ -0,0 +1,13 @@
+from django.apps import AppConfig
+from django.db.models.signals import post_save
+from django.contrib.auth import get_user_model
+
+from .signals import user_save_callback
+
+class V1AppConfig(AppConfig):
```
kave: you know what. i think we should probably start getting into the habit of not using `v1` anymore to name stuff since it'll be a pain when we start referring to the project as `v2` maybe we can call this `CFGovAppConfig` and edit the names?

cfgov/v1/views.py
```diff
@@ -25,9 +29,14 @@
...
 from wagtail.wagtailadmin.views import account
 from wagtail.wagtailadmin import forms
+
+from .forms import login_form
 from .util import password_policy
 from .models import base
+from .models.base import PasswordHistoryItem
+
 
+LoginForm = login_form()
```
kave: maybe you could do `from .forms import login_form as LoginForm` instead?

kurtw:
The above comment should apply to the lockout error as well.

Other than that, I tested everything out and it's clear! Good work @rosskarchner! :+1:

cfgov/v1/views.py
kurtw: There's a comment in the login_form() that explains why he's done this.

kave: ah i see

kave:
minor comments, looks to me :+1:

rosskarchner:
OK, I've edited those errors with a link to the password reset page.

PR #1374, opened 2016-01-27
Updated `lodash` from `3.10.0` from `4.0.1`

anselmbradford:
## Changes

- Updated `lodash` from `3.10.0` from `4.0.1` and requires `lodash.assign` as a module instead of importing the full lodash and calling the assign method.

## Testing

- `npm install && gulp test --sauce=false` should work as before.

## Review

- @sebworks 
- @KimberlyMunoz 
- @jimmynotjim


sebworks:
:+1:

PR #1376, opened 2016-01-27
Refactoring Beta Banner

sebworks:
Refactoring Beta Banner to use a general life cycle (`init`, `initEvents`, `destroy`). Also switched the code to use the Atomic Expandable.

## Testing

- Visit  `http://localhost:3000/`.

## Review

- @anselmbradford 
- @KimberlyMunoz 
- @jimmynotjim 

## Notes

- Currently working on the testing spec and trying to figure out how to test private functions.
- We should also consider passing boolean values instead of integers to set the initial state of the Expandable.

## Screenshots
screen shot 2016-01-27 at 10 54 16 am

### DIFF
--- a/cfgov/jinja2/v1/_layouts/base.html
+++ b/cfgov/jinja2/v1/_layouts/base.html
@@ -132,22 +132,33 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
 <a href="#main" id="skip-nav">Skip to main content</a>
 
 {% if flag_enabled('BETA_NOTICE') %}
-<div class="beta-banner expandable" id="beta-banner">
+<div class="beta-banner
+            m-expandable
+            m-expandable__expanded"
+     id="beta-banner">
     <div class="wrapper wrapper__match-content">
         <div class="beta-banner_head">
-            <span class="cf-icon cf-icon-error-round beta-banner_icon"></span>
+            <span class="cf-icon
+                         cf-icon-error-round
+                         beta-banner_icon"></span>
             This beta site is a work in progress.
         </div>
-        <p class="beta-banner_desc expandable_content">
-            We’re prototyping new designs. Things may not work as expected.
-            Our regular site continues to be at
-            <a href="http://www.consumerfinance.gov/">www.consumerfinance.gov</a>.
-        </p>
-        <button class="btn beta-banner_btn expandable_target" id="beta-banner_btn">
-            <span class="expandable_cue-close">
+        <div class="m-expandable_content">
+            <p class="beta-banner_desc  m-expandable_content-animated">
+                We’re prototyping new designs. Things may not work as expected.
+                Our regular site continues to be at
+                <a href="http://www.consumerfinance.gov/">www.consumerfinance.gov</a>.
+            </p>
+        </div>
+        <button class="btn
+                       beta-banner_btn
+                       m-expandable_target
+                       m-expandable_link"
+                id="beta-banner_btn">
+            <span class="m-expandable_cue m-expandable_cue-close">
                 Collapse <span class="cf-icon cf-icon-up"></span>
             </span>
-            <span class="expandable_cue-open">
+            <span class="m-expandable_cue m-expandable_cue-open">
                 More info <span class="cf-icon cf-icon-down"></span>
             </span>
         </button>
--- a/cfgov/unprocessed/css/pages/beta.less
+++ b/cfgov/unprocessed/css/pages/beta.less
@@ -12,8 +12,12 @@
     font-size: 0.875em;
     transition: padding 0.2s ease-out;
 
-    &.expandable__expanded {
+    &.m-expandable__expanded {
         padding: @grid_gutter-width 0;
+    }
+
+    .m-expandable_content {
+        border-bottom: none;
     }
 
     .respond-to-min(@bp-sm-min, {
@@ -25,7 +29,7 @@
     .h4();
     margin-bottom: 0;
 
-    .expandable__expanded & {
+    .m-expandable__expanded & {
         margin-bottom: unit(5px / @base-font-size-px, em);
     }
 }
@@ -36,6 +40,8 @@
 
 .beta-banner_desc {
     margin-bottom: 0;
+    padding: 0;
+    background: @gold-20;
 }
 
 .beta-banner_btn {
@@ -75,6 +81,12 @@
         // so the button position has to be adjusted accordingly.
         top: unit(-12px/@base-font-size-px, em);
     });
+}
+
+.no-js {
+    .beta-banner_btn {
+        display: none;
+    }
 }
 
 .beta-hero {
--- a/cfgov/unprocessed/js/modules/beta-banner-state.js
+++ b/cfgov/unprocessed/js/modules/beta-banner-state.js
@@ -5,54 +5,65 @@
 
 'use strict';
 
-// Import modules
-var $ = require( 'jquery' );
+// Required modules.
+var Expandable = require( '../molecules/Expandable' );
 var _storage = require( './util/web-storage-proxy' );
 
 // Private variables.
-var _key = 'betaBannerIsCollapsed';
+var _expandable;
+
+// Constants.
+var EXPANDED_STATE = 'betaBannerIsExpanded';
+
 
 /**
  * Set up DOM references and event handlers.
  */
 function init() {
 
-  // DOM references.
-  var btnDOM = $( '#beta-banner_btn' );
-  var bannerDOM = $( '#beta-banner' );
+  // Init Expandable.
+  var betaBannerDom = document.querySelector( '#beta-banner' );
+  var isExpanded = _storage.getItem( EXPANDED_STATE ) !== 'false';
 
-  // Event handlers.
-  btnDOM.click( _betaBannerClicked );
+  _expandable = new Expandable( betaBannerDom );
+  _expandable.init( isExpanded && _expandable.EXPANDED );
 
-  // Initial state.
-  if ( _storage.getItem( _key ) !== 'true' ) {
-    bannerDOM.get( 0 ).expand();
-  }
+  _initEvents();
 }
 
 /**
  * Run when the beta in-progress banner has been clicked.
- * @returns {boolean} always returns false.
  */
-function _betaBannerClicked() {
-  _toggleStorage();
-  return false;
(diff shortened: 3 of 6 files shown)


cfgov/unprocessed/css/pages/beta.less
```diff
@@ -13,10 +13,14 @@
     font-size: 0.875em;
     transition: padding 0.2s ease-out;
 
-    &.expandable__expanded {
+    &.m-expandable__expanded {
         padding: @grid_gutter-width 0;
```
anselmbradford: This is fine for now, but I had a discussion with @jimmynotjim that we'll likely want to abstract out some kind of generic expandable stuff so the atomic expandable isn't being customized outside of the atomic workflow (customizing margin is fine, but changing other parts of the appearance arguably changes it into an atomic element inconsistent with the design docs). We'd talked about it in terms of JS to have something like:

```
           GenericExpandableThing.js
             /         \           \
Expandable.js    SecondaryNav.js    BetaBanner.js
```
_Note: above is using composition **not** inheritance_

The same sort of structure may make sense for the CSS as well. Really food for thought at this point though.

cfgov/unprocessed/css/pages/beta.less
```diff
@@ -13,10 +13,14 @@
...
     }
 
+    .m-expandable_content {
+        border-bottom: none;
```
anselmbradford: The beta banner has a bottom border. Can this border be retained?

cfgov/unprocessed/js/modules/beta-banner-state.js
```diff
@@ -5,38 +5,47 @@
...
  */
 function init() {
 
-  // DOM references.
-  var btnDOM = $( '#beta-banner_btn' );
-  var bannerDOM = $( '#beta-banner' );
+  // Init Expandable.
+  var betaBannerDom = document.querySelector( '#beta-banner' );
+  _expandable = new Expandable( betaBannerDom );
+  _expandable.init( _storage.getItem( COLLAPSED_STATE ) !== 'true' && 3 );
```
anselmbradford: What is the `&& 3` for?

cfgov/unprocessed/js/modules/beta-banner-state.js
```diff
@@ -5,38 +5,47 @@
...
 }
 
 /**
  * Run when the beta in-progress banner has been clicked.
  * @returns {boolean} always returns false.
  */
-function _betaBannerClicked() {
-  _toggleStorage();
-  return false;
+function initEvents() {
```
anselmbradford: `initEvents` should probably be private. Is there a use-case to exposing `initEvents`, which can't be handled by calling or not calling `init`? 

It makes sense to expose it publicly if events are added, then removed, and then added again—but when would that behavior be happening for the beta banner?

cfgov/unprocessed/js/modules/beta-banner-state.js
```diff
@@ -5,38 +5,47 @@
...
+  _expandable.addEventListener( 'click', toggleStorage );
+}
+
+/**
+ * Remove DOM references and event handlers.
```
anselmbradford: Update code comment. No DOM references are removed.

cfgov/unprocessed/js/modules/beta-banner-state.js
```diff
@@ -5,38 +5,47 @@
...
+ */
+function remove() {
```
anselmbradford: Update to `removeEvents`?

cfgov/unprocessed/js/modules/beta-banner-state.js
```diff
@@ -44,15 +53,20 @@ function _betaBannerClicked() {
  * @returns {boolean} Returns value stored in the web storage,
  *   either true or false.
  */
-function _toggleStorage() {
-  var value = _storage.getItem( _key );
+function toggleStorage() {
```
anselmbradford: Maybe `toggleStoredState` instead?

cfgov/unprocessed/css/pages/beta.less
sebworks: So if I don't remove the border then it looks like:

screen shot 2016-01-27 at 12 53 05 pm

cfgov/unprocessed/js/modules/beta-banner-state.js
sebworks: You have to pass that now per https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/unprocessed/js/molecules/Expandable.js#L30 but we should probably change the Expandable to accept Booleans.

cfgov/unprocessed/js/modules/beta-banner-state.js
sebworks: How would you feel about `destroy` instead? It would serve as a general teardown function.

cfgov/unprocessed/js/modules/beta-banner-state.js
sebworks: Yeah, it should definitely be private but I was exposing it to make the code more testable for reasons mentioned here : http://philipwalton.com/articles/why-i-test-private-functions-in-javascript/. 

You shouldn't have to forgo encapsulation to do that though. I will make the change.

cfgov/unprocessed/js/modules/beta-banner-state.js
anselmbradford: Ahh, gotcha, yea so update it to this --> `_expandable.init( _storage.getItem( COLLAPSED_STATE ) !== 'true' && _expandable.EXPANDED );`

cfgov/unprocessed/js/modules/beta-banner-state.js
anselmbradford: `destroy` seems like it would undue whatever `init` does, which maybe is what you're after?

cfgov/unprocessed/css/pages/beta.less
anselmbradford: Gotcha, well that's fine. Like mentioned earlier we'll likely want to abstract this out of the expandables at some point. Alternatively if the padding could be placed on the content inside the expandable then the border could end up at the edge.

cfgov/unprocessed/js/modules/beta-banner-state.js
sebworks: ha, it's in the comments.

cfgov/unprocessed/js/modules/beta-banner-state.js
sebworks: yep, that is it exactly.

anselmbradford: Sounds good!

cfgov/unprocessed/css/pages/beta.less
sebworks: Sounds good, maybe just have a mixin.

anselmbradford:
:+1:

cfgov/unprocessed/css/pages/beta.less
jimmynotjim: The main nav (in mobile view) could probably be added to this list.

cfgov/unprocessed/js/modules/beta-banner-state.js
```diff
@@ -5,38 +5,49 @@
...
  * Set up DOM references and event handlers.
  */
 function init() {
 
-  // DOM references.
-  var btnDOM = $( '#beta-banner_btn' );
-  var bannerDOM = $( '#beta-banner' );
+  // Init Expandable.
+  var betaBannerDom = document.querySelector( '#beta-banner' );
+  var isExpanded = _storage.getItem( COLLAPSED_STATE ) !== 'true';
```
jimmynotjim: Any specific reason to use not equal instead of a strict equal? What if the value is something arbitrary like `foo`?

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -124,22 +124,24 @@
...
-        <button class="btn beta-banner_btn expandable_target" id="beta-banner_btn">
-            <span class="expandable_cue-close">
+        <div class="m-expandable_content">
+            <p class="beta-banner_desc  m-expandable_content-animated">
+                We’re prototyping new designs. Things may not work as expected.
+                Our regular site continues to be at
+                <a href="http://www.consumerfinance.gov/">www.consumerfinance.gov</a>.
+            </p>
+        </div>
+        <button class="btn beta-banner_btn m-expandable_target m-expandable_link" id="beta-banner_btn">
```
jimmynotjim: More than three classnames should be stacked. Double check these files elsewhere.

cfgov/unprocessed/js/modules/beta-banner-state.js
sebworks: If something arbitrary gets passed then `isExpanded` would be set to true which is the default.

cfgov/jinja2/v1/_layouts/base.html
sebworks: fixed in https://travis-ci.org/cfpb/cfgov-refresh/builds/105687904

sebworks:
I've added the Unit Test in https://github.com/cfpb/cfgov-refresh/commit/c1547328b9cf0b1f5507142c40e8ab22d3754dba. @anselmbradford, @jimmynotjim  please review when you get a chance.

anselmbradford:
Just some minor comments :+1:

KimberlyMunoz:
:+1: Good job on this!

cfgov/unprocessed/js/modules/beta-banner-state.js
jimmynotjim: Got it. The naming is a bit confusing, maybe we should rename the stored value as `EXPANDED_STATE` to make it clearer?

sebworks: The only issue is that the local/session storage key is called `betaBannerIsCollapsed`.  Still probably would be clearer if you changed it. What do you think?

jimmynotjim: Yeah, there's no reason not to update that storage key, this isn't live yet so it's only saved on our machines. This also doesn't need to hold up this PR, it's more of a nice to have. Feel free to move this to an issue so you can merge this in.

jimmynotjim:
Fix the conflicts and you're good to go :+1:

Issue #1370, opened 2016-01-26
Social media removed above intro section


duelj:
there are multiple pages that have social media icons showing above the text introduction section that need to be removed. For pages that should have social media they should be positioned below the introduction section. Should match [item introduction](https://github.cfpb.gov/flapjack/Modules-V1/wiki/Item-Introduction) if social media is needed, rather than [text introduction](https://github.cfpb.gov/flapjack/Modules-V1/wiki/Text-Introduction).

##### Example of social media out of position
<img width="1274" alt="screen shot 2016-01-26 at 1 47 32 pm" src="https://cloud.githubusercontent.com/assets/6873734/12590917/60d55b44-c433-11e5-9a40-cea6f60db157.png">

##### Affected pages 
[Protecting against credit discrimination](http://content.refresh.demo.cfpb.gov/consumer-help/managing-credit/protecting-against-credit-discrimination/)
[Protecting against fraud](http://content.refresh.demo.cfpb.gov/consumer-help/planning-retirement/protecting-against-fraud/)
[Sending money abroad Spanish](http://content.refresh.demo.cfpb.gov/consumer-help/sending-money-abroad/es/)
[Money tips and basics for servicemembers](http://content.refresh.demo.cfpb.gov/consumer-help/financial-resources-servicemembers/money-tips/)
[College Credit Card Agreements](http://content.refresh.demo.cfpb.gov/data-research/college-credit-card-agreement/)
[Other languages - Sending Money Abroad](http://content.refresh.demo.cfpb.gov/consumer-help/sending-money-abroad/es/) All languages have the social media above the intro text

KimberlyMunoz:
The template code has this on top of every browse page. Our templates don't have that, so I can remove it and it'll fix this issue.

KimberlyMunoz:
fixed!

PR #1377, opened 2016-01-27
Removing social media from the top of browse pages.

KimberlyMunoz:
This was unnecessarily added and shouldn't be on this type of page. I just checked with @schaferjh to confirm.

This fixes the issues in #1370 that were all occurring on browse-basic pages.

## Review
- @jimmynotjim 
- @anselmbradford 
- @sebworks


anselmbradford:
Oops :+1:

PR #1380, opened 2016-01-27
Archive events

kave:
Created a script to be run on the cmdline by the jenkins cron job hourly to archive past events.

- @kurtw 
- @rosskarchner 
- @richaagarwal 

### Test
- import Event pages `./cfgov/manage.py import-data events eventpage --parent events -u <wagtail_admin_username> -p <wagtail_admin_pw>`
 - WP events should be living under `/events/`
- run archive events script `./cfgov/manage.py runscript archive_events`
 - WP events now live under `/events/archive`


cfgov/scripts/initial_data.py
```diff
@@ -39,3 +41,18 @@ def run():
                 submitted_for_moderation=False,
             )
             revision.publish()
+        # Archived Events Browse Filterable Page
```
kave: creates an Archived Events Browse Filterable Page under events during local dev

rosskarchner:
Looks good, but I haven't had a chance to try it out yet

cfgov/scripts/archive_events.py
```diff
@@ -0,0 +1,27 @@
...
+        events = BrowsePage.objects.get(title='Events')
+        archived_events = BrowseFilterablePage.objects.get(title='Archive')
+
+        for child in events.get_children():
+            event = child.specific
+            if isinstance(event, EventPage):
+                if event.end_dt < timezone.now():
+                    if event.can_move_to(archived_events):
+                        event.move(archived_events, pos='last-child')
+                        print event.title + ' Event .....archived'
```
kurtw: Should this print something if either doesn't exist?

kave: ah true, yea i'll add a print for that

jimmynotjim:
What's the reason for manually archiving them and not doing the same `after now` and `before now` filtering we were doing w/ Wordpress? Is it faster?

kave:
@jimmynotjim good question so in terms of querying the optimization is minimal. The main purposes is it helps structure pages in a logical tree order from the wagtail admin perspective. So it will manually move old events under an Archived events Browse Filterable page.

jimmynotjim:
@kave Thanks for the clarification. Everything looks good, but I'll obviously defer to another BEWD

rosskarchner:
I was wondering if this could be achieved with URL routing, and not moving content around-- but I might be thinking with my Django brain and not my Wagtail brain ;)

The [first example here seems like it might be relevant](http://docs.wagtail.io/en/v1.3.1/reference/contrib/routablepage.html), though.

That shouldn't stop this from going forward-- so definite :+1:

kave:
yea that would be nice if we were only concerned bout the urls. i'll pin it for the future because i can see this being relevant at some point

PR #1382, opened 2016-01-28
Fixing the FCM categories

### LINKED ISSUE #1337: fcm categories need to be updated
The categories available in the CMS aren't the categories outlined in the spec. They should be in line with the design manual and should display in sentence case. 

c5c20af0-b9ea-11e5-9e87-7e0c4ef80d9b

KimberlyMunoz:
The new FCM categories now align with what's in the design manual.
Fixes #1337 

## Testing
Create a sublanding page and check out the new categories in the Featured Content Module

## Preview
image

image

## Review
- @kave 
- @kurtw 
- @richaagarwal


cfgov/v1/models/ref.py
```diff
@@ -121,6 +130,9 @@ def category_label(category):
             if category == slug:
                 return name
 
+def fcm_label(category):
+    for slug, name in fcm_types:
+            return name
```
kurtw: This always returns the first `name` it sets! I think you're looking to return the name _if_ the `slug` and the given `category` are equal.

kurtw:
This all looks great! Just address my comment before I can give a thumbs up.

cfgov/jinja2/v1/_includes/macros/category-icon.html
```diff
@@ -27,7 +27,13 @@
...
         'blog': 'cf-icon-speech-bubble',
-        'newsroom': 'cf-icon-newspaper'
+        'newsroom': 'cf-icon-newspaper',
+        'featured event': 'cf-icon-date',
+        'featured blog': 'cf-icon-speech-bubble',
+        'featured video': 'cf-icon-play-round',
+        'featured tool': 'cf-icon-settings',
+        'featured news': 'cf-icon-newspaper',
+        'featured': 'cf-icon-favorite'
       }
```
jimmynotjim: Can you add a todo to the top of this logic to eventually move it to the backend? The slug really should be passed in with the value instead.

cfgov/v1/models/ref.py
kave: so like
```
if slug == category:
   return name
```

cfgov/jinja2/v1/_includes/macros/category-icon.html
KimberlyMunoz: There's an issue for it already. :)

cfgov/v1/models/ref.py
KimberlyMunoz: Thanks for the catch!

jimmynotjim:
:+1:

cfgov/jinja2/v1/_includes/macros/category-icon.html
jimmynotjim: Whoops, I keep forgetting

cfgov/v1/models/ref.py
KimberlyMunoz: fixed in  a937940

kurtw:
:+1:

PR #1385, opened 2016-01-28
Add CFGOVFormPage and ShareablePage model

kurtw:
- CFGOVPage and CFGOVFormPage inherit from both Page and ShareablePage
- ShareablePage owns the share logic now

This gives us the ability to add form fields to build forms of various components described [here](http://docs.wagtail.io/en/v1.3.1/reference/contrib/forms.html#form-builder). Note that the ShareablePage is abstract.

Next step is to convert the Pages that contain a form (other than the filter form already taken care of) to inherit from CFGOVFormPage.

@kave 
@richaagarwal 
@rosskarchner


kurtw:
Opting for a soon-to-come PR for wagtail that changes FormPage to a mixin.

PR #1389, opened 2016-01-29
Fixing lodash dependency error

sebworks:
Fixing lodash dependency error

# Reveiw

@kave 

# Test

Run `gulp test`

### DIFF
--- a/test/browser_tests/page_objects/page_blog-single.js
+++ b/test/browser_tests/page_objects/page_blog-single.js
@@ -1,6 +1,6 @@
 'use strict';
 
-var _assign = require( 'lodash.assign' );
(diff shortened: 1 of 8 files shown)


kave:
@sebworks hmm so i'm getting like 47 errors locally when it runs browser tests but i'm not convinced its due to the lodash because if you navigate to `/events/field-hearing-on-arbitration/` its erroring out on twitter_options `'dict object' has no attribute 'twitter_options'`

anselmbradford:
LGTM pending Emmanuel's comment

sebworks:
@kave, please test this again

kave:
@sebworks seems to be a fluke :+1:

Issue #1354, opened 2016-01-19
Higher table title character count?


Schlachtmeyer:
There appears to be a 20-character limit on table titles, which I don't think is enough. Maybe 100? 

Please see the Your Money, Your Goals page in Wagtail.

tnferrell:
To Laura's point, I'm kind of for removing character limits altogether for tables. Here's a table I'm trying to input into wagtail. http://www.consumerfinance.gov/guidance/petitions-to-modify-or-set-aside/. The first cell character limits are too low for text and for hyperlinks.

KimberlyMunoz:
I'm trying to see the bug in wagtail itself, but I can't find the table in Your Money, Your Goals. I think the BEWDs can easily increase the character limit on table titles for now though!

schaferjh:
There are no longer tables on Your Money, Your Goals because we made a content/ux decision to convert the information to browse pages instead of putting it all into tables on this sublanding page. 

As for the character limits in tables, is this linked to #1355?

KimberlyMunoz:
No, that's a different front-end issue.

schaferjh:
ok cool, thanks for the carification @KimberlyMunoz. 

@kave this is a higher priority issue now as content has been running into it in a couple of places. One example of a more recent page where they've run into the problem at the bottom of /policy-compliance/small-business-review-panels/, but you can also look at the content board on Jira, where they've flagged all the pages where they've run into problems.

kave:
@schaferjh created a fix for this to be reviewed on the next push to refresh this afternoon

https://github.com/cfpb/cfgov-refresh/pull/1418

Issue #1348, opened 2016-01-15
Required body text in link blob


Schlachtmeyer:
Right now the half-width link blob requires content between the heading and list of links. That's not working for /college-credit-card-agreements-browse/ - do we need to create lead-in text or can this be changed to not required (in Wagtail and on wiki)?

KimberlyMunoz:
Where is that page. I don't see it on the top level.

KimberlyMunoz:
Found it and the issue.

![image](https://cloud.githubusercontent.com/assets/1860176/12621580/3053b256-c4ef-11e5-9668-c171c98496d0.png)

KimberlyMunoz:
BEWDs will have to make that field unrequired.

KimberlyMunoz:
Fixed in refresh!

PR #1392, opened 2016-01-29
Issues

kave:
body not required in halfLinkBlob and header has no char limit on Table organism

Fixed [1348](https://github.com/cfpb/cfgov-refresh/issues/1348) and [1354](https://github.com/cfpb/cfgov-refresh/issues/1354)

@kurtw 
@rosskarchner 
@richaagarwal


kurtw:
LGTM

richaagarwal:
:+1: 

Out of curiosity, why is the migration file so long? Maybe I just need to read up more on StreamFields?

kave:
@richaagarwal so because the edited molecule is used in alot of other models a migration change is propagated to each model affected by this change

PR #1399, opened 2016-02-02
Atomizes mega menu

anselmbradford:
## Changes

- Migrates primary nav code to atomic mega menu.
- Makes mega menu at large sizes expand using translation, like is done at mobile sizes.
- Abstracts out JS for applying removing aria-expanded attributes so that mega menu and global search can share flyout menu code.
- Updates some of the gray brand colors based on values in cf-generator.

## Testing

- Visit any page, expand the About Us menu, resize the page. See TODOs below for known issues.

## Review

- @jimmynotjim 
- @sebworks 
- @KimberlyMunoz 

## Screenshots

screen shot 2016-02-02 at 11 44 38 am

screen shot 2016-02-02 at 11 44 47 am

screen shot 2016-02-02 at 11 44 51 am

## Todos

- Fixed ~~~The bottom drop-shadow of the menu at desktop sizes gets cropped when expanded~~~

- The mega menu and logo swap vertical stacking order at desktop sizes, unfortunately this creates a box that interrupts the flow of the global search.
screen shot 2016-02-02 at 11 49 34 am

- Partially fixed ~~~There needs to be an overlay on the page or some other means to close the menu when clicking off of it~~~  Dark overlay needed https://github.cfpb.gov/github-enterprise-assets/0000/0087/0001/9432/f712b8b6-b85a-11e5-9e9e-dae41d83563a.png)

- The mobile size mega menu has to handle three levels deep. Careers is an example.

- Fixed ~~~When clicking the close button on the hamburger menu, the sub menu flyouts need to close as well so their state isn't remembered when resizing the page and re-opening the menu.~~~

- Partially fixed ~~~Sub-menus need to handle keyboard input and focus properly~~~ Submenus aren't handling when the tab key leaves the menu properly.

- Menu needs to fully expand when JS is turned off.

- Featured Menu Content molecule needs to be added.

- Mega menu needs to show on rollover.


sebworks:
Clicking the `About Us` link on desktop causes the link to change color. Is this the expected behavior? It currently doesn't do this on BETA.

sebworks:
Should the Mega Menu display on mouse over?

sebworks:
Should the Mega Menu be 100% width? It is currently that way on Beta?

sebworks:
There is an issue on mobile when tabbing through elements:

screen shot 2016-02-04 at 8 24 08 am

sebworks:
Hidden elements are obtaining focus when tabbing. Not sure if this is the correct behavior.

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
```diff
@@ -0,0 +1,133 @@
...
+
+   Description:
+
+   Creates a mega menu primary navigation menu.
+
+   ========================================================================== #}
+{% import 'templates/nav/_vars-primary-nav.html' as vars with context %}
+<div class="o-mega-menu">
+    <button class="o-mega-menu_trigger">
+        <span class="cf-icon"></span>
```
sebworks: can we use css to add the icon?

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
```diff
@@ -0,0 +1,133 @@
...
+                           list__unstyled
+                           o-mega-menu_subcontent-list">
+                    {%- for href, id, caption in sub_nav_item %}
+                        <li class="list_item
+                                   o-mega-menu_subcontent-item">
+                            <a class="list_link
+                                      {{ 'list_link__current' if href == request.path else '' }}
+                                      {{ 'u-link__disabled' if href == '' else '' }}
+                                      o-mega-menu_subcontent-link"
+                               {{ 'href=' + href | e if href != '' else '' }}>
```
sebworks: Are you assigning a blank `href` or just omitting `href` if blank?

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
```diff
@@ -0,0 +1,133 @@
...
+            <li class="o-mega-menu_content-item">
+                {% include 'molecules/global-header-cta.html' with context %}
+            </li>
+
+            {%- for nav_title,
+                    sub_nav_landing,
+                    sub_nav_title,
+                    sub_nav_media,
+                    sub_nav_items in vars.nav_items %}
+                <li class="o-mega-menu_content-item">
```
sebworks: unnecessary indent?

cfgov/unprocessed/css/misc.less
```diff
@@ -3,6 +3,25 @@
...
+body {
+    min-width: 320px;
+}
+
+// Creates semi-transparent drop-shadow after element.
+.u-drop-shadow-after() {
+    &:after {
+        position: relative;
+        top: 5px;
+        content: '\a0';
```
sebworks: Might be worth a comment for the `\ao`?

cfgov/unprocessed/css/organisms/header.less
```diff
@@ -62,7 +61,11 @@
 */
 
 .o-header {
+    min-width: 320px;
```
sebworks: Are we going to set this on the body?

cfgov/unprocessed/css/organisms/mega-menu.less
```diff
@@ -5,44 +5,449 @@
...
     }
 
-    .m-global-eyebrow {
+    &_subcontent {
+        &-title {
+            display: block;
+            border-bottom: 1px solid @gray-50;
+        }
+
+        &-link {
```
sebworks: Do you need `subcontent-item-link`? Could you just target the `a` ?

cfgov/unprocessed/css/organisms/header.less
anselmbradford: Probably more atomic to set it on the element that needs it. I could be convinced otherwise though, but the argument would be if you take the header organism on its own it'll break if it gets below 320px (logo will get cropped), so the organism taken in isolation has a min width. 

Having a variable for this seems like a good idea.

cfgov/unprocessed/css/organisms/mega-menu.less
jimmynotjim: This is more BEM, and follows the pattern we've set up previously (ex. `list_link`). My only question is should this be `&_title` and `&_link`?

jimmynotjim:
When you open a layer on small screens, the focus continues on the trigger, which is now hidden, rather than moving into the next level. The same happens in reverse. It used to move with the layers, is this an intended change?

cfgov/unprocessed/css/organisms/mega-menu.less
anselmbradford: So my understanding of BEM is:

```
o-mega-menu                           = (B)lock
o-mega-menu_subcontent                = (E)lement of o-mega-menu block
o-mega-menu_subcontent-link           = (E)lement of o-mega-menu block
o-mega-menu_subcontent-link__disabled = (M)odifier of &_subcontent-link element in o-mega-menu block (this is a hypothetical modifier)
```

I'm not sure in BEM if you'd have an element of an element of a block, like `o-mega-menu_subcontent_link`
@Scotchester Do you have insight into this?

Scotchester: We do try to avoid elements of elements, yes. But doing a multi-level menu without that is going to be pretty difficult... I think the best compromise is often using blocks within blocks.

I'm not very knowledgeable about the atomic structure yet, but doesn't `o-mega-menu` imply that it would contain `m-things`? Would that be a way around it?  Some child molecules of the mega menu organism would be their own block in BEM parlance.

Just spitballing here... this is probably going to be possible to do in an ideologically pure way. Just do what makes the most sense :)

anselmbradford:
> Clicking the About Us link on desktop causes the link to change color. Is this the expected behavior? It currently doesn't do this on BETA.

Checked with UX/Design. No, it should stay blue. Same for the CTA.

> Should the Mega Menu display on mouse over?

Checked with UX/Design. Yes.

> Should the Mega Menu be 100% width? It is currently that way on Beta.

Checked with UX/Design. Yes.

> Hidden elements are obtaining focus when tabbing. Not sure if this is the correct behavior.

@sebworks Fixed in https://github.com/cfpb/cfgov-refresh/pull/1416

> When you open a layer on small screens, the focus continues on the trigger, which is now hidden, rather than moving into the next level. The same happens in reverse. It used to move with the layers, is this an intended change?

No, I'll fix this.

cfgov/unprocessed/css/organisms/header.less
anselmbradford: Note to self: remove https://github.com/cfpb/cfgov-refresh/pull/1399/files#diff-8263f8fac724e6961ecfffd5ed5c96e5R8

cfgov/unprocessed/css/misc.less
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/pull/1416

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
anselmbradford: This is omitting `href` if blank. This is a carryover from the old code. That seems correct, right?

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/commit/b6e3652398fa0cb565b69a4a466a10ed4b828f4f

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
anselmbradford: One indent after for loop is used elsewhere. Is this wrong?

sebworks: I no longer indent the for loops but I'm not sure what everyone else is doing.

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
sebworks: seems right but I'm never sure about the order of operations in jinja

anselmbradford:
Ready for review again. In addition to the updated TODOs in the PR description a few notes specific to earlier feedback:
- Mega menu still needs to show on rollover. I'd like to address this in a separate PR. Added a TODO.
- Tab focus works partially. Focus on the menus works, but tabbing through the menus to the end doesn't work completely as expected (offscreen submenu is tab-able and when submenu is open and user tabs through, submenu closes, but then main menu doesn't). Added TODOs and I'll address this in a separate PR.

cfgov/jinja2/v1/_includes/templates/header.html
```diff
@@ -1,20 +1,20 @@
-<header class="o-header
-               header"
+<header class="o-header"
```
sebworks: Should this file be in the `organisms` folder?

sebworks:
is this the correct appearance at `1040px`?
screen shot 2016-02-10 at 3 36 29 pm

cfgov/jinja2/v1/_includes/templates/header.html
anselmbradford: Yup, good catch. Fixed in https://github.com/cfpb/cfgov-refresh/commit/7321c8b8d227912d596290a19a8c2ce47eac0ca4

cfgov/unprocessed/js/organisms/MegaMenu.js
```diff
@@ -0,0 +1,225 @@
...
+  /**
+   * Whether a target is one of the ones that appear in the desktop view.
+   * @param {HTMLNode} target - The target of a mouse event (most likely).
+   * @returns {boolean} True if the passed target is in the desktop view.
+   */
+  function _isDesktopTarget( target ) {
+    var bodyElem = document.body;
+    target = target.parentNode;
+    do {
+      if ( target === _activeMenuDom ) {
```
sebworks: can we make this a function? Seems like the same code is repeated below ( different targets ).

anselmbradford:
> is this the correct appearance at 1040px?

It's not. I have a `TODO` [here](https://github.com/cfpb/cfgov-refresh/pull/1399/files#diff-08ab6d28d7ebf5e705737702ce9fe7a6R90) for that. I'm swapping the stacking order, but that means the area of the logo extends to the width of the menu, so when the screen resizes down the search box pushes against the logo/menu box and causing it to narrow. If you have any other ideas let me know. Flexbox may be an option since IE9/10 doesn't support pointerevents so will need to display a degraded version of the menu anyway, but I'm not sure if flexbox wouldn't have the same issue. I'll leave it to another PR though if that's okay.

cfgov/unprocessed/js/organisms/MegaMenu.js
anselmbradford: Moved to dom-traverse.js in https://github.com/cfpb/cfgov-refresh/commit/072a94a59637ebbb3efd575ee220760a02f98341

jimmynotjim:
Fonts are displaying as medium weight

jimmynotjim:
Is the underline supposed to show on hover? It currently does not, and looks kind of odd here

screen shot 2016-02-10 at 4 17 16 pm

jimmynotjim:
Out of scope, but wanted to document, the search input and submit button don't align

screen shot 2016-02-10 at 4 18 45 pm

cfgov/jinja2/v1/_includes/organisms/header.html
```diff
@@ -1,20 +1,20 @@
...
     <div class="o-header_content
                 wrapper">
-        <button class="primary-nav_trigger js-primary-nav_trigger">
-            <span class="u-visually-hidden">Menu</span>
-        </button>
 
-        {% import 'molecules/global-header-cta.html' as global_cta with context %}
-        {{ global_cta.render( true ) }}
+        {% import 'molecules/global-header-cta.html' as global_header_cta with context %}
+        {{ global_header_cta.render( true ) }}
```
jimmynotjim: Do you need `.render` here?

anselmbradford:
> Fonts are displaying as medium weight

I _think_ they are correct. Look like the mocks.

> Is the underline supposed to show on hover? It currently does not, and looks kind of odd here

Placeholder for hover state. TODO [here](https://github.com/cfpb/cfgov-refresh/pull/1399/files#diff-8ed4d4fdd74dc1a79427c798d8dedd15R357). Sorry about all the TODOs, just trying to get the files moved so I can focus on the details in specific PRs.

> Out of scope, but wanted to document, the search input and submit button don't align

Yeah, cf-forms issue https://github.com/cfpb/capital-framework/issues/261, also opened DFJR-815 for it.

cfgov/jinja2/v1/_includes/organisms/header.html
anselmbradford: `render` is the macro name, so yes. How else are you thinking?

cfgov/unprocessed/js/organisms/MegaMenu.js
sebworks: can you use https://developer.mozilla.org/en-US/docs/Web/API/Node/contains instead?

anselmbradford: Perfect! Updated in https://github.com/cfpb/cfgov-refresh/commit/0c71f80900397b153971a9fd1eaff6d3bcdc6c2f

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
```diff
@@ -0,0 +1,125 @@
...
+   sub_nav_items:   An array of tuples.
+
+   sub_nav_media:   A string composed of html markup.
+
+   ========================================================================== #}
+
+{% macro _sub_nav(sub_nav_title, sub_nav_landing, sub_nav_items, sub_nav_media) %}
+<div class="o-mega-menu_subcontent" aria-expanded="false">
+    <div class="o-mega-menu_subcontent-wrapper">
+        <div class="wrapper">
```
jimmynotjim: Do we really need two wrappers?

cfgov/unprocessed/css/organisms/header.less
```diff
@@ -83,6 +86,18 @@
             & > .m-global-search {
                 float: right;
             }
+
```
jimmynotjim: Instead of nesting could you use the `m-global-header-cta__horizontal` to float the CTA?

Does Search even need the nesting? There's only one.

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
anselmbradford: `_subcontent` creates a cropped "window" into which the menu slides down into.
`_subcontent-wrapper` sets the full width size, background color, and bottom drop-shadow.
`wrapper` sets the max-width and padding for the content to contain it within the center of the screen.

It would be nice to combine the inner two wrappers, but since they have two different widths that may be tricky.

cfgov/unprocessed/css/organisms/mega-menu.less
```diff
@@ -5,46 +5,488 @@
     - name: Default example
       markup: |
         <div class="o-mega-menu">
+            <button class="o-mega-menu_trigger" aria-expanded="false">
+                <span class="o-mega-menu_trigger-label">
+                    <span class="u-visually-hidden">Menu</span>
+                </span>
+            </button>
+            <nav class="o-mega-menu_content" aria-expanded="false" aria-label="main navigation" role="navigation">
```
jimmynotjim: Check the char counts throughout this file

cfgov/unprocessed/css/organisms/mega-menu.less
```diff
@@ -5,46 +5,488 @@
...
+        margin-bottom: 0;
+    }
 
-        .respond-to-min( @bp-med-min, {
-            display: none;
-        } );
+    &_content {
+        &-link {
+            // Colors for :link, :visited, :hover, :focus, :active.
+            .u-link__colors( @black, @black, @black, @black, @black );
```
jimmynotjim: If you do `@black` once, it'll set all of them. Here and below

cfgov/unprocessed/css/organisms/header.less
anselmbradford: > Instead of nesting could you use the m-global-header-cta__horizontal to float the CTA?

You mean put `float: right;` in `m-global-header-cta__horizontal`? So the rationale not to is to encapsulate the molecule. Having the float in the molecule means it would be aware of its layout. We should be able to move the molecule around without worrying about floats blowing up the layout of other molecules. Does that make sense?

Note: there are floats on the mega-menu trigger and global search. Added a todo for those in https://github.com/cfpb/cfgov-refresh/commit/f3bec89f514f05e43e86189e7949256d9616c598

> Does Search even need the nesting? There's only one.

I merged it into the `.m-global-header-cta` rule, does this work? https://github.com/cfpb/cfgov-refresh/commit/f3bec89f514f05e43e86189e7949256d9616c598

cfgov/unprocessed/css/organisms/mega-menu.less
```diff
@@ -5,46 +5,488 @@
...
+                margin-bottom: unit( @grid_gutter-width / 2 / 18px, em );
+
+                font-size: 1em;
+            }
+
+            &-list {
+                .grid_column(4);
+
+                &:first-child {
+                    border-left: 0;
```
jimmynotjim: Should we wrap these in an elem with `.grid_nested-col-group()`?

cfgov/unprocessed/css/organisms/mega-menu.less
```diff
@@ -5,46 +5,488 @@
...
+                &__current {
+                    cursor: default;
+                }
+            }
+        }
+
+        // Hide global header CTA and eyebrow at desktop size.
+        .m-global-header-cta,
+        .m-global-eyebrow {
+            display: none;
```
jimmynotjim: Instead of nesting, target the `__list` modifiers

cfgov/unprocessed/css/organisms/mega-menu.less
```diff
@@ -5,46 +5,488 @@
...
         }
     } );
+
+    // Large desktop size.
+    .respond-to-min( @bp-lg-min, {
+        &_subcontent-grid {
+            .grid_column( 8 );
+            border-left: 0;
```
jimmynotjim: Could we use `.grid_nested-col-group()` here too?

cfgov/unprocessed/js/modules/FlyoutMenu.js
```diff
@@ -124,6 +124,18 @@ function FlyoutMenu( element, triggerSel, contentSel, altTriggerSel ) {
     this.dispatchEvent( 'collapseEnd', { target: this } );
   }
 
+  /**
+   * @returns {Object}
+   *   Hash of trigger, alternative trigger, and content DOM references.
+   */
+  function _getDom() {
+    return {
+      altTrigger: _altTriggerDom,
```
jimmynotjim: Left align values here and throughout.

cfgov/unprocessed/css/organisms/mega-menu.less
anselmbradford: Sweet! Fixed in https://github.com/cfpb/cfgov-refresh/commit/83f76dc218a3402b12cee0251154849a6aea7a68

jimmynotjim:
It looks like the wrapper is adding too much padding. I think the left edge of the overview link should align with the main nav/logo

screen shot 2016-02-10 at 5 57 48 pm

cfgov/unprocessed/css/organisms/mega-menu.less
anselmbradford: Is there value in having all the desktop stuff wrapped in `.respond-to-min( @bp-med-min, {`? I kind of like the organization of not moving desktop-specific stuff outside of the breakpoint. Also, it seems like the mega-menu should just care that it's not displaying the global header cta and eyebrow at desktop sizes. It doesn't care if they're in a list or horizontal layout. E.g. if the horizontal layout ended up being used in the mega-menu for some reason it shouldn't require an update here to re-hide it at desktop sizes.

cfgov/unprocessed/css/organisms/mega-menu.less
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/commit/c2ce340878e2904d023b2e075cfd8d61e4598347

cfgov/unprocessed/css/organisms/mega-menu.less
anselmbradford: It messes up the Featured Menu Content aside and pushes it to a new line. Added TODO to check it when updating the FMC molecule https://github.com/cfpb/cfgov-refresh/commit/f294bbbe031f4ad11f3568adfae1f00c723d0d2d

cfgov/unprocessed/js/modules/FlyoutMenu.js
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/commit/61a40782b361155fe2cbbe50a21dce76a7763d6b

anselmbradford:
> It looks like the wrapper is adding too much padding. I think the left edge of the overview link should align with the main nav/logo

Fixed and added unit conversion in https://github.com/cfpb/cfgov-refresh/commit/5ad04ebf577c86c7b9c5004e25b6973288ac8011

cfgov/unprocessed/css/organisms/mega-menu.less
anselmbradford: Stacked everything, but with indents it's still wider. Not sure what to do here but may be worth revisiting when docs are updated throughout https://github.com/cfpb/cfgov-refresh/commit/cf2ff8222c7f966e7a39d2f9af67558503a1ebc5

sebworks:
:+1:

sebworks:
Should we be testing this with JS off?

anselmbradford:
> Should we be testing this with JS off?

No. JS and keyboard navigation are still in-progress and best in their own PRs so it's easier to see what's adjusted to make them work.

sebworks:
>No. JS and keyboard navigation are still in-progress and best in their own PRs so it's easier to see what's adjusted to make them work.

cool. Looks good.

cfgov/unprocessed/css/organisms/header.less
jimmynotjim: Makes sense, thanks

cfgov/unprocessed/css/organisms/mega-menu.less
jimmynotjim: Yeah, that makes sense. If we were to reuse these somewhere, like the footer, we'd be hiding them when we wouldn't need to. Carry on.

jimmynotjim:
:+1:

PR #1400, opened 2016-02-02
Update colors

anselmbradford:
## Changes

- Updated brand colors to updates in generator-cf.

## Testing

- `gulp build` should build.

## Review

- @sebworks 
- @jimmynotjim 
- @KimberlyMunoz 

## Todos

- `@gray-50` is deprecated but still needed by cf-layout.

### DIFF
--- a/cfgov/unprocessed/css/brand-palette.less
+++ b/cfgov/unprocessed/css/brand-palette.less
@@ -4,23 +4,100 @@
    ========================================================================== */
 
 // Official palette.
-// Current as of April 9, 2015.
+// Current as of November 13, 2015.
 // To view the colors in the Design Manual, visit:
 // https://cfpb.github.io/design-manual/identity/color-principles.html#palette
+//
+// | Green      | Teal       | Pacific    | Navy       | Red        | Gold       | Neutral    | Black/Gray |
+// | ---------- | ---------- | ---------- | ---------- | ---------- | ---------- | ---------- | ---------- |
+// |            |            |            |            |            |            |            | 101820     |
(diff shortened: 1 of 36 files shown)


cfgov/unprocessed/css/brand-palette.less
```diff
@@ -4,69 +4,143 @@
...
+@neutral:      #796e65;
+@neutral-80:   #948b84;
+@neutral-60:   #b5aca5;
+@neutral-40:   #cdc7c2;
+@neutral-20:   #e3e2df;
+
+
+// Gray family
+
+@black:        #101820; // Also known as "CFPB Black"
```
sebworks: Do you think it should be called `@black` when `@white` is `#fff`?

anselmbradford: This may be to make the black used appear more consistent across screen and print and to visibly look like consistent gradient steps through the grays to black. Black can be created via a mix of RGB, a mix of CMYK, or as 0% C 0% M 0% Y 100% K. So when converting 100% black in CMYK to black in RGB the result is likely not `#000`. See https://github.com/cfpb/design-manual/issues/327#issuecomment-140925068
Since it's not `#000`, the `@black` variable makes it easier to work with. `@white` is provided for consistency and to handle if white were to be tinted to an off-white in the future. 

At any rate, we're kind of the end users of these names since they're decided in the design manual and https://github.com/cfpb/generator-cf/blob/master/app/templates/src/static/css/brand-palette.less, so would need to be changed there.

sebworks:
:+1:

PR #1402, opened 2016-02-02
Squash migrations

kurtw:
- Keeps add_share_perms in it's own script

The only other migration that had custom work done was `cfgov/v1/migrations/0032_cfgovimage_cfgovrendition.py` but I did not think it necessary to keep around.

### To test
- Run `migrate` but `--fake` the `0002` migration with `./cfgov/manage.py migrate v1 0002 --fake` and you should be good.

@kave 
@rosskarchner 
@richaagarwal


richaagarwal:
is this still an active PR?

PR #1393, opened 2016-01-29
DFJR-719 - Remove prototype code

richaagarwal:
"Inside our atomic elements templates (/_includes/ folder) and page templates in the we want to remove all references to prototype stock content." 

## Review

- @kave 
- @rosskarchner 
- @kurtw 
- @jimmynotjim 
- @sebworks 
- @KimberlyMunoz 
- @anselmbradford 

## Notes

- I can't run `gulp test:acceptance` yet :( but did do a visual test on localhost

PR #1407, opened 2016-02-04
Removed manipulation of the `value` constant passed by the CMS

jimmynotjim:
- Fix to help #1393 along.
- There may be other areas we need to replace value where we were improperly manipulating it.

### DIFF
--- a/cfgov/jinja2/v1/_includes/molecules/global-header-cta.html
+++ b/cfgov/jinja2/v1/_includes/molecules/global-header-cta.html
@@ -8,11 +8,15 @@
 
    Creates markup for Global Header Call to Action molecule.
 
-   value.is_in_list: Whether the molecule appears in a list or not.
+   is_horizontal: Whether the molecule appears horizontally or vertically.
+                  defaults to false.
 
    ========================================================================== #}
 
-<div class="m-global-header-cta
-            m-global-header-cta__{{ 'list' if value.is_in_list else 'horizontal' }}">
-    <a href="#">Submit a Complaint</a>
-</div>
+{% macro render( is_horizontal=false ) %}
+    {# TODO: Padding should be on link, not CTA itself #}
+    <div class="m-global-header-cta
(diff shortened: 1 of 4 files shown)


cfgov/jinja2/v1/_includes/molecules/global-header-cta.html
```diff
@@ -8,11 +8,15 @@
 
    Creates markup for Global Header Call to Action molecule.
 
-   value.is_in_list: Whether the molecule appears in a list or not.
+   horizontal: Whether the molecule appears horizontally or vertically.
```
anselmbradford: Update `is_horizontal` docs

cfgov/jinja2/v1/_includes/templates/nav/primary-nav.html
```diff
@@ -45,15 +37,7 @@
...
-            'global_eyebrow': {
-                'is_in_list': true
-            }
-
-        })
-    %}
-    {% set value = global_dict.global_eyebrow %}
-    {% include 'molecules/global-eyebrow.html' with context %}
+    {% from 'molecules/global-eyebrow.html' import render as eyebrow with context %}
+    {{ eyebrow() }}
```
anselmbradford: To following existing conventions this should be `global_eyebrow.render()` so the function call name isn't dependent on the context in which it is imported.

jimmynotjim: I was following @kurtw's example in https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/jinja2/v1/browse-filterable/index.html#L29. I never know which is the right way. Kurt, can you fill us in?

kurtw: I think Ans' approach is more readable and easier to remember.

jimmynotjim: Thanks, done

anselmbradford:
:+1:

sebworks:
:+1:

PR #1409, opened 2016-02-05
Enabling Django Asset compression

sebworks:
Enabling Django Asset compression for performance benefits (one script/caching ).

## Testing

- Visit  `http://localhost:3000/` and view source.

## Review

- @anselmbradford 
- @KimberlyMunoz 
- @jimmynotjim 
- @rosskarchner 
- @kurtw 

## Notes

- I'm not sure if/how this impacts deployment of assets on production. Also not sure how cached assets are expired/refreshed.

## Screenshots
 Before : 

screen shot 2016-02-05 at 7 37 07 am

After:
screen shot 2016-02-05 at 7 37 31 am

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -7,6 +7,7 @@ from django.conf import global_settings
 # Repository root is 4 levels above this file
 REPOSITORY_ROOT = Path(__file__).ancestor(4)
 
+
 # This is the root of the Django project, 'cfgov'
 PROJECT_ROOT = REPOSITORY_ROOT.child('cfgov')
 V1_TEMPLATE_ROOT = PROJECT_ROOT.child('jinja2', 'v1')
@@ -160,6 +161,13 @@ STATIC_ROOT = os.environ.get('DJANGO_STATIC_ROOT', '/var/www/html/static')
 
 MEDIA_ROOT = os.path.join(PROJECT_ROOT, 'f')
 MEDIA_URL = '/f/'
+
+#Enabling compression for use in base.html
+COMPRESS_ENABLED = True
+
+COMPRESS_ROOT = MEDIA_ROOT
+
+COMPRESS_JS_FILTERS = []
 
 # List of finder classes that know how to find static files in
 # various locations.
--- a/config/webpack-config.js
+++ b/config/webpack-config.js
@@ -27,12 +27,10 @@ module.exports = {
   plugins: [
     new webpack.optimize.CommonsChunkPlugin( COMMON_BUNDLE_NAME ),
     new webpack.optimize.CommonsChunkPlugin( COMMON_BUNDLE_NAME,
-                                             [ COMMON_BUNDLE_NAME ] ),
+                                             'common.[chunkhash].js' ),
     // Change warnings flag to true to view linter-style warnings at runtime.
     new webpack.optimize.UglifyJsPlugin( {
       compress: { warnings: false }
     } ),
-    // Wrap JS in raw Jinja tags so included JS won't get parsed by Jinja.
-    new BannerFooterPlugin( '{% raw %}', '{% endraw %}', { raw: true } )
   ]
 };
--- a/cfgov/jinja2/v1/_layouts/base.html
+++ b/cfgov/jinja2/v1/_layouts/base.html
@@ -105,6 +105,8 @@
         docElement.className = docElement.className.replace( /(^|\s)js(\s|$)/, '$1no-js$2' );
     </script>
     <![endif]-->
+    
+    <!--[if IE 9]><script src="{{ static('js/ie/common.ie.js') }}"></script><![endif]-->
 
     {# TODO: jQuery CDN is provided in the head to satisfy
              Google Tag Manager (GTM) requirements.
@@ -171,7 +173,7 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
 {% block javascript %}
 <!--[if gt IE 8]><!-->
 {# Include site-wide JavaScript. #}
-<script type='text/javascript' src=" {{ get_js_bundle_name( '/js/routes/*.js' ) }} "></script>
+<script type='text/javascript' src=" {{ get_js_bundle_name( '/js/routes/common.*.js' ) }} "></script>
 
 {# Remove after sheer has been completely phased out #}
 {% if not page %}
--- a/cfgov/jinja2/v1/_layouts/base.html
+++ b/cfgov/jinja2/v1/_layouts/base.html
@@ -106,8 +106,6 @@
     </script>
     <![endif]-->
 
-    <!--[if IE 9]><script src="{{ static('js/ie/common.ie.js') }}"></script><![endif]-->
-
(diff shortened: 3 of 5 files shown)


jimmynotjim:
This looks good to me, but I'd love to get a BEWD to weigh in on the production question.

cfgov/cfgov/settings/base.py
```diff
@@ -152,6 +152,11 @@
 
 ALLOWED_HOSTS = ['*']
 
+#Enabling compression for use in base.html
+COMPRESS_ENABLED = True
```
rosskarchner: If we're enabling compression in base, maybe we need to disable in local.py? I imagine it makes Javascript debugging harder (not an expert here, though)

rosskarchner:
I think we need to add the extension to our Jinja2 environment, per:
http://django-compressor.readthedocs.org/en/latest/jinja2/#plain-jinja2

That would probably happen on this line:
https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/sheerlike/__init__.py#L82

sebworks:
@rosskarchner, I think we added it here

https://github.com/cfpb/cfgov-refresh/blob/e6a99dd9990697da950b22d76e76e4ce6dbc7065/cfgov/v1/__init__.py#L15

rosskarchner:
@sebworks ah, OK-- sorry for missing that.

cfgov/cfgov/settings/base.py
sebworks: I like it as a file because I can then use https://chrome.google.com/webstore/detail/pretty-beautiful-javascri/piekbefgpgdecckjcpffhnacjflfoddg.

@anselmbradford  what do you think?

anselmbradford: So a few things to consider here: The JS is already being compressed at build-time (see the before PR screenshot) by webpack, so this adds the concatenation and caching of the JS file, with the consequence that the compression is being applied to an already compressed JS file at runtime (instead of build-time) and is creating a separate request for the JS file. I'm not an expert on caching, but if the page itself (with inline compressed JS) is cached, it may be less performant to run compression again and have a separate request for the JS. However, the page content will likely be updated more often than the JS so it does make sense to separate it out. 
On the local.py question, if `DEBUG` is `True`, the docs seem to indicate that `COMPRESS_ENABLED` would be cancelled out.
I suspect webpack will be more flexible in how it handles the JS compression, and it also happens at build-time. So alternatively it might make sense to have webpack concatenate the common.js and page-specific JS files instead and place them at a URL that's loaded in. However, this depends on what django does in regard to caching. Can we place JS at a URL and have django do caching magic without further compression?

sebworks: Yep, working on a fix to allow caching but turning off the compression. Also agree that it would be better to handle the bundle generation Webpack, if possible.

sebworks:
@anselmbradford , I've updated the PR to use webpack for the common.js build. I'm not sure if it's possible to do that for the wagtail pages. Let me know if you know of a way to do that.

kave:
:+1:  looks good to me. Is it possible to also compress the page specific js? 

screen shot 2016-03-04 at 10 58 11 am

sebworks:
hmmm, it should be compressed. Restart your server and run `gulp build`.

config/webpack-config.js
```diff
@@ -27,13 +27,11 @@ module.exports = {
...
     new webpack.optimize.CommonsChunkPlugin( COMMON_BUNDLE_NAME ),
     new webpack.optimize.CommonsChunkPlugin( COMMON_BUNDLE_NAME,
-                                             [ COMMON_BUNDLE_NAME ] ),
+                                             'common.[chunkhash].js' ),
     // Change warnings flag to true to view linter-style warnings at runtime.
     new webpack.optimize.UglifyJsPlugin( {
       compress: { warnings: false }
     } ),
-    // Wrap JS in raw Jinja tags so included JS won't get parsed by Jinja.
-    new BannerFooterPlugin( '{% raw %}', '{% endraw %}', { raw: true } )
```
anselmbradford: Woo! Looks like this can be removed from package.json: https://github.com/cfpb/cfgov-refresh/blob/flapjack/package.json#L11

anselmbradford:
Sorry didn't notice myself tagged in this! Reviewing now..

config/webpack-config.js
```diff
@@ -27,13 +27,11 @@ module.exports = {
   plugins: [
     new webpack.optimize.CommonsChunkPlugin( COMMON_BUNDLE_NAME ),
     new webpack.optimize.CommonsChunkPlugin( COMMON_BUNDLE_NAME,
-                                             [ COMMON_BUNDLE_NAME ] ),
+                                             'common.[chunkhash].js' ),
```
anselmbradford: So how's this working now? It generates `common.js` and `common.*.js` that are duplicates of each other, is that right? Can we just spit out the hashed version?

One consequence of this I'm seeing is `/static_built/` starts to build up a number of hashed common.js files. We may want to include a step in backend.sh or frontend.sh to flush out that directory, if that doesn't currently happen.

anselmbradford:
> I'm not sure if it's possible to do that for the wagtail pages. Let me know if you know of a way to do that.

If it ends up in /static_built/ it'll be used in wagtail, no?

config/webpack-config.js
sebworks: > So how's this working now? It generates common.js and common.*.js that are duplicates of each other, is that right? Can we just spit out the hashed version?

So it's only generating the hashed version. The problem is the directory doesn't get cleaned. If you `gulp clean && gulp build`, then you should only see the hashed version.

sebworks:
> If it ends up in /static_built/ it'll be used in wagtail, no?

So the problem is that you don' t know which combinations of `on-demand` bundles your will be using for a particular page in advance. We just loop the page.media array and include the particular bundles in `base.html`.

anselmbradford:
> So the problem is that you don' t know which combinations of on-demand bundles you will be using for a particular page in advance. We just loop the page.media array and include the particular bundles in base.html.

Ahh, right. So theoretically the best would be to have wagtail run webpack on a click of save? I recall you, me, and @kave talking about that in the fall actually.

sebworks:
If we could connect to the database then we could do it in advance.

sebworks:
@kave , can we connect to the database to get the particular atomic elements on a page?

kave:
@anselmbradford yeah we had looked into this but decided calling the gulp command with every save was overkill, and decided to just gather the page elements `js` on page serve.

kave:
@sebworks what problem is it your trying to solve? because couldn't you just compress the individual js files in the `on-demand` folder ?

sebworks:
@kave, so we are trying to create the Wagtail page bundles in Webpack(in advance) and alleviate the need to use the Django compress tags.

kave:
@sebworks yea so that wouldn't be necessary anymore they way we pull in the page specific js. as long as the js in `routes/on-demand` are compressed and retain their original names then it will work

richaagarwal:

sebworks:
I'm going to close this for now. We should revisit it at a later time.

PR #1410, opened 2016-02-05
Set case insensitive tag lookup for tag name completion in the admin

kurtw:
Opened a PR with torchbox's modelcluster to fix their add tag mechanism because it doesn't actually reflect this setting. https://github.com/torchbox/django-modelcluster/pull/52

@rosskarchner


kave:
:+1:

PR #1416, opened 2016-02-05
Global search - keyboard + hiding

anselmbradford:
## Changes

- Updates Global search to:
  - Handle tabbing through the search
  - Clicking off the search box to close it.
  - Add basic no-js state.

## Testing

- `gulp build`

Keyboard:
- Resize to mobile size.
- tab through to highlight search magnifying glass, continue tabbing to pass by hidden search input.
- click in beta banner.
- tab through to highlight search magnifying glass.
- Press spacebar to expand menu.
- Press tab to tab off of global search and observe it closes.
- Repeat for desktop

Mouse:
- Resize to mobile size.
- Click magnifying glass to expand search.
- Click off global search element to close it.
- Repeat for desktop.

## Review

- @KimberlyMunoz 
- @sebworks 
- @jimmynotjim 

## Screenshots

search

## Todos

- Basic no-js state is present but stacking and padding will need to be added when no-js hamburger menu is added.

### DIFF
--- a/cfgov/unprocessed/css/cf-enhancements.less
+++ b/cfgov/unprocessed/css/cf-enhancements.less
@@ -1369,11 +1369,7 @@ textarea {
   patterns:
     - name: Utility class
       markup: |
-        <h1>
-            <a href="#">
-                <span class="u-hidden">Not shown</span>
-            </a>
-        </h1>
+        <span class="u-hidden">Not shown</span>
       codenotes:
         - .u-hidden;
       notes:
@@ -1386,6 +1382,29 @@ textarea {
 
 .u-hidden {
     display: none;
+}
+
+
+/* topdoc
+  name: Invisible
+  family: cf-core
+  patterns:
+    - name: Utility class
+      markup: |
+        <span class="u-invisible">Not shown</span>
+      codenotes:
+        - .u-invisible;
+      notes:
+        - "Use this class to make something invisible.
+           Useful for hiding off-screen elements that
+           are part of an animation, but you don't want
+           as part of the tabindex."
+  tags:
+    - cf-core
+*/
+
+.u-invisible {
+    visibility: hidden;
 }
 
 
--- a/cfgov/unprocessed/css/molecules/global-search.less
+++ b/cfgov/unprocessed/css/molecules/global-search.less
@@ -94,21 +94,7 @@
     - cfgov-molecules
 */
 
-.m-global-search {
-
-    // Absolute position when there is a flyout at mobile/tablet sizes.
-    .respond-to-max( @bp-sm-max, {
-        // Make the height large so that the overflowing flyout can be hidden,
-        // but not get cropped when it's visible.
-        height: 500px;
-        width: 100%;
-        position: absolute;
-        top: 0;
-        overflow-x: hidden;
-        // TODO: This will be ignored in IE8-10,
-        //       leading to the overflow preventing clicks on the links.
-        pointer-events: none;
-    } );
+.js .m-global-search {
 
     &_trigger {
         height: unit( 45px / 18px, em );
@@ -134,21 +120,50 @@
             content: @cf-icon-search;
             font-size: unit( 20px / 18px, em );
         }
+    }
 
-        // Show "Search" text in trigger at tablet sizes.
-        .respond-to-min( @bp-sm-min {
-            &-label:before {
-                .webfont-medium();
-                content: 'Search';
+    &_content {
+        padding-top: unit( @grid_gutter-width / 4 / @base-font-size-px, em );
+
+        transform: translateX( 100% );
+        transition: transform 0.25s ease-out;
+
+        &[aria-expanded="true"] {
+            display: block;
+
+            transform: translateX( 0 );
+        }
+
+        &-suggestions {
+            display: none;
+
+            .h5,
+            .list {
+                display: inline-block;
             }
-        } );
+        }
+    }
 
-        // Only show flyout and close trigger at tablet/mobile sizes.
-        .respond-to-max( @bp-sm-max, {
+    // Tablet size and below.
+    .respond-to-max( @bp-sm-max, {
+        // Absolute position when there is a flyout at mobile/tablet sizes.
+        // Make the height large so that the overflowing flyout can be hidden,
+        // but not get cropped when it's visible.
+        height: 500px;
+        width: 100%;
+        position: absolute;
+        top: 0;
+        overflow-x: hidden;
+        // TODO: This will be ignored in IE8-10,
+        //       leading to the overflow preventing clicks on the links.
+        pointer-events: none;
+
+        &_trigger {
             height: unit( 60px / 18px, em );
             min-width: unit( 60px / 18px, em );
             padding-top: unit( @grid_gutter-width / 2 / 18px, em );
             padding-bottom: unit( @grid_gutter-width / 2 / 18px, em );
+            display: block;
 
             &[aria-expanded="true"] {
                 background: @gray-10;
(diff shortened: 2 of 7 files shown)


cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -267,6 +258,22 @@
     } );
 }
 
+.m-global-search {
+    // Tab trigger is used to capture press of the tab key so that
+    // global search can be collapsed when it hits this element.
```
jimmynotjim: Just so I'm clear, the user tabs through the entire menu, would normally tab to the next item after the menu, but instead hits this hidden tab-trigger, which has an event observer that in turn closes the menu?

cfgov/unprocessed/js/molecules/GlobalSearch.js
```diff
@@ -37,82 +41,98 @@ function GlobalSearch( element ) {
...
-   * tabbing between the input and the search button.
-   * @param {KeyboardEvent} event The event object for the keyboard key press.
+   * Event handler for when there's a click on the page's body.
+   * Used to close the global search, if needed.
+   * @param {MouseEvent} event The event object for the mousedown event.
    */
-  function _handleTabPress( event ) {
-    if ( event.keyCode === KEY_TAB ) {
-      _tabPressed = true;
+  function _handleBodyClick( event ) {
```
jimmynotjim: This would be really useful for the main menu too.

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: That is correct. I guess it's a little weird in that after the menu closes the focus jumps back up to the beginning of the page. Not sure if that's an issue as it's still all tab-accessible (e.g. after collapsing, continue tabbing through the page past the search button). Perhaps focus should jump back to the closed search button? I'll push a fix for this.

This approach is to avoid a lot of complicated logic on tabbing behavior. The search box needs to close after tabbing from the search button at desktop and mobile size, but needs to instead close after the last of the suggested topics links at tablet size. I believe once the bewd-side comes in, the suggested topics links will not be static. Therefore there'd need to be logic around adding and removing events at different breakpoints and on the bewd-side to identify the last link, if present.

cfgov/unprocessed/js/molecules/GlobalSearch.js
```diff
@@ -122,11 +142,39 @@ function GlobalSearch( element ) {
...
+
+  /**
+   * Event handler for when FlyoutMenu expand transition begins.
+   * Use this to perform post-expandBegin actions.
+   */
+  function _handleExpandBegin() {
+    _contentDom.classList.remove( 'u-hidden' );
+
+    // TODO: Refactor to remove totally infuriating
+    //       bug where select() won't fire without timeout.
```
jimmynotjim: That's a weird bug. Could you select on animation end instead?

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/commit/2c8ab804eaa09543b3a569e93172906825e5c85b

cfgov/unprocessed/js/molecules/GlobalSearch.js
anselmbradford: Yeah, I'm planning to use that there too.

cfgov/unprocessed/js/molecules/GlobalSearch.js
anselmbradford: The CSS would need to be changed from a transition to an animation I think. I can explore in another PR.

cfgov/unprocessed/css/molecules/global-search.less
jimmynotjim: I like it, it's a smart work around to a wicked complicated set of interactions. Nice work!

cfgov/unprocessed/js/molecules/GlobalSearch.js
jimmynotjim: Yeah, doesn't need to be addressed now, I was just throwing out a possible solution.

jimmynotjim:
Should there be an animation on desktop-ish sizes? It looks like it just jumps from closed to open and in reverse.

sebworks:
ditto Jimmys comment about animating on desktop.

KimberlyMunoz:
> tab through to highlight search magnifying glass, continue tabbing to pass by hidden search input.

This step is technically a WCAG failure, but I don't think your PR introduced it.

Clicking on the menu, then clicking on the search icon shows no visible changes. When I click on the menu "x" both things are dismissed, making it seem like the search never appeared. It might be worth talking with UX about this interaction because I think it could be fixed a couple ways.

The search box toggle works really well on Voiceover and through keyboard tabbing.

anselmbradford:
> Should there be an animation on desktop-ish sizes? It looks like it just jumps from closed to open and in reverse.

Fixed in https://github.com/cfpb/cfgov-refresh/commit/ca0e0928d19d01db72ae01af699afd4efb202384

> Clicking on the menu, then clicking on the search icon shows no visible changes. When I click on the menu "x" both things are dismissed, making it seem like the search never appeared. It might be worth talking with UX about this interaction because I think it could be fixed a couple ways.

These will slide back and forth. Search from the right, mega menu from the left. I just tried to isolate the global search behavior here and haven't included the mega menu behavior.

cfgov/unprocessed/js/molecules/GlobalSearch.js
```diff
@@ -17,19 +19,23 @@ var breakpointState = require( '../modules/util/breakpoint-state' );
...
 
   var BASE_CLASS = 'm-global-search';
 
   var _dom =
     atomicCheckers.validateDomElement( element, BASE_CLASS, 'GlobalSearch' );
-  var _triggerDom = _dom.querySelector( '.' + BASE_CLASS + '_trigger' );
+  var _triggerSel = '.' + BASE_CLASS + '_trigger';
+  var _triggerDom = _dom.querySelector( _triggerSel );
+  var _flyoutMenu = new FlyoutMenu( _dom,
+                                    _triggerSel,
```
sebworks: Probably better to place one new line.

sebworks:
Looks good. :+1: . It doesn't always animate out when tabbing on mobile, but maybe that can be addressed later.

cfgov/unprocessed/css/cf-enhancements.less
```diff
@@ -1390,6 +1390,33 @@ textarea {
...
 
 /* topdoc
+  name: Invisible
+  family: cf-core
+  patterns:
+    - name: Utility class
+      markup: |
+        <h1>
+            <a href="#">
+                <span class="u-invisible">Not shown</span>
```
jimmynotjim: Should this class be on the `a`? Cause that'll still be in the tab index.

anselmbradford:
> It doesn't always animate out when tabbing on mobile, but maybe that can be addressed later.

Yeah, I have a poorly worded todo for that [here](https://github.com/cfpb/cfgov-refresh/commit/dd16fd5cb009b6a77f88a32bc6f49b120ea0f683#diff-6b1721a3e0398b4a6b4db0f766fc6206R176). Looks like it's a reflow issue again. It may be worth trying an animation instead of a transition and see if there's a difference.

cfgov/unprocessed/css/cf-enhancements.less
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/commit/d152faa87c74af775afa753df5bec8557e7fe1ca

cfgov/unprocessed/js/molecules/GlobalSearch.js
anselmbradford: fixed in https://github.com/cfpb/cfgov-refresh/commit/b23f666702bb8874c9b9e7daf93185075dd1a000

jimmynotjim:
:+1:

KimberlyMunoz:
:+1:

Issue #1366, opened 2016-01-25
Spacing across different pages / elements


ajbush:
There are several pages that seem to have display bugs with spacing. I think the best way to capture the issues I've seen thus far regarding spacing is to use this issue to document appearances of spacing issues and to keep track of where they were found. At this time I'm not sure if these are global issues or just isolated occurrences but the goal is to be as succinct as possible by using mockups.

**Click links below to view marked up screenshots.**

[Consumer Help](https://cloud.githubusercontent.com/assets/6931351/12563112/99b6d162-c376-11e5-8663-ab4e094b7e92.png)
[Credit Card Data](https://cloud.githubusercontent.com/assets/6931351/12563125/ab709b22-c376-11e5-8b3e-546bafaa9538.png)
[Data and research](https://cloud.githubusercontent.com/assets/6931351/12563134/b7f167e6-c376-11e5-9ace-ec8d7a26a17d.png)
[For communities and educators](https://cloud.githubusercontent.com/assets/6931351/12563145/c14b5216-c376-11e5-80f1-ff24f465a276.png)
[Managing Credit](https://cloud.githubusercontent.com/assets/6931351/12563161/d4473d62-c376-11e5-9290-ce73393b6723.png)
[Planning for retirement](https://cloud.githubusercontent.com/assets/6931351/12563173/e064baac-c376-11e5-9a17-b7bee86373fe.png)
[Protecting Against Credit Discrimination](https://cloud.githubusercontent.com/assets/6873734/12590760/89353e84-c432-11e5-8bdf-625f1d8c98a3.png)
[Protecting Against Fraud](https://cloud.githubusercontent.com/assets/6873734/12597202/ec23091c-c451-11e5-808e-7da64b87e1d1.png)
[Sending Money Abroad](https://cloud.githubusercontent.com/assets/6873734/12597329/9ddc1f68-c452-11e5-94eb-adcb9952fff8.png)
[Information for Servicemembers](https://cloud.githubusercontent.com/assets/6873734/12592131/1b0c36c2-c439-11e5-8d02-5f419d9cfeb8.png)
[Money Tips for Servicemembers](https://cloud.githubusercontent.com/assets/6873734/12592462/78e7276a-c43a-11e5-90e8-a2c86c71307b.png)
[...]

PR #1421, opened 2016-02-08
Bringing the layouts in line with spacing standards

### LINKED ISSUE #1396: General atomic element spacing guidlines
I collected some general spacing thoughts for spacing between molecules and organisms that are consistent across page types. I thought I would see if this helped and make sure we were all on the same page with these before we have more discussion

#### Standard spacing (no rule line)
screen shot 2016-02-01 at 1 00 32 pm
* 30px above 
 * This is true of mols/orgs under heroes as well
 * unless it is a text introduction under a breadcrumb which would be [handled like this](https://github.cfpb.gov/flapjack/Modules-V1/wiki/Breadcrumbs)
* 60px after
 * This is true when the mol/org is the last piece to a section. The next element should be a rule line or mol/org of a new section.
* This is true for the last mol/org on a page before the footer. *(there should always be 60px between the last thing on the page and the footer)*
* **These spacings are not additive! There should not be more than 60px between mol/orgs unless there is a rule line (see below)**

#### Spacing with rule line divider between sections
screen shot 2016-02-01 at 1 00 47 pm
* The rule line acts a mol/org without spacing of it's own 
* mol/org before rule line has standard 60px after, the rule line, and the mol/org that follows the rule has the standard 30px above. 

#### Spacing between multiple 25/75 molecules
screen shot 2016-02-01 at 1 00 53 pm
* The [specs for the 25/75](https://github.cfpb.gov/flapjack/Modules-V1/wiki/25-75-Image-&-Text) do not reflect this but we use this molecule commonly without rule lines on top and bottom. The spacing in the spec is still generally correct without the rules.
* An update that was made in build was allowing for 30px of space between 25/75 molecules where there were multiples of these in the same section stacked on top of each other. 
 * The last 25/75 molecule should still have 60px of space after. 
 * This spacing should not be used if the multiple 25/75 molecules are not part of the same section or are not intended to be grouped. 
 * I do not think any other molecules abide by this spacing structure. 

#### Other instances
* Sidebar mol/orgs abide by the same spacing structure and the slug headings are considered part of a given sidebar mol/org. 
* Full width text molecules follow this structure when the mol/org  above or below is not another full width text molecule. 
 * When multiple full width text molecules are next to each other the spacing should reflect the [web type hierarchy spacing standards](https://github.com/cfpb/design-manual/issues/269#issuecomment-157544901)

@ajbush and @schaferjh - do these guidelines follow your thinking for our general approach to spacing elements on all page types? Are there things  that are missing or need to be added? Do any of these sound incorrect?

KimberlyMunoz:
Fixes a number of spacing issues, specifically with the text-image molecules, the first block of the page on several templates and spacing at the end of pages.

Fixes #1396 and a lot of #1366 

## Review
- @jimmynotjim 
- @anselmbradford 
- @sebworks

### DIFF
--- a/cfgov/unprocessed/css/organisms/image-text-50-50-group.less
+++ b/cfgov/unprocessed/css/organisms/image-text-50-50-group.less
@@ -16,13 +16,13 @@
     - cfgov-molecules
 */
 .o-image-text-50-50-group {
-    // TODO: Find another way to handle spacing
+    // TODO: Find another way to handle spacing here and in link-blob-group
     margin-bottom: unit( @grid_gutter-width * -2 / @base-font-size-px, em );
 
     > h2 {
-          .heading-2();
-          margin-left: unit( @grid_gutter-width / 2 / @font-size, em );
-          margin-right: unit( @grid_gutter-width / 2 / @font-size, em );
+        .heading-2();
+        margin-left: unit( @grid_gutter-width / 2 / @font-size, em );
+        margin-right: unit( @grid_gutter-width / 2 / @font-size, em );
     }
 }
 
--- a/cfgov/jinja2/v1/browse-basic/index.html
+++ b/cfgov/jinja2/v1/browse-basic/index.html
@@ -38,16 +38,22 @@
     {{ page.seo_title }}
 {%- endblock %}
 
-
 {% block content_main_modifiers -%}
     {{ super() }} content__flush-bottom
 {%- endblock %}
 
 {% block content_main %}
     {% for block in page.header -%}
-        <div class="block">
-            {{ render_stream_child(block) }}
-        </div>
+        {% if loop.index == 1 %}
+            <div class="block
+                        block__flush-top">
+                {{ render_stream_child(block) }}
+            </div>
+        {% else %}
+            <div class="block">
+                {{ render_stream_child(block) }}
+            </div>
+        {% endif %}
     {%- endfor %}
 
     {% for block in page.content -%}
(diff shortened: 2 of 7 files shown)


anselmbradford:
Code looks good :+1: How best to test this?

One trivial point, you could stack the classes that call super:
```
{{ super() }}
content__flush-bottom
```

cfgov/unprocessed/css/organisms/image-text-50-50-group.less
```diff
@@ -16,13 +16,13 @@
     - cfgov-molecules
 */
 .o-image-text-50-50-group {
-    // TODO: Find another way to handle spacing
+    // TODO: Find another way to handle spacing here and in link-blob-group
```
jimmynotjim: I've got an idea for this I'm testing on the homepage. Now that it's merged, I'm going to extend the info-unit stuff here.

KimberlyMunoz:
Testing is going to be a pain. The best way I could see is to create templates with one molecule on each page and checking the spacing for the 25-75, 50-50, and link blob specifically.

cfgov/unprocessed/css/organisms/image-text-50-50-group.less
KimberlyMunoz: Discussed offline We'll get this in another PR.

jimmynotjim:
Am I the only one that can still test the demo pages? Built and tested there and it looks good! :+1:

sebworks:
:

PR #1422, opened 2016-02-08
Color and display fix for global header cta

anselmbradford:
## Changes

- Turns out the global header cta (and mega menu) shouldn't have visited link colors. This fixes the colors in the global header cta.
- Sets the padding on the link itself so that the "focus" state occupies the whole padded area.

## Testing

- `gulp build`
- Check global header cta in header. It should always be pacific blue and shouldn't have a visited link color.

## Review

- @jimmynotjim 
- @KimberlyMunoz 
- @sebworks 

## Screenshots

screen shot 2016-02-08 at 12 09 09 pm

### DIFF
--- a/cfgov/unprocessed/css/molecules/global-header-cta.less
+++ b/cfgov/unprocessed/css/molecules/global-header-cta.less
@@ -34,6 +34,8 @@
 
     a {
         .webfont-medium();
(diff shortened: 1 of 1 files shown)


jimmynotjim:
Working locally :+1:

PR #1428, opened 2016-02-09
Removing bottom spacing from last li in a bulleted list

KimberlyMunoz:
Part of the work to bring spacing into Design's spacing specs including fixes cases where the spacing at the bottom of a list adds to the spacing at the bottom of a group. I've fixed this by adding a cf-enhancements rule to remove the padding.

**Before**
image

**After**
image

## Review
- @jimmynotjim 
- @sebworks 
- @anselmbradford

### DIFF
--- a/cfgov/unprocessed/css/cf-enhancements.less
+++ b/cfgov/unprocessed/css/cf-enhancements.less
@@ -1585,6 +1585,17 @@ select[multiple] {
 }
 
 /* topdoc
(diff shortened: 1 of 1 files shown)


jimmynotjim:
Should we wait till this is released? https://github.com/cfpb/capital-framework/pull/259/files

KimberlyMunoz:
I wasn't sure if that already made it into CF or not. This can wait for the CF changes to make it in.

PR #1430, opened 2016-02-09
implemented fix for xss attack possibility in login redirects

kave:
Django `is_safe` would still allow with `http://` in the redirect which left us at risk. resolved by only allowing  https://www.djangoproject.com/weblog/2015/mar/18/security-releases/

@kurtw 
@rosskarchner 
@richaagarwal

### DIFF
--- a/cfgov/v1/views.py
+++ b/cfgov/v1/views.py
@@ -1,4 +1,4 @@
-import time
+import time, re
 from datetime import timedelta
 
 from core.services import PDFGeneratorView, ICSView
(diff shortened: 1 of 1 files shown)


cfgov/v1/views.py
```diff
@@ -147,7 +147,13 @@ def login_with_lockout(request, template_name='wagtailadmin/login.html'):
     """
     Displays the login form and handles the login action.
     """
-    redirect_to = request.GET.get(REDIRECT_FIELD_NAME, '')
+    redirect_to = request.POST.get(REDIRECT_FIELD_NAME,
```
kave: reverted my original change here, realized i introduced a bug where it would lose the redirect url during the Post request

rosskarchner:
Looks good-- I tested with `http://127.0.0.1:8000/admin/login/?next=http://google.com`-- is that adequate?

kave:
@rosskarchner yes that correct. thats the main case we want to avoid

PR #1435, opened 2016-02-10
Spin off ClearableInput behavior

anselmbradford:
Spins clearable input behavior out of the Global Search molecule.

## Additions

- Adds `ClearableInput` class for encapsulating `input-contains-label` clearable label behavior.

## Changes

- Updates Global Search to use ClearableInput instance.

## Testing

- `gulp build`
- Expand global search.
- Type something.
- Click (x) clear.
- Try at different screen sizes.
- Should work and be error-free.

## Review

- @KimberlyMunoz 
- @sebworks 
- @jimmynotjim 

## Todos

- Note that since the clear label is a label of input, it is not directly tab accessible as tabbing to the clear button automatically moves to the associated input. I think this is okay because the delete key can be used to remove the input? This is existing behavior.

### DIFF
--- a/cfgov/unprocessed/js/modules/ClearableInput.js
+++ b/cfgov/unprocessed/js/modules/ClearableInput.js
@@ -1,0 +1,96 @@
+'use strict';
+
+// Required polyfills for IE9.
+if ( !Modernizr.classlist ) { require( '../modules/polyfill/class-list' ); } // eslint-disable-line no-undef, global-require, no-inline-comments, max-len
+
+// Required modules.
+var atomicCheckers = require( '../modules/util/atomic-checkers' );
+
+/**
+ * ClearableInput
+ * @class
+ *
+ * @classdesc Initializes a new ClearableInput molecule.
+ *
+ * @param {HTMLNode} element
+ *   The DOM element within which to search for the molecule.
+ * @returns {Object} A ClearableInput instance.
+ */
+function ClearableInput( element ) {
+  var BASE_CLASS = 'input-contains-label';
+
+  var _dom =
+    atomicCheckers.validateDomElement( element, BASE_CLASS, 'ClearableInput' );
+  var _inputDom = _dom.querySelector( 'input' );
+  var _clearBtnDom = _dom.querySelector( '.' + BASE_CLASS + '_after__clear' );
+
+  var _isClearShowing = true;
+
+  /**
+   * @returns {Object} The ClearableInput instance.
+   */
+  function init() {
+    _clearBtnDom.addEventListener( 'mousedown', _clearClicked );
+    _inputDom.addEventListener( 'keyup', _inputTyped );
+
+    _setClearBtnState( _inputDom.value );
+
+    return this;
+  }
+
+  /**
+   * Event handler for when the clear input label was clicked.
+   * @param {MouseEvent} event - The event object for the mousedown event.
+   */
+  function _clearClicked( event ) {
+    _inputDom.value = _setClearBtnState( '' );
+    _inputDom.focus();
+
+    // Prevent event bubbling up to the input, which would blur otherwise.
+    event.preventDefault();
+  }
+
+  /**
+   * Event handler for when the user typed in the input.
+   */
+  function _inputTyped() {
+    _setClearBtnState( _inputDom.value );
+  }
+
+  /**
+   * @param {string} value - The input value in the search box.
+   * @returns {string} The input value in the search box.
+   */
+  function _setClearBtnState( value ) {
+    if ( _isClearShowing && value === '' ) {
+      _hideClearBtn();
+    } else if ( !_isClearShowing ) {
+      _showClearBtn();
+    }
+    return value;
+  }
+
+  /**
+   * Add a hidden class to the input search label.
+   * Used when there is no text input.
+   */
+  function _hideClearBtn() {
+    _clearBtnDom.classList.add( 'u-hidden' );
+    _isClearShowing = false;
+  }
+
+  /**
+   * Remove a hidden class from the input search label.
+   * Used when there is text input.
+   */
+  function _showClearBtn() {
+    _clearBtnDom.classList.remove( 'u-hidden' );
+    _isClearShowing = true;
+  }
+
+  this.init = init;
+
+  return this;
+}
+
+module.exports = ClearableInput;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,9 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 
 ### Added
 - Added Backend sidebar contact
-- Add Related Metadata molecule to backend 
+- Added Related Metadata molecule to backend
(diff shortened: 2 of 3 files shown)


cfgov/unprocessed/js/modules/ClearableInput.js
```diff
@@ -0,0 +1,92 @@
...
+ *   The DOM element within which to search for the molecule.
+ * @returns {Object} A ClearableInput instance.
+ */
+function ClearableInput( element ) {
+  var BASE_CLASS = 'input-contains-label';
+
+  var _dom =
+    atomicCheckers.validateDomElement( element, BASE_CLASS, 'ClearableInput' );
+  var _inputDom = _dom.querySelector( 'input' );
+  var _clearBtnDom = _dom.querySelector( '.' + BASE_CLASS + '_after__clear' );
```
sebworks: is this BEM format?

cfgov/unprocessed/js/modules/ClearableInput.js
```diff
@@ -0,0 +1,92 @@
...
+      _showClearBtn();
+    }
+    return value;
+  }
+
+  /**
+   * Add a hidden class to the input search label.
+   * Used when there is no text input.
+   */
+  function _hideClearBtn() {
```
sebworks: Why create these functions instead of just setting the classes in `_setClearBtnState`?

cfgov/unprocessed/js/modules/ClearableInput.js
jimmynotjim: Related question, should there be a js only hook instead of using the style hook?

cfgov/unprocessed/js/modules/ClearableInput.js
jimmynotjim: Agreed, unless that function will be doing more in the future, the class list manipulation can live in `_setClearBtnState`

cfgov/unprocessed/js/modules/ClearableInput.js
```diff
@@ -0,0 +1,92 @@
...
+    atomicCheckers.validateDomElement( element, BASE_CLASS, 'ClearableInput' );
+  var _inputDom = _dom.querySelector( 'input' );
+  var _clearBtnDom = _dom.querySelector( '.' + BASE_CLASS + '_after__clear' );
+
+  /**
+   * @returns {Object} The ClearableInput instance.
+   */
+  function init() {
+    _clearBtnDom.addEventListener( 'mousedown', _clearClicked );
+    _inputDom.addEventListener( 'keyup', _inputTyped );
```
jimmynotjim: Should we add a timeout so this isn't triggered a million times in a row?

cfgov/unprocessed/js/modules/ClearableInput.js
anselmbradford: It's so that the implementation for how the clear button is shown/hidden is hidden from where it is called. The clear button could be moved off screen, be set invisible, set to hidden, etc. and the method call signature of where it was used wouldn't change. Also, having `add` to hide the label seems less clear than saying "hideClearBtn." Probably overkill in this case though, I can update it.

cfgov/unprocessed/js/modules/ClearableInput.js
anselmbradford: > is this BEM format?

I think so? Here's the breakdown I see:

`input-contains-label_after__clear`
`input-contains-label` = Block
`_after` = Element
`__clear` = Modifier

> Related question, should there be a js only hook instead of using the style hook?

What're the rules around doing this again, and we were going to move to `data-js-hook` attributes instead?

sebworks: yeah, you don't have to do it in this PR. It should be apart of the JS cleanup we have to do.

cfgov/unprocessed/js/modules/ClearableInput.js
anselmbradford: Not sure how to get around this. We need it to respond immediately on the press of the delete key that removes everything, so we need to be checking whether each key press is the one to remove all content.

jimmynotjim: I'm not understanding the significance of the delete key. `_inputTyped()` is setting the button state as either shown or hidden. My suggestion is to add a timeout on keyup to give us time to check if another key is pressed, so that it's not `show, show, show, show, show` if you were to enter `test` in the search field.

anselmbradford: Say there's a `t` in the input field, the user could either press any key (e.g. `e`) or they may press the delete key. If it's any other key, it's true that `show` will be called again, but if it's the delete key hide would need to be called immediately to hide the clear button when there are no letters. I don't see how a timeout wouldn't introduce scenarios where it would be possible to type fast enough to hit the delete key and have it be ignored. We could have logic that handles the keyup event differently when there's more than one character in the input, but it would need to handle scenarios where the text was selected and delete was hit.

sebworks: @anselmbradford, can we throttle or debounce it?

sebworks: Better yet, why not just check if it's already visible before setting it again?

anselmbradford: So a check would be like this, right:

```
if ( _clearBtnDom.classList.contains( 'u-hidden' ) ) {
    _clearBtnDom.classList.remove( 'u-hidden' );
}
```
In terms of querying the `classList` that doesn't save anything. Perhaps it saves a redraw though. I don't know if attempting to remove a class that isn't there would cause a redraw or not. I'll test it out.

jimmynotjim: Ah right, now I see the delete key dilemma. If you were to delay 200ms or so, the user wouldn't even notice the delay in the clear showing/hiding, but it's just long enough to allow you to type a full word.

@sebworks that could work if you check the key code is delete, but a debounce is probably better

jimmynotjim: Found this https://remysharp.com/2010/07/21/throttling-function-calls

sebworks: Or just keep track of the state like `is_visible`.

anselmbradford: Added flag in https://github.com/cfpb/cfgov-refresh/commit/0b8b87c052b2636ea256c393de78a14ef24916b6

cfgov/unprocessed/js/modules/ClearableInput.js
anselmbradford: I left these in in https://github.com/cfpb/cfgov-refresh/commit/0b8b87c052b2636ea256c393de78a14ef24916b6 because they now do more than set the class. Does this work?

anselmbradford:
Ready for review again! Thanks!

sebworks:
:+1:

jimmynotjim:
:+1:

PR #1443, opened 2016-02-11
Fixing prefooter spacing across project

KimberlyMunoz:
This goes back in the project to make sure spacing in the about pages matches the rest of the wagtail project.

## Test
- Check budget, career and another office page to make sure the spacing matches those on wagtail.

## Review 
- @jimmynotjim 
- @anselmbradford 
- @sebworks

### DIFF
--- a/cfgov/jinja2/v1/offices/_single.html
+++ b/cfgov/jinja2/v1/offices/_single.html
@@ -147,17 +147,20 @@
         {{ office.content | safe }}
     </section>
     {% endif %}
-    <aside>
+    <aside class="block
+                  block__flush-bottom
+                  block__flush-sides
+                  block__bg
+                  block__border-top
+                  block__border-right">
         <div class="block
-                    block__flush-sides
-                    block__flush-bottom
-                    block__border-top
-                    block__border-right">
+                    block__flush-top">
             {% if office.tags %}
                 {% import 'macros/activities-block.html' as activities_block with context %}
                 {{ activities_block.render(office.tags) }}
+        </div>
+        <div class="block">
             {% endif %}
-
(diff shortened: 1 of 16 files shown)


cfgov/jinja2/v1/offices/_single.html
```diff
@@ -147,17 +147,20 @@ <h3 class="h2 qa-resource-title">
         {{ office.content | safe }}
     </section>
     {% endif %}
-    <aside>
+    <aside class="block
```
sebworks: Instead adding all these classes, would it make sense to combine them into one class? The value would be that you aren't just repeating the same block of classes.

KimberlyMunoz: I've seen us go back and forth on whether we want more classes in our markup or a smaller HTML file. I'm okay with either, but would want to make sure we don't go and forth.

sebworks: :1, makes sense.

sebworks:
:+1:

cfgov/jinja2/v1/offices/_single.html
KimberlyMunoz: Discussed in chat and I'm pulling it out since some of it will carry over to the new project and be used in multiple places.

KimberlyMunoz:
Hey @sebworks, could you give it another look? Thanks!

sebworks:
Prefooter is being truncated at mobile widths:
screen shot 2016-02-12 at 2 49 41 pm

KimberlyMunoz:
Fixed!

jimmynotjim:
:+1:

sebworks:
:

PR #1447, opened 2016-02-12
Fix for notifier classes

sebworks:
Updating the notifier markup to use the atomic classes.

## Testing

- Visit  `http://localhost:blog/` and enter invalid email address.

## Review

- @anselmbradford 
- @KimberlyMunoz 
- @jimmynotjim 

## Screenshots
 Before : 

screen shot 2016-02-12 at 9 18 41 am

After:

screen shot 2016-02-12 at 9 18 22 am


anselmbradford:
:+1: Looks like there needs to be more space between "Email address" and the notification, but perhaps that's out of scope here. Also out of scope, but perhaps we should move `cf-icon` into `m-notification_icon`.

Issue #1408, opened 2016-02-04
Filterable List Spacing Issue


KimberlyMunoz:
The results for the filterable list has off spacing. Reported by @kurtw in chat.

## Steps to Replicate

1. Visit [content refresh]/policy-compliance/amicus/briefs/
2. Apply Supreme Court Filter

## Expected Behavior

There should be spacing between the notification and the list of results

## Screenshot

![image](https://cloud.githubusercontent.com/assets/1860176/12833280/98155b2c-cb6d-11e5-918e-9e5b8dcad083.png)

kurtw:
It seems like this is done? @KimberlyMunoz could you confirm and close if so?
![screen shot 2016-03-02 at 10 17 02 am](https://cloud.githubusercontent.com/assets/1412978/13464612/f712d986-e05f-11e5-957d-d9fdb25f6eec.png)

PR #1451, opened 2016-02-12
Adding validation logic for FilterableListControls

sebworks:
Adding validation logic for FilterableListControls

## Testing

-  Goto Wagtail Admin and create a `Browse Filterable Page`
- Add a `Learn Page` as a child of the page you just created.
- Publish the `Browse Filterable Page`. 
- Visit the `Browse Filterable Page` and enter in `043`. You should get an error message.
- Edit the entry to be `04`. The form should now submit. 

## Review

- @anselmbradford 
- @KimberlyMunoz 
- @jimmynotjim 

## Notes

- Need to write unit test and fix JSDoc comments.

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
+++ b/cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
@@ -18,12 +18,14 @@
    ========================================================================== #}
 {% from 'molecules/expandable.html' import expandable with context %}
 
-{% macro _filter_selectable(type, index, label_text, id, name, value) %}
+{% macro _filter_selectable(type, index, label_text, id, name, value, required=None, group=None) %}
     <input class="cf-input"
            type="{{ type }}"
            value="{{ value }}"
            id="filter{{ index }}_{{ id }}"
            name="filter{{ index }}_{{ name }}"
+           {{ 'data-required=' ~ required if required else ''  }}
+           {{ 'data-group=' ~ group if group else ''  }}
            {{ 'checked' if is_filter_selected('filter' ~ index ~ '_' ~ name, value) else '' }}>
     <label for="filter{{ index }}_{{ id }}" >
         {{ label_text if label_text else value }}
@@ -38,6 +40,16 @@
 {% endmacro %}
 
 {% macro _render_filter_fields(controls, form, index) -%}
+     <div class="form-l_col
+                    form-l_col-1">
+            <div class="form-group">
+                <label class="form-label-header" for="email">
+                    Email:
+                </label>
+                 <input data-type='email' id='email' type="text" />
+            </div>
+        </div>
+   
     {% if controls.title %}
         <div class="form-l_col
                     form-l_col-1">
@@ -116,7 +128,7 @@
                                                 form-l_col-1-2">
                                         <div class="form-group">
                                             <label class="form-label-header"
-                                                   for="from_date">
+                                                   for="{{ 'filter' ~ index ~ '_from_date' }}">
                                                 From:
                                             </label>
                                             {{ form.render_with_id(form.from_date,
@@ -127,7 +139,7 @@
                                                 form-l_col-1-2">
                                         <div class="form-group">
                                             <label class="form-label-header"
-                                                   for="to_date">
+                                                   for="{{ 'filter' ~ index ~ '_to_date' }}">
                                                 To:
                                             </label>
                                             {{ form.render_with_id(form.to_date,
--- a/cfgov/unprocessed/css/organisms/filterable-list-controls.less
+++ b/cfgov/unprocessed/css/organisms/filterable-list-controls.less
@@ -33,6 +33,7 @@
     // Add gap between expandable and notification.
     .m-notification {
         margin-top: unit( @grid_gutter-width / @base-font-size-px, em );
+        margin-bottom: unit( @grid_gutter-width / @base-font-size-px, em );
     }
 }
 
--- a/cfgov/unprocessed/js/config/error-messages-config.js
+++ b/cfgov/unprocessed/js/config/error-messages-config.js
@@ -7,13 +7,21 @@
 'use strict';
 
 var ERROR_MESSAGES = {
-  CHECKBOX_ERRORS: {
-    required: 'Please select at least one of the "%s" options.'
+  CHECKBOX: {
+    REQUIRED: 'Please select at least %s of the options.'
   },
-  DATE_ERRORS: {
-    invalid:      'You have entered an invalid date.',
-    one_required: 'Please enter at least one date.'
-  }
+  DATE: {
+    INVALID: 'You have entered an invalid date.',
+    ONE_REQUIRED: 'Please enter at least one date.'
+  },
+  EMAIL: {
+    INVALID: 'You have entered an invalid email address.',
+    REQUIRED: 'Please enter an email address.'
+  },
+  FIELD: {
+    REQUIRED: 'This field is required.'
+  },
+  DEFAULT: 'Error!'
 };
 
-module.exports = ERROR_MESSAGES;
+module.exports = Object.seal( ERROR_MESSAGES );
--- a/cfgov/unprocessed/js/modules/post-filter.js
+++ b/cfgov/unprocessed/js/modules/post-filter.js
@@ -7,7 +7,7 @@
 
 var $ = require( 'jquery' );
 var dateRangeFormatter = require( './util/date-range-formatter' );
-var DATE_ERRORS = require( '../config/error-messages-config' ).DATE_ERRORS;
+var DATE_ERRORS = require( '../config/error-messages-config' ).DATE;
 
 /**
  * Set up jQuery plugin initialization.
@@ -107,24 +107,24 @@ PostFilter.prototype = {
     var fromDateValue = this.$gte.val();
     var toDateValue = this.$lte.val();
     var dateRange = dateRangeFormatter.format( fromDateValue, toDateValue );
-    var error_messages = [];
+    var errorMessages = [];
 
     if ( dateRange && dateRange.isValid ) {
       this.$gte.val( dateRange.startDate );
       this.$lte.val( dateRange.endDate );
     } else {
-      error_messages.push( DATE_ERRORS.invalid );
+      errorMessages.push( DATE_ERRORS.INVALID );
     }
 
     if ( $form.hasClass( 'js-validate_require-date' ) &&
     toDateValue === '' && fromDateValue === '' ) {
-      error_messages.push( DATE_ERRORS.one_required );
+      errorMessages.push( DATE_ERRORS.ONE_REQUIRED );
     }
 
-    if ( error_messages.length ) {
+    if ( errorMessages.length ) {
       event.preventDefault();
       $form.trigger( 'cf_notifier:notify', {
-        message: error_messages.join( '</br>' ),
+        message: errorMessages.join( '</br>' ),
         state:   'error'
       } );
     }
--- a/cfgov/unprocessed/js/modules/util/dom-traverse.js
+++ b/cfgov/unprocessed/js/modules/util/dom-traverse.js
@@ -1,6 +1,5 @@
 'use strict';
 
-// Required polyfills for IE9.
 if ( window.Modernizr && !window.Modernizr.classlist ) { require( '../polyfill/class-list' ); } // eslint-disable-line no-undef, global-require, no-inline-comments, max-len
 
 /**
@@ -48,17 +47,27 @@ function not( elems, exclude ) {
  * Get the nearest parent node of an element.
  *
  * @param {HTMLNode} elem - A DOM element.
- * @param {string} className - CSS classname.
- * @returns {HTMLNode} Nearest parent node that contains className, or null.
+ * @param {string} selector - CSS selector.
+ * @returns {HTMLNode} Nearest parent node that matches the selector.
  */
-function closest( elem, className ) {
+function closest( elem, selector ) {
   elem = elem.parentNode;
-  do {
-    if ( elem.classList && elem.classList.contains( className ) ) {
-      return elem;
+
+  var matchesSelector = elem.matches ||
+                        elem.webkitMatchesSelector ||
+                        elem.mozMatchesSelector ||
+                        elem.msMatchesSelector;
+  var match;
+
+  while ( elem ) {
+    if ( matchesSelector.bind( elem )( selector ) ) {
+      match = elem;
+    } else {
+      elem = elem.parentElement;
     }
-    elem = elem.parentNode;
-  } while ( elem );
+
+    if ( match ) return elem;
+  }
 
   return null;
 }
--- a/cfgov/unprocessed/js/organisms/FilterableListControls.js
+++ b/cfgov/unprocessed/js/organisms/FilterableListControls.js
@@ -2,10 +2,10 @@
 
 // Required modules.
 var atomicCheckers = require( '../modules/util/atomic-checkers' );
-var Expandable = require( '../molecules/Expandable' );
+var ERROR_MESSAGES = require( '../config/error-messages-config' );
+var getClosestElement = require( '../modules/util/dom-traverse' ).closest;
 var Notification = require( '../molecules/Notification' );
-// TODO: Implement form data client-side validation.
-// var validate = require( 'validate' );
+var typeCheckers = require( '../modules/util/type-checkers' );
 
 /**
  * FilterableListControls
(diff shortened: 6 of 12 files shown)


jimmynotjim:
Works well locally, I'm seeing one oddity with the layout of the "Published" text

screen shot 2016-02-16 at 6 04 34 pm

cfgov/unprocessed/js/organisms/FilterableListControls.js
```diff
@@ -2,10 +2,9 @@
```
jimmynotjim: I love all the cleanup you've done in here. :facepunch:

cfgov/unprocessed/js/config/error-messages-config.js
```diff
@@ -7,13 +7,18 @@
...
-  }
+  },
+  EMAIL : {
+    invalid : 'You have entered an invalid email address.',
+    required : 'Please enter an email address.'
+  },
+  DEFAULT: 'Error!'
 };
 
 module.exports = ERROR_MESSAGES;
```
anselmbradford: Since this is constants, maybe `module.exports = Object.seal( ERROR_MESSAGES );`
This could change when L:9 becomes `const ERROR_MESSAGES` at some point.

cfgov/unprocessed/js/config/error-messages-config.js
```diff
@@ -7,13 +7,18 @@
...
 var ERROR_MESSAGES = {
-  CHECKBOX_ERRORS : {
-    required : 'Please select at least one of the "%s" options.'
+  CHECKBOX : {
+    required : 'Please select at least %s of the options.'
   },
-  DATE_ERRORS : {
+  DATE : {
     invalid : 'You have entered an invalid date.',
     one_required : 'Please enter at least one date.'
```
anselmbradford: Linter will complain about snake_case in JS. Also, should these be formatted as constants too (UPPERCASE) ?

cfgov/unprocessed/js/modules/util/dom-traverse.js
```diff
@@ -1,7 +1,6 @@
 'use strict';
 
-// Required polyfills for IE9.
-if ( window.Modernizr && !window.Modernizr.classlist ) { require( '../polyfill/class-list' ); } // eslint-disable-line no-undef, global-require, no-inline-comments, max-len
+require( '../polyfill/class-list' );
```
anselmbradford: Didn't we add `if ( window.Modernizr && !window.Modernizr.classlist )` so class-list polyfill isn't delivered to all browsers?

cfgov/unprocessed/js/modules/util/dom-traverse.js
```diff
@@ -48,23 +47,33 @@ function not( elems, exclude ) {
...
 
   return null;
 }
 
 module.exports = {
-  closest:     closest,
   getSiblings: getSiblings,
-  not:         not
+  not:         not,
+  closest:     closest
```
anselmbradford: Not a big deal, but these were alphabetized.

cfgov/unprocessed/js/organisms/FilterableListControls.js
```diff
@@ -18,157 +17,254 @@ var Notification = require( '../molecules/Notification' );
...
-   * Show error notification.
-  */
-  function _showError() {
-    _notification.setTypeAndContent( _notification.ERROR, 'Error!' );
-    _notification.show();
+   * Set the notification type, msg, and visibility.
+   * @param {String} type The type of notification to display.
+   * @param {String} msg The message to display in the notification.
+   * @param {String} methodName The method to use to control visibility
+                                of the notification.
```
anselmbradford: I believe jsdocs needs an `*` on the blank line and two space indent.

```
* @param {string} methodName The method to use to control visibility
*   of the notification.
```

cfgov/unprocessed/js/organisms/FilterableListControls.js
```diff
@@ -18,157 +17,254 @@ var Notification = require( '../molecules/Notification' );
...
   /**
-   * Show error notification.
-  */
-  function _showError() {
-    _notification.setTypeAndContent( _notification.ERROR, 'Error!' );
-    _notification.show();
+   * Set the notification type, msg, and visibility.
+   * @param {String} type The type of notification to display.
+   * @param {String} msg The message to display in the notification.
+   * @param {String} methodName The method to use to control visibility
```
anselmbradford: Technically would be `string` for string primitives unless it's taking `new String( '…' )`.

cfgov/unprocessed/js/organisms/FilterableListControls.js
```diff
@@ -18,157 +17,254 @@ var Notification = require( '../molecules/Notification' );
...
   }
 
   /**
-   * Show success notification.
-  */
-  function _showSuccess() {
-    _notification.setTypeAndContent( _notification.SUCCESS, 'Success!' );
-    _notification.show();
+   * Determines if you should validate a field.
+   * @param {Field} field A form field.
```
anselmbradford: Type is likely `HTMLNode`

cfgov/unprocessed/js/organisms/FilterableListControls.js
```diff
@@ -18,157 +17,254 @@ var Notification = require( '../molecules/Notification' );
...
+  function shouldValidateField( field, type, isInGroup ) {
+    var isDisabled = field.getAttribute( 'disabled' ) !== null;
+    var isIgnoreType = _defaults.ignoreFieldTypes.indexOf( type ) > -1;
+    var shouldValidate = isDisabled === false && isIgnoreType === false;
+
+    if ( shouldValidate && ( isInGroup || false ) ) {
+      var name = field.getAttribute( 'data-group' ) ||
+                 field.getAttribute( 'name' );
+      var isRequired = field.getAttribute( 'data-required' ) !== null;
+      if ( _fieldGroups.indexOf( name ) > -1 || isRequired === false ) {
```
anselmbradford: Maybe move ` _fieldGroups.indexOf( name ) > -1` to a variable too, similar to `isIgnoreType`.

cfgov/unprocessed/js/organisms/FilterableListControls.js
```diff
@@ -18,157 +17,254 @@ var Notification = require( '../molecules/Notification' );
...
-      valid:   [],
-      invalid: []
+      invalid: [],
+      valid:   []
     };
+    var validatedField;
+
+    _fieldGroups = [];
+
+    fields.forEach( function loopFields( field ) {
```
anselmbradford: I'm more of a fan of using a classic loop to avoid the scope change and overhead of creating a function, but maybe that's just me :-)

cfgov/unprocessed/js/organisms/FilterableListControls.js
```diff
@@ -18,157 +17,254 @@ var Notification = require( '../molecules/Notification' );
...
-      value:  null,
-      label:  null,
-      status: {
-        checkgroup: false
-      }
+  function _validateField( field, isInGroup ) {
+    var emptyStringRegex = /^\s*$/;
+    var value = field.value;
+    var isRequired = field.getAttribute( 'required' ) !== null;
+    var isEmpty = value === '' || value === null ||
```
anselmbradford: Possibly create a utility function to check if a value is empty and call it here.

cfgov/unprocessed/js/organisms/FilterableListControls.js
```diff
@@ -18,157 +17,254 @@ var Notification = require( '../molecules/Notification' );
...
-      label:  null,
-      status: false
-    };
+
+    if ( _validators[type] ) {
+      validation.status = _validators[type]( field, validation );
+    }
+
+    if ( isRequired && isEmpty ) {
+      validation.msg += 'is required.';
```
anselmbradford: Does this need a space before the `is`?

cfgov/unprocessed/js/organisms/FilterableListControls.js
anselmbradford: Here and throughout. `console.dir( field )` in the function should tell you. Not a big deal though.

cfgov/unprocessed/js/config/error-messages-config.js
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/commit/f217937d4eec3537f28765b2f956d29b1609df50

cfgov/unprocessed/js/config/error-messages-config.js
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/commit/f217937d4eec3537f28765b2f956d29b1609df50.

cfgov/unprocessed/js/modules/util/dom-traverse.js
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/commit/f217937d4eec3537f28765b2f956d29b1609df50

cfgov/unprocessed/js/organisms/FilterableListControls.js
sebworks: fixed in https://github.com/cfpb/cfgov-refresh/commit/f217937d4eec3537f28765b2f956d29b1609df50

cfgov/unprocessed/js/organisms/FilterableListControls.js
sebworks: fixed in https://github.com/cfpb/cfgov-refresh/commit/f217937d4eec3537f28765b2f956d29b1609df50

cfgov/unprocessed/js/organisms/FilterableListControls.js
sebworks: I'm using `forEach` here because we have an ESLint has a `no-continue` rule. Generally I would agree with you.

cfgov/unprocessed/js/organisms/FilterableListControls.js
sebworks: fixed in https://github.com/cfpb/cfgov-refresh/commit/f217937d4eec3537f28765b2f956d29b1609df50

cfgov/unprocessed/js/organisms/FilterableListControls.js
sebworks: fixed in https://github.com/cfpb/cfgov-refresh/commit/f217937d4eec3537f28765b2f956d29b1609df50

cfgov/unprocessed/js/organisms/FilterableListControls.js
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/commit/f217937d4eec3537f28765b2f956d29b1609df50

cfgov/unprocessed/js/modules/util/dom-traverse.js
sebworks: Yeah, I mistakenly removed this code. I'm working on PR to remove them as we discussed last week. Fixed in https://github.com/cfpb/cfgov-refresh/commit/f217937d4eec3537f28765b2f956d29b1609df50

cfgov/unprocessed/js/organisms/FilterableListControls.js
sebworks: I need to check on this.

sebworks:
this is ready for review again.

cfgov/unprocessed/css/organisms/filterable-list-controls.less
```diff
@@ -33,6 +33,7 @@
     // Add gap between expandable and notification.
     .m-notification {
         margin-top: unit( @grid_gutter-width / @base-font-size-px, em );
+        margin-bottom: unit( @grid_gutter-width / @base-font-size-px, em );
```
KimberlyMunoz: Fixes #1408!

cfgov/unprocessed/js/modules/post-filter.js
```diff
@@ -107,24 +107,24 @@ PostFilter.prototype = {
...
     var dateRange = dateRangeFormatter.format( fromDateValue, toDateValue );
-    var error_messages = [];
+    var errorMessages = [];
 
     if ( dateRange && dateRange.isValid ) {
       this.$gte.val( dateRange.startDate );
       this.$lte.val( dateRange.endDate );
     } else {
-      error_messages.push( DATE_ERRORS.invalid );
+      errorMessages.push( DATE_ERRORS.INVALID );
```
KimberlyMunoz: Wouldn't this be `errorMessages.push( DATE.INVALID );`?

cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
```diff
@@ -38,6 +40,16 @@
 {% endmacro %}
 
 {% macro _render_filter_fields(controls, form, index) -%}
+     <div class="form-l_col
```
KimberlyMunoz: Is this extra code just for testing?

sebworks: yeah, I have should have added a TODO for that.

KimberlyMunoz: But it'll get removed before merging?

sebworks: yep, just to make it easier to test.

cfgov/unprocessed/js/organisms/FilterableListControls.js
anselmbradford: Mm, right. I like `continue` too and we could relax the rules. I set them strict so that we would see where we need to relax them over time, since if they'd started loose it would be harder to see where they should be more strict. I see why `continue` could be a problem though and how working around it creates less nested logic and more testable code (though the methods are private so it may not matter). In this case you could consider something like this that would separate the steps that are happening:

```
var field;
var type;
var isGroupField;
for ( var i = 0, len = fields.length; i < len; i++ ) {
  field = fields[i];
  type = field.getAttribute( 'data-type' ) ||
             field.getAttribute( 'type' ) ||
             field.tagName.toLowerCase();
  isGroupField = _defaults.groupFieldTypes.indexOf( type ) > -1;

  if ( shouldValidateField( field, type, isGroupField ) === true ) {
    _setFieldStatus( _validateField( field, type, isGroupField ) )
  }
}

// … further down …

function _setFieldStatus( validatedField ) {
  var fieldIsValid = false;
  for ( var prop in validatedField.status ) {
    if ( validatedField.status[prop] === false ) fieldIsValid = true;
  }

  if ( fieldIsValid ) {
    validatedFields.invalid.push( validatedField );
  } else {
    validatedFields.valid.push( validatedField );
  }
}
```

KimberlyMunoz:
:+1: Just one comment above, but it looks like it's working well. Some of the error messages seem a little ambiguous on the "From" and "To" labels of the date picker, but I think UX will have to look at that after we merge.

cfgov/unprocessed/js/organisms/FilterableListControls.js
sebworks: After doing further reading, I'm going to leave it as it exist. For smaller arrays, where you know the amount shouldn't be more than 20/30, I'm actually going to start using `forEach`. I think it's easier to read/understand and that is more valuable then any small performance benefits you would get with a `for`. These are the articles that I read:

http://codeblog.jonskeet.uk/2009/01/29/for-vs-foreach-on-arrays-and-lists/

https://ryanpcmcquen.org/javascript/2015/10/25/map-vs-foreach-vs-for.html

https://www.symphonious.net/2010/10/09/javascript-performance-for-vs-foreach/

cfgov/unprocessed/js/modules/post-filter.js
sebworks: fixed in https://github.com/cfpb/cfgov-refresh/commit/8a828a9fea878b0aaa9975c3a914611071f25e88

cfgov/unprocessed/js/organisms/FilterableListControls.js
anselmbradford: Yeah, sorry to imply it's a performance tweak—this is more about disentangling the steps vs performance. If functions aren't nested it means they can be isolated. In `foreach` one function is nested inside the loop so there is a tight coupling between the steps (loop, do something on each item), whereas in a classic loop, it's a loose coupling. There's one step that calls another. One function is for looping over the array, and another is for setting the value. For testing, having the two functions separate means they can be more easily tested in isolation, easier to document (not that we're utilizing jsdocs yet), and easier to refactor in isolation (e.g. made public, shared between functions, etc.). But again as private methods it may not matter.

sebworks: yeah, I don't think it matters in this because they are private functions and aren't being shared across files.

anselmbradford: :+1:

jimmynotjim:
:+1: Really nice work on this one. @anselmbradford might still want to sign off since he had a lot of comments.

cfgov/unprocessed/js/config/error-messages-config.js
```diff
@@ -7,13 +7,21 @@
 'use strict';
 
 var ERROR_MESSAGES = {
-  CHECKBOX_ERRORS: {
-    required: 'Please select at least one of the "%s" options.'
+  CHECKBOX : {
+    required : 'Please select at least %s of the options.'
```
anselmbradford: Should this be uppercase like the rest?

cfgov/unprocessed/js/modules/util/dom-traverse.js
```diff
@@ -1,6 +1,5 @@
 'use strict';
 
-// Required polyfills for IE9.
 if ( window.Modernizr && !window.Modernizr.classlist ) { require( '../polyfill/class-list' ); } // eslint-disable-line no-undef, global-require, no-inline-comments, max-len
```
anselmbradford: This can be removed if your IE9 fix goes in first.

cfgov/unprocessed/js/organisms/FilterableListControls.js
```diff
@@ -18,157 +18,261 @@ var Notification = require( '../molecules/Notification' );
...
     ]
   };
 
+  var _validators = {
+    date: function( field, currentStatus ) {
+      var status = currentStatus || {};
+      var dateRegex =
+        /^\d{2}$|^\d{4}$|^\d{2}\/(?:\d{4}|\d{2})$|^\d{2}\/\d{2}\/\d{4}$/;
+      if ( field.value && dateRegex.test( field.value ) === false ) {
+        status.msg += ERROR_MESSAGES.DATE.INVALID;
```
anselmbradford: Okay to have `+=` to a string without any space, comma, etc. on the end to separate the concat'd strings?

cfgov/unprocessed/js/config/error-messages-config.js
sebworks: I left this one as lowercase to signify that it needed to be modified (`%s`).

cfgov/unprocessed/js/organisms/FilterableListControls.js
sebworks: yeah, the messages are getting concatenated  in https://github.com/cfpb/cfgov-refresh/pull/1451/files#diff-373a9942243971a4ddce967a7dd02b20R137

cfgov/unprocessed/js/config/error-messages-config.js
anselmbradford: Hm, and it can be modified even though ERROR_MESSAGES is exported with `Object.seal( )`? Interesting, maybe object seal only works on the direct object it wraps. Technically a constant wouldn't be modified at any level of its structure. Not a big deal now, but this may require a change in futureland where `var ERROR_MESSAGES` becomes `const ERROR_MESSAGES`.

anselmbradford: Actually I see where this is used:

`ERROR_MESSAGES.CHECKBOX.required.replace( '%s', required );`

`replace` returns a new string, so `ERROR_MESSAGES.CHECKBOX.required` is constant because it's not modified directly. I'd say make it uppercase as well.

cfgov/unprocessed/js/organisms/FilterableListControls.js
```diff
@@ -18,157 +18,261 @@ var Notification = require( '../molecules/Notification' );
...
+        status.date = false;
+      }
+      return status;
+    },
+    email: function( field, currentStatus ) {
+      var status = currentStatus || {};
+      var regex = [ "^[a-z0-9\u007F-\uffff!#$%&'*+\/=?^_`{|}~-]+(?:\.[a-z0-9",
+                    "\u007F-\uffff!#$%&'*+\/=?^_`{|}~-]+)*@(?:[a-z0-9]",
+                    "(?:[a-z0-9-]*[a-z0-9])?\.)+[a-z]{2,}$" ].join( '' );
```
anselmbradford: Linter doesn't like this double quote, can it be single?

sebworks: I tried it but it wouldn't work because the Regex contains a single quote.  Even when I escaped the single quote in the Regex.

anselmbradford:
:+1: minor comment above with the constant. Wagtail admin is down for me so I can't test there, hope someone else has!
Note on the linter: this is introducing a bunch of `key-spacing` warnings for the space before the colon in, e.g. `DATE : {`. Those could be changed, or the linter rule could be updated to allow for it. I tend to omit the space, but either way is fine with me. Doesn't necessarily have to happen now as the linter will remind about it.

cfgov/unprocessed/js/organisms/FilterableListControls.js
anselmbradford: Ah cool, just needs an eslint ignore for quotes then.

PR #1452, opened 2016-02-12
use clang when NPM needs to compile c++

rosskarchner:
Our travis build had been depending on g++48 from the [ubuntu-toolchain-r](https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test?field.series_filter=precise) PPA, which no longer seems to carry it.

This PR switches us to the built-in clang++ compiler now, which appears to work fine.

@anselmbradford

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ install:
   - pip install virtualenv virtualenvwrapper
   - pip install autoenv
   # Set g++ version for npm nan.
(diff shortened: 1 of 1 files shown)


rosskarchner:
as a bonus, we're probably using clang on our macs anyways.

anselmbradford:
BONUS :100:

PR #1453, opened 2016-02-12
Cleaning up the home page work

jimmynotjim:
Cleaning up the home page work

## Removals

- Removed old beta styles
- Removed extra markup and empty classes

## Changes

- Abstracted info unit into a helper mixin to make it easier to re-use the inline version
- Moved Home page specific layout changes to it's own file
- Fixed spacing

## Testing

- `gulp build` and checkout the home page

## Review

- @KimberlyMunoz 
- @anselmbradford 
- @sebworks 

## Todos

- I've already started updating the half-width link blob, 50/50 and 25/75 in a separate branch to keep this scope on point.

## Checklist


cfgov/jinja2/v1/index.html
```diff
@@ -403,7 +223,7 @@ <h3 class="h1 m-card_heading">98% of consumers</h3>
                 <div class="content-l">
                     <section class="content-l_col
                                     content-l_col-1-3">
-                        <div class="">
+                        <div>
```
KimberlyMunoz: What are these `div`s necessary for?

cfgov/unprocessed/css/pages/home.less
```diff
@@ -0,0 +1,105 @@
...
+    } );
+
+    .respond-to-min( @bp-lg-min, {
+        .grid_column( 4 );
+    } );
+}
+
+.o-well__home {
+    .respond-to-min( @bp-xl-min, {
+        max-width: 1230px - ( 45px * 2 );
```
KimberlyMunoz: Maybe add a comment explaining this for future devs?

KimberlyMunoz:
This looks really solid and much cleaner than before. A few questions, but :+1:.

cfgov/jinja2/v1/index.html
jimmynotjim: Yes and no, these should eventually become related posts, but those had some layout styles that need to be pulled out before they can be used here.

sebworks:
Should the border be inline withe hero?

screen shot 2016-02-16 at 6 46 06 pm

sebworks:
Can we use flexbox to stretch the shorter boxes?

cfgov/unprocessed/css/pages/home.less
```diff
@@ -0,0 +1,106 @@
...
+        // Max-width - both gutters: required due to well spaning the full page
+        max-width: 1230px - ( 45px * 2 );
+        margin: 0 auto;
+        padding-right: unit( 45px / @base-font-size-px, em );
+        padding-left: unit( 45px / @base-font-size-px, em );
+    } );
+}
+
+.o-well__home-hero {
+    background-repeat: no-repeat;;
```
sebworks: extra semicolon

cfgov/unprocessed/css/pages/home.less
```diff
@@ -0,0 +1,106 @@
...
+    } );
+
+    .respond-to-min( @bp-sm-min, {
+        min-height: 380px - (45px * 2);
+        padding-top: unit( 45px / @base-font-size-px, em );
+        padding-bottom: unit( 45px / @base-font-size-px, em );
+
+        background-color: @gray;
+        background-image: url('../img/home-hero_full.jpg');
+        background-size: 1230px;
```
sebworks: might consider making `1230px` a variable

sebworks:
Looks great. :+1:

cfgov/unprocessed/css/molecules/card.less
```diff
@@ -1,30 +1,30 @@
...
 
-  p {
-    margin-bottom: 0;
-  }
+    p {
+        margin-bottom: 0;
+    }
 }
 
 .o-card {
```
anselmbradford: Mentioned in chat.. organism styles shouldn't be hidden in the molecule styles.

cfgov/unprocessed/css/molecules/info-unit.less
```diff
@@ -71,91 +129,39 @@
...
+    Inline info-unit Helper Mixin -
+    This mixin is used to modify an info-unit to an inline arrangement.
+    Most of the time the `.m-info-unit__inline` class will be sufficient,
+    but there are limited situations the inline arrangement needs to be conditioned.
+    See /pages/home.less L:103
+ */
+.u-info-unit-base__inline( @img_size ) {
+    .u-clearfix();
+
+    .m-info-unit_image {
```
anselmbradford: Does this need to go in the mixin since it doesn't depend on `@img_size`?

cfgov/unprocessed/css/pages/beta.less
```diff
@@ -23,64 +23,64 @@
...
-        unit( 14px / @btn-font-size, em );
-    margin-top: @grid_gutter-width / 2;
+    &_btn {
+        .btn__secondary();
+        width: 100%;
+        // Restore standard btn padding, which is removed by .expandable_target.
+        padding:
+            unit( @btn-v-padding / @btn-font-size, em )
+            unit( 14px / @btn-font-size, em );
+        margin-top: @grid_gutter-width / 2;
```
anselmbradford: Does this need unit conversion?

cfgov/unprocessed/css/pages/beta.less
```diff
@@ -23,64 +23,64 @@
...
-        top: 1px;
-        right: 0;
-        .u-link__colors(@black, @dark-gray);
-        .u-link__no-border();
-        text-transform: lowercase;
-    });
+        .respond-to-min(@bp-sm-min, {
+            .btn__link();
+            width: auto;
+            padding: @grid_gutter-width / 2;
```
anselmbradford: Does this need unit conversion?

cfgov/unprocessed/css/pages/home.less
```diff
@@ -0,0 +1,106 @@
...
+
+    .respond-to-max( @bp-xs-max, {
+        padding-bottom: unit( ( @grid_gutter-width * 2 ) / @base-font-size-px, em );
+    } );
+}
+
+.content-l_col__home {
+    + .content-l_col__home {
+        .respond-to-max( @bp-med-max, {
+            margin-top: unit( ( @grid_gutter-width ) / @base-font-size-px, em );
```
anselmbradford: Trivial: Extra unneeded parentheses.

cfgov/unprocessed/css/pages/home.less
```diff
@@ -0,0 +1,106 @@
...
+        } );
+
+        .respond-to-range( @bp-sm-min, @bp-med-max, {
+            &:before {
+                .a-rule-break();
+                content: '';
```
anselmbradford: I'm not sure why, but I've seen `'\a0'` used to fill `content` with empty content. There may be a cross-browser reason. Not sure. Perhaps worth looking into at some point.

cfgov/unprocessed/css/pages/home.less
```diff
@@ -0,0 +1,106 @@
...
+                display: block;
+            }
+        } );
+    }
+
+    .respond-to-min( @bp-sm-min, {
+        .grid_column( 1, 1 );
```
anselmbradford: Trivial: Pretty sure it could be `.grid_column( 1 )` like L:29.

cfgov/unprocessed/css/pages/home.less
```diff
@@ -0,0 +1,106 @@
...
+    } );
+
+    .respond-to-min( @bp-lg-min, {
+        .grid_column( 4 );
+    } );
+}
+
+.o-well__home {
+    .respond-to-min( @bp-xl-min, {
+        // Max-width - both gutters: required due to well spaning the full page
```
anselmbradford: Trivial: typo `spanning`

cfgov/unprocessed/css/pages/home.less
```diff
@@ -0,0 +1,106 @@
...
+        max-width: 1230px - ( 45px * 2 );
+        margin: 0 auto;
+        padding-right: unit( 45px / @base-font-size-px, em );
+        padding-left: unit( 45px / @base-font-size-px, em );
```
anselmbradford: 45px should probably utilize our standard gutter width variable (1 gutter width + 1/2 gutter width).

cfgov/unprocessed/css/pages/home.less
```diff
@@ -0,0 +1,106 @@
...
+    } );
+}
+
+.o-well__home-hero {
+    background-repeat: no-repeat;;
+    border: none;
+
+    .respond-to-max( @bp-xs-max, {
+        padding-bottom: 56.25%;
```
anselmbradford: Perhaps add a code comment on why this magic number is needed.

cfgov/unprocessed/css/molecules/info-unit.less
jimmynotjim: Yes, because it's always used for the inline modifier, no matter what the image size is.

cfgov/unprocessed/css/pages/home.less
jimmynotjim: Responded in the original PR. The mixin assumes the second param is `12` if you only pass one, so that'd make it `1/12` instead of 12/12 like we want. I could just update it to 12, but I was worried if that ever changed what might happen.

jimmynotjim:
> Can we use flexbox to stretch the shorter boxes?

I'd love to use flexbox, but it doesn't seem to be a priority and getting cross-browser working would require more work than we have time for now. I'll add a TODO though so we don't lose it.

jimmynotjim:
> Should the border be inline withe hero?

We'd need to set the wells to border-box, but I was a bit worried what this might do to existing wells. Probably something to consider in a base well style?

sebworks:
I wouldn't add it to the base well style; I would just implement it for the home page.

cfgov/unprocessed/css/pages/home.less
anselmbradford: ah, right, disregard

jimmynotjim:
> I would just implement it for the home page.

That leads to even more well adjustments specific to the homepage (because the padding/width get's fudged). I've been trying to limit one-off tweaks as much as possible to avoid creating new issues down the line.

sebworks:
> That leads to even more well adjustments specific to the homepage (because the padding/width get's fudged). I've been trying to limit one-off tweaks as much as possible to avoid creating new issues down the line.

ok, I would leave it as-is for now. :100:

cfgov/unprocessed/css/pages/beta.less
jimmynotjim: I just deleted the old home page stuff, this will all go away when the banner does, so I'm going to leave it for now.

cfgov/unprocessed/css/pages/home.less
jimmynotjim: Sounds good to me

cfgov/unprocessed/css/pages/home.less
jimmynotjim: I'm going to leave it because we still haven't made these global variables, but I'm adding a TODO for after launch.

jimmynotjim:
This is ready for another review.

cfgov/unprocessed/css/pages/home.less
```diff
@@ -0,0 +1,93 @@
...
+        max-width: 1230px - ( 45px * 2 );
+        margin: 0 auto;
+        padding-right: unit( 45px / @base-font-size-px, em );
+        padding-left: unit( 45px / @base-font-size-px, em );
+    } );
+}
+
+@home-hero_sm: 100%;
+@home-hero_med: 900px;
+@home-hero_lg: 1230px;
```
anselmbradford: Possibly include the units in the variable names since they are mixed units.

cfgov/unprocessed/css/pages/home.less
```diff
@@ -0,0 +1,93 @@
...
+    .respond-to-max( @bp-xs-max, {
+        padding-bottom: 56.25%;
+
+        background-image: url('../img/home-hero_sm.jpg');
+        background-position: center bottom;
+        background-size: @home-hero_sm;
+    } );
+
+    .respond-to-min( @bp-sm-min, {
+        min-height: 380px - ( 45px * 2 );
```
anselmbradford: Add code comment of where `380px` comes from.

anselmbradford:
Two minor comments, otherwise :+1:

cfgov/unprocessed/css/pages/home.less
jimmynotjim: I was purposely leaving them out so they're more generalized for exactly that reason. If I update the small one, I just want to change the value, and not have to update the var name as well.

cfgov/unprocessed/css/pages/home.less
jimmynotjim: That's in the spec

cfgov/unprocessed/css/pages/home.less
```diff
@@ -0,0 +1,94 @@
...
+
+.m-home-hero {
+    .respond-to-min( @bp-sm-min, {
+        .grid_column( 1, 1 );
+    } );
+
+    // Adding a custom breakpoint at 776px to fix specific
+    // layout issues
+    .respond-to-range( 776px, @bp-sm-max, {
+        .grid_column(8, 12);
```
sebworks: Should the grid columns be spaced `( 8, 12 )` like on line https://github.com/cfpb/cfgov-refresh/pull/1453/files#diff-3130a3912e7f8f9ac6af49da87a12794R58?

cfgov/unprocessed/css/molecules/card.less
```diff
@@ -1,30 +1,15 @@
...
-  } );
-
-  .respond-to-range( @bp-sm-min, @bp-med-max, {
-    .grid_column( 1, 1 );
-  } );
 
-  .respond-to-min( @bp-lg-min, {
-    .grid_column(6);
-  } );
+    p {
```
sebworks: Should we add a `m-card_body` class to the p tag and target that instead?

sebworks:
:+1: , looks great.

PR #1457, opened 2016-02-12
Local protractor install fix

anselmbradford:
Checks for protractor global install before npm installs a local protractor.

## Testing

- `./frontend.sh`

## Review

- @jimmynotjim


jimmynotjim:
Ran and works :+1:

PR #1458, opened 2016-02-12
Fix bug where publised pages were showing shared content

kurtw:
- Also solved memory leak of double revisions created for a single Page save
- Added comments for clarity

Explained in code comments.

If a page was published then it's content is updated and shared but not published, the published page will show the updated content.

## Test
- BEFORE pulling down, try publishing a page, then editing and sharing the updated page. You will see the update on both the live version and shared version of the page.
- Pull down the code. Repeat publishing, editing, and sharing. You should see the desired behavior.

@kave 
@richaagarwal 
@rosskarchner

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 
 
 ### Fixed
+- Fix bug where publised pages were showing shared content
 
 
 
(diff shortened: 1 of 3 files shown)


kave:
@kurtw hey so i encountered a bug. 
Scenario
- Create and publish a page
- Add a change and save it.
- It makes the page unavailable on the live site

kurtw:
Bug should be fixed.

kave:
cool looks good :+1:

PR #1460, opened 2016-02-13
Removed all forms from project

KimberlyMunoz:
Since we aren't including forms our v1 launch, they are being removed from the project. Content may need to rework the content, I think it should happen after this PR so they can easily see what they have to work with once we remove the content on the flapjack branch.

## Review
- @jimmynotjim 
- @anselmbradford 
- @sebworks 
- @richaagarwal


richaagarwal:
This looks good to me. At least, it resolves some of my failing tests :)

jimmynotjim:
I hadn't realized we were removing them all. :+1: on cutting scope!

KimberlyMunoz:
One sec, I just realized there's form tests that need to be removed too!

KimberlyMunoz:
Removed!

anselmbradford:
Yeah, I didn't know we were removing them either :+1:

kave:
:+1:

PR #1462, opened 2016-02-16
Remove atoms from the sidebar

kurtw:
Look in the sidefoot streamfield on any page to make sure that these are not present.

Go to each page type to make sure that none of them are broken.

This is what it should look like

@richaagarwal 
@kave


richaagarwal:
@kurtw could you upload a screenshot of what should not be present?

richaagarwal:
:+1:

kave:
:+1:

Issue #1336, opened 2016-01-13
formatting in text introduction


tnferrell:
i'm not sure if this is a duplicative issue, i couldn't find it when i was searching the repo, but the text introduction formatting looks off. example can be seen here: /consumer-help/sending-money-abroad/

jimmynotjim:
@tnferrell looking at the inspector, it seems like either the cms is adding `p` tags or they've been added in the input field in the admin. Because of the classname requirement for the lead paragraph, we have to add the `p` in the template here. Could you take a look at the admin and make sure there aren't any there?

tnferrell:
Unfortunately, there isn't an HTML editor in wagtail, so I can't verify if I've accidentally added code. But, I copied the text from notepad, pasted it in, and the P is selected, which I'm assuming means it's in a paragraph tag. I can't unselect the P. Any time you paste text into the rich text box it's selected by default. Hope this helps.

duelj:
Just wanted to note that this issue relates to issue #1366. Many of the other scenarios of where this type of thing is happening are documented there.

KimberlyMunoz:
This field has to be a rich text field because we might have links or bullet points in the lead paragraph so it's an all or nothing thing for potentially stripping extra `p`s. I went in and I couldn't really remove them when once they're in. @jimmynotjim, why couldn't we wrap this in a `div` instead of a `p`?

jimmynotjim:
We've done that before, but I'm concerned how that affects spacing https://github.com/cfpb/capital-framework/blob/canary/src/cf-core/src/cf-base.less#L333-L344

tnferrell:
Just wanted to add to this that we're seeing the font is fine now. But I think it's supposed to be a larger size? /policy-compliance/enforcement/warning-letters/

jimmynotjim:
Yeah, the main issue is still open, it just turned into Avenir when I fixed the base-font. I say we go w/ a div wrapper. @KimberlyMunoz if you haven't, I can do this today.

KimberlyMunoz:
Hey @jimmynotjim, let me know if this issue is open?

jimmynotjim:
Discussed in scrum, lets go w/ the div wrapper because if a `return` gets pasted in or entered, it messes everything up.

Issue #1466, opened 2016-02-17
Font size changes in Item Introduction / doc detail page


Schlachtmeyer:
For the Reports pages I'm using the doc detail template with the Item Introduction for each report. The font size of the paragraph text seems to depend on whether or not there's a hard return? It starts out as a large font size and then I can force it to a smaller font size by inserting a hard return - and then if I delete the hard return it stays at the smaller size. To me the smaller size looks better, but there should be a better way to make that happen?

Large font size:
![image](https://cloud.githubusercontent.com/assets/14129916/13113041/67b2f5e6-d55b-11e5-92ea-f4327470983c.png)

Small font size:
![image](https://cloud.githubusercontent.com/assets/14129916/13113051/711c2576-d55b-11e5-808d-b8576486c2e9.png)

KimberlyMunoz:
The spec calls for the larger font size, right? @ajbush and @duelj?

This might be the same issue as #1336.

duelj:
@KimberlyMunoz - yes the [spec](https://github.cfpb.gov/flapjack/Modules-V1/wiki/Text-Introduction) calls for the larger (H3) subheading to to the page title. It looks like it is basically the same as the #1336 issue. One of the other things build does not often make use of was the rule line at the bottom of the text/page introduction to define that section from the rest of the page content. I don't know if this is just not getting add in Wagtail, or if it is not part of the organism. I think we will want to make sure it is an option in the CMS, so that we can use it in the future.

Schlachtmeyer:
@duelj For each report I have the choice of a Text Intro or an Item Intro. The Text Intro has no date so I've been using the Item Intro. The Text Intro has the option of a line below, but the Item Intro doesn't. (And I can't get Wagtail to save a page with a Text Intro, so I can't show you!)

KimberlyMunoz:
We discussed in Team Working session and the decision was to put any content that needs to be smaller like that in the Full-Width molecule. We're going to make sure that returns don't break the h3 intended size for that introduction paragraph in #1471.

KimberlyMunoz:
Fix has been pushed up to refresh!

PR #1471, opened 2016-02-18
Changing p tags in introductions to divs

KimberlyMunoz:
In item and text introductions, any hard enter will break the formatting
on rendered pages. This changes the p tags into divs to fix this.

Also resolves issue #1336 and #1466.

## Testing
- Create a page with a text introduction and item introduction. The document detail page is the best place to do both at once.

## Screenshot

image


sebworks:
:+1:

jimmynotjim:
:+1:

PR #1473, opened 2016-02-19
Updates ESLint to 2.0

anselmbradford:
## Changes

- Updates ESLint to 2.0 and fixes linting errors and some warnings.

## Testing

- `./setup.sh && gulp lint` should show 0 linting errors (it will show warnings tho).

## Review

- @jimmynotjim 
- @KimberlyMunoz 
- @sebworks

### DIFF
--- a/test/unit_tests/modules/classes/BreakpointHandler-spec.js
+++ b/test/unit_tests/modules/classes/BreakpointHandler-spec.js
@@ -6,12 +6,12 @@ var sinon = require( 'sinon' );
 var BreakpointHandler;
 var args;
 var BASE_LOC = '../../../../cfgov/unprocessed/js/';
-
+var standardType = require( BASE_LOC + 'modules/util/standard-type' );
 
 beforeEach( function() {
   args = {
-    enter:      function() {},
-    leave:      function() {},
+    enter:      standardType.noopFunct,
+    leave:      standardType.noopFunct,
     breakpoint: 600
   };
 
--- a/.eslintrc
+++ b/.eslintrc
@@ -31,6 +31,8 @@ env:
   browser: true
   node: true
   commonjs: false
+  shared-node-browser: true
+  es6: true
   worker: false
   amd: true
   mocha: false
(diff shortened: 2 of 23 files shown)


test/unit_tests/modules/classes/BreakpointHandler-spec.js
```diff
@@ -10,8 +10,8 @@ var BASE_LOC = '../../../../cfgov/unprocessed/js/';
 
 beforeEach( function() {
   args = {
-    enter:      function() {},
-    leave:      function() {},
+    enter:      function() { return; },
```
sebworks: Why is the return statement necessary? Are we requiring that all functions ( not constructors ) have a return statement?

anselmbradford: No, it's http://eslint.org/docs/rules/no-empty-function 
Basically it can be unclear whether empty functions are intentional or mistakenly left in the codebase so the rule is about leaving a code comment in the function. In this case, the functions are placeholders that are intended to be overridden. I could include a comment, but then would need to unwrap the `{ }` across multiple lines. It seemed clear that these are placeholders, so I included the `return;` to give them a body (which is implicit anyway, so it shouldn't change behavior).

sebworks:
Instead of adding a return statement, how about just creating a `noop` function that we could use throughout the codebase?

KimberlyMunoz:
:+1:. 

I'll go with whatever you and Steven come up with on the no-empty function rule, but linter works well.

test/unit_tests/modules/classes/BreakpointHandler-spec.js
anselmbradford: Added noop in https://github.com/cfpb/cfgov-refresh/commit/293473e2d3156a219db95bd473a4e629221887b4

anselmbradford:
noop added in https://github.com/cfpb/cfgov-refresh/commit/293473e2d3156a219db95bd473a4e629221887b4 
I'll also use this file to add a constant for the `data-js-hook`

sebworks:
:+1:

PR #1475, opened 2016-02-19
Fix main contact info breaking pages

kurtw:
Jinja2 imports were unnecessary and breaking the page that the organism main contact info exists on.

## Removals

- jinja2 imports that were unnecessary and breaking the page

## Testing

- Go to a page with that organism.

## Review

- @richaagarwal 
- @kave


richaagarwal:
weird... I thought I ran into this and removed them earlier.  but :+1:

PR #1477, opened 2016-02-19
Fix WP events processor to import events data

kurtw:
Some events in WP don't have fields that are required in Wagtail. Rather than make then non-required, I've refactored the code so it doesn't fail the imports, but will still not convert the page over.

## Changes

- Changed eventpage.py processor to check to see if start and end dates exist before assignment.

## Testing

- `./cfgov/manage.py import-data events eventpage --parent events -u user -p pass`

## Review

- @richaagarwal 
- @kave 

## Notes

- There are currently 12 events that don't have the necessary fields for conversion.

## Todos

- Add import-data to our build flow and send email notifications on errors


kave:
:+1:

PR #1486, opened 2016-02-23
Accounts for missing start/end dates in Events

kave:
Events page without a start/end date break the template

## Testing

- View an imported event page without a start/end date

## Review

- @kurtw 
- @anselmbradford


kurtw:
:+1:

anselmbradford:
:+1:

PR #1489, opened 2016-02-23
removed unnecessary imports from sidebar contact

kave:
## Removals

- Removed unnecessary imports for sidebar contact organism

## Testing

- View a page with a sidebar contact that and it does not break

## Review

- @KimberlyMunoz 
- @kurtw


kurtw:
:+1:

PR #1491, opened 2016-02-23
Add new manager to deal with shared vs live appropriately

kurtw:
Makes sure the right pages are served by adding a query method to get a set of pages according to their status and hostname.

## Additions

- New manager

## Testing
These should work. Look at the comments and change the command as needed.

```
./cfgov/manage shell
from v1.models import CFGOVPage, AbstractFilterPage
CFGOVPage.objects.shared() # should return shared pages

# this will only really work if you have the two sites set up (content.localhost, localhost)
CFGOVPage.objects.live_shared('content.localhost') #should return shared or live pages 

# the manager should be accessible via the AbstractFilterPage as well
pages = AbstractFilterPage.objects.live_shared('content.localhost')
```

## Review

- @kave 
- @richaagarwal 
- @rosskarchner

### DIFF
--- a/cfgov/v1/forms.py
+++ b/cfgov/v1/forms.py
@@ -199,21 +199,19 @@ class FilterableListForm(forms.Form):
 
     def __init__(self, *args, **kwargs):
         parent = kwargs.pop('parent')
+        hostname = kwargs.pop('hostname')
         super(FilterableListForm, self).__init__(*args, **kwargs)
-        self.set_topics(parent)
-        self.set_authors(parent)
+        self.set_topics(parent, hostname)
+        self.set_authors(parent, hostname)
 
     # Populate Topics' choices
-    def set_topics(self, parent):
-        live_tags = [tag for tags in [page.tags.names() for page in
-                    AbstractFilterPage.objects.live().descendant_of(
-                    parent)] for tag in tags]
-        shared_tags = [tag for tags in [page.tags.names() for page in
-                       AbstractFilterPage.objects.descendant_of(parent)
-                       if page.shared] for tag in tags]
-        all_tags = list(chain(live_tags, shared_tags))
+    def set_topics(self, parent, hostname):
+        tags = [tag for tags in
+                     [page.tags.names() for page in
+                      AbstractFilterPage.objects.live_shared(hostname).descendant_of(parent)]
+                     for tag in tags]
         # Orders by most to least common tags
-        options = most_common(all_tags)
+        options = most_common(tags)
         most = [(option, option) for option in options[:3]]
         other = [(option, option) for option in options[3:]]
         self.fields['topics'].choices = \
@@ -221,10 +219,10 @@ class FilterableListForm(forms.Form):
              ('All other topics', tuple(other)))
 
     # Populate Authors' choices
-    def set_authors(self, parent):
+    def set_authors(self, parent, hostname):
         all_authors = [author for authors in [page.authors.names() for page in
-                       AbstractFilterPage.objects.live().descendant_of(
-                       parent).live()] for author in authors]
+                       AbstractFilterPage.objects.live_shared().descendant_of(
+                       parent)] for author in authors]
         # Orders by most to least common authors
         self.fields['authors'].choices = [(author, author) for author in
                                           most_common(all_authors)]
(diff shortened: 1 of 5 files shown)


kave:
@kurtw can you resolve the merge conflicts?

kurtw:
Ready

cfgov/v1/forms.py
```diff
@@ -198,21 +198,19 @@ class FilterableListForm(forms.Form):
...
-    def set_topics(self, parent):
-        live_tags = [tag for tags in [page.tags.names() for page in
-                    AbstractFilterPage.objects.live().descendant_of(
-                    parent)] for tag in tags]
-        shared_tags = [tag for tags in [page.tags.names() for page in
-                       AbstractFilterPage.objects.descendant_of(parent)
-                       if page.shared] for tag in tags]
-        all_tags = list(chain(live_tags, shared_tags))
+    def set_topics(self, parent, hostname):
+        live_tags = [tag for tags in
```
kave: lets rename `live_tags` ==> `tags` since now they could be a combination of live/shared

cfgov/v1/forms.py
```diff
@@ -220,7 +218,7 @@ def set_topics(self, parent):
              ('All other topics', tuple(other)))
 
     # Populate Authors' choices
-    def set_authors(self, parent):
+    def set_authors(self, parent, hostname):
         all_authors = [author for authors in [page.authors.names() for page in
                        AbstractFilterPage.objects.live().descendant_of(
```
kave: missing `AbstractFilterPage.objects.live_shared(hostname)` ?

cfgov/v1/models/learn_page.py
```diff
@@ -75,6 +75,8 @@ class AbstractFilterPage(CFGOVPage):
     # This page class cannot be created.
     is_creatable = False
 
+    objects = CFGOVPage.objects
```
kave: will it not already inherit this from `CFGOVPage` ?

cfgov/v1/forms.py
kurtw: ah yes, thanks for catching

cfgov/v1/models/learn_page.py
kurtw: No :(

kurtw:
Made comment fixes

cfgov/v1/models/learn_page.py
kave: whack

richaagarwal:
You can also do `./cfgov/manage.py shell_plus` and then you don't have to do the imports

richaagarwal:
Also, I think it needs to be `cfgovpage = BrowseFilterablePage.objects.first()` to grab only one.

richaagarwal:
I get this error when I try to access cfgovpage.objects: ```>>> cfgovpage
<BrowseFilterablePage: hi>
>>> cfgovpage.objects
  File "<console>", line 1, in <module>
  File "/Users/agarwalr/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/django/db/models/manager.py", line 255, in __get__
    raise AttributeError("Manager isn't accessible via %s instances" % type.__name__)
AttributeError: Manager isn't accessible via BrowseFilterablePage instances
>>> ```

kurtw:
@richaagarwal thanks for your comments. I didn't know about `shell_plus`! I put in get() instead because the first BrowseFilterable you get might not be the best one to test. Also, the manager only exists on the CFGOVPage and the AbstractFilterPage.

kurtw:
Also, the class holds the manager not the instance so I was wrong to put that as a testing step to begin with.

kurtw:
Updated the testing directions to not be overtly incorrect.

richaagarwal:
What is `cfgovpage` supposed to be here?: `cfgovpage.objects.live_shared('content.localhost') #should return shared or live pages`

richaagarwal:
Oh, is it supposed to be `CFGOVPage` ?

kave:
ooo `shell_plus` thats handy

kave:
@richaagarwal yea he meant `CFGOVPage`

kave:
:+1:

Issue #692, opened 2015-06-30
Investigate whether z-index:1000 is needed in Slick carousel


anselmbradford:
See https://github.com/cfpb/cfgov-refresh/pull/686#discussion_r33521467 Slick carousel nav buttons have a very high z-index:

```less
        .slick-prev,
        .slick-next {
            height: 30px;
            width: 30px;
            top: 75px;
            z-index: 1000;
```

anselmbradford:
Updated to `z-index: 9` in https://github.com/cfpb/cfgov-refresh/pull/1495

PR #1495, opened 2016-02-24
Added overlay atom

anselmbradford:
Adds atom for covering the page content with a semi-opaque overlay that's shown when the mobile mega menu is shown.

## Additions

- Added overlay atom

## Changes

- Added `rootExpandBegin` and `rootCollapseEnd` events to the mega menu to respond when the base level menu is opened/closed so that events can be tied to when the menu first opens and completely closes.

## Testing

- `gulp build`
- Resize page to mobile size.
- Click hamburger menu, observe overlay appear and menu slide in from left.
- Click hamburger menu, observe menu slide to left and overlay disappear.

## Review

- @KimberlyMunoz 
- @sebworks 
- @jimmynotjim 
- @duelj 
- @ajbush 

## Screenshots

overlay

## Notes

- @duelj @ajbush Since the overlay is a new atom it could be added to the wiki and the mega menu. Note that the mega menu doesn't include the overlay. The overlay reference is handed to the Header organism at the template level, and the Header shows/hides it when the mega menu expands/collapses.

## Todos

- Please ignore no-js and keyboard state for the mega menu (still WIP), but note any no-js or keyboard state issues for the overlay atom directly.

### DIFF
--- a/cfgov/unprocessed/css/atoms/overlay.less
+++ b/cfgov/unprocessed/css/atoms/overlay.less
@@ -1,0 +1,36 @@
+/* topdoc
+  name: Overlay
+  family: cfgov-atoms
+  patterns:
+    - name: Default example
+      markup: |
+          <div class="a-overlay"></div>
+      codenotes:
+        - |
+          Structural cheat sheet:
+          -----------------------
+          .a-overlay
+  tags:
+    - cfgov-atoms
+*/
+
+.a-overlay {
+    // Only show overlay at mobile/tablet size.
+    .respond-to-max( @bp-sm-max, {
+        height: 100%;
+        width: 100%;
+
+        position: fixed;
+        // TODO: Update z-index value with standard z-index variable.
+        z-index: 10;
+
+        background: @black;
+        opacity: 0.6;
+    } );
+}
+
+
+/* topdoc
+    name: EOF
+    eof: true
+*/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Added new content flush sides on small modifier to fix an issue where margin was set on the molecule level instead of the template.
 - Added Info Unit Macro.
 - URL field to the Post Preview organism
+- Frontend: Added overlay atom.
 
 ### Changed
 - Converted the project to Capital Framework v3
--- a/cfgov/jinja2/v1/_layouts/base.html
+++ b/cfgov/jinja2/v1/_layouts/base.html
@@ -129,6 +129,9 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
 
 {% block body %}
 
+{# Overlay for the page. Used for the mobile mega menu. #}
+<div class="a-overlay u-hidden"></div>
+
 <a href="#main" id="skip-nav">Skip to main content</a>
 
 {% if flag_enabled('BETA_NOTICE') %}
--- a/cfgov/unprocessed/css/main.less
+++ b/cfgov/unprocessed/css/main.less
@@ -67,6 +67,7 @@
 
 @import (less) "atoms/rule-break.less";
 @import (less) "atoms/multi-select.less";
+@import (less) "atoms/overlay.less";
 
 
 /* Molecule pieces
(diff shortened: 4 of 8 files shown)


cfgov/unprocessed/css/atoms/overlay.less
```diff
@@ -0,0 +1,37 @@
...
+*/
+
+.a-overlay {
+    // Only show overlay at mobile/tablet size.
+    .respond-to-max( @bp-sm-max, {
+        height: 100%;
+        width: 100%;
+
+        position: fixed;
+        z-index: 10;
```
sebworks: Why did you choose to set it to 10? Might need a note.

cfgov/unprocessed/css/atoms/overlay.less
```diff
@@ -0,0 +1,37 @@
...
+
+        background: @black;
+        // Whitespace in content so element can have dimensions set.
+        content: '\a0';
```
sebworks: can content be used outside of pseudo elements?

jimmynotjim: Nope - https://developer.mozilla.org/en-US/docs/Web/CSS/content

Sidenote, `/a0` creates an invisible character, so it influences sizing as well.

jimmynotjim:
Overlay doesn't show over the beta banner, not sure if that's intentional or not.

jimmynotjim:
Works well locally other than the note above.

sebworks:
screen shot 2016-02-24 at 3 06 49 pm

z -index may need to be higher

sebworks:
If the overlay is present, you scroll, then click on the page it scrolls you to the top.

sebworks:
Might consider fading in the overlay; the flicker is having a teletubby like affect on me.

jimmynotjim:
I thought the scrolling to the top was intentional, because the menu closed. That might be a question for @schaferjh or Erin (I can't get autocomplete to post her username).

sebworks:
looks good, :+1:

anselmbradford:
> Overlay doesn't show over the beta banner, not sure if that's intentional or not.

Yeah, that's intentional. It should only overlay content below the menu.

> Might consider fading in the overlay; the flicker is having a teletubby like affect on me.

Yeah I agree this would be a nice touch. I'll wait till the 3rd level menu is in because that's kind of transition syncing hell and listening to a fadeout/fadein will add more complexity, so best tackled on its own.

cfgov/unprocessed/css/atoms/overlay.less
anselmbradford: It seemed to cover everything, but as you noted elsewhere, not quite everything, so it should be higher! Will fix.

cfgov/unprocessed/css/atoms/overlay.less
anselmbradford: Interesting, I thought it wouldn't set the width and height without display:block or some content in the div, but I guess `position:fixed` takes care of it. Thanks!

anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/commit/5cfcf84a5c61150464c589181560325ff2335347

cfgov/unprocessed/css/atoms/overlay.less
anselmbradford: Added TODO to move this value to a standardized value. Updated Slick carousel arrow value to 9 from 1000 in https://github.com/cfpb/cfgov-refresh/commit/5cfcf84a5c61150464c589181560325ff2335347

anselmbradford: Fixes https://github.com/cfpb/cfgov-refresh/issues/692

schaferjh:
Re: scrolling to the top -- this is fine. 

The way I think about it, in order to open the menu in the first place, you had to be at the top of the page, so that's where you were last time you saw the full screen. Yes, it's possible, that after you've scrolled to the bottom of the somewhat-lengthy About Us (or Consumer Tools) menus, you spot something behind the overlay that catches your eye and want to see it immediately....but I think this is the less likely case.

Keep the scroll to top for now, we'll keep an eye on it in future user testing.

Thanks for checking!

sebworks:
:

jimmynotjim:
:+1:

PR #1496, opened 2016-02-24
Adding specific script for IE legacy versions

sebworks:
Adding specific script for IE legacy versions which allows us to remove legacy code for scripts. 

## Additions

- Adding specific script for IE legacy versions.

## Removals

- Removed conditional `class-list` check from various files.

## Testing
- Browse site on IE9 virtual machine.

## Review

- @anselmbradford 
- @KimberlyMunoz 
- @jimmynotjim 

## Notes

- Wasn't sure if other scripts should be included in the legacy build. 

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_layouts/base.html
+++ b/cfgov/jinja2/v1/_layouts/base.html
@@ -100,6 +100,8 @@
     </script>
     <![endif]-->
 
+    <!--[if IE 9]><script src="{{ static('js/ie/common.ie.js') }}"></script><![endif]-->
+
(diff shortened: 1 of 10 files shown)


anselmbradford:
Looks good!! :+1:
At some point we may want to isolate the tasks so that they fully work independently. So like spin out the modernizr build into its own task and have `scripts:modern` and `scripts:ie` both call it. It would have to handle not running the modernizr build twice though so that it runs once if each of the following tasks are run: `scripts`, `scripts:modern`, and `scripts:ie`.

jimmynotjim:
:+1:

PR #1509, opened 2016-02-26
Mega Menu/Search: Remove toggle / defer collapse

anselmbradford:
## Removals

- Removes `toggle` event from FlyoutMenu.

## Changes

- Adds collapse queue to FlyoutMenu so that if `collapse()` is called while expand is animating, it'll call collapse after the animation has completed.

## Testing

- On refresh.demo, click Search and immediately click Mega Menu. It's possible to have both expanded at the same time, which shouldn't happen:
screen shot 2016-02-26 at 2 46 36 pm
- Now do the same thing with this PR. It should not be possible to show both the mega menu and search expanded.

## Review

- @sebworks 
- @jimmynotjim

### DIFF
--- a/cfgov/unprocessed/js/modules/FlyoutMenu.js
+++ b/cfgov/unprocessed/js/modules/FlyoutMenu.js
@@ -2,6 +2,7 @@
 
 // Required modules.
 var EventObserver = require( '../modules/util/EventObserver' );
+var standardType = require( '../modules/util/standard-type' );
 
 /**
  * FlyoutMenu
@@ -30,6 +31,12 @@ function FlyoutMenu( element, triggerSel, contentSel, altTriggerSel ) {
   // Needed to add and remove events to transitions.
   var _expandEndBinded = _expandEnd.bind( this );
   var _collapseEndBinded = _collapseEnd.bind( this );
+
+  // Set this function to a queued collapse function,
+  // which is called if collapse is called while
+  // expand is animating.
+  var _deferFunct = standardType.noopFunct;
+  var _collapseBinded = collapse.bind( this );
 
   /**
    * @returns {Object} The FlyoutMenu instance.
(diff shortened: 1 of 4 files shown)


cfgov/unprocessed/js/modules/FlyoutMenu.js
```diff
@@ -65,7 +71,7 @@ function FlyoutMenu( element, triggerSel, contentSel, altTriggerSel ) {
    */
   function expand() {
     if ( !_isExpanded && !_isAnimating ) {
-      this.dispatchEvent( 'toggle', { target: this } );
+      _deferFunct = standardType.noopFunct;
```
sebworks: Why do you need to set the `deferFunct` to ` standardType.noopFunct;` everytime `expand` is called?

cfgov/unprocessed/js/modules/FlyoutMenu.js
```diff
@@ -108,11 +119,14 @@ function FlyoutMenu( element, triggerSel, contentSel, altTriggerSel ) {
...
   /**
    * Expand animation has completed.
+   * Call deferred collapse function,
+   * if set (otherwise it will call a noop function).
    */
   function _expandEnd() {
     _isAnimating = false;
     _contentDom.removeEventListener( _transitionEndEvent, _expandEndBinded );
     this.dispatchEvent( 'expandEnd', { target: this } );
+    _deferFunct.bind( this )();
```
sebworks: You should only have to bind the function once.

cfgov/unprocessed/js/modules/FlyoutMenu.js
anselmbradford: The queued collapse should only be called if `expand` was called and then `collapse` was called while the expand animation was underway. If expand is called and the `_deferFunct` has a value  other than noop, it means we can cancel the deferred collapse because that was set between the last call to expand and a call to collapse.

cfgov/unprocessed/js/modules/FlyoutMenu.js
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/commit/82087213261641c8e9f2487be77af9e7671f8bf6

sebworks:
:+1:

jimmynotjim:
Looks good :+1:

PR #1511, opened 2016-02-26
Bugfix adds proper spacing between two inline info-units at all sizes

jimmynotjim:
Bugfix adds proper spacing between two inline info-units at all sizes

## Changes

- Removes the media query wrapping the defined top margin

## Testing

- `gulp build` and if you have local data visit `/policy-compliance/`, otherwise build a page w/ multiple 25/75 molecules

## Review

- @anselmbradford 
- @sebworks 

## Screenshots

__Before__
screen shot 2016-02-26 at 5 52 26 pm

__After__
screen shot 2016-02-26 at 5 48 02 pm

## Notes

- Still uses the !important that I'd like to remove, but that's outside the scope

## Todos

- Still uses the !important that I'd like to remove, but that's outside the scope

## Checklist

- Removes the media query wrapping the defined top margin
- Still uses the !important that I'd like to remove, but that's outside the scope

### DIFF
--- a/cfgov/unprocessed/css/organisms/info-unit-group.less
+++ b/cfgov/unprocessed/css/organisms/info-unit-group.less
@@ -44,15 +44,15 @@
     - cfgov-organisms
 */
 .o-info-unit-group {
+    .content-l_col-1 + .content-l_col-1 {
+        // Adds margin top to info units that follow another
+        margin-top: unit( @grid_gutter-width / @base-font-size-px, em ) !important;
+    }
+
     .respond-to-min(@bp-sm-min, {
         .content-l_col-1-2:nth-of-type(n+3) {
             // Adds margin top to info units that are 3rd and up in a group
             margin-top: unit(@grid_gutter-width / @base-font-size-px, em);
-        }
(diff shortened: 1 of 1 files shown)


cfgov/unprocessed/css/organisms/info-unit-group.less
```diff
@@ -44,16 +44,16 @@
     - cfgov-organisms
 */
 .o-info-unit-group {
+    .content-l_col-1 + .content-l_col-1 {
+        // Adds margin top to info units that follow another
+        margin-top: unit( @grid_gutter-width / @base-font-size-px, em ) !important;
```
sebworks: Might want to add a note why the `!important` is necessary.

sebworks:
:+1:

cfgov/unprocessed/css/organisms/info-unit-group.less
jimmynotjim: I've got a PR coming next that's removing it. I couldn't find any reason the original rule in overrides exists

PR #1518, opened 2016-02-29
Bugfix: added js init scripts for old pages.

anselmbradford:
The secondary nav was not being initialized on some non-wagtail pages. This adds initialization scripts for those routes.

## Additions

- Added js init scripts for non-wagtail pages at `/offices/`, `/sub-pages/`, and `/budget/` routes.

## Testing

- `gulp build`
- visit `/offices/`, `/sub-pages/`, and `/budget/` URLs
- resize to mobile
- navigate section using the "In This Section" secondary nav.

## Review

- @sebworks 
- @KimberlyMunoz 
- @jimmynotjim 
- @kurtw 
- @schaferjh 

## Screenshots

screen shot 2016-02-29 at 4 00 08 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Signal receiver function to unpublish all revisions for a page when a page is unpublished
 - Backend: HomePage Model
 - David Silberman's assets
+- Frontend: Added JS init scripts for /offices/, /sub-pages/, and /budget/.
 
 ### Changed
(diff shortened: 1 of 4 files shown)


jimmynotjim:
:+1: 

Side thought, did this break when the routes were removed? I thought I'd tested these pages when I updated it.

anselmbradford:
> Side thought, did this break when the routes were removed? I thought I'd tested these pages when I updated it.

I'm not sure what broke these. Seems like the atomic template routes wouldn't be applied so wouldn't have an effect when removed.

PR #1519, opened 2016-02-29
Minor linter build file cleanup

anselmbradford:
## Changes

- Fixes errors in `gulp lint:build`

## Testing

- `gulp lint:build` should show 0 errors and 2 TODO warnings.

## Review

- @KimberlyMunoz 
- @sebworks 
- @jimmynotjim

### DIFF
--- a/cfgov/unprocessed/js/modules/util/assign.js
+++ b/cfgov/unprocessed/js/modules/util/assign.js
@@ -11,10 +11,9 @@
 
 
 /**
-* @param {object} JavaScript object.
+* @param {Object} object - JavaScript object.
 * @returns {boolean} True if object is plain Javascript object.
 */
-
 function _isPlainObject( object ) {
   return Object.prototype.toString.call( object ) === '[object Object]';
 }
@@ -24,10 +23,9 @@ function _isPlainObject( object ) {
 * existing properties. When assigning from multiple sources, fields of every
 * next source will override same named fields of previous sources.
 *
-* @param {object} destination object.
-* @returns {object} assigned destination object.
+* @param {Object} destination object.
+* @returns {Object} assigned destination object.
 */
-
 function assign( destination ) {
   destination = destination || {};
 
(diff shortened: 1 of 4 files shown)


jimmynotjim:
:+1:

cfgov/unprocessed/js/modules/util/assign.js
```diff
@@ -11,10 +11,9 @@
 
 
 /**
-* @param {object} JavaScript object.
+* @param {Object} object - JavaScript object.
```
KimberlyMunoz: This might be a silly question, but why is {Object} capitalized and {boolean} not?

sebworks:
:+1:

cfgov/unprocessed/js/modules/util/assign.js
anselmbradford: There are no silly questions! :-) 
So it's due to the jsdocs describing the type of data a function is expecting. Either the constructor of an object (capitalized) or a description (lowercase). JS has primitive datatypes that allow creating variables with ease, like `var thisIsAString = 'hello';` for a string. But it also has wrapper functions for its primitives. With these, new strings can be created in a more object-oriented way like `var thisIsAMoreComplexString = new String( 'hello' );`. Similarly, booleans can be created with `primitive = true;` or `primitive = new Boolean( false );`.

If the jsdocs say `* @param {String} greeting - a word to display.`, it's saying the function implementation should look like `sayGreeting( new String( 'hello' ) );`—it's expecting the type passed in as a parameter to be a String object, not a string primitive. Or as a Boolean, it would be expecting `isItPrimitive( new Boolean( false ) );`. Objects on the other hand are not primitive and are instantiated through `new Object()`, so the type shown in jsdocs is the constructor `Object`. It's not very common to see `new Object()` in JS, but that's just because JS includes a shorthand for `new Object()` as `{}`. Either syntax is still a complex object, not a primitive.

Functionally, the difference likely doesn't matter since JS converts primitives to complex objects as needed to run methods and so forth (as you can see from console logging a primitive—it doesn't have methods). But theoretically a method could be written to expect an object instead of a primitive, so getting the types right is the final polish. So when I'm in the file and see they're wrong I'll update them.

Sidenote: things like `{integer}` is a stylistic type in that it isn't a true type within JS, but `number` doesn't convey whether the number expected is a whole number or a floating point number, so `integer` can be used to convey a whole number primitive.

KimberlyMunoz: :100:

PR #1526, opened 2016-03-01
Bugfix and Minor cleanup

anselmbradford:
## Changes

- Shortened atomic dom checker module and method name so that the call would fit on one line more often.
- Fixed missing flyout data-* attribute for global search.
- Fixed missing aria-hidden value on global search.
- Added some todos and code comments.

## Testing

- `gulp build` should work and site shouldn't have errors.

## Review

- @KimberlyMunoz 
- @sebworks 
- @jimmynotjim


sebworks:
:+1: , we might consider creating a `bindAll` method.

PR #1535, opened 2016-03-03
missing hostname param

kave:
@richaagarwal


cfgov/v1/forms.py
```diff
@@ -221,7 +221,7 @@ def set_topics(self, parent, hostname):
     # Populate Authors' choices
     def set_authors(self, parent, hostname):
         all_authors = [author for authors in [page.authors.names() for page in
-                       AbstractFilterPage.objects.live_shared().descendant_of(
```
kave: missing param value was breaking filter controls

PR #1542, opened 2016-03-07
Mega Menu: Adds Tree and Tree traversal data structure

anselmbradford:
Adds Tree data structure for more easily querying the submenus of the mega menu. Also adds utility tree traversal algorithms.

## Additions

- JS Tree data structure.
- JS Tree traversal algorithms.

## Testing

- Can be tested by adding the following snippet to a page's JS (common.js or Header.js, or whatever):

```
var Tree = require( '../modules/Tree' );
var treeTraversal = require( '../modules/util/tree-traversal' );
/**
 * Make this Tree:
 *        R
 *       / \
 *      A   B
 *    / | \
 *   C  D  E
 */
var tree = new Tree();
tree.init( 'R' );
var nodeR = tree.getRoot();
var nodeA = tree.add( 'A', nodeR );
var nodeB = tree.add( 'B', nodeR );
var nodeC = tree.add( 'C', nodeA );
var nodeD = tree.add( 'D', nodeA );
var nodeE = tree.add( 'E', nodeA );

var level1 = tree.getAllAtLevel( 1 );
for ( var i = 0, len = level1.length; i < len; i++ ) {
  console.log( 'level1:', level1[i].payload );
}

treeTraversal.backtrack( nodeC, function( node ) {
  console.log( 'backtrack:', node.payload );
} );

treeTraversal.bfs( nodeR, function( node ) {
  console.log( 'bfs:', node.payload );
} );

treeTraversal.dfs( nodeR, function( node ) {
  console.log( 'dfs:', node.payload );
} )
```

Outputs:

```
level1: A
level1: B
backtrack: C
backtrack: A
backtrack: R
bfs: R
bfs: A
bfs: B
bfs: C
bfs: D
bfs: E
dfs: R
dfs: A
dfs: C
dfs: D
dfs: E
dfs: B
```

## Review

- @sebworks 
- @KimberlyMunoz 
- @jimmynotjim 

## Todos

- Perfect candidate for adding unit tests to.

### DIFF
--- a/cfgov/unprocessed/js/modules/Tree.js
+++ b/cfgov/unprocessed/js/modules/Tree.js
@@ -1,0 +1,120 @@
+'use strict';
+
+/**
+ * Tree
+ * @class
+ *
+ * @classdesc A tree data structure.
+ * Tree's have one root node, and child nodes that branch.
+ * Like:
+ *
+ *        R
+ *       / \
+ *      A   B
+ *    / | \
+ *   C  D  E
+ *
+ * @returns {Tree} An instance.
+ */
+function Tree() {
+
+  var _root = null;
+  var _levelCache = {};
+
+  /**
+   * @param {Object} data - Data to attach to the root node.
+   * @returns {Tree} An instance.
+   */
+  function init( data ) {
+    _root = new TreeNode( this, data );
+    _levelCache[0] = [ _root ];
+
+    return this;
+  }
+
+  /**
+   * @param {Object} data - Data to attach to new node.
+   * @param {TreeNode} parent - Node on which to add a new child node.
+   * @returns {TreeNode} The newly instantiated and added node.
+   */
+  function add( data, parent ) {
+    var child = new TreeNode( this, data, parent );
+
+    // Save node at each level as a flat array.
+    var level = child.level;
+    if ( _levelCache[level] ) {
+      _levelCache[level].push( child );
+    } else {
+      _levelCache[level] = [ child ];
+    }
+
+    parent.children.push( child );
+
+    return child;
+  }
+
+  /**
+   * @returns {TreeNode} The root node of this Tree.
+   */
+  function getRoot() {
+    return _root;
+  }
+
+  /**
+   * Get all nodes at a particular tree level. For example, returning
+   * Given this Tree and starting at (R):
+   *
+   *        R
+   *       / \
+   *      A   B
+   *    / | \
+   *   C  D  E
+   * Level 0 nodes would return (R).
+   * Level 1 nodes would return (A) and (B).
+   * Level 2 nodes would return (C), (D), and (E).
+   *
+   * @param {number} level - The tree level on which to return nodes.
+   * @returns {Array} A list of all nodes at a particular tree level.
+   */
+  function getAllAtLevel( level ) {
+    return _levelCache[level];
+  }
+
+  // TODO: Implement remove method.
+  // function remove( child, parent ) {
+  //
+  // }
+
+  this.add = add;
+  this.init = init;
+  this.getRoot = getRoot;
+  this.getAllAtLevel = getAllAtLevel;
+
+  return this;
+}
+
+// PRIVATE CLASS
+/**
+ * TreeNode
+ * @class
+ *
+ * @classdesc A node in a tree data structure.
+ *
+ * @param {Tree} tree - The data structure this node is a member of.
+ * @param {Object} data - The data data.
+ * @param {TreeNode} parent - The parent node in the root,
+ *   null if this is the root node.
+ * @param {Arrray} children - List of children nodes.
+ * @returns {TreeNode} An instance.
+ */
+function TreeNode( tree, data, parent, children ) {
+  this.tree = tree;
+  this.data = data;
+  this.parent = parent;
+  this.children = children || [];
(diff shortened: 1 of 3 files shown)


contolini:
This seems like a good candidate for a standalone node module. It would make unit testing easier.

Scotchester:
Relatedly, https://www.npmjs.com/search?q=tree

None of those have many stars, but is there a reason we're writing our own?

KimberlyMunoz:
I'd feel uncomfortable merging this without a few unit tests outlining the cases in the documentation. The documentation is great, but the unit tests would make sure they're never out of date.

cfgov/unprocessed/js/modules/Tree.js
```diff
@@ -0,0 +1,122 @@
...
+
+    return this;
+  }
+
+  /**
+   * @param {Object} payload - Data to attach to new node.
+   * @param {TreeNode} parent - Node on which to add a new child node.
+   * @returns {TreeNode} The newly instantiated and added node.
+   */
+  function add( payload, parent ) {
```
sebworks: Instead of payload, can we call it `data`? I think payload has a different connotation within the computing.

cfgov/unprocessed/js/modules/Tree.js
```diff
@@ -0,0 +1,122 @@
...
+    if ( !parent.children ) { parent.children = []; }
```
sebworks: Why is this check necessary? Doesn't the parent `TreeNode` always have children? https://github.com/cfpb/cfgov-refresh/pull/1542/files#diff-6ea501203f930030236f6ca28adaf8feR116

cfgov/unprocessed/js/modules/util/tree-traversal.js
```diff
@@ -0,0 +1,82 @@
...
+ *    / | \
+ *   C  D  E
+ *
+ * Returns (R) -> (A) -> (B) -> (C) -> (D) -> (E).
+ *
+ * @param {TreeNode} node - Node within a tree to traverse from.
+ * @param {Function} callback - Function to call at each node.
+ */
+function bfs( node, callback ) {
+  var queue = [ node ];
```
sebworks: Will this alway be one element? if so, why put it in the array?

cfgov/unprocessed/js/modules/Tree.js
anselmbradford: I was going off of the advice [here](http://eslint.org/docs/rules/id-blacklist)
> Bad names can lead to hard to decipher code. Using generic names, such as data don’t infer much about the code and the values it receives.

It's specifically a something that gets carried around with a TreeNode, which seemed to be what payload described, but if that has another connotation, what do you think?

cfgov/unprocessed/js/modules/Tree.js
anselmbradford: Mm, good find! I'll remove it.

cfgov/unprocessed/js/modules/util/tree-traversal.js
anselmbradford: No, it gets loaded with additional nodes while traversing the tree, see L:49 https://github.com/cfpb/cfgov-refresh/pull/1542/files#diff-327a6706d58da50fa37f20076162e405R49

sebworks: So why not just loop the children?

cfgov/unprocessed/js/modules/Tree.js
sebworks: I might be skewed by DOM nodes ( `data/dataset` ) but it's what I naturally think about when in comes to nodes. I don't think people normally wouldn't associate the term `payload` with a tree node.  I could be wrong though.

cfgov/unprocessed/js/modules/util/tree-traversal.js
anselmbradford: > So why not just loop the children? You could still call the callback for the parent. Could you make the dfs function generic?

Not sure I follow, do you mean utilize the dfs function so it can be called from the bfs?
Order is going to matter. If, while looping over the children, the grandchild (children of the children) are visited before the siblings, it will be a depth-first search, not a breadth-first search. The queue is used to push the grandchildren into a queue that are then visited after all siblings have been visited.

sebworks: sounds good :+1: , I was just suggesting that we combine the functions but that's a bad ideal.

sebworks:
:+1: , from me. Test would be nice though.

cfgov/unprocessed/js/modules/Tree.js
anselmbradford: I'll update it to `data`, that works.

jimmynotjim:
Ditto the notes on tests, but lets merge it to make reviewing 1572 easier and add tests first thing next sprint. :+1:

cfgov/unprocessed/js/modules/Tree.js
anselmbradford: Fixed!

cfgov/unprocessed/js/modules/Tree.js
anselmbradford: Fixed!

PR #1544, opened 2016-03-07
Data-hook fixes + Minor cleanup

anselmbradford:
## Removals

- Removes unneeded classlist polyfill that is covered by the IE9 polyfill.

## Changes

- Moves code comment above correct line it refers to in common.js
- Updates `attr` variables to `value` to more accurately reflect what it represents.
- Fixes TODO in data-hook.js `contains` method.
- Fixes issue in data-hook.js where if an element didn't already have a `data-js-hook` value, it would get one with a value of `null`.

## Testing

- You can try this on any element and check the page source, such as the overlay in Header.js:
```
var dataHook = require( '../modules/util/data-hook.js' );
dataHook.add( _overlay, 'test' );
dataHook.remove( _overlay, 'test' );
console.log( dataHook.contains( _overlay, 'test' ) );
```

## Review

- @jimmynotjim 
- @sebworks 
- @KimberlyMunoz 

## Notes

- When removing all the values from  `data-js-hook`, the attribute itself still remains on the element (but without any value). I think the overhead of checking whether all values have been removed and then removing the attribute probably isn't worth it.

### DIFF
--- a/cfgov/unprocessed/js/modules/util/data-hook.js
+++ b/cfgov/unprocessed/js/modules/util/data-hook.js
@@ -5,53 +5,59 @@ var standardType = require( './standard-type' );
 
 /**
  * @param {HTMLNode} element - DOM element.
- * @param {string} attr
- *   Attribute to add to the JS data-* hook value.
+ * @param {string} value
+ *   Value to add to the element's JS data-* hook.
  * @throws {Error} If supplied value contains a space,
  *   which would mean it would be two values, which is likely a typo.
  * @returns {string} The value that was added.
  */
-function add( element, attr ) {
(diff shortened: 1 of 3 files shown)


jimmynotjim:
Works locally :+1:

sebworks:
:+1:

KimberlyMunoz:
What is the IE9 Polyfill that replaces ClassList?

anselmbradford:
@KimberlyMunoz You got this covered with https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/unprocessed/js/ie/common.ie.js right?

PR #1545, opened 2016-03-07
Header: Adds transition utilities

anselmbradford:
Updates Global Search animation to use separate transition module, which eventually the mega menu will use also.

## Additions

- Added utility classes for translation and opacity CSS transitions.

## Changes

- Updates flyout menu to support transition utilities.

## Testing

- Global search and mega menu should work the same as on refresh.demo.

## Review

- @jimmynotjim 
- @KimberlyMunoz 
- @sebworks 

## Notes

- The flyout menu can either use no transition, in which case `expandBegin` and `expandEnd` are called subsequent to each other, or a transition can be added, in which case `expandEnd` is called after the `transitionEnd` event fires. The same architecture is present for the collapse methods. A different transition instance can be used for the expand and collapse, if, for example, a menu should move into view when opened, but fade out of view when closed.

## Todos

- This is prep for the transitions used in the mega menu hamburger menu.

### DIFF
--- a/cfgov/unprocessed/css/misc.less
+++ b/cfgov/unprocessed/css/misc.less
@@ -655,6 +655,78 @@
     }
 }
 
+
+/* topdoc
+  name: Transition utility classes.
+  family: cfgov-misc
+  notes:
+    - "Adds transitions utilty classes for transform, opacity,
+       and for the removing the transition duration."
+  tags:
+    - cfgov-misc
+*/
+.u-move-transition {
+    transition: transform 0.25s ease-out;
+}
+
+.u-alpha-transition {
+    transition: opacity 0.25s ease-out;
+}
+
+.u-no-animation {
+    transition-duration: 0s;
+}
+
+/* topdoc
+  name: Move utility classes.
+  family: cfgov-misc
+  notes:
+    - "Utility classes for moving an element using transform translate values."
+  tags:
+    - cfgov-misc
+*/
+.u-move-to-origin {
+    transform: translateX( 0 );
+}
+
+.u-move-left {
+    transform: translateX( -100% );
+}
+
+// TODO: Look into adding a mixin for movement multiples.
+.u-move-left-2x {
+    transform: translateX( -200% );
+}
+
+.u-move-left-3x {
+    transform: translateX( -300% );
+}
+
+.u-move-right {
+    transform: translateX( 100% );
+}
+
+.u-move-up {
+    transform: translateY( -100% );
+}
+
+/* topdoc
+  name: Alpha utility classes.
+  family: cfgov-misc
+  notes:
+    - "Utility classes for setting an element's opacity."
+  tags:
+    - cfgov-misc
+*/
+.u-alpha-100 {
+    opacity: 1;
+}
+
+.u-alpha-0 {
+    opacity: 0;
+}
+
+
 /* topdoc
   name: EOF
   eof: true
--- a/cfgov/unprocessed/css/molecules/global-search.less
+++ b/cfgov/unprocessed/css/molecules/global-search.less
@@ -125,15 +125,6 @@
         padding-top: unit( @grid_gutter-width / 4 / @base-font-size-px, em );
         position: absolute;
 
-        transform: translateX( 100% );
-        transition: transform 0.25s ease-out;
-
-        &[aria-expanded="true"] {
-            display: block;
-
-            transform: translateX( 0 );
-        }
-
         &-suggestions {
             display: none;
 
@@ -302,6 +293,11 @@
 .no-js .m-global-search {
     &_trigger {
         display: none;
+    }
+
+    // JS isn't available to remove u-hidden class, so override it.
+    &_content.u-hidden {
+        display: block;
     }
 }
 
--- a/cfgov/unprocessed/js/modules/FlyoutMenu.js
+++ b/cfgov/unprocessed/js/modules/FlyoutMenu.js
@@ -9,30 +9,52 @@ var standardType = require( '../modules/util/standard-type' );
  * FlyoutMenu
  * @class
  *
- * @classdesc Initializes a new FlyoutMenu molecule.
+ * @classdesc Initializes new FlyoutMenu behavior.
+ * As added JS behavior, this is added through HTML data-js-hook attributes.
  *
- * @param {HTMLNode} element
- *   The DOM element within which to search for the molecule.
+ * Structure is:
+ * flyout-menu
+ *   flyout-menu_trigger
+ *   flyout-menu_content
+ *     flyout-menu_alt-trigger
+ *
+ * The alt-trigger is for a back button, which may obscure the first trigger.
+ * The flyout can be triggered three ways: through a click of the trigger or
+ * through the click of an alt-trigger.
+ *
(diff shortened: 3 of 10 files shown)


cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -303,6 +294,11 @@
     &_trigger {
         display: none;
     }
+
+    // JS isn't available to remove u-hidden class, so override it.
+    &_content.u-hidden {
+      display: block;
```
jimmynotjim: Update to four spaces.

jimmynotjim:
Works locally, we should probably add tests for the header stuff soon, cause there's a lot going on here.

cfgov/unprocessed/js/modules/FlyoutMenu.js
```diff
@@ -9,31 +9,53 @@ var standardType = require( '../modules/util/standard-type' );
...
-  var _altTriggerDom = _dom.querySelector( BASE_SEL + '_alt-trigger]' );
+  var _triggerDom = _dom.querySelector( TRIGGER_SEL );
+  var _contentDom = _dom.querySelector( CONTENT_SEL );
 
-  var _isExpanded = false;
+  if ( !_triggerDom ) { throw new Error( TRIGGER_SEL + ' is missing!' ); }
+  if ( !_contentDom ) { throw new Error( CONTENT_SEL + ' is missing!' ); }
 
-  var _transitionEndEvent = _getTransitionEndEvent( _contentDom );
+  var _altTriggerDom = _dom.querySelector( ALT_TRIGGER_SEL );
```
sebworks: Do you need to throw an error if `_altTriggerDom`?

cfgov/unprocessed/js/modules/FlyoutMenu.js
```diff
@@ -92,14 +114,17 @@ function FlyoutMenu( element ) {
       this.dispatchEvent( 'expandBegin', { target: this } );
       _isExpanded = true;
       _isAnimating = true;
-      // If transition is not supported, call handler directly (IE9/OperaMini).
-      if ( _transitionEndEvent ) {
-        _contentDom.addEventListener( _transitionEndEvent, _expandEndBinded );
+      if ( _expandTransitionMethod ) {
+        _expandTransitionMethod.call();
```
sebworks: Why are you using `call` here? Is it because `_expandTransitionMethod` is bound to a different scope?

cfgov/unprocessed/js/modules/FlyoutMenu.js
anselmbradford: No, the alternative trigger is optional and is checked on https://github.com/cfpb/cfgov-refresh/blob/adds_transitions/cfgov/unprocessed/js/modules/FlyoutMenu.js#L76

cfgov/unprocessed/js/modules/FlyoutMenu.js
anselmbradford: Hm, yeah I think my original thinking was the different scope. Since `_expandTransitionMethod` is not a function defined in `FlyoutMenu`, it seemed deceptive to have it as `_expandTransitionMethod()` since that looks like every call to a private method within `FlyoutMenu`.

cfgov/unprocessed/css/misc.less
```diff
@@ -655,6 +655,77 @@
...
+.u-move-to-origin {
+    transform: translateX( 0 );
+}
+
+.u-move-left {
+    transform: translateX( -100% );
+}
+
+.u-move-left-2x {
+    transform: translateX( -200% );
```
sebworks: what is the value in having `2x` and `3x` utility classes? These seem like they are specific to the menu.  Can they be dynamic where you can pass in a multiple?

cfgov/unprocessed/js/modules/FlyoutMenu.js
sebworks: ok, if it's something that we are following in other places then I'm ok with it but if it's just here then I would probably leave a comment.

cfgov/unprocessed/js/modules/transition/BaseTransition.js
```diff
@@ -0,0 +1,203 @@
...
+      }
+    }
+  }
+
+  /**
+   * @param {string} clazz - A CSS class.
+   * @returns {boolean} False if the class is already applied,
+   *   otherwise true if the class was applied.
+   */
+  function applyClass( clazz ) {
```
sebworks: can we just call this `className`?

cfgov/unprocessed/js/modules/transition/MoveTransition.js
```diff
@@ -0,0 +1,160 @@
+'use strict';
+
+// Required modules.
+var EventObserver = require( '../../modules/util/EventObserver' );
+var BaseTransition = require( './BaseTransition' );
+
+// Exported constants.
+var CLASSES = Object.seal( {
```
sebworks: Using `Object.seal` won't prevent someone from changing the existing values. it will prevent someone from adding new values though.

sebworks:
:+1:

cfgov/unprocessed/css/molecules/global-search.less
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/commit/3d4d642051a9befc68b2f98d92f696c48849c909

cfgov/unprocessed/js/modules/transition/BaseTransition.js
jimmynotjim: :raised_hands: `clazzhands` :raised_hands:

cfgov/unprocessed/js/modules/FlyoutMenu.js
anselmbradford: Updated and added a comment in https://github.com/cfpb/cfgov-refresh/commit/1dcbb2a9ab7d759a71deebbe145689c1c395f2b1

cfgov/unprocessed/js/modules/transition/BaseTransition.js
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/commit/17b34b701b909763b85eed401dbd46e07b0959d1

cfgov/unprocessed/css/misc.less
anselmbradford: Yeah without editing the CSS from JS, I couldn't see a way around this (and I think it would be preferable to keep the JS/CSS separate). Let's re-visit when the full menu is in and maybe you'll see another way around it.

jimmynotjim: I can see the value in the `2x` being a classname available for use directly in markup. Might still be a good idea to create a base mixin that these classes then call.

cfgov/unprocessed/js/modules/transition/MoveTransition.js
anselmbradford: Mm true!

cfgov/unprocessed/css/misc.less
sebworks: sorry, I meant a base mixin as Jimmy mentioned.

sebworks:
:+1:

PR #1549, opened 2016-03-08
Fixed issue with filter clear button

jimmynotjim:
Fixed issue where filter form clear button didn't submit the cleared form updating the results.

## Changes

- Added from submission to clear button click event.

## Testing

- `gulp build` and navigate to `/blog/` set some filters and submit. Then click the reset button and the form should submit again sans-filtered options.

## Review

- @anselmbradford 
- @KimberlyMunoz 
- @sebworks 

## Todos

- We should move this off jQuery when we have the time.
- Ajax would be a great enhancement here

## Checklist


sebworks:
Do we always want to submit the forms on clear?

KimberlyMunoz:
The fix looks good, but would we want the form to submit if there's no filtering being done and the user just wants to clear topics/authors.

:+1:

jimmynotjim:
@sebworks Do we use this anywhere other than on filters? 

@KimberlyMunoz topics/authors are filters, I'm not sure I understand.

sebworks:
nope, :+1:

KimberlyMunoz:
> @KimberlyMunoz topics/authors are filters, I'm not sure I understand.

I meant that if a person were to start filtering and wanted to clear the fields before submitting, it would force a refresh instead of just clearing the fields. It's an edge case, so I don't think it should hold up this fix.

schaferjh:
@KimberlyMunoz raises an interesting edge case, but I agree it shouldn't hold up this fix. In futureland, I think we'll want to see how much we can do without full page refresh, but it's not needed for our initial go.

jimmynotjim:
@KimberlyMunoz Ah I see, that's something we can definitely revisit, this entire thing should be re-written anyway.

@schaferjh Yep, I noted that we could add Ajax to make these feel more dynamic and avoid page reloads altogether, but it's def a v2 thing.

PR #1551, opened 2016-03-08
use the django default password hashing

rosskarchner:
Explicitly calls out that we are using the default Django password hashing configuration

(this is useful for our security documentation)


kave:
:+1:

PR #1554, opened 2016-03-09
Fix sharing bug that would prevent the display of tags

kurtw:
The wagtail_hooks.py file deletes the latest revision and creates a new one. It's not known why a new revision would lose the tag references, but modifying the latest instead of deleting it and creating a new one seems like a better choice anyway.

## Changes

- sharing a page now modifies the wagtail created revision instead of deleting it and creating a new one

## Testing

- Create, save, share, publish a page with tags.
 - Verify that the tags are saved at each step.

## Review

- @kave


kave:
:+1: make sense

Issue #1556, opened 2016-03-09
Extra spacing on sublanding page


Schlachtmeyer:
Not sure why this sublanding page is displaying so much space below the hero.

## Current behavior
![image](https://cloud.githubusercontent.com/assets/14129916/13644148/4ebadaac-e5f2-11e5-8656-254f8f6a3a3a.png)

## Expected behavior
Even adding a header above the half-width link blob doesn't fill in the spacing or make it even with the sidebar content. And the gap between the hero and the sidebar seems extra wide to me?

## Steps to replicate behavior

http://content.refresh.demo.cfpb.gov/data-research/credit-card-data/

KimberlyMunoz:
The hero needs a `block_flush` added to it. That seems to be a FEWD issue, so I'll look into it. 

Here's what it looks like when it's added.

![image](https://cloud.githubusercontent.com/assets/1860176/13650336/2bc0a220-e610-11e5-8051-f2055ff2529c.png)

Schlachtmeyer:
Looks better! I may still try adding a header for the link blob group, but this helps.

KimberlyMunoz:
LGTM!

PR #1557, opened 2016-03-09
Add SublandingFilterablePage type 

richaagarwal:
From Jira: 

We need a quick new page type in Wagtail to satisfy content that we need to forklift into the site – a sublanding page that can house a filterable list. We should make this as a separate option rather than a modification to the existing type. Jessica and Emmanuel chatted/emailed about this briefly already.
Requirements:
Top of the page is the same as the top of the standard sublanding page (can be hero or text intro)
Page content is the same as a browse filterable page (can have a FCM, some copy, and the filter and post previews pulled from children document detail pages)
Side bar is the same as the sidebar options on the standard sublanding page
Development Considerations:
Clone BrowseFilterablePage Model in `browse_filterablte_page.py` ==> SublandingFilterablePage
Relate it to the Sublanding page template or ask FEWDS if they want to create a new template to determine how its rendered on the page
Verify with content if the sublandingFilterablePage will include all the prior sublanding page atomic elements + FilterControls or if there will be other changes

## Additions

- SublandingFilterablePage 

## Testing

- Run migrations
- Add a child page of type 'Sublanding Filterable Page' 

## Review

- @kave @kurtw 

## Screenshots
How the page looks:
screen shot 2016-03-09 at 10 27 15 am

## Notes

- Who from content or FEWD should I tag?

## Checklist


cfgov/v1/models/browse_filterable_page.py
```diff
@@ -124,4 +124,4 @@ def get_form_class(self):
         return forms.EventArchiveFilterForm
 
     def get_template(self, request, *args, **kwargs):
-        return BrowseFilterablePage.template
+        return BrowseFilterablePage.template
```
kave: lets revert this space

richaagarwal: :+1: yep that slipped through, will do

cfgov/v1/models/sublanding_filterable_page.py
```diff
@@ -0,0 +1,66 @@
...
+        FieldPanel('secondary_nav_order'),
+    ]
+
+    # Tab handler interface
+    edit_handler = TabbedInterface([
+        ObjectList(content_panels, heading='General Content'),
+        ObjectList(sidebar_panels, heading='Sidebar'),
+        ObjectList(CFGOVPage.settings_panels, heading='Configuration'),
+    ])
+
```
kave: We will also need to include all the functions from the browseFilterable model `get_context` `get_form_class` etc.

cfgov/v1/models/sublanding_filterable_page.py
```diff
@@ -0,0 +1,66 @@
...
+    template = 'sublanding-page/index.html'
```
kave: in the `sublanding/index.html` lets add the section from the browse filterable template that handles filter controls in the `page.content` loop
```
{% if 'filter_controls' in block.block_type %}
            <div class="block
                    block__flush-top">

                {% set f = forms.pop(0) %}
                {% from 'organisms/filterable-list-controls.html' import render as flc with context %}
                {{ flc(block.value, f, loop.index0) }}
            </div>
        {% else %}
```

kave:
@KimberlyMunoz including a FEWD for frontend verification

kave:
Can you test this page type by adding a `Filter Control` to this page and child documentDetail/learn pages to filter on. so we can be sure the expected behaviour

KimberlyMunoz:
Sublanding pages with heroes shouldn't be able to display breakout slugs. Breakout slugs can only exist with Text Intro molecules.

There's also some weird spacing with the hero, but it looks like it was reported in #1556 so I'll take a look at that outside of this PR.

KimberlyMunoz:
I'm testing this now and you can just remove all the sidebar breakout modules there. I can't seem to get the filters to actually show up when I add it to the page though.

cfgov/v1/models/sublanding_filterable_page.py
kurtw: It probably makes sense to put this in the sublanding-page/index.html for now, but later on things like this will merit a catch-all streamfield file to include in these page templates.

cfgov/v1/models/sublanding_filterable_page.py
kurtw: Same with `get_page_set` and `get_form_specific_filter_data`, although you could just call those methods from the BrowseFilterablePage class instead since they don't actually call the object itself...which also make me think maybe it should be pulled out of the class and into utils or something. It's up to you.

KimberlyMunoz:
:+1:

There's conflicting migrations in Travis, but other than that it looks great!

richaagarwal:
@kave @kurtw if one of you can do a final code review, i'll merge this in!

cfgov/v1/util/filterable_context.py
```diff
@@ -0,0 +1,83 @@
+from .. import forms
+from .util import get_secondary_nav_items
+from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
+from django.db.models import Q
+from ..models import base, molecules, organisms, ref
+from ..models.learn_page import AbstractFilterPage
+
+
```
kave: ooo vary nice

kave:
looks good, just one issue. Some other models like `SublandingPage` get_browsefilterable_posts function call some filterable list methods that have been moved to the util function. So lets move those functions back in like so in the `BrowseFilterablePage` & `SublandingFilterablePage`
```
        context = super(SublandingFilterablePage, self).get_context(request, *args, **kwargs)
        return filterable_context.get_context(self, request, context)

    def get_form_class(self):
        return filterable_context.get_form_class(self)

    def get_page_set(self, form, hostname):
        return filterable_context.get_page_set(self, form, hostname)
``

richaagarwal:
@kave oh good catch, thanks!

PR #1558, opened 2016-03-09
Allow Content editing of HomePage

kave:
## Additions

- Allow editing of Latest Updates Section & Text below Hero

## Testing

- Create a `HomePage` and add content 

## Review

- @kurtw 
- @richaagarwal 
- @jimmynotjim 

## Screenshots

screen shot 2016-03-09 at 1 37 51 pm
screen shot 2016-03-09 at 1 37 42 pm

## Checklist


kave:
@jimmynotjim I noticed the hyperlinks seem to be in a different color when using the half link blob molecules. Are we missing a hyperlink css class or something in the template?

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -213,7 +213,7 @@
 
     {# TODO: Remove Else conditional when this is fully moved to wagtail. #}
     {# Check and include page-level JavaScript. #}
-    {% if page and page.media %}
+    {% if page %}
```
richaagarwal: If there is a page but no page.media, no js will be included - is this intended or is that a situation we no longer anticipate?

sebworks: Does every page have a page object?

schaferjh:
Question about the category options -- will the content editors be able to use all the categories under newsroom? This is an *extra* so feel free to say no for this initial version, but I had a query from EA indicating that they would like to be able to select "press release" "speech" etc (which are currently categories on the newsroom). 

If we did this, we would still want to be able to use main categories like "Blog" "Event" "Report" -- so if it's too finicky to enable both levels of categories right now, that's fine.

cfgov/jinja2/v1/_layouts/base.html
kurtw: The `page` is the page object.

sebworks: No, Im just wondering if we can get rid of the if statement and just have the for loop?

kurtw: Maybe I confused your question. Every time a page is served via wagtail, it will have a `page` object.

kurtw:
This looks good to me :+1:

cfgov/jinja2/v1/_layouts/base.html
sebworks: The code is in `base.html` so it will run on non wagtail pages as well.

kurtw: Oh I see. I don't believe they do. Context is defined [here](https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/sheerlike/__init__.py#L90) and [here](https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/v1/__init__.py#L23) so as long as they don't include a `page` context processor or variable then we should be okay to have this conditional.

sebworks: cool, I would just test this on a Browse page because that is a Wagtail page with no `page.media` object.

kave: @richaagarwal ah yes, your right. i'll add it back in

kave:
@schaferjh do they all have different icons?
screen shot 2016-03-10 at 10 31 30 am

PR #1562, opened 2016-03-09
post preview snapshot is now fixed.

kave:
## Changes

- Post preview snapshot is rendering correctly.

## Testing
- Import Refresh Db
- visit page `policy-compliance/amicus/` to verify 500 error is gone and post preview items render correctly

## Review

- @kurtw 
- @richaagarwal 

## Screenshots
screen shot 2016-03-09 at 4 54 18 pm

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/post-preview.html
+++ b/cfgov/jinja2/v1/_includes/organisms/post-preview.html
@@ -44,7 +44,7 @@
 {% from 'macros/util/format/url.html' import location_image_url as location_image_url %}
 
 {% macro render(post, type='', form_id=0, url='') %}
-    {% set url = url or request.path %}
+    {% set url = url or get_page_state_url({'request':request}, post) %}
     <article class="o-post-preview">
         <div class="meta-header">
             <span class="date meta-header_right">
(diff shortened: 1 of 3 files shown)


schaferjh:
Following up on something I mentioned in the task: We need to be able to select *which* child filterable browse page the post preview snapshot should pull from. Is allowing content editors to make that selection part of this? 

(The use case is for adding it to the Notice and Comment sublanding page which has 2 children filtrable browse pages)

kave:
@schaferjh ah i missed that line... hmm.. i'll have to think about how we can allow this

kurtw:
@schaferjh right now the snapshot looks at all the child browse filterable pages, sorts them by the date_published field, then renders the number that was specified.

kurtw:
@kave [here](https://github.com/cfpb/cfgov-refresh/pull/1562/files#diff-2ba27f15d5ce948b3f9483f62581c805R66) is where self.get_descendants() grabs _any_ page so that should sort out the shared/live thing. We can make this change in another PR as well.

schaferjh:
@kurtw ok, that should work for now. Looking forward to seeing it in action!

kave:
ok great, i'll make the shared/live fix in another PR so content can atleast render the pages for now.

PR #1569, opened 2016-03-10
Fixing Archive Events Script

kave:
Event archive pages now have a specific form & model, so updating script to account for this

- @kurtw


kurtw:
:+1:

PR #1566, opened 2016-03-10
Fixes/Test for External Site Redirect

sebworks:
Fixes/Test for External Site Redirect

## Additions

- Added Unit test for External Site Redirect.

## Removals

- Removed the need for jQuery.

## Changes

- Changed the regex to fix an issue where it wasn't catching query-string params. 

## Testing

- `Visit http://localhost:8000/` and click links in the footer. All non gov links should be redirected.

## Review

- @jimmynotjim 
- @KimberlyMunoz 
- @anselmbradford 

## Notes

- This assumes there is only one level nested spans within an anchor element. 
- The redirect test is very expensive to run on every click. In a beautiful world this wouldn't exist.

## Checklist

* [x] Changes are limited to a single goal (no scope creep)
* [x] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [x] Passes all existing automated tests
* [x] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #1572, opened 2016-03-11
Mega Menu: DFJR-821 Adds full menu/desktop hover state

anselmbradford:
## Additions

- Added Tree and tree traversal algorithms (includes https://github.com/cfpb/cfgov-refresh/pull/1542, with code review revisions.)

## Changes

- Added all launch-state mega menu links.
- Added hover-to-show behavior in desktop mega menu. Fixes DFJR 821.
- Renames transition `flush` method to `remove` and makes it so it removes all transition classes (including the base transition class).
- Adds mouseover the trigger event to the flyout menu.
- Adds `isAnimating` method to the flyout menu.

## Testing

- `gulp build`
- Hovery hover and clickety click on the mege menu items at desktop sizes.
- Resize to mobile, repeat.
- Resize to desktop, repeat.
- Clickety click the search too.

## Review

- @KimberlyMunoz 
- @sebworks 
- @jimmynotjim 

## Screenshots

mobilemegamenu

desktopmegamenu

## Notes

- Includes https://github.com/cfpb/cfgov-refresh/pull/1542, with code review revisions.

## Todos

- Keyboard navigation will be done last, so don't expect it to work.
- The selected state of the menu items at desktop size needs to be worked out. It should be a solid black bar when in a section, but a grey bar when hovering in a section.

### DIFF
--- a/cfgov/unprocessed/css/organisms/mega-menu.less
+++ b/cfgov/unprocessed/css/organisms/mega-menu.less
@@ -125,6 +125,8 @@
 */
 
 .o-mega-menu {
+    position: relative;
+    width: 100%;
 
     &_content-item {
         // Override margin added by CF to list items.
@@ -217,15 +219,10 @@
             border-bottom: 1px solid @gray-50;
             pointer-events: auto;
 
-            transform: translateX( -100% );
-            transition: transform 0.25s ease-out;
-
             .u-drop-shadow-after();
 
             &[aria-expanded="true"] {
                 display: block;
-
-                transform: translateX( 0 );
             }
 
             &-alt-trigger {
@@ -356,12 +353,30 @@
                 }
             }
         }
+
+        // 3rd-level menu - Tablet/Mobile.
+        &_content-3 {
+
+            // TODO: There is similar code set on all three menus.
+            //       Investigate if it can be combined.
+            // Remove final border line of last list item.
+            &-list:last-child {
+                .o-mega-menu_content-3-item:last-child {
+                    border-bottom: none;
+                }
+            }
+        }
+
+        // Submenus - Tablet/Mobile.
+        // TODO: Enable when third-level behaviors are in.
+        //&_content-2,
+        //&_content-3 {
+        //    .u-move-right();
+        //}
     } );
 
     // Desktop sizes.
     .respond-to-min( @bp-med-min, {
-        margin-right: -@grid_gutter-width / 2;
-        margin-left: -@grid_gutter-width / 2;
 
         &_trigger {
             display: none;
@@ -381,11 +396,11 @@
 
             &-item {
                 display: inline-block;
+                margin-right: @grid_gutter-width / 2;
             }
 
             &-link {
                 padding-bottom: @grid_gutter-width;
-                margin-right: @grid_gutter-width / 2;
                 margin-left: @grid_gutter-width / 2;
 
                 font-size: unit( 18px / @base-font-size-px, em );
@@ -396,11 +411,15 @@
 
                 // -6px padding offset is to account for the border height.
                 &[aria-expanded="true"],
-                &__current,
-                // TODO: Show menu on hover and remove :hover selector.
-                &:hover {
+                &__current {
                     padding-bottom: @grid_gutter-width - 6px;
                     border-bottom: 6px solid @black;
+                    cursor: pointer;
+                }
+
+                &:hover {
+                    padding-bottom: @grid_gutter-width - 6px;
+                    border-bottom: 6px solid @gray-50;
                     cursor: pointer;
                 }
             }
@@ -423,8 +442,6 @@
 
                 .o-mega-menu_content-2-wrapper {
                     display: block;
-
-                    transform: translateY( 0 );
                 }
             }
 
@@ -433,8 +450,6 @@
                 background-color: @gray-5;
                 border-top: 1px solid @gray-50;
                 border-bottom: 1px solid @gray-50;
-                transform: translateY( -100% );
-                transition: transform 0.25s ease-out;
 
                 .u-drop-shadow-after();
 
@@ -542,6 +557,13 @@
         position: absolute;
         top: -9999px;
         left: -9999px;
+    }
+}
+
+.no-js .o-mega-menu {
+    // JS isn't available to remove u-hidden class, so override it.
+    &.u-hidden {
+      display: block;
     }
 }
 
--- a/cfgov/unprocessed/js/modules/FlyoutMenu.js
+++ b/cfgov/unprocessed/js/modules/FlyoutMenu.js
@@ -56,22 +56,31 @@ function FlyoutMenu( element ) { // eslint-disable-line max-statements, no-inlin
   var _expandTransitionMethod;
   var _collapseTransitionMethod;
 
+  // Binded events.
+  var _collapseBinded = collapse.bind( this );
   // Needed to add and remove events to transitions.
-  var _expandEndBinded = _expandEnd.bind( this );
   var _collapseEndBinded = _collapseEnd.bind( this );
+  var _expandEndBinded = _expandEnd.bind( this );
+
+  // If this menu appears in a data source,
+  // this can be used to store the source.
+  // Examples include the index in an Array,
+  // a key in an Hash, or a node in a Tree.
+  var _data;
 
   // Set this function to a queued collapse function,
   // which is called if collapse is called while
   // expand is animating.
   var _deferFunct = standardType.noopFunct;
-  var _collapseBinded = collapse.bind( this );
 
   /**
    * @returns {FlyoutMenu} An instance.
    */
   function init() {
     var triggerClickedBinded = _triggerClicked.bind( this );
+    var triggerOverBinded = _triggerOver.bind( this );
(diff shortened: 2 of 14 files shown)


sebworks:
There are times, when hovering on the sub-menu, the nav just closes.

sebworks:
After opening the nav on mobile and resizing to desktop, I got the following:

screen shot 2016-03-11 at 5 23 15 am

cfgov/unprocessed/js/modules/FlyoutMenu.js
```diff
@@ -243,11 +253,19 @@ function FlyoutMenu( element ) { // eslint-disable-line max-statements, no-inlin
...
+      container:  _dom,
       content:    _contentDom,
       trigger:    _triggerDom
     };
   }
 
+  /**
+   * @returns {boolean} True if menu is expanded, false otherwise.
+   */
+  function isAnimating() {
```
sebworks: Consider using a utility `Object.defineProperty` function for places where you want public getters to access private data. The benefit is a property check instead of a function call and write/read control.

https://github.com/cfpb/cfgov-refresh/pull/1566/files#diff-7a844d22219d7d3db1fa7c1e70d7ba45R35

Shouldn't be addressed in this PR.

cfgov/unprocessed/js/modules/FlyoutMenu.js
```diff
@@ -71,7 +71,9 @@ function FlyoutMenu( element ) { // eslint-disable-line max-statements, no-inlin
    */
   function init() {
     var triggerClickedBinded = _triggerClicked.bind( this );
+    var triggerOverBinded = _triggerOver.bind( this );
```
sebworks: I don't think there is a need for the `Binded` suffix. I think it is implied that `this` has the same scope as other functions within the `FlyoutMenu` function.

cfgov/unprocessed/js/modules/FlyoutMenu.js
```diff
@@ -261,10 +279,16 @@ function FlyoutMenu( element ) { // eslint-disable-line max-statements, no-inlin
   this.setCollapseTransition = setCollapseTransition;
   this.getTransition = getTransition;
   this.getDom = getDom;
+  this.isAnimating = isAnimating;
 
   // Public static properties.
   FlyoutMenu.EXPAND_TYPE = 'expand';
```
sebworks: Consider making `expand` and `collapse` constants as well. Might not make sense though if there aren't used within the file.

cfgov/unprocessed/js/modules/FlyoutMenu.js
anselmbradford: Added `TODO` in https://github.com/cfpb/cfgov-refresh/commit/3ee747cfba82bc15ac1040a50a4cab62363da076

anselmbradford:
> There are times, when hovering on the sub-menu, the nav just closes. (on desktop)

Do you see consistency with when this happens?

Does it occur when hovering over the menu item (such as "About Us") and quickly moving downward? I think what's happening is the menu is triggered to be shown and then while it is sliding downward it's possible to move the  cursor below the animating menu, which triggers an event of moving off the menu, causing it to disappear.

bug


> After opening the nav on mobile and resizing to desktop, I got the following:

Ah, yup, seeing this. I'll look into it..

cfgov/unprocessed/js/modules/FlyoutMenu.js
anselmbradford: Yeah the constant is `EXPAND_TYPE` and 'expand' is the value, so references to 'expand' are intended to use `EXPAND_TYPE`. This is a type-safe helper for `getTransition( FlyoutMenu.EXPAND_TYPE )`.

cfgov/unprocessed/js/modules/FlyoutMenu.js
anselmbradford: Removed them in https://github.com/cfpb/cfgov-refresh/commit/7a1bd7ea7db4e9117f8055b6096b38767ee8b8f0 and just set the binding in the call. The variable is only needed where there are repeated calls to the binded function, or if it needs to be removed.

cfgov/unprocessed/css/organisms/mega-menu.less
```diff
@@ -545,6 +560,13 @@
     }
 }
 
+.no-js .o-mega-menu {
```
sebworks: might make for a nice utility class

sebworks:
> Do you see consistency with when this happens?

No, it happens sporadically. Let me see if I can reproduce it consistently.

cfgov/unprocessed/js/modules/FlyoutMenu.js
anselmbradford: Oops, realized why I did this, so there's a call to the binded `_triggerClicked` in both the trigger and alternative trigger. The `Binded` suffix is for cases like https://github.com/cfpb/cfgov-refresh/pull/1572/files#diff-982f166b46c3d02b0eb18e05e6f3e66fR60, where the binded function needs to be shared across methods internally but can't have the same name as the function its binding. Doesn't need to be a `Binded` suffix, but it needs to be some difference in the name.

anselmbradford:
> After opening the nav on mobile and resizing to desktop, I got the following:

Fixed this issue in https://github.com/cfpb/cfgov-refresh/commit/592584a29617e99669107377ba62d91891122ab1

I realized if the mega-menu is set to relative, the width of the drop-down menu would be constrained to the wrapper width, which shouldn't be the case. The drop-down shouldn't rely on the settings outside the menu. I re-arranged the HTML to only wrap the logo/search/cta in the `wrapper`, as it has its own internal wrappers inside the menu lists.

sebworks:
:+1:

sebworks:
> Does it occur when hovering over the menu item (such as "About Us") and quickly moving downward? I think what's happening is the menu is triggered to be shown and then while it is sliding downward it's possible to move the cursor below the animating menu, which triggers an event of moving off the menu, causing it to disappear.

That doesn't cause it but am I'm not able to reproduce it consistently. I say address at a later time.

anselmbradford:
@sebworks Okay, sounds good.
@jimmynotjim @KimberlyMunoz Let me know if you see any other issue outside of the menu disappearing sporadically and keyboard navigation not being ready.

jimmynotjim:
Since we have access to `isAnimating`, can we stop clicks from closing the menu while it's animating?

jimmynotjim:
While hovering, there's a flash between black and gray in the 5px underline/border.

jimmynotjim:
If you hover, then quickly move out and hover again, like your mouse path wasn't exactly straight, the menu stays closed w/ a gray border.

jimmynotjim:
There are a lot of times I move my mouse into the menu, and it closes on me

jimmynotjim:
Example of both of these behaviors.

http://cl.ly/fPNU

anselmbradford:
@sebworks @jimmynotjim I found the random close issue and pushed a fix in https://github.com/cfpb/cfgov-refresh/commit/8d5542094f171aff10e919ac1beb175fbca99b0d, please check!

jimmynotjim:
The scroll-to-top on open/close should be limited to smaller devices, it causes the screen to jump around far too much on desktop sized screens.

http://cl.ly/fPwu

anselmbradford:
> The scroll-to-top on open/close should be limited to smaller devices, it causes the screen to jump around far too much on desktop sized screens.

This is because `focus()` is set on the first item in the menu when it's expanded. I turned that off in https://github.com/cfpb/cfgov-refresh/commit/58bcdbaf472ebd2bea9dab47dddd84e3af38c16b till keyboard navigation is tackled.

sebworks:
Haven't encountered the closing issue but still testing.

cfgov/unprocessed/css/organisms/mega-menu.less
anselmbradford: `no-js` you mean? I think we're tied to the name because it's what modernizr swaps out.

sebworks: no, I mean have classes that `show/hide` depending on if `no-js` is included on the body.

KimberlyMunoz:
Started testing in older browsers, every menu is open in IE8. A CSS rule to hide all those on lt-ie9 should take care of them.
screen shot 2016-03-15 at 12 29 20 pm

KimberlyMunoz:
Opposite problem on IE9, there's no menu at all. 
image

KimberlyMunoz:
Found another weird case, this is on Chrome and IE10.

Steps to replicate:
1. Hover over `Policy and Compliance` to open that flyout menu
2. Move your mouse to `About Us` and open that flyout menu without leaving the menu and flyout menu area (not hard to do since they are next to each other)
3. Move your mouse out of the menu area to close the `About Us` flyour menu.
4. The Policy and Compliance menu remains and is impossible to dismiss.

cfgov/unprocessed/css/organisms/mega-menu.less
anselmbradford: Ah ok, different from https://github.com/cfpb/cf-core/blob/gh-pages/src/cf-utilities.less#L34-L38 ?

PR #1573, opened 2016-03-11
Adds icon to download icons

### LINKED ISSUE #1390: download icon for pdfs missing from document detail template

KimberlyMunoz:
Any link with pdf, doc, docx, xls, xlsx, or zip will get the icon. Fixes #1390 

## Additions

- Adds a download icon to certain links

## Testing

- http://localhost:8000/blog/struggling-to-keep-up-with-student-loan-repayment-2/
- http://localhost:8000/sub-pages/privacy-impact-assessments-pias/

## Review

- @anselmbradford 
- @sebworks 
- @jimmynotjim 

## Checklist

### DIFF
--- a/cfgov/unprocessed/css/misc.less
+++ b/cfgov/unprocessed/css/misc.less
@@ -726,6 +726,29 @@
     opacity: 0;
 }
 
+/* topdoc
+  name: Download icons
+  family: cfgov-misc
+  notes:
+    - "Adds download icons to certain links"
+  tags:
+    - cfgov-misc
+*/
+
+
+a[href$='.pdf']:after,
+a[href$='.doc']:after,
+a[href$='.docx']:after,
+a[href$='.xls']:after,
+a[href$='.xlsx']:after,
+a[href$='.csv']:after,
+a[href$='.zip']:after {
+    .cf-icon();
+    display: inline-block;
+    content: @cf-icon-download;
+    width: 1em;
+    text-align: right;
+}
 
 /* topdoc
   name: EOF
--- a/cfgov/unprocessed/css/pages/sub-pages.less
+++ b/cfgov/unprocessed/css/pages/sub-pages.less
@@ -3,17 +3,6 @@
    Sub-Pages pages at /sub-pages/*.
    ========================================================================== */
 
-.sub-page {
-    a[href $='.pdf']:after {
-        .cf-icon();
-        .u-inline-block();
-        width: 1em;
-        content: @cf-icon-download;
-        color: @pacific;
-        text-align: right;
-    }
-}
-
 .sub-page__grandchild,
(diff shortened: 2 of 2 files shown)


sebworks:
Will this overwrite the existing icons on pages like `/budget/strategic-plan/`?

schaferjh:
A good question, but I'll note that `/budget/strategic-plan/`is one of the ones our content folks have recently moved into Wagtail. :)

cfgov/unprocessed/css/misc.less
```diff
@@ -726,6 +726,29 @@
...
+
+a[href$='.pdf']:after,
+a[href$='.doc']:after,
+a[href$='.docx']:after,
+a[href$='.xls']:after,
+a[href$='.xlsx']:after,
+a[href$='.csv']:after,
+a[href$='.zip']:after {
+    .cf-icon();
+    .u-inline-block();
```
anselmbradford: At this point this could be straight up `display: inline-block`. The utility class just adds a hack for IE7 and below https://github.com/cfpb/cf-core/blob/gh-pages/src/cf-utilities.less#L113

anselmbradford:
Works! :+1: pending Steven's question.

KimberlyMunoz:
Anything with a `icon-link__pdf` like the strategic plan page or list of download links is unaffected, especially on mobile styles, because this selector is as low-level as it gets.

sebworks:
:+1:

anselmbradford:
> Anything with a icon-link__pdf like the strategic plan page or list of download links is unaffected

It looks like `icon-link__pdf` _is_ being overwritten, which doesn't sound like what you want it to do?

E.g.
screen shot 2016-03-11 at 11 12 32 am

…becomes…

screen shot 2016-03-11 at 11 12 28 am

Seems to be because of https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#directly-targeted-elements

jimmynotjim:
> It looks like icon-link__pdf is being overwritten

It's supposed to be the download icon now, the pdf one was removed from the Manual a while back, so I'd actually say this is an ok behavior

schaferjh:
Agree with @jimmynotjim that's the right behavior. It actually will bring those older pages that do remain after the launch in line with our newer ones.

KimberlyMunoz:
I took a look at all the cf-icons that this could potentially overwrite and other than the PDF icon, which I visually thought was the same as this download icon, there probably isn't a use case where we wouldn't want the download icon.

anselmbradford:
:+1: Sounds like `icon-link__pdf` is obsolete now. We'll want to remove that, but doesn't have to happen here.

sebworks:
:

jimmynotjim:
:+1:

PR #1584, opened 2016-03-14
Fix paginator bug

kurtw:
@jimmynotjim


jimmynotjim:
This fixes the 500 error, but now the filtering message is missing. I think the issue is really that posts.paginator doesn't exist.

kurtw:
Should be ready

kave:
:+1: pow

jimmynotjim:
Working, thanks. I'll address the spacing issue in another PR

PR #1585, opened 2016-03-14
Fix filters notifications

richaagarwal:
Do not show a filters notification if there were no filters specified by the user. 

## Additions

- If statement to do this

## Removals

-

## Changes

-

## Testing

- Create a Filterable type page and a child page, and ensure that when no filters are set it does not display any notification, and when there is a filter, it does. 

## Review

- @kave @kurtw 

## Notes

- This doesn't work when the user has "cleared" their filters and applied them again.  

## Todos

- 

## Checklist


cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
```diff
@@ -200,7 +200,9 @@
             {% endfor %}
         {% endif %}
         {% if 'filterable-list' in controls.form_type and posts is defined %}
-            {% if posts.paginator.count == 0 %}
+            {% if not request.GET %}
```
sebworks: Shouldn't if be `{% if request.GET %}`?

richaagarwal: @sebworks request.GET is empty if there are no filters specified, which is the condition we want to be handling here. To `notification`, we pass in `false` as one of the params, so that it does not get displayed.

sebworks: ok, should we pass in a blank message since you will never see that notification unless an error occurs?

sebworks: Nevermind, I think it probably helps with the accessibility of the filters..

kurtw:
Cool :+1:

PR #1586, opened 2016-03-14
Adding phone number to global eyebrow

KimberlyMunoz:
Adds the phone number to the global eyebrow :eyes: 

## Testing

- Test on any page both desktop/mobile

## Review

- @anselmbradford 
- @jimmynotjim 
- @sebworks 

## Screenshots

image

image

## Checklist

### DIFF
--- a/cfgov/unprocessed/css/molecules/global-eyebrow.less
+++ b/cfgov/unprocessed/css/molecules/global-eyebrow.less
@@ -65,7 +65,7 @@
         }
     }
 
-    &_right {
+    &_actions {
         text-align: right;
     }
 
@@ -117,7 +117,7 @@
             padding-bottom: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
         }
 
-        .m-global-eyebrow_right {
+        .m-global-eyebrow_actions {
             padding-top: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
             border-top: 1px solid @gray-40;
             text-align: left;
--- a/cfgov/unprocessed/css/molecules/global-eyebrow.less
+++ b/cfgov/unprocessed/css/molecules/global-eyebrow.less
@@ -65,7 +65,12 @@
         }
     }
 
+    &_right {
+        text-align: right;
+    }
+
     &_languages {
+        display: inline;
         list-style-type: none;
 
         li {
@@ -81,6 +86,10 @@
         a {
             .u-link__no-border();
         }
+    }
+
+    &_phone {
+        padding-left: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
     }
 
     &__horizontal {
@@ -101,7 +110,6 @@
         padding-right: unit( ( @grid_gutter-width / 4 ) * 3 / @base-font-size-px, em );
         padding-top: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
         padding-bottom: unit( @grid_gutter-width / @base-font-size-px, em );
-        border-top: 1px solid @gray-40;
 
         .m-global-eyebrow_tagline {
             background-position-y: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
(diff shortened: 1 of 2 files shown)


cfgov/unprocessed/css/molecules/global-eyebrow.less
```diff
@@ -109,10 +117,16 @@
             padding-bottom: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
         }
 
-        .m-global-eyebrow_languages {
-            padding-left: 37px; //22px flag + 15px spacing
+        .m-global-eyebrow_right {
```
anselmbradford: Probably should have a different name like `&_body` or something since it moves to the bottom at mobile sizes.

anselmbradford:
It looks like the languages and number could both be set to `float: right` to push them to the right and eliminate the necessity of the wrapper div, but perhaps that's more of a hassle to handle clearing the float. Otherwise minor comment of the name that changes layout :+1:

cfgov/unprocessed/css/molecules/global-eyebrow.less
KimberlyMunoz: Sure, I was stuck on a good name. Would `body` work even if the tagline isn't in there?

anselmbradford: Only other ones I can think of are `&_inset` or `&_wrapper`, since it's inside a level from `global-eyebrow`. Or `&_actions`, since it's the functional part of the global eyebrow.

sebworks:
:100:

cfgov/unprocessed/css/molecules/global-eyebrow.less
```diff
@@ -109,10 +117,16 @@
...
             padding-top: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
             border-top: 1px solid @gray-40;
+            text-align: left;
+        }
+
+        .m-global-eyebrow_phone,
+        .m-global-eyebrow_languages {
+          display: block;
+          padding-left: 37px; //22px flag + 15px spacing
```
anselmbradford: `padding-left` could be added to the wrapper around these two it looks like, instead of adding it to them individually.

schaferjh:
Question: If a user is on a calling-capable device, will they be able to tap the number to call? 

(This behavior is a nice to have -- the answer for V1 can be no!)

cfgov/unprocessed/css/molecules/global-eyebrow.less
KimberlyMunoz: I like `&_actions`. Thanks, I'll change that.

KimberlyMunoz:
>If a user is on a calling-capable device, will they be able to tap the number to call?

No, it should be simple to add in v2. We just would have to talk through how it would look and expected interactions on desktop and mobile.

schaferjh:
great, thanks.

cfgov/unprocessed/css/molecules/global-eyebrow.less
KimberlyMunoz: Both the `_phone` and `_languages` have left padding that needs to be overwritten. Including it here handles that and adds the spacing.

jimmynotjim:
@KimberlyMunoz are you sure about the numbers? The ones on the `/contact-us` are automatically dial-able on both Chrome and Safari on iOS. I don't see why it'd be different here.

KimberlyMunoz:
I'm open to making them clickable. I just would want to walk through how they'll look if we do that and make sure design and Jess are aware so we don't switch back and forth.

jimmynotjim:
> In Safari on iOS, telephone number detection is on by default

_https://developer.apple.com/library/ios/featuredarticles/iPhoneURLScheme_Reference/PhoneLinks/PhoneLinks.html_

They're auto-formating text that looks like a number, so this will be clickable on iOS devices. Pretty sure Android does the same. I had thought we were avoiding writing tel links because we wanted to let devices deal with them as needed, but looking at the source on contact-us, we did make them tel links.

KimberlyMunoz:
Yep, that's what I meant, that they aren't tel links right now. Since iOS devices are going to do what they do, we can look at making tel link behavior a bit more standardized in v2 when we have a little more breathing room.

jimmynotjim:
Ok, I was getting mixed up by the reply to @schaferjh's question then. Jess, yes, if you're on a device that's also a phone, the number will be clickable.

schaferjh:
Cool, thanks for the clarification @jimmynotjim

jimmynotjim:
Whoops, forgot to :+1: Good to merge

PR #1595, opened 2016-03-16
Fix sheerlike date functions bug for timezones

kurtw:
Dates were being output with the UTC timezone instead of the appropriate timezone (mainly Eastern). This fixes the functions that are used to convert the given timestamps into the correct timezones.

## Additions

- helper function for DRYness

## Changes

- The functions now use the same logic but return different types and their names reflect that (str vs date obj)

## Fixes

- Sheerlike tests for the templates are now being run 

## Testing

- Run `tox`
- Check out a list page with dates. Really any date being outputted on the site.
- Check out a browsefilterablepage and filter using dates. That should also work.

## Review

- @kave 
- @richaagarwal 

## Screenshots
screen shot 2016-03-16 at 2 31 59 pm

## Notes

- The other sheerlike tests need to be updated too, but that's out of scope for this PR
- Maybe more tests. See if we have time later or post launch

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/macros/time.html
+++ b/cfgov/jinja2/v1/_includes/macros/time.html
@@ -21,7 +21,7 @@
 
 {% macro render(datetime,
                 show_only={'date':true, 'time':true, 'timezone':true},
-                timezone='America/New_York') %}
+                timezone=none) %}
 <span class="datetime">
     {% if show_only.date == true  %}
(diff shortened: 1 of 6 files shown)


richaagarwal:
:+1: other than this part "Check out a browsefilterablepage and filter using dates. That should also work" not working

kurtw:
Yeah that shouldn't be a step because i had the code to fix it locally but didn't include it because it was out of scope. Sorry for the confusion!

PR #1599, opened 2016-03-16
Fix filterable list page set retrieval

kurtw:
A new manager for the pages was necessary to get the `date_published` from the queryset. For whatever reason, even though the objects within the queryset had `date_published` the queryset itself was not filterable on that field. Go figure. 

@richaagarwal 
@kave


richaagarwal:
:+1:

PR #1603, opened 2016-03-17
Mega Menu: Suspend 3rd level menu at desktop

anselmbradford:
## Changes

- Updates incorrect URL and replaces `and` with `&` in menu item text.
- Adds method to suspend/resume trigger events on a flyout menu.
- Suspends trigger events on the 3rd-level trigger at the desktop menu size so that the triggers (e.g. About Us > Careers) won't prevent the default link behavior.

## Testing

- On refresh clicking About Us > Careers just closes the mega menu. Run `gulp build` and see that going from About Us > Careers goes to the careers overview page.

## Review

- @jimmynotjim 
- @KimberlyMunoz 
- @sebworks


KimberlyMunoz:
:+1:

sebworks:
:+1:

PR #1607, opened 2016-03-17
category name change

kave:
## Changes

- category "Administrative Adjudication Process" to "Administrative Filing"

## Testing

- visit page `http://content.localhost:8000/policy-compliance/enforcement/actions/` and verify category is now named "Administrative Filing"

## Review

- @kurtw 
- @richaagarwal 

## Screenshots
Was:
screen shot 2016-03-17 at 1 29 51 pm

now:
screen shot 2016-03-17 at 1 29 23 pm


richaagarwal:
:+1:

cfgov/v1/models/ref.py
```diff
@@ -47,7 +47,7 @@
     )),
     ('Enforcement action', (
         ('fed-district-case', 'Federal District Court Case'),
-        ('admin-adj-process', 'Administrative Adjudication Process'),
+        ('admin-adj-process', 'Administrative Filing'),
```
kurtw: change `admin-adj-process` to `admin-filing`. Just checked the database and there are no pages with this category so we should be good to merge without a data migration.

kave: cool done

PR #1608, opened 2016-03-17
An approach to handling External links in richtext fields

kave:
Benefits of this approach:
- strips out the extra div created by richtext fields resolving css issues caused by this
```
<div class="rich-text">
```
- parses the richt text content and adds a class to any links found to be external.

## Additions

- Global variable to allow changing of the external link css name
- external link filter for richtext fields

## Testing

- Create a Text introduction, and body field add an internal link and add an external link then Preview page

## Review

- @kurtw 
- @richaagarwal 
- @sebworks 

## Screenshots
screen shot 2016-03-17 at 2 11 53 pm

screen shot 2016-03-17 at 4 25 44 pm


cfgov/v1/__init__.py
```diff
@@ -35,12 +39,30 @@ def environment(**options):
...
 
+EXTERNAL_LINK_PATTERN = '(https?:\/\/(?:www\.)?(?![^\?]*(cfpb|consumerfinance).gov)(?!(content\.)?localhost).*)'
+
+
+def external_links(value):
+    soup = BeautifulSoup(expand_db_html(value.source))
+
+    try:
+        for a in soup('a'):
+            p = re.compile(EXTERNAL_LINK_PATTERN)
```
richaagarwal: this re.compile line should be outside of the for loop, ya?

richaagarwal: also not sure if we need to do this in a reg ex, thoughts?

kave: whoops yea i'll move that out of the loop. I thought a regex was the simplest way to handle anything outside of `cfpb.gov`, `consumerfinance.gov`, `localhost` and `content`.

i'm open to suggestions though?

richaagarwal: @kave I don't actually know how much less performant a reg ex is, but since this is being called a lot, I thought it might be more performant to do an if statement, like so:

`if 'cfpb.gov' or 'consumerfinance.gov' in a['href']:` 

which should take care of 3 of those cases. we could throw in localhost in there too, or set up an environment variable. I realize that sounds more complex code-wise, so if it doesn't sound like it'd make that much of a difference performance-wise, I think this is fine!

kave: yea so the regex also covers cases if anyone added a link like `https://www.facebook.com/?redirect=http%3A//test.cfpb.gov/the-bureau` with the conditional that would be considered a valid internal. 
i think for now lets leave it in, and post V1 we can look into a more performant to handle all the cases?

richaagarwal: @kave yeah I thought about that but CFPB is controlling the content, so it didn't seem like that'd be an issue. anyway, yes, that sounds good!  :+1:

richaagarwal:
:+1:

cfgov/v1/__init__.py
```diff
@@ -35,12 +39,32 @@ def environment(**options):
         'choices_for_page_type': ref.choices_for_page_type,
         'is_blog': ref.is_blog,
         'get_page_state_url': share.get_page_state_url,
+        'cfgov_rich_text': external_links,
```
sebworks: This could be very confusing; we are naming the function `cfgov_rich_text ` but it's actually converting external links.

sebworks:
Short term this is fine but we shouldn't be parsing whole text blocks on the server at runtime (It's extremely expensive to do so) .We need to modify the links when they are created or control how they are created. Wagtail is using https://github.com/bergie/hallo for their [text editor](https://github.com/thesgc/shergar/blob/bdf2d5195195b600535b32a2f26e98d035fc82ab/shergar/assets/wagtailadmin/js/vendor/hallo.js) and it has a mechanism for doing that: https://github.com/bergie/hallo#events.

Lets go with this for now and lets work on implementing that JS solution. Thanks for working on this @kave.

cfgov/v1/__init__.py
kave: true, i've changed the name to `parse_links` now

kave:
@sebworks sounds good to me. i'd rather throw out hallo's minimalist editor in general and replace it with something like http://summernote.org/ in the future

kurtw:
:+1:

PR #1615, opened 2016-03-17
Filters no longer showing notification after clearing form

richaagarwal:
Fixes this bug from Jira: "When I use the filter and then clear it, the notification shows up again. (enter filter parameters, click apply filter, page reload with notification and filtered results, click "clear filters", page reloads showing default list but with "filtered results" notification)

## Changes

- Update `clear-form-buttons.js` to remove the input fields themselves, rather than clear the values.

## Testing

- Go to any filterable type page and select a filter and apply it, and then try clearing the form. This will clear the results as well as the notification. 

## Review

- @kave @kurtw 

## Checklist


sebworks:
We should probably try calling `form.reset()`?

cfgov/unprocessed/js/modules/clear-form-buttons.js
```diff
@@ -17,21 +17,10 @@ function init() {
...
-    // Clear checkboxes
-    $form.find( '[type="checkbox"]' )
-    .removeAttr( 'checked' );
-
-    // Clear radio
-    $form.find( '[type="radio"]' )
-    .removeAttr( 'checked' );
-
-    // Clear select options
-    $form.find( 'select option' )
```
jimmynotjim: We'll need to keep the `select` elem one.

jimmynotjim:
It's working with inputs, but be sure to clear a normal `select` as well.

jimmynotjim:
> We should probably try calling form.reset()?

@sebworks I thought about that when I added the submission, but I figured that this needs to be refactored to remove jQuery anyway after launch, so keep the scope simple till then.

richaagarwal:
@sebworks @jimmynotjim I think reset only works on form elements, not the form itself, so it doesn't seem to gain us much? Also I'm not sure but I suspect it would just clear the value, whereas we need to remove the entire element itself. 

@jimmynotjim  Actually, I just noticed the selects wouldn't need to be cleared since we're somehow clearing them anyway (if you check on flapjack). Not sure if that's it's own bug though...

jimmynotjim:
The selects aren't really selects, they're groups of checkboxes in a dropdown, so you're capturing them with the other inputs. If for some reason the original select wasn't replaced, we'd still want to be able to clear it out.

sebworks:
> I think reset only works on form elements, not the form itself, so it doesn't seem to gain us much? Also I'm not sure but I suspect it would just clear the value, whereas we need to remove the entire element itself.'

You are correct and I misspoke; reset will only work on form elements. Having said that, we shouldn't be removing the form fields on submission.  It won't solve the issue if JS is disabled and if for some reason there is a script error (We are rely uplon a jquery to remove the elements), you are left with a blank form. I suspect some form fields aren't being cleared when looping the fields.

jimmynotjim:
> We should probably just make the reset button a link and make the href the request url. This way, no js would be required.

I like this idea for the time being till we update this w/ AJAX requests.

richaagarwal:
> It won't solve the issue if JS is disabled and if for some reason there is a script error (We are rely upon a jquery to remove the elements), you are left with a blank form. I suspect some form fields aren't being cleared when looping the fields.

@sebworks Is the main issue then that this uses JS?  I'm just curious why that is a concern with this one in particular, since we're using it in other places, and I'm not introducing it but modifying it.  In any case, it sounds like this probably should've been a task assigned to FEWDs :)  Can I hand it off to one of you?

sebworks:
It's just that we are removing form fields via an external library and that library may fail. There are just better options. You can assign it to me.

sebworks:
@richaagarwal, I've opened up a new PR to solve this issue https://github.com/cfpb/cfgov-refresh/pull/1650

PR #1616, opened 2016-03-18
Fixes Tags on related metadata

kave:
## Changes

- removed page reference and calling function from its util location

## Review

- @kurtw 
- @richaagarwal 

## Testing

- visit [here](http://content.localhost:8000/policy-compliance/guidance/implementation-guidance/supervisory-matters-appeal-process/)

- verify clicking tag in the side bar routes you to the appropriate filtered page


kurtw:
:+1:

PR #1622, opened 2016-03-18
Datetimes now render with their entered times with the desired timezone

kurtw:
Basically, timezones are saved as UTC which is usually the best idea but when converting to EST is a problem because data entry is in EST not UTC. This strips the UTC and replaces it with the desired timezone (usually EST) instead of converting from UTC.

@richaagarwal 
@kave

### DIFF
--- a/cfgov/sheerlike/templates.py
+++ b/cfgov/sheerlike/templates.py
@@ -11,10 +11,9 @@ def _convert_date(date, tz):
         if type(date) in [datetime.datetime, datetime.date]:
             if isinstance(date, datetime.datetime):
                 pytzone = timezone(tz)
-                if not date.tzinfo:
-                    date = pytzone.localize(date)
-                else:
-                    date = date.astimezone(pytzone)
+                if date.tzinfo:
+                    date = date.replace(tzinfo=None)
+                date = pytzone.localize(date)
     return date
 
 


cfgov/sheerlike/templates.py
```diff
@@ -11,10 +11,9 @@ def _convert_date(date, tz):
...
             if isinstance(date, datetime.datetime):
                 pytzone = timezone(tz)
-                if not date.tzinfo:
-                    date = pytzone.localize(date)
-                else:
-                    date = date.astimezone(pytzone)
+                if date.tzinfo:
+                    date = date.replace(tzinfo=None)
+                date = pytzone.localize(date)
     return date
```
richaagarwal: I don't know if stripping the date of its timezone info is a good idea.  For example, if you started with a timezone that is in ET, you would end up with the same time but without the ET asociated with it.  Then, let's say you localize it to PT, you're *still* using the same time, but you've simply swapped the timezones from ET to PT without updating the datetime itself.

schaferjh:
Just throwing this in here, not sure how related or not but just in case --

@tnferrell observed the following issue while updating events pages in Wagtail:
"Something's wrong with the time on the datepicker. I put in times that start at 10, but the agenda starts at 6:00am."

kurtw:
Yeah this should fix this.

cfgov/sheerlike/templates.py
kurtw: We are assuming that they are entering the time into Wagtail as EST. I noted that we aren't doing the best thing here but we just don't have time to change all of our dates and date logic right now.

richaagarwal:
Ah, finally got what's going on. Didn't realize we were always saving in UTC.  I know it's already merged, but just wanted to :+1: for posterity.

PR #1631, opened 2016-03-18
We've stopped featuring a topic on the Newsroom

jimmynotjim:
What it says.


KimberlyMunoz:
:+1:

Is the featured topic something that's handled by the CMS and going to be going off-and-on? Because if it's variable we should probably remove that series of tests entirely.

schaferjh:
yes to @KimberlyMunoz's question.

jimmynotjim:
I think leaving it silenced will help direct us when we rethink our testing strategy. Seeing all these silenced tests kinda sells us on why it needs to be redone.

PR #1639, opened 2016-03-19
Updates the mega-menu for no-js and IE8-10 fixes **SEE NOTES**

jimmynotjim:
Updates the mega-menu for no-js and IE 8-10 fixes (Part 2 of Jira issue 816)

## Removals

- Removes need for pointer events

## Changes

- Sets up JS only styles instead of menu styles globally scoped to js

## Testing

- `gulp build` and test the menu at multiple browser styles, with JS on and off, it should look and function the same as before.

## Review

- @anselmbradford 
- @KimberlyMunoz 
- @sebworks 

## Notes

- We've decided to show the top level items as links by default on small screens when JS fails to deliver.
- ~~There are script errors from the BaseTransitions files in IE9. I'm having trouble debugging them, but we should discuss if transitions should even be supported in such an old browser. Progressively enhancing a simple toggle of show/hide is probably more appropriate than trying to add complicated scripting that the browser doesn't seem to like.~~ This was fixed in #1641 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,8 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Wrap prefooter section in Browse pages in a conditional to prevent empty prefooter
 - Frontend: Added behaviors for third level mobile mega menu.
 - Frontend: Made Expandables collapse under 600px window size.
+- Updated the Mega Menu layout to avoid pointer events for older IE.
+- Updated the Mega Menu for devices without JS.
 
 ### Removed
 - Removed normalize and normalize-legacy from main less file because CF already includes it.
--- a/cfgov/jinja2/v1/_includes/organisms/header.html
+++ b/cfgov/jinja2/v1/_includes/organisms/header.html
@@ -26,11 +26,11 @@
                          height="50">
                 </noscript>
             </a>
-        </div>
 
-        {% block primary_nav %}
-            {% include 'organisms/mega-menu.html' with context %}
-        {% endblock primary_nav %}
+            {% block primary_nav %}
+                {% include 'organisms/mega-menu.html' with context %}
+            {% endblock primary_nav %}
+        </div>
 
     </div>
 </header>
--- a/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
+++ b/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
@@ -107,7 +107,8 @@
      data-js-hook="flyout-menu_content">
 
     <div class="{{ _classes( nav_depth, '-wrapper' ) }}">
-        <div class="wrapper">
+        {{ '<div class="wrapper">' | safe if nav_depth > 1 else '' }}
+            {# Open wrapper if needed #}
             {% if nav_depth > 1 %}
             <button class="{{ _classes( nav_depth, '-alt-trigger' ) }}"
                     data-js-hook="flyout-menu_alt-trigger">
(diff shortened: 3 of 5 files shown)


cfgov/unprocessed/js/modules/transition/BaseTransition.js
```diff
@@ -198,9 +203,9 @@ function BaseTransition( element, classes ) { // eslint-disable-line max-stateme
   this.setElement = setElement;
 
   // Public static constants.
-  BaseTransition.BEGIN_EVENT = 'transitionBegin';
-  BaseTransition.END_EVENT = 'transitionEnd';
-  BaseTransition.NO_ANIMATION_CLASS = 'u-no-animation';
+  this.BEGIN_EVENT = BEGIN_EVENT;
+  this.END_EVENT = END_EVENT;
+  this.NO_ANIMATION_CLASS = NO_ANIMATION_CLASS;
```
anselmbradford: I know it's IE :rage4:, but this doesn't look like it's a necessary change. It looks like these can just be moved outside the function definition, so that the end of the file would look like:

```

  return this;
}

// Public static constants.
BaseTransition.BEGIN_EVENT = 'transitionBegin';
BaseTransition.END_EVENT = 'transitionEnd';
BaseTransition.NO_ANIMATION_CLASS = 'u-no-animation';

module.exports = BaseTransition;
```

If you then put this at the top of common.js to test, it works in IE9:

```

var BaseTransition = require( '../modules/transition/BaseTransition' );
console.log( BaseTransition.BEGIN_EVENT );
```

anselmbradford: Confirmed fixed by moving constants outside of the constructor and removing redundant firing of `transitionEnd` --> https://github.com/cfpb/cfgov-refresh/pull/1641

anselmbradford:
Green arrows on the first-level menu at mobile could probably be wrapped in `.js` too show they don't show when its not possible to get to a submenu https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/unprocessed/css/organisms/mega-menu.less#L286-L296

anselmbradford:
I can't get the mobile menu trigger to work anymore. The cursor doesn't change anymore and it's not clickable:
screen shot 2016-03-20 at 2 53 06 pm

jimmynotjim:
> I can't get the mobile menu trigger to work anymore. 

Looks like I missed one of the pointer events. Fix has been pushed

> Green arrows on the first-level menu at mobile could probably be wrapped in .js so they don't show when it's not possible to get to a submenu 

> Also pushed

jimmynotjim:
@anselmbradford  @sebworks  @KimberlyMunoz Ready for another review, be sure to pull latest and `gulp build`

anselmbradford:
> I can't get the mobile menu trigger to work anymore.
> Looks like I missed one of the pointer events. Fix has been pushed

I remembered this one is because the menu covers the header before it slides out. So without the `pointer-event: none` it's not possible to click the search button or the logo at mobile, because it's covered by a transparent part of the menu. This only matters for IE9-10 shrunk to a mobile window size. I'd say add a pointer-events specific style rule that hides the search at mobile when no pointer-events are supported.  Modernizr is currently adding a `no-csspointerevents` class to the page. It's not like it eliminates search completely, as this would only be IE9-10 opened or resized to a small window size. This would also eliminate the IE9 bug of swapping between search and the menu as well I believe.

anselmbradford:
LGTM :+1:

KimberlyMunoz:
I like how this has been simplified for no-js and mobile. :clap: 

:+1:

PR #1641, opened 2016-03-20
FlyoutMenu: No transition support fixes

anselmbradford:
## Changes

- When no transition event is present `transitionEnd` was dispatched twice. This fixes that.
- Move public static constants outside constructor.
- Fix order of addEventListener on transition in FlyoutMenu when there is no transition support.

## Test

- `gulp build` and look at dev console in IE9. There should be no js script error. 

## Review

- @jimmynotjim

### DIFF
--- a/cfgov/unprocessed/js/modules/FlyoutMenu.js
+++ b/cfgov/unprocessed/js/modules/FlyoutMenu.js
@@ -1,6 +1,7 @@
 'use strict';
 
 // Required modules.
+var BaseTransition = require( '../modules/transition/BaseTransition' );
 var dataHook = require( '../modules/util/data-hook' );
 var EventObserver = require( '../modules/util/EventObserver' );
 var standardType = require( '../modules/util/standard-type' );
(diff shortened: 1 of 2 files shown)


sebworks:
:+1:

anselmbradford:
@sebworks found one more issue where if transitions weren't supported the `transitionEnd` event would fire before it was being listened for, so `expandEnd` would never fire. Pushed up https://github.com/cfpb/cfgov-refresh/pull/1641/commits/9d1f7739b799cc27989eff22bea6efb38a5fa7ff that fixes the order. Should fix the desktop menu in IE9 (mobile size on IE9 still has issues).

KimberlyMunoz:
:+1: If #1639 doesn't fix the mobile issues, we should note those in issue.

anselmbradford:
The remaining issue I'm seeing on IE9 is:
- Resize to mobile
- Expand mega menu
- Click to second level menu (e.g. About Us)
- Click search button (mega menu collapses, search expands)
- Click mega menu button (search collapses, mega menu should expand but **ONLY OVERLAY SHOWS**)

sebworks:
Late to the game but LGTM

PR #1643, opened 2016-03-20
DFJR 971 - Collapses expandable on mobile

anselmbradford:
Expandables will collapse when page is loaded under 600px width, or if it is resized below 600px. If resized and collapsed, it won't remember that it was expanded at desktop sizes and won't expand when resizing back to desktop.

## Changes

- Makes Expandables collapse under 600px window size.
- General code comment cleanup.
- Removes `addEventListener` check since IE8 is no-js state now.

## Testing

- `gulp build`
- Create a browse filterable page in wagtail and add a filter.
- Check "Is Expanded" in wagtail admin.
- Load the page preview in a small window, note filter is collapsed.
- Load the page preview in a large window, note filter is expanded.
- Resize the page to a small size and note filter collapses.

## Review

- @kurtw 
- @sebworks 
- @jimmynotjim 
- @KimberlyMunoz 

## Todos

- @kurtw Can we add a note next to the "Is Expanded" checkbox in wagtail that says it won't apply under 600px wide?
- If filters have any user entered data, they need `is_expanded=true` passed to `block.value` in https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/jinja2/v1/browse-filterable/index.html#L34

### DIFF
--- a/cfgov/unprocessed/js/molecules/Expandable.js
+++ b/cfgov/unprocessed/js/molecules/Expandable.js
@@ -1,8 +1,9 @@
 'use strict';
 
 // Required modules.
-var EventObserver = require( '../modules/util/EventObserver' );
 var atomicHelpers = require( '../modules/util/atomic-helpers' );
+var breakpointState = require( '../modules/util/breakpoint-state' );
+var EventObserver = require( '../modules/util/EventObserver' );
 
 /**
  * Expandable
@@ -39,6 +40,8 @@ function Expandable( element ) { // eslint-disable-line max-statements, inline-c
 
   // TODO: Replace function of _that with Function.prototype.bind.
   var _that = this;
+  var _collapseBinded = collapse.bind( this );
+  var _expandBinded = expand.bind( this );
 
   /**
    * @param {number} state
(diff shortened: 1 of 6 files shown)


sebworks:
:+1:

cfgov/unprocessed/js/molecules/Expandable.js
```diff
@@ -47,8 +50,10 @@ function Expandable( element ) { // eslint-disable-line max-statements, inline-c
    */
   function init( state ) {
     _calcHeight();
-    if ( state === EXPANDED ||
-         _dom.getAttribute( 'data-state' ) === 'expanded' ) {
+    // Even if expanded is set, don't expand if in mobile window size.
+    if ( !_isInMobile() &&
+         (state === EXPANDED ||
```
jimmynotjim: Minor, add space after `(`

jimmynotjim:
LGTM

cfgov/unprocessed/js/molecules/Expandable.js
anselmbradford: Fixed!

PR #1645, opened 2016-03-21
Filters stay expanded after submitting 

richaagarwal:
BEWD portion of https://github.com/cfpb/cfgov-refresh/pull/1643

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
+++ b/cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
@@ -181,6 +181,9 @@
 
 {% macro render(controls, form, index) %}
     <div class="o-filterable-list-controls">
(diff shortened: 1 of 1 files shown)


jimmynotjim:
Can you update `/cfgov/jinja2/_includes/macros/filters.html` as well for the old Sheer pages?

anselmbradford:
Checking the request in the expandable will potentially lead to expandables that aren't used in a filter being expanded if they are on the same page as a filter. `is_expanded` needs to be passed into the expandable, such as setting `controls.is_expanded = true` before `controls` is passed in here --> https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html#L185

anselmbradford:
LGTM :100:

PR #1648, opened 2016-03-21
Add disabled overview pages

anselmbradford:
## Changes

- Adds logic to show overview—but disabled—links for Consumer Tools and Education Resources.

## Testing

- Check in desktop and mobile menu Consumer Tools and Educational Resources.

## Review

- @KimberlyMunoz 
- @jimmynotjim 
- @sebworks 

## Screenshots

screen shot 2016-03-20 at 9 41 52 pm

screen shot 2016-03-20 at 9 42 11 pm

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
+++ b/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
@@ -120,10 +120,11 @@
                     {# TODO: Remove the check for '#' when real overview pages
                              are added for the Consumer Resources and Education
                              pages. #}
-                    {% if nav_overview_url and nav_overview_url != '#' %}
-                    <a class="{{ _classes( nav_depth, '-overview-link' ) }}
+                    {% if nav_overview_url %}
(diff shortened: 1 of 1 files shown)


schaferjh:
I'm having doubts on this one. @anselmbradford and I discussed it earlier and I was thinking it would be nice to have the overview label as a placeholder at the mobile scale since it serves as a piece of descriptive information even without a link at that scale. However, seeing it grayed out at desktop size worries me that it will just lead to more confusion than help.

Maybe we think about this a bit more before doing it -- interested to hear other thoughts?

KimberlyMunoz:
I think having it greyed out with the rest being blue links signals "coming soon" which is pretty accurate for what these pages are. I'd want to ask someone who hasn't been on the project for as long what their thoughts are.

KimberlyMunoz:
:+1: on the implementation. Glad to see my TODO basically stays the same in this situation.

jimmynotjim:
:+1: dependent on @schaferjh final decision.

anselmbradford:
Usurped by https://github.com/cfpb/cfgov-refresh/pull/1704

PR #1654, opened 2016-03-21
Fixup the publish all script

kurtw:
The actual page object in the database wasn't being set to live. This should do the trick.

@kave


kave:
:+1:

kave:
lol oops wrong button

PR #1661, opened 2016-03-21
Fix timezones

richaagarwal:
This reverts the code @kurtw introduced in https://github.com/cfpb/cfgov-refresh/pull/1622 (sorry Kurt!) but instead handles the issue at the time the datetime is saved. 

cc @kave @rosskarchner 

UPDATE:
- I added a script to update the existing datetimes in the db. 
- I did not update `ItemIntroduction.date` because we don't have any DemoPages yet. It also doesn't seem necessary to update the `HomePage.latest_updates.date` since it seems like we will be updating that anyway. Otherwise I am updating all Datetimes. Dates are always naive so we don't have to worry about those.

ANOTHER UPDATE:

**Testing instructions:**

- Start with a refresh db. 
- Load up a page like [http://localhost:8000/about-us/events/spring-2016-credit-union-advisory-council-meeting-washington-dc/](http://localhost:8000/about-us/events/spring-2016-credit-union-advisory-council-meeting-washington-dc/) and observe the times there
- Run `python cfgov/manage.py runscript fix_datetimes`. 
- Refresh the above page & make sure these changes propagated - times should be 4 hours ahead in this case. 
- Confirm these times are the same as in [www](http://files.consumerfinance.gov/f/201603_cfpb_meeting-agenda_cuac.pdf)

- Go to [http://localhost:8000/the-bureau/leadership-calendar/](http://localhost:8000/the-bureau/leadership-calendar/) and confirm these times match those in [www](http://www.consumerfinance.gov/leadership-calendar/) (Just pick out one or two)

- Create an event page or edit an event page & enter in a start time and an end time. Preview or publish the page and confirm it shows the same time you input. 
- Go into the shell and confirm the times are stored in UTC (and thus are 4 hours ahead)

- You can look at events before Daylight Savings (before March 13) and confirm those are actually 5 hours ahead of what's being displayed on the site.

Or if you think of other scenarios, feel free to try em instead!


kave:
nice, i like the look of this. i'll test it now.

kurtw:
alt

richaagarwal:
lol I don't understand what that gif means. OK i need to add a script to update the existing wrong dates in the db, and then this will be ready

richaagarwal:
This is ready for review again (added script and updated comment above - please read!)

rosskarchner:
@richaagarwal would we get the benefits of this middleware by setting [US_TZ](https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-USE_TZ) to True and [TIME_ZONE](https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-TIME_ZONE) to 'US/Eastern'?

rosskarchner:
Testing notes:
On current flapjack:created an event 11-12 today, and it was stored 11:00-12:00 (so, wrong, as expected)
On this PR:  time stored was 15:00 - 16:100
Back on flapjack, but tweaking TIME_ZONE:  timezone stored was 15:00-16:00

richaagarwal:
@rosskarchner Are you looking at the database values, or what's displayed?  We want to store things in UTC so I don't think tweaking the time zone would work.

rosskarchner:
Yeah, I'm looking at the database values

rosskarchner:
the effect of setting the TIME_ZONE seems to be, interpret incoming form data as being in the specified TZ, and store as UTC.

richaagarwal:
ah, interesting. I'll take a look and modify my PR. We'd still need to run the script though.

richaagarwal:
Are you setting the TIME_ZONE in settings/base.py?

rosskarchner:
Yeah

rosskarchner:
Could the script use timezone.localize and astimezone to do the conversion? That might avoid the DST problem.

richaagarwal:
@rosskarchner Yeah, good call. Thanks for all the feedback!

kurtw:
Could the fix times be a data migration instead?

kurtw:
Also, why is it only EventPages? There are many other types that have date fields. The issue should affect those too right?

richaagarwal:
@kurtw From above: "I did not update ItemIntroduction.date because we don't have any DemoPages yet. It also doesn't seem necessary to update the HomePage.latest_updates.date since it seems like we will be updating that anyway. Otherwise I am updating all Datetimes. Dates are always naive so we don't have to worry about those." 

A data migration doesn't really make sense since this is a one-off thing. We wouldn't want it to run on data that is correct.

kurtw:
> From above: "I did not update ItemIntroduction.date because we don't have any DemoPages yet. It also doesn't seem necessary to update the HomePage.latest_updates.date since it seems like we will be updating that anyway. Otherwise I am updating all Datetimes. Dates are always naive so we don't have to worry about those."

Whoops.

> A data migration doesn't really make sense since this is a one-off thing.

The data migration should run only once and it happens at that state of the database. It makes more sense to make it a data migration because we're moving data at this particular point in time. Putting it into a script means that we can run it at anytime making it more susceptible to breaking something. **Disclaimer**: I just made a script that should have been a data migration because I didn't feel like debugging what seemed like a bug in django orm's `apps` module.

If you can get it to work as a data migration I'd prefer that but if not it's fine as a script. But honestly, **do whatever** because we're getting rid of half of our scripts and our migration history is going away so it ultimately doesn't matter which one you do, I was just saying as a matter of principle.

kurtw:
> lol I don't understand what that gif means. OK i need to add a script to update the existing wrong dates in the db, and then this will be ready

lol it's an upvote! :arrow_up:

richaagarwal:
Updates:  

- The script now saves a revision too, so the db and page content are in sync. 
- Put the middleware back in. The TIME_ZONE setting doesn't seem to be doing what it's intended to do when said to US/Eastern, so I'm just going to leave it as UTC.  
- I've tested this on my end a bunch, but I've been looking at it for a while too, so fresh eyes would be helpful!  Updated the testing instructions above.

@kave @kurtw @rosskarchner

kurtw:
:smile: :+1:

kave:
looks good to me :+1:

PR #1663, opened 2016-03-21
Updates helper method to remove multiple lookups

anselmbradford:
## Changes

- Updates helper method to remove multiple lookups within block object of helper.
- Updates code comment for missing argument.

## Review

- @kurtw 
- @kave 
- @richaagarwal


richaagarwal:
:+1:

PR #1677, opened 2016-03-22
Fix the way the fcm video expands

KimberlyMunoz:
Fix so close and share buttons don't hover over the video.

## Changes

- Add the `featured-content-module_visual` class to the video player instead of around it so that the `left` attribute can be overwritten when the video is playing

## Testing

- Visit http://localhost:8000/the-bureau/

## Review

- @sebworks 
- @anselmbradford 
- @jimmynotjim 

## Screenshots

image

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/macros/video-player.html
+++ b/cfgov/jinja2/v1/_includes/macros/video-player.html
@@ -55,7 +55,8 @@
     {% set button_pos   = options.button_pos or 'center' %}
     {% set is_flexible  = ( options.video.width or options.video.height ) is not defined %}
     <div class="video-player
-                video-player__{{ video_player }}"
+                video-player__{{ video_player }}
+                {{ 'featured-content-module_visual' if is_flexible == false }}"
          {{ "data-id="~video_id | trim if video_id else '' }}
          data-src="{{ video_url }}"
          data-width="{{ options.video.width if options.video.width else '' }}"
--- a/cfgov/jinja2/v1/_includes/molecules/featured-content.html
+++ b/cfgov/jinja2/v1/_includes/molecules/featured-content.html
@@ -83,10 +83,8 @@
         </ul>
     </div>
     {% if value.video.url and value.video.id %}
-        <div class="featured-content-module_visual">
-            {% import 'macros/video-player.html' as video_player %}
-            {{ video_player.render(value) }}
-        </div>
+        {% import 'macros/video-player.html' as video_player %}
+        {{ video_player.render(value) }}
     {% elif value.image %}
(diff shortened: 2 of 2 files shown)


cfgov/jinja2/v1/_includes/macros/video-player.html
```diff
@@ -55,7 +55,8 @@
     {% set button_pos   = options.button_pos or 'center' %}
     {% set is_flexible  = ( options.video.width or options.video.height ) is not defined %}
     <div class="video-player
-                video-player__{{ video_player }}"
+                video-player__{{ video_player }}
+                {{ 'featured-content-module_visual' if is_flexible == false }}"
```
sebworks: Instead of adding `featured-content-module_visual` how about just using `additional_classes` and passing in `featured-content-module_visual`?

sebworks:
:100:  looks great.

cfgov/jinja2/v1/_includes/macros/video-player.html
KimberlyMunoz: Would it be okay to do that and have video_player take in `options` and `additional_classes`? I couldn't figure out a way to have wagtail return a constant for is_fcm or additional classes so that content don't have to mess with that field.

anselmbradford: I'd just caution that if video-player is on the road to being an organism it shouldn't use `additional_classes`, because then it could arbitrarily change and it can't be guaranteed that its fully covered by documentation, tests, etc.

anselmbradford:
:+1:

cfgov/jinja2/v1/_includes/macros/video-player.html
sebworks: so @anselmbradford what you target it from the parent organism?

anselmbradford: My understanding would be if video-player is an organism and it has margin/float adjustments, those would be set in the template. If it's more complex changes, it would be a modifier, or an entirely different organism. 

The issue is that if the video-player organism can take arbitrary classes, it can morph into a unique snowflake that isn't a video-player organism anymore (as originally authored). That means it can't be guaranteed to be covered in documentation from design or be expected to behave the same way in tests as other video-player instances.

sebworks: ok, I think the way to go is a modifier. @KimberlyMunoz, lets go ahead and merge this and I'll fix the issue with the modifier.

jimmynotjim:
:+1:

PR #1678, opened 2016-03-23
Fix view more link

richaagarwal:
Related posts "view more" link was showing up for each post type. This PR fixes that.

## Changes

- Moved view_more block out of a for loop it didn't need to be in.  It's hard to tell from the diff, but it was previously in the` _related_posts_list` macro, which gets called for each post type. So I moved it out of there and put it at the end of the main `render` block.

## Testing

- Check out [http://localhost:8000/about-us/contact-us/](http://localhost:8000/about-us/contact-us/) and confirm that there's only one "View more" link in the sidebar

## Review

- @kave @kurtw 

## Screenshots

Before: 

screen shot 2016-03-23 at 2 27 15 am

After: 

screen shot 2016-03-23 at 2 27 31 am

## Notes

- I made a comment in Jira, but I wonder if we could use this in places where it seems to be hardcoded, e.g. the `about-us` template. 

## Todos

- Depends on outcome of above

## Checklist


rosskarchner:
Here's a funny thing-- on localhost, I get "see all activities" instead of "view more"-- is that expected?

Aside from that, :+1:

schaferjh:
I believe the text for the link can be set within the CMS -- could that be the source of the difference you see, @rosskarchner?

richaagarwal:
@rosskarchner Are you looking at /about-us or /about-us/contact-us ?  I made a note about how this is hardcoded in /about-us in Jira and maybe we want to think about updating that now that this works.

rosskarchner:
@schaferjh that was it-- I found it:
screen shot 2016-03-23 at 12 24 09 pm

richaagarwal:
@rosskarchner Ah! Different versions of the db, I guess :)

PR #1684, opened 2016-03-23
Abstracted dom events for easier reuse throughout the project

jimmynotjim:
Abstracted dom events for easier reuse throughout the project

## Additions

- Creates public bind and fire events
- Creates private parse event helper for bind
- Adds tests for public bind events

## Testing

- `gulp test:unit:scripts`
- `gulp build` and test the Multiselect on `/blog`. It should function exactly the same

## Review

- @anselmbradford 
- @sebworks 
- @KimberlyMunoz 

## Notes

- Not a lot to gain in coverage, but it'll make testing the multiselect and anything else that uses events easier to test in the long run.

__Before__

```
=============================== Coverage summary ===============================
Statements   : 27.28% ( 731/2680 )
Branches     : 23.34% ( 204/874 )
Functions    : 20.05% ( 86/429 )
Lines        : 27.38% ( 727/2655 )
================================================================================
```

__After__

```
=============================== Coverage summary ===============================
Statements   : 27.78% ( 745/2682 )
Branches     : 23.57% ( 206/874 )
Functions    : 20.98% ( 90/429 )
Lines        : 27.89% ( 741/2657 )
================================================================================
```

## Checklist

### DIFF
--- a/cfgov/unprocessed/js/modules/util/dom-events.js
+++ b/cfgov/unprocessed/js/modules/util/dom-events.js
@@ -1,0 +1,21 @@
+'use strict';
+
+/**
+ * Shortcut for binding event listeners to elements
+ * @param  {HTMLNode} elem   The element to attach the event listener to
+ * @param  {Object}   events The list of events to attach to the element
+ */
+function bindEvent( elem, events ) {
+  var callback;
+
+  for ( var event in events ) {
+    if ( events.hasOwnProperty( event ) ) {
+      callback = events[event];
+      elem.addEventListener( event, callback );
+    }
+  }
+}
+
+module.exports = {
+  bindEvent: bindEvent
+}
--- a/cfgov/unprocessed/js/molecules/Multiselect.js
+++ b/cfgov/unprocessed/js/molecules/Multiselect.js
@@ -3,9 +3,10 @@
 // Required modules.
 var arrayHelpers = require( '../modules/util/array-helpers' );
 var typeCheckers = require( '../modules/util/type-checkers' );
-var domTraverse = require( '../modules/util/dom-traverse' );
+var queryOne = require( '../modules/util/dom-traverse' ).queryOne;
 var domCreate = require( '../modules/util/dom-manipulators' ).create;
 var strings = require( '../modules/util/strings' );
+var bindEvent = require( '../modules/util/dom-events' ).bindEvent;
 
 /**
  * Multiselect
@@ -369,7 +370,7 @@ function Multiselect( element ) { // eslint-disable-line max-statements, inline-
   function _bindEvents() {
     var inputs = _list.querySelectorAll( 'input' );
 
-    _bind( _search, {
+    bindEvent( _search, {
       input: function() {
         _evaluate( this.value );
       },
@@ -411,7 +412,7 @@ function Multiselect( element ) { // eslint-disable-line max-statements, inline-
       }
     } );
 
-    _bind( _list, {
+    bindEvent( _list, {
       mousedown: function() {
         _isBlurSkipped = true;
       },
@@ -423,7 +424,7 @@ function Multiselect( element ) { // eslint-disable-line max-statements, inline-
           event.preventDefault();
           event.target.checked = !checked;
 
-          _fire( domTraverse.queryOne( event.target ), 'change', {} );
+          queryOne( event.target ).dispatchEvent( 'change' );
         } else if ( key === KEY_ESCAPE ) {
           _search.focus();
           collapse();
@@ -435,14 +436,14 @@ function Multiselect( element ) { // eslint-disable-line max-statements, inline-
       }
     } );
 
-    _bind( _container, {
+    bindEvent( _container, {
       mousedown: function() {
         _isBlurSkipped = true;
       }
     } );
 
     for ( var i = 0, len = inputs.length; i < len; i++ ) {
-      _bind( inputs[i], {
+      bindEvent( inputs[i], {
         change: _changeHandler
       } );
     }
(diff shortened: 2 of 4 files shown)


cfgov/unprocessed/js/molecules/Multiselect.js
```diff
@@ -352,7 +353,7 @@ function Multiselect( element ) { // eslint-disable-line max-statements, inline-
   function _bindEvents() {
     var inputs = _list.querySelectorAll( 'input' );
 
-    _bind( _search, {
+    domEvents.bindEvent( _search, {
```
anselmbradford: It seems like it might be easier to set the events on `_search`'s parent and then in the event handlers get `event.target`. See http://codepen.io/anon/pen/aNWXYZ
Maybe out of scope with this PR, but a TODO would be good if that makes sense.

anselmbradford: Sorry, I was confused. I had in my head this was binded for each item in the list, but it's a single time. It still may make sense to pull the functions out of the passed object literal so they're not anonymous, but this is fine for now.

cfgov/unprocessed/js/modules/util/dom-events.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+/**
+ * Parses the event passed by the bind shortcut to create the event listener
+ * @param   {string}   evt      The type of event to watch for
```
anselmbradford: Is `evt` a space-separated list of events?

cfgov/unprocessed/js/modules/util/dom-events.js
```diff
@@ -0,0 +1,55 @@
...
+    if ( options.hasOwnProperty( evt ) ) {
+      callback = options[evt];
+      _parseEvent( elem, evt, callback );
+    }
+  }
+}
+
+/**
+ * Shortcut for triggering events on elements
+ * @param  {HTMLNode} target     The element to fire events on
```
anselmbradford: Update `target` to `element`

cfgov/unprocessed/js/modules/util/dom-events.js
```diff
@@ -0,0 +1,55 @@
...
+    function( event ) {
+      elem.addEventListener( event, callback );
+    }
+  );
+}
+
+/**
+ * Shortcut for binding event listeners to elements
+ * @param  {HTMLNode} elem    The element to attach the event listener to
+ * @param  {Object}   options The options for the event listener
```
anselmbradford: Is this a dictionary of event listener handler functions?

cfgov/unprocessed/js/modules/util/dom-events.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Shortcut for triggering events on elements
+ * @param  {HTMLNode} target     The element to fire events on
+ * @param  {string}   type       The event type to create
+ * @param  {Object}   properties The options for the event
+ */
+function fireEvent( elem, type, properties ) {
+  var evt = document.createEvent( 'HTMLEvents' );
```
anselmbradford: `HTMLEvents` should be singular `HTMLEvent`

cfgov/unprocessed/js/modules/util/dom-events.js
```diff
@@ -0,0 +1,55 @@
...
+
+  evt.initEvent( type, true, true );
```
anselmbradford: Should this be `initCustomEvent`? If this is for IE fallback, we'll probably want to wrap it some logic to check for support since `initCustomEvent` is deprecated https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/initCustomEvent

cfgov/unprocessed/js/modules/util/dom-events.js
jimmynotjim: No, it's a single event passed by `bindEvent` below. This is another bit borrowed from awesomeplete and I'm actually unsure why more than one event would ever be passed. I've removed this entire bit because for our purposes, I don't think we'd ever need it.

cfgov/unprocessed/js/modules/util/dom-events.js
```diff
@@ -0,0 +1,55 @@
...
+
+  for ( var j in properties ) {
+    if ( properties.hasOwnProperty( j ) ) {
+      evt[j] = properties[j];
+    }
+  }
+
+  elem.dispatchEvent( evt );
```
anselmbradford: At some point we'll want to maybe investigate overlaps between this and https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/unprocessed/js/modules/util/EventObserver.js#L58 ... though I see how they fill different roles, one for one-off dispatching and the other for creating a `dispatchEvent` method on an instance.

test/unit_tests/modules/util/dom-events-spec.js
```diff
@@ -0,0 +1,73 @@
...
+  beforeEach( function() {
+    input = document.createElement( 'input' );
+    input.type = 'checkbox';
+    input.value = 'test-bind-event';
+    clicked = false;
+  } );
+
+  it( 'should not update the var until the listener is added', function() {
+    input.click();
+    expect( clicked ).to.be.false;
```
anselmbradford: Is this test necessary? It seems like since `clicked` isn't tied to `input` there couldn't be a change in the codebase that would make this fail. I guess it could fail if `bindEvent` were added in the `beforeEach` block, but isn't that making the test test itself as opposed to the codebase API?

cfgov/unprocessed/js/modules/util/dom-events.js
jimmynotjim: I've updated the comment, it's an object containing each of the events to attach.

test/unit_tests/modules/util/dom-events-spec.js
jimmynotjim: It's probably not, it's just showing that until the listener is added, the click shouldn't have any effect on the `clicked` var.

cfgov/unprocessed/js/modules/util/dom-events.js
jimmynotjim: Removed, no longer an issue :)

cfgov/unprocessed/js/modules/util/dom-events.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+/**
+ * Shortcut for binding event listeners to elements
+ * @param  {HTMLNode} elem   The element to attach the event listener to
+ * @param  {Object}   events The list of events to attach to the element
+ */
+function bindEvent( elem, events ) {
```
sebworks: I think it would be useful to provide a parameter that will allow us bind do a certain context other than the element. We often will do `element.addEventListener( 'click', _function.bind( this ) );`

jimmynotjim: You already can as the value of the event key

sebworks: So what is `this` referencing in the callback?

jimmynotjim:
```
    function() {
      this.name = 'test';

      function test( event ) {
        console.log( this.name );
      }
      domEvents.bindEvent( input, {
        click: test.bind( this )
      } );
      input.click();
    }
```

sebworks: ok, :+1:

jimmynotjim:
Rebased with flapjack and updated to a single commit (there were too many changes to rebase all the commits). Ready for a final review.

KimberlyMunoz:
:+1:

cfgov/unprocessed/js/modules/util/dom-events.js
anselmbradford: No idea why I said this, but for posterity it should be plural https://dom.spec.whatwg.org/#dom-document-createevent

PR #1688, opened 2016-03-23
Atomizes beta banner

anselmbradford:
## Removals

- Removes standalone expandable initialization code from sheer pages since it was double-initializing the beta banner. I think this is safe to do, but let me know if you know of any standalone expandables on sheer pages.

## Changes

- Moved banner into header and atomizes the banner architecture (as "Global Banner").

## Testing

- `gulp build` 
- Banner should look and function the same.
- Removing the beta banner code from the header does not break anything.

## Review

- @sebworks 
- @KimberlyMunoz 
- @jimmynotjim 

## Todos

- We haven't taken into account atomic elements that might be removed through a flag and how to gracefully bail out the JS in that case. Perhaps have a utility method in atomic-helpers that checks for the dom and doesn't continue if it's not present.


cfgov/jinja2/v1/_includes/organisms/header.html
```diff
@@ -1,5 +1,41 @@
 <header class="o-header"
         role="banner">
+
+    {% if flag_enabled('BETA_NOTICE') %}
```
anselmbradford: @rosskarchner this flag is okay outside of `base.hml`, correct?

anselmbradford:
Gotta fix the test.. hold tight.

sebworks:
:+1: , probably want to update the spec as well

anselmbradford:
Tests fixed

sebworks:
:+1:

cfgov/jinja2/v1/_includes/organisms/header.html
rosskarchner: Yes!

rosskarchner: it seems like it might choke when being inside an include (that's why my demo broke when at the D+D presentation), but this doesn't seem to have triggered that issue.

anselmbradford: Cool! We'll have to think about how to handle optional HTML on the JS-side. Like if `if flag_enabled('BETA_NOTICE')` is false, the HTML won't render and the JS behaviors attached to the HTML will error and stop all further scripts on the page from loading, essentially breaking the page. I've handled this in a `try…catch` block in this PR, but that should ideally be handled on the backend to exclude the needed JS when `BETA_NOTICE` is absent.

test/unit_tests/modules/beta-banner-state-spec.js
```diff
@@ -7,14 +7,16 @@ var sinon = require( 'sinon' );
...
 var contentDom;
 var contentAnimatedDom;
+var GlobalBanner;
 var targetDom;
-var clickEvent;
+var toggleStoredStateSpy;
 var webStorageProxy;
 
-describe( 'Beta Banner State', function() {
+describe( 'Global Banner State', function() {
```
KimberlyMunoz: Should the name of the file also be changed?

cfgov/jinja2/v1/_includes/organisms/header.html
```diff
@@ -1,5 +1,41 @@
...
+    <div class="m-global-banner">
```
KimberlyMunoz: Because the `m-global-banner` is no longer on the same div as the expandables, some of the LESS needs to change. Remove the `&` on L15 of what used to be beta.less.

test/unit_tests/modules/beta-banner-state-spec.js
anselmbradford: Oops, sure should! Fixed in https://github.com/cfpb/cfgov-refresh/pull/1688/commits/0b1e71b6209e2efef2405eda58ee1f960bf709d5

cfgov/jinja2/v1/_includes/organisms/header.html
anselmbradford: Good catch! Fixed https://github.com/cfpb/cfgov-refresh/pull/1688/commits/c96ec46b19df10e7a6e75174e948fa8cbfae8436

KimberlyMunoz:
:+1:

PR #1689, opened 2016-03-24
URL changes to match new site map

rosskarchner:
Moves /blog, /newsroom, and /about-us to their intended URL, per @schaferjh document

Initial approach was to fix this via server configuration, but this seems like a more sustainable way of doing it...

Also adds some fin-ed-resources URL's that were missing, like the new-ish money-as-you-grow.

## Changes

- change urls.py so that blog,  newsroom, and the-bureau are under /about-us
- adds missing URL's for fin-ed-resources
- change all templates and sheerlike settings to reference the new URL's
- redirects the old URL's to the new ones
- move the-bureau sheer templates under about-us

## Testing

- observe that all links go to new blog and newsroom urls, and that the blog and newsroom (including filter, etc) performs as before.

To test the fin-ed-resources stuff:

- checkout out fin-ed-resources to a sibling directory of cfgov-refresh
- in that directory, `npm install` and `grunt`
- back in cfgov-refresh, `cfgov/manage.py sheer_index`
- `runserver` and observe that you can browse to localhost:8000/money-as-you-grow/

## Review

- @kurtw
- @kave
- @richaagarwal

## Checklist


rosskarchner:
Broadening the scope of this PR to include other URL changes/additions

KimberlyMunoz:
I don't feel great about moving `/blog` to `/about-us/blog`. With the current configuration there are few broken links on the blog when we do the switch over.

`http://beta.consumerfinance.gov/blog/mortgage-moves-how-much-can-you-afford/` replaces `http://www.consumerfinance.gov/blog/mortgage-moves-how-much-can-you-afford/` and the slugs should match in most cases. That's not going to be the case when all the beta links point to `/about-us/blog`.

rosskarchner:
@KimberlyMunoz I'm not sure I understand-- what changes about the slugs?

KimberlyMunoz:
Nothing changes about the slugs, they stay the same and so the relative paths used to match up exactly before. With this change, when the switch over happens the relative paths won't match any more.

rosskarchner:
So, you mean, when a blog post links to another blog post, it'll link to /blog/whatever/?

Is it not enough if that gets caught by the redirect, so the user gets transparently bounced to /about-us/blog/whatever/?

KimberlyMunoz:
It's probably fine with the redirect. We lose some SEO link juice depending on the type of redirect, but it's still workable. I don't feel great about it because it seems like an unnecessary change to me. But it's a small concern that I wanted to raise before we do it, but it shouldn't hold it up or stop it. 

Think of this feedback as one of those annoying "Are you sure" pop-up windows.

imgres
:point_up_2:  it me

schaferjh:
It's a good point @KimberlyMunoz. Part of my thinking around this is that we will be moving the blog and newsroom under "About us" in Wagtail eventually (soon, I hope), and at that time all this change to the urls will happen anyway. So it's a redirect now or later kind of thing.

rosskarchner:
So with the attached change, links are fixed at sheer_index time. Is this a terrible idea?

Here are the rules that are applied: https://github.com/cfpb/cfgov-refresh/pull/1689/files#diff-fe16ef71622313a8f4d91e408a59b891R4

rosskarchner:
travis failure noted

cfgov/processors/processors_common.py
```diff
@@ -0,0 +1,32 @@
+from BeautifulSoup import BeautifulSoup
```
kurtw: this should be `from bs4 import BeautifulSoup`

kurtw:
After addressing the comment, I think it's good to go.

cfgov/processors/processors_common.py
rosskarchner: Thanks, kurt-- I swear there were a few days where we had both versions in our requirements.

kave:
hey ross could you rebase this branch to remove alot of the noise from other commits. thanks

rosskarchner:
done!

kave:
:+1:

rosskarchner:
Added one last commit, moving /hmda/ to /data-research/mortgage-data-hmda/

kave:
:+1:

PR #1691, opened 2016-03-24
Removed low-value browser tests

KimberlyMunoz:
I removed a lot of browser tests that were a bit fragile. Here are the cases were I removed tests.
- If the test was testing if a string existed
- If the test didn't test any logic in the template
- If the test could be broken by the CMS alone

I updated a few tests to include missing stories from the activity log and adjust for adding `- Consumer Financial Protection Bureau` to all our titles.

## Testing

- `gulp test:acceptance` or `gulp test:acceptance --sauce=false`

## Review

- @anselmbradford 
- @jimmynotjim 
- @sebworks 

## Todos

- There's a lot of things that could be tested on wagtail pages, but since we don't have a good way to match what we have locally to beta, this is hard to do. Additionally we run into the same problems of depending on CMS generated content that can break tests.

## Checklist


test/browser_tests/page_objects/page_careers.js
```diff
@@ -1,13 +1,10 @@
...
 
 var _getQAelement = require( '../util/qa-element' ).get;
-var careersSocialSection =
-require( '../shared_objects/careers-social-section' );
-
 
 function Careers() {
 
-  Object.assign( this, careersSocialSection );
+  Object.assign( this );
```
sebworks: We can remove this line.

sebworks:
:+1: , I would remove `qa-hooks` from the templates as well if they are being deleted from the tests.

jimmynotjim:
:+1: When everything is moved to wagtail, we might just have to check items exist, or the text length is greater than zero since we won't be able to guarantee the content.

KimberlyMunoz:
Would it make sense to check high-traffic pages that don't change often for functional things, like expandables working and multi-selects generating properly.

test/browser_tests/page_objects/page_careers.js
KimberlyMunoz: Thanks! Fixed.

KimberlyMunoz:
@sebworks Good looking out on the extra unused qa-hooks. Removed in 033a1ec

sebworks:
:+1: , @KimberlyMunoz.

PR #1694, opened 2016-03-24
Adds test for Tree.js

anselmbradford:
## Additions

- Adds tests to `Tree.js`

## Changes

- Makes `getAllAtLevel` method always return an array, even if it's empty.

## Testing

## Review

- @KimberlyMunoz 
- @jimmynotjim 
- @sebworks 

## Notes

- Not sure how to cover this if statement:

screen shot 2016-03-24 at 2 04 45 pm

### DIFF
--- a/cfgov/unprocessed/js/modules/Tree.js
+++ b/cfgov/unprocessed/js/modules/Tree.js
@@ -78,7 +78,9 @@ function Tree() {
    * @returns {Array} A list of all nodes at a particular tree level.
    */
   function getAllAtLevel( level ) {
-    return _levelCache[level];
+    var levelCache = _levelCache[level];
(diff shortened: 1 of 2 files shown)


anselmbradford:
I'll probably come back to this and make it like the web-storage-proxy, which groups the public methods in describe blocks, but this is a start for now.

jimmynotjim:
I think you're going to need to break this out to test each individual public method on their own `describe`s to catch that condition, mind adding that as a todo to the top of the file?

anselmbradford:
Yup, added in https://github.com/cfpb/cfgov-refresh/pull/1694/commits/4dad378c6a59876f797c19f75a0355c8d4e6378f

jimmynotjim:
:+1:

PR #1695, opened 2016-03-24
Spec file structure fixes

anselmbradford:
## Changes

- Moves Breakpoint test to correct directory to match that in the codebase.
- Adds `-spec` suffix to web-storage-proxy test.

## Review

- @KimberlyMunoz 
- @sebworks 
- @jimmynotjim


sebworks:
:+1:

PR #1697, opened 2016-03-24
Fix printing of error for sidebar-contact-info.html

kurtw:
I noticed that there was an error printing out on the contact info section of a sidebar/footer.

@richaagarwal 
@sebworks


sebworks:
:+1:

sebworks:
wait, we need that code to control the order of the blocks.

PR #1699, opened 2016-03-24
Update the Hero Macro for the new overlay hero styles

jimmynotjim:
Update the Hero Macro for the new overlay hero styles

## Additions

- Adds `overlay` and `white-text` modifiers to the hero options
- Adds `lg_image` for the background to the hero options

## Changes

- Changes the `image` option to `sm_image`
- Fixes naming in the home-hero stylesheet

## Testing

- Will need the CMS side to really test this, but you could create a hero manually on a static page with the following options.

  ```
  {
    'is_overlay': true,
    'is_white_text': true,
    'heading': 'Test Overlay Hero',
    'body': 'Test overlay hero body content',
    'lg_image': 'http://beta.consumerfinance.gov/static/img/home-hero_lg.jpg`,
    'sm_image': 'http://beta.consumerfinance.gov/static/img/home-hero_sm.jpg`,
  }
  ```

## Review

- @KimberlyMunoz 
- @anselmbradford 
- @sebworks 

## Screenshots

screen shot 2016-03-24 at 3 24 23 pm
screen shot 2016-03-24 at 3 26 30 pm

## Notes

- This is using the Homepage image even though it's not the homepage. It's what we have on beta currently that makes testing across machines simple. The flag/building image will need to be set in the CMS when the new page is published.
- After discussion with @ajbush, we decided it was better to keep the image the same as the default hero at smaller screen sizes, rather than span edge-to-edge and remove the bottom padding. It's more consistent and leads to a simpler code base.
- See https://github.com/cfpb/capital-framework/pull/301 for style changes

## Todos

- Wire the CMS to the macro

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/molecules/hero.html
+++ b/cfgov/jinja2/v1/_includes/molecules/hero.html
@@ -10,35 +10,46 @@
 
    Create a hero molecule.
 
-   value:           An object with the following requirements.
+   value:               An object with the following requirements.
 
-   value.heading:   A string containing title of module.
+   value.is_overlay:    A boolean to change the hero to an overlay hero.
+   value.is_white_text: A boolean to change the text color to white.
 
-   value.body:      A string containing the content to display below the
-                    heading (typically a paragraph).
+   value.heading:       A string containing title of module.
 
-   value.image:     A image object containing the URL of the image to be placed
-                    adjacent to the text.
+   value.body:          A string containing the content to display below the
+                        heading (typically a paragraph).
+
+   value.lg_image:      An image object containing the URL of the image to be
+                        placed behind the hero.
+
+   value.sm_image:      A image object containing the URL of the image to be
+                        placed adjacent to the text.
 
 
    value.background_color:
-                    A string containing the color code (in any format) to
-                    set as the hero background color.
+                        A string containing the color code (in any format) to
+                        set as the hero background color.
 
-   value.links:     A tuple of items to create a list of links, containing:
+   value.links:         A tuple of items to create a list of links, containing:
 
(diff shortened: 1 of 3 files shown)


cfgov/jinja2/v1/_includes/molecules/hero.html
```diff
@@ -10,35 +10,46 @@
 
    Create a hero molecule.
 
-   value:           An object with the following requirements.
+   value:               An object with the following requirements.
 
-   value.heading:   A string containing title of module.
+   value.is_overlay:    A boolean to change the hero to an overlay hero.
+   value.is_white_text: A boolean to change the text color to white.
```
anselmbradford: It seems a little inconsistent that the `background_color` can be set to anything, but the text has a flag for white or not. Would it make sense to lock down the background colors, or open up the text colors so they're handled the same way?

jimmynotjim: The BG color is for the width of the entire header, but the image only sits within the wrapper. This is only for use when the image is dark, and only for > 600px device width. When it's less than that, it will always be black text. The approved set of colors for the entire header are all light and work with black text. @ajbush or @duelj correct me if I'm wrong.

KimberlyMunoz: Locking down the approved background colors seems like a task for v2 when the designers have more bandwidth, and opening up the text colors seems unfriendly to content designers who shouldn't need to know what `#000` and `#fff` mean if those are going to be the only two options they need to use.

sebworks:
:+1:

KimberlyMunoz:
:+1:

PR #1701, opened 2016-03-25
Adds atomic-helper spec

anselmbradford:
## Additions

- Adds spec for atomic-helpers utility method.

## Testing

## Review

- @jimmynotjim 
- @sebworks 
- @KimberlyMunoz

### DIFF
--- a/test/unit_tests/modules/util/atomic-helpers-spec.js
+++ b/test/unit_tests/modules/util/atomic-helpers-spec.js
@@ -1,0 +1,51 @@
+'use strict';
+
+var BASE_JS_PATH = '../../../../cfgov/unprocessed/js/';
+
+var chai = require( 'chai' );
+var expect = chai.expect;
+var jsdom = require( 'jsdom' );
+
+var atomicHelpers = require( BASE_JS_PATH + 'modules/util/atomic-helpers' );
+
+describe( 'atomic-helpers', function() {
+
+  var containerDom;
+  var expandableDom;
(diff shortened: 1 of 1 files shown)


test/unit_tests/modules/util/atomic-helpers-spec.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+var BASE_JS_PATH = '../../../../cfgov/unprocessed/js/';
+
+var chai = require( 'chai' );
+var assert = chai.assert;
```
KimberlyMunoz: Could we stick with expect/BDD. There should be a way to use .throw with the expect syntax.

http://chaijs.com/api/bdd/#method_throw

anselmbradford: Ahh sure! What's the difference between using expect and assert? Just a style?

KimberlyMunoz: Yeah, it's mostly the syntax for how you write the tests. Expect is a style for behavior driven development, Assert is a style more for test driven development. There's a good explanation of the differences here. 

http://chaijs.com/guide/styles/

anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/pull/1701/commits/6c4e6138b154fa642dab80dd973d720ca1690b5b

KimberlyMunoz:
:+1:

sebworks:
:+1:

PR #1705, opened 2016-03-25
Updated the layout for the level 1 menu items to match design

jimmynotjim:
Updated the layout for the level 1 menu items to match design

## Changes

- Uses cf-grid to set each level 1 menu item as 1/5 of the header

## Testing

- `gulp build` and check any page, only the level 1 items should be updated.

## Review

- @schaferjh 
- @anselmbradford 
- @KimberlyMunoz 
- @sebworks 

## Screenshots

screen shot 2016-03-25 at 10 15 21 am

## Checklist

### DIFF
--- a/cfgov/unprocessed/css/organisms/mega-menu.less
+++ b/cfgov/unprocessed/css/organisms/mega-menu.less
@@ -415,7 +415,7 @@
 
             &-item {
                 display: inline-block;
-                margin-right: @grid_gutter-width / 2;
+                margin-right: @grid_gutter-width;
 
                 .no-js &:hover {
                     .o-mega-menu_content-2 {
@@ -426,7 +426,7 @@
 
             &-link {
                 padding-bottom: @grid_gutter-width;
-                margin-left: @grid_gutter-width / 2;
+                margin-left: @grid_gutter-width;
 
                 font-size: unit( 18px / @base-font-size-px, em );
 
@@ -562,6 +562,12 @@
 
     // Large desktop size.
     .respond-to-min( @bp-lg-min, {
+
+        &_content-1 {
+            &-item {
+                margin-right: @grid_gutter-width * 1.5;
+            }
+        }
         // 2nd-level menu - Large desktop.
          &_content-2 {
             &-grid {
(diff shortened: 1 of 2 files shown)


schaferjh:
To provide some context on this, and also to pull @ajbush and @duelj into the convo:

We know that the current header on beta is a little tight in spacing which leads to some hover craziness --  we got quite a few comments about it during the bureau-wide QA period we just held. We would like to create a more refined solution to this problem but at the moment, spacing things out using the columns as @jimmynotjim has done here is a quick way for us to add more breathing room to the hover target spots. However, it does mean that the spacing between items is a bit wonky.

I'm open to discussion on this and thinking about how we can modify this "quick fix" to make it bearable for now. One thing to keep in mind, is we want to lock the header by end of day Sunday/Monday morning so that we can start the work of propagating it to other templates.

duelj:
I don't think non-equal spacing is a good idea. One of the alternative mock ups when designing the header was use use the [full width for top level nav spacing](https://github.com/cfpb/cfgov-refresh/pull/1705#issuecomment-201348910). I would be more in favor of this approach, or at least larger equal spaces between top level nav items. 
screen shot 2016-03-25 at 12 49 41 pm

jimmynotjim:
The problem is there's currently no way to set equal spacing that's also responsive, at least until we've tested and implemented flexbox (which isn't going to happen before launch). We could try to add a bit more space between each item at specific breakpoints, but that still won't space them perfectly to the width of the header at all times.

jimmynotjim:
Here's what that would look like at each breakpoint

__901px__
screen shot 2016-03-25 at 12 06 59 pm

__1021px__
screen shot 2016-03-25 at 12 07 59 pm

__1231px__
screen shot 2016-03-25 at 12 08 31 pm

Looks a lot better, but it'll never be perfect.

schaferjh:
Yeah, I definitely like that latest version better @jimmynotjim

duelj:
@jimmynotjim - the last mock ups look a lot better. I'm in. I know we talked about the Amazon style target area, or putting a delay on the hover. Is this still something we are looking into as well in combination with the spacing?

schaferjh:
Yeah, that's the kind of thing I see us exploring when we create the "more refined solution" I mentioned above.

jimmynotjim:
@duelj I've never been able to get the Amazon style delay to work without making things worse somewhere else. I haven't tried since joining though, so it's something we can look at in v2.

schaferjh:
:+1:

sebworks:
:+1:

cfgov/unprocessed/css/organisms/mega-menu.less
```diff
@@ -425,8 +425,10 @@
             }
 
             &-link {
+                //display: inline-block;
```
KimberlyMunoz: Remove this line?

KimberlyMunoz:
:+1: This is a nice change that really helps the usability of this menu. :100:

PR #1706, opened 2016-03-25
`has_active_filters` needs to be accessible from all filterable type …

kurtw:
SublandingFilterablePage's are breaking because `has_active_filters` needs to be accessible from all filterable type pages so it's moved into the fitlerable_context.py in the utils module in v1 to accomplish that.

## Changes

- moved `has_active_filters` to a larger scope

## Testing

- go to a sublanding and browse fitlerable page and test out the filter controls, making sure that the notifications show up appropriately.

## Review

- @richaagarwal 
- @sebworks

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
+++ b/cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
@@ -181,7 +181,7 @@
 
 {% macro render(controls, form, index) %}
     <div class="o-filterable-list-controls">
-        {% set has_active_filters = page.has_active_filters(request) %}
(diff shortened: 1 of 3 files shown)


sebworks:
:+1: , thanks @kurtw.

sebworks:
@kurt, why doesn't `sublanding_filterable_page` inherit from `BrowseFilterablePage`?

kurtw:
Because there are different blocks in the Streamfields and Django does not let you override a database field of a parent model.

richaagarwal:
Dang you beat me to this!  Thanks for fixing my stuff.

kurtw:
No problemo

PR #1707, opened 2016-03-25
Fix ordering of related posts to be date descending

kurtw:
The data retrieved from elasticsearch was not ordered by date so this fixes that by making the most recent related post first.

## Additions

- Add `sort` list to the `Query.get_tag_related_documents` in sheerlike

## Testing

- go to http://beta.consumerfinance.gov/about-us/operations/ and not that the orderiing is wonky
- go to http://content.localhost:8000/about-us/operations/ and note the ordering is date desc

## Review

- @richaagarwal 
- @rosskarchner 

## Screenshots
screen shot 2016-03-25 at 12 02 13 pm

### DIFF
--- a/cfgov/sheerlike/query.py
+++ b/cfgov/sheerlike/query.py
@@ -237,7 +237,8 @@ class Query(object):
     def get_tag_related_documents(self, tags, size=0):
         query_file = json.loads(file(self.filename).read())
         doc_type = query_file['query']['doc_type']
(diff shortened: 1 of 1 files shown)


richaagarwal:
This looks good but I think it would be easier to test with the other PR merged in (https://github.com/cfpb/cfgov-refresh/pull/1708), since those event dates wouldn't all be the same then. But :+1:

PR #1708, opened 2016-03-25
Fix related posts dates

kurtw:
In the related posts molecule, the Event dates are corresponding to the last updated date instead of the start date of the event. This fixes that by showing the date in the following priority if it exists on the post:

1. `start_dt`
2. `date_published`
3. `last_revision_created_at`
4. `date`

## Additions

- More dates to try when rendering related posts

## Testing

- Go to http://content.localhost:8000/policy-compliance/ and check one of the related event's date
- Go to that event's page and verify that the date before is also the start date.
- Go back to the poly com page and verify that the other related post types' dates match up still

## Review

- @richaagarwal

### DIFF
--- a/cfgov/jinja2/v1/_includes/molecules/related-posts.html
+++ b/cfgov/jinja2/v1/_includes/molecules/related-posts.html
@@ -88,7 +88,10 @@
                 </p>
                 {% endif %}
                 <p class="date">
-                    {% set date = posts[i].latest_revision_created_at or  posts[i].date %}
+                    {% set date = posts[i].start_dt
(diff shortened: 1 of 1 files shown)


richaagarwal:
Hey, it looks good on the policy compliance page, but I checked another page out (http://localhost:8000/about-us/events/archive-past-events/fall-consumer-advisory-board-meeting-in-washington-dc/) and it's not in order.

richaagarwal:
Woops, nevermind, I see the ordering issue is taken care of in another PR. :+1:

PR #1712, opened 2016-03-25
Fix extra scroll bar on IE

KimberlyMunoz:
Fixes the extra scrollbar on the x-axis with a CSS fix.

## Testing

- Visit localhost/blog on IE either with a VM or Sauce Labs

## Review

- @jimmynotjim 
- @sebworks 
- @anselmbradford 

## Screenshots

image

## Notes

- The `blur` binding on IE is way more aggressive than it is on Chrome. When you grab the scrollbar to scroll, it runs the `collapse()` method. This doesn't affect people who use a scrollmouse or keyboard. I'll play with it some more, but does anyone have any ideas.

### DIFF
--- a/cfgov/unprocessed/css/atoms/multi-select.less
+++ b/cfgov/unprocessed/css/atoms/multi-select.less
@@ -26,7 +26,7 @@
 
         // Styles
         box-sizing: border-box;
-        overflow: scroll;
+        overflow-y: scroll;
         position: absolute;
         z-index: 10;
 


sebworks:
This works but whenever you click on the scrollbar the dropdown closes. Tested on VM running IE11 and windows 7. Probably out of scope for this fix though. :+1:

jimmynotjim:
Crud, I hadn't tested w/ dragging the scrollbar, just mouse/touch scrolling. Not really sure how scrollbars end up triggering things. Maybe we could attach mouse down/up listeners to the list and avoid closing if a user is interacting there?

KimberlyMunoz:
@sebworks and @jimmynotjim, I noticed that but wasn't sure if I could fix it and got this up in the meantime. Can y'all check out #1717 instead and I'll close this one?

PR #1718, opened 2016-03-25
Protect URLs from potentially leaking unpublished pages

kurtw:
Live pages have the possibility of linking to pages that are not yet live, which is fine. However, what's not fine is the information that's leaked from that link. This protects that information by replacing non-live page links on the live site with `#`. It also fixes the 404 issue for Featured Content modules.

## Additions

- context function that returns the protected URL

## Changes

- page links should use `get_protected_url` instead of `get_page_state_url`

## Testing

- Go to http://localhost:8000/data-research/research-reports/ and click on the See Full Report link on the featured content component to make sure that link is okay.
- Go to that full report's edit page and unpublish it. 
- go back to http://localhost:8000/data-research/research-reports/ and verify that the link to the full report has been replaced with a `#`

## Review

- @richaagarwal 
- @rosskarchner 
- @kave


richaagarwal:
Good catch!  :+1:

PR #1719, opened 2016-03-25
Flyout Menu: Adds aria-expanded on alt trigger

anselmbradford:
## Changes

- I'm unclear on what should have aria-expanded when the triggers to show/hide some content are outside or inside the content to be shown. 
[This](https://www.w3.org/WAI/GL/wiki/Using_the_WAI-ARIA_aria-expanded_state_to_mark_expandable_and_collapsible_regions#Example_2:_Collapsing_and_expanding_subtrees_of_a_tree) makes it seem like it should only be on the content, while [this](http://v4-alpha.getbootstrap.com/components/collapse/#example) makes it seem like it should be on the content and the triggers.
We have this structure for flyout menus (in pseudocode):

```
.trigger
.content
   .alt-trigger
```

So `.trigger` and `.content` are siblings and `.alt-trigger` is a child of `.content`.
Clicking either of the triggers toggles content.

This PR updates the FlyoutMenu to set `aria-expanded` on the alt-trigger, as it was only being applied to the trigger and content before, but is this correct?

## Testing

- `gulp build`
- Menu and search should work as before, but at mobile sizes the "Back" button should get an `aria-expanded` attribute.

## Review

- @KimberlyMunoz 
- @sebworks 
- @jimmynotjim

### DIFF
--- a/cfgov/unprocessed/js/modules/FlyoutMenu.js
+++ b/cfgov/unprocessed/js/modules/FlyoutMenu.js
@@ -108,6 +108,19 @@ function FlyoutMenu( element ) { // eslint-disable-line max-statements, no-inlin
 
   /**
    * Event handler for when the search input trigger is hovered over.
+   * @param {HTMLNode} elem - The element to set.
+   * @param {boolean} value - The value to set on `aria-expanded`,
+   *   casts to a string.
+   * @param {string} The cast value.
+   */
+  function _setAriaExpandedAttr( elem, value ) {
+    var strValue = String( value );
+    elem.setAttribute( 'aria-expanded', strValue );
+    return strValue;
+  }
+
+  /**
+   * Event handler for when the search input trigger is hovered over.
    */
   function _triggerOver() {
     if ( !_suspended ) {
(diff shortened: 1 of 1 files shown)


KimberlyMunoz:
It should be on the triggers. In that first example, someone would tab through the `li` and hear that it is unexpanded when it gets to one with `aria-expanded=false`. We probably should remove them from `.content`.

I'm not convinced that putting `aria-expanded` on the alt-triggers is useful. In our menu, you have a bunch of menus within menus. Adding "expanded" to the alt trigger could mean so many things. Is the back menu expanded? If we really wanted to make these more accessible we should do something like `Back <span class="u-visually-hidden"> to the Main Menu/About Menu/Careers Menu</span>` to give better context.

KimberlyMunoz:
Going through this again, I changed my mind. I think knowing you are in an expanded portion of the menu can help orient folks. I think there's better context that can be added, but this is still an improvement.

KimberlyMunoz:
:+1:

sebworks:
:100: , looks good.

PR #1725, opened 2016-03-28
Fix tag url for blogs and newsrooms

kurtw:
The topic tag urls were using url parameters that are Wagtail page specific for every type of page. This changes up the tags.html code so that the topic tags will work for pages that are in sheer.

## Additions

- New tags.html render() parameter to tell whether or not the page is a sheer page.

## Testing

- Go to a blog post or a newsroom item and click on the topic tag that should now link back to the newsroom or blog landing page with a working url parameter to filter based on the tag selected.

## Review

- @kave 
- @richaagarwal


richaagarwal:
:+1:

PR #1729, opened 2016-03-28
Disable GTM on menu links that don't go anywhere

anselmbradford:
## Changes

- Adds a `data-gtm_ignore="true"` to menu links that have their default behavior disabled so they don't actually work like links.

## Testing

- `gulp build` and view the source for the desktop menu. The main links should have `data-gtm_ignore="true"` attributes set.

## Review

- @sebworks 
- @jimmynotjim 
- @KimberlyMunoz


sebworks:
:+1: , I have no way of testing that this solves their issue but it doesn't break anything.

jimmynotjim:
Same as Steven :+1:

anselmbradford:
Yeah they just want it pushed up to beta to test.

PR #1732, opened 2016-03-28
Disables GTM in return to top link

anselmbradford:
## Changes

- Addendum to https://github.com/cfpb/cfgov-refresh/pull/1729

## Testing

- `gulp build`, resize to mobile, click back to top should still work.

## Review

- @sebworks 
- @jimmynotjim 
- @KimberlyMunoz

PR #1735, opened 2016-03-29
Disables GTM in return to top link

anselmbradford:
## Changes

- Addendum to https://github.com/cfpb/cfgov-refresh/pull/1729

## Testing

- `gulp build`, resize to mobile, click back to top should still work.

## Review

- @sebworks 
- @jimmynotjim 
- @KimberlyMunoz 

## TODO

- Footer code could remove jquery dependency. Dom lookups aren't needed and the animation can be covered in a function or two.


sebworks:
:+1:

anselmbradford:
I realized the test in https://github.com/cfpb/cfgov-refresh/pull/1732 would need more refactoring with making a dom query, which isn't a priority right now, so this PR just adds the GTM disabling code to the existing return to top button.

PR #1739, opened 2016-03-29
Fix JS bug

kurtw:
Initializing the blocks to get a Media.js attribute was more costly and error prone than just reusing the objects that contain it anyway.

## Removals

- unnecessary and error prone block initialization

## Testing

- Go to  http://localhost:8000/policy-compliance/amicus/briefs/ and make sure the JS is initialized properly
- Go to these to make sure they render properly:
 - localhost:8000/privacy/
 - localhost:8000/accessibility/
 - localhost:8000/about-us/project-catalyst/
 - localhost:8000/office-civil-rights/
 - localhost:8000/policy-compliance/enforcement/

## Review

- @kave 
- @richaagarwal


anselmbradford:
LGTM :+1: 

block

kave:
works for me :+1:

PR #1742, opened 2016-03-29
Apply Wagtail page seo_title or title

kurtw:
Simply checks for a Wagtail page's `seo_title` then `title`. If it's a sheer page, just append the bureau's title to whatever is there.

## Additions

- Bureau title to all page's titles.

## Testing

- Go to any page and inspect the title to see the bureau's title appended.

## Review

- @richaagarwal 
- @kave 
- @rosskarchner

### DIFF
--- a/cfgov/jinja2/v1/about-us/index.html
+++ b/cfgov/jinja2/v1/about-us/index.html
@@ -1,7 +1,7 @@
 {% extends 'layout-2-1-bleedbar.html' %}
 
 {% block title -%}
-    About us
+    About us | Consumer Financial Protection Bureau
 {%- endblock %}
 
 {% block content_main %}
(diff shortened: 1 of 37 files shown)


cfgov/jinja2/v1/about-us/index.html
```diff
@@ -1,7 +1,7 @@
 {% extends 'layout-2-1-bleedbar.html' %}
 
 {% block title -%}
-    About us
+    About us | Consumer Financial Protection Bureau
 {%- endblock %}
```
richaagarwal: I'm confused about all of these where the title is hardcoded. It would always render "About us" so why the or statement here?  And, shouldn't it just be calling super() to make use of the logic added in `base.html` ?

kave: @richaagarwal i don't think this is an `or` its just a pipe symbol

kave:
@kurtw needs to be updated but looks straight forward :+1:

cfgov/jinja2/v1/about-us/index.html
richaagarwal: oh lolol right.  okay but shouldn't we still be using `{{ post.title }}` there?

kurtw: There is no `post` context variable for these types of pages. They exist as static content.

richaagarwal: Got it. :+1:

PR #1743, opened 2016-03-29
Add screenshot.capture() test utility

anselmbradford:
Using this to check a test suite I'm building, so spinning it off here.

## Additions

- Adds screenshot.js test utility for capturing a screenshot in a running test.

## Testing

- In a test (such as `blog.js`) add `var screenshot = require( '../util/screenshot' );` (or whatever path is needed) and add `screenshot.capture()` to a test to observe, such as:

```js
  it( 'should properly load in a browser', function() {
    expect( page.pageTitle() ).toContain( 'Blog' );
    screenshot.capture();
  } );

```

- Find screenshot in `test/screenshot.png`

## Review

- @KimberlyMunoz 
- @sebworks 
- @jimmynotjim

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -96,6 +96,7 @@ collectstatic/*
 test/macro_tests/src/
 test/unit_test_coverage/
 test/perf_test_results/
+test/screenshot.png
 htmlcov/
 .tox/
 .coverage
--- a/test/browser_tests/util/screenshot.js
+++ b/test/browser_tests/util/screenshot.js
@@ -1,0 +1,39 @@
+/* ==========================================================================
+   Utility to capture a screenshot of the current browser window.
+   ========================================================================== */
+
+'use strict';
+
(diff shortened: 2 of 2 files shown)


jimmynotjim:
Works locally. I'd consider giving it a unique name based on timestamp or test name.

anselmbradford:
> Works locally. I'd consider giving it a unique name based on timestamp or test name.

So how I was using this was generating a single screenshot when debugging a single test, so I could keep the screenshot open and it'd refresh because it gets overwritten. I think having the option to generate a timestamped/test name version would be good, but perhaps an enhancement on this PR? That seems like it would be for an option to generate test suite screenshots automatically.

jimmynotjim:
Ah, now I see what you were doing. I think the option is something you can add later. :+1: from me

KimberlyMunoz:
:+1:

PR #1746, opened 2016-03-29
Filter tag slugs

kurtw:
The tag/author names can cause JS errors, breaking filterable list functions. This makes the filter use slugs instead.

## Changes

- data retrieval for tags and authors gets both slug and name instead of just name
- related-metadata links back to the tag's slug

## Review

- Go to http://localhost:8000/data-research/research-reports/ and make sure filtering still works.
- Go to a report and click one of the topic links and make sure that it applies the topic filter correctly.

## Testing

- @kave 
- @richaagarwal

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/post-preview.html
+++ b/cfgov/jinja2/v1/_includes/organisms/post-preview.html
@@ -164,9 +164,9 @@
                     </span>
                 {% endfor %}
 
-                {% if post.tags.names() | length %}
+                {% if post.tags.all() | length %}
                     {%- import 'tags.html' as tags %}
-                    {{ tags.render(post.tags.names(), page_url, true, false, form_id) }}
+                    {{ tags.render(post.tags.all(), page_url, true, false, form_id) }}
                 {% endif %}
             </div>
             {% if post.secondary_link_url and post.secondary_link_text %}
--- a/cfgov/jinja2/v1/_includes/tags.html
+++ b/cfgov/jinja2/v1/_includes/tags.html
@@ -35,11 +35,17 @@
     <ul class="tags_list">
     {% for tag in tags %}
         <li class="tags_tag">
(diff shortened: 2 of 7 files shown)


richaagarwal:
So, locally I have a tag that JS doesn't like, and causes this: ```Uncaught SyntaxError: Failed to execute 'querySelector' on 'Document': 'Consumers' voices' is not a valid selector." ``` But... this PR is supposed to fix that, right?

kurtw:
@richaagarwal that's correct.

richaagarwal:
Got it. Well, in that case, that's still not working :(

richaagarwal:
But, by the way, I don't think we need to do this right now. @schaferjh said it's OK if tags can't contain special characters.

schaferjh:
It is true that special characters in tags feels like a nice to have to me, but I'm not sure how those special characters are related to the larger purpose of this PR?

kurtw:
@richaagarwal the special characters cannot be in the values of the options because the JS will break, as you're seeing. PM me and we can figure out what's going on. I do not get this error.

@schaferjh the special characters are relevant because they break the filters' scripts to get it to render that way. 

The filter should look like this:
screen shot 2016-03-29 at 3 03 35 pm
The missing groups of "Most Frequent" and "Other" are due to JS, and @jimmynotjim knows about this.

kave:
tested, works fine :+1:

schaferjh:
cool, all sounds fine. I would only say, I give you permission to strip special characters if they are a pain point at the moment, but looks like you have things sorted. Carry on! :smile:

richaagarwal:
screen shot 2016-03-30 at 5 08 00 pm

My confusion was whether this PR is supposed to allow tags to have special characters in it or not.  If it doesn't, that's cool, because it doesn't seem like it's necessary, and we can just remove them. But, if we are, locally I have 'Consumers' voices' as a tag, and it causes a js error and things don't work.  You might have more recently pulled from Refresh, where I fixed it by removing that tag, so you might not be seeing that. But if we're not handling those, I'll go ahead and remove it and test :)

PR #1748, opened 2016-03-29
Auto-set the view more text and URL

richaagarwal:
This is part 1 of cleaning up "View more..." links to be consistent. The name is set to "View more" and the URL is set to "/activity-log".  When related posts are loaded, the URL will be dynamically updated to include filters from the tags. 

## Testing

- Create a page with Related Posts, add a bunch of tags, publish and view. Check that the view more link is correct (should have filters).

## Review

- @kave @kurtw 

## Notes

- Technically the URL doesn't need to be set in advance and it should be read-only or hidden within the admin panel, but I'm not sure how to do this. This doesn't seem too pressing as long as content editors know to leave those fields alone. @kurtw @kave @schaferjh 

- We no longer need to backfill the URLs, since they will be automatically generated. Do we still want to go in and update all the text to "View more", or is that not as pressing? @schaferjh 

## Checklist


cfgov/v1/models/atoms.py
```diff
@@ -53,8 +53,13 @@ class Hyperlink(blocks.StructBlock):
...
     url = blocks.CharBlock(default='/', required=False)
 
-    def __init__(self, required=True):
+    def __init__(self, required=True, text=None, url=None):
         self.required = required
+        if text:
+            self.base_blocks['text'].meta.default = text
+        if url:
+            self.base_blocks['url'].meta.default = url
+
```
richaagarwal: I couldn't think of a better way to do this, so if there's a cleaner way, let me know :)

schaferjh:
Just checking -- will the link to /activity-log/ be modified by the topic tags associated with the Related posts module the link is appearing in? (That would be my preference, but certainly a consistent link is better than what we got going on now.)

richaagarwal:
@schaferjh Hmm, it seems like they all have been linking to /activity-log anyway, but if you have examples of those links being more specific, I can look into this.

schaferjh:
For example here: 
http://beta.consumerfinance.gov/about-us/advisory-groups/ 
The view more links to /activity-log/?filter_tags=Academic%20Research%20Council&Community%20Bank%20Advisory%20Council&Consumer%20Advisory%20Board&Credit%20Union%20Advisory%20Council
Per the topic tags on the page.

(oh! and look there was an example I could test your fix for the multiple links! and yes! it looks better on refresh! :+1: )

richaagarwal:
Ah got it. Hmm, off the top of my head I can't think of an easy way to do that, but I will think about it more.

cfgov/v1/models/atoms.py
kave: ¯\_(ツ)_/¯ looks good to me

richaagarwal:
Updated to include filters within link. Ready for review again.

kave:
:+1: tested, works as expected

kave:
:+1: tested, looks good

kurtw:
Let's make an issue so that we update this so there's only a text field instead of a Hyperlink atom since the URL field is useless now.

cfgov/v1/models/base.py
```diff
@@ -125,9 +127,20 @@ class CFGOVPage(Page):
         ObjectList(settings_panels, heading='Configuration'),
     ])
 
+    def generate_view_more_url(self):
+        tags = []
+        for tag in self.tags.names():
```
kurtw: #1746 is what's fixing JIRA issue 1080 and that uses `self.tags.slugs()` now.

kave:
actually hold up, can you verify hyperlinks that had been previously been created are not affected by this

cfgov/v1/models/atoms.py
kave: hmm i'm seeing the case where if you set that text param somewhere. it'll be set for ALL hyperlinks everywhere instead of that one instance

richaagarwal:
@kave Hmm, it doesn't look like it's impacting previous fields. Where did this happen for you?

@kurtw Yeah, technically they're both useless because I think they always want the field to be called "View more", but I'm also not sure if they'll want to be able to edit both (in which case, I'd have to make some updates to the URL logic, but that's doable). @schaferjh Any thoughts here?

schaferjh:
If we're automating the link I think we're fine automating the "view more" text and removing the option from the editor.

richaagarwal:
Just updated this. I'll create an issue for getting rid of the Hyperlink.

kave:
:+1:

PR #1752, opened 2016-03-30
Adds unit test for BaseTransition

anselmbradford:
## Additions

- Adds unit test for BaseTransition

## Changes

- Updates transition methods to return false if they're called before the transition is initialized.

## Testing

- `gulp test:unit:scripts`

## Review

- @KimberlyMunoz 
- @sebworks

### DIFF
--- a/cfgov/unprocessed/js/modules/transition/AlphaTransition.js
+++ b/cfgov/unprocessed/js/modules/transition/AlphaTransition.js
@@ -3,13 +3,14 @@
 // Required modules.
 var EventObserver = require( '../../modules/util/EventObserver' );
 var BaseTransition = require( './BaseTransition' );
+// TODO: import Function.bind polyfill.
 
 // Exported constants.
-var CLASSES = Object.seal( {
+var CLASSES = {
   BASE_CLASS: 'u-alpha-transition',
   ALPHA_100:  'u-alpha-100',
   ALPHA_0:    'u-alpha-0'
-} );
+};
 
 /**
  * AlphaTransition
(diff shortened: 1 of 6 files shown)


sebworks:
:+1:

cfgov/unprocessed/js/modules/transition/AlphaTransition.js
```diff
@@ -43,28 +45,37 @@ function AlphaTransition( element ) {
...
 
   /**
    * Remove a "transition-duration: 0" utility CSS class.
+   * @returns {boolean}
+   *   Whether the element's CSS classes where touched or not.
+   *   False means the transition has not been initialized.
    */
   function animateOff() {
-    _baseTransition.animateOff();
+    return _baseTransition.animateOff();
```
sebworks: Should these be prefixed with `is`?

anselmbradford: Hmm, that's a good question. I don't think so in this case. What this method is doing is effectively turning the transition animation off by adding a class to the dom that sets the duration to 0, so `animateOff` describes the action that happens, which is different from something that checks the state, like `isExpanded`.

Calling this method when the associated dom element was not initialized would look the same as when it was initialized (returning undefined), so as a convenience to testing, I updated it to return true when the dom is touched, or false if it's not. However, that's not the purpose of the method (like `isDomTouched` or something), so it could return something else, like `undefined` in all cases (like classList.remove(…) does), or return the instance so the methods can be chained. But a boolean seemed to make the most sense—implying animate off either happened or didn't.

sebworks: Yeah, I guess my worry is that someone wouldn't know that you only are returning a boolean to make the code more testable. The original intent might be lost and someone might start using those boolean return values in ways you didn't intend.

Also, the transitions all seem to be repeating the same methods,  why not make them mixins?

anselmbradford: Mix them in like EventObserver? Yeah I can look into that. I can have the methods return the instance too, so they can be chained.

sebworks:
:+1:

cfgov/unprocessed/js/modules/transition/AlphaTransition.js
jimmynotjim: So is this PR still relevant or should it be closed in favor mixing them into `EventObserver`?

anselmbradford: No, it's still relevant. `BaseTransition` would just follow the pattern of `EventObserver`.

jimmynotjim:
Sounds good :+1:

anselmbradford:
Ready for review again

jimmynotjim:
:+1:

PR #1754, opened 2016-03-30
Updated header tests to support responsive testing

anselmbradford:
## Changes

- Updated header tests to support responsive testing.
- Moved stylesheet tests to a homepage test and nested promise `then` blocks to properly handle async. It looked to me like the test could run before the promises would resolve. Let me know if I'm missing something here.

## Testing

- `gulp build`
- `gulp test:acceptance --specs=organisms/header.js --sauce=false`
- `gulp test:acceptance --specs=organisms/header.js --sauce=false --windowSize=610,700`

## Review

- @sebworks 
- @KimberlyMunoz 

## Notes

- It doesn't seem like we need the header shared object because the header organism test should be the only thing needing it and can access its variables directly.


test/browser_tests/spec_suites/organisms/header.js
```diff
@@ -0,0 +1,118 @@
...
+      describe( 'at page load', function() {
+        it( 'should include Header', function() {
+          expect( _dom.header.isDisplayed() ).toBe( true );
+        } );
+
+        it( 'should include logo', function() {
+          expect( _dom.logo.isDisplayed() ).toBe( true );
+        } );
+
+        it( 'should include Mega Menu', function() {
```
sebworks: Instead of `should include`, should this be `should display`?

anselmbradford: Sure, I can update throughout

test/browser_tests/spec_suites/homepage.js
```diff
@@ -0,0 +1,41 @@
...
+    browser.get( '/' );
+  } );
+
+  it( 'should load the stylesheet', function() {
+    browser.executeScript( _getStylesheets ).then(
+      function( styleSheets ) {
+        browser.styleSheets = styleSheets;
+
+        browser.getCapabilities().then( function( cap ) {
+          browser.name = cap.get( 'browserName' );
```
sebworks: I'm not sure if this is necessary due to how `Control Flow ordering`. https://github.com/seleniumhq/selenium/issues/677. I just don't know enough about it and it's within Protractor.

test/browser_tests/spec_suites/organisms/header.js
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/pull/1754/commits/d9177fa74b300d92e11d1362c2e3627d5641c181

sebworks:
:+1:

test/browser_tests/spec_suites/homepage.js
anselmbradford: You mean the nesting of `then()` calls? Yeah I wasn't 100% certain either, but the stylesheets weren't coming through as I was expecting and with this change they're all good.
I get `Expected undefined to contain 'http://localhost:8000/static/css/main.css'` if I don't nest them, so I figured it was a problem with the expect() call before the promise resolved.

sebworks: :+1:

PR #1756, opened 2016-03-30
New Hero Options

kave:

## Additions

- New hero options
 - is_white_text
 - is_overlay
 - background_color

## Review

- @kurtw 
- @richaagarwal 
- @jimmynotjim 

## Testing 

- Create a demo page with a hero using these new options


kave:
@jimmynotjim the white text class doesn't seem to be working

screen shot 2016-03-30 at 12 46 00 pm

screen shot 2016-03-30 at 12 45 45 pm

cfgov/jinja2/v1/_includes/molecules/hero.html
```diff
@@ -46,10 +46,12 @@
...
                 if value.background_color else '' }}>
+
+    {% if value.background_image %}
+            {% set background_image=image(value.background_image, 'original') %}
+    {% endif %}
     <div class="wrapper hero_wrapper"
-          {{ 'style="background-image: ' +
-                value.background_image + ';"'
-                if value.background_image else '' }}>
+         style="background-image: url( {{  background_image.url if background_image else '' }} );">
```
anselmbradford: Why the structural change from L:50-L:52? If `background_image` isn't set this will end up outputting `style="background-image: url( '' );"`. It should remove the whole inline style if the background image URL is absent.

kave: @anselmbradford good point, i've pushed up change

cfgov/v1/models/molecules.py
```diff
@@ -63,12 +64,17 @@ class Hero(blocks.StructBlock):
     heading = blocks.CharBlock(required=False)
     body = blocks.RichTextBlock(required=False)
 
-    image = atoms.ImageBasic()
+    background_image = ImageChooserBlock(required=False,
+                                         help_text='An image object containing the URL of the image to be placed behind the hero.')
+
+    image = atoms.ImageBasic(required=False)
```
kurtw: Why this here and ImageChooserBlock above?

kave: this is a remnant from when we had not overridden the Wagtail Image Model and included alt. `ImageBasic` needs to be deleted. but i'd rather that done in a separate Issue.

kurtw:
:+1:

richaagarwal:
This is how white on black background looks for me, which seems a little strange, but otherwise :+1: 

screen shot 2016-03-30 at 5 41 22 pm

schaferjh:
Our initial use case for this uses black text on the image (see GHE//flapjack/Modules-V1/issues/112#issuecomment-72108) so if the white text needs to be left as a to-be-fixed later, that's okay. We'll just all have to remember to fix it later. :wink:

kave:
@schaferjh oh okay, that works. i'll create an issue for it to be picked up later

schaferjh:
:+1:  makin' things work...

jimmynotjim:
@kave The new styles are all missing, we need the [Capital Framework](https://github.com/cfpb/capital-framework/pull/301) side to be merged and our project updated to see them. This should all sort itself out after that happens.

Issue #1760, opened 2016-03-30
gulp test:unit:macro broken


anselmbradford:
Might explain our flaky `./setup.sh test` step on Travis?

## Current behavior

- `gulp test:unit:macro` outputs:

```
…
  File "./cfgov/v1/models/__init__.py", line 11, in <module>
    from .browse_filterable_page import *
  File "./cfgov/v1/models/browse_filterable_page.py", line 14, in <module>
    from .. import forms
  File "./cfgov/v1/forms.py", line 16, in <module>
    from sheerlike.templates import get_date_obj
ImportError: cannot import name get_date_obj
```

## Expected behavior

- `gulp test:unit:macro` should pass.

## Steps to replicate behavior

1. Run `gulp test:unit:macro`

Issue #830, opened 2015-07-31
Is gulp-load-plugins using `$` a convention?


anselmbradford:
`gulp-load-plugins` is mascarading as jquery in [our gulp tasks](https://github.com/cfpb/cfgov-refresh/blob/flapjack/gulp/tasks/test.js#L4). Is this a convention? It doesn't appear to be according to the plugin [GH repo](https://github.com/jackfranklin/gulp-load-plugins#usage).

KimberlyMunoz:
@jimmynotjim, now that you're back. Could we change this to something else?

jimmynotjim:
Sure, looking at the docs, their recommendation is to use `plugins` instead. I think I picked up `$` from a tutorial when I first started using Gulp a while back.

PR #1761, opened 2016-03-30
Fixes macro unit tests

anselmbradford:
## Changes

- Fixes https://github.com/cfpb/cfgov-refresh/issues/1760 with feedback from @rosskarchner 
- Exchanges running of macro unit tests via `exec` with `spawn` in hopes of making the macro unit tests more stable.

## Testing

- `gulp test:unit:macro` should pass.
- `gulp test --sauce=false` (or `gulp test` shouldn't stall)

## Review

- @KimberlyMunoz 
- @sebworks 
- @rosskarchner

### DIFF
--- a/gulp/tasks/test.js
+++ b/gulp/tasks/test.js
@@ -4,7 +4,6 @@ var gulp = require( 'gulp' );
 var $ = require( 'gulp-load-plugins' )();
 var sitespeedio = require( 'gulp-sitespeedio' );
 var childProcess = require( 'child_process' );
-var exec = childProcess.exec;
 var spawn = childProcess.spawn;
 var config = require( '../config' ).test;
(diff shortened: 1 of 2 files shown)


gulp/tasks/test.js
```diff
@@ -4,7 +4,6 @@ var gulp = require( 'gulp' );
 var $ = require( 'gulp-load-plugins' )();
```
sebworks: Can we not use the `$` as people closely associate with jQuery?

sebworks:
:+1:

gulp/tasks/test.js
anselmbradford: Yeah, there's this https://github.com/cfpb/cfgov-refresh/issues/830 open for that.

sebworks: awww, ok

PR #1763, opened 2016-03-31
fixed incorrect variables and added default to post_date_description …

kave:
fixed incorrect variables and added default to post_date_description to account for post preview snapshot organism

### Testing 
- Load Refresh db
- visit http://localhost:8000/policy-compliance/amicus/
 -  should render correctly and post previews should have `Published` as default text next to date since it wasn't available from the backend
 - screen shot 2016-03-30 at 11 24 04 pm

@kurtw @richaagarwal


sebworks:
:+1:

PR #1766, opened 2016-03-31
Fix bug in page.media code

kurtw:
Expandables were not showing up correctly

## Changes

- List block type blocks are now accessed correctly in page.media code

## Testing

- Go [here](http://content.localhost:8000/cfpb-ombudsman/ombudsman-faqs/) and verify that expandables work right.

## Review

- @kave 
- @richaagarwal 

## Screenshots
screen shot 2016-03-31 at 10 49 55 am

### DIFF
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -298,8 +298,7 @@ class CFGOVPage(Page):
             for child in block.child_blocks.values():
                 self._add_block_js(child, js)
         elif issubclass(type(block), blocks.ListBlock):
-            for child in block.child_block.child_blocks.values():
-                self._add_block_js(child, js)
+            self._add_block_js(block.child_block, js)
 
     # Assign the Media js to the dictionary appropriately
     def _assign_js(self, obj, js):


richaagarwal:
:+1:

sebworks:
There is a the following error on this page:

screen shot 2016-03-31 at 11 44 59 am

sebworks:
@anselmbradford , do you know why we might get that error?

anselmbradford:
Yeah, that's because the page is trying to initialize a new ExpandableGroup by passing it a dom reference that doesn't exist. The code would be something like this on a page that didn't have `o-expandable-group`:

```
// Oops o-expandable-group doesn't exist, dom = undefined
var dom = document.query( 'o-expandable-group' ); 

var expandableGroup = new ExpandableGroup( dom );
```

anselmbradford:
So the Ombudsman page does have `o-expandable-group`, so where `new ExpandableGroup(…)` is called, the dom reference passed into the constructor is not querying the page to pick out that class correctly.

kave:
:+1:

PR #1768, opened 2016-03-31
Ensures IE prefixed styles aren't double-generated

anselmbradford:
## Changes

- Updates autoprefixer version.
- Removes possibility of IE7-8 prefixing from non-ie stylesheet.

## Testing

- `gulp build` should pass. If it doesn't you'll need to run `./frontend.sh`
- Site should still look fine on IE8.

## Review

- @sebworks 
- @KimberlyMunoz


KimberlyMunoz:
LGTM

sebworks:
:+1:

PR #1773, opened 2016-03-31
Fix filter controls and sublanding page

kurtw:
Updated FilterControls organism's `show_preview_categories` flag was broken on the front end. In fixing this, it unearthed another bug which is solved in here as well. Basically, the posts in the Post Preview Snapshot organism's tags and categories links would be broken because the form_id was never gotten from the pages that the posts existed on.

## Changes

- The `get_browsefilterable_posts` now returns a tuple that is the id of the form where the post came from and the post.

## Testing

- Go [here](http://content.localhost:8000/data-research/research-reports/) and see categories displayed on the post previews. Some don't have them because they don't have categories.
- Go [here](http://content.localhost:8000/policy-compliance/amicus/) and make sure that the posts have categories and that the categories and tags have links that work to link back to the filterable page with the appropriate parameters.

## Review

- @kave 
- @richaagarwal

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/post-preview.html
+++ b/cfgov/jinja2/v1/_includes/organisms/post-preview.html
@@ -45,15 +45,18 @@
 
 {% macro render(post, controls, form_id=0, url='') %}
     {% set type = controls.page_type or '' %}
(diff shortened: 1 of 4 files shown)


kave:
:+1:

PR #1781, opened 2016-04-01
proof-of-concept: alterative approach for making in-progress work visible on content server

rosskarchner:
This can _probably_ wait until post-launch. 

Basically, this makes it so that on the "Staging" site (like content.refresh) , all drafts are visible, without needing to be "shared". It _doesn't_ make brand new, un-published content show up on lists (like on the events page). I'm not yet sure if this is a big problem or not, or how big the effort would be to make that work.

In this world 'publishing' is publishing to production, and nothing special needs to be done to make drafts visible to stakeholders.
## Additions
- If the STAGING_HOSTNAME environment variable is defined, instead of the default wagtail catch-all URL pattern, it adds one that points to [this function](https://github.com/rosskarchner/cfgov-refresh/blob/alternate-sharing-idea/cfgov/core/views.py#L19), which, if the current hostname matches STAGING_HOSTNAME, will serve the latest draft of any content.
## Removals
- Fully implemented, this would involve removing most/all of our "sharing" logic. That hasn't been done on this branch _yet_
## Testing
- check out this PR
- In wagtail, go into 'sites' and make sure the secondary site has a hostname of 'content.localhost' and port 8000 (this may not be the case if you've recently imported another database, like refresh)
  screen shot 2016-04-01 at 8 42 50 am
- edit any wagtail page (like Amicus) to add a bit of your own flavor, and SAVE (not publish or share) it. confirm that the change is NOT visible on plain old localhost:

screen shot 2016-04-01 at 8 30 23 am

screen shot 2016-04-01 at 8 30 41 am
- confirm that your change IS visible on content.localhost:

screen shot 2016-04-01 at 8 31 10 am
## Review
- @richaagarwal @kave @kurtw @tnferrell  @sarahsimpson09 @Schlachtmeyer @schaferjh
## Todos
- [ ] remove (or at least disable) sharing code
## Checklist
- [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
- [ ] New functions include new tests
- [ ] Placeholder code is flagged
- [ ] Visually tested in supported browsers and devices
- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

### DIFF
--- a/cfgov/cfgov/urls.py
+++ b/cfgov/cfgov/urls.py
@@ -9,18 +9,17 @@ from sheerlike.views.generic import SheerTemplateView
 from sheerlike.feeds import SheerlikeFeed
 from sheerlike.sites import SheerSite
 
+from core.views import preview_if_content_server
+
 from v1.views import LeadershipCalendarPDFView, unshare, renderDirectoryPDF, \
     change_password, password_reset_confirm, cfpb_login, create_user, edit_user
 
 from wagtail.wagtailadmin import urls as wagtailadmin_urls
 from wagtail.wagtaildocs import urls as wagtaildocs_urls
 from wagtail.wagtailcore import urls as wagtail_urls
-from django.views.generic import RedirectView
 
 from transition_utilities.conditional_urls import include_if_app_enabled
 
-from wagtail.wagtailadmin.forms import PasswordResetForm
-from wagtail.wagtailadmin.views import account
 
 fin_ed = SheerSite('fin-ed-resources')
 
@@ -251,7 +250,10 @@ if settings.DEBUG :
 
 # Catch remaining URL patterns that did not match a route thus far.
 
-urlpatterns.append(url(r'', include(wagtail_urls)))
+if 'STAGING_HOSTNAME' in os.environ:
+    urlpatterns.append(url(r'^((?:[\w\-]+/)*)$', preview_if_content_server, name='wagtail_serve'))
+else:
+    urlpatterns.append(url(r'', include(wagtail_urls)))
 
 from sheerlike import register_permalink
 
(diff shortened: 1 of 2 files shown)


kurtw:
So elegant. So beautiful. Shield your eyes folks. 

I'll test this very soon.

rosskarchner:
:sunglasses:

schaferjh:
So, before this PR, I believe we had 3 additive states for content:
- entered and saved in Wagtail but not visible outside of Wagtail (draft)
- saved in Wagtail and visible on content (shared)
-saved in Wagtail, visible on content, visible on refresh (live)

This appears to get rid of draft, correct? If yes, explain more about why we want to do that?

rosskarchner:
I guess my hunch was the technical complexity of providing that draft state outweighs the benefit. I certainly may have that wrong.

rosskarchner:
Also, the "bug" of things not showing up in list views may provide some of that benefit, unless unshared-draft-changes-to-already-shared-content is an important use-case.

schaferjh:
It's kind of hard to know at the moment -- certainly in our current state of entering a ton of content and pushing it live, getting rid of draft makes sense. But before we do so, I'd like to get an opinion or two from our content folks. I'll ask our team working session in a few mins.

schaferjh:
Chatted about this with @Schlachtmeyer @sarahsimpson09 and @tnferrell and we all see a value for having the draft state. Especially as we move into the future of pages that need to be reviewed in the shared state by multiple people and therefore we might need to make changes to a draft while the shared version is stable. 

If we can maintain the technical complexity, I think it is a benefit to the content edit, staging, and review process before publishing to www.

rosskarchner:
OK! I'll close this, though I may tinker with the idea some more (after launch!), to see if there is a more natural/less code way of dealing with drafts.

PR #1783, opened 2016-04-01
Fix for no max res image

sebworks:
Certain youtube videos don't have a max res image. This fixes that by using Javascript to check for the image and using a default if not present.

## Changes

- Made changes to the JS to create the image.

## Testing

- Visit `http://localhost:8000/about-us/the-bureau/` and turn JS on and off.

## Review

- @anselmbradford 
- @KimberlyMunoz 

## Screenshots

-  
screen shot 2016-04-01 at 10 51 28 am
screen shot 2016-04-01 at 10 51 21 am

## Notes
- This also fixes an error in the Jinja template due to no `else` condition.

## Todos

- I will eventually replace this method with a call to the Youtube data API, which should return a list of available images. This method will however be slower, more complex, and require a API key.

## Checklist

### DIFF
--- a/cfgov/unprocessed/css/video-player.less
+++ b/cfgov/unprocessed/css/video-player.less
@@ -48,13 +48,23 @@
     }
 }
 
-// Video image container related styles
+// Video image related styles
 
 .video-player_image-container {
     position: relative;
     // @black not used because it's not totally black.
     background-color: #000;
     text-align: center;
+}
+
+.video-player_image {
+    opacity: 0;
+    transition: opacity .25s;
+
+    .no-js &,
+    &-loaded {
+        opacity: 1;
+    }
 }
 
 .video-player_play-btn {
--- a/cfgov/unprocessed/js/modules/YoutubePlayer.js
+++ b/cfgov/unprocessed/js/modules/YoutubePlayer.js
@@ -11,7 +11,8 @@ var YoutubePlayer;
 
 var CLASSES = Object.freeze( {
   VIDEO_PLAYER_SELECTOR: '.video-player__youtube',
-  IFRAME_CLASS_NAME:     'video-player_iframe__youtube'
+  IFRAME_CLASS_NAME:     'video-player_iframe__youtube',
+  IMAGE_LOADED_STATE:    'video-player_image-loaded'
 } );
 
 var API = {
@@ -19,6 +20,8 @@ var API = {
   constructor: YoutubePlayer,
 
   SCRIPT_API: 'http://www.youtube.com/iframe_api',
+
+  IMAGE_URL: 'https://img.youtube.com/vi/%video_id%/maxresdefault.jpg',
 
   iFrameProperties: {
     id: CLASSES.IFRAME_CLASS_NAME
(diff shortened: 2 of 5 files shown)


anselmbradford:
If there is text in the video title it looks like it'll be cropped at tablet size. Perhaps the stacked style should happen at a wider breakpoint?
screen shot 2016-04-01 at 11 12 31 am

cfgov/unprocessed/js/modules/YoutubePlayer.js
```diff
@@ -20,6 +21,8 @@ var API = {
 
   SCRIPT_API: 'http://www.youtube.com/iframe_api',
 
+  IMAGE_URL: 'https://img.youtube.com/vi/%video_id%/maxresdefault.jpg',
```
anselmbradford: Technically wouldn't be a constant because the id changes. Pretty sure L:86 would break if we updated this to a `const` declaration.

cfgov/unprocessed/js/modules/YoutubePlayer.js
```diff
@@ -60,6 +67,26 @@ var API = {
...
+   * TODO: Replace this method by calling the Youtube data API.
+   * https://developers.google.com/youtube/v3/getting-started#fields
+   */
+  loadImage: function( ) {
+    if ( this.videoId ) {
+      var _image = this.childElements.image;
+      _image.classList.add( CLASSES.IMAGE_LOADING_STATE );
+      var image = document.createElement( 'img' )
+      image.onload = function onImageLoad() {
+        if ( image.naturalWidth && image.naturalWidth != 120 ) {
```
anselmbradford: Possibly add a comment above saying what the magic number `120` is.

cfgov/unprocessed/js/modules/util/data-set.js
```diff
@@ -19,7 +21,9 @@ function _toCamelCase( str ) {
  * @returns {Object} The data set.
  */
 function dataSet( element ) {
-  if ( document.documentElement.dataset ) return element.dataset;
+  if ( document.documentElement.dataset ) {
+    return _assign({}, element.dataset );
```
anselmbradford: Run `gulp lint` and fix errors.

cfgov/unprocessed/css/video-player.less
```diff
@@ -57,6 +57,15 @@
     text-align: center;
 }
 
+.video-player_image {
+    opacity: 1;
+    transition: opacity 2s;
```
anselmbradford: We could explore having unified transitions through the `u-alpha-transition`, `u-alpha-100`, and `u-alpha-0` classes, but it may not be a fit here since the utility transition duration is shorter.

Also, if you need events around the transition, AlphaTransition has them available something like this:

```
var transition = new AlphaTransition( document.querySelector( '.video-player_image' ) );
transition.addEventListener( 'transitionEnd', _handleFadeOut );
transition.fadeTo0(); // NOTE: we should update this to fadeOut/fadeIn
```

cfgov/unprocessed/js/modules/YoutubePlayer.js
sebworks: It's crazy because the youtube image will 404 but not fire `onerror`. Will add a comment.

cfgov/unprocessed/js/modules/YoutubePlayer.js
sebworks: Will this return a new string or replace the current string?

KimberlyMunoz:
Instead of changing the breakpoints, can we create a TODO for the designers to provide another default image that doesn't cut off the text?

schaferjh:
> Instead of changing the breakpoints, can we create a TODO for the designers to provide another default image that doesn't cut off the text?

+1 to this idea

cfgov/unprocessed/js/modules/YoutubePlayer.js
anselmbradford: Oo right right, replace is a new string. Carry on, all good.

anselmbradford:
> If there is text in the video title it looks like it'll be cropped at tablet size. Perhaps the stacked style should happen at a wider breakpoint?

@sebworks I realized this behavior wasn't introduced in this PR.

@KimberlyMunoz @schaferjh Sounds good. There's an existing layout that kicks in at small sizes that I was thinking could happening at a wider breakpoint:

screen shot 2016-04-04 at 8 19 51 am

But I realized there is a much wider range of sizes where it gets cropped:

screen shot 2016-04-04 at 8 19 36 am

So probably easiest would be to have a preview image that doesn't include text for now. Or leave it as is as a bug for v2.

anselmbradford:
:+1: Looks good, just needs a comment for https://github.com/cfpb/cfgov-refresh/pull/1783#discussion_r58226689

schaferjh:
I'm okay leaving it as a bug for v2 with the understanding @sebworks that you're using the About Us video in the FCM as an example here that that for that specific case on the live site, we'll see Cordray's image as the preview image across sizes?

sebworks:
That is correct @schaferjh. I may have a fix though.

jimmynotjim:
> If there is text in the video title it looks like it'll be cropped at tablet size. Perhaps the stacked style should happen at a wider breakpoint?

This is a limitation of the Featured Content Module, and related to the Google trademark being cut off from Google Maps images. I'm starting to think we need to go back to design on this module and rethink how the images are handled, there are non-text cases where it crops poorly as well.

KimberlyMunoz:
:+1:

jimmynotjim:
:+1:

cfgov/unprocessed/css/video-player.less
sebworks: Yeah, you might want to add the capability to pass an options object which allowed you to specify duration and time.

cfgov/unprocessed/js/modules/util/data-set.js
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/1783/commits/03e64c145f02123a25f7dde4f447bbc5411df6b3.

sebworks:
That is correct @schaferjh. I may have a fix though

> So I tried to convert the image to an SVG but that didn't produce results that were that much better. I'm going to go ahead and merge this PR.

PR #1576, opened 2016-03-11
Remove chosen

jimmynotjim:
Removes the Chosen.js plugin from the code base in favor of our new multi-select, includes fixes for improper option values from @kurtw 

## Additions

- WordPress processors for proper option values

## Removals

- Removes Chosen library and it's initialization and styles

## Changes

- Updates common router to initialize the multi-select script on any select with the `multi` attribute
- Fixes the values for author and tag options to remove special characters

## Testing

- checkout branch and run `python cfgov/manage.py sheer_index`
- run `gulp build`
- navigate to `/blog`, `/newsroom`, and `/activity-log` and check out the two multi-select options

## Review

- @schaferjh 
- @duelj 
- @sebworks 
- @KimberlyMunoz 
- @anselmbradford 

## Screenshots

screen shot 2016-03-11 at 12 58 21 pm

## Todos

- There's a bug in the multi-select where selected options aren't displayed after form submission (though the filters are applied). Bugfix for that in a forthcoming PR.

## Checklist

* [x] Changes are limited to a single goal (no scope creep)
* [x] Code can be automatically merged (no conflicts)
* [x] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [x] Passes all existing automated tests
* [x] New functions include new tests
* [x] New functions are documented (with a description, list of inputs, and expected output)
* [x] Placeholder code is flagged
* [x] Visually tested in supported browsers and devices
* [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #1795, opened 2016-04-02
Organize/cleanup style/scripts tasks

anselmbradford:
## Changes

- Creates separate webpack config objects for modern and IE.
- Moves IE script uglifying to webpack Uglifier plugin instead of gulp.
- Moves modernizr generation to its own task `gulp scripts:polyfill`. (@jimmynotjim – you're right, this makes sense.)
- Moves task functions to named functions.
- Remove `replace` for chosen-sprite, since chosen has been removed in https://github.com/cfpb/cfgov-refresh/pull/1576

## Testing

- `gulp build` site should work on modern browsers and IE9 as before.

## Review

- @sebworks 
- @KimberlyMunoz 
- @jimmynotjim

### DIFF
--- a/config/webpack-config.js
+++ b/config/webpack-config.js
@@ -14,7 +14,7 @@ var webpack = require( 'webpack' );
 var JS_ROUTES_PATH = '/js/routes';
 var COMMON_BUNDLE_NAME = 'common.js';
 
-module.exports = {
+var modernConf = {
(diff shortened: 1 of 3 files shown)


sebworks:
:+1:

jimmynotjim:
Nice :+1:

PR #1749, opened 2016-03-29
Split up enhancements to make them modular

jimmynotjim:
Split up enhancements to make them modular

## Additions

- individual enhancement files

## Removals

- main enhancements file

## Testing

- `gulp build`, everything should be the same

## Review

- @KimberlyMunoz 
- @anselmbradford 
- @sebworks 

## Notes

- Breaks up all of the enhancements to their appropriate CF file name making them a bit more modular.
- When trying to create a non-project header, enhancements was including a lot of extra styles.

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #1796, opened 2016-04-02
Dfjr 930 non v1 header - PROTOTYPE

anselmbradford:
Adds a method for packaging and viewing an atomic component in (relative) isolation, for moving to other projects.

## Changes

- Contains https://github.com/cfpb/cfgov-refresh/pull/1749 and https://github.com/cfpb/cfgov-refresh/pull/1795
- Adds header initialization script to /on-demand/ routes folder.
- Wraps header organism in a macro and allow passing in of option to hide banner (in addition to flag set in CMS).
- Adds /on-demand/ directory in unprocessed css directory for creating a custom stylesheet for a specific component.
- Adds a private `/test-fixture/` route for viewing a component in a stripped down webpage template.

## Testing

- Run `gulp scripts:atomic && gulp styles:atomic`
- Visit `localhost:8000/test-fixture/?atomic=header`
There should be a functional header without any other content.

## Review

- @jimmynotjim 
- @sebworks 
- @KimberlyMunoz 
- @schaferjh 
- @rosskarchner 
- @Scotchester 

## Screenshots

screen shot 2016-04-02 at 1 21 52 pm

screen shot 2016-04-02 at 1 22 02 pm

## Notes

- The idea of this PR is to have some encapsulated HTML/JS/CSS that could be copied over to responsive non-V1 pages, in a reproducible fashion, so that updates can be carried over easily in the future. However, some of our styles have the expectation of being global. I've wrapped the ones I could in an `.o-header` scope, but the others will need extra attention to see they don't break the existing sites. See `/unprocessed/css/on-demand/header.less` for which ones are not wrapped in a scope (all atomic components will have atomic prefix scope, so will not collide on pages that don't have atomic components).

### DIFF
--- a/cfgov/cfgov/urls.py
+++ b/cfgov/cfgov/urls.py
@@ -250,6 +250,7 @@ if 'selfregistration' in settings.INSTALLED_APPS:
     urlpatterns.append(pattern)
 
 if settings.DEBUG :
+    urlpatterns.append(url(r'^test-fixture/$', SheerTemplateView.as_view(template_name='test-fixture/index.html'), name='test-fixture'))
     urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
 
 # Catch remaining URL patterns that did not match a route thus far.
--- a/cfgov/jinja2/v1/test-fixture/index.html
+++ b/cfgov/jinja2/v1/test-fixture/index.html
@@ -1,0 +1,124 @@
+{# ==========================================================================
+
+   Minimal template for rendering out an atomic component in sudo isolation.
+
+   ========================================================================== #}
+
+{% set atomic_type = request.GET.get( 'atomic' ) %}
+{# Set to the Header organism by default. #}
+{% set atomic_type = 'header' if atomic_type == None else atomic_type %}
+
+<!DOCTYPE html>
+
+<!--[if IE 8]>         <html lang="en" class="no-js lt-ie10 lt-ie9"> <![endif]-->
+<!--[if IE 9]>         <html lang="en" class="no-js lt-ie10"> <![endif]-->
+<!--[if gt IE 8]><!--> <html lang="en" class="no-js"> <!--<![endif]-->
+
+<head prefix="og: http://ogp.me/ns# fb: http://ogp.me/ns/fb# {% block og_article_prefix %}{% endblock %}">
+
+<!--
+    ===========
+    GLOBAL META
+    ===========
+-->
+
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta charset="utf-8">
+    {% block meta_viewport %}
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    {% endblock meta_viewport %}
+
+<!--
+    ==================
+    PAGE-SPECIFIC META
+    ==================
+-->
+
+    <title>{% block title %}MISSING TITLE{% endblock title %}</title>
+    <meta name="description"
+          content="{% block desc %}Prototyping for the consumerfinance.gov refresh{% endblock %}">
+
+    <!-- Open Graph properties -->
+        <!--   Required  -->
+        <meta property="og:title" content="{% block og_title %}{{ self.title() }}{% endblock %}">
+        <meta property="og:type" content="{% block og_type %}website{% endblock %}">
+        <meta property="og:url" content="{{ request.url }}">
+        <meta property="og:image"
+              content="http://www.consumerfinance.gov/wp-content/themes/cfpb_nemo/_/img/logo.svg">
+        <!--   Optional -->
+        <meta property="og:description" content="{% block og_desc %}{{ self.desc() }}{% endblock %}">
+        <meta property="og:site_name" content="Consumer Financial Protection Bureau">
+        <!--   Facebook -->
+        <meta property="fb:app_id" content="210516218981921">
+        {% block og_article_author %}{% endblock %}
+    <!-- End of Open Graph properties -->
+
+    {# TODO: Explicit favicon link needed for testing. Remove for production. #}
+    <link rel="shortcut icon" type="image/x-icon" href="/favicon.ico">
+{#
+    ============
+    ONDEMAND CSS
+    ============
+    The atomic CSS file. This includes legacy IE-specific prefixing.
+#}
+<link rel="stylesheet" href="{{ static('css/' + atomic_type + '.css') }}">
+
+{#
+    ================
+    POLYFILL SCRIPTS
+    ================
+    The following scripts must be included in the <head>
+    and are used to polyfill missing functionality in legacy browsers.
+#}
+    {# Customized Modernizr build that includes html5shiv.
+       Built via gulp-modernizer in `scripts.js` task. #}
+    <script src="{{ static('js/modernizr.min.js') }}"></script>
+
+    <!--[if lt IE 9]>
+    <script>
+        // If in IE8 reverse no-js/js class change made by modernizr.
+        var docElement = document.documentElement;
+        docElement.className = docElement.className.replace( /(^|\s)js(\s|$)/, '$1no-js$2' );
+    </script>
+    <![endif]-->
+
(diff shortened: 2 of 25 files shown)


cfgov/jinja2/v1/test-fixture/index.html
```diff
@@ -0,0 +1,116 @@
...
+<!--
+    ============
+    BODY SCRIPTS
+    ============
+    The number of scripts here must be kept to a minimum.
+-->
+
+{% block javascript %}
+<script type='text/javascript'>
+    {% include '/js/atomic/' + atomic_type + '.js' %}
```
sebworks: Consider adding `ignore missing` .

anselmbradford: Added in https://github.com/cfpb/cfgov-refresh/pull/1796/commits/245870449d5108bb31cc96376fa5d2be7cf74927

sebworks:
:+1:

rosskarchner:
/cc @KimberlyMunoz 

this miiiiight br crazy, what if this was always available (not subject to settings.DEBUG), and in the Wordpress theme, used [file-get-contents](http://php.net/manual/en/function.file-get-contents.php) to pull it that URL.

Scotchester:
@rosskarchner The one used by WordPress needs to be non-responsive.

schaferjh:
@Scotchester https://github.cfpb.gov/CFPB/cfpb_nemo/pull/111

Scotchester:
@schaferjh I know; I've been reviewing it ;) Just responding to Ross's idea.

Scotchester:
@anselmbradford This seems like a pretty good idea. How do you conceive of the workflow for applying the result of this to another project? Is it related to @rosskarchner's #1798, which I haven't delved into yet? (Going to now.)

rosskarchner:
@Scotchester @anselmbradford I think the answer for that here: https://github.com/cfpb/cfgov-refresh/pull/1806

With that, we can override nemo_header.html and friends in cfgov-refresh

anselmbradford:
@Scotchester So for any non-v1 page, I think this would do it:

- Copy the header HTML directly from /test-fixture/organism=header or copy the jinja template if it is supported or use Ross' global_include https://github.com/cfpb/cfgov-refresh/pull/1798.
- Copy the `<head>` scripts for `modernizr.min.js` and `common.ie.js` for IE.
- Insert `header.js` in the footer (or in the `<head>` with `defer` and `async` attributes set).
- Insert `header.css` in the `<head>`.

The JS should be self-contained. For the CSS we have to either:
a) Add the header styles and check that the global styles don't break something on the sites they are added to.
b) Pull out the styles attached to `body`, etc. that are being applied globally and wrap them in `.o-header {}` and re-generate `header.css`.

Also, I realized one missing piece is main.ie.css, I suppose there needs to be a `header.ie.css` as well that can be exported and wrapped in IE9- conditional comments.

cfgov/unprocessed/css/enhancements/base.less
```diff
@@ -21,4 +21,7 @@ body {
   .webfont-regular();
 }
 
-
+/* topdoc
```
jimmynotjim: Argh! Thanks for catching all of these, in my haste to finish I totally forgot about them

cfgov/cfgov/urls.py
```diff
@@ -250,6 +250,7 @@
     urlpatterns.append(pattern)
 
 if settings.DEBUG :
+    urlpatterns.append(url(r'^test-fixture/$', SheerTemplateView.as_view(template_name='test-fixture/index.html'), name='test-fixture'))
```
jimmynotjim: Minor, add spaces around parens and semantic line breaks.

cfgov/jinja2/v1/test-fixture/index.html
```diff
@@ -0,0 +1,118 @@
...
+    ===========
+-->
+
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta charset="utf-8">
+    {% block meta_viewport %}
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    {% endblock meta_viewport %}
+
+<!--
```
jimmynotjim: The meta info could be packaged up for reuse on both v1 and non-v1 instead of duplicating

gulp/tasks/scripts.js
```diff
@@ -27,32 +32,65 @@ gulp.task( 'scripts:modern', function() {
...
     } ) );
-} );
+}
 
-gulp.task( 'scripts:ie', function() {
+/**
+ * Bundle IE9-specific script.
+ * @returns {PassThrough} A source stream.
+ */
+function scriptsIe() {
```
jimmynotjim: Should this be `scriptsIE` since it's an initialism?

config/webpack-config.js
```diff
@@ -36,3 +36,40 @@ module.exports = {
...
+    filename: 'common.ie.js'
+  },
+  plugins: [
+    new webpack.optimize.UglifyJsPlugin( {
+      compress: { warnings: false }
+    } )
+  ]
+};
+
+var atomicConf = {
```
jimmynotjim: All of our scripts should eventually be atomic, could this and others be updated to `onDemand` instead for clarity?

cfgov/cfgov/urls.py
anselmbradford: Doesn't look like bewd's are spacing it on the python-side. It'd be hundreds of edits for the parens in this file. Prob best for another PR.

anselmbradford:
I added IE-prefixing directly to `header.css` in https://github.com/cfpb/cfgov-refresh/pull/1796/commits/2fbca41f411d8ada30605232174c976aadb3b57c Probably easiest to more follow encapsulate the styles.

gulp/tasks/scripts.js
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/pull/1796/commits/49335e3b7094033c1037c11fda4f773feb98ba0e

config/webpack-config.js
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/pull/1796/commits/49335e3b7094033c1037c11fda4f773feb98ba0e

anselmbradford:
Pushed an update to clean up the /test-fixture/ template to remove unneeded content below meta and to import the script via src.

cfgov/jinja2/v1/test-fixture/index.html
anselmbradford: Does it look like everything in page meta could be exported in the same macro?

rosskarchner:
is the *html* produced any different from pulling in header.html? say if, #1798 were merged it could be grabbed with global_include

anselmbradford:
> is the html produced any different from pulling in header.html? say if, #1798 were merged it could be grabbed with global_include

Yes, the overlay that shows/hides under the mega-menu at mobile is not part of the Header, but is a separate atom. The full header markup is:
```
<div class="a-overlay u-hidden"></div>
{% import 'organisms/header.html' as o_header with context %}
{{ o_header.render( show_banner=false ) }}
```

We could wrap the overlay in a template too, and:

```
{{ global_include('atom/overlay.html') }}
{{ global_include('organisms/header.html') }}
```

Does that work? See the updated `/test-fixture/index.html` which I've included code comments in for the various parts. How do you want to handle the styles and JS?

anselmbradford:
@rosskarchner Oh also, how do we handle parameters like `{{ o_header.render( show_banner=false ) }}` with the global_include?

rosskarchner:
hrm, we... don't

anselmbradford:
@rosskarchner I changed the default for the banner. Ready for review!

rosskarchner:
I think I figured out the issue-- global_include-ing header.html doesn't *execute* the macro. So, maybe we need a small file that just does that

rosskarchner:
:+1: :+1:  One of the year's ten best!

cfgov/jinja2/v1/test-fixture/index.html
jimmynotjim: Everything except title and description

PR #1809, opened 2016-04-04
Check if there are params before adding them

richaagarwal:


cfgov/jinja2/v1/external-site/index.html
```diff
@@ -6,7 +6,10 @@
 {% set request_url = request_copy.pop('ext_url', None) %}
 {% set params = request_copy | urlencode %}
 {% if request_url %}
-   {% set request_url = request_url[0] + '&' + params %}
+   {% set request_url = request_url[0] %}
+   {% if params %}
+      {% set request_url = request_url + '&' + params %}
+   {% endif %}
```
kurtw: Okay. Could be shortened to `{% set request_url = request_url[0] + '&' + params if params else request_url[0] %}`

richaagarwal: Functionally they're the same, but I feel like that would be less readable. Is there a benefit to doing it in one line?

richaagarwal: Actually, nevermind, mine would potentially set twice. I'll update, thanks!

PR #1814, opened 2016-04-05
Fixing browser tests

KimberlyMunoz:
A lot of tests broke when career pages and Doing Business With Us moved. This fixes all those broken tests.

## Testing
- Run `gulp test:acceptance --sauce=false`

## Review
- @anselmbradford 
- @sebworks 
- @jimmynotjim 
- @richaagarwal


sebworks:
:+1:

Issue #1672, opened 2016-03-22
Most of the JS is broken on Blackberry


KimberlyMunoz:
The Mega menus, all expandables (Including the in this section) nav do not function.

## Current behavior

When you visit beta.consumerfinance.gov, most functions that require javascript do not work, but js styles are applying, it seems there's a JS error very early in the page.

## Expected behavior

Mega menu, expandables and other JS features should either work or fall back to no-js styles

## Steps to replicate behavior

1. Visit beta.consumerfinance.gov on your Blackberry
2. Menus and Expandables do not open

## Affected Browsers
- Blackberry

schaferjh:
oh Blackberry. :sob:

anselmbradford:
Have any JS been found that works? Beta banner is broken too. Here is the load order of the first scripts in the footer (modernizr, IE scripts, and GTM are loading in the header):

```
require( 'jquery' );
require( 'jquery-easing' );
require( 'cf-expandables' );

// Global modules.
require( '../modules/beta-banner-state' ).init();
```

PR #1828, opened 2016-04-06
Adding Fixes for BB7 and BB10

sebworks:
Adding Fixes for BB7 and BB10.  Fixes issue https://github.com/cfpb/cfgov-refresh/issues/1672.

## Additions

-  Added entry in `package.json` for `ES5-Shim`. 
-  Added `Shim/Sham` ES5 scripts to support BB7 and BB10.
-  Added build script for ES5.

## Changes

- Modified `Expandable.js` to fix issue adding transition without a prefix.
- Modified `YoutubePlayer.js` to fix issues surrounding lack of polyfill support for Object literal getters/setters.
- Modified `Base.html` to load es5 shim scripts. 

## Testing

- ~~Visit the site via Sauce Labs using a BB7 or BB10.~~ I'm working on an alternative solution for testing that doesn't make you hate Blackberry. If you were so inclined you could download the 9930 simulator and install it on your IE VM. http://swdownloads.blackberry.com/Downloads/contactFormPreload.do?code=060AD92489947D410D897474079C1477&dl=FE6AAFDF4E58488965BA8500DF214588&check1=A

## Review
- @anselmbradford 
- @KimberlyMunoz 
- @Jimmynotjim

## Screenshots

screenshot2016-04-05at7 46 45pm

## Notes

- Loading the shim scripts in `base.html` is less than ideal. Need to think of a better solution that doesn't revolve around loading polyfills via modules. 

- An alternative approach is to polyfill exactly what we need (This approach however wouldn't protect us going forward and limits our ability to use new aspects of the language). If we went with this approach we should implement the polyfills as functions as Modernizr has done and not violate the general principal of isolated module scope.

  Most of the new ES5 features we are using in a minor capacity with the exception of `Function.bind`. The rest could simply be changed if we so desired.  This is a list of features we would need to polyfill/augment:

  - `Object.seal`
  - `Object.freeze`
  - `Object.create`
  - `Object.defineProperty`
  - `Function.bind`

-

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_layouts/base.html
+++ b/cfgov/jinja2/v1/_layouts/base.html
@@ -97,6 +97,7 @@
     {# Customized Modernizr build that includes html5shiv.
        Built via gulp-modernizer in `scripts.js` task. #}
     <script src="{{ static('js/modernizr.min.js') }}"></script>
+    <script src="{{ static('js/es5-shim.js') }}"></script>
 
     <!--[if lt IE 9]>
     <script>
--- a/cfgov/unprocessed/js/molecules/Expandable.js
+++ b/cfgov/unprocessed/js/molecules/Expandable.js
@@ -36,6 +36,7 @@ function Expandable( element ) { // eslint-disable-line max-statements, inline-c
 
   var _state = COLLAPSED;
   var _transitionEndEvent = _getTransitionEndEvent( _content );
+  var _transitionPrefix = _getTransitionPrefix( _transitionEndEvent );
   var _contentHeight;
 
   // TODO: Replace function of _that with Function.prototype.bind.
@@ -168,7 +169,7 @@ function Expandable( element ) { // eslint-disable-line max-statements, inline-c
   function _transitionHeight( callback, duration ) {
     if ( _transitionEndEvent ) {
       _content.addEventListener( _transitionEndEvent, callback );
-      _content.style.transition = 'height ' + duration + 's ease-out';
+      _content.style[_transitionPrefix] = 'height ' + duration + 's ease-out';
     } else {
       // TODO: Remove callback-return ESLint ignore
       callback(); // eslint-disable-line callback-return, inline-comments, max-len
@@ -367,6 +368,22 @@ function _getTransitionEndEvent( elm ) {
     }
   }
   return transition;
+}
+
+/**
+ * @param {string} transitionEnd The browser-prefixed transition end event.
+ * @returns {string} The browser-prefixed transition event.
+ */
+function _getTransitionPrefix( transitionEnd ) {
+  var TRANSITION_PREFIXES = {
+    webkitTransitionEnd: '-webkit-transition',
+    MozTransition      : '-moz-transition',
+    OTransition        : '-o-transition',
+    transitionend      : 'transition'
+  };
+
+  return TRANSITION_PREFIXES[transitionEnd] ||
+         TRANSITION_PREFIXES.transitionend;
 }
 
 /**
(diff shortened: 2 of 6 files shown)


KimberlyMunoz:
I don't think Sauce has a BB7 or BB10 tester, but everything else looks good. Thanks for sharing that articles on polyfills on chat, I think it makes sense as an approach. :+1: 

image

cfgov/unprocessed/js/molecules/Expandable.js
```diff
@@ -370,6 +371,20 @@ function _getTransitionEndEvent( elm ) {
 }
 
 /**
+ * @param {string} transitionEnd The browser-prefixed transition end event.
+ * @returns {string} The browser-prefixed transition event.
+ */
+function _getTransitionPrefix( transitionEnd ) {
+  var TRANSITION_PREFIXES = {
+    transitionend      : 'transition',
+    webkitTransitionEnd: '-webkit-transition'
```
anselmbradford: Do you need the other platforms prefixed here as well? https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/unprocessed/js/modules/transition/BaseTransition.js#L167-L172

gulp/tasks/scripts.js
```diff
@@ -84,13 +84,31 @@ function scriptsOnDemand() {
     } ) );
 }
 
+function scriptsEs5Shim() {
```
anselmbradford: Minor: Add jsdoc above. Return type is `@returns {PassThrough} A source stream.` I'm pretty sure.

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -97,6 +97,7 @@
     {# Customized Modernizr build that includes html5shiv.
        Built via gulp-modernizer in `scripts.js` task. #}
     <script src="{{ static('js/modernizr.min.js') }}"></script>
+    <script src="{{ static('js/es5-shim.js') }}"></script>
```
sebworks: This doesn't have to go in base.html. This could be added as the first script in common.js.

cfgov/unprocessed/js/molecules/Expandable.js
sebworks: yeah, great point. Will make that change.

cfgov/unprocessed/js/shims/es5-shim.js
```diff
@@ -0,0 +1,11 @@
+/* ==========================================================================
+   ES 5 Shim/Sham, added for BB7.
+   ========================================================================== */
+
+'use strict';
+
+if( window.Modernizr && window.Modernizr.es5 === false ) {
+  // Global modules.
+  require( 'es5-shim' ); // eslint-disable-line global-require
+  require( 'es5-sham' ); // eslint-disable-line global-require
```
anselmbradford: Heh, I wondered what a sham was https://github.com/es-shims/es5-shim/commit/11f9ef990b941240d9145c24c540b8e553ce12f9#diff-04c6e90faac2675aa89e2176d2eec7d8R6

anselmbradford:
> Loading the shim scripts in base.html is less than ideal. Need to think of a better solution that doesn't revolve around loading polyfills via modules.

I was looking at https://github.com/azproduction/autopolyfiller awhile ago. Would that break best-practices? Last I looked I think it was missing some important stuff though.

cfgov/unprocessed/js/shims/es5-shim.js
sebworks: yeah, I think it is fine for our case; the only methods that would silently fail are newer Object methods if you passed a an erroneous type.

https://github.com/es-shims/es5-shim/blob/master/es5-sham.js#L460

anselmbradford:
This looks like a well-researched PR that I think would do fine for Friday. :+1: I'll try getting the emulator going too.

PR #1830, opened 2016-04-06
Related posts header and related logic updated

kurtw:
Header and relation logic for the blog's related posts have been updated.

## TEST

[here](http://localhost:8000/about-us/blog/older-americans-are-not-alone-in-the-fight-to-stop-financial-abuse/) should be an updated slug and posts that are sorted based on date and have a high enough score to be related to the current post. 

@kave


kave:
:+1:

PR #1831, opened 2016-04-06
Adds on demand templates

anselmbradford:
## Additions

- Adds on-demand templates for global include. 

## Changes

- Updates test-fixture path to pull in on-demand templates.

## Testing

- `/test-fixture/?atomic=header` should spit out the header.
- `/test-fixture/?atomic=footer` should spit out the footer (CSS will be broken because it's not atomic).

## Review

- @Scotchester 
- @rosskarchner 
- @jimmynotjim 
- @sebworks 
- @KimberlyMunoz 

## Todos

- footer.css needs to be generated with `styles:ondemand` task.

### DIFF
--- a/cfgov/jinja2/v1/_includes/on-demand/footer.html
+++ b/cfgov/jinja2/v1/_includes/on-demand/footer.html
@@ -1,0 +1,8 @@
+{# ==========================================================================
+
+   Render a footer for use in the global_include used in non-V1 projects.
+
+   ========================================================================== #}
+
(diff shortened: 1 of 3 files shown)


Scotchester:
In the CSS folder you called it `on-demand` (with a dash). Let's stick to that convention, if we can.

anselmbradford:
@Scotchester updated!

Scotchester:
:+1:

PR #1835, opened 2016-04-06
Remove popular posts for now

kurtw:
Just get rid of Popular posts for now, since it's causing problems.

@kave 
@richaagarwal


kave:
let it go! :+1:

PR #1840, opened 2016-04-06
Add browser tests for various page states

kurtw:
This is the beginning of testing Wagtail page states.

### Testing

- Run `gulp test:acceptance:wagtail --specs=wagtail/wagtail-page-states-pages.js --sauce=false --windowSize=610,700`

### Review

- @kave 
- @richaagarwal 
- @jimmynotjim 
- @anselmbradford

### DIFF
--- a/test/browser_tests/page_objects/page_wagtail_states_pages.js
+++ b/test/browser_tests/page_objects/page_wagtail_states_pages.js
@@ -1,0 +1,68 @@
+'use strict';
+
+function DraftPage() {
+
+  this.get = function() {
+    var baseUrl = '/draft-page/';
+    browser.get( baseUrl );
+  };
+
+  this.pageTitle = function() { return browser.getTitle(); };
+
+}
+
+function SharedPage() {
+
+  this.get = function() {
+    var baseUrl = '/shared-page/';
+    browser.get( baseUrl );
+  };
+
+  this.getStaging = function() {
+    var baseUrl = 'content.' + process.env.HTTP_HOST + ':' + process.env.HTTP_PORT + '/shared-page/';
+    browser.get( baseUrl );
+  };
+
+  this.pageTitle = function() { return browser.getTitle(); };
+
+}
+
+function SharedDraftPage() {
+
+  this.get = function() {
+    var baseUrl = '/shared-draft-page/';
+    browser.get( baseUrl );
+  };
+
+  this.pageTitle = function() { return browser.getTitle(); };
+
+}
+
+function LivePage() {
+
+  this.get = function() {
+    var baseUrl = '/live-page/';
+    browser.get( baseUrl );
+  };
+
+  this.pageTitle = function() { return browser.getTitle(); };
+
+}
+
+function LiveDraftPage() {
+
+  this.get = function() {
+    var baseUrl = '/live-draft-page/';
+    browser.get( baseUrl );
+  };
+
+  this.pageTitle = function() { return browser.getTitle(); };
+
+}
+
+module.exports = {
+  draftpage: DraftPage,
+  sharedpage: SharedPage,
+  livepage: LivePage,
+  livedraftpage: LiveDraftPage,
+};
--- a/test/browser_tests/page_objects/page_wagtail_states_pages.js
+++ b/test/browser_tests/page_objects/page_wagtail_states_pages.js
@@ -19,7 +19,8 @@ function SharedPage() {
   };
 
   this.getStaging = function() {
-    var baseUrl = 'content.' + process.env.HTTP_HOST + ':' + process.env.HTTP_PORT + '/shared-page/';
+    var baseUrl = 'content.' + process.env.HTTP_HOST + ':' + 
+      process.env.HTTP_PORT + '/shared-page/';
     browser.get( baseUrl );
   };
 
@@ -61,8 +62,8 @@ function LiveDraftPage() {
 }
 
 module.exports = {
-  draftpage: DraftPage,
-  sharedpage: SharedPage,
-  livepage: LivePage,
-  livedraftpage: LiveDraftPage,
+  draftpage:     DraftPage,
+  sharedpage:    SharedPage,
+  livepage:      LivePage,
+  livedraftpage: LiveDraftPage
 };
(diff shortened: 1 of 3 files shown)


test/browser_tests/page_objects/page_wagtail_states_pages.js
```diff
@@ -0,0 +1,68 @@
...
+}
+
+function SharedPage() {
+
+  this.get = function() {
+    var baseUrl = '/shared-page/';
+    browser.get( baseUrl );
+  };
+
+  this.get_staging = function() {
```
anselmbradford: Minor: Should be `getStaging`
bewds: run `gulp lint` and it'll warn you about style issues.

test/browser_tests/page_objects/page_wagtail_states_pages.js
```diff
@@ -0,0 +1,69 @@
...
+    var baseUrl = '/live-draft-page/';
+    browser.get( baseUrl );
+  };
+
+  this.pageTitle = function() { return browser.getTitle(); };
+
+}
+
+module.exports = {
+  draftpage: DraftPage,
```
jimmynotjim: Left align the values in these pairs to avoid linter complaints

jimmynotjim:
Tests pass, there's a few other linting warnings. None seem major so we can probably tackle those later.

gulp/tasks/test.js
```diff
@@ -154,6 +154,35 @@ gulp.task( 'test:perf', sitespeedio( {
...
+      );
+  });
+});
+
+gulp.task( 'test:acceptance',
+  [
+    'test:acceptance:browser'
+  ]
+);
+
```
anselmbradford: Delete edits to test.js L:157-L:185. This is duplicating the tasks.

test/browser_tests/spec_suites/wagtail/wagtail-page-states-pages.js
```diff
@@ -0,0 +1,84 @@
...
+  var page;
+
+  beforeAll( function() {
+    page = new DraftPage();
+    page.get();
+  } );
+
+  it( 'should not load in a browser',
+    function() {
+      expect( 404 );
```
anselmbradford: I don't think this'll do what you expect. It's checking whether the number 404 is 404, which it always will be true. You can change the number here and it'll still pass.

Looks like checking response headers is complicated http://stackoverflow.com/questions/25137881/how-to-use-protractor-to-get-the-response-status-code-and-response-text It might be easier to find something unique in the 404 pages and check for that.

test/browser_tests/spec_suites/wagtail/wagtail-page-states-pages.js
```diff
@@ -0,0 +1,84 @@
...
+      page.get();
+      expect( 404 );
+    }
+  );
+
+  it( 'should properly load in a browser',
+    function() {
+      page.getStaging();
+      expect( page.pageTitle() ).toContain(
+        'Shared Page | Consumer Financial Protection Bureau' );
```
anselmbradford: This test expectation is failing for me. Says `Page not found at /shared-page/`. Maybe I don't have the latest db locally though.

KimberlyMunoz: I grabbed the DB and it was still failing. When I ran the publish all shared pages scripts, it pased, but I think Shared Pages probably shouldn't be able to load in the browser until published, right?

richaagarwal: @kurtw get the same error as Ans. Is this navigating to content.localhost?

test/browser_tests/spec_suites/wagtail/wagtail-page-states-pages.js
richaagarwal: Yeah, also one thing that is useful/ we should be doing as we write tests is to make sure our tests fail with other input.

test/browser_tests/spec_suites/wagtail/wagtail-page-states-pages.js
richaagarwal: Ok so I don't think `page.getStaging();` is doing what you want it to, because if I replace it with `browser.get('content.localhost:8000/shared-page');` the test passes

richaagarwal:
Other than the comments this looks good. 

One question/feedback I have, though not meant to be a blocker - why did you abstract the functions into `page_wagtail_states_pages.js` ? The naming is sort of confusing, but beyond that it seems like it's not actually needed to keep things DRY. Instead of `page.get();`, it would just be `browser.get('/draft-page/');` or `browser.get('/shared-draft-page')` etc., with no need to create a Page object.  They're all used exactly once, so unless we anticipate using them more than once, this seems needlessly complex.

test/browser_tests/spec_suites/wagtail/wagtail-page-states-pages.js
kurtw: Everyone, check that your Wagtail sites have been set correctly. Make sure they're `content.localhost` port 8000 and `localhost` port 8000. Also, make sure that `HTTP_HOST` and `HTTP_PORT` env variables are set.

kave:
@richaagarwal yea we wanted to keep our tests inline with the convention stared by the other browsers tests, hence the `page.get()`

test/browser_tests/spec_suites/wagtail/wagtail-page-states-pages.js
richaagarwal: @kurtw Meant to respond to this sooner, but yeah my sites and env variables are set correctly.

anselmbradford:
FWIW I didn't use page objects for the organism tests (e.g. https://github.com/cfpb/cfgov-refresh/blob/flapjack/test/browser_tests/spec_suites/organisms/mega-menu.js#L28) for the reasons @richaagarwal pointed out

richaagarwal:
:+1:

richaagarwal:
Should this be merged in or closed?

kurtw:
I'm coming back to this as per @richaagarwal and @anselmbradford comments about DRYness over consistency.

kurtw:
Ready for another review

richaagarwal:
👍

PR #1853, opened 2016-04-07
Reduce mobile header height + add styles per spec feedback

anselmbradford:
## Changes

- Reduces height of mega menu and global search triggers.
- Creates `@mobile_trigger_ht__px` to track header height.
- Adds right-side border to mega-menu hamburger menu.

## Testing

- `gulp build`
- Boot up the iOS simulator for an iPhone 5.
- C-c-check it out and click the menu and search buttons.

## Review

- @sebworks 
- @jimmynotjim 
- @KimberlyMunoz 
- @ajbush @duelj 

## Screenshots

screen shot 2016-04-07 at 2 47 55 pm

### DIFF
--- a/cfgov/unprocessed/css/molecules/global-search.less
+++ b/cfgov/unprocessed/css/molecules/global-search.less
@@ -1,5 +1,5 @@
 
-// TODO: Move the theme variables to cf-enhancements.
+// TODO: Move the theme variables to /enhancements/ for CF.
 /* topdoc
   name: Theme variables
   family: cf-core
@@ -12,12 +12,14 @@
         - |
           @margin__em
           @margin_half__em
+          @mobile_trigger_ht__px
   tags:
     - cf-core
 */
 
 @margin__em: unit( @grid_gutter-width / @base-font-size-px, em );
 @margin_half__em: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
+@mobile_trigger_ht__px: 54px;
 
 /* topdoc
   name: Global Search molecule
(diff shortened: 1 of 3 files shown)


Scotchester:
The conventional wisdom I've heard is that it's not good to prevent the user from zooming if they want to.

But let's back up a little... rather than sizing the logo down smaller than our standards officially allow, would simply centering it correctly fix the problem? Right now it's sitting significantly too far to the right.

That zooming behavior when invoking search is wack, though...

sebworks:
@Scotchester could we use `srcset` http://alistapart.com/article/responsive-images-in-practice?

anselmbradford:
@Scotchester It looks like it's too far to the right, but it actually has a 10px gap on both left and right, which is more visible when the menu is clicked:

Before:

screen shot 2016-04-07 at 11 18 07 am

After:

screen shot 2016-04-07 at 11 17 56 am

I suppose we could reduce the left/right logo margin to 5px/5px. How does that sound?

Scotchester:
That's not what I'm seeing:

screen shot 2016-04-07 at 11 25 45

anselmbradford:
Weird.. have you `gulp build`? 

screen shot 2016-04-07 at 11 31 48 am

Scotchester:
I'm just looking at straight-up Beta and flapjack.demo.

duelj:
Good catch, the design for the default hamburger menu has a border around which I think helps the logo feel more grounded with the spacing. I am not sure where that got lost along the way. The selected state for the search was originally a rectangle so it allowed extra room for the logo (as seen below) The mobile header was also supposed to be 60px tall with the 4px green stroke rather than the current 64px. This would make the selected state boxes smaller as well to gain some more space.
screen shot 2016-04-07 at 12 51 14 pm


If we use the logo at 35px tall we should also be in the clear on spacing for the menu and search. The DM allows for "variance" from the 40px tall spec. and there is also a small execution spec. that says the logo should be wider than 54px which the 35px tall version comes in at 70px wide so it's find for that guideline.

jimmynotjim:
Let's not prevent users from zooming, it's an important feature, especially for those with less than ideal vision.

anselmbradford:
Okay, totally updated this PR per @duelj feedback. Ready for review again. See top for new PR state.

jimmynotjim:
:+1:

PR #1858, opened 2016-04-07
Fix download links from becoming relative

kurtw:
Download links were being set to relative links so `files.consumerfinance.gov` wouldn't have ever been hit. This prevents this from happening.

@rosskarchner @kave @richaagarwal


kave:
:+1:

PR #1864, opened 2016-04-08
Remove output formatting to get back what was put in

kurtw:
Looks like the `html` formatter from before caused some weird output for `>`-like symbols. This jsut removes formatting entirely so we're just getting back what was put in to begin with. Safe enough, though the docs say that it might output invalid html, but that seems to be the case if the html was invalid to begin with.

`cfgov/manage.py sheer_index`

http://localhost:8000/about-us/blog/mortgage-moves-what-kind-of-mortgage-will-you-get/ check for correct output of apostrophes

http://localhost:8000/about-us/the-bureau/leadership-calendar/ check descriptions for correct output of `>`'s
screen shot 2016-04-08 at 11 32 37 am
@rosskarchner @kave @richaagarwal

### DIFF
--- a/cfgov/sheerlike/external_links.py
+++ b/cfgov/sheerlike/external_links.py
@@ -14,7 +14,7 @@ def process_external_links(doc):
 def _process_data(field):
     if isinstance(field, basestring):
         soup = BeautifulSoup(field, 'html.parser')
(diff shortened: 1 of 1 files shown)


kave:
:+1:

PR #1874, opened 2016-04-09
Hotfix field sort

kurtw:
Change the query's sort field so we actually get calendar events by day.

@kave @rosskarchner @richaagarwal @sebworks @anselmbradford @jimmynotjim @KimberlyMunoz


sebworks:
:+1:

richaagarwal:
:+1:  do we also want to update the size?

PR #1875, opened 2016-04-09
Hotfix: Changes overlay click behavior for iOS/BB

anselmbradford:
iOS was not dismissing the menu on click of the mega menu semi-opaque overlay.
BB was dismissing it too soon.

## Changes

- Adds `cursor: pointer` to the styles for the overlay so that iOS bubbles the click event up to the body, which dismisses the menu in the mega menu script.
- Changes the event type to `click` instead of `mousedown`, so that a `mouseup` event is required to the dismiss the menu (hopefully this fixes the issue on BB).

## Testing

- `gulp build`
- Load in iOS simulator (try iPhone 5)
- Click hamburger menu, scroll down till the menu isn't visible, and click the overlay. Menu should close

## Review

- @jimmynotjim 
- @sebworks 
- @KimberlyMunoz

### DIFF
--- a/cfgov/unprocessed/css/atoms/overlay.less
+++ b/cfgov/unprocessed/css/atoms/overlay.less
@@ -26,6 +26,10 @@
 
         background: @black;
         opacity: 0.6;
+
+        // Cursor needs to be set for iOS to capture clicks of the overlay
+        // and bubble them up to the body.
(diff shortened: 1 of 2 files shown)


jimmynotjim:
Works locally :+1:

PR #1877, opened 2016-04-10
updated tags to account for events

kave:
This event pages tags were not routing to the correct url. that has been fixed

e.g [here](http://localhost:8000/about-us/events/spring-2016-community-bank-advisory-council-meeting-washington-dc/)

@kurtw @richaagarwal


kurtw:
If it turns out it only has to link back to the current events filterable list then :+1: otherwise there should be some logic to determine if the event is from the archive or not.

PR #1881, opened 2016-04-11
fixed mailTo link rendering in richtext

kave:
go [here](http://localhost:8000/about-us/project-catalyst/run-trial-disclosure/) an click step 1 or 3 in the middle of the page.

@kurtw @richaagarwal

### DIFF
--- a/cfgov/processors/processors_common.py
+++ b/cfgov/processors/processors_common.py
@@ -19,6 +19,6 @@ def update_path(path):
 def fix_link(link):
     urldata = urlsplit(link['href'])
     new_path = update_path(urldata.path)
-    new_href = urlunsplit((None, None, urldata.path, urldata.query,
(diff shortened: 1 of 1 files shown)


richaagarwal:
:+1:  just a note - let's be sure to re-import wordpress posts once this is pushed, since those are affected too

rosskarchner:
Thanks @richaagarwal for pointing out that I tried to fix the same bug with #1883.

rosskarchner:
Somewhere along the way, this function stopped doing what it's supposed to do. I checked both this, my version (#1883) and before @kurtw's recent refactoring of this code, all basically do nothing:

```
>>> link = {'href':'/blog/foo/'}
>>> fix_link(link)
>>> print link
{'href': '/blog/foo/'}
>>> 
```

If it was working, that would have become /about-us/blog.

This isn't a critical feature, though, and I did confirm that with this PR, we are no longer mangling mailto's.

👍

PR #1884, opened 2016-04-11
Fix mobile text intro

KimberlyMunoz:
Added a check to avoid blank link lists on mobile.

## Testing

- Visit http://localhost:8000/about-us/events/ on mobile

## Review

- @anselmbradford 
- @sebworks 
- @jimmynotjim 

## Screenshots

It should _not_ look like this

image

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/molecules/text-introduction.html
+++ b/cfgov/jinja2/v1/_includes/molecules/text-introduction.html
@@ -27,14 +27,14 @@
 <h1>{{ value.heading }}</h1>
 <div class="lead-paragraph">{{ parse_links(value.intro) | safe }}</div>
 {{ parse_links(value.body) | safe }}
-{% if value.links.url %}
-  <ul class="list list__links">
-      {% for link in value.links %}
-          <li class="list_item">
-              <a class="list_link" href="{{ link.url }}">
-                  {{ link.text }}
-              </a>
-          </li>
-      {% endfor %}
-  </ul>
-{% endif %}
+{% for link in value.links %}
+    {% if link.text %}
+        {% if loop.first %}<ul class="list list__links">{% endif %}
+            <li class="list_item">
+                <a class="list_link" href="{{ link.url }}">
+                    {{ link.text }}
+                </a>
+            </li>
+        {% if loop.last %}</ul>{% endif %}
+    {% endif %}
+{% endfor %}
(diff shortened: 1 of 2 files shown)


cfgov/jinja2/v1/_includes/molecules/text-introduction.html
```diff
@@ -27,12 +27,14 @@
...
-<ul class="list list__links">
-    {% for link in value.links %}
-        <li class="list_item">
-            <a class="list_link" href="{{ link.url }}">
-                {{ link.text }}
-            </a>
-        </li>
-    {% endfor %}
-</ul>
+{% if value.links.url %}
```
anselmbradford: If `for link in value.links` exists further down, will `value.links.url` ever have a value since it's presumably an array? Wouldn't it need to be something like `value.links | count > 0` and then further down have `if value.link.url`

KimberlyMunoz: This is how we do the check for item-introduction and it seems to work on http://localhost:8000/about-us/careers/.

jimmynotjim:
:+1:

cfgov/jinja2/v1/_includes/molecules/text-introduction.html
anselmbradford: Where's the bit in /careers/ where this is done? Trying to find it..
Do we have a way to test this when the links value has a URL? I worried it may always end up being false and never show, even when we want it to.

KimberlyMunoz: Hmmm... You know, the futzing around with the item introduction, it actually seems a little off. Going to update both of these in a second, the testing instructions will be a little more involved though.

KimberlyMunoz:
Re-wrote the way we check for blank authors/links. It's clunky, but in a perfect world people would delete the blank author and link objects in the CMS and we don't live in that perfect world. :cry:

The new testing directions: 

1. Log in the CMS and create a Document Detail Page with an item intro and text intro. Leave it with the default, a blank link for both and add titles and a date for item introduction and check.

image
2. Add one author, link for each

image
3. Add two authors/links for each

image

sebworks:
:+1:

jimmynotjim:
:+1:

schaferjh:
Oooo, this is nice. I just made a backlog task for changing the Text Introduction organism in the CMS so that the link option is not there be default. (That's why there are so many blank links, the field shows up when we add the module in the CMS, but it doesn't error if you don't enter link information and most introductions don't have a link that follows.)

KimberlyMunoz:
Thanks Jess, I added some context to the final commit message so we can clean this up after that backlog task is in and we clear out the unnecessary blank links in the CMS. Someday... 🌈

PR #1887, opened 2016-04-11
Scroll back to top on menu clicks

anselmbradford:
Always go to the top of the page as navigating between menus.

## Changes

- Scroll back to top on menu changes.
- Turn suspended checks into guard clauses.

## Testing

- `gulp build`
- Open in mobile view.
- Go to About Us, scroll down till Careers button is at the top of the screen.
- Click Careers entries. 
Should scroll to top and slide careers menu into view.

## Review

- @jimmynotjim 
- @KimberlyMunoz 
- @sebworks

### DIFF
--- a/cfgov/unprocessed/js/organisms/MegaMenuMobile.js
+++ b/cfgov/unprocessed/js/organisms/MegaMenuMobile.js
@@ -74,17 +74,16 @@ function MegaMenuMobile( menus ) {
    * @param {Event} event - A FlyoutMenu event.
    */
   function handleEvent( event ) {
-    if ( !_suspended ) {
-      if ( event.type === 'triggerClick' ) {
-        _handleTriggerClickBinded( event );
-      } else if ( event.type === 'expandBegin' ) {
-        _handleExpandBeginBinded( event );
-      } else if ( event.type === 'collapseBegin' ) {
-        _handleCollapseBeginBinded( event );
-      } else if ( event.type === 'collapseEnd' ) {
-        _handleCollapseEndBinded( event );
-      }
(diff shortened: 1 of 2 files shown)


jimmynotjim:
Nice fix :+1:

cfgov/unprocessed/js/organisms/MegaMenuMobile.js
```diff
@@ -74,17 +74,16 @@ function MegaMenuMobile( menus ) {
...
-        _handleTriggerClickBinded( event );
-      } else if ( event.type === 'expandBegin' ) {
-        _handleExpandBeginBinded( event );
-      } else if ( event.type === 'collapseBegin' ) {
-        _handleCollapseBeginBinded( event );
-      } else if ( event.type === 'collapseEnd' ) {
-        _handleCollapseEndBinded( event );
-      }
-    }
+    if ( _suspended ) { return; }
```
jimmynotjim: Mixed `{}` and not for single line conditions. We should standardize on this, if not for the project in the entire front-end team

KimberlyMunoz: I'm surprised the linter doesn't have a rule for this.

anselmbradford: Fixed! We are allowing single line conditionals to not have braces [here](https://github.com/cfpb/cfgov-refresh/blob/flapjack/.eslintrc#L540). We could swap that to false, but will need to add braces in a few places.

kave:
👍

PR #1891, opened 2016-04-12
Fix for BB on non V1 pages

sebworks:
Fix for BB on non V1 pages. We have opted not to include a polyfill or shim for various reasons. Instead, we have chosen to use an internal function to augment `Function.prototype.bind`. I think this approach is the least invasive when including the header/footer modules on other sites. It's not a complete approach as we are using `.bind` instead of `fnBind` in other modules.

This fix doesn't work on the emulator but does on a real device. What a fun time that was discovering this fact. 

## Changes

- Changed `Function.prototype.bind` instances to use internal function.

## Removals

- Removed instances of `Object.seal` to avoid shimming/shamming/polyfill augmenting. We will address immutable constants at a later date.

## Testing

- Browse the site paying close attention to the nav at various breakpoints. Visit `http://localhost:8000/test-fixture/?atomic=header` and do the same. 

## Review

- @KimberlyMunoz 
- @anselmbradford 
- @jimmynotjim 

## Todos

-

## Checklist

### DIFF
--- a/cfgov/unprocessed/js/modules/util/assign.js
+++ b/cfgov/unprocessed/js/modules/util/assign.js
@@ -9,10 +9,10 @@
 
 'use strict';
 
-var fnBind = require( './fnBind' ).fnBind;
+var fnBind = require( './fn-bind' ).fnBind;
 
 /**
-* @param {Object} object - JavaScript object.
+* @param {object} object - JavaScript object.
 * @returns {boolean} True if object is plain Javascript object.
 */
 function _isPlainObject( object ) {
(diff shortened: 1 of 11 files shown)


cfgov/unprocessed/js/modules/util/assign.js
```diff
@@ -9,10 +9,10 @@
 
 'use strict';
 
-var fnBind = require( './fnBind' ).fnBind;
+var fnBind = require( './fn-bind' ).fnBind;
 
 /**
-* @param {Object} object - JavaScript object.
+* @param {object} object - JavaScript object.
```
anselmbradford: This one was good capitalized. The only lowercase ones would be `string`, `boolean`, `number`, `null`, and `undefined`

anselmbradford:
LGTM 👍

PR #1900, opened 2016-04-14
fixed leadership calendar filtering for pdf downloads

kave:
The redirect was tossing the `GET` request params, so i updated the url path created by the service.

Also created a ELASTICSEARCH_BIGINT variable as es doesn't currently have a way to return all documents without pagination of some sort which won't work for a pdf creation.

@kurtw @richaagarwal @rosskarchner 

## Testing
- Testing this locally will be a pain for you, wait to test on flapjack after merging.


richaagarwal:
:+1: for getting it on flapjack, let us know the testing instructions once it's merged and pushed to refresh

rosskarchner:
👍 
I did a thing with [the housing counselor PDF download ](https://github.com/cfpb/cfgov-refresh/pull/1907/files#diff-ce8b16726af60f2157fbb75fb8b61a7cR17) that, if debug is on and PDF reactor isn't set up, responds with the URL that it *would* have rendered

kave:
@rosskarchner oh thats a good idea, i'll add it in too

PR #1901, opened 2016-04-14
cleaned up init data

kave:
@kurtw @richaagarwal


richaagarwal:
It would be ideal to get this logic into `setup_sites.py` instead and use that, but looks good :+1:

PR #1905, opened 2016-04-15
Set up seo description content for pages

kurtw:
Sets it up for `page`

@kave @richaagarwal @rosskarchner


richaagarwal:
:+1:  Was `seo_description` just a placeholder before?

rosskarchner:
👍   I wonder if the same text should go into the OpenGraph description?

https://github.com/cfpb/cfgov-refresh/blob/set-seo-description/cfgov/jinja2/v1/_layouts/base.html#L64

kave:
👍

PR #1911, opened 2016-04-16
Molecules tests

richaagarwal:
@rosskarchner @kave @kurtw  - Just adds some more tests for molecules. 

- Refresh db 
- Run `gulp test:acceptance:wagtail --specs=molecules.js`


kurtw:
Each one failed for me :( Are there any other things I must do?

richaagarwal:
I would make sure you've reset the database. Ideally we would just have that as part of the test script; not sure why we don't :\

rosskarchner:
I got 2/8 --

When you say "refresh DB", do you mean run initial-test-data.sh, or load a recent snapshot of the 'refresh' server?

rosskarchner:
it worked! 👍

PR #1912, opened 2016-04-18
Pdfreactor Exception message

kave:
- included an additional check for pdfReactor to render a better exception message for local developers 
- fixed imports for PDFreactor in HousingCounselorPDFView breaking the views.

- @kurtw 
@richaagarwal @rosskarchner

### DIFF
--- a/cfgov/legacy/views.py
+++ b/cfgov/legacy/views.py
@@ -8,7 +8,7 @@ if six.PY2:
         sys.path.append(os.environ.get('PDFREACTOR_LIB'))
         from PDFreactor import *
     except:
-        PDFReactor = None
+        PDFreactor = None
 
 
 class HousingCounselorPDFView(PDFGeneratorView):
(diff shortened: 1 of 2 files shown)


rosskarchner:
hold a sec!

rosskarchner:
👍  made a few tweaks I had been working on in a different branch, but should be good now

cfgov/legacy/views.py
```diff
@@ -13,9 +21,6 @@ def get_render_url(self):
     def get(self, request):
 
         self.request = request
-        if settings.DEBUG and PDFReactor is None:
-            return HttpResponse("PDF Reactor is not configured, can not render %s" % self.get_render_url())
```
rosskarchner: why remove this? (though the capitalization was incorrect, it should be PDFreactor)

kave: yeah the capitalization was causing errors, thought it was related to pdfReactor not loading correctly. i've put it back in,  ~_~ its been a slow morning.

kurtw:
Seems legit

PR #1913, opened 2016-04-18
included post date description field to the post preview snaphot

kave:
@kurtw @richaagarwal 

### testing 
- Edit [page](http://localhost:8000/admin/pages/40/edit/) and publish
screen shot 2016-04-18 at 12 24 02 pm

- view change [here](http://localhost:8000/policy-compliance/amicus/) 

 
before:
screen shot 2016-04-18 at 12 21 02 pm

after:
screen shot 2016-04-18 at 12 20 45 pm


schaferjh:
This looks like a viable solution! Just wondering -- is there anyway it could automatically pull from the editable field in the child filterable list? Making it another editable field at this level just introduces the possibility of a mismatch created by a content entry error. Or, that's a refinement we can look into later!

kave:
@schaferjh i had thought about that, but in scenarios where there are multiple child filterable pages, how do we select which takes precedence?

schaferjh:
> in scenarios where there are multiple child filterable pages, how do we select which takes precedence?

Ah, now that gets at something I think *should* be editable at this level: the content entry person should be able to select which child filterable page the snapshot is linked to.

kave:
@schaferjh hmmm that may take me sometime to implement, as i'll need to override some wagtail admin templates. Can we have this as mvp and make that a story for post launch?

schaferjh:
For launch we just need the date descriptions to match. So if the MVP way to accomplish that is adding in the editable field per your initial idea, that's fine. But let's definitely add a TODO/backlog item for the more long term fix. Thanks!

PR #1914, opened 2016-04-18
included netloc in the url

kave:
@kurtw @richaagarwal 

### Testing
- Verify you can view the pdf [here](http://localhost:8000/policy-compliance/amicus/briefs/sheriff-gillie/)


rosskarchner:
Hmmm-- my intention there was to remove the domain from links that point to www, making them relative. In this case, I think it's the unnecessary *addition* of http is the real bug.

rosskarchner:
👍

PR #1917, opened 2016-04-18
included a flag to allow/disallow access to Django-Admin & Wagtail Admin

kave:
@kurtw @rosskarchner @richaagarwal

### DIFF
--- a/cfgov/cfgov/urls.py
+++ b/cfgov/cfgov/urls.py
@@ -26,25 +26,7 @@ from wagtail.wagtailadmin.views import account
 fin_ed = SheerSite('fin-ed-resources')
 
 urlpatterns = [
-    url(r'^django-admin/login', cfpb_login, name='wagtailadmin_login'),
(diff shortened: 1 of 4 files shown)


cfgov/cfgov/urls.py
```diff
@@ -251,6 +233,37 @@
     # Report redirects
     url(r'^reports/(?P<path>.*)$', RedirectView.as_view(url='/data-research/research-reports/%(path)s', permanent=True)),
 ]
+
+if settings.DEBUG:
```
rosskarchner: It might be a little more transparent to put this logic in settings.base?

like,

`ALLOW_ADMIN_URL = DEBUG or os.environ.get('ALLOW_ADMIN_URL', False)`

This also assumes false if ALLOW_ADMIN_URL is unset, which seems safer.

kave: this is true @rosskarchner i've refactored it

PR #1918, opened 2016-04-18
resolved request param validation by wrapping in a form

kave:
@rosskarchner 

created the form, and included some missing imports that we're likely breaking that [url](http://localhost:8000/save-hud-counselors-list/?zip=22153)

i'll verify on the build site, when this is merged


rosskarchner:
👍 👍

PR #1928, opened 2016-04-19
django admin now uses wagtail user creation, consistent with change password

kave:
@kurtw @richaagarwal @rosskarchner 

### Testing
- Log into Django-Admin and create a new user


richaagarwal:
👍

PR #1934, opened 2016-04-20
return None if no viable ancestor is found

kave:
@kurtw @richaagarwal 

### Testing 
- get the latest refresh dump
- observe this [page](http://localhost:8000/policy-compliance/rulemaking/small-business-review-panels/arbitration-agreements/) now loads correctly


sebworks:
👍

kurtw:
`article.html` is using this. Could you provide some checking to make sure that nothing breaks? For example, ancestor is passed into `get_protected_url` but that function doesn't account for a page that is None.

kave:
donzo

PR #1942, opened 2016-04-20
Adds overflow:hidden to body at mobile + iPad fixes

anselmbradford:
## Changes

- Hides overflow at mobile sizes.
- Adds halt and clearTransitions methods for halting in-progress transitions and clearing transitions, respectively.
- Added brackets to single-line if statements, because CodeClimate prefers that.
- Cleanup on mobile mega menu suspend method.

## Testing

- Open iPad Retina simulator portrait orientation.
- Navigate to `flapjack.demo.cfpb.gov`
- Open menu.
- Rotate to landscape.
- Click any menu link.
- Rotate to portrait.
- It should look like:

screen shot 2016-04-20 at 1 28 30 pm

Now, `gulp build` and navigate to `localhost:8000` and repeat the steps above. Off-screen content should not be visible.

screen shot 2016-04-20 at 1 34 41 pm

## Review

- @sebworks 
- @jimmynotjim 
- @KimberlyMunoz 

## Todo

- There's a weird double-closure of the menu on iPad at mobile sizes, but is *kinda* looks like an intentional effect, so I think it'll be okay for launch. I think it's due to the deferred collapse function that was added for multiple clicks on the global search flyout menu.

### DIFF
--- a/cfgov/unprocessed/js/modules/FlyoutMenu.js
+++ b/cfgov/unprocessed/js/modules/FlyoutMenu.js
@@ -53,11 +53,12 @@ function FlyoutMenu( element ) { // eslint-disable-line max-statements, no-inlin
   var _isExpanded = false;
   var _isAnimating = false;
 
-  var _expandTransition;
-  var _collapseTransition;
-  var _expandTransitionMethod;
+  var _expandTransition = null;
+  var _expandTransitionMethod = null;
   var _expandTransitionMethodArgs = [];
-  var _collapseTransitionMethod;
+
+  var _collapseTransition = null;
+  var _collapseTransitionMethod = null;
   var _collapseTransitionMethodArgs = [];
 
   // Binded events.
@@ -303,6 +304,24 @@ function FlyoutMenu( element ) { // eslint-disable-line max-statements, no-inlin
   }
 
   /**
+   * Clear the transitions attached to this FlyoutMenu instance.
+   */
+  function clearTransitions() {
+    var transition = getTransition( FlyoutMenu.EXPAND_TYPE );
+    if ( transition ) { transition.remove(); }
+    transition = getTransition( FlyoutMenu.COLLAPSE_TYPE );
+    if ( transition ) { transition.remove(); }
+
+    _expandTransition = null;
+    _expandTransitionMethod = null;
+    _expandTransitionMethodArgs = [];
+
+    _collapseTransition = null;
+    _collapseTransitionMethod = null;
+    _collapseTransitionMethodArgs = [];
+  }
+
+  /**
    * @param {string} type
    *   (Optional) The type of transition to return.
    *   Accepts 'expand' or 'collapse'.
@@ -310,7 +329,7 @@ function FlyoutMenu( element ) { // eslint-disable-line max-statements, no-inlin
    *   as type-safe constants passed into this method.
    *   If neither or something else is supplied, expand type is returned.
    * @returns {MoveTransition|AlphaTransition}
-   *   A transition instance set on this instance, or undefined if none is set.
+   *   A transition instance set on this instance, or null if none is set.
    */
   function getTransition( type ) {
     if ( type === FlyoutMenu.COLLAPSE_TYPE ) {
@@ -404,6 +423,7 @@ function FlyoutMenu( element ) { // eslint-disable-line max-statements, no-inlin
   this.collapse = collapse;
   this.setExpandTransition = setExpandTransition;
   this.setCollapseTransition = setCollapseTransition;
+  this.clearTransitions = clearTransitions;
   this.getData = getData;
   this.getTransition = getTransition;
   this.getDom = getDom;
--- a/cfgov/unprocessed/js/modules/transition/BaseTransition.js
+++ b/cfgov/unprocessed/js/modules/transition/BaseTransition.js
@@ -26,6 +26,7 @@ function BaseTransition( element, classes ) { // eslint-disable-line max-stateme
   var _transitionEndEvent;
   var _transitionCompleteBinded;
   var _addEventListenerBinded;
+  var _isAnimating = false;
   var _isFlushed = false;
 
   /**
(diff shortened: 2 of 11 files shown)


anselmbradford:
@sebworks with this change we can probably remove the invisibility on the global search and I *think* it'll take care of the Android issues.

sebworks:
ok, testing this now.

sebworks:
👍

jimmynotjim:
Anyone else seeing the second level when opening closing after re-orienting to portrait?

sebworks:
Anyone else seeing the second level when opening closing after re-orienting to portrait?

> Which device are you using?

jimmynotjim:
iPad2 in the simulator

anselmbradford:
Yeah, I see the issue. Fix upcoming.

anselmbradford:
Ready for review again

cfgov/unprocessed/js/modules/FlyoutMenu.js
```diff
@@ -53,11 +53,12 @@ function FlyoutMenu( element ) { // eslint-disable-line max-statements, no-inlin
...
 
-  var _expandTransition;
-  var _collapseTransition;
-  var _expandTransitionMethod;
+  var _expandTransition = null;
+  var _expandTransitionMethod = null;
   var _expandTransitionMethodArgs = [];
-  var _collapseTransitionMethod;
+
+  var _collapseTransition = null;
```
sebworks: Why the switch from undefined to `null`?

jimmynotjim:
Hate to be the bearer of yet more bad news, but I'm still seeing the 2nd level when opening/closing the first level after following the steps. Looks like the `aria-expanded` on `.o-mega-menu_content-2` isn't being switched to `false` on rotation.

cfgov/unprocessed/js/modules/FlyoutMenu.js
anselmbradford: I think `null` is more appropriate for these. Each variable here is a thing, whether it is a transition or a placeholder for a transition. They're individual things. `undefined` is global singleton essentially and doesn't imply the variable is a placeholder for an object.

cfgov/unprocessed/js/modules/transition/BaseTransition.js
```diff
@@ -87,10 +88,25 @@ function BaseTransition( element, classes ) { // eslint-disable-line max-stateme
   }
 
   /**
+   * Halt an in-progress animation and call the complete event immediately.
+   */
+  function halt() {
+    if ( !_isAnimating ) { return; }
+    _dom.style.transitionDuration = '0';
```
sebworks: Should we include the prefixes for Moz and Ms?

anselmbradford: Yes. I was going of the transitionEnd event, but these are different.

jimmynotjim:
Must not have had the latest, the 2nd level is now fixed. Waiting on sign-off for an error discussed in chat.

cfgov/unprocessed/js/modules/transition/BaseTransition.js
anselmbradford: https://github.com/cfpb/cfgov-refresh/pull/1942/commits/fd6bcb33afb19bac39e133c2db6ed407db094433

jimmynotjim:
👍 Nice work sticking it out on this, it definitely wasn't an easy one.

cfgov/unprocessed/js/modules/FlyoutMenu.js
sebworks: Declaring a variable by using `var a` doesn't assign it to a global singleton (window.undefined). JS is a weakly typed language so there is no implication of the type of data the variable will hold aside from naming conventions.  I would only assign variable to null when the default type expected is null i.e when querying a for a dom element that might not be found.

sebworks:
👍

cfgov/unprocessed/js/modules/FlyoutMenu.js
anselmbradford: Hmm, undefined vars actually do get set to the global singleton https://jsfiddle.net/xLc6fkze/

So the original problem I ran into was our linter prevents use of undefined, because I presume `_collapseTransition = undefined` would be setting a value (albeit undefined) to a variable that should not have a value. An undefined value on a variable should imply it's never been initialized before with a value. Since the transition can be added and removed, the default value needs to be reset. So I'm explicitly giving it a default value, similar to how `_collapseTransitionMethodArgs` has a default of an empty array. If I didn't initialize it as `null`, it could return three possible values: `null`, `undefined`, or an actual transition object, so setting it to null removes the ambiguity of having it return either undefined or null. I imagine code like dom querying is doing something similar behind the scenes.

sebworks: So why not just use https://github.com/cfpb/cfgov-refresh/blob/1ee49f2425f279a56aa85ce817024a989de494d3/cfgov/unprocessed/js/modules/util/standard-type.js#L25?



> Hmm, undefined vars actually do get set to the global singleton https://jsfiddle.net/xLc6fkze/


You are talking about a primitive value and not a singleton. 

From the ECMAScript spec:

> 4.3.9 undefined value
> 
> primitive value used when a variable has not been assigned a value
> 
> 4.3.10 Undefined type
> 
> type whose sole value is the undefined value
> 
> 4.3.11 null value
> 
> primitive value that represents the intentional absence of any object value
> 
> 4.3.12 Null type
> 
> type whose sole value is the null value

anselmbradford: Ah! I missed that we had undefined in standard-type. Sure, that works. I just didn't want three possible states. So what would be an appropriate use of null in authoring? From the ES spec I'm not seeing how it doesn't fit as a placeholder in this scenario.

anselmbradford: https://github.com/cfpb/cfgov-refresh/pull/1975

PR #1947, opened 2016-04-21
Fixed the mobile menu for on-demand django pages

jimmynotjim:
Due to the `.o-header` scoping, some base styles were overwriting our mega-menu styles. By scoping the mega-menu to the header organism I was able to match the specificity and set the cascade properly.

## Changes

- Wrapped the mega-menu import in the header organism to match the scoping.

## Testing

- `gulp styles:ondemand` and open http://localhost:8000/test-fixture/?atomic=header in a smaller browser window. The left side of the nav should have the proper gutter instead of being pushed up agains the window as we're currently seeing on `/complaint`

## Review

- @anselmbradford 
- @sebworks 

## Screenshots

__Before__

screen shot 2016-04-20 at 7 22 26 pm

__After__

screen shot 2016-04-20 at 7 22 21 pm

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/header.html
+++ b/cfgov/jinja2/v1/_includes/organisms/header.html
@@ -56,7 +56,7 @@
     {% import 'molecules/global-eyebrow.html' as global_eyebrow with context %}
     {{ global_eyebrow.render( true ) }}
 
-    <div class="o-header_content">
+    <div class="o-header_content no-js">
 
         <div class="wrapper">
             {% import 'molecules/global-header-cta.html' as global_header_cta with context %}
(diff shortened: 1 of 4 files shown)


sebworks:
👍

cfgov/jinja2/v1/_includes/organisms/header.html
```diff
@@ -95,6 +95,11 @@
         </div>
 
     </div>
+    <script>
+      /* TODO: scope no-js to individual molecules/organisms. This is a hack to fix no-js on the on-demand header, while also allowing the menu to be scoped to o-header for on-demand pages. */
```
anselmbradford: This could already go into the Header.js init function, but it's fine here with the TODO too.

anselmbradford:
👍

KimberlyMunoz:
This will require an update to cfgov-sheer-templates. I'd merge them in at the same time if possible to avoid any weirdness from changing the markup and not having the styles.

cfgov/jinja2/v1/_includes/organisms/header.html
jimmynotjim: Like before, I'd like to make it more obvious we should remove it as soon as possible.

Scotchester:
We need to get this done ASAP, because, per @KimberlyMunoz's observation, I'll need to make another cfgov-sheer-templates fix and then update the 4 or 5 sites that use that before we can freeze.

PR #1949, opened 2016-04-21
Related Posts now use category to filter out related content

kurtw:
Creates a lookup function in ref.py and then uses it to create the related posts query in base.py.

@kave @rosskarchner


kave:
👍

PR #1950, opened 2016-04-21
Removed related posts from sidebar of imported Newsroom items

kurtw:
After this gets merged in the import-data script needs to be run for these pages with the `overwrite` flag.

@kave @rosskarchner


kave:
verified related posts are gone

PR #1957, opened 2016-04-21
Added the Digital Gov search script

jimmynotjim:
Including this script allows Digital Gov to generate our global site search results. Do not edit, this script comes straight from Digital Gov - http://search.digitalgov.gov/manual/code.html

## Additions

- Digital Gov search loader script

## Testing

- navigate to any new page and inspect the `head`. Just after the IE conditional and jQuery loader should be two search.usa.gov scripts injected by this loader script in the footer.

## Review

- @rosskarchner 
- @sebworks 
- @anselmbradford 

## Screenshots

screen shot 2016-04-21 at 8 28 23 am

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,8 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Added unit test specs for all files to test (excluding config, polyfills and jQuery plugins).
 - Added no-js and js classes to the on-demand header.
 - Added link to Livestream FAQ.
-- Flag for database routing for content.consumerfinance.gov
+- Flag for database routing for content.consumerfinance.gov.
+- Added the Digital Gov search script.
 
 ### Changed
 
--- a/cfgov/jinja2/v1/_includes/on-demand/footer.html
+++ b/cfgov/jinja2/v1/_includes/on-demand/footer.html
@@ -6,3 +6,14 @@
 
 {% import 'organisms/footer.html' as footer with context %}
 {{ footer.render() }}
+
+<script type='text/javascript'>
+//<![CDATA[
+    var usasearch_config = { siteHandle: 'cfpb' };
+
+    var script = document.createElement( 'script' );
+    script.type = 'text/javascript';
+    script.src = 'http://search.usa.gov/javascripts/remote.loader.js';
+    document.getElementsByTagName( 'head' )[0].appendChild( script );
+//]]>
+</script>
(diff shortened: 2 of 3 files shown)


sebworks:
Is this already included on the non V1 pages?

jimmynotjim:
@sebworks According to @rosskarchner each project has implemented it on their own so we only need to worry about pages generated by us.

sebworks:
👍

rosskarchner:
well, more accurately, the code is still in cfpb_nemo and the django base templates in cfpb_common.

Paging @Scotchester though-- this code doesn't seem to be in the cfgov-sheer-templates repo and it's downstream projects. This isn't a launch issue, but should be fixed at some point.

rosskarchner:
also, 👍

Scotchester:
@rosskarchner @anselmbradford Is it possible to include this in the on-demand `header.js` build?

jimmynotjim:
@Scotchester I'd put it in the footer instead. I can add that to this PR if you want before you push the other on-demand bits

Scotchester:
@jimmynotjim That'd be great, but why footer instead of header?

jimmynotjim:
@Scotchester the loader script injects the other scripts that it gets from search.usa.gov into the `head`, so I'd rather not block page render waiting for that reply.

Scotchester:
@jimmynotjim `header.js` doesn't go in the head, it gets imported into a project's combined `main.min.js`, which goes in at the end of `body`.

Scotchester:
But are we sure this needs to be added to cfgov-sheer-templates? Search seems to be working fine for me on OAH and KBYO, as two examples.

jimmynotjim:
I don't think we should be bundling this in our compiled script files, since it's generated for us.

Scotchester:
@jimmynotjim Mmm, I guess that makes sense.

Still not sure if it's needed at all. See above comment about search working already on Sheer pages.

Scotchester:
Also, existing Sheer sites (which have definitely not had this code implemented before), are being indexed already: http://search.consumerfinance.gov/search?utf8=✓&sc=0&query=Know+Before+You+Owe&m=&affiliate=cfpb&commit=Search

rosskarchner:
I think it gets indexed more proactively if it has the script-- and it's how we get counted at https://analytics.usa.gov/

Scotchester:
Oh that's interesting, I didn't know the two were related.

Okay, :shipit:

PR #1958, opened 2016-04-21
Add view more back in

kurtw:
I think this bit of code got incorrectly rolled back in a rebase or something.

@kave @rosskarchner @richaagarwal

### DIFF
--- a/cfgov/jinja2/v1/_includes/molecules/related-posts.html
+++ b/cfgov/jinja2/v1/_includes/molecules/related-posts.html
@@ -39,13 +39,7 @@
                 {{ _related_posts_list(post_type_list, block.value.limit) }}
             </div>
         {% endfor %}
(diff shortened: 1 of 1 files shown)


rosskarchner:
where should I be able to test this? I'm not seeing it on blog posts, for instance

screen shot 2016-04-21 at 11 06 34 am

schaferjh:
Not seeing them on Blog is a seperate issue: I think @kurtw has another fix up for that, but now I can't find it in the list of PRs...

While that is happening, this module shows up on all the reports pages under data-research, so you could look there.

kurtw:
@rosskarchner go to any of the reports http://localhost:8000/data-research/research-reports

kave:
verified `view_more` display and routes to the correct activity-log with filter 👍

rosskarchner:
Looks good

PR #1960, opened 2016-04-21
Fixes up the related posts again to correctly include category based …

kurtw:
This should fix it up well now. It applies a descendant_of query for each parent of the types of data we want to pull in, as well handles filtering the categories that are selected for each query as well.

## Changes

- Refactored related_posts query for category handling...again

## Testing

- Go to http://localhost:8000/data-research/research-reports/issue-brief-social-security-claiming-age-and-retirement-security/ and see that related posts pull in Data, research & reports related content. If you change the category selection, the results change with it.

## Review

- @kave 
- @richaagarwal 
- @rosskarchner 

## Notes

- also fixes the return value of a ref.py function

### DIFF
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -141,27 +141,37 @@ class CFGOVPage(Page):
         return get_protected_url({'request': request}, activity_log) + '?' + tags
 
     def related_posts(self, block, hostname):
+        from . import AbstractFilterPage
         related = {}
+        queries = {}
         query = models.Q(('tags__name__in', self.tags.names()))
-        if block.value['specific_categories']:
+        if [c for c in block.value['specific_categories'] if c]:
             categories = ref.related_posts_category_lookup(block.value['specific_categories'])
             query &= Q(('categories__name__in', categories))
-        # TODO: Add other search types as they are implemented in Django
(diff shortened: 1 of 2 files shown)


rosskarchner:
I'm not sure it's working for me:

screen shot 2016-04-21 at 10 58 48 am

schaferjh:
Yeah, the page in your screenshot should definitely have News items as well as Blog items, @rosskarchner

kurtw:
@rosskarchner @schaferjh this is the expected output. The posts listed on the right are pages that both have the tags of the current page and the category selected. The only categories we have are in Newsroom and Blog. The category for that page is `Data, research & reports` which is a Blog category so no Newsroom items will show.

schaferjh:
No, that's not how this was working previously @kurtw. The restriction only applied to the type the restriction comes from. So if I did a restriction for blogs, but had news checked, I was still getting news content.

kurtw:
Apparently, the category should only affect where it lives.

kurtw:
@rosskarchner @kave @richaagarwal @schaferjh Updated to reflect behavior change.

rosskarchner:
mmm, that's some nice further reading 👍 

screen shot 2016-04-21 at 12 30 38 pm

PR #1962, opened 2016-04-21
Android fix - Remove global search invisibility

anselmbradford:
Remove global search invisibility to make search work on Android 4.

## Changes

- Moves global search content out of view when aria-expanded is false.
- Removes logic for setting/removing invisibility on global search content.

## Testing

- `gulp build` and check on Android 4 in Sauce. Search should be clickable.

## Review

- @sebworks 
- @jimmynotjim 

## Screenshots

screen shot 2016-04-21 at 10 59 07 am


anselmbradford:
Turns out iPad ignores the overflow rules on body 😭

PR #1963, opened 2016-04-21
concantenation of lists of posts when the form id is the same

kave:
@kurtw @richaagarwal 

this [page](http://localhost:8000/policy-compliance/notice-opportunities-comment/) was not getting the latest posts since it was overriding the 2nd browse filterable page since the form is was the same in both `0`


kurtw:
👍

PR #1965, opened 2016-04-21
our final battle

kave:
@rosskarchner @kurtw @richaagarwal 

url had to be relative to the localhost, it had problems resolving the `build` hostname


rosskarchner:
image

PR #1966, opened 2016-04-21
Stop archive pages from being included in FilterablePage lists

kurtw:
This is a terrible way to handle this and should be changed after launch. Excludes pages from being included in the `get_descendants` and `descendant_of` queries if their parent's title has `archive` (case insensitive) in it.

@kave


kave:
👍

PR #1967, opened 2016-04-21
Delays hover state on menu

anselmbradford:
## Changes

- Delays hover state on header.

## Testing

- `gulp build`
- if quickly moving over the menu it should not open.
- When over the menu item it should open after a short 150ms delay.
- Check in iPad simulator

## Review

- @sebworks 
- @jimmynotjim 
- @KimberlyMunoz 

## Todos

- iPad simulator (Apple) has a slide down transition the first time each menu is opened, after which it works as expected where when switching between menus there is no slide down transition.

### DIFF
--- a/cfgov/unprocessed/css/organisms/mega-menu.less
+++ b/cfgov/unprocessed/css/organisms/mega-menu.less
@@ -425,6 +425,7 @@
             }
 
             &-list {
+                display: inline-block;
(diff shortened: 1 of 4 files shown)


Scotchester:
Works as advertised :+1:

jimmynotjim:
👍

PR #1971, opened 2016-04-21
Fixes disappearing search close button

anselmbradford:
Fixed disappearing search close button when swapping device orientation.

## Changes

- Makes trigger invisible when `aria-expanded` is `true` at desktop sizes (via CSS instead of JS).

## Testing

- `gulp build`
- open in iPad simulator.
- in landscape, open search.
- rotate to portrait. Search close button should be visible.

## Review

- @jimmynotjim 
- @sebworks 
- @KimberlyMunoz


sebworks:
👍

jimmynotjim:
👍

PR #1979, opened 2016-04-22
Add rest of models

kurtw:
Though I don't think these are entirely necessary, this adds more models that use the auth user.

@rosskarchner @richaagarwal @kave


kave:
👍

rosskarchner:
I think we're past the need for this, at this point.

kurtw:
No need for this anymore

PR #1999, opened 2016-04-26
Make further reading and list filter results distinct

kurtw:
Duplicate results are showing up in the Further Reading sections of pages. The same type of queries are happening on the filterable lists, so it's possible that it's happening there (though not observed yet).

Also, `descendant_of_q` is changed to `child_of_q`to prevent undesired results.

## Additions

- `.distinct()` to queries to return unique rows in the results

#### Before checking out the branch

- Go into wagtail to make a new page
- Add a Related Posts molecule to the Sidefoot with specified categories `Data, research, & reports` and `Press release`
- Add the tags to the page: `Academic research council`, `Research`, and `Payday loans`
- Publish and go to the page
- View repeated results

#### After checking out the branch

- Go to your page and see unique results

## Review

- @richaagarwal 
- @rosskarchner 
- @kave

### DIFF
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -153,7 +153,7 @@ class CFGOVPage(Page):
         for parent_slug, search_type, search_type_name, search_query in search_types:
             try:
                 parent = Page.objects.get(slug=parent_slug)
-                search_query &= Page.objects.descendant_of_q(parent)
+                search_query &= Page.objects.child_of_q(parent)
                 if 'specific_categories' in block.value:
                     specific_categories = ref.related_posts_category_lookup(block.value['specific_categories'])
                     choices = [c[0] for c in ref.choices_for_page_type(parent_slug)]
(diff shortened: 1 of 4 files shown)


cfgov/v1/models/base.py
```diff
@@ -153,7 +153,7 @@ def related_posts(self, block, hostname):
         for parent_slug, search_type, search_type_name, search_query in search_types:
             try:
                 parent = Page.objects.get(slug=parent_slug)
-                search_query &= Page.objects.descendant_of_q(parent)
+                search_query &= Page.objects.child_of_q(parent)
                 if 'specific_categories' in block.value:
```
richaagarwal: was `descendant_of_q` returning the parent in its results?

kurtw: no, it would return archived pages which is generally something we want to avoid unless we specifically want them

richaagarwal: oh okay, I'm noticing that it's still returning the parent then. http://localhost:8000/about-us/blog/toddlers-to-teens-how-to-kick-start-your-childs-saving-habits/

kurtw: that page didn't have any related posts, but when i added them i didn't see the parent included. by default, those queries exclude it. could you send a screenshot?

richaagarwal: screen shot 2016-04-26 at 11 38 40 am

kurtw: Oh, this isn't the parent but the current page. I'll add in an exclude clause

kurtw: Fixed!

richaagarwal:
👍

rosskarchner:
👍

PR #2000, opened 2016-04-26
Update refresh from flapjack branch

rosskarchner:


Scotchester:
Is all of this stuff good to go to production today with my reg commenting stuff?

PR #2002, opened 2016-04-26
Adding Fix for GTM on subnav items

sebworks:
Adds GTM ignore attribute to subnavs items in Jinja 

## Review

- @Scotchester 

## Notes

We currently add the ignore attribute via JS but unsure why.

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
+++ b/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
@@ -66,7 +66,8 @@
                   {{ _classes( nav_depth, '-link__has-children' ) if nav_children | count > 0 else '' }}
                   {{ _classes( nav_depth, '-link__current' ) if nav_url == request.path else '' }}"
            {{ '' if nav_url == '' else 'href=' + nav_url | e }}
-           {{ 'data-js-hook=flyout-menu_trigger' if nav_children | count > 0 else '' }}>
+           {{ 'data-js-hook=flyout-menu_trigger' if nav_children | count > 0 else '' }}
+           data-gtm_ignore="true">
             {{ nav_caption }}
         </a>
         {% if nav_children | count > 0 %}


Scotchester:
> We currently add the ignore attribute via JS but unsure why.

No removal of this JS code?

anselmbradford:
> We currently add the ignore attribute via JS but unsure why.

It's because there are cases where GTM loads but other JS doesn't. Currently it's used in links that have their default behavior prevented via JS, so when the link's default is prevented, GTM ignore is added. If the default is not prevented, GTM ignore is not added.

Scotchester:
@anselmbradford So we're okay with having it all the time on these subnav links, as done in this PR?

anselmbradford:
> @anselmbradford So we're okay with having it all the time on these subnav links, as done in this PR?

What's the context here? It should already be [added](https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/unprocessed/js/modules/FlyoutMenu.js#L97)?

anselmbradford:
Ahh seeing Kelley's email. Since this is happening on specific pages, I think this is a symptom of issues in the structure of those pages. You can inspect the source in `/adult-financial-education/` for example and the GTM ignore is in the source for the menu links. But it also has `Uncaught TypeError: f is not a function` in the console.

sebworks:
@anselmbradford, do we want to have the GTM ignore code added if JS is disabled?

sebworks:
> Ahh seeing Kelley's email. Since this is happening on specific pages, I think this is a symptom of issues in the structure of those pages. You can inspect the source in /adult-financial-education/ for example and the GTM ignore is in the source for the menu links. But it also has Uncaught TypeError: f is not a function in the console.

I'm only seeing the error when resizing. If view the page at mobile I don't see error. I also only see the data ignore attribute on the first level nav items.

anselmbradford:
> @anselmbradford, do we want to have the GTM ignore code added if JS is disabled?

The "no-js treatment" for IE8, for instance, loads GTM but not the other JS. So if it's in the HTML the menu clicks in that browser would not be registered.

> I'm only seeing the error when resizing. If view the page at mobile I don't see error. 

Yeah I'm seeing that too.

> I also only see the data ignore attribute on the first level nav items.

Really? I see it at all levels. Is it only in certain states that you don't see it?

sebworks:
You seen the data attribute on the `adult-financial-education` link at mobile widths?

sebworks:
> The "no-js treatment" for IE8, for instance, loads GTM but not the other JS. So if it's in the HTML the menu clicks in that browser would not be registered.

I see your point but if JS is disabled or there is a JS error then unwanted events get registered.

sebworks:
> Really? I see it at all levels. Is it only in certain states that you don't see it?

yeah, I'm seeing the attribute. I even see the attribute after resizing and encountering that error.

anselmbradford:
> I see your point but if JS is disabled or there is a JS error then unwanted events get registered.

So if JS is completely disabled, GTM would be disabled too (right, it's all JS as far as I know)? If unwanted events are getting registered from a JS error, the underlying error should really be fixed.

sebworks:
> the underlying error should really be fixed

I agree but the header and footer are on sites in which you don't control the JS. Anywho, I'm going to close this PR as I don't think it's the approach we want to take.

PR #2004, opened 2016-04-27
Make name fields optional for reg comment form

Scotchester:
- Also update form layout slightly
- Remove tests that are no longer needed

**Review:**
- @rosskarchner
- @kurtw
- @richaagarwal
- @sebworks
- @amymok


sebworks:
LGTM

PR #2008, opened 2016-04-28
Add tests for external/internal link handling

richaagarwal:
This adds unit tests for the way we handle external and non CFPB links, functionality for which can be found in `cfgov/v1/__init__.py`.  

I removed `cfgov/v1/tests.py` since it was not being used and does not match the structure we're using elsewhere for unit tests.

To test, run `tox` (I don't know of a way that runs the test individually (that actually works), but if you know of a way, please comment!)

@rosskarchner 
@kave 
@kurtw

### DIFF
--- a/cfgov/v1/tests.py
+++ b/cfgov/v1/tests.py
@@ -1,3 +1,0 @@
-from django.test import TestCase
-
-# Create your tests here.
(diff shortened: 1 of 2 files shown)


kave:
👍

rosskarchner:
👍

kurtw:
👍

PR #2015, opened 2016-04-28
Allow init_command to be set by env

richaagarwal:
I need `default_storage_engine` in my setup, and maybe others do too? 

If you need `storage_engine`, it's set to that by default, but double-check `STORAGE_ENGINE` is not being set to anything else in your `.env`

Testing:  Make sure you can still run the app

@kave 
@kurtw 
@rosskarchner


rosskarchner:
Seems to work for me!

PR #2019, opened 2016-04-29
Removing Multiselect from common.js

sebworks:
All sheer pages should be using the `FilterableListControl` so I'm moving the `Multiselect` out of `common.js`. 

## Changes

- Moved code from `common.js` to `FilterableListControl.js`.

## Testing

- Run gulp build and visit the blog. Expand the filter and use the Multiselect. It should function without error

## Review

- @anselmbradford 

## Checklist

### DIFF
--- a/cfgov/unprocessed/js/organisms/FilterableListControls.js
+++ b/cfgov/unprocessed/js/organisms/FilterableListControls.js
@@ -4,6 +4,7 @@
 var atomicHelpers = require( '../modules/util/atomic-helpers' );
 var ERROR_MESSAGES = require( '../config/error-messages-config' );
 var getClosestElement = require( '../modules/util/dom-traverse' ).closest;
+var Multiselect = require( '../molecules/Multiselect' );
 var Notification = require( '../molecules/Notification' );
 var validators = require( '../modules/util/validators' );
 
@@ -44,6 +45,8 @@ function FilterableListControls( element ) {
   function init( ) {
     _notification = new Notification( _dom );
     _notification.init();
+
+    atomicHelpers.instantiateAll( 'select[multiple]', Multiselect );
 
     _initEvents();
   }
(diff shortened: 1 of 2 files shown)


anselmbradford:
LGTM 👍

sebworks:
@anselmbradford , how do you feel about having to pass in a selector? Shouldn't you just be able to call `AtomicHelpers.instantiate( Multiselect )`. If you chose to pass in a collection of elements or a selector then we would use that but if would not then we would use the default selector. If need be we could expose the selector like so:  `Multiselect.getSelector` or `MultiSelect.selector`.

We also need to switch all the atomic elements to use js data hooks; `select[multiple]` is pretty generic.

anselmbradford:
@sebworks Yeah, you're saying make it optional, right? I wouldn't entirely remove the ability to pass in a selector, since it allows it to be scoped to a certain part of the page, but as an optional parameter that seems fine. There is the `BASE_CLASS` constant that's present in the atomic components ([e.g.](https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/unprocessed/js/organisms/ExpandableGroup.js#L19)) that could and probably should be exposed through a getter, like you wrote. For example, in ExpandableGroup where there's this bit:

```
var _domChildren = _dom.querySelectorAll( '.m-expandable' );
```

It should be:

```
// Return constant value from a getter, like `return '.' + BASE_CLASS;`
var _domChildren = _dom.querySelectorAll( Expandable.getSelector() );
```

That way the Expandable only ever knows what selector it needs and that knowledge isn't bled into other components.

jimmynotjim:
How does that look when we move it outside the project and into it's own "plugin"? I tried to keep it as least opinionated as possible for that reason.

sebworks:
> How does that look when we move it outside the project and into it's own "plugin"? I tried to keep it as least opinionated as possible for that reason.

It just means that the atomic JS elements will have default hooks that you can override.

PR #2020, opened 2016-04-29
refactored form to use USZipCodeField() inorder to not strip leading zeros

kave:
@rosskarchner @kurtw 

This resolves issues with people entering correct zipcodes with leading `0`'s but pdfreactor was stripping them on the backend.


rosskarchner:
👍

PR #2022, opened 2016-04-29
Reg comment test fixes

rosskarchner:
I couldn't get these two tests to pass. At least on my machine, it turned out to be necessary to patch django.conf.settings instead of cfgov.settings.base.

@kurtw @richaagarwal @kave


richaagarwal:
Nice, I had this problem too. 👍

rosskarchner:
I'm gonna merge this.

PR #2025, opened 2016-05-02
[DFJR-1323] Remove choice validator from topics and authors for view more link

kurtw:
View More links contain tags that are not valid for the Activity Log that it links to, breaking the results of the filterable list. That prevents Django from invalidating the form for these topics and/or authors. So, this PR changes the fields used for topics and authors with fields that validate the selection based off the list of any topics or author on the site. This prevents erroneous and potentially dangerous parameters from being passed in, while letting us keep our logic to generate a view more url simple.

## Additions

- MultipleChoiceField that validated against the list of all tags/authors instead of just the ones listed

## Testing

- Go to `http://www.consumerfinance.gov/activity-log/?filter0_topics=Regulation+V&filter0_topics=Credit+report&filter0_topics=Regulations` and see the error and no results
- Pull down branch code
- Go to `http://localhost:8000/activity-log/?filter0_topics=Regulation+V&filter0_topics=Credit+Reporting&filter0_topics=Regulations` and see results returned but only Regulations selected for topics
- Go to `http://localhost:8000/activity-log/?filter0_topics=Regulation+V&filter0_topics=Credit+Reporting&filter0_topics=Regulations&filter0_topics=blahblahblah` and see the error.

## Review

- @richaagarwal 
- @rosskarchner 
- @kave 

## Todos

- This probably isn't the _best_ way of handling this, so later we should pull the logic in the form to set all of the topics/authors, queries, page set, etc into a helper file.

### DIFF
--- a/cfgov/v1/forms.py
+++ b/cfgov/v1/forms.py
@@ -161,6 +161,18 @@ class CalenderPDFFilterForm(forms.Form):
             raise forms.ValidationError(ERROR_MESSAGES['DATE_ERRORS']['one_required'])
         return cleaned_data
 
+class MultipleChoiceField(forms.MultipleChoiceField):
+    def validate(self, value):
+        tags = Tag.objects.all().values_list('name', flat=True)
+        for tag in value:
+            if tag not in tags:
+                raise ValidationError(
+                    self.error_messages['invalid_choice'],
+                    code='invalid_choice',
+                    params={'value': tag},
+                )
+
+
 
 class FilterableListForm(forms.Form):
     title_attrs = {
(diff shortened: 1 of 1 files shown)


kave:
so is `Regulation V` valid or not?

kurtw:
So, `Regulation V` isn't _actually valid_ for Activity log but we don't show an error for it because it doesn't affect the output of what _is_ valid. The reason we're doing this is because View more links are generated with the current page's tags, regardless of it any particular tag is available in the Activity log. 

Before you mention it, **yes** it makes more sense to how the View more link is generated. And we will! Just not now. There's code that already does this but it's embedded in the form logic, so that's why there's a TODO to make sure something is done about it.

kurtw:
Actually, I'm moving forward with testing the filter which will give us the opportunity to do this in the right way. Closing for future PR on testing and refactoring filters.

PR #2026, opened 2016-05-03
dbrouter shortcut function for model lookup by contenttype

kave:
### Testing
- Ensure you have the legacy db setup with a dump from content
- Clone knowledgebase app as a sibling to cfgov-refresh
 - build frontend assets `./setup.sh`
- cd back into cfgov-refresh 
 - `pip install -e ../knowledgebase`
 - `pip install simplejson`
 -  `./cfgov/manage.py update_index knowledgebase`
- Enter django admin and select a [spanish answer](http://localhost:8000/django-admin/knowledgebase/spanishanswer/1945/?_changelist_filters=p%3D9) 
- Then click `view on site` site and see then site should render
screen shot 2016-05-03 at 10 18 04 am

@rosskarchner 
@kurtw


cfgov/legacy/views.py
```diff
@@ -66,3 +75,92 @@ def generate_pdf(self):
...
+        get_absolute_url = obj.get_absolute_url
+    except AttributeError:
+        raise http.Http404(_("%(ct_name)s objects don't have a get_absolute_url() method") %
+                           {'ct_name': content_type.name})
+    absurl = get_absolute_url()
+
+    # Try to figure out the object's domain, so we can do a cross-site redirect
+    # if necessary.
+
+    # If the object actually defines a domain, we're done.
```
rosskarchner: I'm not sure this stuff below is necessary-- we can trust that once someone clicks 'view on site', it will be on the same host as the current django-admin session

kave: @rosskarchner alright i'll simplify it.

kurtw:
Nice 💯

rosskarchner:
👍

PR #2030, opened 2016-05-03
Removes unused parameter

anselmbradford:
## Changes

- Removes unused parameter from `get_secondary_nav_items` function.

## Review

- @richaagarwal 
- @kave 
- @kurtw

### DIFF
--- a/cfgov/v1/util/util.py
+++ b/cfgov/v1/util/util.py
@@ -89,7 +89,7 @@ def instanceOfBrowseOrFilterablePages(page):
 
 
 # For use by Browse type pages to get the secondary navigation items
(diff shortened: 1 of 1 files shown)


richaagarwal:
Seems fine to remove, but I would defer to @kave since he introduced it in
this commit:
https://github.com/cfpb/cfgov-refresh/commit/bcff9984e2547dfad2012269a0f40802163770b2

kurtw:
This can be removed. [This line](https://github.com/cfpb/cfgov-refresh/pull/2030/files#diff-a908c7015d65f8eb2bca865d03f8485aR117) does the job of the parameter in question.

PR #2031, opened 2016-05-03
Move V1 modules

kurtw:
This moves v1 modules around to places that make a bit more sense. This is a preliminary step in refactoring our code for testing purposes.

## Additions

- `v1/atomic_elements` module

## Changes

- atomic elements live in the `v1/atomic_elements` module

## Testing

- Pull down and visit the various page models to make sure they can still pull in the right code.
- Also if it just flat out won't run the server then there's something up

## Review

- @kave 
- @rosskarchner 
- @richaagarwal


richaagarwal:
👍 but is there a reason they don't live in `/models/atomic_elements` ?

kurtw:
Because they aren't actually models. They're classes used by a model field to generate HTML from the field's value, among other things.

richaagarwal:
Ah ok 👍

rosskarchner:
👍

kave:
👍  after conflicts resolving

PR #2034, opened 2016-05-04
Fixes linter errors

anselmbradford:
## Changes

- Fixes linter errors and warnings

## Testing

- `gulp lint` should have no errors.

## Review

- @jimmynotjim 
- @sebworks

### DIFF
--- a/cfgov/unprocessed/js/molecules/Multiselect.js
+++ b/cfgov/unprocessed/js/molecules/Multiselect.js
@@ -107,7 +107,7 @@ function Multiselect( element ) { // eslint-disable-line max-statements, inline-
     for ( var i = 0, len = list.length; i < len; i++ ) {
       item = list[i];
 
-      // If the value isn't valid kill the script and propt the developer
+      // If the value isn't valid kill the script and prompt the developer.
       if ( !strings.stringValid( item.value ) ) {
         console.log( '\'' + item.value + '\' is not a valid value' );
 
@@ -241,8 +241,8 @@ function Multiselect( element ) { // eslint-disable-line max-statements, inline-
    * @param   {string} value The value of the option the user has chosen
    */
   function _updateSelections( value ) {
-    var optionIndex
-      = arrayHelpers.indexOfObject( _optionsData, 'value', value );
+    var optionIndex =
+      arrayHelpers.indexOfObject( _optionsData, 'value', value );
     var option = _optionsData[optionIndex] || _optionsData[_index];
 
     if ( option ) {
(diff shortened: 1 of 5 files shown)


cfgov/unprocessed/js/molecules/Multiselect.js
```diff
@@ -107,7 +107,7 @@ function Multiselect( element ) { // eslint-disable-line max-statements, inline-
     for ( var i = 0, len = list.length; i < len; i++ ) {
       item = list[i];
 
-      // If the value isn't valid kill the script and propt the developer
+      // If the value isn't valid kill the script and prompt the developer.
       if ( !strings.stringValid( item.value ) ) {
         console.log( '\'' + item.value + '\' is not a valid value' );
```
anselmbradford: This should probably throw an error instead of a console log.

sebworks: Can we change the name from `strings` to `stringHelpers`?

anselmbradford: Sounds good to me.

sebworks:
👍

cfgov/unprocessed/js/molecules/Multiselect.js
jimmynotjim: NOOOOO! The problem was that errors were stopping all JS on the page, which is why the checker was added in the first place.

test/browser_tests/spec_suites/organisms/mega-menu.js
```diff
@@ -52,7 +52,8 @@ describe( 'MegaMenu', function() {
         } );
       } );
 
-      describe( 'when mouse moves from one link to another after a delay', function() {
+      describe( 'when mouse moves from one link ' +
```
jimmynotjim: Why break a text string up instead of breaking on the comma?

cfgov/unprocessed/js/molecules/Multiselect.js
anselmbradford: Why are these errors showing up again? Shouldn't they be fixed at the source? This relies on the developer actually checking the console, which isn't guaranteed. `console.log` shouldn't be in production code in my opinion. If it should gracefully fail what about throwing an error and handling the error in a `try… catch` block where that error would matter.

test/browser_tests/spec_suites/organisms/mega-menu.js
anselmbradford: Breaking on the comma requires indenting the whole block differently because a function follows the comma, whereas breaking the string handles the line length without further changes to indent. If indenting the block is preferable though, I'm fine with it either way.

PR #2037, opened 2016-05-04
Implements standard-type unit tests

anselmbradford:
## Additions

- Implements standard-type unit tests

## Testing

- `gulp test:unit` should pass

## Review

- @sebworks 
- @jimmynotjim

### DIFF
--- a/test/unit_tests/modules/util/standard-type-spec.js
+++ b/test/unit_tests/modules/util/standard-type-spec.js
@@ -7,5 +7,15 @@ var expect = chai.expect;
 var standardType = require( BASE_JS_PATH + 'modules/util/standard-type' );
 
 describe( 'standard-type', function() {
-  // TODO: Implement tests.
+  it( 'should include a standard JS data hook', function() {
+    expect( standardType.JS_HOOK ).to.equal( 'data-js-hook' );
+  } );
+
+  it( 'should include a non operational function', function() {
+    expect( standardType.noopFunct() ).to.be.undefined;
+  } );
+
+  it( 'should include a standard undefined reference', function() {
+    expect( standardType.UNDEFINED ).to.be.undefined;
+  } );
 } );


test/unit_tests/modules/util/standard-type-spec.js
```diff
@@ -7,5 +7,15 @@ var expect = chai.expect;
 var standardType = require( BASE_JS_PATH + 'modules/util/standard-type' );
 
 describe( 'standard-type', function() {
-  // TODO: Implement tests.
+  it( 'should include a standard JS data hook', function() {
+    expect( standardType.JS_HOOK ).to.equal( 'data-js-hook' );
+  } );
+
+  it( 'should include an filler function', function() {
+    expect( standardType.noopFunct() ).to.be.undefined;
```
sebworks: should we test that `noopFunct` is a function?

sebworks:
👍

test/unit_tests/modules/util/standard-type-spec.js
anselmbradford: I don't think we need to, if the function does not exist on standardType it'll throw an error that it doesn't exist. And you can't call anything but functions with `()`.

sebworks: ok, I would change the language to read `should include a non operational function`

jimmynotjim:
👍

test/unit_tests/modules/util/standard-type-spec.js
anselmbradford: Updated in https://github.com/cfpb/cfgov-refresh/pull/2037/commits/2350084f2114e1c6a4dfcad4f0e25ab1d0f60cb8 .. though technically the test isn't checking if it's non operational, just that it returns undefined. It could be doing something, but I don't think it's one worth setting up a spy for, given its name and location.

PR #2039, opened 2016-05-04
Removes validators complexity warning

anselmbradford:
## Changes

- Creates private `_checkableInput()` method that both `checkbox()` and `radio()` share. This removes the public exposure of the `type` parameter, since it would have been possible to call `checkbox()` and pass in a type that was neither `checkbox` nor `radio`. Also moves variable default checks to the top of the method.
- Minor updates to jsdocs to add trailing periods, further explain return type, and adjust line lengths.

## Testing

- Not 100% sure where this is used, but the `validators-spec.js` should pass with `gulp test:unit`.
- `gulp lint` should no longer have a complexity warning for `validators.checkbox()` method.

## Review

- @jimmynotjim 
- @sebworks


jimmynotjim:
It's used in the FilterableControls 👍 on the fix

PR #2040, opened 2016-05-04
Removes dom-traverse complexity warning

anselmbradford:
## Changes

- Moves prefixed matches() method lookup to private `_getMatchesMethod`.
- Adds inline if statement brackets to satisfy Code Climate.

## Testing

- `gulp lint` should not show a complexity warning for `dom-traverse.js`.

## Review

- @sebworks 
- @jimmynotjim


jimmynotjim:
👍

sebworks:
👍

PR #2042, opened 2016-05-04
Adds aria-pressed to primary flyout menu trigger

anselmbradford:
## Changes

- Adds aria-pressed to primary flyout menu trigger to prepare expandables to share transition code.

## Testing

- `gulp test:unit` should pass.
- When inspecting mega menu hamburger menu at mobile, the trigger to show/hide the menu should toggle `aria-pressed` from false to true. The submenu links should as well. The back button _should not_, but maybe it should also? It's not clearly a toggle, but neither are the submenu links.

## Review

- @jimmynotjim 
- @sebworks


jimmynotjim:
👍

sebworks:
👍

PR #2047, opened 2016-05-05
Adds print styles for major site features that aren't print applicable

anselmbradford:
## Additions

- Adds print styles for major site features that aren't print applicable.
- Moves the logo on print so it is aligned correctly.
- Adds missing topdoc EOF comments.

## Testing

- `gulp build` visit the homepage and go to File > Print and look at the preview. Should not include the mega menu, search, hero, social media icons, banner and footer buttons, or gaps for images that are background images.
- Check /about-us/careers/application-process/ for the hidden breadcrumbs and secondary nav too.

## Review

- @jimmynotjim 
- @sebworks 
- @schaferjh 

## Screenshots

screen shot 2016-05-05 at 12 48 32 pm

## Todos

- Heros in CF need print styles added to hide areas where background images appear.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 ## Unreleased
 
 ### Added
+- Frontend: Added print styles to hide major site features that aren't
+  print applicable.
 
 ### Changed
 
--- a/cfgov/unprocessed/css/breadcrumbs.less
+++ b/cfgov/unprocessed/css/breadcrumbs.less
@@ -27,3 +27,10 @@
         }
     }
 }
+
+// Hide on print.
+.respond-to-print( {
+    .breadcrumbs {
+        display: none;
+    }
+} );
--- a/cfgov/unprocessed/css/molecules/call-to-action.less
+++ b/cfgov/unprocessed/css/molecules/call-to-action.less
@@ -33,7 +33,6 @@
     }
 }
 
-
 /* topdoc
     name: EOF
     eof: true
--- a/cfgov/unprocessed/css/molecules/featured-menu-content.less
+++ b/cfgov/unprocessed/css/molecules/featured-menu-content.less
@@ -77,7 +77,6 @@
     } );
 }
 
-
 /* topdoc
     name: EOF
     eof: true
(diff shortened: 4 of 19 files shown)


jimmynotjim:
Any idea why it might look like it's also getting small screen style?

screen shot 2016-05-05 at 1 29 13 pm

anselmbradford:
> Any idea why it might look like it's also getting small screen style?

Because of the width of a piece of paper (minus margin) and printing resolution I assume. 8" x 72dpi = 576px

schaferjh:
> Because of the width of a piece of paper (minus margin) and printing resolution I assume. 8" x 72dpi = 576px

Is that an assumption all printers will make? We're going to confuse people who go to print a page from their spacious desktop browser window and see everything reformatted on the printed page. 

I'm fine going with this for now because in the current situation the mobile menu flyout makes printing almost un-usuable, but I don't think this can be the long term fix if it reformats all pages like this.

jimmynotjim:
@anselmbradford how'd you get it to show correctly for you?

anselmbradford:
To be honest I haven't printed a page in a long time, so this would require some research, but the page size can be set with the `@page` rule, like:

```
@page {
  size: 17in 22in;
}
```

Which at 72dpi fit to 8.5 x 11 would print the page at 144dpi at 8.5 x 11 and look like:

screen shot 2016-05-06 at 5 57 42 pm

However, I'm pretty sure setting that is *not* a best practice unless the printing environment is known and controlled. The resolution can be set on the printer or even on the display (72dpi or 96dpi), so a set value for the page size would not be guaranteed to come out as expected. Visit another responsive site like cnn.com and you'll see that the print preview is the mobile view as well. Do any of you have a printer where you can experiment with setting the resolution higher and seeing that you're getting a desktop version of the site?

jimmynotjim:
Hmm, wonder if we can control the ppi vs the dpi. Pixels do not have to match dots when printing. I'll try playing with this tomorrow.

KimberlyMunoz:
Mine looks similar to Ans's first screenshot and isn't too bad on the blog pages where the content is less graphics-based than the homepage.

jimmynotjim:
Must be some odd setting on this machine. 👍 based on Kim's comment

anselmbradford:
Okay, I'll merge and we can investigate further if @schaferjh's stakeholder contacts aren't happy with the way it's printing on their printers.

sebworks:
👍

PR #2048, opened 2016-05-05
Bad zips should return 400 error

rosskarchner:
This one is pretty subtle.

The old behavior was, we were raising an Exception, and letting Django sort out the rest. This meant, in production, a 500 error. My main concern is that this was polluting the 'Error' rate in New Relic.

Also, I did some reading, and it seems like HTTP 400 is a more appropriate response.

So, when DEBUG is False, we now return a simple failure message with HTTP status 400. We may wish to create a full-fledged error template for 400, like we do 404 and 500-- but users of the /find-a-housing-counselor/ tool will never get this far, so I'm confident the error will mostly be seen by bots.

## Testing

- Hit /save-hud-counselors-list/?zip=9021 and note the exception message you get
- temporarily set DEBUG to False in cfgov/settings/local.py
- Hit /save-hud-counselors-list/?zip=9021 and note the new error message
- If you want to get really fancy, use curl and see the HTTP status:

`curl -I http://localhost:8000/save-hud-counselors-list/?zip=9021`

should look something like this:

```
HTTP/1.0 400 BAD REQUEST
Date: Thu, 05 May 2016 16:50:36 GMT
Server: WSGIServer/0.1 Python/2.7.11
X-Frame-Options: SAMEORIGIN
Content-Type: text/html; charset=utf-8
```

## Review

- @kurtw @kave @richaagarwal 

## Checklist

### DIFF
--- a/cfgov/legacy/views.py
+++ b/cfgov/legacy/views.py
@@ -34,7 +34,7 @@ class HousingCounselorPDFView(PDFGeneratorView):
             url = '%s://%s/%s' % (request.scheme, 'localhost', api_url)
             return url
         else:
-            raise InvalidZipException(form.errors)
+            raise InvalidZipException(form.errors['zip'].as_text())
 
     def get(self, request):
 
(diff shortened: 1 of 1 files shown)


richaagarwal:
Nice 👍

kurtw:
👍

PR #2055, opened 2016-05-05
DFJR-994 Adds initialization flag for atomic components

anselmbradford:
## Changes

- Adds initialization flag for atomic components so that they bail out of their `init` function if it has already been called.

## Testing

- `gulp build` and inspect the page. Any atomic component should have `data-js-hook="[atomic base class] init"` attributes.
- In Wagtail, it should be possible to add an expandable and an expandable group to a page, or expandable and secondary navigation, and they will both still work.

## Review

- @kave
- @kurtw 
- @jimmynotjim 
- @sebworks

### DIFF
--- a/cfgov/unprocessed/js/modules/util/atomic-helpers.js
+++ b/cfgov/unprocessed/js/modules/util/atomic-helpers.js
@@ -6,6 +6,9 @@
 'use strict';
 
 var dataHook = require( './data-hook' );
+var standardType = require( './standard-type' );
+
+var INIT_FLAG = standardType.STATE_PREFIX + 'atomic_init';
 
 // TODO: Update baseClass to baseSel to handle CSS selector instead of a class.
 /**
@@ -70,11 +73,11 @@ function _verifyClassExists( element, baseClass ) {
  *   false otherwise.
  */
 function setInitFlag( element ) {
-  if ( dataHook.contains( element, 'init' ) ) {
+  if ( dataHook.contains( element, INIT_FLAG ) ) {
     return false;
   }
 
-  dataHook.add( element, 'init' );
+  dataHook.add( element, INIT_FLAG );
 
   return true;
 }
@@ -98,7 +101,7 @@ function instantiateAll( selector, Constructor ) {
 
 // Expose public methods.
 module.exports = {
-  checkDom: checkDom,
+  checkDom:       checkDom,
   instantiateAll: instantiateAll,
-  setInitFlag: setInitFlag
+  setInitFlag:    setInitFlag
 };
--- a/cfgov/unprocessed/js/molecules/Expandable.js
+++ b/cfgov/unprocessed/js/molecules/Expandable.js
@@ -5,6 +5,7 @@ var atomicHelpers = require( '../modules/util/atomic-helpers' );
 var breakpointState = require( '../modules/util/breakpoint-state' );
 var EventObserver = require( '../modules/util/EventObserver' );
 var fnBind = require( '../modules/util/fn-bind' ).fnBind;
+var standardType = require( '../modules/util/standard-type' );
 
 /**
  * Expandable
@@ -52,7 +53,9 @@ function Expandable( element ) { // eslint-disable-line max-statements, inline-c
    *   or undefined if it was already initialized.
    */
   function init( state ) {
-    if ( !atomicHelpers.setInitFlag( _dom ) ) { var inst; return inst; }
+    if ( !atomicHelpers.setInitFlag( _dom ) ) {
+      return standardType.UNDEFINED;
+    }
 
     _calcHeight();
     // Even if expanded is set, don't expand if in mobile window size.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,6 +94,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 
 ### Changed
 - Updated event times to show EDT.
+- Frontend: Added init flag when initializing atomic components.
 
 
 ## 3.0.0-3.3.11 - 2016-05-03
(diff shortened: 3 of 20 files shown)


cfgov/unprocessed/js/molecules/Expandable.js
```diff
@@ -48,9 +48,12 @@ function Expandable( element ) { // eslint-disable-line max-statements, inline-c
   /**
    * @param {number} state
    *   Allows passing of EXPANDED flag to set expanded state.
-   * @returns {Object} The Expandable instance.
+   * @returns {Expandable|undefined} An instance,
+   *   or undefined if it was already initialized.
    */
   function init( state ) {
+    if ( !atomicHelpers.setInitFlag( _dom ) ) { var inst; return inst; }
```
sebworks: Why not just use the standard `UNDEFINED` type here?

sebworks:
What do you think about just limiting the scope of the selectors to direct descendants?

`template > o-*, template > m-*, template > a-*` 
`organism > m-*, organism > a-*` 
`molecule > a-*`

anselmbradford:
> What do you think about just limiting the scope of the selectors to direct descendants? The parent would always be responsible for instantiating and initializing it's atomic children.

Maybe? It seems like there could be issues with it scaling. 
Using the expandable/expandable group as an example:
Say you have this:

```
templateA -
  - expandableA
  - expandableB
  - expandableGroupA - 
      - expandableAA
      - expandableAB

templateB -
  wrapper
    - expandableA
    - expandableB
    - expandableGroupA - 
        - expandableAA
        - expandableAB
```

If it can be guaranteed molecule `expandableA` and `expandableB` are direct children of template `templateA`, that'd work, but I'm not sure we can since each template may have different wrappers between it and the expandables to handle layout needs, like shown in `templateB`. Do we know that expandables will always now and in the future be direct children of browse pages and doc detail pages (or other pages)? If there is the need to have a different selector based on page type, the selector would potentially have to adapt in some way depending on the type of page that was being visited. If a wrapper was added in the future and whoever added it didn't know about the expandable scoping it'd silently break on some page types and not others—though that theoretically could be caught in tests.

sebworks:
yeah, we would have to use `not selectors` for the templates like:

`template o-*, template m-*:not(o-* m-*) , template a-*:not(o-* a-*, m-* a-*) `

anselmbradford:
What would the actual selectors look like? I mocked up a pen to see what the possible configurations would be: http://codepen.io/anon/pen/rePVyz but getting an error on the `not`.

sebworks:
@anselmbradford , ok I'm looking at the code pen now.

sebworks:
yep, it won't let you do a complex query within a `not` statement. I could work around this by using the he `not` helper utility or https://developer.mozilla.org/en-US/docs/Web/API/NodeFilter. I'm not sure it's worth it though. What do you think @anselmbradford ?

anselmbradford:
I think the approach we use here would depend on whether we want the atomic components to manage flagging their own initialization state or if there's a global approach further up that manages what gets queried and initialized or not.

Take this bit:
```
template -
  - expandable
  - expandableGroup - 
      - expandable
  - secondaryNavigation - 
      - expandable
  -globalBanner - 
      - expandable
```

A query for `expandable` would need to retrieve expandables when they appear on their own, but not when they appear inside an organism. So the query would either have to exclude expandables that appeared inside things that have an `o-` prefix or list out the organisms (expandableGroup, secondaryNavigation, globalBanner) that should be ignored. The issue I see here is some atomic things may not have the prefix https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/jinja2/v1/_includes/molecules/hero.html#L44, but could theoretically include an expandable, and any additional organisms that came along and decided to include an expandable would need to be added into the query in any context where the page was being queried for expandables.

Hmm, thinking about it further, I'm leaning towards having atomic components manage their own initialization state. For performance reasons, the global query could still be modified to only return standalone expandables, but at least if the query was set too wide, the expandables initialized inside an organism would not get initialized a second time by a stand-alone expandables query.

sebworks:
>> For performance reasons, the global query could still be modified to only return standalone expandables, but at least if the query was set too wide, the expandables initialized inside an organism would not get initialized a second time by a stand-alone expandables query.

I'm not sure it should care about if it's already been initialized for a given dom element. Anywho, lets go with this for now. 👍

cfgov/unprocessed/js/modules/util/atomic-helpers.js
```diff
@@ -56,5 +100,6 @@ function instantiateAll( selector, Constructor ) {
 // Expose public methods.
 module.exports = {
   checkDom: checkDom,
-  instantiateAll: instantiateAll
+  instantiateAll: instantiateAll,
+  setInitFlag: setInitFlag
```
jimmynotjim: Left align your values to appease the linter

cfgov/unprocessed/js/molecules/Expandable.js
jimmynotjim: Ditto this, it's not clear why you're setting the var without a value.

jimmynotjim:
What's you're plan for the empty tests? Coverage has been dropping, I don't want to see us drop too far from when we released.

cfgov/unprocessed/js/molecules/Expandable.js
anselmbradford: Okay, mostly did this because that variable is limited to the if block and not elsewhere and it wouldn't require importing a module to copy this line to new components.

sebworks: ok, I would just add a comment and probably rename it to `var UNDEFINED`;

anselmbradford:
> What's you're plan for the empty tests? 

Filled out the `setInitFlag` ones. The `instantiateAll` is todo'd for another PR.

anselmbradford:
Ready for review again! Thanks!

sebworks:
LGTM

jimmynotjim:
👍

anselmbradford:
@sebworks @KimberlyMunoz @jimmynotjim 
Ready for review again. Updated per our chat in FEWD V1 room.

sebworks:
👍

PR #2059, opened 2016-05-06
Isolates secondary navigation

anselmbradford:
Isolates secondary navigation so that it can be moved to responsive non-v1 pages.

## Additions

- Adds on-demand secondary navigation.

## Changes

- Moves old secondary-navigation and logic into atomic secondary nav and added TODOs so there is only one secondary-navigation.html to refer to.
- Updates `nav_link` to use `render` method.

## Testing

- `gulp build && gulp scripts:ondemand && gulp styles:ondemand`
Secondary navigation should work in the following contexts:
  - wagtail: `/policy-compliance/rulemaking/final-rules/?page=6#o-pagination-content_list`
  - non-wagtail: `/about-us/careers/application-process/`
  - on-demand: `/test-fixture/?atomic=secondary-navigation`

## Review

- @sebworks 
- @jimmynotjim 
- @KimberlyMunoz 

## Screenshots

screen shot 2016-05-06 at 2 35 17 pm

## Notes

- Includes relevant parts of https://github.com/cfpb/cfgov-refresh/pull/2055 while that gets sorted out.
- Used in https://github.com/cfpb/cfgov-sheer-templates/pull/14 and https://github.cfpb.gov/CFGOV/fin-ed-resources/pull/69


sebworks:
Is there a way to cut down on the amount of css that we are including with the footer, header, and side-nav? I don't think the sites that are including these atomic elements expect that they are going to be getting 150 kb of css for a header, footer, and side nav. I think there is a reasonable level of expectation that we have to meet. Per page css?

anselmbradford:
> Is there a way to cut down on the amount of css that we are including with the footer, header, and side-nav? I don't think the sites that are including these atomic elements expect that they are going to be getting 150 kb of css for a header, footer, and side nav. I think there is a reasonable level of expectation that we have to meet. 

Yeah, it's not ideal, but in the current architecture the way around it would be going into each `unprocessed/css/on-demand/*.less` file and manually walking through the imported CSS and picking out the redundant styles and then making sure it still works on the pages that are only header + footer and header + secondary nav + footer. Or perhaps grouping all the styles together and running [uncss](https://github.com/giakki/uncss), but I tried that a little and found it still required manual cleanup. I'm not sure either of those options are worth it versus directly bringing the non-v1 pages affected into v1. They're also currently [buggy](https://jira.demo.cfpb.gov/browse/DFJR-1394) on www. Perhaps @schaferjh has input on the timeline for bringing pages like `/tax-preparer-resources/` into v1?

> Per page css?

This could be an option, so have a less file in `unprocessed/css/on-demand/` that defines a combined header + footer and header + side navigation + footer CSS file. Is this what you mean?

jimmynotjim:
Could we create an on demand page that includes all three, then run something like https://github.com/ben-eb/gulp-uncss ?

jimmynotjim:
Should the on demand example be showing the mobile styles at all browser widths?

screen shot 2016-05-06 at 5 15 17 pm

schaferjh:
> I'm not sure either of those options are worth it versus directly bringing the non-v1 pages affected into v1. They're also currently buggy on www. Perhaps @schaferjh has input on the timeline for bringing pages like /tax-preparer-resources/ into v1?

The pages within Educational Resources are slated to be integrated into Wagtail as they are updated/reviewed by the Consumer Experience project. From what I know about that project, I believe that means leaving them as is for at least another 6 mos or more. 

That timeline assumes they are acceptably functional in their current state and it's not a pain for them to be allowed to exist as-is. The disappearing-side-nav bug @anselmbradford points out is worrisome. Are there other issues with allowing these pages to exist as is? Would it be useful to have a conversation offline about some of this?

anselmbradford:
> Could we create an on demand page that includes all three, then run something like https://github.com/ben-eb/gulp-uncss ?

I did try uncss when I originally built the on-demand header and footer and found it unfavorable in what it produced, requiring a lot of additional manual cleanup, but that said I think it would be worth looking into again.

> Should the on demand example be showing the mobile styles at all browser widths?

I'd think so, but it's grabbing the styles in secondary-navigation.less, which would indicate that the hiding at desktop for the secondary navigation is happening somewhere else, which seems like an error. I can look into where desktop hiding of the secondary nav is defined.

> That timeline assumes they are acceptably functional in their current state and it's not a pain for them to be allowed to exist as-is. The disappearing-side-nav bug @anselmbradford points out is worrisome. Are there other issues with allowing these pages to exist as is?

So in the ramp up to v1 launch, in my mind the global on-demand header and footer were a stop-gap to get non-v1 pages into the appearance/functionality of v1 pages. The JS is isolated, but as pointed out in this thread the CSS is not. That is because the CSS is not totally atomic. The atomic parts of the CSS are isolated, but for example, the Capital Framework parts of the header and the footer are redundant with each other and are each delivered to the page. If this will go to more permanent pages a better solution will be needed for this ideally, as it is being propagated --> https://github.com/cfpb/college-costs/pull/152/files#diff-a6224f3cdcddb48919e4dcaf6b24f540R62

>  Would it be useful to have a conversation offline about some of this?

Perhaps if this will be a permanent fixture in parts of the site. It's a performance issue, for what that's worth.

Scotchester:
I agree with @anselmbradford, I think this is a necessary evil for the time being. I'll make every effort I can to move these pages into Wagtail when I don't have more pressing project work, but I think we need to do this for now.

@sebworks and @jimmynotjim can we proceed with the review of this so we know it's good before translating the work to the sidenav code in cfgov-sheer-templates?

sebworks:
👍

PR #2061, opened 2016-05-06
Added base pagination test before working on pagination bugs

jimmynotjim:
There are no tests for pagination interaction. Before taking on fixing the existing filter/pagination combo bug, I'm adding some baseline tests of the functionality.

## Additions

- pagination organism tests
- filtering/pagination template tests

## Testing

- `gulp test:acceptance --sauce=false --specs=blog.js,organisms/pagination.js,templates/browse-filterable.js`

## Review

- @anselmbradford 
- @sebworks 
- @KimberlyMunoz 

## Notes

- The last test is the reported bug, fix forthcoming.

## Checklist

### DIFF
--- a/test/browser_tests/shared_objects/filter.js
+++ b/test/browser_tests/shared_objects/filter.js
@@ -16,8 +16,11 @@ var filter = {
 
   searchFilterHideBtn: _getFilterElement( '.m-expandable_cue-close' ),
 
-  searchFilterSubmitBtn: _getFilterElement( 'btn[type="submit"]' )
+  searchCategoryLabel: _getFilterElement(
+    'label[for="filter1_categories_at-the-cfpb"]'
+  ),
 
+  searchFilterSubmitBtn: _getFilterElement( 'input[type="submit"]' )
 };
 
 module.exports = filter;
--- a/test/browser_tests/spec_suites/organisms/pagination.js
+++ b/test/browser_tests/spec_suites/organisms/pagination.js
@@ -1,0 +1,39 @@
+'use strict';
+
+var Blog = require(
+    '../../page_objects/page_blog.js'
+  );
+
+describe( 'Pagination', function() {
+  var page;
+
+  beforeEach( function() {
+    page = new Blog();
+    page.get();
+  } );
+
+  it( 'should navigate to the second page', function() {
+    page.paginationNextBtn.click()
+    browser.sleep( 1000 );
+
+    expect( browser.getCurrentUrl() ).toContain( 'page=2' );
+  } );
+
+  it( 'should navigate to the first page', function() {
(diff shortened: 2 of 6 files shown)


test/browser_tests/shared_objects/filter.js
```diff
@@ -16,8 +16,11 @@ var filter = {
 
   searchFilterHideBtn: _getFilterElement( '.m-expandable_cue-close' ),
 
-  searchFilterSubmitBtn: _getFilterElement( 'btn[type="submit"]' )
+  searchCategoryLabel: _getFilterElement(
```
sebworks: should we use a `qa-hook` here?

test/browser_tests/spec_suites/organisms/pagination.js
```diff
@@ -0,0 +1,39 @@
...
+  var page;
+
+  beforeEach( function() {
+    page = new Blog();
+    page.get();
+  } );
+
+  it( 'should navigate to the second page', function() {
+    page.paginationNextBtn.click()
+    browser.sleep( 1000 );
```
sebworks: Is there an alternative to using sleep here? https://github.com/angular/protractor/issues/685

jimmynotjim: Nope, you have to wait for the next page to load, and there doesn't seem to be anything within Protractor that does this because the content itself is mostly the same (we could never rely on the results because they change too often).

test/browser_tests/shared_objects/filter.js
jimmynotjim: Possibly? I think we need to rework this list of checkboxes anyway, they should be in a `ul`, but that's out of the scope of this PR.

test/browser_tests/spec_suites/organisms/pagination.js
sebworks: Could you use: https://github.com/angular/protractor/issues/685#issuecomment-40249972?

jimmynotjim: @sebworks I tried something similar to check the url on load, then wait for it to not match, but I couldn't get it to work. Any ideas?

```
  it( 'should navigate to the second page', function() {
    var currentUrl = browser.getCurrentUrl();
    page.paginationNextBtn.click()

    browser.wait( function() {
      return browser.getCurrentUrl().then(function( url ) {
        return ( currentUrl !== url );
      } );
    } );

    expect( browser.getCurrentUrl() ).toContain( 'page=2' );
  } );
```

KimberlyMunoz:
Assuming there isn't an alternative to `browser.sleep`, :+1:

sebworks:
:, I'll take a look at it. Go ahead and merge at your convenience.

PR #2073, opened 2016-05-11
When the user resets their password, we should probably save it

rosskarchner:
Currently, when users attempt to use the 'forgot my password' feature, we aren't actually saving it.

This adds the missing 'form.save()' call to make the password change permanent.


kurtw:
👍

kave:
👍

rosskarchner:
@kave could you get this onto content(at least) ASAP?

PR #2074, opened 2016-05-11
legacy blog: duplicate content panels were causing pages not to save correctly

kave:
legacy blog: duplicate content panels were causing pages not to save correctly so additional edit boxes no longer override changed content

## Removals

- duplicate content panels were causing pages not to save correctly so additional edit boxes no longer override changed content

## Testing

- Load a db dump
- Edit and Publish a legacy blog page

## Review

- @jimmynotjim 
- @kurtw

### DIFF
--- a/cfgov/v1/models/blog_page.py
+++ b/cfgov/v1/models/blog_page.py
@@ -21,10 +21,8 @@ class BlogPage(AbstractFilterPage):
         ('image_text_50_50_group', organisms.ImageText5050Group()),
     ])
 
-    content_panels = AbstractFilterPage.content_panels
-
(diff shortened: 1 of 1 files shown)


jimmynotjim:
No clue on the code but it fixes the bug. Should we add some kind of test for this? These posts won't ever be transitioned to Wagtail components entirely right? That'd be a lot of manual work.

kave:
@jimmynotjim i don't think a test is necessary for this case since the form used to save the wagtail page was just setup incorrectly.
and no i believe these pages will stay as is

kurtw:
yikes 👍

jimmynotjim:
👍

PR #2080, opened 2016-05-12
Fixes issue with spacing after the last item in a full-width component

jimmynotjim:
The full-width text molecule always lives within a block, which has 60px of bottom spacing. The addition of bottom spacing to the last child of the full-width text molecule was being stacked with the block’s 60px increasing the space more than spec’d. Removing the bottom margin from the last child of the full-width text molecule fixes this issue.

## Changes

- Forced the last child of a full-width text molecule to remove any set padding.

## Testing

- `gulp build` and check out http://localhost:8000/about-us/project-catalyst/. The paragraph wrapping `Get the data` should have 0 bottom margin.

## Review

- @Scotchester 
- @KimberlyMunoz 
- @sebworks 
- @anselmbradford 

## Screenshots

screen shot 2016-05-12 at 2 04 45 pm

## Notes

- Thanks to @Scotchester for reporting this bug

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 ### Fixed
 
 - Fixed a validation bug in the Multiselect.
+- Fixed issue with spacing after the last-child.
 
 
 ## 3.0.0-3.3.14 - 2016-05-11
--- a/cfgov/unprocessed/css/enhancements/typography.less
+++ b/cfgov/unprocessed/css/enhancements/typography.less
@@ -383,6 +383,7 @@ dd {
     - cf-core
 */
 
+p:last-child,
 ul:last-child,
 .list__links .list_item:last-child {
     margin-bottom: 0;
--- a/cfgov/unprocessed/css/molecules/full-width-text.less
+++ b/cfgov/unprocessed/css/molecules/full-width-text.less
@@ -39,7 +39,7 @@
     max-width: 41.8em;
 
     &__serif {
-      font-family: Georgia, 'Times New Roman', serif;
+        font-family: Georgia, 'Times New Roman', serif;
     }
 }
 


anselmbradford:
To make sure I understand: The last child of the full-width text molecule will always be a paragraph atom, which adds a bottom margin, making the `!important` necessary?

Scotchester:
It doesn't seem like the `!important` is needed to me. The base rule has very low specificity, just applying directly to all `p` elements.

Scotchester:
Is this solution broad enough? Surely there are last children in other types of modules that also should get 0 bottom margin. [This rule](https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/unprocessed/css/enhancements/typography.less#L376-L389) already exists. Would it be too far-reaching to just add `p` elements to that?

jimmynotjim:
@Scotchester ah thanks I had forgotten about that rule. Do we have any worries that targeting all `p:last-child` will be too broad? I limited it to `full-width-nav` just because I was worried there might be unintended consequences.

Scotchester:
I tried thinking of a scenario in which a last-child `p` would be in a container that doesn't have its own bottom margin, and came up empty. I think it's worth a shot.

jimmynotjim:
Ha, I already found unintended consequences, but this is probably something we should fix anyway, that `div` is unnecessary. 

screen shot 2016-05-13 at 10 05 13 am

Scotchester:
@jimmynotjim Hmm. Any idea why that's there in the first place?

jimmynotjim:
@Scotchester Downside of rich text in Wagtail, and @kave says we can't do anything about it at the moment.

Scotchester:
So... we shouldn't do the universal `p:last-child` then, right?

jimmynotjim:
@Scotchester Yeah, reverting for now. Hopefully we can customize that on wagtail at some point. One thing that'll still be a problem is if a rich text field is the last item in a full-width-text

jimmynotjim:
Ok, @kurtw has a fix for the divs, but until that's in, this should probably wait. Feel free to merge while I'm away or when I get back on 5/26

kurtw:
Fix for removing "rich-text" classed div. https://github.com/cfpb/cfgov-refresh/pull/2096

jimmynotjim:
Ready for review again.

Scotchester:
👍

Scotchester:
Could we get one more sign-off on this, @anselmbradford @KimberlyMunoz @sebworks? Would like to get this in production before Jimmy returns from leave.

anselmbradford:
👍

PR #2081, opened 2016-05-12
Fixes "Login loop"

richaagarwal:
If a user logs in but doesn't have access to the Wagtail admin, it currently keeps redirecting them to the login screen. This displays a helpful error page instead. 

@rosskarchner @kave @kurtw 

Known issue:
If you do not have permissions to view the django-admin page, it will just keep you on the wagtail admin page. Not sure how to get it to display something else or if this is an issue.


rosskarchner:
I tested this yesterday, and it works as described-- thanks!  👍

cfgov/templates/wagtailadmin/access_denied.html
```diff
@@ -0,0 +1,11 @@
+{% overextends "wagtailadmin/login.html" %}
+{% load i18n admin_static %}
+
+
+{% block furniture %}
+    <div class="content-wrapper">
+        <h1>Sorry, you do not have permission to access the Wagtail admin page.</h1>
+        <h2>Did you mean to access the <a href="/django-admin/">{% trans 'Django admin' %} page?</h2>
+        <h2>Otherwise, you may want to <a href="/django-admin/logout/">{% trans 'Log out' %}</a> and log in as a user with the proper permissions.</h2>
```
anselmbradford: Why is log out wrapped in `trans`, but the rest of the sentence isn't?

richaagarwal: Ah, I wasn't quite sure how this was being used elsewhere - updated the template to make a little more sense, thanks!

PR #2082, opened 2016-05-12
Added browser tests for the multiselect

jimmynotjim:
There were no tests, fixing that with some baseline tests of: initialization, default results, filtering, clearing results, and keyboard highlighting.

## Additions

- Multiselect objects for testing
- Tests of basic behaviors

## Testing

- `gulp test:acceptance --sauce=false --specs=molecules/multiselect.js`

## Review

- @KimberlyMunoz 
- @anselmbradford 
- @sebworks 

## Notes

- Hit me with other behaviors you think we should test! These seemed like the most important to start with.

## Todos

- Add even more tests (but don't get so out of hand it becomes brittle).

## Checklist

### DIFF
--- a/test/browser_tests/shared_objects/multiselect.js
+++ b/test/browser_tests/shared_objects/multiselect.js
@@ -1,0 +1,24 @@
+'use strict';
+
+var _multiSelect =
+  element.all( by.css( '.cf-multi-select' ) ).first();
+
+function _getMultiSelectElement( selector ) {
+  return _multiSelect.element( by.css( selector ) );
+}
+
+var multiSelect = {
+  multiSelect: _multiSelect,
+
+  multiSelectChoices: _getMultiSelectElement( '.cf-multi-select_choices' ),
+
+  multiSelectHeader: _getMultiSelectElement( '.cf-multi-select_header' ),
+
+  multiSelectSearch: _getMultiSelectElement( '.cf-multi-select_search' ),
+
+  multiSelectFieldset: _getMultiSelectElement( '.cf-multi-select_fieldset' ),
+
+  multiSelectOptions: _getMultiSelectElement( '.cf-multi-select_options' )
+};
+
+module.exports = multiSelect;
--- a/test/browser_tests/spec_suites/molecules/multiselect.js
+++ b/test/browser_tests/spec_suites/molecules/multiselect.js
@@ -1,0 +1,208 @@
+'use strict';
+
+var Blog = require(
+    '../../page_objects/page_blog.js'
+  );
+
+describe( 'Multiselect', function() {
+  var page;
+
+  beforeEach( function() {
+    page = new Blog();
+    page.get();
+
+    page.searchFilterBtn.click();
+    browser.sleep( 1000 );
+  } );
+
+  describe( 'on page load', function() {
+    it( 'should create the multiselect', function() {
+      expect( page.multiSelect.isDisplayed() ).toBe( true );
+    } );
+
+    it( 'should not show the dropdown', function() {
+      expect( page.multiSelectFieldset.isDisplayed() ).toBe( false );
+    } );
+  } );
+
+  describe( 'when interacting with search input', function() {
+    it( 'should show the dropdown on click', function() {
+      page.multiSelectSearch.click();
+      browser.sleep( 1000 );
+
+      expect( page.multiSelectFieldset.isDisplayed() ).toBe( true );
+    } );
+
+    xit( 'should show the dropdown on focus', function() {
+      // Couldn't get this to tab over, not sure why
+      var prevInput = browser.element(
+        by.css( 'label[for="filter1_categories_info-for-consumers"]' )
+      );
+      prevInput.click();
+      browser.sleep( 1000 );
+      prevInput.sendKeys( protractor.Key.TAB );
+      browser.sleep( 1000 );
+
+      expect( page.multiSelectFieldset.isDisplayed() ).toBe( true );
+    } );
+
+    it( 'should hide the dropdown when losing focus', function() {
+      page.multiSelectSearch.click();
+      browser.sleep( 1000 );
+      page.multiSelectSearch.sendKeys( protractor.Key.TAB );
+      browser.sleep( 1000 );
+
+      expect( page.multiSelectFieldset.isDisplayed() ).toBe( false );
+    } );
+  } );
+
+  describe( 'when typing in search input', function() {
+    it( 'should return the matched results', function() {
+      page.multiSelectSearch.sendKeys( 'students' );
+      browser.sleep( 1000 );
+
+      var results = browser.element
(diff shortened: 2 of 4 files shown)


test/browser_tests/spec_suites/molecules/multiselect.js
```diff
@@ -0,0 +1,81 @@
...
+
+describe( 'Multiselect', function() {
+  var page;
+
+  beforeEach( function() {
+    page = new Blog();
+    page.get();
+  } );
+
+  it( 'should create the itial state of the multiselect', function() {
```
anselmbradford: `itial`

anselmbradford:
Do you really need the shared object? What's the use case for another test getting access to the multiselect dom that couldn't be directly covered in the multiselect test suite?

anselmbradford:
Consider breaking the describe blocks down further. So currently there's:

```
describe "Multiselect"
  it …
  it …
  it …
```

It's hard to see how the it blocks are grouped by user interaction. Instead, what about breaking it down into the page/user state actions:

```
describe "Multiselect"
  describe "on page load"
    it "should not show drop-down"
    it …
  describe "after interacting with input"
    it "should show drop-down on click"
    it "should hide drop-down on click into and out of the input"
    it …
  describe "after checking options"
    it "should check option"
    it "should show option"
    it …
  describe "after removing options"
    it "should disappear when clicked"
    it …
  describe "after typing text in the input"
    it "should return results on matched text entered"
    it "should NOT return results on unmatched text entered"
    it …
```

test/browser_tests/spec_suites/molecules/multiselect.js
jimmynotjim: Why can't I type that word?!?!

KimberlyMunoz:
These are good tests to start!

I'd add one to test that when the multiselect and the choices loses focus, that the multiselect closes.

KimberlyMunoz:
Oh, but I think my suggestion and Ans suggestions could go in another PR, so :+1:. I think we should keep the multiselect in its own shared object since it's getting its own test suite that we could hypothetically add to any page with a multiselect, but that should get hashed out.

sebworks:
👍

test/browser_tests/shared_objects/multiselect.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+var _multiSelect =
+  element( by.css( '.cf-multi-select' ) );
+
+function _getFilterElement( selector ) {
```
sebworks: Should this be  `_getMultiSelectElement`?

test/browser_tests/shared_objects/multiselect.js
```diff
@@ -0,0 +1,24 @@
...
+  multiSelectHeader: _getFilterElement( '.cf-multi-select_header' ),
+
+  multiSelectSearch: _getFilterElement( '.cf-multi-select_search' ),
+
+  multiSelectFieldset: _getFilterElement( '.cf-multi-select_fieldset' ),
+
+  multiSelectOptions: _getFilterElement( '.cf-multi-select_options' )
+};
+
+module.exports = filter;
```
sebworks: Should this be `multiSelect`?

test/browser_tests/shared_objects/multiselect.js
jimmynotjim: Whoops, hold over from the filter test I cloned.

jimmynotjim:
@sebworks thanks for catching those.

test/browser_tests/shared_objects/multiselect.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+var _multiSelect =
+  element( by.css( '.cf-multi-select' ) );
```
anselmbradford: Maybe turn this into `elements.all( by.css( '.cf-multi-select' ) ).first();` (I think that's right).. so we don't get a bunch `WARNING - more than one element found for locator By(css selector, .cf-multi-select) - the first result will be used`

anselmbradford:
If possible, remove the `browser.sleep` calls and replace with logic to check if the expected dom element has become visible or not. Looks like there's nearly 30 seconds in waiting that'll both slow down the tests and potentially add brittleness around execution time (e.g. a test would erroneously fail if for some reason any click took longer than a second to update the page)—this'll be particularly important when running these tests in the cloud where they're run on an instance that's sharing resources with other tests.

Anyway, that can happen in another PR if you want.

👍

test/browser_tests/spec_suites/molecules/multiselect.js
```diff
@@ -0,0 +1,208 @@
...
+
+    it( 'should highlight the first item', function() {
+      page.multiSelectSearch.click();
+      browser.sleep( 1000 );
+      page.multiSelectSearch.sendKeys( protractor.Key.ARROW_DOWN );
+
+      var results = browser.element
+        .all( by.css( '.cf-multi-select_options li input' ) );
+
+      expect( results.first().getInnerHtml() ).
```
anselmbradford: Dot should be moved to L:108

test/browser_tests/shared_objects/multiselect.js
jimmynotjim: What I didn't get is without the `all` it should automatically get the first. Changing to your suggestion fixes it though so ¯\_(ツ)_/¯

jimmynotjim:
@anselmbradford yeah, I'd like to get away from sleeping the browser, but with all the animations I was getting errors because if an animation hasn't finished the test complains that the element isn't in view. I'll try and explore that more in another PR next week.

test/browser_tests/spec_suites/molecules/multiselect.js
```diff
@@ -0,0 +1,208 @@
...
+  describe( 'when interacting with search input', function() {
+    it( 'should show the dropdown on click', function() {
+      page.multiSelectSearch.click();
+      browser.sleep( 1000 );
+
+      expect( page.multiSelectFieldset.isDisplayed() ).toBe( true );
+    } );
+
+    xit( 'should show the dropdown on focus', function() {
+      // Couldn't get this to tab over, not sure why
```
KimberlyMunoz: Is this a bug with protractor or with the multiselect?

KimberlyMunoz:
:+1:

test/browser_tests/spec_suites/molecules/multiselect.js
jimmynotjim: It works correctly in the browser, so it's either an issue with Protractor or with how I'm trying to implement the interaction.

KimberlyMunoz: Do you mind adding a note to that effect for people trying to fix these tests later?

jimmynotjim: Good call, updated.

PR #2090, opened 2016-05-13
Updates on-demand secondary navigation

anselmbradford:
## Changes

- Updates on-demand secondary navigation CSS to use CF resources by reference.
- Updates on-demand secondary navigation JS to check for presence of dom before initializing.

## Testing

- Needed updates for fin-ed-resources that have already been tested in https://github.cfpb.gov/CFGOV/fin-ed-resources/pull/70

## Review

- @Scotchester


Scotchester:
:+1:

PR #2091, opened 2016-05-16
Fixed a validation bug in the Multiselect

jimmynotjim:
When an author (or any string) contains a period, it is not a valid value. Updated the string validator to check strings for periods. Now the Multiselect will skip initialization and log a helper message if the value contains a period.

## Changes

- Added period to string validation regex.
- Added a test for the string validation using a period in the test string.

## Testing

- `gulp test:unit:scripts`

## Review

- @anselmbradford 
- @KimberlyMunoz 
- @kurtw 

## Notes

We really should get these values fixed in the DB to properly stringified values

## Checklist

### DIFF
--- a/test/unit_tests/modules/util/strings-spec.js
+++ b/test/unit_tests/modules/util/strings-spec.js
@@ -57,6 +57,14 @@ describe( 'Strings stringValid()', function() {
     }
   );
 
+  it( 'should return false when testing a string containing a period',
+    function() {
+      string = 'Some P. Name';
+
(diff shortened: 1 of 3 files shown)


test/unit_tests/modules/util/strings-spec.js
```diff
@@ -57,6 +57,14 @@ describe( 'Strings stringValid()', function() {
     }
   );
 
+  it( 'should return false when testing a string containing a period',
```
KimberlyMunoz: Thank you for writing a test after discovering a new bug. 💯

PR #2092, opened 2016-05-16
Removes redundant script includes

anselmbradford:
## Changes

- Updates FilterableListControls to instantiate its internal expandable.
- Removes redundant script includes in the wagtail ondemand script include.

## Testing

- `gulp build` and `/blog/` should still work.

## Review

- @kurtw 
- @jimmynotjim 
- @KimberlyMunoz 

## Notes

- Organisms should instantiate their own molecules. `expandable.js` should only ever need to be included when there is a standalone expandable on the page.

### DIFF
--- a/cfgov/unprocessed/js/organisms/FilterableListControls.js
+++ b/cfgov/unprocessed/js/organisms/FilterableListControls.js
@@ -3,6 +3,7 @@
 // Required modules.
 var atomicHelpers = require( '../modules/util/atomic-helpers' );
 var ERROR_MESSAGES = require( '../config/error-messages-config' );
+var Expandable = require( '../molecules/Expandable' );
 var getClosestElement = require( '../modules/util/dom-traverse' ).closest;
 var Multiselect = require( '../molecules/Multiselect' );
 var Notification = require( '../molecules/Notification' );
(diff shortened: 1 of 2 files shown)


jimmynotjim:
Should we add a browser test for this while we're thinking about it?

anselmbradford:
> Should we add a browser test for this while we're thinking about it?

For the organism directly? It looks like it's already used somewhere https://github.com/cfpb/cfgov-refresh/blob/flapjack/test/browser_tests/shared_objects/filterable-list-control.js

jimmynotjim:
No, for the blog page, that way we'll know next time (like when we move this to use the flyout rather than expandable) something changes we aren't breaking things.

jimmynotjim:
Also, those shared objects are only used in other tests because the filters are hidden, there's no test for the filter itself yet https://github.com/cfpb/cfgov-refresh/tree/flapjack/test/browser_tests/spec_suites/organisms

kurtw:
👍

anselmbradford:
@jimmynotjim I'll get the tests in another PR.

jimmynotjim:
@anselmbradford ok, lets not forget, our coverage continues to drop since launch

PR #2100, opened 2016-05-17
Events Updates

kave:
### Changes
- Rename Events body field => Subheading 
- Switch render location of live/future body fields to under the map (new body location)
- Change date used for post preview from date published to actual event date. 

@kurtw @richaagarwal

### DIFF
--- a/cfgov/v1/models/learn_page.py
+++ b/cfgov/v1/models/learn_page.py
@@ -122,7 +122,7 @@ class AgendaItemBlock(blocks.StructBlock):
 
 class EventPage(AbstractFilterPage):
     # General content fields
-    body = RichTextField(blank=True)
+    body = RichTextField('Sub-Heading',blank=True)
     archive_body = RichTextField(blank=True)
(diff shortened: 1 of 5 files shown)


cfgov/v1/models/learn_page.py
```diff
@@ -122,7 +122,7 @@ class Meta:
 
 class EventPage(AbstractFilterPage):
     # General content fields
-    body = RichTextField(blank=True)
+    body = RichTextField('Sub-Heading',blank=True)
```
kurtw: I really think renaming the field is better than changing the label. Either way still requires a migration and we can still remain consistent. _stopping now_

richaagarwal: Why is there a hyphen between sub and heading?

kave: @richaagarwal this is a good question. no good reason. i've removed it

kave:
@richaagarwal ready for review again

richaagarwal:
👍

PR #2104, opened 2016-05-18
Check against Activity Log topics when generating View More link

kurtw:
View more link to Activity log from a page with tags that are not in the Activity log causes filter to give error.

## Additions

- Generates the list of tags to check against when forming the link

## Testing

-Visit http://www.consumerfinance.gov/policy-compliance/rulemaking/final-rules/fair-credit-reporting-act-disclosures/
http://www.consumerfinance.gov/policy-compliance/ , go to the View more link, notice the error
- go to them locally and visit the link without problems on the activity log

## Review

- @kave 
- @richaagarwal 
- @rosskarchner

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 ## Added
 
 ### Changes
+- Check against Activity Log topics when generating View More link
 
 ### Removed
 
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -131,11 +131,15 @@ class CFGOVPage(Page):
     ])
 
     def generate_view_more_url(self, request):
-        tags = []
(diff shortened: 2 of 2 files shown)


richaagarwal:
The only error I am seeing is a validation error, which I see on both flapjack and this branch: 
screen shot 2016-05-18 at 11 40 57 pm

kurtw:
@richaagarwal for whatever reason, this branch isn't updating on github but git says everything's up to date. While I troubleshoot, did you make sure you were visiting the `View more` link on the branch? If you go to the link in the description, hit the view more link, then just change it to localhost, then you'll still see the error.

richaagarwal:
@kurtw Oh, ha I think I must not have updated to localhost.  Cool, looks good 👍

richaagarwal:
Just fyi though, I'm seeing a different error locally on flapjack now. `'list object' has no attribute 'paginator'`. Just to be aware of when you merge in...

PR #2106, opened 2016-05-18
Fix category filtering

kurtw:
Category filtering for blog was broken due to a broken condition and broken logic in generating the query.

## Testing

- go to newsroom and activity log to make sure filtering on blog only returns blog but also filtering with blog will return blogs and whatever was filtered with it as well.

## Review

- @kave 
- @richaagarwal 

## Todos

- write acceptance test


richaagarwal:
From this page, http://localhost:8000/about-us/blog/save-the-date-albuquerque/, if I click 'View more', I get the error ```ValueError at /activity-log/
'blog' is not in list
```

kave:
👍

kurtw:
Thanks @richaagarwal

PR #2108, opened 2016-05-19
missing template assignment

kave:
@sebworks 

### Testing
- Verify event archive page is renderable and filters are working correctly


sebworks:
👍

PR #2109, opened 2016-05-19
removed legacy templates

kave:
## Removals

- Old newsroom sheer template and unnecessary ES queries

## Testing

- run `./cfgov/manage.py sheer_index -r`
- visit `http://localhost:8000/about-us/newsroom/press-resources/` and see that it renders correctly now

## Review

- @sebworks


richaagarwal:
👍

Issue #571, opened 2015-05-19
Remove unneeded this lookups


anselmbradford:
See https://github.com/cfpb/cfgov-refresh/pull/552#discussion_r30545473

anselmbradford:
Only remains in BreakpointHandler I think https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/unprocessed/js/modules/BreakpointHandler.js#L72

sebworks:
This has been addressed.

PR #2114, opened 2016-05-19
Fixes init flag issue in cloned nodes

anselmbradford:
## Changes

- Adds `destroyInitFlag()` method to `atomic-helpers.js`.
- Adds `destroy()` method to `Expandables.js` to allow reversing calls to `init()`.
- Fixed issue where cloned expandables were not initializing on the bureau structure page.
- Removed `self` references in ContentSlider. (Part of https://github.com/cfpb/cfgov-refresh/issues/571)
- Adds extra small tests to bureau structure page.
 
## Testing

- `gulp test:acceptance --sauce=false --specs=bureau-structure.js --windowSize=480,900` should pass.
- http://localhost:8000/about-us/the-bureau/bureau-structure/ at mobile and desktop should work as expected (nodes and expandables expand).

## Review

- @sebworks 
- @KimberlyMunoz

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,10 +18,17 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 ### Added
 
 ### Changes
+- Frontend: Added `destroyInitFlag()` method to `atomic-helpers.js`.
+- Frontend: Added `destroy()` method to `Expandables.js` to allow
+  reversing calls to `init()`.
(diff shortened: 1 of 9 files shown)


sebworks:
I think we should refactor ContentSlider in the future but it's out of scope for this PR. 👍

anselmbradford:
> I think we should refactor ContentSlider in the future but it's out of scope for this PR. 

Agreed. It'd fix the test issues noted in this PR too.

PR #2115, opened 2016-05-20
Refactoring footer-button and removing jQuery

sebworks:
Refactoring footer-button and removing jQuery

## Changes

- Refactored `footer-button.js` to remove jQuery and use requestAnimationFrame for smooth scrolling to the top of page. 

## Testing

- Visit `localhost:8000` on mobile an click the `Back to Top` button. It should smoothly scroll you back to the top of the page.
- Run 'gulp test:unit:scripts`

## Review

- @KimberlyMunoz 
- @anselmbradford 

## Todos

- Need to write further acceptance and unit tests.

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/footer.html
+++ b/cfgov/jinja2/v1/_includes/organisms/footer.html
@@ -21,9 +21,10 @@
 
             <a class="btn
                       btn__secondary
-                      js-return-to-top
                       o-footer_top-button"
-               href="#">
+                data-gtm_ignore="true"
+                data-js-hook="behavior_return-to-top"
+                href="#">
                 Back to top <span class="cf-icon cf-icon-arrow-up"></span>
             </a>
 
--- a/cfgov/unprocessed/js/modules/footer-button.js
+++ b/cfgov/unprocessed/js/modules/footer-button.js
@@ -1,26 +1,59 @@
 /* ==========================================================================
    Footer Button: Scroll to Top
+
+   Code copied from the following with minimal modifications :
+
+   - http://stackoverflow.com/questions/21474678/
+     scrolltop-animation-without-jquery
    ========================================================================== */
+
 'use strict';
 
-var $ = require( 'jquery' );
+// Required modules.
+var behavior = require( './util/behavior' );
 
-// TODO: Refactor this file to remove jquery dependency.
-//       http://stackoverflow.com/questions/21474678/
-//       scrolltop-animation-without-jquery
 /**
  * Set up event handler for button to scroll to top of page.
  */
 function init() {
-  var duration = 300;
+  if ( 'requestAnimationFrame' in window === false ) {
+    return;
+  }
 
-  // Disable Google Tag Manager tracking on this link.
-  $( '.js-return-to-top' ).attr( 'data-gtm_ignore', 'true' );
-
-  $( '.js-return-to-top' ).click( function( event ) {
+  behavior.attach( 'return-to-top', 'click', function( event ) {
     event.preventDefault();
-    $( 'html, body' ).animate( { scrollTop: 0 }, duration );
+    _scrollToTop();
   } );
+}
+
+/**
+ *  Duration of the scroll to top of the page.
+ */
+function _scrollToTop() {
+  var SCROLL_DURATION = 300;
+  var SCROLL_STEP_DURATION = 10;
+  var scrollHeight = window.scrollY;
+  var scrollStep = Math.PI / ( SCROLL_DURATION / SCROLL_STEP_DURATION );
+  var cosParameter = scrollHeight / 2;
+  var scrollCount = 0;
+  var scrollMargin;
+
+  window.requestAnimationFrame( _step );
+
+  /**
+   * Decrement scroll Y position.
+   */
+  function _step() {
+    window.setTimeout( function() {
+      if ( window.scrollY !== 0 ) {
+        window.requestAnimationFrame( _step );
+        scrollCount += 1;
+        scrollMargin = cosParameter - ( cosParameter *
+                       Math.cos( scrollCount * scrollStep ) );
+        window.scrollTo( 0, scrollHeight - scrollMargin );
+      }
+    }, SCROLL_STEP_DURATION );
+  }
 }
 
 module.exports = { init: init };
--- a/cfgov/unprocessed/js/modules/focus-target.js
+++ b/cfgov/unprocessed/js/modules/focus-target.js
@@ -17,7 +17,7 @@ var attachBehavior = require( './util/behavior' ).attach;
  */
 function init() {
 
-  attachBehavior( 'a[href^="#"]', 'click', function behavior() {
+  attachBehavior( 'a[href^="#"]:not([href="#"])', 'click', function behavior() {
     var anchorSelector = this.getAttribute( 'href' );
     var anchorElement = document.querySelector( anchorSelector );
     if ( anchorElement ) {
--- a/test/unit_tests/organisms/Footer-spec.js
+++ b/test/unit_tests/organisms/Footer-spec.js
@@ -1,10 +1,83 @@
 'use strict';
-var BASE_JS_PATH = '../../../cfgov/unprocessed/js/';
 
 var chai = require( 'chai' );
 var expect = chai.expect;
-var Footer = require( BASE_JS_PATH + 'organisms/Footer' );
+var jsdom = require( 'mocha-jsdom' );
+var sinon = require( 'sinon' );
+
+var BASE_JS_PATH = '../../../cfgov/unprocessed/js/';
+var FooterButton = require( BASE_JS_PATH + 'modules/footer-button.js' );
+
+var footerBtnDom;
+var bodyDom;
+var sandbox;
+var lastTime = 0;
+
+var HTML_SNIPPET =
+    '<a class="btn btn__secondary o-footer_top-button" ' +
+        'data-gtm_ignore="true" data-js-hook="behavior_return-to-top" ' +
+        'href="#">' +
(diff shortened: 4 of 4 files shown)


cfgov/jinja2/v1/_includes/organisms/footer.html
```diff
@@ -21,10 +21,10 @@
 
             <a class="btn
                       btn__secondary
-                      js-return-to-top
                       o-footer_top-button"
-               href="#">
-                Back to top <span class="cf-icon cf-icon-arrow-up"></span>
+                data-js-hook="behavior_return-to-top"
+                href="#">
+                Back to top <span clasbs="cf-icon cf-icon-arrow-up"></span>
```
anselmbradford: `clasbs` --> `class`

cfgov/unprocessed/js/modules/footer-button.js
```diff
@@ -1,26 +1,64 @@
...
 function init() {
-  var duration = 300;
+  var scrollDuration = 300;
+  var behaviorElement = dataHook.get( 'return-to-top' );
 
-  // Disable Google Tag Manager tracking on this link.
-  $( '.js-return-to-top' ).attr( 'data-gtm_ignore', 'true' );
+  if ( behaviorElement === null ||
+       'requestAnimationFrame' in window === false ) {
+    return;
```
anselmbradford: This should probably throw an error.

anselmbradford: Well not for the `requestAnimationFrame`, but for the missing element.

cfgov/unprocessed/js/modules/util/data-hook.js
```diff
@@ -60,8 +60,22 @@ function contains( element, value ) {
   return values.indexOf( value ) > -1 ? true : false;
 }
 
+/**
+ * @param {string} behavior
+ *  Behavior type used to find the element within the dom.
+ * @returns {HTMLNode} if it exists in the dom, null otherwise.
+ */
+function get( behavior ) {
```
anselmbradford: Can we name this `getElementFor` or `getElemFor` or something? `get` is generic and may imply this is getting the data-hook (`data-js-hook`).

cfgov/unprocessed/js/modules/util/data-hook.js
```diff
@@ -60,8 +60,22 @@ function contains( element, value ) {
...
+  var behaviorSelector =
+  standardType.JS_HOOK + '*=' + standardType.BEHAVIOR_PREFIX + behavior;
+  var behaviorElement = document.querySelector( '[' + behaviorSelector + ']' );
```
anselmbradford: It may help to pass in the scope, so instead of requiring querying on `document`, have that be the default that can be overridden by a passed in element. This would be helpful for behaviors that have several pieces and don't want to search the whole document again.

jimmynotjim:
> Need to write further acceptance and unit tests.

Can you add a Jira task specifically for this file if the tests won't be added in this PR. Thanks.

cfgov/unprocessed/js/modules/footer-button.js
```diff
@@ -1,26 +1,64 @@
...
+  }
 
-  $( '.js-return-to-top' ).click( function( event ) {
+  // Disable Google Tag Manager tracking on this link.
+  behaviorElement.setAttribute( 'data-gtm_ignore', 'true' );
```
jimmynotjim: Can we add this attribute in the markup? Seems like we're crossing some streams here.

cfgov/unprocessed/js/modules/footer-button.js
```diff
@@ -1,26 +1,64 @@
...
+  function _step() {
+    setTimeout( function() {
+      if ( window.scrollY !== 0 ) {
+        requestAnimationFrame( _step );
+        scrollCount += 1;
+        scrollMargin = cosParameter - cosParameter *
+                       Math.cos( scrollCount * scrollStep );
+        window.scrollTo( 0, scrollHeight - scrollMargin );
+      }
+    }, 15 );
```
jimmynotjim: Where's the `15` come from? Should it be a var?

cfgov/unprocessed/js/modules/util/data-hook.js
jimmynotjim: What about `find`? That might better fit with the other single term functions in this file.

cfgov/unprocessed/js/modules/footer-button.js
anselmbradford: The rationale for keeping it in the JS elsewhere is that (a) GTM lives only in JavaScript and (b) in some cases if the default link behavior is being prevented by a particular script, if that script were to be disabled/removed the links should _not_ continue ignoring GTM. In that case it makes sense to set `data-gtm_ignore` where `event.preventDefault()` is set.

cfgov/unprocessed/js/modules/util/data-hook.js
anselmbradford: I'd still mention the element... buuuut I'm going on vacation and won't be checking GH for about 9 days. In any case, I look at it like this --> `add`, `remove`, and `contains` are all talking about values inside the data-hook and follow the classList API. By that logic, dataHook.get/dataHook.find would be looking for the value inside the data-hook. However, by retrieving the parent element this method is going the other direction and is talking about outside of the data-hook, so that difference is worth noting to prevent confusion that get/find would be returning the data-hook value.

cfgov/unprocessed/js/modules/footer-button.js
jimmynotjim: The default behavior of an empty anchor link is still to move the user to the top of the page. If for some reason GTM loaded but this script didn't, we should still ignore the interaction in GTM.

cfgov/unprocessed/js/modules/util/data-hook.js
sebworks: So I'm going to remove this function and use [behavior.find](https://github.com/cfpb/cfgov-refresh/blob/1e8a673b248ca58fa55b8e9cfd9804f9f7f433a9/cfgov/unprocessed/js/modules/util/behavior.js#L127) instead.

cfgov/unprocessed/js/modules/util/data-hook.js
sebworks: So I'm going to remove this function and use [behavior.find](https://github.com/cfpb/cfgov-refresh/blob/1e8a673b248ca58fa55b8e9cfd9804f9f7f433a9/cfgov/unprocessed/js/modules/util/behavior.js#L127) instead.

cfgov/unprocessed/js/modules/footer-button.js
anselmbradford: May mean the link should be injected with JS since it doesn't do anything if this script isn't here.

sebworks: The button still functions without JS. I'm still trying to figure out if GTM works without JS.

jimmynotjim: The link doesn't need to be injected, the jump instead of scroll fallback is perfectly fine. I'm just not convinced we should be injecting metric tracking manipulation into a script that's simply adjusting the button unless that adjustment were to fundamentally change the behavior, which this doesn't. With or without the scrolling action, we don't want this button tracked in GTM. Writing the attr to the markup separates the concerns _and_ ensures GTM isn't tracking the button clicks if this file were to fail in some way.

richaagarwal:

jimmynotjim:
@sebworks if you need a second hand getting this through let me know

cfgov/jinja2/v1/_includes/organisms/footer.html
vccabral: seems legit

sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2115/commits/6d1c6acdb567feaecb0248214349bd32cf794dd2

cfgov/unprocessed/js/modules/footer-button.js
sebworks: So the `behavior.find` method will throw an error. Fixed in https://github.com/cfpb/cfgov-refresh/pull/2115/commits/6d1c6acdb567feaecb0248214349bd32cf794dd2.

cfgov/unprocessed/js/modules/footer-button.js
sebworks: I've moved the GTM code to the markup because you will never want to track the event.

cfgov/unprocessed/js/modules/footer-button.js
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2115/commits/6d1c6acdb567feaecb0248214349bd32cf794dd2

sebworks:
@jimmynotjim, @anselmbradford please give it look when you get a chance.

jimmynotjim:
👍 Can someone on Platform run the tests? I'm so far back I can't get things running correctly.

chosak:
Can confirm that browser tests pass for me. 👍

cfgov/unprocessed/js/modules/footer-button.js
```diff
@@ -1,26 +1,60 @@
...
-  $( '.js-return-to-top' ).click( function( event ) {
+  behavior.attach( 'return-to-top', 'click', function( event ) {
     event.preventDefault();
-    $( 'html, body' ).animate( { scrollTop: 0 }, duration );
+    _scrollToTop();
   } );
 }
 
+/**
+ * @param {integer} scrollDuration
```
anselmbradford: param is not in implementation.

anselmbradford:
👍

PR #2120, opened 2016-05-20
Fixes #936 - updates show_only name

### LINKED ISSUE #936: Refactor show_only to be more inclusive of multiple values
`_includes/macros/time.html` includes a parameter `show_only`, which actually allows more than one date parameter to be shown, so a better name should be chosen.

See https://github.com/cfpb/cfgov-refresh/pull/925#discussion_r37897845

anselmbradford:
## Changes

- Fixes https://github.com/cfpb/cfgov-refresh/issues/936

## Review

- @jimmynotjim 
- @KimberlyMunoz 
- @sebworks

### DIFF
--- a/cfgov/jinja2/v1/_includes/macros/time.html
+++ b/cfgov/jinja2/v1/_includes/macros/time.html
@@ -10,32 +10,32 @@
 
    Renders a time element with a standard datetime format when given:
 
-   datetime:  A date timestamp to format.
+   datetime:       A date timestamp to format.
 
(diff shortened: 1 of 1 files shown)


sebworks:
Are we just changing the object name here?

anselmbradford:
Yup, easy one that should have been tackled a long time ago! https://github.com/cfpb/cfgov-refresh/issues/936

sebworks:
👍

PR #2124, opened 2016-05-20
Fix for newsroom

richaagarwal:
@kave


anselmbradford:
👍

cfgov/v1/models/browse_filterable_page.py
```diff
@@ -68,18 +68,17 @@ def get_form_class(self):
...
-                    if 'blog' in categories:
-                        del categories[categories.index('blog')]
-                    if not categories:
-                        only_blog = True
-
+                if categories:
+                    if 'blog' not in categories:
+                        get_blog = False
+                    else:
+                        if len(categories) == 1:
```
sebworks: can you use  `elif ` here?

sebworks:
Are we changing the logic from only including the blog if no categories to always include the blog?

cfgov/v1/models/browse_filterable_page.py
richaagarwal: @sebworks how so?

anselmbradford: +1 @sebworks ... I think he means:
```
if categories:
    if 'blog' not in categories:
        get_blog = False
    elif len(categories) == 1:
        only_blog = True
```

richaagarwal: Oh, right :P  Hmm. I like `elif` for when there are multiple conditions that are unrelated, e.g. `if x == 0: ... elif x == 1: ... elif x == 2: ... ` but as simply a combination of `else` and `if` it seems to be less readable.  Does anyone have stronger thoughts on this?

sebworks: ok, that explanation works for me.

sebworks:
👍

cfgov/v1/models/browse_filterable_page.py
sebworks: I would just add it might help to add comments here. Just so people understand the underlying reasons for the logic.

PR #2131, opened 2016-05-24
Use bare value of RichText field if value type is not RichText

kurtw:
This should fix the issue along with the other fields that we haven't been catching either.

## Changes

- Use bare value of RichText field if value type is not RichText

## Testing

- Add link to a Contact's body field and view it in a sidebar on a page. Make sure the link is valid.

## Review

- @richaagarwal

### DIFF
--- a/cfgov/v1/__init__.py
+++ b/cfgov/v1/__init__.py
@@ -93,7 +93,8 @@ def parse_links(soup):
 def external_links_filter(value):
     if isinstance(value, RichText):
         return parse_links(BeautifulSoup(expand_db_html(value.source), 'html.parser'))
-    return value
+    else:
+        return parse_links(BeautifulSoup(expand_db_html(value), 'html.parser'))
 
 
 @contextfunction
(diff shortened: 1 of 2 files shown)


kave:
👍

cfgov/v1/__init__.py
```diff
@@ -93,6 +93,8 @@ def parse_links(soup):
 def external_links_filter(value):
     if isinstance(value, RichText):
         return parse_links(BeautifulSoup(expand_db_html(value.source), 'html.parser'))
+    else:
+        return parse_links(BeautifulSoup(expand_db_html(value), 'html.parser'))
     return value
```
richaagarwal: `return value` would never be hit here. are we sure that there isn't any other condition that could be met than the two RichText conditions?

richaagarwal:
Other than comment I left, 👍

cfgov/v1/__init__.py
kave: yea good question, is there a reason for that last return?

kurtw:
Hah, whoops. Merging after that comment is addressed.

PR #2132, opened 2016-05-24
Removes macropolo

anselmbradford:
Removes macropolo. This doesn't necessarily have to go in, but is a snapshot reference of where it is used in the project if it is updated.

## Removals

- Macropolo jinja unit tests.

## Review

- @richaagarwal 
- @kave 
- @kurtw 
- @KimberlyMunoz 
- @sebworks 
- @jimmynotjim


richaagarwal:
Awesome, thanks for doing this.  It seems like we only lose two tests (the two .json files) based on my reading of `test_macros.py`, so I'd be in favor of removing this and rewriting those tests another more standard way.

anselmbradford:
Cool. Anyone else have an opinion on whether it should stay or go before I merge this?

sebworks:
👍

rosskarchner:
👍

KimberlyMunoz:
I'd ask @willbarton about what the future of Macro Polo is before merging.

kave:
👍

willbarton:
Go for it.

I am happy to develop Macro Polo further if there's a need, but if there's not, feel free to remove it!

PR #2085, opened 2016-05-13
Filter form testing

kurtw:
Refactor and test Filterable List functionality. 

## Additions

- Tests

## Changes

- filterable_context.py is now called filterable_list.py

## Testing

- `tox`

## Review

- @kave 
- @richaagarwal 
- @rosskarchner 

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #2133, opened 2016-05-24
Fix Notices and Comment

kurtw:
hotfix for bad call to get_page_set from changing its location in #2085 

## Review

- @rosskarchner 
- @kave


kave:
👍

PR #2135, opened 2016-05-24
resolved notices page 500s, due to filtering refactorings

kave:
@kurtw


kave:
pushing directly to beta to create a tag. This has already been testing and committed into flapjack

PR #2138, opened 2016-05-25
Change validation for Event live stream URLs

kurtw:
Some of the URLs from ustream don't contain `embed` which invalidates the URL. This removes that constraint.

## Testing

- Try to save `https://www.ustream.tv/video` as a URL in the Live Stream URL field on an Event page.

## Review

- @kave


sebworks:
@kurtw , all the livestreams should be embeds.

kurtw:
@sebworks you're right, they should have the correct url. Thanks.

PR #2118, opened 2016-05-20
Refactoring focus target to remove jQuery

sebworks:
Refactoring focus target to remove jQuery

## Additions

- Added `cfgov/unprocessed/js/modules/util/Behavior.js` to provide easy mechanism for adding behaviors.

## Changes

- Modified `cfgov/unprocessed/js/modules/focus-target.js` to remove jQuery and uitilize Behavior utility.

## Testing

- Visit `http://localhost:8000/` and tab once. Click on the `Skip to main content` button in the upper left-hand corner. Observe that the `#main` element should have its `tab-index` attribute set to `-1`.

## Review

- @anselmbradford 
- @KimberlyMunoz 

## Notes

-

## Todos

-

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #2142, opened 2016-05-25
Moves behavior dom check to utility

anselmbradford:
## Changes

- Removes unneeded constant exports from FlyoutMenu.
- Moves behavior DOM check to utility method.
- Removes TODO for atomic-helper, as behavior shouldn't use that for clarity.

## Testing

- `gulp build && gulp test:unit && gulp test:acceptance --sauce=false`

## Review

- @sebworks 
- @jimmynotjim 

## Notes

- Adds to `/util/behavior.js` created in https://github.com/cfpb/cfgov-refresh/pull/2118

### DIFF
--- a/cfgov/unprocessed/js/modules/behavior/FlyoutMenu.js
+++ b/cfgov/unprocessed/js/modules/behavior/FlyoutMenu.js
@@ -2,6 +2,7 @@
 
 // Required modules.
 var BaseTransition = require( '../../modules/transition/BaseTransition' );
+var behavior = require( '../../modules/util/behavior' );
 var dataHook = require( '../../modules/util/data-hook' );
 var EventObserver = require( '../../modules/util/EventObserver' );
 var fnBind = require( '../../modules/util/fn-bind' ).fnBind;
@@ -12,13 +13,16 @@ var standardType = require( '../../modules/util/standard-type' );
  * @class
  *
  * @classdesc Initializes new FlyoutMenu behavior.
+ * Behaviors are functionality that can be shared between different pieces
+ * of markup. They are not strictly atomic, though they likely are used
+ * on atomic components.
  * As added JS behavior, this is added through HTML data-js-hook attributes.
  *
  * Structure is:
- * flyout-menu
- *   flyout-menu_trigger
- *   flyout-menu_content
- *     flyout-menu_alt-trigger
+ * behavior_flyout-menu
+ *   behavior_flyout-menu_trigger
+ *   behavior_flyout-menu_content
+ *     behavior_flyout-menu_alt-trigger
  *
  * The alt-trigger is for a back button, which may obscure the first trigger.
  * The flyout can be triggered three ways: through a click of the trigger or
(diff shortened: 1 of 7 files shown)


cfgov/unprocessed/js/modules/util/behavior.js
```diff
@@ -0,0 +1,70 @@
...
+  var dom;
+
+  if ( dataHook.contains( element, behaviorDataAttr ) ) {
+    dom = element;
+  }
+
+  // If it's not on the passed DOM node,
+  // query the node to see if it's in the children,
+  // if the passed DOM node isn't null.
+  if ( !dom && element ) {
```
sebworks: instead of doing `!dom` can you just add a return statement above?

cfgov/unprocessed/js/modules/util/behavior.js
```diff
@@ -0,0 +1,70 @@
...
+    var selector = '[' + standardType.JS_HOOK + '=' + behaviorDataAttr + ']';
+    dom = element.querySelector( selector );
```
sebworks: Why check the children to determine if they have a behavior?

anselmbradford: Say you have this HTML:
```
<div data-js-hook="behavior_flyout-menu">
  …
  <button data-js-hook="behavior_flyout-menu_trigger">
  …
  <div data-js-hook="behavior_flyout-menu_content">
```

To check that the behavior dom is what it's expecting and to retrieve dom references, this can be called:
```
// element is an existing DOM node reference to `<div data-js-hook="behavior_flyout-menu">`
var dom = behavior.checkBehaviorDom( element, 'behavior_flyout-menu' );
var triggerDom = behavior.checkBehaviorDom( element, 'behavior_flyout-menu_trigger' );
var contentDom = behavior.checkBehaviorDom( element, 'behavior_flyout-menu_content' );
```

The child query inside the validity check allows only the parent DOM node reference to be passed in.

cfgov/unprocessed/js/modules/util/behavior.js
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2142/commits/a05671ca802cc590aa60a6b6ad9d9eecfe59ab6a

sebworks: One more on line 59.

cfgov/unprocessed/js/modules/util/behavior.js
sebworks: It's just a little strange because you don' t know whether it's on the element itself or on a child element. How about making it `findDomByBehavior` instead?

cfgov/unprocessed/js/modules/behavior/FlyoutMenu.js
```diff
@@ -31,24 +35,14 @@ function FlyoutMenu( element ) { // eslint-disable-line max-statements, no-inlin
...
-
-  var _triggerDom = _dom.querySelector( TRIGGER_SEL );
-  var _contentDom = _dom.querySelector( CONTENT_SEL );
 
-  if ( !_triggerDom ) { throw new Error( TRIGGER_SEL + ' is missing!' ); }
-  if ( !_contentDom ) { throw new Error( CONTENT_SEL + ' is missing!' ); }
+  // Verify that the expected dom attributes are present.
+  var _dom = behavior.checkBehaviorDom( element, BASE_CLASS );
+  var _triggerDom = behavior.checkBehaviorDom( element, BASE_CLASS + '_trigger' );
+  var _contentDom = behavior.checkBehaviorDom( element, BASE_CLASS + '_content' );
```
jimmynotjim: Would `_actionDom` be a more accurate naming convention to go with the idea that this is a behavior and not necessarily an atomic component.

test/unit_tests/modules/util/behavior-spec.js
```diff
@@ -0,0 +1,60 @@
...
+
+  var HTML_SNIPPET = '<div data-js-hook="behavior_flyout-menu">' +
+                     '<div data-js-hook="behavior_flyout-menu_content">' +
+                     '</div></div>';
+  var initdom = jsdom.jsdom( HTML_SNIPPET );
+  var document = initdom.defaultView.document;
+
+  var containerDom;
+  var behaviorElmDom;
+  var selector = '[data-js-hook=behavior_flyout-menu';
```
jimmynotjim: Missing a closing bracket here

jimmynotjim: Nvm, seeing how this is used on L:24-25. It's a bit odd, I'd personally rather they do `'[` + selector + ']'` and `'[' + selector + '_content ]'` so it's clear where they begin and end.

anselmbradford: Makes sense, updated in https://github.com/cfpb/cfgov-refresh/pull/2142/commits/e1d6e961302bcf3879de10222636289663a053c8

cfgov/unprocessed/js/modules/behavior/FlyoutMenu.js
anselmbradford: Mm, that's a good question. I'd say pressing the trigger performs the action of showing and hiding the content, and therefore the contentDom is the content, not the action.

sebworks:
👍

cfgov/unprocessed/js/modules/behavior/FlyoutMenu.js
jimmynotjim: Ah right, guess that would be more like resultDom. Keep it as is.

jimmynotjim:
👍

PR #2117, opened 2016-05-20
Add Wagtail module for Social Media molecule

Scotchester:
This PR enables the Social Media molecule to be used as a module in Wagtail, available to the Sidefoot.

## Additions

- Ability to add Social Media module to Sidefoot streams in Wagtail

## Changes

- Tweak styling of the Social Media molecule

## Testing

1.  Pull down this branch
1. `gulp styles`
1. `./cfgov/manage.py migrate`
1. `./cfgov/manage.py runserver`
1. Login to your local Wagtail at http://localhost:8000/admin/
1. Add the new Social Media module to a page's Sidefoot
1. Preview that page
1. Also review an existing Blog page and the Footer to ensure the Social Media molecule there looks as expected.

## Review

- @anselmbradford 
- @KimberlyMunoz 
- @sebworks 
- @kave 

## Screenshots

screen shot 2016-05-20 at 10 06 10

## Notes

- I added 30px of top margin to `.post_body`, because bottom margin was removed from the Social Media molecule's atoms to be consistent to the atomic design spacing principles as I (and @anselmbradford) understand them. 30px seems consistent with the latest designs for articles, though those haven't been updated in quite a while. Please comment if you have any concerns with this. /cc @duelj @ajbush 

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #2143, opened 2016-05-25
Hotfix: Add value default back

anselmbradford:
## Changes

- Add default value for `value` back in from https://github.com/cfpb/cfgov-refresh/pull/2117, otherwise non-wagtail pages that use the social media molecule fail (e.g. /about-us/the-bureau/about-director/)

## Testing

- `gulp test:acceptance --sauce=false`

## Review

- @Scotchester 
- @kave 
- @kurtw 

## Notes

- Not sure if there's a better way to provide the default, but this seems to work inside and out of wagtail.

### DIFF
--- a/cfgov/jinja2/v1/_includes/molecules/social-media.html
+++ b/cfgov/jinja2/v1/_includes/molecules/social-media.html
@@ -32,7 +32,7 @@
 
    ========================================================================== #}
 
-{% macro render( value ) %}
+{% macro render( value={} ) %}
 
 {% set blurb = (value.blurb or (page.seo_title if page) or value.title or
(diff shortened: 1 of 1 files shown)


sebworks:
👍

Scotchester:
Gah, sorry. One note, though: the bug is not because it's a non-Wagtail page; it's because that page calls `social_media.render()` with no parameters.

:+1:

anselmbradford:
> Gah, sorry. One note, though: the bug is not because it's a non-Wagtail page; it's because that page calls social_media.render() with no parameters.

Hm, I suppose those calls could be updated to `social_media.render({})`, but this seems safer.

Scotchester:
Yeah, this is definitely the right fix.

PR #2144, opened 2016-05-25
Remove `atomicName` parameter

anselmbradford:
`atomicName` parameter was passed around purely to provide a more descriptive error message. However, the class name can be used instead, since—if the atomic naming is correct—the class name will include the atomic name. This simplifies the atomic helper utilities.

## Removals

- `atomicName` parameter from `atomicHelpers.checkDom`.

## Changes

- Updates associated calling code and tests.

## Testing

- `gulp test:unit:scripts` should pass. `gulp build` and no change in site functionality.

## Review

- @sebworks 
- @jimmynotjim


sebworks:
👍

PR #2148, opened 2016-05-26
Removes console.log statement

anselmbradford:
Oops, noticed a console.log slipped through and is logging "value" in the unit test output.

## Removals

- Removes console.log statement from unit test.

## Testing

- `gulp lint:tests` should have no errors.

## Review

- @sebworks 
- @jimmynotjim


sebworks:
👍 , thanks

PR #2149, opened 2016-05-26
Adds BASE_CLASS and dom validity check to multiselect.

anselmbradford:
## Changes

- Adds `BASE_CLASS` and dom validity check to multiselect to align it with the other atomic components lifecycle.
- Fixes some minor linter warnings.

## Testing

- `gulp test:unit:scripts && gulp test:acceptance --sauce=false` should pass (expect for the 2 acceptance test errors we're getting from wagtail on all PRs).

## Review

- @jimmynotjim 
- @sebworks 

## Todos

- Added a todo to scope the multiselect to markup that has the base class. If global behavior is needed, a separate script like this can be added to common.js or similar:
```
var mselects = document.querySelector( 'select[multiple]' );
for ( var i, len = mselects.length; i < len; i++ ) {
  mselects.classList.add( 'cf-multi-select' );
}
```


cfgov/unprocessed/js/molecules/Multiselect.js
```diff
@@ -21,6 +21,14 @@ var strings = require( '../modules/util/strings' );
  */
 function Multiselect( element ) { // eslint-disable-line max-statements, inline-comments, max-len
 
+  var BASE_CLASS = 'cf-multi-select';
+
+  // TODO: Add 'cf-multi-select' class to <select> menu markup
```
jimmynotjim: Any reason not to update them in this PR? https://github.com/cfpb/cfgov-refresh/blob/e1ed9c3ae5e1b5815b0615a8caea8ab796ae0ed2/cfgov/v1/forms.py#L81-L88

jimmynotjim: On second thought, should we be using a js-hook rather than a classname? We don't want the styles applied to the native select.

test/unit_tests/molecules/Multiselect-spec.js
```diff
@@ -71,7 +73,7 @@ describe( 'Multiselect', function() {
       option.value = 'Foo\'';
       multiselect.init();
       selectDom = document.querySelectorAll( 'select[multiple]' );
-      multiselectDom = document.querySelectorAll( '.cf-multi-select' );
+      multiselectDom = document.querySelectorAll( 'div.cf-multi-select' );
```
jimmynotjim: Related to the question above, if the classname isn't used to the native select, this won't need the `div` chaining.

cfgov/unprocessed/js/molecules/Multiselect.js
anselmbradford: Possibly, though I'd lean toward no. Leave the js-hook for things that don't do any of their own styling. This is kind of a weird case because it's saying "style this thing when it gets swapped out with a div," so the select isn't receiving styles but it's swapped with something that does. It seems unclear to shuffle from a js-hook attribute to a class during the swap. And attaching styles to the js-hook is counter to its intended use. 
 What would be the downside of scoping the styles to `div.cf-multi-select`? Also, potentially there could be fallback styles that would be useful at `select.cf-multi-select`

jimmynotjim: Default multiselects are already styled (although very basically), so that's not an issue.

If we used a data-hook for the init, the styles would still only be attached to the class name, not the hook. The hook would only be for scripting, which is the intended use. As a side note, this component will support single option selects in the future using radio buttons instead of checkboxes, so naming will change at some point.

richaagarwal:

anselmbradford:
Can we give it another week so I have time to re-visit after re-group?

richaagarwal:
@anselmbradford Sure thing, do you think you can re-visit by the end of this week?

anselmbradford:
@richaagarwal Yeah I think so! Thanks for your patience, I know it's been open awhile

anselmbradford:
Closing and re-opening against master branch

PR #2156, opened 2016-05-31
Move assets from cfpb_nemo to this repo, make nonresponsive JS updates

rosskarchner:
This makes the world better in a few ways. Right now, if you need to tweak a Django app that depends on the nonresponsive header, you need to change:

- your app (maybe)
- this repo (maybe)
- our wordpress theme (what?)

This *at least* removes that last step. I've also made some requested changes to the nonresponsive header JS.

Someone (FEWD) should:

- [x] Make the CSS changes necessary to correct the issue in the screenshot bellow
- [ ] Convert Nemo's grunt tasks to our gulpfile

That can either happen in a subsequent PR, or just commit to this branch.

screen shot 2016-05-26 at 1 41 53 pm

In chat, @anselmbradford said:
```
/v1-nonresponsive-header.css needs a manual update likely :-\

Needs the styles from https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/unprocessed/css/on-demand/header.less with all the responsive bits deleted
```

I also embraced scope creep a little, and fixed a few annoying-but-not-user-facing bugs with static files. These are mostly things that generate emails and errors when we're being security scanned, or caused ugly-looking python errors when running locally in debug mode.

## Additions

- a whole bunch of files

## Testing

- As long as pages like http://localhost:8000/jobs/supervision/ don't look busted, we should be fine

## Review

- @anselmbradford @richaagarwal @KimberlyMunoz @sebworks 

## Notes

-

## Todos

- [ ] Adapt cfpb_nemo to point to assets in /static/nemo

## Checklist


richaagarwal:
👍 (though I can't speak to any of the front-end work :o )

cfgov/cfgov/settings/base.py
```diff
@@ -183,9 +183,9 @@
 # various locations.
 STATICFILES_FINDERS = (
     'sheerlike.finders.SheerlikeStaticFinder',
-    'transition_utilities.finders.NoPHPFileSystemFinder',
     'django.contrib.staticfiles.finders.AppDirectoriesFinder',
-    'compressor.finders.CompressorFinder',
+    'django.contrib.staticfiles.finders.FileSystemFinder',
+    #'compressor.finders.CompressorFinder',
```
vccabral: should we cut this line?

.gitignore
```diff
@@ -85,7 +85,7 @@ docs/*/pdfreactor-fonts.html
 #################
 *.egg-info
 .installed.cfg
-static/
+/static/
```
vccabral: why is static in ignore? I don't understand the specifics of how cfgov's static assets are setup but why would we want to not checkin the static dir? Is this the collect static dir?

cfgov/cfgov/settings/base.py
rosskarchner: I re-enabled compressor-- I'm not sure we're using it, but disabling it is out of scope for this PR.

.gitignore
rosskarchner: Pretty sure it's a relic of some Bad Old Days where developers needed to run collectstatic. As written, this rule was ignoreing *every* directory called 'static', which was getting in my way. There may not be a reason for this line at all, since there's not a top-level 'static' directory anyways.

rosskarchner:
I've rebased this on top of #2170, so I can begin work on integrating the two

jimmynotjim:
@rosskarchner do you still need the grunt task converted to gulp?

vickumar1981:
Don't check in static/nemo/.DS_Store.

rosskarchner:
I'm not sure it *needs* done right now, but it might be something to do the next time someone wants to change the CSS or JS of the Spanish sites, or the scripts getting stuffed into scripts.min here: https://github.cfpb.gov/CFPB/cfpb_nemo/blob/master/Gruntfile.js#L67

rosskarchner:
@vickumar1981 fixed!

vickumar1981:
👍  LGTM.

@jimmynotjim   is this cool to merge?

rosskarchner:
My understanding is, this is waiting on #2170 which is waiting on #2166. 

Wheeeeeeeeee!

rosskarchner:
@vickumar1981 merged

PR #2158, opened 2016-05-31
Fix issue that probably came about from renaming site --> request 

richaagarwal:
This fixes an issue we were seeing on several pages that `'Site' object has no attribute 'site'`. @kurtw I assume when you renamed `site` to `request`, this was simply an oversight. https://github.com/cfpb/cfgov-refresh/commit/b30c2ec6c6223f9557aab57588fa529e0c66593c  Tests should all pass now! 

@rosskarchner @jimmynotjim


rosskarchner:
Is this similar to the hotfix that went up last week?

richaagarwal:
Okay, wait, I take it back. I think the deeper issue here is that `get_breadcrumbs` is being called with different types - sometimes request, and sometimes request.site

rosskarchner:
👍

PR #2165, opened 2016-06-03
Django auth

richaagarwal:
We want to consolidate the user creation/permission flow, so this is a first attempt at that.  

We now use the Django default view to create a user, where one can set the appropriate permissions. The password that is set does not have to adhere to the password policy, but it also expires immediately, forcing the user to reset their password when they log in (which is what they should be doing anyway). 

One known issue:
- E-mail addresses do not have a unique constraint, so the admin must ensure that they are not creating multiple users with the same e-mail address. If this happens, there'd be weirdness with resetting the password since that uses the e-mail address to reset, and not the username.

To test:
- Pull down this branch, and log in as a superuser 
- From the Django admin, add a user (with a unique email address). Set password to anything of your liking. Make sure to give the user Staff status at the minimum.
- Log out
- Try to log in with that username and password. Notice you are locked out and need to reset your password.
- Reset the password for the email address you provided. Get the reset password link from the console. 
- Try some non-compliant passwords and note that the policy is enforced.
- Try one that is compliant and reset the password successfully.
- Log in with the username and new password

You can also test that the password policy works when trying to change a password as well. 

@kave 
@rosskarchner 
@kurtw


cfgov/v1/signals.py
```diff
@@ -29,17 +29,13 @@ def new_phi(user):
...
-    try:
-        current_password_data = user.passwordhistoryitem_set.latest()
-    
-        if user.password != current_password_data.encrypted_password:
+    if kwargs['created']:
+        # If user was just created, expire the password and unlock it
+        new_phi(user, expiration_days=0, locked_days=0)
+    else:
+        if user.password != user.passwordhistoryitem_set.latest().encrypted_password:
             new_phi(user)
```
richaagarwal: By the way, getting rid of the `try` block here because it seems like this was only needed to check the condition where the user was a newly created one, but if that's not the case do chime in!

cfgov/v1/auth_forms.py
```diff
@@ -36,14 +36,6 @@ class UserEditValidationMixin(PasswordValidationMixin):
     user_attribute = 'instance'
 
 
-class CFGOVUserCreationForm(UserEditValidationMixin, UserCreationForm):
-    pass
```
kave: @richaagarwal hmm related to the email weirdness.. this is true... lets enforce the email unique here

cfgov/v1/auth_forms.py
```diff
@@ -36,14 +36,6 @@ class UserEditValidationMixin(PasswordValidationMixin):
...
-
-
-class CFGOVUserEditForm(UserEditValidationMixin, UserEditForm):
-    pass
```
kave: and here

cfgov/v1/auth_forms.py
richaagarwal: @kave I'm hesitant to add that back in, as part of the point of this PR is we're going with the Django defaults.  Eventually I think we'll want to upgrade our Django version, where we can hook up the password policies much more seamlessly.

kurtw: We probably shouldn't be breaking those things though. I'm all for using Django as much as we can, but if there's a requirement (especially dealing with users resetting passwords which seems to be a frequent occurrence) then it should remain unbroken.

_waits for counter arguments_

kave: yea this creates more technical debt, and future bugs with user creation which i'd rather avoid. I think its important to enforce the email uniqueness.

richaagarwal: This isn't breaking it; it's already broken.

kurtw: Email uniqueness hasn't been enforced?

richaagarwal: Nope, it's an existing issue I was pointing out.

kurtw: Yeah, @richaagarwal you're right we haven't been enforcing that. I think fixing it is out of scope for this PR.

richaagarwal: Cool, any other issues, or is this OK to be merged?

rosskarchner:
Seems to work, altough somewhere along the way I got sent to the wagtail login /admin/ instead of /django-admin/. I see if I can isolate this.

richaagarwal:
@rosskarchner Yeah, I think that's where we have been redirecting the user after they reset the password. What would you suggest for that?

rosskarchner:
I wonder if we can preserve the value of the 'next'  parameter through the password reset process?

PR #2166, opened 2016-06-03
Adding mobile tests for the header

KimberlyMunoz:
Adding some missing tests to the header to test switching between the search and mega menu.

## Testing

- `gulp test:acceptance --sauce=false --windowSize=400,610 --specs=header.js` 

## Review

- @jimmynotjim 

## Checklist

These tests look at switching between the mega menu and search.

### DIFF
--- a/cfgov/unprocessed/css/misc.less
+++ b/cfgov/unprocessed/css/misc.less
@@ -665,7 +665,7 @@
     - "Adds transitions utilty classes for transform, opacity,
        and for the removing the transition duration."
     - "`.u-move-transition__disabled` is used for disabling tests
-        in protractor tess"
+        in protractor tests"
   tags:
     - cfgov-misc
 */
@@ -681,8 +681,11 @@
     transition-duration: 0s;
 }
 
-.u-move-transition__disabled * {
-  transition: none !important;
+// The mega-menu code doesn't work with a transition of 0ms
+.u-move-transition__disabled {
+  .u-move-transition {
+    transition: transform 1ms ease;
+  }
 }
 
 /* topdoc
--- a/cfgov/unprocessed/css/misc.less
+++ b/cfgov/unprocessed/css/misc.less
@@ -664,6 +664,8 @@
   notes:
     - "Adds transitions utilty classes for transform, opacity,
        and for the removing the transition duration."
+    - "`.u-move-transition__disabled` is used for disabling tests
+        in protractor tess"
   tags:
     - cfgov-misc
 */
@@ -677,6 +679,10 @@
 
 .u-no-animation {
     transition-duration: 0s;
+}
+
+.u-move-transition__disabled * {
+  transition: none !important;
 }
 
 /* topdoc
--- a/test/browser_tests/spec_suites/organisms/header.js
+++ b/test/browser_tests/spec_suites/organisms/header.js
@@ -136,7 +136,7 @@ describe( 'Header', function() {
         } );
 
         describe( 'then click search', function() {
-          it( 'should show the search and hide menu', function() {
+          it( 'should show the search and hide megamenu', function() {
             browser.driver.actions().click( _dom.megaMenuTrigger ).perform();
             browser.driver.actions().click( _dom.globalSearchTrigger ).perform();
             browser.driver.wait( _dom.globalSearchContent.getAttribute( 'aria-expanded' ) )
@@ -157,12 +157,8 @@ describe( 'Header', function() {
                 expect( _dom.globalSearchContent.getAttribute( 'aria-expanded' ) ).toBe( 'false' );
                 return isTrue == 'true'
               } );
-
         } );
       } );
-
     } );
   }
 } );
-
-//expect( _dom.globalSearch.getAttribute('aria-expanded') ).toBe( true );
--- a/test/browser_tests/spec_suites/organisms/header.js
+++ b/test/browser_tests/spec_suites/organisms/header.js
@@ -43,6 +43,7 @@ describe( 'Header', function() {
 
   beforeEach( function() {
     browser.get( '/' );
+    browser.executeScript('document.body.className = "u-move-transition__disabled";');
   } );
 
   if ( browser.params.windowWidth > breakpointsConfig.bpLG.min ) {
@@ -139,11 +140,7 @@ describe( 'Header', function() {
           it( 'should show the search and hide megamenu', function() {
             browser.driver.actions().click( _dom.megaMenuTrigger ).perform();
             browser.driver.actions().click( _dom.globalSearchTrigger ).perform();
-            browser.driver.wait( _dom.globalSearchContent.getAttribute( 'aria-expanded' ) )
-              .then( function( isTrue ) {
-                expect( _dom.megaMenuContent.getAttribute( 'aria-expanded' ) ).toBe( 'false' );
-                return isTrue == 'true'
-              } );
+            expect( _dom.megaMenuContent.getAttribute( 'aria-expanded' ) ).toBe( 'false' );
           } );
         } );
       } );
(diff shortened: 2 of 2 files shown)


test/browser_tests/spec_suites/organisms/header.js
```diff
@@ -135,11 +134,31 @@ describe( 'Header', function() {
...
+
+              } );
+          } );
+        } );
+      } );
+
+      describe( 'click search, then click mega menu', function() {
+        it( 'should show the megamenu and hide search', function() {
+          browser.driver.actions().click( _dom.globalSearchTrigger ).perform();
+          browser.driver.wait( _dom.megaMenuContent.isDisplayed() );
```
jimmynotjim: Should this be search content is displayed instead?

test/browser_tests/spec_suites/organisms/header.js
```diff
@@ -135,11 +134,31 @@ describe( 'Header', function() {
...
         } );
+
+        describe( 'then click search', function() {
+          it( 'should show the search and hide menu', function() {
+            browser.driver.actions().click( _dom.megaMenuTrigger ).perform();
+            browser.driver.wait( _dom.megaMenuContent.isDisplayed() );
+            browser.driver.actions().click( _dom.globalSearchTrigger ).perform()
+              .then( function() {
+                expect( _dom.globalSearch.isDisplayed() ).toBe( true );
+
```
anselmbradford: Trivial: extra line could be removed.

test/browser_tests/spec_suites/organisms/header.js
```diff
@@ -135,11 +134,31 @@ describe( 'Header', function() {
               expect( _dom.overlay.isDisplayed() ).toBe( true );
             } );
         } );
+
+        describe( 'then click search', function() {
+          it( 'should show the search and hide menu', function() {
```
anselmbradford: Trivial: Make this the inverse of L:152, so `should show search and hide the megamenu`

test/browser_tests/spec_suites/organisms/header.js
anselmbradford: > EDIT: Which makes me wonder if wait is even doing anything or can be removed from all these tests

This is a tricky one because the `wait` line could be removed and the test would still pass because it's checking that the mega menu shows after clicking the trigger, which will always be true (if the menu is working). Ideally it should be testing that the search is displayed, then after clicking the menu trigger, check that the megamenu is displayed and search is no longer displaying.

KimberlyMunoz: I wrote it this way because the old bug from v1 launch was that the megamenu wouldn't display after hitting search on certain browsers. But let me see about adding in search no longer displaying.

jimmynotjim: But you're opening the search, waiting for the menu, then opening the menu. That should be open search, wait for search, then open the menu. That's why I'm not so sure the wait is doing anything here and why it might not be doing anything elsewhere. It should be timing out, because the menu was never triggered and therefore wouldn't ever return that it's content is being displayed.

anselmbradford: `wait` is waiting for a promise to resolve, so it probably needs:

```
browser.driver.wait( _dom.megaMenuContent.isDisplayed() ).then( function() {
  …
} )
```

KimberlyMunoz: I was looking at the wrong lines, now I see it. So I tried testing something, it looks like 

```expect( _dom.globalSearch.isDisplayed() ).toBe( true );
 expect( _dom.megaMenuContent.isDisplayed() ).toBe( true );
```

always resolve as true when I test on smaller screens. I haven't been able to write a test where that's false. I'm going to read documentation to see if there's another test I can use, but does anyone have ideas?

anselmbradford: This is what I was running into as well. `isDisplayed()` has cases where it will return `true` even when the element is offscreen and visible (in absolute positioned elements, and, it seems, with elements [that have tranform applied](https://github.com/seleniumhq/selenium-google-code-issue-archive/issues/8054#issuecomment-192200898)). See some background here: https://github.com/angular/protractor/issues/759

With the depressing conclusion:
> sometimes elements that you don't think are displayed are being reported as displayed. I think this is because WebDriver is not smart enough to cover all cases of elements that don't have display:false but are not on the screen for some reason. If you can find a reduced case showing one of these issues we can report to WebDriver, but otherwise there's nothing Protractor can do.

KimberlyMunoz: I've been playing around with `expect( _dom.globalSearchContent.getAttribute( 'aria-expanded' ) ).toBe( 'true' );` this morning, but it's tricky to get that into a wait promise.

KimberlyMunoz: I tried fixing it up in fbe92b4. Try kicking the tires on it @anselmbradford @jimmynotjim

test/browser_tests/spec_suites/organisms/header.js
KimberlyMunoz: Fixed in 8ba6540

test/browser_tests/spec_suites/organisms/header.js
KimberlyMunoz: Fixed in 8ba6540

KimberlyMunoz:
I worked on this with @contolini and I don't think our promises are doing much. So instead, we add a css class to testing environments to turn off transitions and avoid the need to pepper our tests with `.wait`s.

jimmynotjim:
It's so much faster!!! 👍 

Not sure if @anselmbradford is available to sign off, maybe @contolini or @Scotchester can run the test and give a second +1

cfgov/unprocessed/css/misc.less
```diff
@@ -664,6 +664,8 @@
   notes:
     - "Adds transitions utilty classes for transform, opacity,
        and for the removing the transition duration."
+    - "`.u-move-transition__disabled` is used for disabling tests
+        in protractor tess"
```
KimberlyMunoz: *tests

test/browser_tests/spec_suites/organisms/header.js
```diff
@@ -135,11 +135,23 @@ describe( 'Header', function() {
               expect( _dom.overlay.isDisplayed() ).toBe( true );
             } );
         } );
+
+        describe( 'then click search', function() {
+          it( 'should show the search and hide megamenu', function() {
+            browser.driver.actions().click( _dom.megaMenuTrigger ).perform();
+            browser.driver.actions().click( _dom.globalSearchTrigger ).perform();
+            expect( _dom.megaMenuContent.getAttribute( 'aria-expanded' ) ).toBe( 'false' );
+          } );
```
imuchnik: Do we need an assert that search is displayed?

test/browser_tests/spec_suites/organisms/header.js
```diff
@@ -135,11 +135,23 @@ describe( 'Header', function() {
...
 
-      // TODO: Add tests for clicking between menu and search.
-      // describe( 'click search when mega menu is showing', function() {} );
-      // describe( 'click mega menu when search is showing', function() {} );
+      describe( 'click search, then click mega menu', function() {
+        it( 'should show the megamenu and hide search', function() {
+          browser.driver.actions().click( _dom.globalSearchTrigger ).perform();
+          browser.driver.actions().click( _dom.megaMenuTrigger ).perform()
+          expect( _dom.globalSearchContent.getAttribute( 'aria-expanded' ) ).toBe( 'false' );
+        } );
```
imuchnik: Same comment here as above. Do we want to assert the mega menu is displayed?

KimberlyMunoz:
When I added the tests @imuchnik suggested and we discovered a weird bug!

image

When you disable the animation by setting to 0ms, the mega menu code starts acting incredibly buggy and not hiding certain things. It becomes impossible to dismiss the search menu (the x never appears) and it's just generally weird. I've updated it to 1ms transition (from 500ms originally) and included notes about it.

sebworks:
👍

PR #2169, opened 2016-06-06
Fix event livestream

richaagarwal:
This fixes an issue we were seeing where the event stream wasn't going live until an hour after the start date of the event.

We are now using the event's go live date instead of the event's start time, treating it as an ET time, and comparing it to the current time in ET to determine if it's past, present, and future (along with the end time).

Eventually we will need to clean up the app from converting things to ET but that relies on a fix here: https://github.com/torchbox/wagtail/issues/2406

To test:
Edit an event page to have a livestream, and set the start time to the current time in ET (or close to it). Confirm that it shows the livestream if it's past and does not if it's future.

Also, you can run the test:
`Run tox -- v1.tests.test_event_state`

@jimmynotjim @kave @kurtw @rosskarchner


cfgov/jinja2/v1/events/event.html
```diff
@@ -1,8 +1,8 @@
 {% extends 'layout-2-1-bleedbar.html' %}
 {% import '_vars-events.html' as vars with context %}
 {% import 'templates/streamfield-sidefoot.html' as streamfield_sidefoot with context %}
-{% if page.start_dt and page.end_dt %}
-    {% set event_state = when(page.start_dt | date('%c'), page.end_dt | date('%c')) %}
+{% if page.live_stream_date and page.end_dt %}
+    {% set event_state = when(page.live_stream_date, page.end_dt) %}
```
kurtw: Why are we using `live_stream_date` now?

jimmynotjim: Because the stream goes live before the event, and the only thing event_state is used for is checking if it should show the live stream during the event or the saved video after the event.

jimmynotjim:
Tested locally and working 👍

cfgov/jinja2/v1/events/event.html
kave: @jimmynotjim not exactly event_state is also used to determine event description text https://github.com/cfpb/cfgov-refresh/blob/45fb6f96ce4f760bbbe0fb1a10708e9474ce21ec/cfgov/jinja2/v1/events/_event-detail.html#L31

this will effect events that don't have live streams

kave: We may need to have a separate variable called `live_stream_event_state` and revert this particular change.
Also a frequent content user error is to check `live_stream_availability` and not enter a date then they end up with 500 errors when the template looks for `page.live_stream_date`. 
so either we'll need to wrap the  `page.live_stream_date` in a conditional that checks for  `live_stream_availability` first or add a form validation check

kurtw: Yeah we'll need to change this back to `start_date` and do something like what @kave suggested.

richaagarwal: Okay, that sounds good.

richaagarwal: Wait, @kave, what's a scenario that an event needs the event state, but doesn't have a livestream?

rosskarchner:
Success! 👍

richaagarwal:
Ready for one last review!

kurtw:
👍  merge away!

kave:
👍

PR #2164, opened 2016-06-02
Updating mobile tests and adding mega menu tests

KimberlyMunoz:
This updates failing tests at the small breakpoint for filters and adds
tests for the mega menu at small breakpoints. 

Finishes up with a failing test for keyboard tabbing on mobile. Ideally,
tabbing should trap focus on a menu until something is selected.

## Testing 

`gulp test:acceptance --sauce=false --windowSize=400,610`

## Review

- @jimmynotjim 

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #2160, opened 2016-06-01
Created non-responsive on-demand styles for non-v1 sections of the site.

jimmynotjim:
As part of #2156, we need a stylesheet without the small screen styles for limited sections of the site that aren't yet updated. Updated the ondemand gulp task to write a separate nonresponsive stylesheet.

## Changes

- Updated gulp task to write both responsive and non-responsive styles for ondemand needs.
- Updated the test fixture for ondemand includes to allow for the nonresponsive stylesheet to be loaded for visual testing.

## Testing

- `gulp styles:ondemand` and navigate to `http://localhost:8000/test-fixture/?atomic=header&suffix=nonresponsive`. Header should keep the large screen styles (though the menu will still go offscreen due to JS, upcoming fix for that).

## Review

- @KimberlyMunoz 
- @rosskarchner 
- @richaagarwal 

## Screenshots

screen shot 2016-06-01 at 12 12 10 pm

## Notes

- It's looking like we'll need to separate out the small screen vs large screen menus to make this simpler. Getting the JS to work with both, but only show the large screen behaviors on these older pages is proving to be extremely complicated. If anyone has objections or other ideas, please chime in cc @anselmbradford 

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #2170, opened 2016-06-07
Creates nonresponsive header script for non-v1 sections of the site

jimmynotjim:
PR #2160 only addressed non-responsive styles, leading to incorrect behavior at smaller screens on the non-responsive test fixture. Added a gulp task to create a nonresponsive script that rewrites `isInDesktop` to always return true.

## Additions

- Added gulp task for nonresponsive scripts
- Added webpack config for nonresponsive gulp task

## Changes

- Moved `isInDesktop` to breakpoint state file to resolve TODOs

## Testing

- `gulp scripts:ondemand && gulp styles:ondemand`
- navigate to `http://localhost:8000/test-fixture/?atomic=header&suffix=nonresponsive` 

## Review

- @KimberlyMunoz 
- @anselmbradford 
- @rosskarchner 

## Screenshots

screen shot 2016-06-07 at 8 49 22 am

# Notes

- We should wait for #2164 & #2166 to be merged to make sure this PR doesn't break the existing functionality

## Checklist

### DIFF
--- a/config/webpack-config.js
+++ b/config/webpack-config.js
@@ -66,8 +66,19 @@ var onDemandConf = {
   ]
 };
 
+var onDemandHeaderRawConf = {
+  context: path.join( __dirname, '/../', paths.unprocessed,
+                      JS_ROUTES_PATH + '/on-demand' ),
+  entry:   './header.js',
+  output: {
+    path:     path.join( __dirname, 'js' ),
+    filename: '[name]'
+  }
+};
+
 module.exports = {
-  onDemandConf: onDemandConf,
-  ieConf:       ieConf,
-  modernConf:   modernConf
+  onDemandHeaderRawConf: onDemandHeaderRawConf,
+  onDemandConf:          onDemandConf,
+  ieConf:                ieConf,
+  modernConf:            modernConf
 };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 
 ### Added
 
+- Added nonresponsive header script for non-v1 sections of the site
+
 ### Changed
 
 ### Removed
--- a/cfgov/jinja2/v1/test-fixture/index.html
+++ b/cfgov/jinja2/v1/test-fixture/index.html
@@ -118,7 +118,7 @@
     (b) In the <head> with the `defer` and `async` attributes
         set on the <script> tag.
 #}
-<script type="text/javascript" src="{{ static('js/atomic/' + atomic_type + '.js') }}"></script>
+<script type="text/javascript" src="{{ static('js/atomic/' + atomic_type + suffix + '.js') }}"></script>
 
 </body>
 
--- a/cfgov/unprocessed/css/organisms/header.less
+++ b/cfgov/unprocessed/css/organisms/header.less
@@ -76,11 +76,15 @@
     z-index: 10;
     background-color: @white;
 
-    // Desktop size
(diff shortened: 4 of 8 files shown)


rosskarchner:
💯 Looks great

config/webpack-config.js
```diff
@@ -66,8 +66,19 @@ var onDemandConf = {
...
+    path:     path.join( __dirname, 'js' ),
+    filename: '[name]'
+  }
+};
+
 module.exports = {
-  onDemandConf: onDemandConf,
-  ieConf:       ieConf,
-  modernConf:   modernConf
+  nonResponsiveConf: nonResponsiveConf,
```
anselmbradford: I'm leaning toward naming this something like `onDemandUnminifiedConf` so it describes what the config does, not where it's used. There's nothing about the configuration that makes it non-responsive. It's the on-demand scripts, just without minification. Its use in https://github.com/cfpb/cfgov-refresh/pull/2170/files#diff-9d68bb3ecce3a165761fdcb2b7966095R94 is what makes it non-responsive, and that function has a name that conveys that. Haha, naming is the hardest ;)

jimmynotjim: It's also only the header, how does that affect naming?

anselmbradford: `onDemandHeaderRawConf`? Kinda a mouthful, but not terrible. At any rate, aiming to make it more descriptive of what it does vs where it's used would be good, that way if it ends up being used somewhere else it'll still make sense.

rosskarchner:
Do you think we'll be able to merge this this week?

jimmynotjim:
@rosskarchner I think so, just wanted to wait on #2166 to be sure I didn't break anything in this process.

anselmbradford:
👍

rosskarchner:
@jimmynotjim cool

rosskarchner:
Should every non-responsive page include secondary-navigation.nonresponsive.css

jimmynotjim:
@rosskarchner is there any way to load it conditionally? It would be great to only load it when it's needed, but due to the static-ness of those pages, I'm not sure how difficult it'll be to do that.

rosskarchner:
What's the scenario where it's needed?

jimmynotjim:
@rosskarchner honestly, I don't know. Do we have a list of non-responsive url somewhere?

rosskarchner:
Off the top of my head, we have:

/company-signup/
/credit-cards/agreements/
/jobs/supervision/
/jobs/technology-innovation-fellows/
... and some stray wordpress pages

jimmynotjim:
@rosskarchner in that case, I assume it can be removed. @anselmbradford do you know why it was ever added to begin with?

anselmbradford:
@jimmynotjim Why on-demand secondary navigation was added to begin with? It is for GHE/CFGOV/fin-ed-resources (responsive non-v1 pages).

rosskarchner:
so, there's no use case for non-responsive, secondary navigation CSS?

jimmynotjim:
Nope, and it looks like Fin-Ed loads the secondary nav on it's own, so we shouldn't even need to load it globally.

Scotchester:
To elaborate: FinEd gets `secondary-nav.less` from cfgov-sheer-templates, which gets it from cfgov-refresh.

rosskarchner:
Alright, then in my next commit I'll remove the reference from nonresponsive django template

jimmynotjim:
That PR is going to be a bit longer, so let's go ahead and merge this. @rosskarchner Let me know if I need to help getting it implemented on the pages it's required.

PR #2171, opened 2016-06-07
Change page.media to use OrderedDict

kurtw:
`set` makes the JS files list from the unique but since it is unordered it can interfere with the JS instantiation. This uses OrderedDict (since python has no built-in implementation of something like OrderedSet) to get both order and uniqueness.

## Changes

- OrderedDict replaces set in use of maintaining uniqueness and order.

## Testing

- Add ExpandableGroup to a page and test its functionality.

## Review

- @Scotchester


Scotchester:
Works! :+1:

Scotchester:
Will merge after Travis finishes.

PR #2174, opened 2016-06-08
3.0.0-3.3.21 Release

kurtw:
## Added
- Ability to use Social Media molecule as a Wagtail module in the Sidefoot.
- Frontend: Added task for generating JavaScript code docs with `gulp docs`.
- Browser tests for most of the organisms generated by Wagtail
- `cfgov/scripts/_atomic_helpers.py` to provide streamfield data
- Test for present, past, and future event states

### Changes
- Use bare value of RichText field if value type is not RichText.
- Check against Activity Log topics when generating View More link.
- Breadcrumb and sidenav link generation gets most appropriate version of page.
- Made Text Introduction's `has_rule` option have an effect.
- Tidied up some of the template logic around using `render_block`.
- Changed class of FCM category slug to remove extra spacing.
- Updated gulp task to write both responsive and non-responsive styles for ondemand needs.
- Updated the test fixture for ondemand includes to allow for the nonresponsive stylesheet to be loaded for visual testing.
- Promotes Expandables from molecule to organism
- Changes global banner expandable Less to resolve cascade issue
- Maintain order and uniqueness in JS file lists by using `OrderedDict` instead of `set`
- initial_test_data script now uses streamfield block data from a new file called `_atomic_helpers.py`
- Use bare value of RichText field if value type is not RichText
- Check against Activity Log topics when generating View More link
- initial_test_data script now uses streamfield block data from a new file called _atomic_helpers.py.
- `when` now compares datetimes against the current time in ET, not UTC
- `when` optionally uses the event stream's start time, if there is one 

### Removed
- Event RSVP email link button.
- `atomicName` parameter from `checkDom` atomic helper.
- Unused gov delivery view function in jobmanager
- PostPreview organism from streamfield block choices since it was unused
- `convert_to_datetime` since it duplicated logic in `_convert_date`

## Review

- @rosskarchner 
- @kave 
- @richaagarwal 
- @KimberlyMunoz 
- @anselmbradford 
- @jimmynotjim 
- @sebworks


kave:
👍

PR #2176, opened 2016-06-08
Updating imports of optional apps.  Condition flag is no longer used.

vickumar1981:
## Removals

- Removes check for 'condition' flag on optional apps b/c it is no longer used

## Changes

- Updates imports of optional apps

## Review

- @kave @kurtw @rosskarchner


kurtw:
code looks good but I'll defer to @rosskarchner since he set it up

rosskarchner:
👍

kave:
👍

PR #2180, opened 2016-06-09
Login Page Reduction Act

rosskarchner:
This gives us a single page for all logins, removing the need for us (and users!) to keep /admin/login and /django-admin/login straight.

Also:
- provides an intermediary view that sniffs the permissions of the page
you are trying to reach. If you don't have appropriate permissions, then you can
navigate to a page you DO have permissions for (Django admin or wagtail)
- If you log in without requesting a specific page, and you have access to
both Django admin and wagtail, you're presented with a page providing both options
- If you log in without requesting a specific page, and you have access to
ONLY ONE of django admin or wagtail, you're redirected to the one you can
access.

## Testing

- Create a user with ONLY wagtail access (active, NOT staff, and a member of moderator and/or editor group)
 - log in via http://localhost:8000/login/ . You should be sent to the wagtail admin page
 - try to reach /django-admin/ , you should see a page explaining that you don't have access to that, with a link back to wagtail
- Create a user with ONLY django admin access (active, staff, and NOT a member of moderator and/or editor groups)
 - log in via http://localhost:8000/login/ . You should be sent to the django-admin admin page
 - try to reach /admin/ , you should see a page explaining that you don't have access to that, with a link back to django-admin
- Create a user with BOTH django-admin and wagtail access (active, staff, and a member of moderator and/or editor group)
 - log in via http://localhost:8000/login/ . You should see a welcome page offering links to both Wagtail and django admin
- Go ahead and do a password reset, just to confirm that it's not broken

## Review

- @richaagarwal @kurtw @kave @Scotchester 

## Screenshots

screen shot 2016-06-08 at 11 26 24 pm

screen shot 2016-06-08 at 11 37 19 pm

## Todos

- It'd be nice if the new templates had appropriate HTML titles (they all say wagtail login)
- now that most of our validation logic is encapsulated in forms, I wonder if we can use the [built in authentication views](https://docs.djangoproject.com/en/1.9/topics/auth/default/#all-authentication-views)

## Checklist

### DIFF
--- a/cfgov/templates/welcome.html
+++ b/cfgov/templates/welcome.html
@@ -1,13 +1,12 @@
 {% overextends "wagtailadmin/login.html" %}
 {% load i18n admin_static %}
 
-
 {% block furniture %}
     <div class="content-wrapper">
-        <h1>{% trans 'Sorry!' %} <br/>
-        	{% trans 'You do not have permission to access this page.' %}
+        <h1>{% trans 'Welcome' %} <br/>
+        You've logged in to the consumerfinance.gov editorial tools.	
         </h1>
-        <h2>Perhaps you could try:</h2>
+        <h2>Continue on to...</h2>
         <ul>
         {% for name, url in destinations %}
             <li><a href="{{url}}"> {{name}}</a></li>
--- a/cfgov/templates/welcome.html
+++ b/cfgov/templates/welcome.html
@@ -12,6 +12,5 @@
             <li><a href="{{url}}"> {{name}}</a></li>
         {% endfor %}
         </ul>
-        <p>{% trans 'Otherwise, you may want to' %} <a href="/logout/">{% trans 'LOG OUT ' %}</a></p>
     </div>
 {% endblock %}
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -16,6 +16,7 @@ SECRET_KEY = os.environ.get('SECRET_KEY', os.urandom(32))
 # Use the django default password hashing
 PASSWORD_HASHERS = global_settings.PASSWORD_HASHERS
 
+
 try:
     import mysql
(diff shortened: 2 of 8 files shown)


Scotchester:
> It'd be nice if the new templates had appropriate HTML titles (they all say wagtail login)

Darn. If we continue to ride the existing Wagtail admin templates' coattails, there's no way to accomplish this. The extend chain leads back to [this file](https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailadmin/templates/wagtailadmin/skeleton.html), wherein there is no way to override the "Wagtail" in the `<title>` element.

Scotchester:
I'll try to do a more thorough review of this tomorrow.

kurtw:
Nice flow! 👍

richaagarwal:
This is awesome!! Tested locally.  👍

cfgov/templates/welcome.html
```diff
@@ -0,0 +1,17 @@
...
+        <h1>{% trans 'Welcome' %} <br/>
+        You've logged in to the consumerfinance.gov editorial tools.	
+        </h1>
+        <h2>Continue on to...</h2>
+        <ul>
+        {% for name, url in destinations %}
+            <li><a href="{{url}}"> {{name}}</a></li>
+        {% endfor %}
+        </ul>
+        <p>{% trans 'Otherwise, you may want to' %} <a href="/logout/">{% trans 'LOG OUT ' %}</a></p>
```
richaagarwal: Minor thing, but the wording of "Otherwise" doesn't quite make sense anymore since before it was if you had insufficient permissions.

rosskarchner: Do you think the logout link is inappropriate?

rosskarchner: I went ahead an eliminated that line

PR #2188, opened 2016-06-14
Datetimes are now saved in UTC with the eastern timezone

kurtw:
Timezones were saved in UTC but being treated in code as eastern. Now, they are saved as UTC with their hour difference applied as is best practice. For example, before this change 4:00am was saved as 04:00:00 UTC and now 12:00am is saved as 08:00:00 UTC and Django will apply the eastern timezone when appropriate. [See this for more details on the settings.](https://docs.djangoproject.com/en/1.9/ref/settings/#std:setting-TIME_ZONE)

This also changes the block types that are used in EventPage AgendaItemBlock's and ItemIntroduction blocks. Before, they were unnecessarily using DateTimeBlock's and now they use their appropriate block types.

## Additions

- A new migration that changes the block types and migrates the datetimes on the pages to the appropriate values.

## Changes

- The timezone is now `America/New_York` and dates displayed in templates and input into forms will be treated as such.

## Testing

- Preferably start with a fresh database dump
- **Before pulling down**
 - Go to the homepage and see dates in latest updates
 - Go to http://localhost:8000/about-us/events/archive-past-events/fall-community-bank-advisory-council-meeting-in-washington-dc/
 - see the start date and agenda times
- **After pulling down** 
 - run `cfgov/manage.py migrate`
  - Go to the homepage and see dates in latest updates unchanged
  - Go to http://localhost:8000/about-us/events/archive-past-events/fall-community-bank-advisory-council-meeting-in-washington-dc/
  - witness that they haven't been affected on the frontend
  - open a shell
  - run:
```
page = EventPage.objects.get(slug='fall-community-bank-advisory-council-meeting-in-washington-dc')
print page.start_dt
```
  - witness `page.start_dt` is saved as UTC with hour difference applied.
  - create a new EventPage with a future start time and agenda items with times
  - publish and see all the times are correct

## Review

- @kave 
- @richaagarwal 
- @rosskarchner 

## TODO

- Later add the timezone selection for events and logic to handle different timezones

## Checklist

### DIFF
--- a/cfgov/sheerlike/templates.py
+++ b/cfgov/sheerlike/templates.py
@@ -7,6 +7,9 @@ def _convert_date(date, tz):
     if date and isinstance(date, basestring):
         date = parser.parse(date,
                             default=datetime.datetime.today().replace(day=1))
+    if isinstance(date, datetime.datetime) and tz:
+        this_tz = timezone(tz)
+        return date.astimezone(this_tz)
     return date
 
 
--- a/cfgov/sheerlike/templates.py
+++ b/cfgov/sheerlike/templates.py
@@ -4,18 +4,9 @@ from pytz import timezone
 
 
 def _convert_date(date, tz):
-    if date:
-        if isinstance(date, basestring):
-            date = parser.parse(date,
-                                default=datetime.datetime.today().replace(day=1))
-        # TODO: Event times are currently stored as UTC in the database, but they're actually ET times. 
-        # We will need to fix this once Wagtail fixes its handling of TIME_ZONE. 
-        if type(date) in [datetime.datetime, datetime.date]:
-            if isinstance(date, datetime.datetime):
-                pytzone = timezone(tz)
-                if date.tzinfo:
-                    date = date.replace(tzinfo=None)
-                date = pytzone.localize(date)
+    if date and isinstance(date, basestring):
+        date = parser.parse(date,
+                            default=datetime.datetime.today().replace(day=1))
     return date
 
 
--- a/cfgov/sheerlike/templates.py
+++ b/cfgov/sheerlike/templates.py
@@ -9,6 +9,8 @@ def _convert_date(date, tz):
                             default=datetime.datetime.today().replace(day=1))
     if isinstance(date, datetime.datetime) and tz:
         this_tz = timezone(tz)
+        if date.tzinfo == None:
+            return this_tz.localize(date)
         return date.astimezone(this_tz)
     return date
 
--- a/cfgov/v1/migrations/0090_auto_20160609_1603.py
+++ b/cfgov/v1/migrations/0090_auto_20160609_1603.py
@@ -1,0 +1,151 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+import json
+import pytz
+import datetime
+from dateutil import parser
+
+from django.conf import settings
+from django.contrib.contenttypes.models import ContentType
+from django.db import migrations, models
+import wagtail.wagtailcore.fields
+import wagtail.wagtailcore.blocks
+
+
+def convert_datetimes(apps, schema_editor):
+    HomePage = apps.get_model('v1.HomePage')
+    DemoPage = apps.get_model('v1.DemoPage')
+    Page = apps.get_model('wagtailcore.Page')
+    PageRevision = apps.get_model('wagtailcore.PageRevision')
+
+    def specific(page):
+        content_type = ContentType.objects.get_for_id(page.content_type_id)
+        model_class = content_type.model_class()
+        if model_class is None:
+            return page
+        elif isinstance(page, model_class):
+            return page
+        else:
+            return content_type.get_object_for_this_type(id=page.id)
+
+    local = pytz.timezone('America/New_York')
+    def timezone_conversion(dt):
+        if isinstance(dt, basestring):
+            dt = parser.parse(dt)
+        naive = dt.replace(tzinfo=None)
+        localized_dt = local.localize(naive)
+        return localized_dt.astimezone(pytz.utc)
+
+    home = HomePage.objects.filter(slug='cfgov').first()
+    if home:
+        for update in home.latest_updates:
+            if update.block_type == 'posts':
+                for post in update.value:
+                    post['date'] = timezone_conversion(post['date'])
+        home.save()
+        revision = PageRevision.objects.filter(page=home).order_by('-id').first()
+        content = json.loads(revision.content_json)
+        if 'latest_updates' in content:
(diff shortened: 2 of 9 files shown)


richaagarwal:
This looks great. I'm going to test it locally.  

It's worth noting (for anyone else that hasn't been following this closely) that this fixes some of the issues we were seeing before by getting rid of `DateTimeBlock` fields entirely. Just something to be aware of if we ever plan on reintroducing those.

kurtw:
Thanks for pointing that out!

cfgov/sheerlike/templates.py
```diff
@@ -8,8 +8,6 @@ def _convert_date(date, tz):
         if isinstance(date, basestring):
             date = parser.parse(date,
                                 default=datetime.datetime.today().replace(day=1))
-        # TODO: Event times are currently stored as UTC in the database, but they're actually ET times. 
-        # We will need to fix this once Wagtail fixes its handling of TIME_ZONE. 
         if type(date) in [datetime.datetime, datetime.date]:
             if isinstance(date, datetime.datetime):
                 pytzone = timezone(tz)
```
richaagarwal: Don't we also need to update this function so that it's no longer stripping the timezone info at line 17? `date = date.replace(tzinfo=None)`

kurtw: At first I thought it was a good idea to keep it in because eventually we'll want to add the ability to localize times to a specific timezone. After thinking about it a little longer, this should probably go. Thanks for the heads up.

kurtw:
Added a TODO

cfgov/sheerlike/templates.py
```diff
@@ -4,16 +4,9 @@
...
-        if type(date) in [datetime.datetime, datetime.date]:
-            if isinstance(date, datetime.datetime):
-                pytzone = timezone(tz)
-                if date.tzinfo:
-                    date = date.replace(tzinfo=None)
-                date = pytzone.localize(date)
+    if date and isinstance(date, basestring):
+        date = parser.parse(date,
+                            default=datetime.datetime.today().replace(day=1))
     return date
```
richaagarwal: I don't think we have to remove the timezone logic entirely, it's just that we shouldn't be stripping the timezone info.  Here's what I had done in my former PR: ```def _convert_date(date, tz):
    if date:
        if isinstance(date, basestring):
            date = parser.parse(date,
                                default=datetime.datetime.today().replace(day=1))
        if type(date) in [datetime.datetime, datetime.date]:
            if isinstance(date, datetime.datetime):
                pytzone = timezone(tz)
                if not date.tzinfo:
                    date = pytzone.localize(date)
                else:
                    date = date.astimezone(pytzone)

    return date
```

If we are going to get rid of it though, we just need to make sure other functions that call this don't break. `when` in `query.py` relies on this function.

File: cfgov/v1/migrations/0090_auto_20160609_1603.py [hunk 3]
```diff
@@ -0,0 +1,143 @@
...
+        ),
+        migrations.RunPython(convert_datetimes),
+        migrations.AlterField(
+            model_name='eventpage',
+            name='agenda_items',
+            field=wagtail.wagtailcore.fields.StreamField([(b'item', wagtail.wagtailcore.blocks.StructBlock([(b'start_time', wagtail.wagtailcore.blocks.TimeBlock(required=False, label=b'Start')), (b'end_time', wagtail.wagtailcore.blocks.TimeBlock(required=False, label=b'End')), (b'description', wagtail.wagtailcore.blocks.CharBlock(max_length=100, required=False)), (b'location', wagtail.wagtailcore.blocks.CharBlock(max_length=100, required=False)), (b'speakers', wagtail.wagtailcore.blocks.ListBlock(wagtail.wagtailcore.blocks.StructBlock([(b'name', wagtail.wagtailcore.blocks.CharBlock(required=False)), (b'url', wagtail.wagtailcore.blocks.URLBlock(required=False))], required=False, icon=b'user')))]))], blank=True),
+        ),
+    ]
```
richaagarwal: Does this migration also take care of the `approved_go_live_at` and `expired_at` dates?

cfgov/sheerlike/templates.py
kurtw: Sure, we could put that in there. I just removed it for now because we don't have that issue since django is now rendering these datetimes in the templates correctly and we currently don't have a way to handle different timezones. `when` works fine with the change. If you want to add a commit that changes `_convert_date()` to the above code then I'm cool with it.

cfgov/v1/migrations/0090_auto_20160609_1603.py
kurtw: It does not! 😦  Thanks!

kurtw: Those are the only other user entered dates.

kurtw: Well `approved_go_live_at` and `go_live_at` are the same date just different fields. There's talk of removing the latter.

rosskarchner:
I can confirm the testing instructions work.

rosskarchner:
In order to get times to appear correctly on the leadership calendar (per #2198), I tweaked _convert_date  (in sheerlike/templates.py) to do timezone conversion. It seems like maybe it was always supposed to do that (why else would it take a timezone argument)?

```python
def _convert_date(date, tz):                                                     
    if date and isinstance(date, basestring):                                    
        date = parser.parse(date,                                                
             default=datetime.datetime.today().replace(day=1))    
             if isinstance(date, datetime.datetime) and tz:                               
                 this_tz = timezone(tz)                                                   
                 return date.astimezone(this_tz)                                          
         return date  
```

(the `if isinstance` block is what I added)

I think this PR is the place to make such a change

cfgov/sheerlike/templates.py
richaagarwal: Thanks, I can confirm that everything is working as expected with the event livestream. @rosskarchner's change essentially does what I mentioned here, but you're right that `when` works without it. 

However, it's weird - it only works without it because somehow the page's event times being passed in to `when`, which are stored in UTC, are passed in as ET. Do you know why that might be? It's called in `event.html` and if you put in a debug statement in the first line of `when`, you'll see that `starttime` or `streamtime` are in ET.  Just curious; perhaps that's a feature of the whole TIME_ZONE setting.

richaagarwal:
Thanks for the hard work on this @kurtw ! It looks like all the issues have been addressed, so 👍

richaagarwal:
Oh, and probably the CHANGELOG should be updated.

kurtw:
Haha I always do that last now since it ALWAYS gets updated. Updating then merging!!

PR #2189, opened 2016-06-14
Fix scheduled publishing

kurtw:
**This should probably go in after the timezone PR is merged #2189**

Scheduled publishing was broken because we changed the `route` function to point to the latest revision when publishing instead of the page itself. When running the management command `publish_scheduled_pages`, it would publish the revisions that were appropriate to publish but it wouldn't hit our logic to update the revision to `live=True` so the changes are never seen. This fixes that by making it hit our logic when calling `publish` on a revision anywhere using signal passing.

## Additions

- Signal to call our revision logic to update the revision appropriately.
- New function to flush akamai

## Removals

- [Here](https://github.com/cfpb/cfgov-refresh/pull/2189/files#diff-44507cc860654b805baac659b58d2c53L64) I removed the `publish()` call from this function since it doesn't provide any value and caused infinite loops when introducing the signal handler.

## Changes

- Every time a revision of a page is published, its revision is updated instead of only when a page is created or edited.
- Akamai is flushed for every page publish call

## Testing

- Make sure the regular workflow of saving, sharing, and publishing a page works.
- Publish a page
- Make an edit and set the Go Live At datetime to be just after the current time
- Publish it
- Note the status of SCHEDULED
- Go to the page and note that the edits have not been published
- once the Go Live At datetime that you set is now or greater, run `cfgov/manage.py publish_scheduled_pages`
- Visit the page again and see the changes. The admin should also now say that the page's status is LIVE

## Review

- @kave 
- @rosskarchner 
- @richaagarwal 

## Todos

- At some point we really need to rethink how we're handling our Shared vs Live logic since it doesn't fall in line with Wagtail's method of serving appropriate versions of pages.

## Checklist

### DIFF
--- a/cfgov/v1/wagtail_hooks.py
+++ b/cfgov/v1/wagtail_hooks.py
@@ -1,6 +1,7 @@
 import os
 import json
 from urlparse import urlsplit
+from django.utils import timezone
 
 from django.conf import settings
 from django.http import Http404
@@ -22,8 +23,14 @@ def share_the_page(request, page):
     is_sharing = bool(request.POST.get('action-share', False))
 
     check_permissions(parent_page, request.user, is_publishing, is_sharing)
-    share(page, is_sharing, is_publishing)
-    configure_page_revision(page, is_sharing, is_publishing)
+
+    is_live = False
+    if is_publishing and not (page.go_live_at and page.go_live_at > timezone.now()):
+        is_live = True
+
+    share(page, is_sharing, is_live)
+    configure_page_revision(page, is_sharing, is_live)
+    flush_akamai(page, is_live)
 
 
 def check_permissions(parent, user, is_publishing, is_sharing):
@@ -33,8 +40,8 @@ def check_permissions(parent, user, is_publishing, is_sharing):
         is_sharing = is_sharing and parent_perms.can_publish()
 
 
-def share(page, is_sharing, is_publishing):
-    if is_sharing or is_publishing:
+def share(page, is_sharing, is_live):
+    if is_sharing or is_live:
         page.shared = True
         page.has_unshared_changes = False
     else:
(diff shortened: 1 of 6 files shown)


richaagarwal:
I wasn't aware of this particular issue with scheduled publishing, but I believe the entire process still won't run exactly as expected because of the issue with the timezones. The `approved_go_live_at` date is currently going to be off by 4 hours, as well as the `expired_at` date. Your other PR should take care of this, and I don't think they need to go in in any particular order, but just wanted to make sure we're on the same page.

kurtw:
Sure, that's why the beginning of the PR description is `This should probably go in after the timezone PR is merged #2189`

richaagarwal:
lol, my bad 😁

rosskarchner:
Confession: I don't know how scheduled publishing is supposed to work

kurtw:
Hmm I'll try to explain @rosskarchner. Whenever an editor wants to schedule a page, they put a datetime in the Go Live At field in the Configuration tab and hit _Publish_. That sets the page's status to SCHEDULED. 

When you run `manage.py publish_scheduled_pages`, it looks at all the revisions that have a `approved_go_live_date <= datetime.now()` and then runs `publish()` on each revision of the set. (`revision.approved_go_live_date` is gotten from `page.go_live_date` and there's talk of removing it)

kurtw:
What broke: vanilla wagtail goes straight to the page object if the page is `LIVE` to get the content to serve. Since we need to hold multiple servable states, we need to use the revisions in order to point to the right version of the page because the single page object can't hold the content for both states at the same time.

When publish() is run via the management command, it doesn't know to hit our custom code to update the revision object that's supposed to be served because it thinks it just has to update the page. 

What this code does is break out the logic that updates the revision so we can call it **whenever** published is called and tie it to the signal "page_published".


Hope this helps.

richaagarwal:
👍

richaagarwal:
(after merge ) :)

cfgov/v1/wagtail_hooks.py
```diff
@@ -42,28 +43,33 @@ def share(page, is_sharing, is_publishing):
...
+# is set to True and return that revision as a page object to serve the request
+# so here we configure the latest revision to fall in line with that logic.
+#
+# This is also used as a signal callback when publishing in code or via
+# management command like publish_scheduled_pages.
+def configure_page_revision(page, is_publishing):
     latest = page.get_latest_revision()
     content_json = json.loads(latest.content_json)
     content_json['live'] = is_publishing
     content_json['shared'] = is_sharing or is_publishing
```
kave: `is_sharing` param was removed

richaagarwal: @kave Could you elaborate?

kurtw: Turns out when this was rebased [this line](https://github.com/cfpb/cfgov-refresh/pull/2189#discussion-diff-69183927R56) changed. Updating now.

kurtw:
Bumping for final review!

richaagarwal:
👍

richaagarwal:
@kurtw Can this be merged in now?

PR #2193, opened 2016-06-15
Leadership calendar error handling

kave:
@rosskarchner

during any failures on PDFReactors part, it should at minimum redirect to the filtered print page


rosskarchner:
In #2198 I removed the whole /print/ URL, but made /pdf/ fail in a similar way-- if the PDF can not be generated, then, it will return the rendered HTML.

rosskarchner:
I'm gonna close this for now.

Issue #2199, opened 2016-06-17
Page number form in filterable list pagination molecule drops parameters


kurtrwall:
## Current behavior
- The page number field in the pagination molecule has a `Go` button next to it to go to the page specified in the input. The issue is [here](https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/jinja2/v1/_includes/molecules/pagination.html#L63), where the parameters from the filter are made into hidden input fields. If you select multiple options of the same vein, say "Category", then only the last parameter in the URL will be used.
- Much less worrisome but still a bug, the last page visited number is maintained in the URL if using that "Go" form.
## Expected behavior
- All parameters should be maintained when going to different pages using the "Go" form.
- The page parameter in the URL should be singular and be the current page visited.
## Steps to replicate behavior
1. Select two or more "Categories" on Blog and filter the results.
2. Scroll to the bottom where the pagination molecule is and increment the number and hit "Go"
3. Note that the URL only contains the last category selected, it's not selected in the filter anymore, and there are 2 page parameters in the URL

@anselmbradford was the one that first reported the duplicated `page` parameter.

anselmbradford:
Reproduced in https://[GHE]/CFGOV/platform/issues/2681

PR #2201, opened 2016-06-20
Updates pagination to support on-demand generation + stepper

anselmbradford:
Updates pagination to support on-demand generation and adds a stepper variant as mentioned in [GHE]/flapjack/Modules-V1/issues/122

## Changes

- Updated pagination to support "stepper" modifier.
- Updated pagination to work in on-demand workflow.
- Updated pagination to support multiple pagination molecules on a single page.
- Moved pagination tests from /organisms/ to /molecules/ where they belong.
- Converted `m-pagination_submit-btn` id to class. 

## Testing

- `gulp styles:ondemand`
- visit /test-fixture/?atomic=pagination to see regular pagination component.
- visit /test-fixture/?atomic=pagination.is_stepper to see stepper variant.
- visit /blog/ and check the pagination with and without filter work correctly.
- `gulp test:acceptance --sauce=false` should pass.

## Review

- @kurtw 
- @Scotchester 
- @sebworks 
- @jimmynotjim 
- @KimberlyMunoz 

## Screenshots

screen shot 2016-06-17 at 2 45 10 pm

screen shot 2016-06-17 at 2 45 21 pm

screen shot 2016-06-17 at 2 45 34 pm

screen shot 2016-06-17 at 2 45 44 pm


kurzn:
@anselmbradford we don't want to ever display the versions that let users "jump" to a specific page number. So we would only ever have ones with "previous/next" 

Also the text should be "Step" instead of "Page"

anselmbradford:
@kurzn Cool, updated to step:

screen shot 2016-06-20 at 11 43 23 am

kurzn:
All good here, then!

kurtw:
Code looks look but why is the ondemand template file needed?

sebworks:
Should we have a base iterator class or module?  Child stepper and paginator modules? Seems like we no longer are using semantic classes if a stepper is called a paginator

cfgov/jinja2/v1/_includes/molecules/pagination.html
```diff
@@ -1,64 +1,92 @@
...
+                Default is 0.
 
    ========================================================================== #}
 
-{% if posts and posts.paginator.num_pages > 1 %}
+{% macro render(total_pages, curr_page, is_stepper=false, fragment_id='', index=0) %}
+{% if total_pages > 1 and curr_page <= total_pages %}
 {% from 'macros/util/url_parameters.html' import url_parameters %}
-
+{% set fragment_id = '#' + fragment_id if fragment_id else '' %}
```
sebworks: Can we separate the logic into paginator and stepper macros? Seems like we are jamming a lot of logic into the template.

cfgov/jinja2/v1/_includes/molecules/pagination.html
```diff
@@ -1,64 +1,92 @@
...
          aria-label="Pagination">
-        {%- if posts.number > 1 %}
-        <a class="btn btn__super m-pagination_btn-prev"
-           href="?page={{ posts.previous_page_number()
-                          ~ url_parameters(request.GET)
-                          ~ '#o-filterable-list-controls' }}">
-        {%- else %}
+        {%- if curr_page > 1 %}
+        <a class="btn
+                  btn__super
```
jimmynotjim: This is all the same as the `else` condition. Can we limit duplication and put the condition on printing the `href`?

cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
```diff
@@ -181,7 +181,7 @@
 
 {% macro render(controls, index) %}
     <div class="o-filterable-list-controls"
-         id="o-filterable-list-controls">
+         id="o-filterable-list-controls-{{ index | string }}">
```
jimmynotjim: This is going to break the filter that relies on this `id` to load to this section the page after filtering.

cfgov/unprocessed/css/molecules/pagination.less
```diff
@@ -92,6 +92,7 @@
 
     label,
     &_label {
+        line-height: @pagination-height__px;
```
jimmynotjim: Did you test this in Firefox? I always run into trouble when trying to use line-height like this

cfgov/unprocessed/css/on-demand/pagination.less
```diff
@@ -0,0 +1,40 @@
...
+@import (reference) "../cf-theme-overrides.less";
+
+// Enhancements that are on a migration path to being added to CF.
+// CF Core
+@import (reference) "../enhancements/vars.less";
+@import (reference) "../enhancements/media-queries.less";
+@import (less) "../enhancements/base.less";
+@import (less) "../enhancements/layout.less";
+
+
```
jimmynotjim: I'm a bit worried about the number of on-demand components we're using. Each duplicates a lot of CF in individually delivered files increasing page load exponentially. Is there some way we can combine all the on-demand stuff? If you're loading the pagination, you're going to need the header/footer. If you're not loading the pagination, that little extra bit isn't going to even be noticeable compared to all the duplication we're creating. We shouldn't have to worry about polluting existing styles because we've prefixed everything with atomic type.

cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
anselmbradford: @jimmynotjim The pagination on L:228 is passed the same fragment identifier with this index set. Is there some other reason this would not work? As is there will be problems if there are ever two instances of filterable-list-controls on the same page. Note Kurt and I found https://github.com/cfpb/cfgov-refresh/issues/2199 when testing, but that is an existing issue.

richaagarwal:
Can this be merged or closed?

anselmbradford:
Can we give it another week so I have time to re-visit after re-group?

richaagarwal:
Yup!

sebworks:
@anselmbradford, is this ready for review?

anselmbradford:
@sebworks It isn't, or rather it won't be different from what you've seen.

cfgov/jinja2/v1/_includes/molecules/pagination.html
anselmbradford: There's also the `btn__disabled` class, but yeah they could probably be combined. I have this open again in https://github.com/cfpb/cfgov-refresh/pull/2267 but can't test your update at the moment. It think it would be a useful refactor though.

cfgov/unprocessed/css/on-demand/pagination.less
anselmbradford: We can't unfortunately guarantee that a component will appear with another component (like a Salesforce app may use the header/footer, but a custom Salesforce component that looks like pagination but isn't). It might work better to see if we can get a css redundancy remover to work. When I tried that though it mangled things.

anselmbradford:
Closing stepper portion of this for now till there's bandwidth to add it here or in CF. Stepper is used in consumer complaint intake form. Pagination parts are moved to https://github.com/cfpb/cfgov-refresh/pull/2267

PR #2202, opened 2016-06-21
Fixing TestMinimumPasswordAge.test_too_soon

rosskarchner:
TestMinimumPasswordAge.test_too_soon assumed that new users were created with a password that can't be changed until 24 hours later. That assumption is no longer valid, so I've altered the test to make certain that the user is not allowed to change their password.

### DIFF
--- a/cfgov/v1/tests/test_password_policy.py
+++ b/cfgov/v1/tests/test_password_policy.py
@@ -1,3 +1,5 @@
+import datetime
+
 from django.test import TestCase
 from django.core.exceptions import ValidationError
 from django.utils import timezone
@@ -53,6 +55,11 @@ class TestMinimumPasswordAge(TestWithUser):
 
     def test_too_soon(self):
         user = self.get_user()
+        history_item = user.passwordhistoryitem_set.latest()
+        current_locked_until = history_item.locked_until
+        new_locked_until = current_locked_until.replace(year=datetime.MAXYEAR)
+        history_item.locked_until = new_locked_until
+        history_item.save()
         with self.assertRaises(ValidationError):
             password_policy.validate_password_age(user)
 


cfgov/v1/tests/test_password_policy.py
```diff
@@ -53,6 +55,11 @@ class TestMinimumPasswordAge(TestWithUser):
 
     def test_too_soon(self):
         user = self.get_user()
+        history_item = user.passwordhistoryitem_set.latest()
+        current_locked_until = history_item.locked_until
+        new_locked_until = current_locked_until.replace(year=datetime.MAXYEAR)
+        history_item.locked_until = new_locked_until
+        history_item.save()
         with self.assertRaises(ValidationError):
```
richaagarwal: I think that this test, as is, lacks enough context to be meaningfully testing anything. We're now setting it up so that the user is guaranteed to be locked, and all `can_change_password` does is to check if something is locked or not.  I think we need to be more explicit about what "too soon" means. This probably means a test for a new user, a test for an existing user that recently changed their password, etc. I can take those on, but do you agree this isn't actually testing anything as is?

rosskarchner: Hmm...  I'll have to think about it some more. It seems like "throws a validation error if locked_until is in the future" is a meaningful test, though. 

Maybe the naming is real flaw.

richaagarwal: It sounds like what we want to test is "When an account is locked, a user cannot change their password", but we're still not entirely testing that, because this is calling `validate_password_age`. What if someone removed that validation on changing a password? The test would still pass. But if we want to simply test the validation function itself, then yes, we should be really clear that's what we're doing. That said, all `validation_password_age` does is call `can_change_password`, which is essentially `return timezone.now() > locked_until`, so writing a unit test for that seems a bit overkill as all it's really testing is that timezone.now and the > operator work as intended. 

We may also want to test, "When a user changes their password, they cannot change it again for another 24 hours." etc.  

I hope I am not being too pedantic :) But I do think it's insanely important for our tests to test what we think they test.

rosskarchner:
I don't think it's pedantic-- and I think the tests you describe might be better. Maybe I'm just being defensive about a test I wrote ;)

As it exists here, this test would fail if a change was introduced that altered the behavior of `can_change_password` to no longer correctly takes `locked_until` into account-- which seems like it'd be a thing worth knowing.

richaagarwal:
Okay, I think I'm convinced that it's meaningful, but that we should also write additional tests. I'm going to go ahead and merge so that all our tests are passing again. Thanks! :)

PR #2206, opened 2016-06-23
Setup fixes

cfarm:
Fixes for issues found during project setup, following standalone instructions in INSTALL.md

## Additions

- add command for running `setup.sh` - it was unclear from the written instructions that you *need* to run this as step 1 of the install in both vagrant + standalone scenarios. I found this out for standalone, because the MySQL script fails until you run `setup.sh` first.

## Removals

- duplicate browser-sync dependency. During setup.sh got this warning message:

  ```
  npm WARN package.json Dependency 'browser-sync' exists in both dependencies and devDependencies, using 'browser-sync@2.11.2' from dependencies`
 ```
- setup.sh step in front end section is redundant since i added it in step 1. I kept the note about re-installing, though.

## Changes

- change 'frontend' references in `backend.sh` script to say 'backend'

## Testing

-

## Review

- @user

## Notes

-

## Todos

-

## Checklist

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -69,7 +69,18 @@ cfgov start django
 
 ## Stand-alone installation
 
-### 1. Back-end setup
+### 1. Environment variables setup
+
+The project uses a number of environment variables.
+The `setup.sh` script will create a `.env` file for you
+from the `.env_SAMPLE` file found in the repository,
+if you don't already have one:
+
+```sh
+./setup.sh
+```
+
+### 2. Back-end setup
 
 #### Virtualenv & Virtualenvwrapper Python modules
 
@@ -196,7 +207,7 @@ pip install git+git://github.com/rosskarchner/govdelivery
   in the [Project Configuration](https://github.com/cfpb/cfgov-refresh/blob/flapjack/INSTALL.md#4-project-configuration).
 
 
-## 2. Front-end setup
+## 3. Front-end setup
 
 The cfgov-refresh front-end currently uses the following frameworks / tools:
 
@@ -217,7 +228,7 @@ The cfgov-refresh front-end currently uses the following frameworks / tools:
 npm install -g gulp bower
 ```
 
-## 3. Install dependencies
+## 4. Install dependencies
 
 > **NOTE:**
   Protractor (for the test suite)
@@ -240,7 +251,7 @@ Next, install dependencies with:
   [updating all the project dependencies](README.md#updating-all-dependencies).
 
 
-## 4. Project configuration
+## 5. Project configuration
 
 The project uses a number of environment variables.
 The `setup.sh` script will create a `.env` file for you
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -36,7 +36,6 @@
     "webpack-stream": "3.1.0"
   },
   "devDependencies": {
-    "browser-sync": "2.11.2",
```
anselmbradford: Would removing this in `dependencies` work and keeping it in `devDependencies`?

cfarm: yeah I think either would work, what's the justification for each?

anselmbradford: `devDependencies` don't get installed on Travis, while `dependencies` do. Since `browser-sync` is only used locally, it would be preferable in `devDependencies`.

INSTALL.md
```diff
@@ -65,7 +69,18 @@ cfgov start django
...
 
-### 1. Back-end setup
+### 1. Environment variables setup
+
+The project uses a number of environment variables.
+The `setup.sh` script will create a `.env` file for you
+from the `.env_SAMPLE` file found in the repository,
+if you don't already have one:
+
+```sh
```
jimmynotjim: Is this needed here as well as where you added it above?

package.json
anselmbradford: Yeah! Just check output on Travis. I moved them over one-by-one when `devDependencies` was originally added and for some reason Travis didn't like them all moved over. It would be great to get at the underlying reason why though

Scotchester:
There's one semi-significant issue with this PR: You can't run (well, you can't successfully _complete_) `setup.sh` if you don't have your system-level dependencies installed.

KimberlyMunoz:
> There's one semi-significant issue with this PR: You can't run (well, you can't successfully complete) setup.sh if you don't have your system-level dependencies installed.

Is that an issue with this PR or an existing issue? We should keep the scope small and document outside issues.

Scotchester:
It's an issue with the PR, because her updated doc instructs people to run `setup.sh` before installing those dependencies.

KimberlyMunoz:
That explanation is helpful. So it's an issue with the documentation.

cfarm:
But I updated the documentation because you need to run `setup.sh` because the MySQL script fails until you run the script first. In the current order, the MySQL step comes first.

Scotchester:
@cfarm If one hasn't done the stuff in ["Front end setup"](https://github.com/cfpb/cfgov-refresh/blob/flapjack/INSTALL.md#2-front-end-setup), the first run of `setup.sh` (where you added it at the very start of the Standalone section) will fail.

Scotchester:
I think it will also fail because the virtualenv hasn't been created when the bit at the end of `.env` tries to activate it. Well, it won't fail, but the pip dependencies will be installed globally, which is no bueno.

Scotchester:
Closing in favor of #2217. Please open a new PR to reorganize devDependencies, if someone want to do that work.

PR #2217, opened 2016-06-30
Refactor setup scripts and improve installation guide

Scotchester:
The primary goal of these changes is to make installing as simple as running `setup.sh` after installing the system-level dependencies.

A secondary goal was making sure that `backend.sh` couldn't `pip install` without being in a virtualenv.

There are changes in here that warrant some further explanation. I will go through and make line notes about them, but wanted to go ahead and post this so people can start testing it out.

## Testing

1. Pull down branch.
2. Follow instructions in `INSTALL.md`
3. `./runserver.sh` and visit <http://localhost:8000/>

## Review

- @contolini 
- @cfarm 
- @richaagarwal 
- @kurtw 
- @kave 
- @rosskarchner 
- @chosak 
- @anselmbradford 
- @jimmynotjim 
- @sebworks 
- @KimberlyMunoz 

## Notes

- I abandoned the earlier iteration of this that tried to roll up the virtualenv activation into `back-end.sh`. The instructions now require that you set that up manually, and `backend.sh` will exit if you aren't in a virtualenv.

## Todos

- Take another pass at tidying up the README.

## Checklist


.env_SAMPLE
```diff
@@ -33,15 +33,16 @@ export ALLOW_ADMIN_URL=True
 #export AKAMAI_HOST=<akamai_hostname>
 
 ############################
-# Mysql Database Server.
+# MySQL Database Server
 ############################
 
 export MYSQL_NAME=v1
+export MYSQL_HOST=localhost
```
Scotchester: @contolini and I discovered that `create-mysql-db.sh` required this env var to be set.

.env_SAMPLE
```diff
@@ -33,15 +33,16 @@ export ALLOW_ADMIN_URL=True
...
 export MYSQL_USER=root
 #export MYSQL_PW=<password>
-#export MYSQL_HOST=<hostname>
 #export MYSQL_PORT=<portnumber>
 
+export STORAGE_ENGINE="SET default_storage_engine=MYISAM"
```
Scotchester: We heard it said that this env var needed to be set. Can't remember exactly why.

.env_SAMPLE
```diff
@@ -72,7 +73,7 @@ export SHEER_ELASTICSEARCH_INDEX=content
 
 # Name of the virtualenv created for Sheer in
 # https://github.com/cfpb/cfgov-refresh/blob/flapjack/INSTALL.md#sheer
-export VIRTUAL_ENV=cfgov-refresh
+export VENV_NAME=testing
```
Scotchester: `VIRTUAL_ENV` is set automatically when activating a virtualenv, so we should use a unique name.

Also, now remembering that I should probably set this back to cfgov-refresh before merging this.

kurtw: What's wrong with `cfgov-refresh`? `testing` seems too generic to me. I have a lot of virtualenvs.

kurtw:
There's a python script that crashes my session. Trying to figure out where

.env_SAMPLE
anselmbradford: Hm, I don't remember ever setting this.

INSTALL.md
```diff
@@ -63,207 +63,277 @@ cfgov assets
...
-```
-mysql.server start
-```
+#### Front-end dependencies
 
-Then run the MYSQL creation script from project root directory:
-```
-./create-mysql-db.sh
-```
+The cfgov-refresh front end currently uses the following frameworks / tools:
```
anselmbradford: Trivial: mix of `Front-end` and `front end` in the file.

Scotchester: That's right. As a noun, "front end" is not hyphenated. As an adjective, it is.

Scotchester: Though I grant that I may have been imperfect in attempting to abide by that rule.

INSTALL.md
```diff
@@ -63,207 +63,277 @@ cfgov assets
...
+"flapjack-demo-refresh-mysql-dump".
+You'll get a download of `refresh_dump.sql.gz`; unzip that, then run:
+
 ```bash
-mkvirtualenv cfgov-refresh
+./refesh-data.sh /path/to/refresh_dump.sql
+```
+
+This will remove the initial Wagtail admin user that was created by
+the `initial-data.sh` script that was called by `backen.sh`.
```
anselmbradford: `backend.sh`?

Scotchester: Good catch, thanks.

INSTALL.md
```diff
@@ -63,207 +63,277 @@ cfgov assets
...
+**Continue following the [usage instructions](README.md#usage) in the README.**
+
+
+## Optional steps
+
+### Load a database dump from the Build server
+
+If you're installing this fresh, the initial data you receive will not be
+as extensive as you'd probably like it to be.
+You can get a database dump from the Build server by running the Jenkins job
```
anselmbradford: This could use some more context for where to go to find the Jenkins job.. if that can be placed in public GH.

INSTALL.md
```diff
@@ -63,207 +63,277 @@ cfgov assets
...
+./cfgov/manage.py createsuperuser
 ```
 
-The new virtualenv will activate right away. To activate it later on
-(say, in a new terminal session) use the command `workon cfgov-refresh`.
-You’ll know you have a virtual environment activated if you see the name of it in
-parentheses before your terminal prompt. Ex:
+### Install Protractor locally
+
+Protractor (for the test suite) can be installed globally
```
anselmbradford: Trivial: `test suites` (each describe block is a suite).

.env_SAMPLE
contolini: @richaagarwal discovered this one

INSTALL.md
```diff
@@ -63,207 +63,277 @@ cfgov assets
...
+
 ```bash
-(cfgov-refresh)$
+npm install -g protractor && webdriver-manager update
 ```
 
-#### GovDelivery
+### Install dependencies for working with the GovDelivery API
 
 Install the following GovDelivery dependencies into your virtual environment:
```
anselmbradford: Should this say what GovDelivery is used for?

anselmbradford: nvm, see it says below

INSTALL.md
Scotchester: Yeah, I think we have to leave that as an exercise for the reader :)

richaagarwal:
Is this ready for review?

.env_SAMPLE
richaagarwal: This is another one of those quirks of each of our set-ups... some people need this (I do), but most don't. If you get a storage_engine error, you need this.

kurtw: Essentially it's caused by a difference between MySQL versions and storage engines.

activate-virtualenv.sh
```diff
@@ -0,0 +1,31 @@
...
+# ==========================================================================
+
+# Set script to exit on any errors.
+set -e
+
+# Presumes the varialble VIRTUAL_ENV is already set to the name of the venv
+
+echo 'Activating virtualenv, if not already activated...'
+
+source $(brew --prefix)/bin/virtualenvwrapper.sh
```
chosak: This assumes that this script is being run on a Mac with virtualenvwrapper installed via Brew, which prevents its usage in Vagrantbox and (potentially) Docker-based setups. Any reason this couldn't be changed to remove the reference to `brew --prefix` and just add a documented requirement that virtualenvwrapper be installed and on the default PATH?

activate-virtualenv.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Activate virtualenv, if not already activated
+if ! type "$workon" &>/dev/null; then
```
chosak: This code seems to have been moved from [here](https://github.com/cfpb/cfgov-refresh/blob/flapjack/.env_SAMPLE#L132), but this check seems obsolete now that virtualenvwrapper is always invoked above (if that's the case, `workon` should always be available).

This code seems to have a slight error -- it should say `type "workon"` not `type "$workon"` as the former checks whether `workon` exists (what we want) and the latter checks the value of the environment variable `$workon`. If this variable isn't set (I couldn't find any documentation suggesting that virtualenv or virtualenvwrapper sets this), this test will always pass and setup or activate the virtualenv - not a problem in practice but still extra logic that isn't doing anything.

activate-virtualenv.sh
Scotchester: I _think_ my intent with this line was to handle the case where folks have it installed but haven't added it to their `.bash_profile`, but I can't remember for sure. In any case, I'm moving in the direction of your last suggestion, anyway, to address the other problem of the poor error handling with `source` invocation, so this will go away.

kurtw:
`workon` is a virtualenvwrapper command. You should try to source the activate file for that virtualenv.

.travis.yml
```diff
@@ -22,6 +22,7 @@ install:
   # Project initialization.
   - chmod +x ./setup.sh
   - ./setup.sh test
+  - workon cfgov-refresh
```
willbarton: Why does Travis need to use a virtualenv?

Scotchester:
Ok, paired a bunch with Andy on this on Thursday, and thought we had a solution that let `backend.sh` activate the virtualenv if it wasn't active, but that didn't work out with Travis, so we went back to requiring that as a manual step before running `setup.sh`.

`INSTALL.md` has been updated accoridngly, and this is ready for any final review folks want to do.

chosak:
:+1:

INSTALL.md
```diff
@@ -63,207 +63,286 @@ cfgov assets
 cfgov start django
 ```
 
+
 ## Stand-alone installation
```
willbarton: This title doesn't match the anchor link at the top. Can you change that to #stand-alone-installation?

Scotchester: I believe you are mistaken, sir.

willbarton: Apparently I am. It's weird that the anchor link isn't working for me.

I also wonder, should we reverse the order of stand-alone and vagrant as they appear in the document, since we're not sure if the Vagrant box still works, or even if we want to keep the Vagrant box?

willbarton:
👍

INSTALL.md
Scotchester: I could reverse them, and also add a note to the Vagrant section that it's not currently working, YMMV, etc.?

willbarton:
Ship it!

200_s

INSTALL.md
```diff
@@ -1,242 +1,190 @@
 # Installation and Configuration for cfgov-refresh
 
 There are two ways to install:
- - [Vagrant-box installation](INSTALL.md#vagrant-box-installation)
- - [Stand-alone installation](INSTALL.md#stand-alone-installation)
+- [Stand-alone installation](#stand-alone-installation)
+- [Vagrant-box installation](#vagrant-box-installation)
+  :warning: These instructions are not currently working.
```
vickumar1981: @Scotchester @willbarton   FYI, the vagrant instructions are working.   

The vagrant box doesn't load all the optional applications, however.

willbarton: It's failing for me because it installs MySQL 5.1 and (default_storage_engine)[https://github.com/cfpb/cfgov-refresh/blob/refactor-setup/.env_SAMPLE#L45] is a setting that was introduced in 5.5.

Scotchester:
@willbarton and I agreed to move forward with this and fix Vagrant stuff in another PR.

PR #2218, opened 2016-06-30
Hotfix: stop trying to process invalid URL's with PDFReactor

rosskarchner:
This is a hotfix that should stop *some* of the bleeding with PDFReactor and our web servers.

We now reject zip+4 codes, because the backend API doesn't support them. PDFReactor doesn't seem to deal well with parsing our error pages.

This also fixes a bug where we were sending error responses with HTTP status 200 that may have been aggravating things.

### DIFF
--- a/cfgov/legacy/forms.py
+++ b/cfgov/legacy/forms.py
@@ -3,4 +3,4 @@ from localflavor.us.forms import USZipCodeField
 
 
 class HousingCounselorForm(forms.Form):
-    zip = USZipCodeField()
+    zip = USZipCodeField(max_length=5)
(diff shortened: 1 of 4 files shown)


cfgov/legacy/forms.py
```diff
@@ -3,4 +3,4 @@
 
 
 class HousingCounselorForm(forms.Form):
-    zip = USZipCodeField()
+    zip = USZipCodeField(max_length=5)
```
kave: i think we'll need a new migration script for this

willbarton: That's not a model, just a form.

kave: lol whoops misread

PR #2221, opened 2016-07-06
WIP - Upgrade wagtail 

richaagarwal:
WIP, but feel free to refer to this if you'd like to test things out on the latest wagtail.


cfgov/v1/models/learn_page.py
```diff
@@ -40,7 +40,7 @@ class AbstractFilterPage(CFGOVPage):
     # General content tab panels
     content_panels = CFGOVPage.content_panels + [
         StreamFieldPanel('header'),
-        StreamFieldPanel('content'),
+        # StreamFieldPanel('content'),
```
richaagarwal: Temporary - need to figure out why we get this error when it's not commented out: `django.core.exceptions.FieldDoesNotExist: AbstractFilterPage has no field named 'content' `

kave: This looks like a copy/paste error that was introduced a long time ago, since the `content` StreamField correctly does not exist here. There must have been an update in their field checking that its finally catching this bug.

e.g. of where content does exist `https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/v1/models/landing_page.py#L18`

cfgov/v1/migrations/0032_cfgovimage_cfgovrendition.py
```diff
@@ -14,16 +14,16 @@ def migrate_images_data(apps, schema_editor):
     UserModel = apps.get_model(settings.AUTH_USER_MODEL)
```
kave: after this update, we really should squash these migrations. All this stuff is 100% unnecessary

PR #2226, opened 2016-07-13
Update spacing in item introduction

virginiacc:
Update spacing between header elements in item introduction and use item introduction template for blog/newsroom header

## Additions
- add 'and' between author names in item introduction byline
- add an else condition in category-slug class output

## Changes
- update spacing between elements in item introduction 
- use item introduction for article header

## Testing

- Navigate to a [blog post](http://localhost:8000/about-us/blog/older-americans-are-not-alone-in-the-fight-to-stop-financial-abuse/) and a [page with item introduction](http://localhost:8000/data-research/research-reports/2014-consumer-response-annual-report/). Check that:
  - Header element spacing matches specs in bug report (issue 10)
  - Multiple author names are joined by 'and'

## Review

- @contolini 
- @sebworks 
- @Scotchester 

## Screenshots
screen shot 2016-07-13 at 7 13 03 am

screen shot 2016-07-13 at 7 13 32 am

## Notes

- Lead paragraph is output in item introduction even when it has no content, causing slight spacing discrepancies
- Article template still needs to be converted to use item introduction template if possible
- If article template can't use item introduction template, consider moving byline generation from both templates into its own macro

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/item-introduction.html
+++ b/cfgov/jinja2/v1/_includes/organisms/item-introduction.html
@@ -32,38 +32,35 @@
 {% set page_url = get_protected_url(ancestor) %}
 
 <div class="o-item-introduction">
-    <div class="h4">
-        {{ category_slug.render(value.category, page_url, index=form_id) }}
-    </div>
+    {{ category_slug.render(value.category, page_url, index=form_id) }}
     <h1>{{ value.heading }}</h1>
 
     {% if value.paragraph %}
         <div class="lead-paragraph">{{ parse_links(value.paragraph) | safe }}</div>
     {% endif %}
+
     {% if value.date or page.authors.all() %}
-        <p>
+        <div class="meta">
     {% endif %}
-    {% for author in page.authors.names() %}
-        {% if loop.index == 1  %}
-            By <a href="{{ page_url }}?filter{{ form_id }}_authors={{ author }}">
-                {{ author }}
+    {% set num_authors = page.authors.all()|length %}
+        {% if num_authors %}
+            <span class="byline">
+            {%- for author in page.authors.names() -%}
+                {{- 'By ' if loop.index == 1 else ' and ' }}
+                <a href="{{ page_url }}?filter{{ form_id }}_authors={{ author }}">
+                    {{ author }}
                 </a>
-        {% elif loop.last == true %}
-            and <a href="{{ page_url }}?filter{{ form_id }}_authors={{ author }}">
-                {{ author }}
-                </a>
-        {% else %}
-            , <a href="{{ page_url }}?filter{{ form_id }}_authors={{ author }}">
-                {{ author }}
-                </a>
+            {%- endfor -%}
+                &ndash;
+            </span>
         {% endif %}
-    {% endfor %}
-    {% if value.date %}
-        {{ time.render(value.date, {'date':true}) }}
-    {% endif %}
+        <span class="date">
+            {{ time.render(value.date, {'date':true}) }}
+        </span>
     {% if value.date or page.authors.all() %}
-        </p>
+        </div>
     {% endif %}
+
     {% if value.has_social %}
         {{ social_media.render() }}
     {% endif %}
--- a/cfgov/jinja2/v1/_includes/organisms/item-introduction.html
+++ b/cfgov/jinja2/v1/_includes/organisms/item-introduction.html
@@ -32,8 +32,15 @@
 {% set page_url = get_protected_url(ancestor) %}
 
 <div class="o-item-introduction">
-    {{ category_slug.render(value.category, page_url, index=form_id) }}
-    <h1>{{ value.heading }}</h1>
+    {% if value.category is string %}
+        {{ category_slug.render(value.category, page_url, index=form_id) }}
+    {% else %}
+        {% for category in page.categories.all() %}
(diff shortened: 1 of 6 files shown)


cfgov/jinja2/v1/_includes/organisms/item-introduction.html
```diff
@@ -32,41 +32,45 @@
...
+            {{ category_slug.render(category.name, get_protected_url(ancestor), 'post_slug', index=index) }}
+        {% endfor %}
+    {% endif %}
+    <h1>{{ value.heading | safe }}</h1>
 
     {% if value.paragraph %}
         <div class="lead-paragraph">{{ parse_links(value.paragraph) | safe }}</div>
     {% endif %}
+
     {% if value.date or page.authors.all() %}
```
sebworks: should we just create a variable like `has_authors = page.authors.all()|length > 0`. It could be used below as well.

cfgov/jinja2/v1/_includes/organisms/item-introduction.html
```diff
@@ -32,41 +32,45 @@
...
-                {{ author }}
+    {% set num_authors = page.authors.all()|length %}
+        {% if num_authors %}
+            <span class="byline">
+            {%- for author in page.authors.names() -%}
+                {{- 'By ' if loop.index == 1 else ' and ' }}
+                <a href="{{ page_url }}?filter{{ form_id }}_authors={{ author }}">
+                    {{ author }}
                 </a>
+            {%- endfor -%}
```
sebworks: if there are no author names are we just displaying `&ndash;`?

cfgov/unprocessed/css/organisms/item-introduction.less
```diff
@@ -35,11 +35,28 @@
 */
 
 .o-item-introduction {
+    // visually approximates 60px
```
sebworks: I'm not sure we are still doing eyeball adjustments; maybe @jimmynotjim / @Scotchester  can enlighten me.

cfgov/unprocessed/css/organisms/item-introduction.less
```diff
@@ -35,11 +35,28 @@
...
+    margin-bottom: unit( 55px / @base-font-size-px, em);
+
     .short-desc {
-        padding-bottom: unit( @grid_gutter-width / 2 / @base-font-size-px, em)
+        padding-bottom: unit( (@grid_gutter-width / 2) / @base-font-size-px, em);
```
sebworks: I don't think parens are necessary but I could be wrong. I think the order of operations takes care of the precedence.

cfgov/unprocessed/css/organisms/item-introduction.less
Scotchester: That's correct, @sebworks, we've stopped doing visual approximations.

cfgov/jinja2/v1/_includes/organisms/item-introduction.html
virginiacc: Good point -- I'll make that change.

cfgov/jinja2/v1/_includes/organisms/item-introduction.html
virginiacc: The `&ndash;` should only show if `num_authors` evaluates to True.

cfgov/unprocessed/css/organisms/item-introduction.less
virginiacc: Good to know, thanks! So we'd use the full 60px value here?

Scotchester: Correct!

cfgov/jinja2/v1/_includes/organisms/item-introduction.html
sebworks: Is there a scenario in which we have authors and no author names? Is so, then the markup would just be ` <span class="byline">&ndash;</span> `. I'm not sure that it' s a valid scenario.

virginiacc: Oh, I see...hmm. The authors are a list of names added as tags to the page, so I wouldn't think that would happen, unless there's the possibility of an empty tag?

virginiacc:
Recommended changes implemented & ready for review.

sebworks:
👍

cfgov/unprocessed/css/post.less
```diff
@@ -373,31 +373,6 @@
     - cfgov-post
 */
 
-.post_heading {
```
sebworks: Are we using these classes in any other places? https://github.com/cfpb/cfgov-refresh/search?utf8=%E2%9C%93&q=post_meta

virginiacc: Oh, good catch -- they're also referenced in a browser test. I'm looking into whether these classes are still necessary elsewhere & if we can substitute the item introduction template in any other places they're used...

sebworks: Yeah, the sections of the browser tests referencing these classes can be deleted. We are trying to move away from testing dom element references and towards focusing on functionality.

virginiacc:
I temporarily restored the removed `.post` less styles. I think the event detail page can also be transitioned to use the item introduction template, but I'll put that in a separate PR so it won't hold up this one.

sebworks:
👍

PR #2227, opened 2016-07-13
Speedup Python unit tests by removing DB dependence

chosak:
Speedup Python unit tests by removing their dependence on the database. This speeds up the current Python unit tests from roughly two minutes to roughly two seconds:

```sh
$ time tox -e fast
...
Ran 129 tests in 0.506s
...
real    0m1.783s
user    0m1.381s
sys     0m0.313s
```

## Additions

- This PR introduces an alternate tox environment named `fast` that runs without recreating the virtualenv each time, invoked using `tox -e fast`. It also invokes tests using `python -m unittest discover` instead of `./manage. py test`, removing any dependence on the Django database. The standard tox environment (that recreates the DB and uses it each time) is still run if you just invoke `tox`.

## Changes

- There were several dozen tests that had some dependence on the database. Many of these were  using the built-in [test client](https://docs.djangoproject.com/en/1.9/topics/testing/tools/#the-test-client) which seems to check the `Site` table. Instead I use the [RequestFactory](https://docs.djangoproject.com/en/1.9/topics/testing/advanced/#django.test.RequestFactory) which lets you create requests and test views against them without going through the urlconf.
- Some of the existing unit tests had some painful mocking going on; I've tried to make things less painful but there are still some refactors that it would be good to have others review. See additional PR comments for some detail.
- This modifies the current `gulp test` command to run all backend unit tests; before it was only running the ones in `core.test` for some reason.

## Testing

- From the command line you can now do:

```sh
$ tox -e fast
```

Alternatively to run a subset of tests you can do something like:

```sh
~/Projects/cfgov-refresh/cfgov $ source .tox/fast/bin/activate
(fast) ~/Projects/cfgov-refresh/cfgov $ DJANGO_SETTINGS_MODULE=cfgov.settings.test python -m unittest core.tests.test_views
```

## Review

- @richaagarwal 
- @grapesmoker 
- @rosskarchner 
- @willbarton 

## Checklist

### DIFF
--- a/cfgov/cfgov/urls.py
+++ b/cfgov/cfgov/urls.py
@@ -37,16 +37,16 @@ urlpatterns = [
 
     url(r'^home/(?P<path>.*)$', RedirectView.as_view(url='/%(path)s', permanent=True)),
 
-    url(r'^owning-a-home/static/(?P<path>.*)$', RedirectView.as_view(url='/static/owning-a-home/static/%(path)s')),
-    url(r'^owning-a-home/resources/(?P<path>.*)$', RedirectView.as_view(url='/static/owning-a-home/resources/%(path)s')),
+    url(r'^owning-a-home/static/(?P<path>.*)$', RedirectView.as_view(url='/static/owning-a-home/static/%(path)s', permanent=True)),
+    url(r'^owning-a-home/resources/(?P<path>.*)$', RedirectView.as_view(url='/static/owning-a-home/resources/%(path)s', permanent=True)),
     url(r'^owning-a-home/', include(SheerSite('owning-a-home').urls)),
 
     # the two redirects are an unfortunate workaround, could be resolved by
     # using static('path/to/asset') in the source template
 
-    url(r'^tax-time-saving/static/(?P<path>.*)$', RedirectView.as_view(url='/static/tax-time-saving/static/%(path)s')),
+    url(r'^tax-time-saving/static/(?P<path>.*)$', RedirectView.as_view(url='/static/tax-time-saving/static/%(path)s', permanent=True)),
     url(r'^tax-time-saving/', include(SheerSite('tax-time-saving').urls)),
-    url(r'^know-before-you-owe/static/(?P<path>.*)$', RedirectView.as_view(url='/static/know-before-you-owe/static/%(path)s')),
+    url(r'^know-before-you-owe/static/(?P<path>.*)$', RedirectView.as_view(url='/static/know-before-you-owe/static/%(path)s', permanent=True)),
     url(r'^know-before-you-owe/', include(SheerSite('know-before-you-owe').urls)),
 
     url(r'^adult-financial-education/', include(fin_ed.urls_for_prefix('adult-financial-education'))),
@@ -186,10 +186,10 @@ urlpatterns = [
             template_name='regulation-comment/reg-comment-form-test.html'),
         name='reg-comment-form-test'),
 
-    url(r'^feed/$', RedirectView.as_view(url='/about-us/blog/feed/')),
-    url(r'^feed/blog/$', RedirectView.as_view(url='/about-us/blog/feed/')),
-    url(r'^feed/newsroom/$', RedirectView.as_view(url='/about-us/newsroom/feed/')),
-    url(r'^newsroom-feed/$', RedirectView.as_view(url='/about-us/newsroom/feed/')),
+    url(r'^feed/$', RedirectView.as_view(url='/about-us/blog/feed/', permanent=True)),
+    url(r'^feed/blog/$', RedirectView.as_view(url='/about-us/blog/feed/', permanent=True)),
+    url(r'^feed/newsroom/$', RedirectView.as_view(url='/about-us/newsroom/feed/', permanent=True)),
+    url(r'^newsroom-feed/$', RedirectView.as_view(url='/about-us/newsroom/feed/', permanent=True)),
 
     url(r'^about-us/$', SheerTemplateView.as_view(template_name='about-us/index.html'), name='about-us'),
 
--- a/cfgov/core/tests/test_views.py
+++ b/cfgov/core/tests/test_views.py
@@ -1,7 +1,8 @@
+import django
 import json
 import sys
 
-from mock import call, patch
+from mock import Mock, call, patch
 
 if sys.version_info[0] < 3:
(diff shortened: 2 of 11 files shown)


cfgov/cfgov/urls.py
```diff
@@ -37,16 +37,16 @@
 
     url(r'^home/(?P<path>.*)$', RedirectView.as_view(url='/%(path)s', permanent=True)),
 
-    url(r'^owning-a-home/static/(?P<path>.*)$', RedirectView.as_view(url='/static/owning-a-home/static/%(path)s')),
-    url(r'^owning-a-home/resources/(?P<path>.*)$', RedirectView.as_view(url='/static/owning-a-home/resources/%(path)s')),
+    url(r'^owning-a-home/static/(?P<path>.*)$', RedirectView.as_view(url='/static/owning-a-home/static/%(path)s', permanent=True)),
```
chosak: In Django 1.9 the default value of permanent for [RedirectView](https://docs.djangoproject.com/en/1.9/ref/class-based-views/base/#redirectview) changes from false to true. If you do not specify a value for this you get RemovedInDjango19Warnings when testing against this file. This change just makes explicit what is (in the current version of Django 1.8) currently implicit.

cfgov/core/tests/test_views.py
```diff
@@ -1,221 +1,257 @@
...
+            'email': 'fake@example.com',
+        }
+
+        answers = [
+            ('batman', 'robin'),
+            ('hello', 'goodbye'),
+        ]
+
+        if include_answers:
+            post.update({'questionid_' + q: a for q, a in answers})
```
chosak: This code remains quite hacky although hopefully an improvement on what was there before. It'd be better to refactor the underlying govdelivery code to allow for nicer testing of individual methods. For this PR I tried to focus solely on maintaining existing tests while decoupling from the DB.

cfgov/sheerlike/loaders.py
```diff
@@ -1,5 +1,4 @@
 from jinja2.loaders import FileSystemLoader
-import fslib
```
chosak: This import isn't present in any requirements files and isn't used anywhere, so I've removed it.

cfgov/v1/models/browse_filterable_page.py
```diff
@@ -12,7 +12,6 @@
 
 from .base import CFGOVPage
 from .learn_page import AbstractFilterPage
-from .. import forms
```
chosak: This had to be removed and moved down into various `get_form_class` methods because otherwise there would be an infinite import loop involving `v1.models`, `v1.forms`, and `v1.util`. I've added a few additional unit tests to validate this change (see below).

gulp/tasks/test.js
```diff
@@ -42,7 +42,6 @@ function testUnitScripts( cb ) {
 function testUnitServer() {
   spawn(
     'tox',
-    [ 'cfgov/core/tests' ],
```
chosak: Modifies `gulp test` behavior to run all Python unit tests, not just the ones under `core.test`.

cfgov/core/tests/test_views.py
willbarton: :+1: That seems like the right approach to me. 

Want to open up a "refactor gov delivery" issue to go on the backlog?

chosak: Thanks, I've created a new task for this.

tox.ini
```diff
@@ -17,3 +17,7 @@ changedir = {toxinidir}/cfgov
 commands=
     python manage.py test {posargs}
     
+[testenv:fast]
```
willbarton: Is there a reason to keep this env separate from the one above?

chosak: There are two different and valuable use cases. First, start a brand new venv, create a database with all migrations, and then run the set of unit tests. This takes a long time but makes sure that everything works from scratch. Second, is reuse the existing venv (avoids having to pip install each time) and avoid using the DB at all (for now, although we should discuss how to handle this going forward). This is fast and makes it easy for developers to run tests quickly.

Travis and Jenkins should do the former, and I think it's useful to have the latter for day-to-day work. My instinct is that it's best to leave the default be the former when running `tox`, and explicitly use the faster one with `tox -e fast`, but we could change it so that the default is fast and you have to do something like `tox -e rebuild` to do everything.

willbarton: Sounds good to me! 👍

cfgov/v1/tests/test_password_policy.py
```diff
@@ -40,47 +41,55 @@ def test_good_passwords(self):
...
-        user = self.get_user()
-        with self.assertRaises(ValidationError):
-            password_policy.validate_password_history(user, 'password')
-
-    def test_new_password(self):
-        user = self.get_user()
-        password_policy.validate_password_history(user, 'new_password')
+    def test_can_set_new_password(self):
+        user = self.get_user('password1')
+        password_policy.validate_password_history(user, 'password2')
```
richaagarwal: can you pass in the parameters explicitly here, so `self.get_user(last_password='password1')` ?

richaagarwal:
👍 other than my comment.  I can't really speak to what's in `test_views.py` so maybe @rosskarchner can do a quick review as well?

cfgov/v1/tests/test_password_policy.py
chosak: Updated, thanks.

cfgov/v1/tests/test_password_policy.py
```diff
@@ -81,12 +81,12 @@ def test_cant_set_password_if_locked(self):
...
     def test_can_set_new_password(self):
-        user = self.get_user('password1')
+        user = self.get_user(last_password='password1')
         password_policy.validate_password_history(user, 'password2')
 
     def test_cant_reuse_password(self):
         password = 'password'
-        user = self.get_user(password)
+        user = self.get_user(last_password=password)
         self.assertRaises(
```
richaagarwal: This is nitpicky, but could you update this so it's consistent with the test above - e.g. passing in the string in both cases or passing in a variable in both cases.

richaagarwal: Actually, nevermind - I think it's slightly clearer to use the string but I see why you used a variable in `test_cant_resuse_password` and not the other, since it's used twice.

richaagarwal:
👍

PR #2228, opened 2016-07-14
Fix pagination bug

kurtw:
There is a bug in the [Leadership Calendar](http://www.consumerfinance.gov/about-us/the-bureau/leadership-calendar) where a day's events that split across multiple pages output out of order and/or missing. Also, events that have a start date within the hour threshold of being saved in UTC as the next day are showing up in UTC day's table instead of it's actual day's table. Additionally, there was a `pdb` statement left in there from the last PR that needed to be removed because [invalid form submission results in server errors](http://www.consumerfinance.gov/about-us/the-bureau/leadership-calendar/?filter_range_date_gte=01).

## Changes
The reason for the pagination bug is somewhat simple but took some maneuvering to overcome. We served events to the page in most recent to least then grouped events by the day and reversed the order within that day. This caused the strange behavior seen when days were split between pages.

- In order to subvert this, I changed the the display of each page to show events in **iterations of 5 days**. 
- I also refactored the view function and fixed the template for the new logic to display events.

## Testing

- `gulp test`
- go to leadership calendar page
- click on pagination button to see each page have 5 days of events that _do not_ bleed events for a single day across pages
- filter the events for a date range
 - filter across 6 days to see 2 pages with one that has 1 day. There's some logic that handles that specifically that should be tested.
- filter by calendar
- test that the PDF works

## Review

- @rosskarchner 
- @richaagarwal 
- @willbarton 
- @schaferjh re: **"5 days per page"** (I thought this flowed better since some days were very short)
- anyone else i should've tagged

#### Page 1
screen shot 2016-07-14 at 1 21 08 pm

#### Page 2
screen shot 2016-07-14 at 1 20 57 pm

#### Page 1
screen shot 2016-07-14 at 1 35 04 pm

#### Page 2
screen shot 2016-07-14 at 1 35 12 pm

## Notes

- There are a few more things that I noticed that the Platform team should be aware of:
 - Invalid date input gets passed into the backend and invalidates the form. I've programmed it to return nothing but **there's no error message shown, only a warn notification stating there's no results.** Additionally, on the JS date validator, the date formats that are invalid but not caught send that request to the server and then get fixed in the form after page load which is confusing because the user submits something invalid and then sees a valid date with no results. This is better than before, where it would just throw a 500 error, but definitely isn't desired. I felt that this was out of scope for the pagination fix.
- To the same point, there are certain date formats that are caught by the JS validator and some that are not. The validator should probably be smarter.
- There are two places where leadership calendar PDF handling is located. [Here](https://github.com/cfpb/cfgov-refresh/blob/f3f5dd156dd121c1db8ae62b484d83ebee7c9b37/cfgov/cal/views.py#L125) and [here](https://github.com/cfpb/cfgov-refresh/blob/flapjack/cfgov/core/services.py#L30). This should be refactored. Again, out of scope for this PR.

## Todos

- Add tests
- Add a CHANGELOG entry

## Checklist

### DIFF
--- a/cfgov/cal/views.py
+++ b/cfgov/cal/views.py
@@ -17,7 +17,7 @@ from django.template import RequestContext
 from django.template.loader import get_template
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
 from django.core.exceptions import ObjectDoesNotExist
-from django.db.models import Max, Min
+from django.db.models import Max, Min, Q
 
 from cal.models import CFPBCalendar, CFPBCalendarEvent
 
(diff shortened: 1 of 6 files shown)


schaferjh:
re: 5 days -- this concerns me a little because it introduces the possibility for a seriously long page BUT I agree with you that when short days occur it would be better to have at least five items showing. I'm going to say 👍  for now. thanks!!

kurtw:
@schaferjh Okay, and we can change this easily if it ever becomes a problem!

kurtw:
re: TODO - create tests

My testing environment is acting up so I can't confidently create any new tests at the moment. Will push up some tests when possible

richaagarwal:
@kurtw Could you clarify this "Also, events that have a start date within the hour threshold of being saved in UTC as the next day are showing up in UTC day's table instead of it's actual day's table."  Not sure what it means / having trouble parsing it

rosskarchner:
@richaagarwal that means, that an event at say, 11PM EST on a monday would be grouped in tuesday's events, because the timezone wasn't taken into account when grouping events.

richaagarwal:
Okay, so just so I'm clear -
1) We currently group events according to UTC
2) We want to group them according to ET
3) This is fixed in this PR

Is this correct?
@rosskarchner @kurtw

rosskarchner:
@richaagarwal yeah--

Here is what we have now:

screen shot 2016-07-18 at 1 34 28 pm

That first event under 10/20 should really be the last event under 10/19.

With this PR, the event is in the right place:

screen shot 2016-07-18 at 1 36 04 pm

kurtw:
Thanks for explaining @rosskarchner

rosskarchner:
👍

willbarton:
:+1:

PR #2225, opened 2016-07-13
Wagtail Upgrade to 1.5.2 

richaagarwal:
We're upgrading to the latest version of Wagtail for a few reasons:
- Fix issues a couple of us see with old Wagtail migrations when setting up environment
- Allow us to make use of new features in the latest releases of Wagtail:
http://docs.wagtail.io/en/v1.4.1/releases/1.4.html,
http://docs.wagtail.io/en/v1.5/releases/1.5.html, 
http://docs.wagtail.io/en/v1.5.2/releases/1.5.html
- Because upgrading is good

This PR upgrades Wagtail to 1.5.2 (currently the latest stable version) but the changes you see are mostly related to migrations - squashing old ones and adding a couple new ones. 

## Notes

- When this eventually gets deployed to prod, we will need to run the specific migration steps mentioned in the first testing step.
- To ease seeing what's changed, don't go to the "Files changed" tab, go to specific commits (this is the main one: https://github.com/cfpb/cfgov-refresh/pull/2225/commits/2e19c88b32343cfd98d5ddd821e64933b0ee9a82) since the squash migrations commit is huge. 

## todo
- I will update the CHANGELOG after this is reviewed

## Additions

- Two new migrations needed for the Wagtail upgrade from 1.3 --> 1.5.2

## Removals

- Many old migrations 
- The content section from AbstractFilterPage, which apparently was never supposed to be there but we get an error now because Wagtail is stricter.

## Changes

- Most old migrations are now squashed into `0001_initial.py` 

## Testing

Two options. 
### 1) Run against a prod database (longer)
- Pull down this branch `wagtail-spike-2`, run `pip install -r requirements/base.txt`
- Grab a prod db dump (which can't be posted here, but there's a link in our chat room) 
- Run `./refresh-data.sh prod_dump.sql`
- Run `python cfgov/manage.py migrate --fake`
- Run `python cfgov/manage.py migrate --fake wagtailcore 0023_alter_page_revision_on_delete_behaviour`
- Run `python cfgov/manage.py migrate --fake v1 0002_share_perms`
- Run `python cfgov/manage.py migrate`

For each of these, you will get this dialogue:
```
Any objects related to these content types by a foreign key will also
be deleted. Are you sure you want to delete these content types?
If you're unsure, answer 'no'.
```
Say no for the first 3, yes for the last. 

The app should now run properly and you should be able to run `gulp test` and `tox` successfully without any test failures.  I don't have testing instructions beyond that, but feel free to play around in the app.

[...]

PR #2234, opened 2016-07-21
Separate browser tests into functional and content suites

chosak:
The current set of browser acceptance tests (run using `gulp test:acceptance`) include some that rely on a complete content database and some that can be run using a minimal set of generic content. This PR separates the tests into two different suites so that the functional subset can be easily run without needing a complete database dump.

## Additions

- Added suites to [test/browser_tests/environment.js](https://github.com/cfpb/cfgov-refresh/blob/cleanup-browser-tests/test/browser_tests/environment.js#L27) to group browser tests into logical groups.
- Added support for the `--suite` parameter when running `gulp test:acceptance`, for example:

 ```sh
 $ gulp test:acceptance --suite=functional
 $ gulp test:acceptance --suite=functional,content
 ```

 Support for `--specs` is still supported; however when passing `--suite`, the suite argument will take precedence. If neither of these arguments are provided, all tests will still run.

## Changes

- Updated `TEST.md` to describe new command-line usage.
- Modified `gulp test:acceptance` so that the `initial-test-data.sh` script is always run (removing a hardcoded check against the `wagtail` spec directory) to load data needed to pass the functional tests. This is suboptimal; it'd be more ideal to use a separate test database (maybe even the same Django DB as the unit tests) and should be revisited in future.

## Testing

- Run the functional browser test suite like this:

```sh
$ gulp test:acceptance --suite=functional
...
48 specs, 0 failures
Finished in 23.152 seconds
```

- The content-dependent browser test suite (which requires a complete database) can be run like this:

```sh
$ gulp test:acceptance --suite=content
...
255 specs, 0 failures, 27 pending specs
Finished in 100.738 seconds
```

- The full set of tests (all suites) can still be run like this:

```sh
$ gulp test:acceptance
...
303 specs, 0 failures, 27 pending specs
Finished in 121.615 seconds
```

- Individual or specific specs can still be run like this:

```sh
$ gulp test:acceptance --specs=foo*.js
...
5 specs, 0 failures
Finished in 2 seconds
```

- Running `gulp test` will now run the linter, the Python unit tests, and the functional test suite.

## Review

- @richa
- @sebworks

## Checklist


chosak:
Trying to close and reopen this PR to see if it will [trigger a seemingly missed Travis build](http://stackoverflow.com/questions/19589260/force-travis-to-build-a-specific-pull-request-or-commit-sha-that-it-missed).

test/browser_tests/conf.js
```diff
@@ -69,12 +69,31 @@ function _isSauceCredentialSet() {
...
+    var suitesArray = params.suite.split( ',' );
+    var specsArray = [];
+    for ( var i = 0, suite; suite = suitesArray[i++]; ) {
+      var suiteSpecs = environment.suites[suite];
+      if ( suiteSpecs.constructor === Array ) {
+        for ( var j = 0, len = suiteSpecs.length; j < len; j++ ) {
+          specsArray.push( suiteSpecs[j] );
+        }
+      } else {
+        specsArray.push(suiteSpecs);
```
anselmbradford: Trivial: Add space in parens.

chosak: Thanks, fixed.

richaagarwal:
This is great, thanks Andy! Does `gulp test` still run everything?  Do we want to modify that?  Ideally we'd still have a single command people can reliably call.

chosak:
@richaagarwal yes, `gulp test` still runs `lint`, `test:unit`, and `test:acceptance`. So right now the default is "run everything, including things that need the prod DB". I'd like to change Travis to add `gulp test:acceptance --suite:functional` so that it runs the non-content tests.

test/browser_tests/conf.js
```diff
@@ -69,12 +69,31 @@ function _isSauceCredentialSet() {
...
   var parsedParams = {};
 
-  if ( _paramIsSet( params.specs ) ) {
+  // If one or more suites are specified, use their specs.
+  if ( _paramIsSet( params.suite ) ) {
+    var suitesArray = params.suite.split( ',' );
+    var specsArray = [];
+    for ( var i = 0, suite; suite = suitesArray[i++]; ) {
+      var suiteSpecs = environment.suites[suite];
+      if ( suiteSpecs.constructor === Array ) {
```
sebworks: Probably safer to use `Array.isArray`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

gulp/tasks/test.js
```diff
@@ -95,6 +95,9 @@ function _getProtractorParams() {
   // If --version=number flag is added on the command-line.
   params = _addCommandLineFlag( params, commandLineParams, 'version' );
 
+  // If --suite=suite1,suite2 flag is added on the command-line.
+  params = _addCommandLineFlag( params, commandLineParams, 'suite' );
+
   return params;
```
richaagarwal: Not a huge deal, but can we do something like `_addCommandLineFlags` (plural) here, that passes in 'version', 'suite', & all the other flags?

test/browser_tests/conf.js
```diff
@@ -69,12 +69,31 @@ function _isSauceCredentialSet() {
...
+        for ( var j = 0, len = suiteSpecs.length; j < len; j++ ) {
+          specsArray.push( suiteSpecs[j] );
```
sebworks: You could probably just use a hack to clone the array like `[1,2,3].slice(0)`. http://stackoverflow.com/questions/3978492/javascript-fastest-way-to-duplicate-an-array-slice-vs-for-loop

chosak: Thanks @sebworks, `concat` is pretty useful, and much nicer than how I had been doing it.

gulp/tasks/test.js
chosak: I agree that this code isn't the clearest; I left it as is just to keep the set of changes in this PR small.

chosak:
Per @richaagarwal feedback, I changed this PR so that if you simply run `gulp test` from the command line, it'll now run only the functional browser tests (along with the Python unit tests and the linter). You can still run the full set of browser tests using `gulp test:acceptance` or with some of the command-line options described above.

richaagarwal:
👍

gulp/tasks/test.js
richaagarwal: Makes sense!

sebworks:
Getting the following errors when running `gulp test:acceptance --suite=functional`:

screen shot 2016-07-22 at 10 47 59 am

chosak:
@sebworks this might be an issue I encountered recently regarding a missing `HTTP_PORT` environment variable. Can you please try the solution described in the PR comments starting [here](https://github.com/cfpb/cfgov-refresh/pull/2225#issuecomment-232711269)?

sebworks:
yep, that was the issue. Great work @chosak. 👍

PR #2237, opened 2016-07-21
Adds ability to auto-fix linter errors

anselmbradford:
## Additions

- Adds ability to auto-fix linter errors with a command-line `--fix` flag on the linter task.

## Testing

- Run `gulp lint --fix` and look at your changed files in git. See that `Header.js` and `es5-shim.js` have spacing issues that are automatically fixed.

## Review

- @sebworks 
- @contolini 
- @chosak 
- @jimmynotjim 
- @virginiacc
- @KimberlyMunoz

### DIFF
--- a/TEST.md
+++ b/TEST.md
@@ -162,3 +162,19 @@ To audit a page's WCAG and Section 508 accessibility:
   4. To test a page aside from the homepage, add the `--u=<path_to_test>` flag.
      For example, `gulp test:a11y --u=contact-us`
      or `gulp test:a11y --u=the-bureau/bureau-structure/`.
+
+# Source code linting
+
+The default test task includes linting of the JavaScript source, build,
+and test files.
+Use the `gulp lint` command from the command-line to run the ESLint linter,
+which checks the JavaScript against the rules configured in `.eslintrc`.
+[See the ESLint docs](http://eslint.org/docs/rules/)
+for detailed rule descriptions.
+
+There are a number of options to the command:
+ - Use `gulp lint:build` to only lint the build scripts.
+ - Use `gulp lint:test` to only lint the test scripts.
+ - Use `gulp lint:scripts` to only lint the project source scripts.
+ - Add the `--fix` flag (like `gulp lint --fix`) to auto-fix
+   some errors, where ESLint has support to do so.
--- a/gulp/tasks/lint.js
+++ b/gulp/tasks/lint.js
@@ -4,6 +4,7 @@ var gulp = require( 'gulp' );
 var plugins = require( 'gulp-load-plugins' )();
 var configLint = require( '../config' ).lint;
 var handleErrors = require( '../utils/handle-errors' );
+var minimist = require( 'minimist' );
 
 /**
  * Generic lint a script source.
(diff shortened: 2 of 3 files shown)


sebworks:
This is awesome. 👍

chosak:
Agree, this is very helpful. 👍

jimmynotjim:
Whaaaaa O_O This is amazing!

TEST.md
```diff
@@ -162,3 +162,16 @@ To audit a page's WCAG and Section 508 accessibility:
   4. To test a page aside from the homepage, add the `--u=<path_to_test>` flag.
      For example, `gulp test:a11y --u=contact-us`
      or `gulp test:a11y --u=the-bureau/bureau-structure/`.
+
+# Source code linting
+
+The default test task includes linting of the JavaScript source, build,
+and test files.
+Use the `gulp lint` command from the command-line to run the ESLint linter,
+which defines a number of [rules](http://eslint.org/docs/rules/).
```
Scotchester: Suggest changing this line to:

```markdown
which checks the JavaScript against the rules configured in `.eslintrc`.
[See the ESLint docs](http://eslint.org/docs/rules/) for detailed rule descriptions.
```

TEST.md
```diff
@@ -162,3 +162,16 @@ To audit a page's WCAG and Section 508 accessibility:
...
+There are a number of options to the command:
```
Scotchester: Start a new paragraph here.

gulp/tasks/lint.js
```diff
@@ -4,16 +4,21 @@ var gulp = require( 'gulp' );
...
 function _genericLint( src ) {
-  return gulp.src( src )
-    .pipe( plugins.eslint() )
+  // Grab the --fix flag from the command-line, if available.
+  var commandLineParams = minimist( process.argv.slice( 2 ) );
+  var willFix = commandLineParams.fix || false;
+  return gulp.src( src, { base: './' } )
+    .pipe( plugins.eslint( { fix: willFix } ) )
     .pipe( plugins.eslint.format() )
+    .pipe( gulp.dest( './' ) )
```
Scotchester: Why is this line added?

gulp/tasks/lint.js
```diff
@@ -4,16 +4,21 @@ var gulp = require( 'gulp' );
...
  * @param {string} src The path to the source JavaScript.
  * @returns {Object} An output stream from gulp.
  */
 function _genericLint( src ) {
-  return gulp.src( src )
-    .pipe( plugins.eslint() )
+  // Grab the --fix flag from the command-line, if available.
+  var commandLineParams = minimist( process.argv.slice( 2 ) );
+  var willFix = commandLineParams.fix || false;
+  return gulp.src( src, { base: './' } )
```
Scotchester: Why is the second parameter added here?

TEST.md
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2237/commits/73499da952846a2e8f6e41fe09c96066384bceb2

TEST.md
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2237/commits/7a1d73933d8530fe6e68733647ae37dc04a1d539

gulp/tasks/lint.js
anselmbradford: So that if the files are auto-fixed they can be written to disk.

gulp/tasks/lint.js
anselmbradford: The `base` property is used to set the base path of the source file glob. By setting it to the current directory, and then also providing the same setting in `gulp.dest`, each source file path is used in the destination, effectively overwriting the files. I wasn't sure how else to pass in the destination file paths, which differ widely.

gulp/tasks/lint.js
Scotchester: Shouldn't they be written back to `gulp.src`?

anselmbradford: `gulp.src` just reads files as far as I know. `gulp.dest` is what writes them out.

jimmynotjim: @Scotchester `gulp.dest` in this instance isn't the same as the chained variable naming we've been using like `styles.dest`, it's a method for writing to disk.

Scotchester: Gotcha.

Scotchester:
:+1:

PR #2236, opened 2016-07-21
Fix link icons

kurtw:
The logic handling links doesn't insert the correct markup.

## Additions

- Added download link regex pattern and logic to parse_link function for rich text.

## Removals

- CSS to add icon classes for files

## Changes

- Markup is inserted and manipulated differently for affected links matching the patterns.

## Testing

- Add cfpb link
- Add .gov link
- Add external link
- add file link
- make sure markup is correct for each one

## Review

- @Scotchester 
- @willbarton 
- @richaagarwal 
- @virginiacc 
- @contolini 

## Notes

-

## Todos

- Change where logic is located for efficiency
- Possible refactor for readability
- Write tests

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #2244, opened 2016-07-26
Consolidates environment variable references

anselmbradford:
## Changes
- Consolidated all environment variables in `config/environment.js`.
- Ignored `console.log` in tests and enforced `no-process`.
- Updated `STAGING_HOSTNAME` to `DJANGO_STAGING_HOSTNAME` environment var.
- Updated `HTTP_HOST` to `TEST_HTTP_HOST` environment var.
- Updated `HTTP_PORT` to `TEST_HTTP_PORT` environment var.
- Allows passing of port to `runserver.sh`.
- Added misnamed and unreferenced environment variables to .env.

## Removed
- Removed unused `SELENIUM_URL` environment variable.

## Testing

- Grab `clear variable state` and `export variable defaults` sections from `.env_SAMPLE` and add to your local `.env`.
- Update values in your local `.env` for:
  - `STAGING_HOSTNAME` > `DJANGO_STAGING_HOSTNAME`
  - `HTTP_HOST` > `TEST_HTTP_HOST`
  - `HTTP_PORT` > `TEST_HTTP_PORT`
  - add `SITE_DESC`
  - add `DJANGO_HTTP_PORT`
  - remove `SELENIUM_URL`
- Reload your environment with `source ./.env` while in the root directory.
- Stop the server and re-start with `./runserver.sh`
- Edit `DJANGO_HTTP_PORT` in django settings section to change port (e.g. `8001`).
- Reload settings with `source ./.env`.
- Stop the server and start again with `./runserver.sh` and note port change from before.
- All other test tasks, etc. should run as before:
  - `gulp test:a11y`
  - `gulp test --sauce=false`
  - `gulp test` (on sauce).

## Review

- @chosak 
- @rosskarchner
- @richaagarwal
- @vickumar1981
- @vccabral

## Screenshots

screen shot 2016-07-26 at 4 24 14 pm

## Notes

- All settings that have defaults should probably be unset to begin with and then set to their defaults in `.env` and then the individual settings should be commented out by default so that they are only enabled in order to override the default.

### DIFF
--- a/.env_SAMPLE
+++ b/.env_SAMPLE
@@ -3,26 +3,13 @@
 # The file may be manually initialized with . ./.env from the project root.
 # --------------------------------------------------------------------------
 
-###################################################################
-# DO NOT EDIT VALUES. Project configuration - clear variable state.
-###################################################################
-
-unset DJANGO_STAGING_HOSTNAME
-unset DJANGO_HTTP_PORT
-unset TEST_HTTP_HOST
-unset TEST_HTTP_PORT
-unset SAUCE_USERNAME
-unset SAUCE_ACCESS_KEY
-unset SAUCE_SELENIUM_URL
-unset ACHECKER_ID
-
 ############################
 # Django Application Server.
 ############################
 
-#export DJANGO_STAGING_HOSTNAME=<staging_hostname>
+export DJANGO_STAGING_HOSTNAME='content.localhost'
 #export DJANGO_STATIC_ROOT=<path_to_static_files>
-#export DJANGO_HTTP_PORT=8000
+export DJANGO_HTTP_PORT=8000
 
 #############################################
 # V1 - application feature related variables.
@@ -122,9 +109,9 @@ export WORDPRESS=http://www.consumerfinance.gov
 # TESTING (optional) - for handling testing of the site.
 ########################################################
 
-#export SITE_DESC=cfgov
-#export TEST_HTTP_HOST=localhost
-#export TEST_HTTP_PORT=8000
+export SITE_DESC=cfgov
+export TEST_HTTP_HOST=localhost
+export TEST_HTTP_PORT=8000
 
 # Web service ID for accessibility testing via http://achecker.ca site.
 #export ACHECKER_ID=<web_service_id>
@@ -140,20 +127,6 @@ export TOXENV=py27
 #export SAUCE_ACCESS_KEY=<sauce_access_key>
 #export SAUCE_SELENIUM_URL=localhost:4445/wd/hub
 #export SAUCE_TUNNEL=<sauce_tunnel_id>
-
-#######################################################################
-# DO NOT EDIT VALUES. Project configuration - export variable defaults.
-#######################################################################
-
-export DJANGO_STAGING_HOSTNAME=${DJANGO_STAGING_HOSTNAME:-'content.localhost'}
-export DJANGO_HTTP_PORT=${DJANGO_HTTP_PORT:-8000}
-export SITE_DESC=${SITE_DESC:-cfgov}
-export TEST_HTTP_HOST=${TEST_HTTP_HOST:-localhost}
-export TEST_HTTP_PORT=${TEST_HTTP_PORT:-${DJANGO_HTTP_PORT}}
-export SAUCE_USERNAME=${SAUCE_USERNAME:-''}
-export SAUCE_ACCESS_KEY=${SAUCE_ACCESS_KEY:-''}
-export SAUCE_SELENIUM_URL=${SAUCE_SELENIUM_URL:-localhost:4445/wd/hub}
-export ACHECKER_ID=${ACHECKER_ID:-''}
 
 ###############################################
 # Project configuration - set up working state.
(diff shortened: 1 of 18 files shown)


.env_SAMPLE
```diff
@@ -125,6 +139,21 @@ export TOXENV=py27
 #export SAUCE_USERNAME=<sauce_username>
 #export SAUCE_ACCESS_KEY=<sauce_access_key>
 #export SAUCE_SELENIUM_URL=localhost:4445/wd/hub
+#export SAUCE_TUNNEL=<sauce_tunnel_id>
+
+#######################################################################
+# DO NOT EDIT VALUES. Project configuration - export variable defaults.
+#######################################################################
+
+export DJANGO_STAGING_HOSTNAME=${DJANGO_STAGING_HOSTNAME:-'content.localhost'}
```
chosak: Love the advanced bash scripting trick here, but isn't this way of settings defaults bash-specific? I know others (like @grapesmoker) prefer to use other shells.

.env_SAMPLE
```diff
@@ -108,9 +122,9 @@ export WORDPRESS=http://www.consumerfinance.gov
 # TESTING (optional) - for handling testing of the site.
 ########################################################
 
-#export HTTP_HOST=localhost
-#export HTTP_PORT=8000
-#export SELENIUM_URL=http://localhost:4444/wd/hub
+#export SITE_DESC=cfgov
+#export TEST_HTTP_HOST=localhost
+#export TEST_HTTP_PORT=8000
```
chosak: Should this whole `TESTING` section be removed since you're resetting these three variables below in the `Project configuration` section?

anselmbradford: This originally came from needing an easy way to swap port values to/from the default, so I set it up as three step:
 1. unset the value
 2. set override (commented out by default)
 3. set default.

So the unset step (1) and the default step (3) would always be enabled and should remain untouched for consistency and ease of copying from `.env_SAMPLE`. To set the port, for example, L:127 would be uncommented and changed to `export TEST_HTTP_PORT=8001` and `source ./.env` would be run. To reset it back, comment out L:127 and re-run `source ./.env`. 

If that's too complicated (or doesn't work outside of bash) I'm happy to updated to whatever works for you'll. It's more important I think that the `process.env` calls are consolidated in one spot.

.env_SAMPLE
anselmbradford: mm, I'm not sure. @grapesmoker let me know if you see a better way.

willbarton: I'm a zsh user, and I can confirm that I see the exact same results in zsh as I do in bash:

```
%> export DJANGO_STAGING_HOSTNAME=${DJANGO_STAGING_HOSTNAME:-'content.localhost'}
%> echo $DJANGO_STAGING_HOSTNAME 
content.localhost

%> export DJANGO_STAGING_HOSTNAME=test
%> export DJANGO_STAGING_HOSTNAME=${DJANGO_STAGING_HOSTNAME:-'content.localhost'}
%> echo $DJANGO_STAGING_HOSTNAME
test
```

chosak: The [fish](https://fishshell.com/) shell [has issues](http://unix.stackexchange.com/q/81421) with this syntax:

```sh
~> export DJANGO_STAGING_HOSTNAME=${DJANGO_STAGING_HOSTNAME:-'content.localhost'}
${ is not a valid variable in fish.
```

.env_SAMPLE
chosak: Thanks, that's a helpful explanation. It feels like this is getting somewhat complicated, when all we need is just a way to set a bunch of environment variables. Did you specifically want to change the behavior so that you don't have to do anything special to `.env` in order to run the tests? Right now the comment in `.env_SAMPLE` tells you to uncomment those test variables, but `TEST.md` doesn't mention them.

What if we left `.env` as it was (commented out test variables, with defaults), made `TEST.md` more explicit about the need to set those, and then added a check in `environment.js` that actually fails if any of those variables are unset? That way you could still easily modify `.env` to override defaults as needed.

anselmbradford: > Did you specifically want to change the behavior so that you don't have to do anything special to .env in order to run the tests?

Yeah, I wanted it so that certain defaults are set out of the box and everything should work. There's no reason to force editing of `.env` unless the user wants to change to a special setup (such as running the server on a different port).

I realized the other reason I put the defaults in `.env` was so they're set in one file. If you search the changed files for `8000` you'll see that that is set as a default value in several places across the codebase, which should really only happen in one spot.

Still mulling how to approach this. Throwing an error in `environment.js` would mean the error wouldn't be thrown when the environment vars are set in `.env`.

anselmbradford:
@chosak I decided to strip the defaults out of this PR so the other fixes can go in and we can loop back later to how the defaults are set up. Let me know how it looks!

chosak:
Thanks. I get three failures when running `gulp test`, all from `v1.tests.test_links.ImportDataTest`. I believe this is due to #2236 - I've left a more detailed comment [there](https://github.com/cfpb/cfgov-refresh/pull/2236#issuecomment-236069301).

anselmbradford:
Thanks @chosak! I'll merge shortly and hopefully those errors will be sorted soon.

PR #2245, opened 2016-07-26
Removed interactiveTestPort variable

anselmbradford:
I couldn't find anywhere that this was used. Seems safe to remove... or?

## Removals

- Removed `interactiveTestPort` test variable.

## Review

- @jimmynotjim 
- @imuchnik 
- @sebworks 
- @KimberlyMunoz


sebworks:
Not sure if it's necessary for us. This is the PR https://github.com/angular/protractor/pull/2010.

LGTM.

PR #2253, opened 2016-07-29
Add PageSpeed Insights performance test

contolini:
In preparation for our upcoming performance work, here's a gulp task that pipes localhost to [Google PageSpeed Insights](https://developers.google.com/speed/pagespeed/insights/). It'll let us see how local front-end changes affect browser performance.

## Additions

- `gulp test:perf` task

## Testing

- `npm install`
- `gulp test:perf` will run `localhost:8000` through PSI
- `gulp test:perf --u=about-us` will check `localhost:8000/about-us`

## Review

- @sebworks 
- @virginiacc 

## Screenshots

screen shot 2016-07-29 at 2 12 52 am

## Notes

- It uses [localtunnel](https://www.npmjs.com/package/localtunnel) to expose localhost to Google so complaints about GZIP compression can probably be ignored.
- The values in the second half of the report are "[rule impacts](https://developers.google.com/speed/docs/insights/v1/getting_started)" that are relative to each other and indicate how important the rule is. In the screenshot above, minimizing the render blocking resources is six times more valuable than prioritizing visible content (score of 48 vs 8). So we really need to stop loading jQuery from a CDN at the top of every page... cc @ascott1


sebworks:
Are there any security concerns with using `localtunnel`?

gulp/tasks/test.js
```diff
@@ -123,6 +124,30 @@ function _getWCAGParams() {
 }
 
 /**
+ * Processes command-line and environment variables
+ * for passing to the psi executable.
+ * An optional URL path comes from the command-line `--u=` flag.
+ * A PSI "strategy" (mobile vs desktop) can be specified with the `--s=` flag.
+ * @returns {Array} Array of command-line arguments for psi binary.
+ */
+function _getPSIParams( callback ) {
```
sebworks: Should this be named  `_createPSITunnel` since it's the main purpose of the function?

gulp/tasks/test.js
```diff
@@ -123,6 +124,30 @@ function _getWCAGParams() {
...
+  var commandLineParams = minimist( process.argv.slice( 2 ) );
+  var host = process.env.HTTP_HOST || 'localhost'; // eslint-disable-line no-process-env, no-inline-comments, max-len
+  var port = process.env.HTTP_PORT || '8000'; // eslint-disable-line no-process-env, no-inline-comments, max-len
+  var urlPath = _parsePath( commandLineParams.u );
+  var url = host + ':' + port + urlPath;
```
sebworks: Should this just be `path` instead of  `urlPath`?

sebworks:
💯 , this great. It might be nice to have a check that ensures that the server is running. Otherwise, you get a weird error.

higs4281:
Too slick! It runs against my local install of paying for college, so other projects can make use.

gulp/tasks/test.js
contolini: 👍

gulp/tasks/test.js
contolini: It's *also* grabbing command line but I think you're right that it's primary purpose is the tunnel.

contolini:
@sebworks I had some concerns about that at first but our local db dumps come from prod and we're not sending anything non-public to localtunnel (it never sees any admin pages just the same stuff that's on cf.gov) so I feel okay about it.

sebworks:
ok, you may want run it by Lamin. It's good on my end. 👍

virginiacc:
This is awesome! 👍 

Is there an issue for changing the way we load jQuery?

sebworks:
👍

contolini:
@virginiacc I think we load jQuery the way we do because the JS snippets that are loaded from Google Tag Manager rely on $ for DOM traversal. We'll need to log into GTM and convert the snippets to vanilla JS.

contolini:
Yep, I think I failed to update `npm-shrinkwrap.json`. Stand by.

PR #2261, opened 2016-07-29
Only run accessibility tests when requested by the user

contolini:
## Before

Running `gulp test` ran lots of tests including our new accessibility tests. This is annoying because it reports a lot of accessibility problems that we won't have time to address anytime soon.

## After

Running `gulp test` doesn't include the accessibility tests. Run `gulp test --a11y` to include them.

## Review

- @chosak 
- @richaagarwal 

## Notes

- It kills me to create a second config file (a11y_conf.js) but it was the only way I could figure out how to modify the `plugins` property via a command line option. Only the [onPrepare](https://github.com/angular/protractor/blob/a74175d4b37fed0b86ca9646b749585a5ff1ba52/docs/referenceConf.js#L195-L209) method seems to have access to the protractor globals and by then it's too late to add a plugin.

### DIFF
--- a/TEST.md
+++ b/TEST.md
@@ -154,8 +154,8 @@ run `gulp test:unit:server` from the command-line in the project root.
 
 # Accessibility Testing
 
-Whenever `gulp test:acceptance` is run, every webpage is checked for WCAG and
-Section 508 compliancy using Protractor's
+Run the acceptance tests with an `--a11y` flag (i.e. `gulp test:acceptance --a11y`)
+to check every webpage for WCAG and Section 508 compliancy using Protractor's
 [accessibility plugin](https://github.com/angular/protractor-accessibility-plugin).
 
 If you'd like to audit a specific page, use `gulp test:a11y`:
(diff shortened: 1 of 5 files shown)


chosak:
Nice approach. It's too bad that you can't modify the `config` object somehow once you're in that method.

Some `gulp lint` errors:

```sh
cfgov-refresh/gulp/tasks/test.js
   24:21  error    Unexpected use of process.env                     no-process-env
   82:38  error    ["a11y"] is better written in dot notation        dot-notation
```

(Although these don't actually fail the build for some reason.)

Minor request: update `TEST.md` [here](https://github.com/cfpb/cfgov-refresh/blob/master/TEST.md#accessibility-testing) to mention this command-line param?

richaagarwal:
@chosak Gulp lint errors aren't supposed to fail the build, afaik. 

Other than Andy's comments, 👍

contolini:
Fixed the lint warning and update the docs. Thanks!

chosak:
Thank you @contolini! 👍

PR #2262, opened 2016-07-29
Adds BASE_CLASS and dom validity check to multiselect.

anselmbradford:
## Changes

- Adds `BASE_CLASS` and dom validity check to multiselect to align it with the other atomic components lifecycle.
- Fixes some minor linter warnings.

## Testing

- `gulp test:unit:scripts && gulp test:acceptance --sauce=false` should pass (expect for the 2 acceptance test errors we're getting from wagtail on all PRs).

## Review

- @jimmynotjim 
- @sebworks 

## Todos

- Added a todo to scope the multiselect to markup that has the base class. If global behavior is needed, a separate script like this can be added to common.js or similar:
```
var mselects = document.querySelector( 'select[multiple]' );
for ( var i, len = mselects.length; i < len; i++ ) {
  mselects.classList.add( 'cf-multi-select' );
}
```

### DIFF
--- a/test/unit_tests/molecules/Multiselect-spec.js
+++ b/test/unit_tests/molecules/Multiselect-spec.js
@@ -32,7 +32,6 @@ describe( 'Multiselect', function() {
 
   beforeEach( function() {
     sandbox = sinon.sandbox.create();
-    sandbox.stub( window.console, 'log' );
 
     document.body.innerHTML = HTML_SNIPPET;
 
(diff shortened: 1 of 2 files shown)


anselmbradford:
@jimmynotjim added your comments from chat; ready for review

test/unit_tests/molecules/Multiselect-spec.js
```diff
@@ -59,14 +58,18 @@ describe( 'Multiselect', function() {
...
+        var choices =
+          document.querySelectorAll( '.cf-multi-select_choices li' );
 
         expect( choices.length ).to.equal( 1 );
         expect( choices[0].innerHTML ).to.contain( 'Debt collection' );
       }
     );
 
     it( 'should log a helpful tip if passed a bad option value', function() {
+      // Note: sandbox.stub console.log will prevent regular `console.log(…)`
```
jimmynotjim: Could probably just update this to a TODO to match the error one you added in the main file.

jimmynotjim:
Minor comment, otherwise 👍

test/unit_tests/molecules/Multiselect-spec.js
anselmbradford: Fixor'd

PR #2250, opened 2016-07-28
Fix bugs on leadership calendar

kurtrwall:
Related to [GHE]/CFGOV/platform/issues/131

There are template errors and mishandled PDF logic that is fixed with this.

## Additions

- Macros in the leadership-calendar-table.html template for modularity and clarity.

## Changes

- Failed PDFreactor calls return the print template rendered instead of just passing.
- Fixed print template macro call
- Refactored templates and view logic
- Fixed markup for "checked" calendars

## Testing

- Go to leadership calendar
- Filter on a calendar
- Filter on dates
- Click Download PDF button on the bottom controls form
- If PDFreactor is set up then it should return a pdf. Otherwise, it should direct you to the print page with all the events.

## Review

- @rosskarchner 
- @willbarton 
- @Scotchester 
- @chosak 
- @virginiacc 
- @contolini 

## Todos

- Tests

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #2263, opened 2016-07-29
Fixing error notification on the Leadership PDF Calendar

sebworks:
Fixing error notification on the Leadership PDF Calendar

**This PR contains commits from https://github.com/cfpb/cfgov-refresh/pull/2250**

## Changes

- Updated `cfgov/v1/util/util.py` to make the error message dynamic through string interpolation. 
- Updated `cfgov/jinja2/v1/about-us/the-bureau/leadership-calendar/print/index.html` to fix issue with logic and formatting. 
- Updated `cfgov/cal/views.py` to add validation for the `CalendarPDFForm1`.
- Updated `cfgov/jinja2/v1/_includes/macros/filters.html` to fix issue rendering the notification.

## Testing

- Disable Javascript in the browser.
- Visit the `http://localhost:8000/about-us/the-bureau/leadership-calendar/` page. 
- Click the `Download PDF` button. You should see an error notification once redirected. 
- Enter in `123456` in the `From :` field. Click the `Download PDF` button. You should see an error notification once redirected. 

## Review

- @richaagarwal 
- @chosak 
- @kurtrwall 

## Screenshots

screen shot 2016-07-29 at 5 50 56 pm

## Notes

- Most of the code was contained in https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/forms.py. It probably makes sense to delete that code at this time, since much of it has been migrated to views.py. 

- I am removing all query-string parameters when redirecting. It might make sense to add back the valid parameters. 

- There are two forms on the leadership-calendar page sharing the same query-string parameters. We might want to consider an alternative approach.

## Todos

- Write Tests. 

## Checklist


chosak:
Would it be possible to have the error redirect go to the download options section (via a hash on the URL) instead of to the top of the page? Currently if there's an error it's not immediately clear what happened and you have to scroll down to see the message.

cfgov/cal/views.py
```diff
@@ -10,17 +10,19 @@
...
 from django.template.loader import get_template
 from django.core.paginator import Paginator, EmptyPage, PageNotAnInteger
 from django.core.exceptions import ObjectDoesNotExist
 from django.db.models import Max, Min, Q
+from django.contrib import messages
 
 from cal.models import CFPBCalendar, CFPBCalendarEvent
-
+from v1.util.util import ERROR_MESSAGES
+from sheerlike.templates import get_date_string
```
chosak: Alphabetize imports.

cfgov/cal/views.py
```diff
@@ -31,6 +33,31 @@
...
+class FilterDateField(forms.DateField):
+    def clean(self, value):
+        if value:
+            try:
+                value = get_date_string(value)
+            except ValueError:
+                msg = self.error_messages['invalid']
+                if '%s'in msg:
+                    msg = msg % value
+                raise forms.ValidationError( msg , code='invalid')
```
chosak: Extra space before `msg`.

cfgov/cal/views.py
```diff
@@ -63,11 +120,13 @@ def display(request, pdf=False):
     """
     display (potentially filtered) html view of the calendar
     """
+
     template_name='about-us/the-bureau/leadership-calendar/index.html'
-    form = CalendarFilterForm(request.GET)
+    form = CalendarPDFForm(request.GET) if pdf else CalendarFilterForm(request.GET)
     context = {'form': form}
+    form_is_vaild = form.is_valid()
```
chosak: Typo, `vaild`.

cfgov/cal/views.py
```diff
@@ -122,8 +181,16 @@ def display(request, pdf=False):
             'range_end':range_end,
         })
 
-        if pdf and PDFreactor:
-            return pdf_response(request, context)
+        if pdf:
+            template_name = 'about-us/the-bureau/leadership-calendar/print/index.html'
+            if PDFreactor:
+                return pdf_response(request, context)
```
chosak: If you don't have PDFreactor installed (for example for local development or if the import fails for some reason), you just get redirected to the main page. Could the `pdf_response` method just return the generic PDF error message in that case instead?

cfgov/cal/views.py
```diff
@@ -122,8 +181,16 @@ def display(request, pdf=False):
...
+
+    elif form_is_vaild == False and pdf:
+        for error in form.errors.itervalues():
+            messages.error(request, str(error), extra_tags='leadership-calendar')
+
+        return HttpResponseRedirect('/the-bureau/leadership-calendar/')
```
chosak: Use `reverse` here instead of hardcoding URL.

cfgov/cal/views.py
```diff
@@ -150,4 +217,5 @@ def pdf_response(request, context):
         response['Content-Disposition'] = 'attachment; filename=cfpb-leadership.pdf'
         return response
     except (urllib2.HTTPError, urllib2.URLError, BadStatusLine):
-        pass
+        messages.error(request, 'Error Creating PDF', extra_tags='leadership-calendar')
+        return HttpResponseRedirect('/the-bureau/leadership-calendar/')
```
chosak: Use `reverse` here instead of hardcoding URL.

cfgov/cal/views.py
```diff
@@ -43,6 +70,36 @@ def clean_filter_calendar(self):
...
+    def clean(self):
+        cleaned_data = super(CalendarPDFForm, self).clean()
+        from_date_empty = 'filter_range_date_gte' in cleaned_data and \
+                          cleaned_data['filter_range_date_gte']  == None
+        to_date_empty = 'filter_range_date_lte' in cleaned_data and \
+                        cleaned_data['filter_range_date_lte'] == None
+
+        if from_date_empty and to_date_empty :
+            raise forms.ValidationError(ERROR_MESSAGES['DATE_ERRORS']['one_required'])
+        return cleaned_data
```
chosak: Entering invalid dates in the filter form (`CalendarFilterForm`) still raises an unhandled exception. Can/should this functionality be unified between the forms (through a common base class, for example), or is that outside the scope of this PR?

cfgov/cal/views.py
```diff
@@ -122,8 +181,16 @@ def display(request, pdf=False):
...
-        if pdf and PDFreactor:
-            return pdf_response(request, context)
+        if pdf:
+            template_name = 'about-us/the-bureau/leadership-calendar/print/index.html'
+            if PDFreactor:
+                return pdf_response(request, context)
+
+    elif form_is_vaild == False and pdf:
+        for error in form.errors.itervalues():
+            messages.error(request, str(error), extra_tags='leadership-calendar')
```
chosak: Using `str` here doesn't properly handle unicode values in the inputs. For example, try putting `ë` as one of the date fields in the PDF download form. This will raise an unhandled exception and cause a 500 error.

cfgov/cal/views.py
```diff
@@ -43,6 +70,36 @@ def clean_filter_calendar(self):
...
+        super(CalendarPDFForm, self).__init__(*args, **kwargs)
+
+    def clean_filter_calendar(self):
+        calendar_names = self.cleaned_data['filter_calendar']
+        calendars = CFPBCalendar.objects.filter(title__in=calendar_names)
+        return calendars
+
+    def clean(self):
+        cleaned_data = super(CalendarPDFForm, self).clean()
+        from_date_empty = 'filter_range_date_gte' in cleaned_data and \
```
vickumar1981: Comparisons to None should be done using the is and is not operators according to PEP 8.

https://www.python.org/dev/peps/pep-0008/

cfgov/cal/views.py
```diff
@@ -122,8 +181,16 @@ def display(request, pdf=False):
             'range_end':range_end,
         })
 
-        if pdf and PDFreactor:
-            return pdf_response(request, context)
+        if pdf:
+            template_name = 'about-us/the-bureau/leadership-calendar/print/index.html'
```
vickumar1981: ditto here, with using the url's reverse.  no magic strings.

cfgov/cal/views.py
```diff
@@ -63,11 +120,13 @@ def display(request, pdf=False):
...
     template_name='about-us/the-bureau/leadership-calendar/index.html'
-    form = CalendarFilterForm(request.GET)
+    form = CalendarPDFForm(request.GET) if pdf else CalendarFilterForm(request.GET)
     context = {'form': form}
+    form_is_vaild = form.is_valid()
 
-    if form.is_valid():
+    if form_is_vaild:
         cal_q = Q(('active', True))
         if form.cleaned_data.get('filter_calendar', None):
```
vickumar1981: Don't need None here.  it's the default optional for a get.

http://www.tutorialspoint.com/python/dictionary_get.htm

vickumar1981: Don't need None here. it's the default optional for a get.

http://www.tutorialspoint.com/python/dictionary_get.htm

cfgov/cal/views.py
```diff
@@ -10,17 +10,19 @@
 from rest_framework import generics
 
 from django.shortcuts import render, render_to_response
-from django.http import HttpResponse
+from django.http import HttpResponse, HttpResponseRedirect
 from django import forms
 from django.utils.safestring import mark_safe
```
vickumar1981: this import as well as a bunch of others are never used.

Here's the list:

date, datetime
from dateutil.relativedelta import relativedelta
from rest_framework import generics
from django.utils.safestring import mark_safe
from django.template import RequestContext
from django.core.exceptions import ObjectDoesNotExist

cfgov/cal/views.py
```diff
@@ -122,8 +181,16 @@ def display(request, pdf=False):
...
         })
 
-        if pdf and PDFreactor:
-            return pdf_response(request, context)
+        if pdf:
+            template_name = 'about-us/the-bureau/leadership-calendar/print/index.html'
+            if PDFreactor:
+                return pdf_response(request, context)
+
+    elif form_is_vaild == False and pdf:
```
vickumar1981: `elif not form_is_valid and pdf:`

cfgov/cal/views.py
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2263/commits/8b1241190c12d6ca8aef3094ad43dcefe3459d7e

cfgov/cal/views.py
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2263/commits/8b1241190c12d6ca8aef3094ad43dcefe3459d7e.

cfgov/cal/views.py
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2263/commits/8b1241190c12d6ca8aef3094ad43dcefe3459d7e.

cfgov/cal/views.py
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2263/commits/8b1241190c12d6ca8aef3094ad43dcefe3459d7e

cfgov/cal/views.py
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2263/commits/8b1241190c12d6ca8aef3094ad43dcefe3459d7e.

cfgov/cal/views.py
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2263/commits/8b1241190c12d6ca8aef3094ad43dcefe3459d7e.

cfgov/cal/views.py
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2263/commits/8b1241190c12d6ca8aef3094ad43dcefe3459d7e.

cfgov/cal/views.py
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2263/commits/8b1241190c12d6ca8aef3094ad43dcefe3459d7e.

cfgov/cal/views.py
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2263/commits/8b1241190c12d6ca8aef3094ad43dcefe3459d7e.

cfgov/cal/views.py
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2263/commits/7436a9bcc457106b52242f8ba888c4d225a79970.

cfgov/cal/views.py
sebworks: Hmmmm, I'm not sure on this issue . An alternative approach might be to just manually create the PDF. 

https://blog.idrsolutions.com/2013/01/understanding-the-pdf-file-format-overview/#helloworld

cfgov/cal/views.py
sebworks: Yeah, I think a base form or just creating a Date range widget probably makes sense. 

https://groups.google.com/forum/#!topic/django-filter/lbi_B4zYq4M

chosak:
Thanks for fixes! 👍

rosskarchner:
👍

kurtrwall:
👍  after merge conflicts

sebworks:
@cfpb/cfgov-backends, I'm not sure how to resolve the testing error. Any help would be appreciated.

cfgov/cal/views.py
```diff
@@ -46,6 +66,36 @@ def clean_filter_calendar(self):
         calendars = CFPBCalendar.objects.filter(title__in=calendar_names)
         return calendars
 
+class CalendarPDFForm(forms.Form):
+    filter_calendar = FilterCheckboxList(
+            choices = [(c.title, c.title) for c in CFPBCalendar.objects.all()],
```
chosak: This usage of `CFPBCalendar.objects.all()` is problematic here because it means you are trying to make a DB query at module import time, which isn't always possible if the database isn't ready.

cfgov/cal/views.py
```diff
@@ -31,6 +26,32 @@
...
+        if value:
+            try:
+                get_date_string(value)
+            except ValueError:
+                msg = self.error_messages['invalid']
+                if '%s'in msg:
+                    msg = msg % value
+                raise forms.ValidationError(msg , code='invalid')
+
+        return super(FilterDateField, self).clean(value)
```
chosak: @sebworks I modified this code; previously it was returning the output of `get_date_string` which was a string, but what I think needs to happen here is that it needs to call the parent `DateField.clean`, which actually [converts it](https://github.com/django/django/blob/master/django/forms/fields.py#L398) to a Python date. If I didn't do this I'd get errors when trying to generate a filtered output.

chosak:
@sebworks I cleaned up those broken tests a bit and got them passing. I'm able to validate the error message behavior and also generate print layout calendars locally. I don't have PDFreactor locally so can't test regular PDF generation. Is this something that you can validate?

sebworks:
> @sebworks I cleaned up those broken tests a bit and got them passing. I'm able to validate the error message behavior and also generate print layout calendars locally. I don't have PDFreactor locally so can't test regular PDF generation. Is this something that you can validate?

@chosak, I don't know of a way to test locally. Maybe we could use a virtual box?

cfgov/cal/views.py
sebworks: So we want to override the value so the form will accept flexible date formats like : 

`04/2016`
`2016`

I think the code is good; the only change would be to override the value like:

`value = get_date_string(value)`

sebworks:
@chosak, thanks for helping me on this PR.

cfgov/cal/views.py
chosak: Thanks, I understand now. That's a helpful clarification. I'll correct and also add unit tests here.

chosak:
@sebworks I refactored the forms and form elements into a separate module and added a whole bunch of tests around the expected behavior. I tested this locally using calendar data and filtering and download seem to work as expected. Would you mind taking another look?

cfgov/cal/tests/test_forms.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def test_no_value_not_required(self):
+        self.get_field(required=False).validate([])
+
+    def test_no_value_required(self):
+        field = self.get_field()
+        self.assertRaises(forms.ValidationError, field.validate, None)
+
+    def test_invalid_value_passes(self):
+        self.get_field().validate('d')
```
chosak: @sebworks just wanted to call out this test specifically. Currently because the calendar options are hardcoded on the frontend, it's possible that a user might make a query for a calendar even when there's no data on the backend. This happens if you test this code against a local DB without calendar data, for example. The pre-existing behavior is that this search just returns nothing silently, instead of actually failing validation and reporting an error to the user. So my changes to the form code and these tests just keep that existing behavior.

cfgov/cal/forms.py
```diff
@@ -0,0 +1,77 @@
...
+
+class CalendarPDFForm(CalendarFilterForm):
+    def __init__(self, *args, **kwargs):
+        kwargs['error_class'] = FilterErrorList
+        super(CalendarPDFForm, self).__init__(*args, **kwargs)
+        self.fields['filter_calendar'].required = True
+
+    def clean(self):
+        super(CalendarPDFForm, self).clean()
+        from_date = self.cleaned_data.get('filter_range_date_gte')
```
sebworks: If we call clean first and an invalid value is submitted, it will be assigned a value of `None` via the `clean` step. Which means we will get two validation errors: One for the invalid date and the other for not submitting at least one date value. Even if the date is invalid, it should still meet the one date requirement because an validation error will get raised in the `clean` step.


You can reproduce this by submitting '123451' and noticing we get two errors.

sebworks:
💯 , this is really clean. Only the one comment.

cfgov/cal/forms.py
chosak: Thanks for catching this @sebworks. Just to clarify one thing: when a Django form [tries to clean](https://github.com/django/django/blob/master/django/forms/forms.py#L383) a field, if that `clean` method raises a `ValidationError`, then the field doesn't get added to `form.cleaned_data` at all (nort assigned a value of `None`). Instead an error for that field gets added to `form.errors` (which can be checked with `form.has_error`). It's only if the field is empty that it gets added to `form.cleaned_data` as `None`.

sebworks: Thanks for the explanation @chosak.

PR #2265, opened 2016-08-01
Rewrite and restructure browser tests --> python tests

richaagarwal:
The "Wagtail" suite of browser tests have now all been converted to unit tests that use a test database. A lot of things have been rewritten and moved around, but generally **there is no new test coverage nor have any tests been (functionally) removed**.

There is no longer an `initial_test_data` as the data is created for each test that needs it, although I make use of `initial_data` as Andy set up in this PR: https://github.com/cfpb/cfgov-refresh/pull/2269

## Testing

As before, you can run `gulp test`, but this just runs `gulp lint` and the unit tests now. It is faster to do the following:

`tox -e fast`
`gulp lint`

You can also still run integration (which need to be updated as well) and content tests like so, but as before, they will fail unless you have the relevant prod data locally.
`gulp test:acceptance --suite=integration`
`gulp test:acceptance —suite=content`

## Review

- @chosak 
- @rosskarchner 
- @willbarton 
- @grapesmoker 

## Todos

- There are two tests commented out in `test_organisms.py` relating to the e-mail in Contact Info, as they are failing and they need to be looked into further. They both make the following assertion:
`self.assertContains(response, 'test@example.com')`

## Checklist


cfgov/v1/tests/wagtail/test_molecules.py
```diff
@@ -0,0 +1,206 @@
...
+from v1.models.landing_page import LandingPage
+from v1.models.sublanding_page import SublandingPage
+from v1.models.learn_page import LearnPage, DocumentDetailPage
+from v1.models.browse_page import BrowsePage
+from v1.models.browse_filterable_page import BrowseFilterablePage
+from v1.models.sublanding_filterable_page import SublandingFilterablePage
+from wagtail.wagtailcore.blocks import StreamValue
+from helpers import publish_page
+
+c = Client()
```
vickumar1981: might want to rename that variable.  also, if multiple tests are using this, which it looks like they are, then it makes sense to move it into a base class and have your tests inherit from your base test class.

cfgov/v1/tests/wagtail/test_organisms.py
```diff
@@ -0,0 +1,224 @@
...
+				atomic.contact_address
+			], 
+			True
+		)
+		contact.save()
+		return contact 
+
+	def test_well(self):
+		"""Well content correctly displays on a Landing Page"""
+		landing_page = LandingPage(
```
vickumar1981: a lot of these tests repeat similar blocks:

- create a page
- add either content or heading 
- publish it
- assert that the page contains a known value

i think maybe we can refactor some of that out to reduce redundancy.

richaagarwal: @vickumar1981 That was somewhat intentional, at least for now, as I wanted to move the tests generally as-is into this new format.  I do tend to prefer verbosity over abstraction, fwiw. For each step you've outlined, it's already one line, except for creating a page, which will be different each time anyway (although open to suggestions on how to deal with those). Any functions that were actually used multiple times, I put into `helpers.py`.

richaagarwal: Just want to add a little bit about what I mean by preferring verbosity - 

I think it should be easy to read a test, not have to decipher it for every single one by looking up the function that's doing the bulk of the work. A little bit is fine, but if I were to refactor it in the way you seem to be suggesting, the tests on their own would not be that readable.

cfgov/v1/tests/wagtail/test_molecules.py
```diff
@@ -0,0 +1,206 @@
...
+
+class MoleculesTestCase(TestCase):
+
+	def test_text_intro(self):
+		"""Text introduction value correctly displays on a Browse Filterable Page"""
+		bfp = BrowseFilterablePage(
```
vickumar1981: does this function take 3 arguments?

cfgov/v1/tests/wagtail/test_organisms.py
vickumar1981: readability and abstraction aren't mutually exclusive.  abstraction helps to make the code more readable.

Robert Martin says, "Duplicate code always represents a missing abstraction"

http://www.informit.com/articles/article.aspx?p=1313447

This is also fairly intuitive to understand.  Say, the slug-name changes for the landing page, or the url we want to load the response from.  As written, one would have to make that fairly simple change in a lot of different places, increasing the time and cost it takes to maintain.

cfgov/v1/tests/wagtail/test_organisms.py
```diff
@@ -0,0 +1,224 @@
...
+		)
+		contact = self.get_contact()
+		sublanding_page.content = StreamValue(
+			sublanding_page.content.stream_block, 
+			[atomic.main_contact_info(contact.id)],
+			True
+		)
+		publish_page(child=sublanding_page)
+		response = c.get('/sublanding/')
+		# self.assertContains(response, 'test@example.com')
```
vickumar1981: remove commented code.

cfgov/v1/tests/wagtail/test_organisms.py
```diff
@@ -0,0 +1,224 @@
...
+		)
+		contact = self.get_contact()
+		landing_page.sidefoot = StreamValue(
+			landing_page.sidefoot.stream_block, 
+			[atomic.sidebar_contact(contact.id)],
+			True
+		)
+		publish_page(child=landing_page)
+		response = c.get('/landing/')
+		# self.assertContains(response, 'test@example.com')
```
vickumar1981: ditto.

cfgov/v1/tests/wagtail/test_page_states.py
```diff
@@ -0,0 +1,99 @@
+from django.test import TestCase
+from django.test import Client
+
+from scripts import _atomic_helpers as atomic
```
vickumar1981: i don't think this import is used.

cfgov/v1/tests/wagtail/test_organisms.py
richaagarwal: @vickumar1981 I understand the urge to abstract this and I think it makes sense in most cases to do what you're saying. In this case, though, these tests may appear to be duplicative but they're not.  The name of the slug could be anything; it's only dependent on the test data I create, which is created in each individual test.  I think I could have a few more helper methods, like "initialize_landing_page" for pages that are created more than once.

cfgov/v1/tests/wagtail/test_organisms.py
richaagarwal: There is a note about this in my PR summary. Please read that before continuing to review, thanks.

cfgov/v1/tests/wagtail/test_page_states.py
```diff
@@ -0,0 +1,99 @@
...
+		shared_draft.title = 'Draft Page Updates'
+		shared_draft.shared = False
+		save_page(page=shared_draft)
+
+		www_response = c.get('/page/')
+		self.assertEqual(www_response.status_code, 404)
+
+		staging_response = c.get(
+			'/page/', 
+			HTTP_HOST=os.environ.get('DJANGO_STAGING_HOSTNAME')
```
vickumar1981: I think this variable can be loaded once.  the 'DJANGO_STAGING_HOSTNAME' environment setting isn't going to change between tests.

cfgov/v1/tests/wagtail/test_page_titles.py
```diff
@@ -0,0 +1,99 @@
+from django.test import TestCase
+from django.test import Client
+
+from scripts import _atomic_helpers as atomic
```
vickumar1981: All the imports from here on below don't look like they're being used.

richaagarwal: They are, see `globals()[page_type]`.

cfgov/v1/tests/wagtail/test_organisms.py
```diff
@@ -0,0 +1,224 @@
...
+		learn_page.content = StreamValue(
+			learn_page.content.stream_block, 
+			[atomic.table],
+			True
+		)
+		publish_page(child=learn_page)
+		response = c.get('/learn/')
+		self.assertContains(response, 'table hyperlink')
+		self.assertContains(response, 'table text')
+		self.assertContains(response, 'table text blob')
```
vickumar1981: if it contains 'table text blob', doesn't that mean it also contains 'table text'?

cfgov/v1/tests/wagtail/test_page_titles.py
vickumar1981: ah, my fault.  👍

cfgov/v1/tests/wagtail/test_molecules.py
willbarton: I would agree with renaming the variable... especially if you're looking at the file for a single test, it may be non-obvious at first what `c` is.

cfgov/v1/tests/wagtail/test_organisms.py
willbarton: I agree with @richaagarwal here. I'm coming to the codebase new to Wagtail, and I'm still trying to wrap my head around some of the ways we're using it. For projects with good tests, I often find they're some of the best ways into learning how the code works. I like these broken out the way they are.

cfgov/v1/tests/wagtail/test_organisms.py
richaagarwal: Ah yes, thanks.  I took these assertions from the previous code and didn't notice that.

cfgov/v1/tests/wagtail/test_page_states.py
richaagarwal: Thanks, will remove!

willbarton:
@richaagarwal I think overall this looks fantastic, and a lot better than the tests that relied on `initial_test_data.py`. 

I can't run them locally, however, for the same reason that Travis fails:

```
  File "/Users/bartonw/repos/platform/cfgov-refresh/cfgov/cfgov/test.py", line 9, in <module>
    from scripts import initial_data, initial_test_data
ImportError: cannot import name initial_test_data
ERROR: InvocationError: '/Users/bartonw/repos/platform/cfgov-refresh/.tox/fast/bin/python manage.py test'
```

[It looks like `initial_test_data.run()` is removed in `test.py`, but the import isn't.](https://github.com/cfpb/cfgov-refresh/commit/edba2c9f60110fcb6b3144a98ef384f1dd4bf1d2#diff-a21fa67a0ea8b49d72120a14a91a30ebL20)

With that import removed, it works as advertised.

richaagarwal:
@willbarton Thanks! And good catch; for some reason they were still running fine for me. Removing the import now.

willbarton:
@richaagarwal that did it. 👍

cfgov/v1/tests/wagtail_pages/helpers.py
```diff
@@ -0,0 +1,19 @@
+from v1.models.home_page import HomePage
+from django.contrib.auth.models import User
+
+
+def save_page(page):
+	admin_user = User.objects.get(username='admin')
+	page.save()
```
chosak: Any test that publishes a page ends up altering the database state for other tests. How would you feel about adding `tearDown` methods to the tests that cleanup the changes they make, so as to keep tests independent?

cfgov/v1/tests/wagtail_pages/test_organisms.py
```diff
@@ -0,0 +1,223 @@
...
+        contact.contact_info = StreamValue(
+            contact.contact_info.stream_block,
+            [
+                atomic.contact_email,
+                atomic.contact_phone,
+                atomic.contact_address
+            ], 
+            True
+        )
+        contact.save()
```
chosak: Same comment above regarding altering database state. Can this test case have a `tearDown` that cleans up these contacts?

cfgov/v1/tests/wagtail_pages/test_page_titles.py
```diff
@@ -0,0 +1,99 @@
...
+            globals()[page_type](
+                title = 'Title ABCD',
+                slug = 'page',
+            )
+        )
+        response = django_client.get('/page/')
+        self.assertContains(response, 'Title ABCD')
+
+    def test_sublanding_page(self):
+        self.page_loads_with_correct_title(page_type = 'SublandingPage')
```
chosak: PEP8 [discourages](https://www.python.org/dev/peps/pep-0008/#pet-peeves) these extra spaces.

cfgov/v1/tests/wagtail_pages/test_page_titles.py
```diff
@@ -0,0 +1,99 @@
...
+'''
+
+django_client = Client()
+
+class PageTitlesTestCase(TestCase):
+    """ Tests that all Wagtail page types load and display the title correctly """
+
+    def page_loads_with_correct_title(self, page_type):
+        publish_page(
+            globals()[page_type](
```
chosak: Any reason you can't just refer to the class directly, e.g.

```python
def page_loads_with_correct_title(self, page_cls):
    publish_page(page_cls(title='foo', slug='bar'))

def test_sublanding_page(self):
    self.page_loads_with_correct_title(SublandingPage)
```

richaagarwal: Ah, I didn't know you could do that!  I'll make that change later; I don't think it's critical.

richaagarwal: Actually nevermind I'll do it now, simple enough. Thanks!

cfgov/v1/tests/wagtail_pages/test_page_titles.py
richaagarwal: Which ones @chosak ?

willbarton: `self.page_loads_with_correct_title(page_type = 'SublandingPage')` 

should be 

`self.page_loads_with_correct_title(page_type='SublandingPage')`

richaagarwal: @willbarton @chosak I'm trying to find it but I don't see anything in the PEP guidelines that Andy linked to that states no spaces around =.

richaagarwal: In fact, it says the opposite: "Always surround these binary operators with a single space on either side."

willbarton: It's in [Other recommendations, under Whitespace](https://www.python.org/dev/peps/pep-0008/#other-recommendations):

> Don't use spaces around the = sign when used to indicate a keyword argument or a default parameter value.

This is the sort of thing that, if we were running a Python linter in Travis like we are for JavaScript, it would definitely complain about.

richaagarwal: Ah okay thanks for linking me!  I will update them.

vickumar1981: There's also a few other places where the PR isn't compliant with PEP8, btw.   There's a couple of ways, I think, that you guys could decide upon if you want to standardize the code formatting.

1)  pip has a module for checking pep8 compliance: https://pypi.python.org/pypi/pep8

2) put the settings in your IDE.  if you use sublime, you can use something like this: https://packagecontrol.io/packages/Python%20PEP8%20Autoformat, and it'll automatically point out wherever there's some PEP8 issue.  If you use PyCharm, that feature should be built in.

richaagarwal:
Updated the spacing.  Going to merge now but still open to additional feedback. Thanks all for your input!

anselmbradford:
Should the atomic components in the `/integration/` directory retain the `/molecules/`, `/organisms/`, etc. subdirectories that are used in the implementations (https://github.com/cfpb/cfgov-refresh/tree/master/cfgov/unprocessed/js/molecules, https://github.com/cfpb/cfgov-refresh/tree/master/cfgov/unprocessed/js/organisms)?

PR #2266, opened 2016-08-01
Send new users password reset email

chosak:
Currently new users in Wagtail [must](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/signals.py#L35) reset their password after their account has been created through the admin. Previously they'd have to try to login first with a temporary password and then request a reset email. This PR automatically sends a password reset email to users created through the Wagtail admin (`/admin/users/add/`).

## Additions

- Custom `WAGTAIL_USER_CREATION_FORM` setting overrides default admin form to use `v1.auth_forms.UserCreationForm`, which triggers a user email send on save.
- Added new `v1.email` module (and relevant tests) to contain email-related methods.
- Added new Django admin action to send password reset emails to users.

## Testing

- Run `tox` to verify that all Python unit tests pass, including the new ones in this PR.
- To manually test on a local server:
 1. Open the Wagtail admin to [http://localhost:8000/admin/users/] (or wherever you're running your dev server) 
 1. Click "Add a User" (in the upper right)
 1. Choose a username, email (doesn't matter for local testing), first and last name, and password
 1. Select roles (if desired) and then click "Add User"
 1. Note debug server output in console includes text of password reset email, including link to URL at [http://localhost:8000] (or wherever you were running your dev server)
- To verify that email links properly use the default Wagtail Site:
 1. Open the Wagtail admin to Site configuration at [http://localhost:8000/admin/sites/] (or wherever you're running your dev server)
 1. Change the default site to some custom hostname (e.g. `testhost`) and set the port to 443
 1. Click "Save" to commit your changes (no need to restart the dev server)
 1. Follow the manual test steps described above.
 1. Note debug server output now includes link to URL at [https://testhost](https://testhost) (or whatever hostname you chose)
 1. Revert default Site changes (if you wish) to restore previous functionality
- To verify the Django admin functionality:
 1. Open the Django admin to the user management page at [http://localhost:8000/django-admin/auth/user/] (or wherever you're running your dev server)
 1. Check the checkbox next to one or more users
 1. From the custom actions dropdown, choose "Send password reset email", and click "Go"
 1. Note debug server output showing password reset email(s)

## Review

- @cfpb/cfgov-backends 

## Screenshots

Sending password reset emails through the Django admin:

image

After sending the password reset emails:

image

## Notes

- Sending users email like this is complicated slightly by the need to know the hostname, port, and scheme of the server you want message links to go to. After investigating several alternatives, using the Wagtail [Site](http://docs.wagtail.io/en/v1.5.3/reference/pages/model_reference.html#site) framework seemed to be the most straightforward approach, especially because this framework is already being used elsewhere. This PR uses the `Site` with `is_default_site=True` (there can be only one) as the base URL to use for email links.

## Checklist


cfgov/v1/auth_forms.py
```diff
@@ -122,4 +122,11 @@ def confirm_login_allowed(self, user):
             raise ValidationError("This account is temporarily locked; please try later or <a href='/admin/password_reset/' style='color:white;font-weight:bold'>reset your password</a>")
 
 
+class UserCreationForm(wagtailforms.UserCreationForm):
+    def save(self, commit=True):
+        user = super(UserCreationForm, self).save(commit=commit)
 
+        if commit:
+            send_password_reset_email(user.email)
```
chosak: It would nice to be able to put a `try/except` around this, and then log a warning if the email send fails (but still return to the Wagtail admin). Is there an existing convention about logging this kind of non-critical error?

cfgov/v1/email.py
```diff
@@ -0,0 +1,42 @@
+import os
```
vickumar1981: i don't think this import is used.

cfgov/v1/admin.py
```diff
@@ -1,7 +1,28 @@
...
+
+admin.site.unregister(User)
+
+@admin.register(User)
+class UserAdmin(UserAdmin):
+    actions = UserAdmin.actions + ['send_password_reset_email']
+
+    def send_password_reset_email(self, request, queryset):
+        for user in queryset:
+            send_password_reset_email(user.email, request)
```
vickumar1981: this might be problematic if you choose a lot of users.

Django has a built-in function to send bulk emails in django.core.mail.send_mass_mail().

chosak: Thanks, that's good feedback. There's lots of ways this is suboptimal: what if one of the sends fails for some reason, then it breaks halfway through, then you lose track of who was mailed? Also as you mention if there are too many emails selected, the request might take too long and cause the server to time out. Using [send_mass_mail](https://docs.djangoproject.com/en/1.8/topics/email/#send-mass-mail) instead would require a significant rewrite of the email sending code (actually implementing it, instead of just piggybacking on Django's [PasswordResetForm](https://github.com/django/django/blob/1.8.14/django/contrib/auth/forms.py#L195) as I do now). I thought this was a reasonable balance for what is really just a helper admin utility.

richaagarwal: If we're going to overload this function, can we call the args more explicitly - e.g. `email = user.email, request = request` ?

kurtrwall:
I am not able to produce the debug output of the reset link.

chosak:
@kurtrwall which step, specifically?

kurtrwall:
> To manually test on a local server:
 - Open the Wagtail admin to http://localhost:8000/admin/users/
  - Click "Add a User" (in the upper right)
  - Choose a username, email (doesn't matter for local testing), first and last name, and password
  - Select roles (if desired) and then click "Add User"
  - Note debug server output in console includes text of password reset email, including link to URL at http://localhost:8000

kurtrwall:
`Note debug server output in console includes text of password reset email, including link to URL at http://localhost:8000` this one.

chosak:
@kurtrwall I just retested this branch from scratch, using these steps:

1. `$ ./drop-db.sh`
1. `$ ./initial-data.sh`
1. `$ python cfgov/manage.py createsuperuser`
1. Try logging in as new superuser at `http://localhost:8000/admin`, follow password reset flow
1. Open superuser pw reset link (dumped in debug server output) to set superuser pw
1. Login to admin at `http://localhost:8000/admin` as superuser
1. Follow manual test steps described above to see correct pw link in debug output:

image

Do you have any custom environment settings, or anything else that might be affecting how this is working?

kurtrwall:
Looks like it was my environment. Thanks for the responses. 👍

rosskarchner:
👍 and like I said in chat, I really like the form submission approach.

cfgov/v1/auth_forms.py
willbarton: Do you want to open a ticket in the GHE platform repo to discuss this? I doubt we'll reach any conclusions for this PR, but we should definitely be logging things like that, and it'd be nice to establish a convention and document it (in @grapesmoker's awesome docs collection).

willbarton:
Have not tested it locally, but like @rosskarchner I really like using the form submission.

chosak:
Thanks for the review @kurtrwall @rosskarchner @willbarton.

PR #2268, opened 2016-08-01
Minor activity log test js cleanup

anselmbradford:
Minor activity log test js cleanup

## Changes

- Minor line length cleanup

## Review

- @virginiacc 
- @sebworks 
- @contolini

### DIFF
--- a/test/browser_tests/page_objects/page_activity-log.js
+++ b/test/browser_tests/page_objects/page_activity-log.js
@@ -2,12 +2,11 @@
 
 var filter = require( '../shared_objects/filter.js' );
 var filterableListControl =
-require( '../shared_objects/filterable-list-control.js' );
+  require( '../shared_objects/filterable-list-control.js' );
 var pagination = require( '../shared_objects/pagination' );
 
 function ActivityLog() {
-  Object.assign( this, filter, filterableListControl,
-    pagination );
+  Object.assign( this, filter, filterableListControl, pagination );
 
   this.get = function() {
     browser.get( '/activity-log/' );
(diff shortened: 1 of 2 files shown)


sebworks:
👍

test/browser_tests/page_objects/page_activity-log.js
```diff
@@ -2,12 +2,11 @@
 
 var filter = require( '../shared_objects/filter.js' );
 var filterableListControl =
-require( '../shared_objects/filterable-list-control.js' );
+  require( '../shared_objects/filterable-list-control.js' );
```
sebworks: Why the space here?

anselmbradford: I thought we indented after an equals on the newline to show it was part of the previous line when reading down the left-hand margin, to avoid confusion with something like this:

```
var filterableListThingy;
require( '../filterable-list-dadoo.js' );
```


but looks like we have both styles ([indent](https://github.com/cfpb/cfgov-refresh/blob/master/test/browser_tests/page_objects/page_leadership-calendar.js#L25), [indent](https://github.com/cfpb/cfgov-refresh/blob/master/test/browser_tests/shared_objects/filter.js#L4), [no indent](https://github.com/cfpb/cfgov-refresh/blob/master/test/browser_tests/page_objects/page_newsroom.js#L30), [no indent](https://github.com/cfpb/cfgov-refresh/blob/master/test/browser_tests/page_objects/page_blog.js#L31)), so update to whatever you'll are using.

sebworks: awwww, let's go with the indentation. I'll use that going forward.

PR #2269, opened 2016-08-02
Setup initial test data when running Python unit tests

chosak:
This PR tries to setup required initial test data before running the Python unit tests, by implementing a custom Django TestRunner class.

## Additions

- Added new `cfgov/cfgov/test.py` file that contains a `cfgov.test.TestDataTestRunner` class.

## Removals

- Removed check against `settings.DEBUG` in `cfgov/scripts/initial_data.py` as this seemed like an unnecessary check vs. just not running the script. Is there anywhere this script is ever run in production?

## Changes

- Python testing using `manage.py test` now uses the custom test runner, which does these steps before running the tests:
 1. Checks to see if the Wagtail [initial_data](https://github.com/torchbox/wagtail/blob/v1.5.3/wagtail/wagtailcore/migrations/0002_initial_data.py) migration has been run (it normally is when running `tox` or `manage.py test`, but not when running `tox -e fast` which skips all migrations), and, if not, runs it to create the Wagtail root page and related models.
 1. Runs our `cfgov/scripts/initial_data.py` to setup initial pages like home, Events, and Archive.
 1. Runs our `cfgov/scripts/initial_test_data.py` to setup specific testing data.

## Testing

- Run `tox` and `tox -e fast` and notice that the initial test data scripts are run before the tests are.

## Review

- @cfpb/cfgov-backends 

## Todos

- If this approach makes sense, we could consider having the Django unit test runner also kick off the Protractor browser tests, so that those can run against the test database.

## Checklist

### DIFF
--- a/cfgov/scripts/initial_data.py
+++ b/cfgov/scripts/initial_data.py
@@ -9,66 +9,65 @@ from v1.models import HomePage, BrowseFilterablePage
 
 
 def run():
-    if settings.DEBUG:
-        print 'Running script \'scripts.initial_data\' ...'
-        admin_user = None
-        site_root = None
-        events = None
+    print 'Running script \'scripts.initial_data\' ...'
+    admin_user = None
+    site_root = None
+    events = None
 
-        admin_user = User.objects.filter(username='admin')
-        if not admin_user:
-            admin_user = User(username='admin',
-                              password=make_password(os.environ.get('WAGTAIL_ADMIN_PW')),
-                              is_superuser=True, is_active=True, is_staff=True)
-            admin_user.save()
-        else:
-            admin_user = admin_user[0]
+    admin_user = User.objects.filter(username='admin')
+    if not admin_user:
+        admin_user = User(username='admin',
+                            password=make_password(os.environ.get('WAGTAIL_ADMIN_PW')),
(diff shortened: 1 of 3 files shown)


richaagarwal:
👍

richaagarwal:
I've merged this PR into my branch `rewrite-wagtail-tests` where it is useful. Should we leave this PR open as a stand-alone or wait until my PR is merged in?

cfgov/scripts/initial_data.py
```diff
@@ -9,66 +9,65 @@
 
 
 def run():
-    if settings.DEBUG:
-        print 'Running script \'scripts.initial_data\' ...'
-        admin_user = None
-        site_root = None
-        events = None
+    print 'Running script \'scripts.initial_data\' ...'
```
willbarton: Can we `from __future__ import print_function` and use `print()` here, since we're changing it?

chosak: Thanks, good suggestion. Implemented in 041e15c.

chosak:
Per discussion with @richaagarwal, going to merge this in to keep PR small and unblock other work that might depend upon this test data.

PR #2271, opened 2016-08-03
Minor cleanup

anselmbradford:
## Changes

- Adds/removes/cleans up comments.
- Removes trailing whitespace.
- Updates `bower.json` and `package.json` with current changelog version.

## Review

- @sebworks 
- @virginiacc

### DIFF
--- a/drop-db.sh
+++ b/drop-db.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # ==========================================================================
-# Setup script for MYSQL Databse Instantiation
+# Setup script used to drop MySQL database.
 # NOTE: Run this script while in the project root directory.
 # ==========================================================================
 
@@ -35,10 +35,3 @@ else
   $MYSQL -uroot -e "$SQL"
   ok "Database $1 dropped."
 fi
-
-
-
-
-
-
-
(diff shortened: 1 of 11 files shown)


sebworks:
👍

drop-db.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # ==========================================================================
-# Setup script for MYSQL Databse Instantiation
+# Setup script for MySQL database instantiation—used to drop the database.
```
sebworks: Should this just be `Setup script used to drop MySQL database`. It's not as though the script has any knowledge on how it's being used.

anselmbradford: Fixed!

initial-test-data.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # ==========================================================================
-# Initialization script for a wagtail user and imported data
+# Initialization script for a wagtail user and imported data.
```
sebworks: Should this ` Initialization script for a wagtail user and imported test data.`?

sebworks:
💯, merge at will.

initial-test-data.sh
anselmbradford: Fixed!

PR #2272, opened 2016-08-03
created the first part of the manual via mkdocs

grapesmoker:
This is the first step to putting together a comprehensive manual of how cfgov-refresh operates. I've included instruction on installation and also building/deployment. Next step is front-end stuff.


grapesmoker:
You can see the compiled form of the manual by doing
```
pip install mkdocs
cd manual
mkdocs serve
```

manual/docs/build-and-deploy.md
```diff
@@ -0,0 +1,214 @@
...
+The following teams and individuals have the necessary access to deploy cf.gov releases:
+
+- Ross Karchner
+- Scott Cranfill
+- Will Barton
+- Bill Higgins
+- Serghei Gorobet
+- The Software Delivery team
+
+### Feature flags
```
willbarton: When I added this section and the Wagtail pages vs Django views section below to the working doc, I envisioned them being separate in their final Markdown form.

manual/docs/build-and-deploy.md
```diff
@@ -0,0 +1,214 @@
...
+{% endif %}
+```
+
+Feature flags are implemented as a Django app in cfgov-refresh/cfgov/flags. 
+
+#### Feature Flag Hygiene
+
+Feature flags should be rare and ephemeral. Changes should be small and frequent, and not big-bang releases, and flags that are no longer used should be cleaned up and removed from code and the database.
+
+### Wagtail pages vs. Django views
```
willbarton: This section generally was a work-in-progress. We can go ahead and add it, but there may be heavy modification necessary. I'd prefer to have some input from @rosskarchner and @richaagarwal and anyone else who knows Wagtail as to whether this is a reasonable starting point or not.

richaagarwal: This seems like a fine starting point for me as we want to encourage content be created via Wagtail going forward. I'm not sure the HelloWorldPage example is a useful one though as, to get the benefits of the CMS, that should be defined via page attributes.

manual/docs/build-and-deploy.md
grapesmoker: I'm happy to break this out however people want. Or feel free to move it around if you like. I just took the whole page that was available and turned it into Markdown.

grapesmoker:
closing this in favor of the PR that follows our standard procedure

Issue #2264, opened 2016-07-29
[BUG] Some headers not receiving correct CSS/JS


anselmbradford:
## Current behavior

- On-demand header fails in IE8 because `Node` is not defined in https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/js/modules/util/behavior.js#L74

## Expected behavior

- Check existence of `Node` support before calling it or check `nodeType` instead (see http://stackoverflow.com/questions/13377887/javascript-node-undefined-in-ie8-and-under)

## Steps to replicate behavior

1. Run `gulp scripts:ondemand`
2. Open `/test-fixture/` in sauce labs (won't work in local VM unless localhost routing is set up).
3. See error in JS console that a function is expected for Node.

## Screenshots

__Current__

![screen shot 2016-07-29 at 7 42 49 pm](https://cloud.githubusercontent.com/assets/704760/17266307/aaffdfe6-55c4-11e6-8304-5081b5810e3a.png)

anselmbradford:
In on-demand header `behaviorElement` where this error is thrown is actually the string `a[href^="#"]``, but it hits the second `else if` that checks `Node` before getting to the third `String` check.

sebworks:
I thought IE8 was getting the no JS experience.

anselmbradford:
@sebworks ah right, good point! The on-demand header must not be falling back properly.

sebworks:
@anselmbradford, can i close this issue?

anselmbradford:
I believe this is still a cfgov-refresh issue in that https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/legacy/templates/front/base_update.html looks like it needs [this code](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/legacy/templates/front/base_nonresponsive.html#L128-L136). @rosskarchner do you know what pages are using `base_update.html` linked above?

PR #2277, opened 2016-08-04
Add/Correct IE8/9 fallback code for base and on-demand templates

anselmbradford:
base_update.html does not include IE8 fallback script to set IE visitors to the no-js state, very likely leading to JS errors on the pages that use that template.

Fixes https://github.com/cfpb/cfgov-refresh/issues/2264
The on-demand header was not including the nonresponsive CSS when it was in IE8. This PR includes the correct CSS where needed.

## Additions

- Adds conditional code blocks for IE8/9.
- Adds non-responsive CSS for IE8.

## Changes

- General code style cleanup of inline JS in template.

## Removes

- Inline JS for adding js/no-js flags that should be set in modernizr.

## Testing

- No sure how to test this. @rosskarchner do you know how to test changes to the `base_update.html` template?
- For the on-demand bits: 
 1. Run `gulp scripts:ondemand && gulp styles:ondemand`
  2. Open `/test-fixture/` in sauce labs (won't work in local VM unless localhost routing is set up).
  3. Menu should be visible and interactive and there shouldn't be any JS errors in the dev console.

## Review

- @rosskarchner 
- @sebworks 
- @virginiacc 
- @jimmynotjim 

## Notes

- All site pages should include `common.ie.js`. If you visit consumerfinance.gov/askcfpb/ and search the source code you will see it is missing `common.ie.js`.

### DIFF
--- a/cfgov/legacy/templates/front/base_update.html
+++ b/cfgov/legacy/templates/front/base_update.html
@@ -76,7 +76,7 @@ please send an email with your full name to tech@cfpb.gov.
     <link rel="stylesheet" href="/static/nemo/_/c/eot.css">
     <![endif]-->
     <link rel="stylesheet" href="/static/nemo/_/c/styles.css{{ STATIC_VERSION }}">
-   
+
     {% endblock %}
 
     {% block app_css %}
@@ -85,8 +85,17 @@ please send an email with your full name to tech@cfpb.gov.
     {% block page_css %}
     {% endblock %}
 
+    <!--[if lt IE 9]>
+    <link rel="stylesheet" href="{% static 'css/header.nonresponsive.css' %}"/>
+    <link rel="stylesheet" href="{% static 'css/footer.nonresponsive.css' %}"/>
+    <![endif]-->
+
+    <!--[if gt IE 8]><!-->
     <link rel="stylesheet" href="{% static 'css/header.css' %}"/>
     <link rel="stylesheet" href="{% static 'css/footer.css' %}"/>
+    <!--<![endif]-->
+
+
     <!-- Feeds -->
     <link rel="alternate" type="application/rss+xml" title="Consumer Financial Protection Bureau &raquo; Feed" href="/feed/" />
     <link rel="alternate" type="application/rss+xml" title="Consumer Financial Protection Bureau &raquo; Comments Feed" href="/comments/feed/" />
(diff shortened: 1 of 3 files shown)


cfgov/legacy/templates/front/base_update.html
```diff
@@ -174,25 +183,27 @@
...
 </script>
 
 <script type="text/javascript">
     // If there are no breadcrumbs on this page, remove the ul tag so that screen
     // readers don't read it (since the ul has nothing in it).
     $(function(){
-        var numBreadCrumbs = $("ul.bread").children("li").length;
-        if(numBreadCrumbs == 0){
-            $("ul.bread").remove();
+        var numBreadCrumbs = $( 'ul.bread' ).children( 'li' ).length;
```
sebworks: @anselmbradford, can this be done in css?

sebworks:
👍

cfgov/legacy/templates/front/base_update.html
anselmbradford: I don't see why not, but prob best to test the removal and I have no idea how to test this template. I just learned this was used on paying for college, askcfpb, and complaint and it would be helpful if it followed the base.html template that the rest of the site uses.

sebworks: Would be nice to have a diagram ( or documentation) on where the various `base.html` files are used.

higs4281: also used on `/retirement/before-you-claim/`

higs4281:
@willbarton @anselmbradford 
- I don't think this will stop us from deploying (our product owner might take issue with that)
- we just want to clarify this fix's path and timeline to build/production
- would be nice to have a means of testing the fix before it reached build, but we probably don't have time

willbarton:
@higs4281 Ok, I think I understand now. Thanks, and sorry for being a bit dense.

> I don't think this will stop PFC2 from deploying (our product owner might take issue with that)

Can you clarify what about this will stop PFC2 from deploying, this going in or this not going in?

> we just want to clarify this fix's path and timeline to build/production

Right now it'll deploy to build when it gets merged. From there, its normal path to production would be to deploy to beta.cf.gov as part of our next release on 8/16, and then to prod on 8/22. If you need it sooner we can apply it as a hotfix to the active release branch whenever it gets merged.

> would be nice to have a means of testing the fix before it reached build, but we probably don't have time

That I think you'll need @anselmbradford or @sebworks to help with.

higs4281:
@willbarton If it gets deployed to build on merge, I think we're fine for testing.

The issue is considered "critical" for the project; I'll get back to you on the need for hotfix

higs4281:
> Right now it'll deploy to build when it gets merged. 
@willbarton re: "when"    
is there still an scm watcher that detects merges on refresh or does a build/fab-deploy sequence need to be started by a human?

higs4281:
@willbarton deployment to prod on the current release timeline (8/22) is fine.

PR #2278, opened 2016-08-04
Removes django base.html

anselmbradford:
## Removals

- Removes django base.html

## Testing

- @rosskarchner can you test this? Thanks!

## Review

- @rosskarchner


rosskarchner:
👍

PR #2279, opened 2016-08-04
WIP: Refactor and test PDF generation code

chosak:
Work in progress: This code creates a new standalone module for PDF generation using PDFreactor, which is currently part of the leadership calendar code in `cal.views`. 

## Additions

- New `core.pdf` module that encapsulates PDF generation using PDFreactor.
- Tests for `core.pdf`.
- New management command (`./manage.py make_pdf`) to make a PDF from a webpage using the new code.

## Testing

- Run `tox` or `tox -e fast` to execute new unit tests in `core.tests.test_pdf`
- Run new management command, e.g. `./manage.py make_pdf http://www.google.com output.pdf` (this requires having the `pdfreactorwebservice` running locally).

## Review

- @rosskarchner @kurtrwall 

## Notes

- This code is not ready to be merged in but serves as an example for how we might want to do this. Potentially a separate PR would be to actually modify `cal.views` to use this new code. The integration should be fairly straightforward, with the single task to create a `base_url` from the `HttpRequest` object as is done [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cal/views.py#L161).
- It appears that at some point this PDF generation code was separate, in [`core.services`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/core/services.py#L30), although this seems defunct.

## Checklist


vccabral:
👍  great PR overall.

willbarton:
Closing this for now, as we look at possibly upgrading PDFReactor.

grapesmoker:
One thing I wanted to note about PDFreactor is that if you look at the module, the class that comes with version 6 just wraps a bunch of http requests. It maintains some internal state in the form of options, but it's not like it even holds a connection or anything, it just calls urllib. So you can instantiate the pdfreactor singleton once on startup and just use it everywhere, as long as you set whatever options you want for the generation (which you should be doing anyway).

PR #2280, opened 2016-08-04
first PR for manual

grapesmoker:
This is pretty much the same PR as #2272 but with additional front-end content and existing in a branch rather than my personal fork.


manual/docs/front-end-testing.md
```diff
@@ -0,0 +1,38 @@
+# Front-End Testing
```
willbarton: The front-end team actively maintains their guides at https://github.com/cfpb/front-end. I really think we're better off just linking to them there.

manual/docs/installation.md
```diff
@@ -0,0 +1,368 @@
+# Clone the repository
```
willbarton: What's the difference between this file and `manual/docs/installation.md` and `INSTALL.md` in the root of the repo? Can we reduce the duplication here?

manual/docs/build-and-deploy.md
```diff
@@ -0,0 +1,214 @@
...
+
+There should be no drop in test coverage for cfgov-refresh.
+
+Regular releases of cf.gov on our release cadence are automated, presuming:
+
+- All unit tests pass 
+- All functional tests pass
+- There is no reduction in test coverage from the last release tagging
+
+### Current deployment process
```
willbarton: This is one of several sections I envisioned being broken out into separate files when we were done with the working google doc for this. I can take care of that.

manual/docs/build-and-deploy.md
```diff
@@ -0,0 +1,214 @@
...
+The following teams and individuals have the necessary access to deploy cf.gov releases:
+
+- Ross Karchner
+- Scott Cranfill
+- Will Barton
+- Bill Higgins
+- Serghei Gorobet
+- The Software Delivery team
+
+### Feature flags
```
willbarton: This is one of several sections I envisioned being broken out into separate files when we were done with the working google doc for this. I can take care of that.

manual/docs/build-and-deploy.md
```diff
@@ -0,0 +1,214 @@
...
+{% endif %}
+```
+
+Feature flags are implemented as a Django app in cfgov-refresh/cfgov/flags. 
+
+#### Feature Flag Hygiene
+
+Feature flags should be rare and ephemeral. Changes should be small and frequent, and not big-bang releases, and flags that are no longer used should be cleaned up and removed from code and the database.
+
+### Wagtail pages vs. Django views
```
willbarton: This is one of several sections I envisioned being broken out into separate files when we were done with the working google doc for this. I can take care of that.

We might also want to fill it out with more useful content as we convert the jobmanager Django app to Wagtail.

willbarton:
There seems to be a lot of duplication from the INSTALL.md, README.md, TEST.md, and the front-end guides. I feel like we should probably scale back our initial work here and focus on getting the new documents we're generating for this repo and cf.gov in place first, then we can revisit whether we should move that content here.

The other thing is that there are also a lot of documents spread around various GHE repos that could possibly be brought into here first.

My other main thought is: do we need this to live in `manual/docs/`? can we just use `docs/`?

willbarton:
I broke up the content from the build & deploy document along my originally intended lines, and I had some other thoughts.

Seeing this in mkdocs, I think overall I like have the INSTALL, README, TEST, etc content in there, but I think that's something we need to discuss as a team before we do. Let's reduce the scope of change this PR is making, and see if we can use this as a starting point to replace the existing `docs/` directory. Does that sound fair?

grapesmoker:
I didn't want to touch the existing `docs` directory because I am not quite sure what it is and even though no one else seems to know either, there's a lot of work there that seems like we might not want to throw away. Hence `manual`.

With regards to duplication, that is not surprising, since it is copied verbatim from those documents. Again, the idea is to consolidate all the information about `cfgov-refresh` into one place. Otherwise, everyone has to navigate across multiple repositories any time they're trying to figure out how to do anything with the repo. Fragmented documentation is bad, in my view, so I'm trying to gather it all in one place so that anyone working on the project can immediately access any and all information about the project. I'm not sure why we shouldn't have this in one place; the text is already written so putting it into the manual is trivial. We don't need to limit ourselves to having only new content in here.

willbarton:
> I didn't want to touch the existing docs directory because I am not quite sure what it is and even though no one else seems to know either, there's a lot of work there that seems like we might not want to throw away. Hence manual.

That's fair, but I think... if no one knows what it is, it must not be very useful. I say replace it. We'll always have the git history if we need it.

> Fragmented documentation is bad, in my view, so I'm trying to gather it all in one place so that anyone working on the project can immediately access any and all information about the project. I'm not sure why we shouldn't have this in one place; the text is already written so putting it into the manual is trivial. We don't need to limit ourselves to having only new content in here.
 
Alright, then I think I would ask that with this PR, you strip down the README, remove INSTALL and TEST, and just add links to the docs in their new location. If we're relocating, let's relocate!

willbarton:
I would add, this is awesome work, and we should get it in so we can start pointing people to it!

README.md
```diff
@@ -208,240 +57,12 @@ please [file a bug report](https://github.com/cfpb/cfgov-refresh/issues/new?body
...
 We welcome your feedback and contributions.
-See the [contribution guidelines](CONTRIBUTING.md) for more details.
+See the contribution guidelines in the manual for more details.
 
 Additionally, you may want to consider
 [contributing to the Capital Framework](https://cfpb.github.io/capital-framework/contributing/),
 which is the front-end pattern library used in this project.
 
-<!-- TODO: Perhaps we want to split this out into a separate page? -->
-## Development tips
```
willbarton: Excellent that this has been separated out!

willbarton:
There's only one other little thing... do you think we could just put mkdocs.yml in the repo root, and have docs/*.md instead of docs/docs/*.md?

Other than that, I say update `master` so we can get it merged. We'll iterate on it as needed from there!

grapesmoker:
Let me check to see if that change will work. If it will.

rosskarchner:
👍  this looks great!

grapesmoker:
Took Will's suggestion to move the docs up a level.

willbarton:
:shipit:

contolini:
👍

PR #2281, opened 2016-08-05
Adding email address to info-unit molecule

sebworks:
Adding email address to info-unit molecule

## Changes

- Modified  `cfgov/jinja2/v1/_includes/molecules/contact-email.html` to use `email.text` before `email.url`.

- Modified  `cfgov/jinja2/v1/_includes/molecules/info-unit.html` to use display email address if `link.text` contains and email address.

## Testing

- Create a `Sublanding` page via Wagtail. 
- Add a `Half Width Link Blob Group` molecule. 
- Enter `email@example.com` in the link text field. Preview.
- Notice that the link has the email formatting. 

## Review

- @richaagarwal 
- @chosak 
- @virginiacc 

## Screenshots

screen shot 2016-08-05 at 9 15 14 am

## Todos

-

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/molecules/info-unit.html
+++ b/cfgov/jinja2/v1/_includes/molecules/info-unit.html
@@ -71,6 +71,12 @@
             <ul class="list list__links">
             {% for link in value.links %}
                 <li class="list_item">
+                    {% if link.text is defined and link.text.find('@') > -1 %}
+                        {% if not value.emails is defined %}
+                            {% do value.update({'emails': [link]}) %}
+                        {% endif %}
+                        {% include 'contact-email.html' with context %}
+                    {% else %}
                     <a class="{{ 'btn m-info-unit_btn' if value.is_button
                                   else 'list_link' }}"
                        href="{{ link.url }}">
@@ -78,6 +84,7 @@
                                  after testing is completed #}
                         {{ link.text if link.text else 'Learn More' }}
                     </a>
+                    {% endif %}
                 </li>
             {% endfor %}
             </ul>
--- a/cfgov/jinja2/v1/_includes/molecules/info-unit.html
+++ b/cfgov/jinja2/v1/_includes/molecules/info-unit.html
@@ -73,7 +73,7 @@
                 <li class="list_item">
                     {% if link.text is defined and link.text.find('@') > -1 %}
                         {% if not value.emails is defined %}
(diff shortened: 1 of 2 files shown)


cfgov/jinja2/v1/_includes/molecules/info-unit.html
```diff
@@ -71,13 +71,20 @@
             <ul class="list list__links">
             {% for link in value.links %}
                 <li class="list_item">
+                    {% if link.text is defined and link.text.find('@') > -1 %}
```
sebworks: I'm not sure if this is the best way to determine if a url is an email address.

chosak:
In order to get this to work properly, I had to set an email address (like `user@example.com`) for both the link Text and Url fields. This generates the "EMAIL" header and a correct link like:

```html
<a href="mailto:user@example.com">user@example.com</a>
```

If I set the text to `user@example.com` but leave the URL as the default `/`, it produces the correct visual markup, but the link is wrong:

```html
<a href="mailto:/">user@example.com</a>
```

cfgov/jinja2/v1/_includes/molecules/info-unit.html
chosak: [Here's](http://emailregex.com/) a somewhat hardcore regex for matching email addresses.

sebworks: Yeah, we can use Djangos [validate_email](https://github.com/django/django/blob/12b4280444b58c94197255655e284e4103fe00a9/django/core/validators.py#L239) method but I'm sure where that code would need to reside. Would it be on the the Hyperlink molecule? A context function?

sebworks:
>If I set the text to user@example.com but leave the URL as the default /, it produces the correct visual markup, but the link is wrong:

Fixed in https://github.com/cfpb/cfgov-refresh/pull/2281/commits/5c9d8bf48eb34e3d10488d6283af85cd38c40c80. I've reverted the change to `contact-email.html` and instead just assigned `link.url` the value of `link.text`. Currently, only the `link.text` is required by [default](https://github.com/cfpb/cfgov-refresh/blob/e0e0f874adb344f5afdd47a6404a7acdb5004478/cfgov/v1/atomic_elements/atoms.py#L56).

chosak:
Thanks. Besides the email address check (which I'm not sure where it should live), this looks good to me. 👍

PR #2282, opened 2016-08-05
Adding third width link blob group

sebworks:
Adding third width link blob group, which is necessary for the http://www.consumerfinance.gov/about-us/careers/working-at-cfpb/ page.

## Additions

- Added `cfgov/jinja2/v1/_includes/organisms/third-width-link-blob-group.html` to handle three column layouts.
- Added `cfgov/jinja2/v1/_includes/organisms/info-unit-group.html` to serve as a generic blog group builder.

## Changes

- Modified `cfgov/v1/atomic_elements/molecules.py` to use the `link-blob` template.
- Modified `cfgov/unprocessed/css/organisms/info-unit-group.less` to reflect the add three column layout. 
- Changed the name of `cfgov/_includes/molecules/half-link-blob.html` to  `half-width-link-blob-group.html`
- Modified `cfgov/jinja2/v1/_includes/organisms/half-width-link-blob-group.html` to call `info-unit-group.html` macro.

## Testing

~~~Currently unable to test the third width link blob group. Requires changes in Wagtail.~~~ Instructions for the third-width link blob group are below:

- Create a `Browse` page via Wagtail. 
- Add a `Third Width Link Blob Group` molecule. 
- Add 3 half-width link blobs. Preview.
- Notice that the organism renders three link blobs in separate columns.
- Return to Wagtail
- Check the `Has top border` and `Has bottom border` checkboxes, preview again.
- Notice that the organism renders as before but with borders.

## Review

- @chosak 
- @richaagarwal 
- @virginiacc 
- @jimmynotjim 


## Checklist


cfgov/v1/atomic_elements/molecules.py
```diff
@@ -18,7 +18,7 @@ class HalfWidthLinkBlob(blocks.StructBlock):
 
     class Meta:
         icon = 'link'
-        template = '_includes/molecules/half-width-link-blob.html'
+        template = '_includes/molecules/link-blob.html'
```
Scotchester: The class here should be renamed to just `LinkBlob` right?

Scotchester:
Looks like we need some further Wagtail model changes to take this over the finish line. Would you like some assistance with that, @sebworks ?

sebworks:
yes, that would be wonderful @Scotchester .

schaferjh:
Which page templates would this be available on? Just sublanding?

And does this mean that anytime I add three `half width link blobs` they show up in 1/3 columns on the same row? Or can I still get them to show up in two rows -- two on top, one on the bottom?

Scotchester:
Well, @schaferjh, that gets somewhat to how it's implemented in Wagtail. I was assuming (though it could be done differently) that the Wagtail user would choose between half-width and third-width layouts when creating their module, and whatever the number of items inside, it'd stick to that.

schaferjh:
Sure it's an implementation thing, I'd just like us to think about it now while we're making it. :) 

I will think on it a bit a comment later.

Scotchester:
Sounds good :)

sebworks:
> Which page templates would this be available on?

Browse basic and home pages. These are some places where the third-width columns are being used:

http://www.consumerfinance.gov/about-us/the-bureau/about-deputy-director/

http://www.consumerfinance.gov/

http://www.consumerfinance.gov/about-us/careers/application-process/.

> And does this mean that anytime I add three half width link blobs they show up in 1/3 columns on the same row? Or can I still get them to show up in two rows -- two on top, one on the bottom?

We are really just talking about containers controlling the widths of their children (i.e, organisms controlling the widths of molecules). Without placing the `link-blob` molecule in a organism then it should be full-width. There really shouldn't be such thing as a `half-width link-blob` molecule.  It should be full width until placed in a half-width or third-width organism.

schaferjh:
> There really shouldn't be such thing as a half-width link-blob molecule. It should be full width until placed in a half-width or third-width organism.

Ok, but I think you're speaking from a code perspective, and I was speaking from a content manager perspective. 😄  As long as when we implement this in Wagtail, we make it clear how to make one choice or the other, it's cool.

> > Which page templates would this be available on?

> Browse basic and home pages.

That sounds good for covering our bases for current use, but I think we should add this to Sublanding pages as well. 

Ok, bugging you about implementation is over. Once this is merged, please help me remember to make the implementation task!

Scotchester:
@schaferjh We should include the back-end implementation in this PR, in my opinion. It should be pretty straightforward.

> As long as when we implement this in Wagtail, we make it clear how to make one choice or the other, it's cool.

So what do we think is the most straightforward way to do this for the Wagtail user? I see two obvious options:

1. Create a new "Third width link blob group" module that mirrors the existing "Half width link blob group" module and simply outputs the three-column layout.
2. Rename the existing "Half width link blob group" module to just "Link blob group" and add a new option inside of it to toggle between two-col and three-col layouts.

schaferjh:
Actually, first a point of clarification -- is there a role for images @sebworks? The home page implementation you point to includes images, so just trying to figure out if you were looking to include them here as well.

The reason the question comes up is that we've implemented blobs with images and blobs without as two seperate implementations in Wagtail. So trying to decipher if this is a moment to either update both of those implementations or think about a way to merge a couple of things....

Scotchester:
@schaferjh Huh. You're right, and I forgot about that...

It would seem to me that they could both be part of the same module. The individual unit in a 50-50 Image and Text module is just a Half Width Link Blob with an image, no?

schaferjh:
Essentially, yes. Addressing/streamlining that is on our backlog, but not super high priority. Just trying to figure out how this new work might play into those modules.

sebworks:
> Actually, first a point of clarification -- is there a role for images @sebworks? The home page implementation you point to includes images, so just trying to figure out if you were looking to include them here as well.

Yes, because the `half-width-link blob`, `25-75`, and columns on the hom  all use the `m-info-unit` molecule. We currently don't provide a way to add images to the  `half-width-link-blob` but the capability exists in the m-info-unit.

schaferjh:
@sebworks and is the `m-info-unit` also the basis of the `50-50 Image and Text` organism?

sebworks:
Yep, here is the list https://github.com/cfpb/cfgov-refresh/search?utf8=%E2%9C%93&q=info_unit

jimmynotjim:
Thought, should we figure out how to remove the layouts from the three info-unit modules and just use info-unit on the backend, inside a 1/2 or 1/3 layout organism? It seems we crossing concerns by continuing to use the 50/50 image & text, 25/75 image & text, and half width link blob.

jimmynotjim:
Continued thought, the 25/75 is a bit more complicated, but if we converted full-width-text/media to a full-width organism, an info-unit inside could use the 25/75 style layout. With a half-width and third-width organism the organism would control all the layout and just default to the vertically stacked info-unit.

cfgov/v1/atomic_elements/molecules.py
sebworks: I'm going to leave it as it exist for now because it might cause some confusion. We still have the concept of the `half-width` link blob in Wagtail.

sebworks:
I've added the `third-width-link-blob-group` to Wagtail.  I've also added the ability to add borders to the  `third-width-link-blob-group` and `half-width-link-blob-group` components. These components are available on the `sublanding`, `landing`, and `browse` pages.


screen shot 2016-08-23 at 6 49 31 am
screen shot 2016-08-23 at 6 47 32 am

richaagarwal:
This looks really clean. Thanks @sebworks! 👍

richaagarwal:
Actually, I just noticed one thing - the new third width link blob group is the same model definition as the half width, they just have different templates. Is there a way to avoid that duplication?

richaagarwal:
👍

PR #2285, opened 2016-08-05
pinned base requirements and added requirements for the manual separa…

grapesmoker:
This PR modifies `requirements/base.txt` and adds the file `requirements/manual.txt` to pin the `pip` dependencies of the project to their current versions. `manual.txt` is separated from the base requirements so that you have the option, should you choose to take it, of installing only the things you need to run the project and not the things you need to also display the manual.

## Additions
`requirements/manual.txt`

## Removals
Removed a few dependencies that did not seem to actually be used anywhere. 

## Changes
In addition to the above, reordered the requirements files to be in alphabetical order. Makes it easier to immediately see differences.

## Testing
`pip install -r requirements/base.txt` works.

## Review

@willbarton @richaagarwal @rosskarchner @chosak 

## Checklist


requirements/base.txt
```diff
@@ -1,31 +1,45 @@
...
+git+https://github.com/rosskarchner/govdelivery.git#egg=govdelivery
+html5lib==0.9999999
+icalendar==3.9.0
 lxml==3.6.0
 pbr==1.9.1
-django-compressor==2.0
\ No newline at end of file
+py==1.4.31
+python-dateutil==2.1
+pytz
```
willbarton: Can we pin pytz?

requirements/test.txt
```diff
@@ -1,4 +1,3 @@
 -r base.txt
 mock==1.3.0
-tox==2.1.1
```
willbarton: Do we need tox in the base requirement and not test?

rosskarchner: I don't think so?

chosak: You only need tox if you're actually running the tests. Travis [uses](https://github.com/cfpb/cfgov-refresh/blob/master/.travis.yml#L34) `cfgov.settings.test` when running them.

requirements/base.txt
grapesmoker: I tried. It seems that pytz has this weird versioning that uses a YYYY.MM.DD format. Pip does not understand how to resolve this and therefore complains about it. I'm not sure what the right solution here is; it was previously left unpinned for, presumably, the same reason. Open to suggestions as to what to do.

rosskarchner: I think a requirements file can point right to the tarball or wheel?  try replacing the pytz line with the URL:

https://pypi.python.org/packages/ba/c7/3d54cad4fb6cf7bf375d39771e67680ec779a541c68459210fcfdc3ba952/pytz-2016.6.1-py2.py3-none-any.whl#md5=735126cd1c1ddc824b7c94e091d7173e

willbarton:
@grapesmoker can you move tox back to the test settings? Then I think we can get this merged.

rosskarchner:
👍 after the change @willbarton mentions above

grapesmoker:
So, I tried to add the tox requirement back into test, and this breaks Travis because apparently duplicating requirements Is Bad:

```Double requirement given: tox==2.3.1 (from -r ./requirements/test.txt (line 4)) (already in tox==2.3.1 (from -r ./requirements/base.txt (line 41)), name='tox')```

@willbarton @rosskarchner please advise

rosskarchner:
Does anyone see a reason for it to be in base.txt? I feel like it can be removed there

willbarton:
Yeah, we don't need tox in both places, and there's no reason for it to be in the base requirements.txt that I can see.

willbarton:
:shipit:

rosskarchner:
👍

PR #2286, opened 2016-08-05
Allow MEDIA_ROOT to be set by environment

willbarton:
This PR allows the MEDIA_ROOT setting to (optionally) be set by an environment variable. This, along with a change to our dev server environment, will avoid our MEDIA_ROOT being recreated with each deployment to our dev server (and therefore all media being flushed).

## Changes

- Allows MEDIA_ROOT to be set by an environment variable, if it exists.

## Testing

- `export MEDIA_ROOT=/tmp/media`
- Run the server
- In the Wagtail admin, upload an image.
- It should be in `/tmp/media/original_images` instead of `cfgov/f/original_images`

## Review

- @richaagarwal 
- @chosak 
- @rosskarchner


richaagarwal:
👍

cfgov/cfgov/settings/base.py
```diff
@@ -177,7 +177,8 @@
 # Absolute path to the directory static files should be collected to.
 STATIC_ROOT = os.environ.get('DJANGO_STATIC_ROOT', '/var/www/html/static')
 
-MEDIA_ROOT = os.path.join(PROJECT_ROOT, 'f')
+MEDIA_ROOT = os.environ.get('MEDIA_ROOT', 
+                            os.path.join(PROJECT_ROOT, 'f'))
```
anselmbradford: I really feel like defaults for environment variables should be centralized where the environment variables are set.

contolini: Where would that happen? In the `.env` file?

richaagarwal: @contolini @anselmbradford We can't set defaults in a `.env` file but I think what Ans is getting at is that it should be really clear what our environment variables are at any given point by examining the environment `os.environ`. So rather than fall back on a value if the environment is not set, it would be better if we _set_ the environment values with the defaults. It could be done in this file.

willbarton: > So rather than fall back on a value if the environment variable is not set, it would be better if we set the environment variables with the default values. It could be done in this file.

We can do it that way if that's the consensus, but I'm very much not sold on the benefit of *not* having sane defaults in the actual configuration if the environment doesn't happen to contain the variable.

On top of this, by defaulting to `PROJECT_ROOT/f`, this is consistent with the previous behavior.

richaagarwal: @willbarton You would explicitly set the default value, so "if the environment doesn't happen to contain the variable" doesn't seem relevant.

richaagarwal: That said, I don't think this should block this particular PR. We should have this conversation though.

willbarton: @richaagarwal Ah, sorry, I misunderstood, you're talking about something different from @anselmbradford. Set the default in `os.environ` at the top of the base settings file?

So is that something that should be done for all our environment variables then, and so a separate piece of work, or is that a blocker for this PR?

anselmbradford: > Where would that happen? In the .env file?

That's how I set it in [this experiment](
https://github.com/cfpb/cfgov-refresh/pull/2244/commits/59a00604e4a00c32d4200e8ccd3e581c9eb10b37#diff-eb951766c3e0a8d12740b6236ded5a9fR145). If there's a more elegant way to set the defaults that may make sense to do it there. It seems counter to what environment variables are—essentially constants—to reference them but when they're not set have (potentially different) fallback values set throughout the codebase. The original reference should be what gets the fallback value at the time it is set.

willbarton: > That's how I set it in this experiment.

That fallback won't work if the .env file isn't sourced though.

> It seems counter to what environment variables are—essentially constants—to reference them but when they're not set have (potentially different) fallback values set throughout the codebase.

The Django settings file isn't exactly throughout the codebase, in this case (I can't speak for other env vars). If I were coming to the project, that would be the first place I'd be looking for default values, not a hidden file.

anselmbradford: > That fallback won't work if the .env file isn't sourced though.

Should the app work at all if `.env` isn't sourced? Seems like sourcing that is a requirement

> The Django settings file isn't exactly throughout the codebase, in this case (I can't speak for other env vars). If I were coming to the project that would be the first place I'd be looking for default values (indeed, that's where I have been), not a hidden file.

Makes sense, as mentioned in chat my concern was just that Python/JS could access the same envvars, so if defaults are set at a layer before that happens for each it would be slightly more robust.

richaagarwal:
I vote for merging this in ASAP since some people rely on images in build.  I can explain more about what I mean with setting env variables in a BEWD huddle some time, or another discussion if appropriate.

willbarton:
@richaagarwal I think @rosskarchner's [drama-free-django fix](https://github.com/cfpb/drama-free-django/pull/2) makes build better. But I agree.

willbarton:
:shipit:

PR #2294, opened 2016-08-10
Alter Job table to allow case-sensitive uniqueness

chosak:
Legacy job listing data in the `jobmanager.models.Job` table includes listings that have slugs that are the same term but different case, e.g. `managing-counsel` and `Managing-counsel`. This PR adjusts the way this table is created so that it treats these two as different when checked against a uniqueness constraint.

The Django `Job` model field `slug` has a uniqueness constraint ([`unique=True`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jobmanager/models.py#L63)), but default MySQL character encoding and collation treats `varchar`s as the same [even with different case](http://dev.mysql.com/doc/refman/5.7/en/case-sensitivity.html). This means that a standard MySQL DB (such as one run on a developer machine) cannot load the job listing data currently in the production database.

The production database has had its schema manually changed in a way that deviates from what the Django ORM creates; this PR attempts to bring the Django ORM back in compliance with what actually exists on the production DB.

## Additions

- New migration that sets `utf8` character set and `utf8_bin` collation on the `jobmanager_job` table (for model `jobmanager.models.Job`).
- New tests for `jobmanager.models.Job` that verify that the slug `UNIQUE` constraint is not violated by `Job` models with slugs that differ only by case.

## Changes

- Modified `Job.save` so that it uses `django.utils.timezone.now` instead of `datetime.now` to avoid [naive DateTimes](https://docs.djangoproject.com/en/1.8/topics/i18n/timezones/#naive-and-aware-datetime-objects) (which avoids `RuntimeWarning` messages when creating/saving these models in unit tests, and is generally good practice).

## Testing

- Run `tox` to verify that all Python unit tests pass.
- Follow these steps with a new or existing database to verify table collation:

```sh
 $ python cfgov/manage.py dbshell
 mysql> show create table jobmanager_job;
 ```

You should see output including `DEFAULT CHARSET=utf8 COLLATE=utf8_bin`.
- Follow these steps with a new database to verify that a production DB fixture (generated by `manage.py dumpdata jobmanager`) can be loaded:
 
```sh
 $ ./drop-db.sh
 $ ./create-mysql-db.sh
 $ python cfgov/manage.py 
 $ python cfgov/manage.py loaddata jobmanager.json
 ```

The fixture load should complete successfully and you should see output like `Installed 1000 object(s) from 1 fixture(s)`.

## Review

- @cfpb/cfgov-backends, especially @rosskarchner

## Todos

- There are likely other inconsistencies that should be identified between existing database tables and their configuration in code.

## Checklist

### CODE REVIEW

File: cfgov/jobmanager/migrations/0002_auto_20160809_1619.py [hunk 1]
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class MySQLOnlyRunSQL(migrations.RunSQL):
```
chosak: Unfortunately Django doesn't provide a way to disable migrations per-backend. The character set and collation alterations that need to happen to this table aren't supported properly in sqlite, so this migration needs to be skipped when running the tests.

cfgov/jobmanager/models.py
```diff
@@ -102,8 +103,8 @@ def get_absolute_url(self):
 
     def save(self):
         if self.date_created == None:
-            self.date_created = datetime.now()
-        self.date_modified = datetime.now()
+            self.date_created = timezone.now()
+        self.date_modified = timezone.now()
```
chosak: This doesn't significantly change the behavior of this field for legacy data (or going forward until this is moved to Wagtail) but eliminates generation of `RuntimeWarning` exceptions that complain about use of naive `DateTime`.

cfgov/jobmanager/models.py
```diff
@@ -102,8 +103,8 @@ def get_absolute_url(self):
 
     def save(self):
         if self.date_created == None:
```
vickumar1981: use is None.

cfgov/jobmanager/tests/test_models.py
```diff
@@ -0,0 +1,16 @@
...
+
+class JobTestCase(TestCase):
+    def test_slug_uniqueness(self):
+        mommy.make(Job, slug='abcde')
+        with self.assertRaises(IntegrityError):
+            mommy.make(Job, slug='abcde')
+
+    def test_slug_case_sensitive_uniqueness(self):
+        mommy.make(Job, slug='abcde')
+        mommy.make(Job, slug='ABCDE')
```
vickumar1981: shouldn't this test assert something?

cfgov/jobmanager/models.py
chosak: Yep, I agree there are some improvements that could be made to the existing job code. I tried to keep the scope of this PR limited to the particular issue of altering the table.

cfgov/jobmanager/tests/test_models.py
chosak: It's verifying that you can create two `Job` objects with slugs that differ only by case. The test only passes if these two don't violate the uniqueness constraint (which means that differing cases are unique). Without that (default MySQL behavior, without this new migration) the second `make` call will raise an `IntegrityError`.

vickumar1981: perhaps then the test should test that an error is raised if uniqueness gets violated, i.e. assert_integrity_error_for_uniqueness_violation, and do an assertRaises(IntegrityError).

We want to ensure that this error is being raised.  That's the test, IMO.

chosak: That's what the test above (`test_slug_uniqueness`) does.

cfgov/jobmanager/models.py
vickumar1981: generally, i follow this rule: http://programmer.97things.oreilly.com/wiki/index.php/The_Boy_Scout_Rule

if i touch a file, and there's an improvement to be made, i want to leave the file in a better state than when i began making changes.

chosak: Fair point @vickumar1981! Although it's a grey area between picking up trash from the campground and mowing, weeding, planting some flowers... I've made the improvement you suggested and even added a couple of new tests.

willbarton: @vickumar1981 I think some context here, too is worth noting: this is a piece of larger work @chosak is doing to move the jobmanager models to Wagtail entirely.

cfgov/jobmanager/tests/test_models.py
vickumar1981: yeah, i see.  so a way to make that more clear might be:

```
    _integrity_error = False
    try:
        mommy.make(Job, slug='abcde')
        mommy.make(Job, slug='ABCDE')
    except IntegretyError:
        _integrity_error = True
    self.assertFalse(_integrity_error, 'Integrity error raised.')
```

chosak: Thanks again for the feedback, modified in ac1b9ed to make it a bit clearer as you suggested.

vickumar1981:
👍

rosskarchner:
:shipit:

PR #2295, opened 2016-08-10
Introducing TableBlocks!

richaagarwal:
Replaces our old `Table` defined in `organisms.py` with Wagtail's new `TableBlock`, including a data migration.

Note that the old field has not yet been removed (this will have to come later),
& the old field is still displayed in the admin panel (still working on this/ is a TODO). Feedback welcome in the meantime.

Other TODOs:
- I will update the `CHANGELOG` when this is ready for merging.

Before:  
screen shot 2016-08-12 at 9 16 02 am

After:  
screen shot 2016-08-11 at 2 21 01 pm

Assignees: @willbarton @rosskarchner @grapesmoker @chosak


richaagarwal:
After more poking, I still can't seem to get the old Table hidden in the admin panel.  Wagtail's documentation says setting `editable=False` should work (see http://docs.wagtail.io/en/v1.4.5/reference/pages/panels.html#hiding-fields and https://github.com/torchbox/wagtail/issues/2712#issuecomment-226152515). Could a @cfpb/cfgov-backends take a look?

rosskarchner:
@richaagarwal Are there consequences to simply removing lines like [this](https://github.com/cfpb/cfgov-refresh/blob/table-block/cfgov/v1/models/learn_page.py#L91)  (it worked in my quick test)?

richaagarwal:
@rosskarchner yup, that's not an option if we want to be able to run the data migration

rosskarchner:
I'm a terrible person, but this does it.

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -5,6 +5,7 @@
 from . import atoms, molecules
 from ..util import ref
 from ..models.snippets import Contact as ContactSnippetClass
+from wagtail.contrib.table_block.blocks import TableBlock
```
chosak: Nitpicky, but can third-party imports go in a separate group before our own imports? See the second bullet in PEP8 [here](https://www.python.org/dev/peps/pep-0008/#imports).

cfgov/v1/models/browse_page.py
```diff
@@ -11,6 +11,7 @@
 from .base import CFGOVPage
 from ..atomic_elements import molecules, organisms
 from ..util.util import get_secondary_nav_items
+from wagtail.contrib.table_block.blocks import TableBlock
```
chosak: Same comment about import order from above.

cfgov/v1/models/learn_page.py
```diff
@@ -13,6 +13,8 @@
 
 from ..atomic_elements import molecules, organisms
 from .base import CFGOVPage, CFGOVPageManager
+from wagtail.contrib.table_block.blocks import TableBlock
```
chosak: Same comment about import order from above.

cfgov/v1/models/sublanding_page.py
```diff
@@ -9,6 +9,8 @@
 from .base import CFGOVPage
 from ..atomic_elements import molecules, organisms
 from ..util import filterable_list, util
+from wagtail.contrib.table_block.blocks import TableBlock
```
chosak: Same comment about import order from above.

chosak:
This looks good to me. One minor thing I noticed, not sure if this should be a blocker or not. The use of `|safe` [here](https://github.com/cfpb/cfgov-refresh/pull/2295/files#diff-34ca73bf8aa2a2e8fa1a80a1d6e4e077R3) is a bit risky. For example, if you enter a `<script>` tag in the table like this:
image
if you preview or view the published page, you see this:
image
Is there a convention that we use around input in other atomic elements? @sebworks

sebworks:
> Is there a convention that we use around input in other atomic elements? @sebworks

I don't think it's an issue because the content editors should never be entering scripts.

chosak:
@sebworks agreed. Just wanted to point this out as part of general sanity-checking of inputs. Not sure how/if Wagtail handles this for other widgets.

cfgov/v1/atomic_elements/organisms.py
richaagarwal: For sure, thanks for pointing that out!

richaagarwal:
@chosak @sebworks Yeah, without it, the html within the table doesn't render. @rosskarchner was suggesting this may be a Wagtail bug, since it seems like we shouldn't have to specify this.

schaferjh:
@richaagarwal yes, I will!

schaferjh:
@richaagarwal I don't see the new tables on build? Just the old tables.

schaferjh:
@richaagarwal Found it, used it, pretty darn cool. 👍  

When this deploys to production, we'll have to make sure Sarah and Risa have a heads up since it's such a different interaction.

PR #2296, opened 2016-08-10
Converting Jobs Description template to Wagtail

sebworks:
Converting Jobs Description template to Wagtail. This is based on the jobs UI from Django Admin and the Event template.

## Additions

- Modified the existing [template](https://github.com/cfpb/cfgov-refresh/blob/e0e0f874adb344f5afdd47a6404a7acdb5004478/cfgov/jinja2/v1/about-us/careers/_single.html) to allow usage within Wagtail.

## Testing

- No great way to test until implemented in Wagtail.

## Review

- @richaagarwal 
- @chosak 

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/job-description-page/index.html
+++ b/cfgov/jinja2/v1/job-description-page/index.html
@@ -1,0 +1,144 @@
+{% extends 'layout-2-1-bleedbar.html' %}
+
+{% import 'templates/streamfield-sidefoot.html' as streamfield_sidefoot with context %}
+
+{% block content_main_modifiers -%}
+    {{ super() }} content__flush-bottom
+{%- endblock %}
+
+{% block content_main %}
+
+{% import 'molecules/social-media.html' as social_media with context %}
+    <section class="block
+                    block__flush-top
+                    block__sub">
+        <h1>{{ page.job_title }}</h1>
+        <dl>
+            <dt>Division/Office: </dt>
+            <dd>{{ page.job_category }}</dd>
+            <dt>Expiration Date:</dt>
+            <dd>
+                {% if page.job_close_date %}
+                    {{ time.render(page.job_close_date, {'date':true}) }}
+                {% endif %}
+            </dd>
+            <dt>Region:</dt>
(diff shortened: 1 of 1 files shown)


chosak:
@sebworks per recent data model discussion, we need to support jobs that don't have a USAJobs link, but may instead have some alternate instructions, likely including an email address to send to. I know that the backend models are still in progress, but can you modify the template [here](https://github.com/cfpb/cfgov-refresh/pull/2296/files#diff-c574ab9aa2258b4c8c4dff6879ddb04aR116) so that it supports this alternate application method?

Let's assume that jobs will have either a USAJobs link OR an email address, and, in both cases, there MIGHT be some additional special instruction text to display before the link.

sebworks:
>Let's assume that jobs will have either a USAJobs link OR an email address, and, in both cases, there MIGHT be some additional special instruction text to display before the link.

Fixed in https://github.com/cfpb/cfgov-refresh/pull/2296/commits/cb67b689ab17bbb4c81a9eaad86a597392a18518

chosak:
This looks good to me as a start, although it will probably need a bit more iteration. I'm at the point where this would be useful to have to test the backend pieces, so I'd vote to merge this in now because nothing else is dependent on it at this point. 👍 

One minor quibble that I ran into when testing with this locally. I was getting errors due to the format specifier `{:,d}` used [here](https://github.com/cfpb/cfgov-refresh/pull/2296/files#diff-c574ab9aa2258b4c8c4dff6879ddb04aR30). Could we use '{:,.2f}' instead?

```py
>>> '{:,d}'.format(Decimal('100000.00'))
ValueError: Invalid format specifier: ,d
>>> '{:,.2f}'.format(Decimal('100000'))
'100,000.00'
```

sebworks:
>  I was getting errors due to the format specifier {:,d} used here. Could we use '{:,.2f}' instead?

Fixed in https://github.com/cfpb/cfgov-refresh/pull/2296/commits/8aef3619ea41ec1e4b25b6187af5d1a9aefc4b61

PR #2297, opened 2016-08-10
Welcome page links on build server

hillaryj:
Adds helpful links to the demo Wagtail.

## Additions

Adds links to Wagtail How-to and CFPB Design Manual.

## Review

- @kurtrwall 

## Checklist


kurtrwall:
👍

cfgov/templates/welcome.html
```diff
@@ -3,14 +3,42 @@
 
 {% block furniture %}
     <div class="content-wrapper">
+
         <h1>{% trans 'Welcome' %} <br/>
-        You've logged in to the consumerfinance.gov editorial tools.	
+        {% if 'build' in request.site.hostname %}
```
chosak: This check won't work if you e.g. access build from its IP address or via a specific server.

cfgov/templates/welcome.html
```diff
@@ -3,14 +3,42 @@
...
         <ul>
-        {% for name, url in destinations %}
-            <li><a href="{{url}}"> {{name}}</a></li>
-        {% endfor %}
+            {% for name, url in destinations %}
+                <li><a href="{{url}}"> {{name}}</a></li>
+            {% endfor %}
         </ul>
+
+        {% if 'build' in request.site.hostname %}
```
chosak: Is there any reason why we wouldn't want these helpful links to be in prod Wagtail if they're useful on build?

schaferjh: Not necessarily. Just something we thought of here since build-Wagtail was thought of as the place for iterating/learning about Wagtail.

cfgov/templates/welcome.html
kurtrwall: Did you try using http://127.0.0.1/login/ ? It works for me. The `request.site` is a Wagtail site object. The site object's hostname is what is being checked against, it doesn't use the hostname from the request's url.

chosak: Thanks, @kurtrwall, my misunderstanding. I withdraw the comment!

hillaryj: Is there a reliable way to check whether it's the build server otherwise?

kurtrwall: Good question. Possibly an env variable?

rosskarchner:
👍

cfgov/templates/welcome.html
willbarton: I kind of agree with @rosskarchner on L22. Why not just expose this everywhere, and then we don't have to worry about special-casing build (especially as we've been exploring options other than build for the Wagtail playground)?

cfgov/templates/welcome.html
willbarton: I kind of agree with @rosskarchner. Why not just expose this everywhere, and then we don't have to worry about special-casing build (especially as we've been exploring options other than build for the Wagtail playground)?

cfgov/templates/welcome.html
willbarton: Deleted my previous comment here... @schaferjh is there a reason we can't have a generally useful sentence here that doesn't require a check to see if we're using build or we're not using build?

cfgov/templates/welcome.html
```diff
@@ -3,14 +3,42 @@
...
+            <h2>Helpful links</h2>
+            <ul>
+                <li>
+                    <a href="https://github.cfpb.gov/pages/flapjack/wagtail-documentation/">
```
willbarton: Other thought, can the wagtail documentation be made public, so we don't have our GHE URL in public code here?

cfgov/templates/welcome.html
schaferjh: Not sure I understand correctly. I am actually purposefully asking that the build version of Wagtail have different welcome language to reinforce that it's not actually connected to the live site. Just like the change to the log in screen that says "Log in to demo Wagtail" instead of "Log in to consumerfinance.gov".

willbarton: Ok. Would it make more sense to special-case production than build?

Scotchester: @willbarton What difference does it make? It's a single-level if-else statement, either way.

cfgov/templates/welcome.html
Scotchester: Agreed, there is no reason to make this conditional to the build server.

cfgov/templates/welcome.html
willbarton: @Scotchester but if we special-case prod, then we get the welcome message that this isn't production everywhere *except* prod, and we get the helpful documentation links below everywhere *except* prod.

This would prevent us from having to special-case other hostnames in the future if we standup development servers that are not called "build".

Scotchester: 1. As noted below, I don't see a point in restricting the helpful links below to any particular server.
2. If we're taking servers other than Build and Content into consideration, then we probably need to reevaluate this whole thing. "Demo Wagtail" is meant to specifically refer to the playground on Build.

willbarton: @Scotchester It is intended to refer to the playground, yes, but the platform team hasn't decided if build is the definitive place for that, or if we want to standup a separate server. See GHE/CFGOV/platform/issues/164.

willbarton: Taking a step back though, if the use-case is to be able to have different welcome language that's server-dependent, it might make sense to have that as a configurable string.

Scotchester: Good points.

cfgov/templates/welcome.html
schaferjh: Um, no I am not comfortable making the Wagtail documentation public at this moment given the purpose and content.

It's an interesting idea in the long run but not something we can just do right now. I don't know how to get around the GHE link. Perhaps we can't include that link here if it means putting GHE in the public.

schaferjh: @willbarton is the GHE link in public code a blocker?

schaferjh: @hillaryj While Adam is digging into the GHE linking issue, for this PR, please remove the link here, replace with a line of text and keep the design manual links. In total:

Helpful links:
- Design manual - Global elements
- Design Manual - Page components
- CFPB-specific Wagtail documentation (please see link in CF.gov Platform team repo under "playground")

I realize this isn't optimal, but it will help us move this PR through for now. Thanks all for the discussion!


Design Manual - Global elements: https://cfpb.github.io/design-manual/global-elements/global-elements.html
Design Manual - Page components: https://cfpb.github.io/design-manual/page-components/page-components.html

cfgov/templates/welcome.html
hillaryj: So is the consensus to leave the Demo Wagtail for now and intend to update for different servers in the future?

willbarton: I think it's fine for now. Let's get this merged!

willbarton:
:shipit:

PR #2298, opened 2016-08-10
Convert jobmanager.models.ApplicantType to Wagtail Snippet

chosak:
As part of converting the `jobmanager` app to Wagtail, this PR creates a new `ApplicantTypeSnippet` as an equivalent replacement for `jobmanager.models.ApplicantType`.

## Additions

- New `jobmanager.models.ApplicantTypeSnippet` model serves as a new Wagtail [Snippet](http://docs.wagtail.io/en/v1.5.3/topics/snippets.html) for managing job applicant types.
- New management command `convert_jobs_to_wagtail` to convert existing Django `ApplicantType` instances into new Wagtail `ApplicantTypeSnippet` instances.

 To convert instances without persisting to the database:
 ```sh
 $ python cfgov/manage.py convert_jobs_to_wagtail --dry-run
 converted 44 ApplicantType(s)
 ```
 To convert instances and save to the database:
 ```sh
 $ python cfgov/manage.py convert_jobs_to_wagtail
 converted 44 ApplicantType(s)
 saved 44 ApplicantTypeSnippet(s)
 ```

## Changes

- `jobmanager.models` changed from a single file to a directory with multiple modules, including `jobmanager.models.django` for existing Django models.

## Testing

- Run `tox` to confirm that all existing unit tests pass.
- With a new or existing database:

 - Run migrations and start your development server:

   ```sh
   $ ./runserver.sh
   ```

 - Navigate to [http://localhost:8000/admin/snippets/](http://localhost:8000/admin/snippets/), select new "Job applicant type" snippet.
 - Select "Add job applicant type" and define a new model instance.
 - Hit "Save" to persist your new applicant type.
- With an existing `ApplicantType` fixture file (generated by running `manage.py dumpdata jobmanager.ApplicantType`):

 - Run migration step above.
 - Load fixture:

  ```sh
   $ python cfgov/manage.py loaddata applicant_type.json
   ```
   
 - Convert fixture data to snippets:

  ```sh
  $ python cfgov/manage.py convert_jobs_to_wagtail
  ```

 - Navigate to [http://localhost:8000/admin/snippets/jobmanager/applicanttypesnippet/](http://localhost:8000/admin/snippets/jobmanager/applicanttypesnippet/) to see list of available applicant types.

## Review

- @cfpb/cfgov-backends @sebworks @schaferjh 

#### Top-level Snippet page:
image

#### List of available applicant types:
image

#### Editing an existing applicant type:
image

## Notes

- This new snippet maintains all of the existing fields and constraints on `ApplicantType`. Notably, this does not include any sort of `unique=True` constraint on the model name or slug.

## Checklist

### DIFF
--- a/cfgov/jobmanager/management/commands/convert_jobs_to_wagtail.py
+++ b/cfgov/jobmanager/management/commands/convert_jobs_to_wagtail.py
@@ -19,9 +19,11 @@ class Command(BaseCommand):
 
     def convert_applicant_types(self, commit=False):
         snippets = [
-            ApplicantTypeSnippet(**{k: getattr(at, k) for k in (
-                'applicant_type', 'slug', 'description', 'active',
-            )}) for at in ApplicantType.objects.all()
+            ApplicantTypeSnippet(
+                applicant_type_label=at.applicant_type,
+                description=at.description,
+                active=at.active
+            ) for at in ApplicantType.objects.all()
         ]
 
         count = len(snippets)
--- a/cfgov/jobmanager/management/commands/convert_jobs_to_wagtail.py
+++ b/cfgov/jobmanager/management/commands/convert_jobs_to_wagtail.py
@@ -1,0 +1,32 @@
+from __future__ import print_function
+
+from django.core.management.base import BaseCommand, CommandError
+
+from jobmanager.models import ApplicantType, ApplicantTypeSnippet
+
+
+class Command(BaseCommand):
+    help = 'Converts Jobmanager models to Wagtail'
+
+    def add_arguments(self, parser):
+        parser.add_argument('-d', '--dry-run', action='store_true',
+                            help='Create models but do not persist them')
+
+    def handle(self, *args, **options):
+        commit = not options['dry_run']
+
+        self.convert_applicant_types(commit=commit)
+
+    def convert_applicant_types(self, commit=False):
+        snippets = [
+            ApplicantTypeSnippet(**{k: getattr(at, k) for k in (
+                'applicant_type', 'slug', 'description', 'active',
+            )}) for at in ApplicantType.objects.all()
+        ]
+
+        count = len(snippets)
+        print('converted {} ApplicantType(s)'.format(count))
+
+        if snippets and commit:
+            ApplicantTypeSnippet.objects.bulk_create(snippets)
+            print('saved {} ApplicantTypeSnippet(s)'.format(count))
--- a/cfgov/jobmanager/migrations/0002_applicanttypesnippet.py
+++ b/cfgov/jobmanager/migrations/0002_applicanttypesnippet.py
@@ -15,13 +15,12 @@ class Migration(migrations.Migration):
             name='ApplicantTypeSnippet',
             fields=[
                 ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
-                ('applicant_type', models.CharField(max_length=255)),
-                ('slug', models.SlugField()),
+                ('applicant_type_label', models.CharField(max_length=255)),
                 ('description', models.TextField()),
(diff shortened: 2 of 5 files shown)


schaferjh:
A few questions: 
* what is the "slug" used for? Does it need to be visible to content managers? When creating a new Job Applicant Type is that auto generated or would a content manager have to create it?
* Could we call the label (first thing in the snippet view) "Applicant type label" to convey that this text will appear as the label/header of the applicant type description?
* I'm a little freaked out by the screenshot of "List of available applicant types:" because it includes grades and the applicant labels shouldn't include grade information. Is that screenshot using dummy data or do you need feedback on specifically which applicant types should be created for jobs going forward?

chosak:
@schaferjh:

-  `slug` actually doesn't appear to be used anywhere, as far as I can tell. It doesn't show up [in the current job listing template](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/about-us/careers/_single.html#L117). I could remove it unless you see a need to have this. All of the existing job-related models appear to have slugs, but I believe only the main `Job` slug may actually be used in the listing URL.
- Yes, I will modify that field name and display label as you suggest.
- That screenshot comes from the existing set of applicant types currently in production (there are 44 of these right now). Were grades once mixed into this table, perhaps in an earlier iteration? If we want to migrate over all existing/legacy jobs we will likely need to keep all of them around.

schaferjh:
* I support removing the slug for clarity as long as @rosskarchner isn't aware of some reason why it's needed.
* Thanks!
* My preference would be to only have snippets listed that are active. Does unchecking the "Active" box in the snippet remove it from the list of possible options that a content manager will see?

Also, where did we come down on whether or not we're migrating over all the legacy jobs? Is that still an open question? If so, maybe we should consult the stakeholder on whether they need access to legacy job data. If we DON'T mirgrate legacy jobs, what happens to that data?

chosak:
No, currently, like the slug, the `active` field doesn't do anything, either in the existing code or in this new version. So presumably it should be removed as well.

Unfortunately, after doing some digging, it doesn't appear that there is a way to filter Snippets within the Wagtail admin. For example, to say, "when choosing a Snippet in the admin Page editor, only allow content creators to choose those with, say, active=True". The way the Wagtail code is written also [makes it difficult](https://github.com/torchbox/wagtail/blob/v1.5.3/wagtail/wagtailsnippets/views/chooser.py#L21) for us to implement this behavior ourselves.

So, that means that using this Snippet-based approach, we'd have to either:

1. Always keep all Applicant Types around, even those that are deprecated. We could possibly mitigate this somehow by modifying the display name in the admin to something like "(DISABLED) Name" if `active=False` for that Snippet. We're only talking about 44 of these today, so it's not like there'd be a ton to filter out.
 1.  If we did this we'd probably want to investigate the feasibility of creating a PR to Wagtail to allow for some kind of configurable filtering of snippets in the admin.
2. Delete any Applicant Types that we no longer want to allow to be selected in the Wagtail Admin. This would cause issues with any Jobs that used those legacy Applicant Types.

A third option is not to use Snippets at all, and to retain the Django model, but that would mean that any changes/edits to the Applicant Types in future would have to use the Django admin, not Wagtail.

I have a slight preference for option 1, although it's a bit hacky. If don't want to migrate over all legacy jobs, this won't be as much of an issue, although if we ever want to disable one of these in future we'll have the same problem.

schaferjh:
I also have a preference for option 1 + possibly only migrating the data we need. I'll respond to your comment on GHE.

chosak:
Made the change to remove `slug` and rename "Applicant type label" in 3bd8dbc (see updated screenshot above). Left in `active` pending decision on whether to migrate over legacy applicant types.

chosak:
Closing this PR in favor of ModelAdmin-based approach in #2314.

Issue #496, opened 2015-04-29
Should we be using Google Analytics and Google Tag Manager snippets concurrently?


anselmbradford:
In https://github.com/cfpb/cfgov-refresh/blob/flapjack/_layouts/base.html#L113-L129 we have both Google Analytics (GA) and the Google Tag Manager (GTM) code snippets.

I noticed in https://support.google.com/tagmanager/answer/2574370 the following:

> A container snippet is provided for each container. Copy and paste this snippet immediately after the opening <body> tag on every page on the site. **Be sure to remove tags (e.g. Google Analytics, AdWords Conversion Tracking, etc)** from the site pages at the same time.

I'm not super familiar with GTM, but according to http://stackoverflow.com/questions/23640645/what-is-the-difference-between-google-tag-manager-and-google-analytics we can output GA code from GTM.

bethannenc:
That is true - once we have GTM fully implemented we can remove the old GA code and rely just on GTM. However, we have not removed the GA code as of yet because we wanted to let the two accounts run in parallel to make sure GTM was implemented correctly. We are almost at that point where we would feel comfortable removing the code - most likely by June 1. For new content, we are comfortable just implementing GTM and not adding the GA code snippet. 

I hope that helps - let me know if you have any other questions.

anselmbradford:
@bethannenc Do you know why GTM has a jQuery dependency and can that be removed/replaced with native code? http://www.googletagmanager.com/gtm.js?id=GTM-KMMLRS

bethannenc:
GTM itself does not have a jQuery dependency. However, some of the tags that we have written are written in jQuery and they will not work if jQuery is not on the page. We could fix that by translating all of our tags in jQuery into JS but that would be a major pain. @wernerc - thoughts?

KimberlyMunoz:
Thanks for asking this question. I had the same question for eRegs.

cc @ascott1

mistergone:
My goal is to try to have all Tag Manager code fit the following:

- For general use on any given page (link click handlers, for instance), use native Javascript
- For specific pages that have complex analytics, use jQuery to ease the writing of said code (assuming the page already uses jQuery, which most do). In these cases, we need to ask developers to export jQuery to `window.jQuery`. Good examples are the Paying for College cost comparison tool, Owning a Home's rate checker tool, etc.
[...]

PR #2300, opened 2016-08-11
WIP: Removing render-blocking javascript

contolini:
One of cf.gov's biggest performance faux pas is loading jquery in the `<head>`. After exploring all sorts of ways to mitigate this (including exporting all the GTM snippets in an attempt to rewrite them all without jQuery), I finally tried just removing it from the head and added it to the common.js bundle.

And, surprisingly, it seems to work fine. I'm still in the process of testing this across all devices but so far everything appears to be working as expected.

Is this a horrible idea @sebworks @virginiacc? Have I overlooked something?


jimmynotjim:
Awesome. I think we tried this and GTM barfed, but maybe something has been updated there to make it work?

KimberlyMunoz:
I think the way we consistently got the problem is testing on slow devices. If you test in Chrome Developer tools with 2G speeds, then GA would load before our JS and cause errors.

anselmbradford:
Yeah, per @jimmynotjim's comment, IIRC we moved jquery out of the head (and removed it entirely on pages that didn't use it) and it broke the GTM code that assumes jquery is in the head when it executes. The larger problem that would be great to address is that the GTM code should be moved to version control (maybe even this repo). We essentially have JavaScript introduced by GTM that circumvents all our code review process.

contolini:
@KimberlyMunoz Good to know, I'll look into that.

@anselmbradford AFAIK GTM is not conducive to external version control, it would involve a lot of copying and pasting. It has its own versioning system that keeps a snapshot of every version of our GTM container.

Here's a dump of all the GTM snippets for the curious: `https://[GHE]/raw/gist/contolini/a2bcfd2426611553ffd608e5f8f26da5/raw/b9c251ac7286ded2eae24146aa3204bfb21bca16/cfpb-gtm.json` If you search for `$` there are ~ 200 occurrences.

anselmbradford:
> @anselmbradford AFAIK GTM is not conducive to external version control, it would involve a lot of copying and pasting. It has its own versioning system that keeps a snapshot of every version of our GTM container.

Chuck looked into injecting scripts directly into the object GTM adds to the page, potentially making it so the code could live here and not in GTM @wernerc where was the thread where you discussed that?

sebworks:
I think this explains it in detail:

https://github.com/cfpb/cfgov-refresh/issues/496

sebworks:
How about this:
```
<noscript><iframe src="//www.googletagmanager.com/ns.html?id=GTM-KMMLRS"
height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>

<script>
    window.loadGTM = function loadGTM() {
        (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
        new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
        j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
        '//www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
        })(window,document,'script','dataLayer','GTM-KMMLRS');

        delete window.loadGTM;
    }
</script>

<script src="//ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js" onload="window.loadGTM()" async="true" ></script>
<script>if (!window.jQuery) { document.write('<script src="{{ static('js/jquery.min.js') }}" window.onload="loadGTM()" async="true"><\/script>'); }
</script>
```

This ensures that jQuery is loaded before loading GTM and adds the async tag to the jQuery scripts. This might break V1 scripts in some cases because we are still dependent upon jQuery is some places. Although, hopefully that will change soon.

mistergone:
These issues with GTM looking for jQuery on page load should be resolved (GTM should no longer barf on page load). There might be some specific pages where this still occurs, but I'm not currently aware of any (I think?). If this issue comes up again, or if someone sees a console error coming from GTM, please let me know!

contolini:
Thanks for the feedback, everyone! Super helpful. After chatting with folks, reading through the previous threads, and testing several script variants, I'm going to pause work on this for now. The odds of introducing a breaking change are high and the performance benefit doesn't justify the risk. We can reapproach it after the CF atomic stuff has wrapped up. Or if someone smarter than I wants to take a stab at it, please do.

PR #2301, opened 2016-08-11
Add handlers.py to v1/models to refactor JS filename handling

kurtrwall:
**This is the start of a series of PR's that deal with refactoring the CFGOVPage class to set up work to be done with filterable pages.**

This abstracts the JS filename handling for a page into it's own class for a more OO approach.

## Additions

- v1/models/handlers.py to hold the handler classes (so far just JSHandler)
- JSHandler: a class to handle the JS filename collection

## Removals

- JS handling code from CFGOVPage

## Changes

- JS filename handling now resides in handlers.py in the JSHandler class

## Testing

- run `tox -- v1.tests.models.test_handlers`
- Run a shell and grab a page that has expandables, expandable group, or filterable list in it's content
 - execute `page.media` and observe an OrderedDict returned with the appropriate filenames

## Review

- @Scotchester 
- @cfpb/cfgov-backends 

## Checklist

### DIFF
--- a/cfgov/v1/atomic_elements/molecules.py
+++ b/cfgov/v1/atomic_elements/molecules.py
@@ -6,6 +6,7 @@ from wagtail.wagtailimages.blocks import ImageChooserBlock
 from . import atoms
 from ..util import util, ref
 
+JS_MOLECULES = []
 
 def isRequired(field_name):
     return [str(field_name) + ' is required.']
--- a/cfgov/v1/atomic_elements/organisms.py
+++ b/cfgov/v1/atomic_elements/organisms.py
@@ -6,6 +6,13 @@ from . import atoms, molecules
 from ..util import ref
 from ..models.snippets import Contact as ContactSnippetClass
 
+JS_ORGANISMS = [
+    'BaseExpandable',
+    'Expandable',
+    'ExpandableGroup',
+    'FilterControls'
+]
+
 
 class Well(blocks.StructBlock):
     content = blocks.RichTextBlock(required=False, label='Well')
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -31,6 +31,7 @@ from modelcluster.tags import ClusterTaggableManager
 
 from sheerlike.query import QueryFinder
 
+from .handlers import JSHandler
 from .. import get_protected_url
 from ..atomic_elements import molecules, organisms
 from ..util import util, ref
@@ -307,50 +308,11 @@ class CFGOVPage(Page):
     def add_page_js(self, js):
         js['template'] = []
 
-    # Retrieves the stream values on a page from it's Streamfield
-    def _get_streamfield_blocks(self):
-        lst = [value for key, value in vars(self).iteritems()
-               if type(value) is StreamValue]
-        return list(chain(*lst))
-
-    # Gets the JS from the Streamfield data
-    def _add_streamfield_js(self, js):
-        # Create a dictionary with keys ordered organisms, molecules, then atoms
-        for child in self._get_streamfield_blocks():
-            self._add_block_js(child.block, js)
-
-    # Recursively search the blocks and classes for declared Media.js
-    def _add_block_js(self, block, js):
-        self._assign_js(block, js)
-        if issubclass(type(block), blocks.StructBlock):
-            for child in block.child_blocks.values():
-                self._add_block_js(child, js)
-        elif issubclass(type(block), blocks.ListBlock):
-            self._add_block_js(block.child_block, js)
-
-    # Assign the Media js to the dictionary appropriately
-    def _assign_js(self, obj, js):
-        try:
-            if hasattr(obj.Media, 'js'):
-                for key in js.keys():
-                    if obj.__module__.endswith(key):
-                        js[key] += obj.Media.js
-                if not [key for key in js.keys() if obj.__module__.endswith(key)]:
-                    js.update({'other': obj.Media.js})
-        except:
-            pass
-
     # Returns all the JS files specific to this page and it's current Streamfield's blocks
     @property
     def media(self):
-        js = OrderedDict()
-        for key in ['template', 'organisms', 'molecules', 'atoms']:
-            js.update({key: []})
-        self.add_page_js(js)
-        self._add_streamfield_js(js)
-        for key, js_files in js.iteritems():
-            js[key] = OrderedDict.fromkeys(js_files).keys()
-        return js
+        js_handler = JSHandler(self)
+        return js_handler.get_js_dict()
 
 
 class CFGOVPageCategory(Orderable):
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -1,5 +1,6 @@
-import os
 import json
+import os
+import urllib
 from itertools import chain
 from collections import OrderedDict
 
@@ -36,7 +37,8 @@ from .. import get_protected_url
 from ..atomic_elements import molecules, organisms
 from ..util import util, ref
 
-import urllib
+PAGE_HANDLERS = [JSHandler]
+
 
 
 class CFGOVAuthoredPages(TaggedItemBase):
@@ -215,6 +217,13 @@ class CFGOVPage(Page):
     def get_prev_appropriate_siblings(self, hostname, inclusive=False):
         return self.get_appropriate_siblings(hostname=hostname, inclusive=inclusive).filter(path__lte=self.path).order_by('-path')
 
+    def get_context(self, request, *args, **kwargs):
+        context = super(CFGOVPage, self).get_context(request, *args, **kwargs)
+        for handler_class in PAGE_HANDLERS:
+            handler = handler_class(self, request)
+            handler.process(context)
+        return context
+
     @property
     def status_string(self):
         page = CFGOVPage.objects.get(id=self.id)
@@ -307,12 +316,6 @@ class CFGOVPage(Page):
     # 'template' is used as the key for front-end consistency
     def add_page_js(self, js):
         js['template'] = []
-
-    # Returns all the JS files specific to this page and it's current Streamfield's blocks
-    @property
-    def media(self):
-        js_handler = JSHandler(self)
-        return js_handler.get_js_dict()
 
 
 class CFGOVPageCategory(Orderable):
--- a/cfgov/v1/models/handlers.py
+++ b/cfgov/v1/models/handlers.py
@@ -17,7 +17,7 @@ class Handler(object):
 
 class JSHandler(Handler):
     """
-        Gathers all the JS files specific to this page and it's current
+        Gathers all the JS files specific to this page and its current
         Streamfield's blocks and puts them in the template context.
     """
     def __init__(self, page, request):
--- a/cfgov/v1/models/handlers.py
+++ b/cfgov/v1/models/handlers.py
@@ -1,0 +1,56 @@
+from collections import OrderedDict
+from itertools import chain
+from wagtail.wagtailcore import blocks
+from wagtail.wagtailcore.blocks.stream_block import StreamValue
+
+
+class JSHandler:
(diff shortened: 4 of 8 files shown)


cfgov/v1/models/handlers.py
```diff
@@ -0,0 +1,56 @@
...
+
+    # Assign the Media js to the dictionary appropriately
+    def assign_js(self, obj):
+        try:
+            if hasattr(obj.Media, 'js'):
+                for key in self.js_dict.keys():
+                    if obj.__module__.endswith(key):
+                        self.js_dict[key] += obj.Media.js
+                if not [key for key in self.js_dict.keys()
+                        if obj.__module__.endswith(key)]:
```
chosak: Use of `obj.__module__` makes me a bit nervous. What's an example of the module name you're checking against here? Something like `v1.atomic_elements.atoms`? If in future we were to break `atoms.py` into a directory like `atoms/whatever.py`, will this code break?

cfgov/v1/models/handlers.py
```diff
@@ -0,0 +1,56 @@
...
+                    self.js_dict.update({'other': obj.Media.js})
+        except:
```
chosak: [PEP8](https://www.python.org/dev/peps/pep-0008/#id49) discourages using bare `except` statements like this (search for "bare except"). What is this `try/except` protecting against? Are there "expected" exceptions that we can/want to ignore?

cfgov/v1/models/handlers.py
kurtrwall: Exactly, it will break. Good catch. I'll try to think of something more robust.

cfgov/v1/models/handlers.py
kurtrwall: It's been a while since this was written, so I'll do some poking to get this resolved.

Scotchester:
I'm not sure that I have enough understanding of what's going on to review this PR, but hooray that work has started!

kurtrwall:
Changed this up a bit for @chosak's comments, and moved the handler logic into `get_context` for generic handler class processing.

cfgov/v1/models/handlers.py
```diff
@@ -6,11 +6,27 @@
...
         self.page = page
+        self.request = request
+
+    def process(self, context):
+        raise NotImplementedError
+
+
+class JSHandler(Handler):
+    """
+        Gathers all the JS files specific to this page and it's current
```
chosak: nitpick: should be "its"

richaagarwal:
👍

cfgov/v1/models/handlers.py
```diff
@@ -0,0 +1,78 @@
...
+        Gathers all the JS files specific to this page and its current
+        Streamfield's blocks and puts them in the template context.
+    """
+    def __init__(self, page, request):
+        super(JSHandler, self).__init__(page, request)
+        self.js_dict = OrderedDict()
+
+    def process(self, context):
+        self.generate_js_dict()
+        context['media'] = self.js_dict
```
chosak: In this PR there's only a single `Handler` in [this](https://github.com/cfpb/cfgov-refresh/pull/2301/files#diff-8e8a3e28d46ee12516bea4f5b50bf5a4R222) loop. If there are multiple handlers in future, won't assigning `context['media']` like this potentially overwrite some other media that another handler might have put in the dictionary?

cfgov/v1/models/handlers.py
```diff
@@ -0,0 +1,78 @@
...
+    def add_block_js(self, block):
+        self.assign_js(block)
+        if issubclass(type(block), blocks.StructBlock):
+            for child in block.child_blocks.values():
+                self.add_block_js(child)
+        elif issubclass(type(block), blocks.ListBlock):
+            self.add_block_js(block.child_block)
+
+    # Assign the Media js to the dictionary appropriately
+    def assign_js(self, obj):
```
chosak: Would be nice to have some unit tests of this function, since the logic here is a bit complicated. When is that `AttributeError` raised?

cfgov/v1/models/handlers.py
```diff
@@ -0,0 +1,78 @@
...
+        try:
+            if hasattr(obj.Media, 'js'):
+                class_name = type(obj).__name__
+                for key in self.js_dict:
+                    if key in ATOMIC_JS:
+                        if class_name in ATOMIC_JS[key]:
```
vickumar1981: i believe you and combine these two if statements with an and.

cfgov/v1/tests/models/test_handlers.py
```diff
@@ -0,0 +1,72 @@
...
+        self.page = mock.Mock()
+        self.request = mock.Mock()
+        self.handler = Handler(self.page, self.request)
+
+    def test_process_raises_NotImplementedError(self):
+        with self.assertRaises(NotImplementedError) as nie:
+            self.handler.process({})
+
+    def test_get_streamfield_blocks_returns_list(self):
+        assert type(self.handler.get_streamfield_blocks()) is list
```
vickumar1981: i think you can use assertIsInstance for this case.

cfgov/v1/tests/models/test_handlers.py
```diff
@@ -0,0 +1,72 @@
...
+    @mock.patch('v1.models.handlers.JSHandler.generate_js_dict')
+    def test_process_sets_context_media_od(self, mock_generate_js_dict):
+        context = {}
+        self.js_handler.process(context)
+        assert 'media' in context
+        assert type(context['media']) is OrderedDict
+
+    @mock.patch('v1.models.handlers.JSHandler.add_streamfield_js')
+    def test_generate_js_dict_returns_OrderedDict(self, mock_add_sf_js):
+        assert type(self.js_handler.js_dict) is OrderedDict
```
vickumar1981: ditto here.

cfgov/v1/models/handlers.py
```diff
@@ -0,0 +1,78 @@
...
+    def __init__(self, page, request):
+        self.page = page
+        self.request = request
+
+    def process(self, context):
+        raise NotImplementedError
+
+    # Retrieves the stream values on a page from it's Streamfield
+    def get_streamfield_blocks(self):
+        lst = [value for key, value in vars(self.page).iteritems()
```
vickumar1981: i think you can use just values() here b/c the key isn't being used.

cfgov/v1/models/handlers.py
```diff
@@ -0,0 +1,78 @@
...
+               if type(value) is StreamValue]
+        return list(chain(*lst))
```
vickumar1981: i'm not sure about this line.  there's a typecast to a list b/c itertools.chains returns an iterable.  chain is normally used to combine iterables.  i only see a single iterable here - lst.

cfgov/v1/models/handlers.py
```diff
@@ -0,0 +1,78 @@
...
+                            break
+                    elif key == 'other':
+                        self.add_files(key, obj.Media.js)
+                        break
+        except AttributeError:
+            pass
+
+    def add_files(self, key, filenames):
+        self.js_dict[key] += [name for name in filenames
+                              if name not in self.js_dict[key]]
```
vickumar1981: might be cleaner here to use a set.  from what i get, you're adding 'names' from  'filenames' which don't exist in the 'js_dict[key]'.

i.e.,
```
self.js_dict[key] = list(set(self.js_dict[key] + filenames))
```

cfgov/v1/atomic_elements/molecules.py
```diff
@@ -6,6 +6,7 @@
 from . import atoms
 from ..util import util, ref
```
vickumar1981: util is not used in this file.  and neither is ValidationError.  please remove.

cfgov/v1/models/base.py
```diff
@@ -1,5 +1,6 @@
-import os
 import json
```
vickumar1981: please remove the following imports:

```
from itertools import chain
from collections import OrderedDict

from django.core.urlresolvers import reverse
...
from wagtail.wagtailcore import blocks
from wagtail.wagtailcore.blocks.stream_block import StreamValue
...
from wagtail.wagtailcore.templatetags.wagtailcore_tags import slugurl
...
from sheerlike.query import QueryFinder
```

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -6,6 +6,13 @@
 from ..util import ref
 from ..models.snippets import Contact as ContactSnippetClass
 
+JS_ORGANISMS = [
```
vickumar1981: Why not make this a list of types instead of strings?

below you define page handlers:

`PAGE_HANDLERS = [JSHandler]`

cfgov/v1/models/handlers.py
vickumar1981: :+1:  this is a really slick way to flatten a list.

cfgov/v1/models/handlers.py
kurtrwall: @chosak This is true, I'll make a change that checks for the existence of `media`. There's a future PR that actually uses Wagtail's hooks module to register these handlers instead.

cfgov/v1/models/handlers.py
kurtrwall: `AttributeError` is raised when the `obj` object doesn't have a `Media` property. I could change it to check for the property instead.

cfgov/v1/tests/models/test_handlers.py
kurtrwall: I cannot find `assertIsInstance`. [Here's the mock class.](https://docs.python.org/3/library/unittest.mock.html#the-mock-class)

cfgov/v1/models/handlers.py
kurtrwall: This is true, but this serves another purpose too. To maintain the order of the files that are included in the list, it cannot be casted to a set since the ordering is lost when doing so.
```python
>>> lst = ['a', 'b', 'c']
>>> new = ['a', 'd']
>>> lst += new
>>> lst
['a', 'b', 'c', 'a', 'd']
>>> set(lst)
set(['a', 'c', 'b', 'd'])
>>> lst = list(set(lst))
>>> lst
['a', 'c', 'b', 'd']
```

cfgov/v1/atomic_elements/organisms.py
kurtrwall: I'm not sure, but this is changing in a follow PR where I'm using Wagtail's hooks module instead of importing the handlers and adding them to a list. Much cleaner.

cfgov/v1/models/handlers.py
kurtrwall: I think the splat is cleaner :D I'm open to reasons why it's not though.

cfgov/v1/models/handlers.py
chosak: Could you remove the `try/except` by doing something like:

```py
if hasattr(obj, 'Media') and hasattr(obj.Media, 'js'):
```

cfgov/v1/tests/models/test_handlers.py
chosak: I think @vickumar1981 might mean [this](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertIsInstance), e.g.

```sh
self.assertIsInstance(self.handler.get_streamfield_blocks(), list)
```

cfgov/v1/atomic_elements/organisms.py
chosak: Do you still want to merge this PR, or is it better to wait for the future, improved one?

cfgov/v1/models/handlers.py
kurtrwall: Yes, I'll do that now.

kurtrwall:
Thanks everyone for your comments and suggestions. I'm closing this PR in favor one that supersedes some of the logic contained within this PR as well as updates and includes more tests.

PR #2309, opened 2016-08-12
Create new virtualenvs with python2.7 explicitly

willbarton:
Our `activate-virtualenv.sh` script should create a virtualenv with python2.7 explicitly. 

It's possible for the default `python` to be Python 3.x, or for a user to have `VIRTUALENVWRAPPER_PYTHON` set to Python 3.x, in which case the virtualenv will be created with that version of Python.

Currently it is not possible to set up the codebase for Python 3 because of MySQL-python, which does not yet support Python 3. 

## Additions

- Add a test for `python2.7`'s existence in the user's path, and use it explicitly when creating the virtualenv.

## Review

- @cfpb/cfgov-backends 
- @Scotchester


Scotchester:
I don't think I can test this anytime soon, but I approve of the concept.

chosak:
😢 that we can't support Python 3, but 👍 to this check.

Is it worth adding a note in the documentation somewhere about the code only working with Python 2.7?

willbarton:
@chosak I'm of two minds about it... our inability to support Python 3 is purely down to our use of [MySQL-python](https://pypi.python.org/pypi/MySQL-python/). [Django recommends mysqlclient](https://docs.djangoproject.com/en/1.10/ref/databases/#mysql-db-api-drivers) as an alternative.

I think I'd rather explore the best way to fix the problem (and see if @rosskarchner @richaagarwal @kurtrwall can provide any context around MySQL-python vs mysqlclient). It's a silly reason not to support Python 3.

But my goal here is to make sure the base setup instructions don't cause failure because they expect a version they aren't going to get.

activate-virtualenv.sh
```diff
@@ -25,11 +25,17 @@ if ! type "workon" &>/dev/null; then
   source $virtualenvwrapper_path
 fi
 
+if ! which python2.7 &>/dev/null; then
```
vickumar1981: what if i have python 2.7 installed but the symlink is /usr/bin/python which is the default?

i think the reason we have a /usr/bin/python2.7 (not really standard) on our dev servers is b/c we can't remove python 2.6.

willbarton: The assumption is that if there is a `python2.7` in our `PATH` somewhere, it is expected to actually point to a Python 2.7 interpreter. I think that's a reasonable assumption, unless I'm misunderstanding the concern.

Otherwise, I think if you've got a slightly esoteric Python setup that you're afraid this will break, it's easy enough to make the virtualenv yourself. These scripts are only really intended for local development anyway. I know for a fact the only one we use as part of the deployment process is `frontend.sh`. 

That all being said, I think we all agree that (1) these scripts are fragile, and make too many assumptions as is, (2) are hard to troubleshoot when they do fail, and (3) the real problem here is our dependency on a Python 3-incompatible library.

rosskarchner:
👍

higs4281:
👍 script works with new check

PR #2317, opened 2016-08-16
Remove CF component demo dependency

contolini:
AFAIK this is cruft (and jenkins doesn't like it).

## Removals

- `cf-component-demo` dependency

## Review

- @virginiacc 
- @sebworks 
- @anselmbradford

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
   },
   "devDependencies": {
     "browser-sync": "2.11.2",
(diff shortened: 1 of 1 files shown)


Scotchester:
Correct, it is no longer used.

Scotchester:
:+1:

sebworks:
👍

PR #2318, opened 2016-08-16
added a test for get_protected_url

grapesmoker:
This PR adds a test for get_protected_url, which was a blocker for a few other things that need unit tests.

## Review

@chosak @willbarton @richaagarwal 

## Checklist


cfgov/v1/tests/test_links.py
```diff
@@ -1,8 +1,12 @@
 from django.test import TestCase
-from v1 import parse_links
-from bs4 import BeautifulSoup
+from v1 import parse_links, get_protected_url
+# from bs4 import BeautifulSoup
```
chosak: Remove commented import.

cfgov/v1/tests/test_links.py
```diff
@@ -22,6 +26,41 @@ def test_gov_link(self):
...
+        self.page.live = True
+        self.page.specific.shared = True
+        self.context = mock.MagicMock()
+
+        def context_dict(name):
+            return {'request': self.request}[name]
+
+        self.context.__getitem__.side_effect = context_dict
+
+    def test_get_protected_url(self):
```
chosak: Can you break this test into separate tests, each with a single assert? Is there some dependence between them?

richaagarwal: Yeah, want to echo @chosak's comment here & suggest that the name of the test represents what's being tested in that single test, and/or have comments to explain what each test is testing.

cfgov/v1/tests/test_links.py
```diff
@@ -22,6 +26,41 @@ def test_gov_link(self):
...
 
 
+class LinkTest(TestCase):
+
+    def setUp(self):
+
+        self.page = mock.MagicMock()
+        self.request = mock.MagicMock()
+        self.request.url = 'http://localhost:8000/about-us/blog/'
+        self.page.url = 'http://localhost:8000/about-us/blog/we-want-hear-public-about-payday-loans/'
```
vickumar1981: if page.url simply appends onto the request url, it might be better to write it using string interpolation, i.e.:

`self.page.url = "%swe-want-hear..." % self.request.url`

cfgov/v1/tests/test_links.py
```diff
@@ -22,6 +26,41 @@ def test_gov_link(self):
...
+
+        self.context.__getitem__.side_effect = context_dict
+
+    def test_get_protected_url(self):
+
+        result = get_protected_url(self.context, None)
+        self.assertEqual(result, '#')
+        self.page.url = None
+        result = get_protected_url(self.context, self.page)
+        self.assertIs(result, None)
```
vickumar1981: unit test has a method called assertIsNone for this case.

cfgov/v1/tests/test_links.py
grapesmoker: The values here are purely incidental; they don't really represent anything.

cfgov/v1/tests/test_links.py
grapesmoker: There's no dependence here; I was testing the various possibilities that get_protected_url could return, which to my mind made sense to combine into a single unit test, since the unit being tested is the function. But if people think the different return values should each have their own test, I can certainly do that.

cfgov/v1/tests/test_links.py
grapesmoker: how do i scrolled autocomplete

cfgov/v1/tests/test_links.py
willbarton: I agree with @chosak and @richaagarwal. We want to test the various possibilities, but they should probably be separate tests.

cfgov/v1/tests/test_links.py
vickumar1981: probably depends on your ide.  i use pycharm.

Here's the list of assert methods available in unittest:

https://docs.python.org/2/library/unittest.html#assert-methods

grapesmoker: sorry, that was a badly made joke. I know how to scroll autocomplete.

grapesmoker:
fixorated

cfgov/v1/tests/models/test_sublanding_page.py
```diff
@@ -0,0 +1,15 @@
...
+class SublandingPageTestCase(TestCase):
+
+    def setUp(self):
+        self.request = mock.MagicMock()
+        self.limit = ''
+        pass
+
+    def test_get_browsefilterable_posts(self):
+
+        pass
```
chosak: Can this file be removed? It doesn't appear to be doing anything.

grapesmoker: Sorry, don't know how that got in there. That's for a different set of tests I'm working on. I'll get rid of it.

cfgov/v1/tests/test_links.py
```diff
@@ -22,6 +25,51 @@ def test_gov_link(self):
...
+        self.assertEqual(result, self.page.url)
+
+    def test_get_protected_url_non_live_page(self):
+        self.page.live = False
+        result = get_protected_url(self.context, self.page)
+        self.assertEqual(result, '#')
+
+    def test_get_protected_url_staging(self):
+        self.page.live = False
+        self.request.url = 'http://content.localhost:8000/path/to/page/'
```
richaagarwal: is this line needed?  it doesn't seem like it would update context on its own.

grapesmoker: Yes because originally the request.url is initialized to a different value.

richaagarwal:
@grapesmoker Thanks for your work on this!  Apologies for not chiming in with this sooner, but could you either add a comment to each test to describe what the expected behavior is, or could you update the names to be more descriptive - e.g. the last test `test_get_protected_url_staging` should say something like "Prod URL on a Staging page should get converted to Staging URL".  I realize that's a mouthful for the name of the test, so it'd be good to have stuff like that in a comment at the very least.

grapesmoker:
No problem, I'll do that.

cfgov/v1/tests/test_links.py
```diff
@@ -22,6 +25,68 @@ def test_gov_link(self):
...
+        Confirm that if a live page is requested from the same host as the request
+        we get back the original page url.
+        """
+        self.page.url = 'http://localhost:8000/path/to/some/other/page/'
+        result = get_protected_url(self.context, self.page)
+        self.assertEqual(result, self.page.url)
+
+    def test_get_protected_url_non_live_page(self):
+        """
+        Confirm that if a non-live page is request, we get back the hash.
```
richaagarwal: I might update this to "Confirm that if a non-live page is requested from a live page, we get back the hash" since it's only true under those conditions.

cfgov/v1/tests/test_links.py
```diff
@@ -1,8 +1,11 @@
...
+from v1 import parse_links, get_protected_url
+
+import mock
+
 
 class ImportDataTest(TestCase):
+
     def test_external_link(self):
         link = '<a href="https://wwww.google.com">external link/a>'
         output = str(parse_links(link))
```
vccabral: why is parse_links used? Seems like you can no longer use the expressive parts of beautiful soup now. Consider a refactor to at least pass the params to the find_all. This is preventing you from doing a direct compare on the remaining returned list (which could just use a css selector and assert the length) .

grapesmoker: I have no idea, I did not write this code.

richaagarwal: I did, but yeah, let's keep PR review to the changed code.

cfgov/v1/tests/test_links.py
richaagarwal: @grapesmoker See comment above. Still waiting on this to be addressed before I can give it a thumbs up!

richaagarwal: also, @schaferjh, could you take a look at the conditions described here & confirm that this is the expected behavior?

grapesmoker: @richaagarwal noted, will correct

schaferjh: Thanks for the tag @richaagarwal. Based on my initial understanding, I think this is the desired behavior but I don’t fully understand the test descriptions. Below are my not-a-developer questions, could you help me out @grapesmoker? 

> Confirm that in the absence of a page, get_protected_url returns the hash. 

I don’t know what “the absence of a page” might mean. Also not sure what returning “the hash” means as a behavior?

> Confirm that in the absence of a page URL, get_protected_url returns None.

Does this mean that any link would be null without a URL? If yes, +1

 > Confirm that if a live page is requested from the same host as the request we get back the original page url.

The reference to “original page url" throws me, but if this means that when a page is requested from www, we get back the requested page at www, then +1

> Confirm that if a non-live page is request, we get back the hash. aka, "Confirm that if a non-live page is requested from a live page, we get back the hash”

I think this is confirming that a staged page cannot be requested from www, which is good. Again, not sure what returning the hash means as a behavior.

> Make sure that if we're on the staging site and request a page from that site which does not point to the same host, that we replace the hostname with the staging hostname.

I think this means that if I’m on staging.cf.gov/about-us/ and I request www.cf.gov/about-us/blog for whatever reason, I get back staging.cf.gov/about-us/blog. If yes, +1

grapesmoker: > I don’t know what “the absence of a page” might mean. Also not sure what returning “the hash” means as a behavior?

"Absence of a page" means "the page object is None." Sorry, should have been clearer on that.

> Does this mean that any link would be null without a URL? If yes, +1

Yes.

> The reference to “original page url" throws me, but if this means that when a page is requested from www, we get back the requested page at www, then +1

Yes, this is what that means. 

> I think this is confirming that a staged page cannot be requested from www, which is good. Again, not sure what returning the hash means as a behavior.

Again, I should have been clearer on this. This function is actually a template tag that "protects" a URL so that, as you say, a staged URL cannot be requested from www. The return value of that function is the hash, '#', which is interpolated into the template, which is what I meant.

> I think this means that if I’m on staging.cf.gov/about-us/ and I request www.cf.gov/about-us/blog for whatever reason, I get back staging.cf.gov/about-us/blog. If yes, +1

Correct.

schaferjh: @grapesmoker thank you! Based on your write up then, yes this is the expected behavior.

richaagarwal:
👍

chosak:
:+1:

grapesmoker:
If there are no more comments, I'll update and merge?

richaagarwal:
@grapesmoker go for it!

PR #2321, opened 2016-08-18
HOLD - Add StoryPage model and template

Scotchester:
⚠️ **ON HOLD** per the comments below, in order to set up stories to just be another kind of Document Detail page.

This PR adds the model class and the template needed to create pages of type Story, to be used to migrate [Everyone Has a Story](http://www.consumerfinance.gov/everyone-has-a-story/) into Wagtail pages.

## Additions

- `StoryPage` class / page type
- New template for individual Story pages
- New categories to which Stories can be assigned
- New category icon associations for the new categories

## Changes

- Fixes a rendering bug in the video player that would cause a black bar at the bottom of the preview image
- Fixes a bug with selecting the Archive Image for Events

## Testing

1. Pull down the PR code
1. `gulp styles`
1. `./cfgov/manage.py runserver`
1. Log in at <http://localhost:8000/admin/>
1. Create a new Sublanding Filterable page.
1. Create a child page of that with type Story.
1. Enter some basic data in the child page, but definitely include:
   - Item Introduction with Category and Title
   - A YouTube video URL, like https://www.youtube.com/embed/Alzp47ipDnY
   - A Preview Description, Preview Image, and Category (same one selected for the Item Introduction) on the Configuration tab
1. Return to the test Sublanding Filterable page and add a Filter Controls module with page type of Story
1. Publish and preview the test pages

## Review

- @cfpb/cfgov-backends 
- @cfpb/cfgov-frontends 

### Sublanding page

screen shot 2016-08-17 at 17 49 54

### Story page

screen shot 2016-08-18 at 12 46 05

## Notes and questions

- Should the "Story" grouping for the new categories be called something more generic, like "Financial products"?
- I can't figure out why the image's alt property is only returning the title, even if the alt field has something entered in it in Wagtail. (See TODO note on `cfgov/jinja2/v1/story-page/index.html` L35)

## Todo (in another PR)

- Extend the filterable list output so that it can output in a 50-50 type of format. See mockup at https://[GHE]/design-ops/CE-Debt-collection/issues/1

## Checklist


cfgov/v1/models/learn_page.py
```diff
@@ -39,6 +39,7 @@ class AbstractFilterPage(CFGOVPage):
 
 
     # Configuration tab panels
+    header_panel = [StreamFieldPanel('header')]
```
kurtrwall: What's this?

Scotchester: I had to add that to get the standard Header panel from AbstractFilterPage into my subclass. See this line: https://github.com/cfpb/cfgov-refresh/pull/2321/files#diff-c53bb3e17622a41cfcf254dedc4920dcR135

kurtrwall:
Fix up that migration, rebase from master, and I'll come check again. Otherwise, it's all good from me.

cfgov/v1/models/learn_page.py
```diff
@@ -106,6 +107,47 @@ class DocumentDetailPage(AbstractFilterPage):
     template = 'document-detail/index.html'
 
 
+class StoryPage(AbstractFilterPage):
```
chosak: Can this page be added as a new Python module (separate file) instead of adding it to `learn_page.py`, which is already rather crowded?

More generally, are there additional task around other content or logic for migration of Everyone Has a Story that might motivate treating it as a separate app from `v1`?

cfgov/v1/models/learn_page.py
```diff
@@ -106,6 +107,47 @@ class DocumentDetailPage(AbstractFilterPage):
...
+    video_url = models.URLField("YouTube URL", blank=True,
+                                help_text="Format: https://www.youtube.com/embed/video_id. "
+                                          "It can be obtained by clicking on Share > Embed on YouTube.",
+                                validators=[RegexValidator(
```
vickumar1981: might want to use URLValidator here.  It's a subclass of RegexValidator.

chosak:
Can you create some unit tests around programmatically creating a `StoryPage` and rendering it to a template? See also @richaagarwal tests [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/tests/wagtail_pages/test_page_titles.py).

Scotchester:
Rebased and updated the migration. (Had to force push, if you have already pulled down this branch.)

cfgov/v1/models/learn_page.py
Scotchester: I don't think that EHAS is significantly different enough from other types of content that it would warrant its own app, but I'm not entirely sure what our philosophy is on when things should be turned into separate apps.

Scotchester: Oops, missed your first point in this comment.

I could, but it seemed logical to follow the pattern of declaring other subclasses of AbstractFilterPage in this same file. Should I break all of them out into their own file?

schaferjh:
Left field question: Why did you create a new page type rather than adding necessary functionality to the Doc Detail or Learn templates rather than creating a new page type? 

I mean, I recognize that the work is done and it's not a big issue. But Blog, Newsroom and this template are all essentially variations of Learn pages and at some point in the future we'll want to streamline them.

Scotchester:
@schaferjh Because I need to be able to put them in a filterable list on a sublanding filterable page, and you have to give that a page type.

Scotchester:
@chosak Added `StoryPage` to the test you linked to. Are there other tests it should also be added to?

cfgov/v1/models/learn_page.py
Scotchester: I'm intrigued. But, given the specificity required for the regex passed to it, what would be the advantage of using URLValidator?

schaferjh:
> Because I need to be able to put them in a filterable list on a sublanding filterable page, and you have to give that a page type.

But Document detail pages can be in a filterable list from a sublanding page?

cfgov/v1/models/learn_page.py
chosak: There's a lot of stuff mixed into those files right now, but maybe for the scope of this PR it might be best to just make a new file for your new page.

It's probably fair to say that the philosophy about when to break out things into separate apps is still evolving (@willbarton @rosskarchner). The `jobmanager` app is separate and has its own page model and related things, but that's mostly because it has some Django models. It doesn't seem like this Story PR has enough to warrant being separate.

chosak:
> Are there other tests it should also be added to?

I don't think there are any other tests that go through a set of page types like that one does.

Scotchester:
> But Document detail pages can be in a filterable list from a sublanding page?

@schaferjh I finally see what you're getting at – but it's not _just_ doc detail pages, it's a bunch of different, I'll call them "sub-types" of doc detail pages, e.g., Amicus Briefs, Final Rules, etc.

What I _don't_ understand still is how to, when creating a new doc detail page, relate it to those sub-types. I.e., what do you do to a new doc detail page to trigger it for inclusion when "Amicus Brief" is selected as the page type in a filterable list?

Anyone besides @kurtrwall able to answer this easily? If not, it's okay to wait for him to return, because the story rollout is going to be delayed due to shifting priorities.

schaferjh:
> What I don't understand still is how to, when creating a new doc detail page, relate it to those sub-types. I.e., what do you do to a new doc detail page to trigger it for inclusion when "Amicus Brief" is selected as the page type in a filterable list?

The filter module grabs all the document detail child pages below the page it is placed on. 

So, I related a new document detail page to the Amicus filter by creating a document detail child page under the Amicus browse filterable page. I related a new document detail page to the Final Rules section, by creating the new doc detail page as a child to the Final Rules browse filterable page. Step and repeat for all content types within filterable lists across the site. 

Kurt may have a more dev-friendly explanation, but that's how it works for a content manager.

Scotchester:
@schaferjh 

Ok, after doing a bit more thorough testing, you are correct! And I now realize that the "Page type" field in the Filter Controls module tells the module _which categories_ to display in the filter box. I incorrectly assumed it autodetected the available categories and spit them out.

So, I think you're right. We should not create a new page type for this.

As an MVP, I could theoretically do nothing and just use the Doc Detail type straight-up, because YouTube video embed code _can_ just be dropped inside a rich text field.

But what do you think about creating a new module for using our fancy video player in any main content area?

schaferjh:
>But what do you think about creating a new module for using our fancy video player in any main content area?

@Scotchester sounds good to me!

sebworks:
@Scotchester, this is some old code that might be helpful. I was in the process of removing some of the logic from the template. https://github.com/cfpb/cfgov-refresh/commit/17a7c32e61282db65bbad76d5791ed4bbee5e252

Scotchester:
@sebworks Cool, thanks! I will probably take a run at this next week.

cfgov/v1/models/learn_page.py
vccabral: You can chain the validators but I think @scotchest is right, this is better. More specific.

cfgov/jinja2/v1/story-page/index.html
```diff
@@ -0,0 +1,54 @@
+{% extends 'layout-2-1-bleedbar.html' %}
+
+{% import 'macros/video-player.html' as video_player %}
+{% import 'organisms/item-introduction.html' as item_introduction with context %}
+{% import 'templates/render_block.html' as render_block with context %}
+{% import 'templates/streamfield-sidefoot.html' as streamfield_sidefoot with context %}
+
+{% block content_main_modifiers -%}
+    {{ super() }} content__flush-bottom
```
vccabral: I have never seen parenthesis in a template block, is this a jinja thing?

cfgov/v1/models/learn_page.py
```diff
@@ -110,6 +111,47 @@ class DocumentDetailPage(AbstractFilterPage):
...
+                                    regex='^https?:\/\/www\.youtube\.com\/embed\/.*$'
```
vccabral: This can be made a little more specific JUXJtZex9PI => [\w\d]{10}

cfgov/v1/models/learn_page.py
```diff
@@ -155,8 +197,8 @@ class EventPage(AbstractFilterPage):
         related_name='+'
     )
     flickr_url = models.URLField("Flickr URL", blank=True)
-    youtube_url = models.URLField("Youtube URL", blank=True,
-    help_text="Format: https://www.youtube.com/embed/video_id. It can be obtained by clicking on Share > Embed on Youtube.",
+    youtube_url = models.URLField("YouTube URL", blank=True,
+    help_text="Format: https://www.youtube.com/embed/video_id. It can be obtained by clicking on Share > Embed on YouTube.",
```
vccabral: 👍

cfgov/v1/models/learn_page.py
```diff
@@ -155,8 +197,8 @@ class EventPage(AbstractFilterPage):
...
     validators=[ RegexValidator(regex='^https?:\/\/www\.youtube\.com\/embed\/.*$')])
```
vccabral: If you inherit from a validator you can make it reusable.

vccabral: I see this validator used twice (it also might just be a method, or a callabe, cant remember now)

cfgov/jinja2/v1/story-page/index.html
rosskarchner: Yes-- Jinja doesn't automatically call bare functions (ie, it behaves more like Python)

cfgov/v1/models/learn_page.py
sebworks: Great point @vccabral, I don't why the hell i didn't do that.

willbarton:
@Scotchester What's the status of this PR?

Scotchester:
Sorry for the delay on this, Will. The project driving it is undergoing some rethinking, so this has been backburnered for a while.

I may reuse this branch, but as the purpose of the update will fundamentally change (see https://github.com/cfpb/cfgov-refresh/pull/2321#issuecomment-241111801), I think it makes sense to close this PR and open a new one when the time comes.

PR #2325, opened 2016-08-19
Removes potentially misleading instructions from backend.sh

anselmbradford:
## Changes

- Remove `backend.sh` message about activating virtual environment.

## Testing

- Open a new tab and cd to the project directory.
- Run `deactivate` if a virtual environment was activated.
- Run `./backend.sh` and see that the message output does not include the virtualenv.

## Review

- @richaagarwal 
- @chosak 
- @kurtrwall

### DIFF
--- a/backend.sh
+++ b/backend.sh
@@ -16,8 +16,7 @@ init() {
 
   # Ensure that we're in a virtualenv
   python -c 'import sys; sys.real_prefix' 2>/dev/null || (
-    echo 'Please activate your virtualenv before running this script:' \
-         '$ workon cfgov-refresh' &&
+    echo 'Please activate your virtualenv before running this script.' &&
     exit 1
   )
 }


chosak:
This seems okay to me, with the big caveat of "we have too many setup scripts that interact with each other in an overly complicated way". One thing is, can we add a check before this to enforce that `VENV_NAME` is set? Since we're forcing use of `bash` in this script, we could do something like:

```sh
: "${VENV_NAME?Looking for virtualenv name in \$VENV_NAME}"
```

Although, as I write this, part of me thinks we should just remove this variable and the suggested `workon` message from this script. It assumes that people are using `virtualenvwrapper` (which they might not be), and then adds this other dependency on an external environment variable, which really has nothing to do with what this script is actually doing.

anselmbradford:
@chosak Sure, I'm fine with whatever. I just noticed this because due to legacy reasons with my involvement in the project my virtualenv is `flapjack` not `cfgov-refresh` so the instructions it spits out were wrong. Which direction do you want to go with this PR? Instead of adding logic to check if `VENV_NAME` is set here, I would suggest ensuring `VENV_NAME` has a default set further up. Not sure where we ended up on that earlier discussion around defaults.

chosak:
@anselmbradford I don't feel strongly either. I notice that [`activate-virtualenv.sh`](https://github.com/cfpb/cfgov-refresh/blob/master/activate-virtualenv.sh#L8) already has the check for this environment variable, which seems to be the only place that it's used. I'd probably lean towards removing the reference to it here just to keep its use limited to that single place.

willbarton:
I agree with @chosak. If we're thinking about this from the "too many setup scripts that interact in a non-trivial way" problem @chosak mentioned, then I think the suggestion of removing the reference here is the ideal way to solve this problem.

anselmbradford:
@chosak @willbarton Removed the reference in https://github.com/cfpb/cfgov-refresh/pull/2325/commits/de97935f397b91c5721f716a180aaf309b746026

willbarton:
:shipit:

richaagarwal:
@anselmbradford I see there was some discussion above.  Is the goal of this PR now to just decouple activating the virtual environment from the `backend.sh` script?

anselmbradford:
@richaagarwal The instructions it gave in some cases was wrong, so yeah I removed those instructions per the earlier discussion.

richaagarwal:
@anselmbradford The title of the PR currently says "Updates backend.sh message to use envvar value" but it doesn't use any kind of envvar.  Could you update the title and description of the PR to reflect what it is - both to help me but mostly to avoid confusion in the future.

anselmbradford:
@richaagarwal Sure! Updated it

richaagarwal:
👍

PR #2334, opened 2016-08-23
Squash gulp lint:build warnings/errors

anselmbradford:
## Changes

- Updated ESLint dot-notation rule to support `catch` block in a Promise.
- Updated `gulp test:perf` task to use a Promise.
- Added `.eslintrc` override for gulp tasks to allow process.exit and console logging.

## Testing

- `gulp lint:build` should only warn about TODOs.
- `gulp test:perf` should work as expected (@contolini plz see that I didn't break anything)

## Review

- @contolini 
- @sebworks 
- @virginiacc

### DIFF
--- a/gulp/tasks/test.js
+++ b/gulp/tasks/test.js
@@ -81,7 +81,9 @@ function _getProtractorParams( suite ) {
 
   var commandLineParams = minimist( process.argv.slice( 2 ) );
 
-  var configFile = commandLineParams.a11y ? 'test/browser_tests/a11y_conf.js' : 'test/browser_tests/conf.js';
+  var configFile = commandLineParams.a11y ?
+                  'test/browser_tests/a11y_conf.js' :
+                  'test/browser_tests/conf.js';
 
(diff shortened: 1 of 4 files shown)


gulp/tasks/test.js
```diff
@@ -200,18 +212,19 @@ function testA11y() {
  * Run PageSpeed Insight tests.
  */
 function testPerf() {
-  _createPSITunnel( function( err, url, tunnel ) {
-    if ( err ) {
-      return plugins.util.log( err );
-    }
+  _createPSITunnel().then( function( params ) {
+    plugins.util.log( 'PSI tests checking URL: http://' + params.url );
```
contolini: `params.url` is an [array](https://github.com/cfpb/cfgov-refresh/pull/2334/files#diff-a8715b09d86dd13f31e1bb9e3ccc0f71R169) so use `params.url.join( ' ' )` here to get it to display a little nicer.

anselmbradford: It is done!

PR #2335, opened 2016-08-23
Hotfix: Spanish molecule translations

Scotchester:
We're launching Money as You Grow in Spanish on September 1st. The pages (unlike the existing English pages, which live in the fin-ed-resources repo) will be built in Wagtail. Consequently, they need the microcopy on a few choice modules to be in Spanish.

This needs to be a hotfix because the next regularly-scheduled release is Tuesday, September 6th.

## Additions

- Added support for Spanish-language cue labels to the Expandables organism.
- Added support for Spanish-language heading to the Social Media molecule.

## Testing

1. Pull branch
2. `gulp styles`
3. `runserver`
4. Create a new Browse page in your local Wagtail instance.
5. Add an Expandable to the main content area, and a Social Media module (in sharing mode) to the sidefoot.
6. Preview it.
7. Observe the cue labels on the Expandable and the heading on the Social Media module _en español_.

## Review

- @cfpb/cfgov-frontends 

## Screenshots

screen shot 2016-08-23 at 14 54 30

## Notes

- This is a quick-and-dirty, won't-scale-for-more-languages implementation. If someone is willing to help me figure out how to actually test for the language of the page in the template for an atomic module and output the text appropriately there, that would probably be the better route to go.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,17 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 
 ---------------------------------------
 
+## Unreleased
+
+### Added
+
+### Changed
+- Added support for Spanish-language cue labels to the Expandables organism.
+- Added support for Spanish-language heading to the Social Media molecule.
+
+### Removed
+
+
 ## 3.7.0
 
 ### Added
--- a/cfgov/jinja2/v1/_includes/molecules/social-media.html
+++ b/cfgov/jinja2/v1/_includes/molecules/social-media.html
@@ -44,6 +44,7 @@
             m-social-media__{{ 'share' if is_share_view else 'follow' }}">
     {% if is_share_view %}
         <div class="h5 m-social-media_heading">Share this</div>
+        <div class="h5 m-social-media_heading m-social-media_heading__es">Compartir este</div>
     {% endif %}
 
     <ul class="list
--- a/cfgov/jinja2/v1/_includes/organisms/expandable.html
+++ b/cfgov/jinja2/v1/_includes/organisms/expandable.html
@@ -48,11 +48,13 @@
                 <span class="o-expandable_cue
                              o-expandable_cue-open">
                     <span class="o-expandable_cue-label">Show</span>
+                    <span class="o-expandable_cue-label o-expandable_cue-label__es">Mostrar</span>
                     <span class="cf-icon cf-icon-plus-round"></span>
                 </span>
                 <span class="o-expandable_cue
                              o-expandable_cue-close">
                     <span class="o-expandable_cue-label">Hide</span>
+                    <span class="o-expandable_cue-label o-expandable_cue-label__es">Ocultar</span>
                     <span class="cf-icon cf-icon-minus-round"></span>
                 </span>
             </span>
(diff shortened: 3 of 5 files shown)


Scotchester:
You can proceed with testing, but I'll rebase my feature branch off of the 3.7 branch and force push shortly.

Scotchester:
Fixed.

Question: Should I be updating the change log to show "3.7.1" instead of "Unreleased"? And update the version in `package.json` accordingly?

willbarton:
@scranfill I believe so, but I don't think we're really documented the process of hotfixing.

@rosskarchner?

vccabral:
@Scotchester, what is the exact mechanism to do translations? Django supports internationalization natively but this appears to be a css/js approach? Just checking due to my ignorance on how you all have made your architectural decisions. 

https://docs.djangoproject.com/en/1.10/topics/i18n/translation/

sebworks:
yeah, can we reuse Djangos internationalization features?

sebworks:
This doesn't seem like it scales well because you end having to hardcode language support in CSS and HTML.

Scotchester:
@vccabral @sebworks That's what I was getting at with my note in the OP. Wagtail has some guidance on this, too: http://docs.wagtail.io/en/v1.5/advanced_topics/i18n/#creating-sites-with-multiple-languages

But it wasn't immediately clear to me how to leverage that to do what I need. I'd be happy to pair with someone on it if they can assist me.

As to how it works now, @vccabral If the page's language is set to Spanish in Wagatail (on its Configuration tab), it outputs `lang="es"` on L30 here: https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_layouts/content-base-sidebar-first.html#L29-L30.

rosskarchner:
@Scotchester @willbarton That's a good question--- maybe unreleased in this PR, and it'll be up to the person tagging the hotfix to edit the changelogs beforehand.

I think we would need a sprint to really get a handle on doing multiple languages in wagtail/django RIGHT.

There's a naive version that doesn't seem to complicated, if the molecule templates have access to the page object (do they?), then we could add some branching logic based on the language (which should be available in page.language)

Scotchester:
> There's a naive version that doesn't seem to complicated, if the molecule templates have access to the page object (do they?)

Fairly certain the answer to this is "no", which is why I assumed it was too complex for me to take on myself, but I'll give it a quick test.

Scotchester:
@rosskarchner Doesn't work immediately (`'page' is undefined`) but I see that there are other module templates that are accessing `page`.

Talking out of my butt here, but maybe it's that the `page` context is not passed down through the `render_stream_child` method?

sebworks:
This looks good on my end. Probably makes sense to have Jimmy or Kim look at this as well.

jimmynotjim:
👍

rosskarchner:
👍 I think this is fine for something needed next week, but we should get a real handle on localization soon.

Scotchester:
Ok, seems like this is mergeable for now.

@rosskarchner @willbarton Please let me know if you'd like me to do anything to ensure that this gets deployed before 9/1. (I may have one other thing to go in the hotfix deployment: cfpb/design-manual#439)

rosskarchner:
We'll create and deploy 3.7.2 tomorrow

PR #2337, opened 2016-08-24
Be less "clever" when setting BaseURL

rosskarchner:
The 'master' version of this fix should probably use environment variables (but default to localhost if the variable is unset)

### DIFF
--- a/cfgov/cal/views.py
+++ b/cfgov/cal/views.py
@@ -146,7 +146,7 @@ def pdf_response(request, context):
     stylesheet_url = '/static/css/pdfreactor-fonts.css'
     pdf_reactor = PDFreactor()
 
-    pdf_reactor.setBaseURL("%s://%s/" % (request.scheme, request.get_host()))
+    pdf_reactor.setBaseURL("http://localhost/")
     pdf_reactor.setLogLevel(PDFreactor.LOG_LEVEL_WARN)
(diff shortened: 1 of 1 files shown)


willbarton:
This looks okay, I'm not sure I understand what using localhost will do here though.

rosskarchner:
When PDF reactor needs to reach back to Django to retrieve assets (like CSS and images) , it uses the Base URL to resolve relative URL's

willbarton:
Ah, alright.

:+1:

anselmbradford:
Perhaps this is too clever, but would it ever not be `localhost`? Like should it use https://github.com/cfpb/cfgov-refresh/blob/master/.env_SAMPLE#L121 or equivalent?

willbarton:
@anselmbradford We've talked about moving pdfreactor off to its own server (or dockerization, or some way to isolate it). In that case, it would not be localhost.

rosskarchner:
I'm going to go ahead and merge this, since I made the change live on prod anyways

chosak:
> I'm going to go ahead and merge this, since I made the change live on prod anyways

I insist that we get T-shirts with this @willbarton.

willbarton:
:+1:

rosskarchner:
I've done that twice today, but I swear, it's never happened before!

PR #2339, opened 2016-08-24
Hotfix: Decouple leadership calendar PDF template from base.html

rosskarchner:
While further troubleshooting PDF's, I discovered that PDFReactor was really choking on main.css. Removing it made the difference between **painfully slow** (takes multiple minutes, sometimes times out) to reasonably zippy.

What I've done here, is take the template and remove it from the "base.html" inheritance chain. I've added the necessary CSS directly to the template. This isn't front-end code that we'll be proud of, but with this change, this feature is now usable.

## Testing

- Review this PDF generated with this code:  [cfpb-leadership (22).pdf](https://github.com/cfpb/cfgov-refresh/files/435530/cfpb-leadership.22.pdf)

- If you have a working PDF reactor, give it a try

(note that this PDF exposes a logic flaw in the application itself-- that event from March 14th should not appear there. This is not PDF-specific though)

## Review

- @cfpb/cfgov-backends 

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/about-us/the-bureau/leadership-calendar/print/index.html
+++ b/cfgov/jinja2/v1/about-us/the-bureau/leadership-calendar/print/index.html
@@ -1,23 +1,198 @@
-{% extends 'base.html' %}
+<html>
+<head>
+<title> Leadership calendar </title>
 
-{% block title -%}
-    Leadership calendar
-{%- endblock %}
 
-{% block desc -%}
-    Leadership calendar | Consumer Financial Protection Bureau
-{%- endblock %}
+<style type="text/css">
+@media print {
 
-{% block css %}
-    {{ super() }}
-    {% import 'pdfreactor-footer.html' as pdfreactor_footer %}
-    {{ pdfreactor_footer.render('consumerfinance.gov/about-us/the-bureau/leadership-calendar') }}
-{% endblock %}
+    @page {
+        size: letter;
+        padding: 0;
+        margin: 0.35in 0.35in 1.25in;
+        /*border: 1px solid red;*/
+        counter-increment: page;
+    }
 
-{% block javascript %}
-{% endblock %}
+    @page {
+        @bottom-center {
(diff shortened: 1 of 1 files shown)


richaagarwal:
I think this is fine if it speeds it up; but I am curious why it was choking on main.css? Are there things we can do there to improve its performance?

richaagarwal:
Might also be worth tagging the @cfpb/cfgov-frontends here in case they have more insight into `main.css` performance issues.

rosskarchner:
I'm not sure how to investigate further-- This PDFReactor version might just predate the trend of stuffing a lot of CSS into a big minified file, or maybe we're using CSS3 features that were cutting edge circa 2013. The logs are pretty useless.

willbarton:
Hmm. Could this be related to some of the performance issues that @virginiacc and @contolini have been working on?

willbarton:
As a hotfix for PDFReactor I'm ok with this though, pending front-end thoughts. 

I have not, ahem, tested it locally, however.

sebworks:
👍 , this is good.  We should have never been using `main.css`.

chosak:
When I run this using my local version of PDFReactor, the generated PDF uses different fonts, and I see messages like this in the output log:

> [150][09:20:40.516][WARNING] - Could not create font e (font-family: "AvenirNextLTW01-Regular"; src: _b (url: http://localhost/static/fonts/pdfreactor/46cf1067-688d-4aab-b0f7-bd942af6efd8.ttf, remote: true); bold: false; italic: false)
[150][09:20:40.517][WARNING] - Could not create font e (font-family: "AvenirNextLTW01-Demi"; src: _b (url: http://localhost/static/fonts/pdfreactor/a0f4c2f9-8a42-4786-ad00-fce42b57b148.ttf, remote: true); bold: true; italic: false)

I'm using version 6.3.6828.3 locally. I do have these fonts installed, so perhaps this is a Mac thing:

image

rosskarchner:
That's interesting-- but I don't think it should block us

rosskarchner:
I'm interpreting @willbarton  "pending front-end thoughts" and @sebworks 👍  as equalling 👍 👍  ;)

higs4281:
@chosak if PDFReactor depends on web fonts from origin http://www.consumerfinance.gov, they won't be available locally because of CORS

PR #2340, opened 2016-08-24
Job listing conversion script

chosak:
This PR adds a management command to convert from legacy Django `Job` model instances to tje new Wagtail `JobListingPage` pages created in #2314.

## Additions

- New management command, `manage.py convert_jobs_to_wagtail`. This command only converts jobs that are still open or closed within the last thirty days. The script requires the slug of the "parent" page for the new jobs pages.

## Testing

- In order to run this script, you'll need some `Job` model instances in your database that are less than a month old. These can be obtained from a production data dump.
- You can test in a "dry run" mode that creates the new pages but doesn't actually save them to the database:

 ```sh
python cfgov/manage.py convert_jobs_to_wagtail events --dry-run
using parent page Events
1/10 converting job Program Manager (Service Design Manager) CN-60/60 (program-manager-service-design-manager-cn-6060)
2/10 converting job Supervisory IT Specialist (Internet) CN-60/60 (supervisory-it-specialist-internet-cn-6060)
3/10 converting job Attorney-Advisor (Fair Lending Enforcement) CN-51,52,53/60 (attorney-advisor-fair-lending-enforcement-cn-51525)
4/10 converting job Attorney-Advisor (Fair Lending Enforcement) CN-71 (attorney-advisor-fair-lending-enforcement-cn-71)
5/10 converting job Stakeholder Support Team Lead CN-53/53 (stakeholder-support-team-lead-cn-5353)
6/10 converting job Senior Acquisition Support Analyst CN-53/53 (senior-acquisition-support-analyst-cn-5353)
7/10 converting job Personnel Research Psychologist CN-51/53 (personnel-research-psychologist-cn)
8/10 converting job Senior Policy Advisor, Fair Lending CN-60/60 (senior-policy-advisor-fair-lending-cn-6060)
9/10 converting job Acting Assistant Director, Consumer Lending, Collections & Reporting Markets CN-82 (acting-assistant-director-consumer-lending-collect)
10/10 converting job Reporting Analyst CN-51/52 (reporting-analyst-cn-5152)
```

- Or, remove the `--dry-run` parameter to actually create or update your local pages.

## Review

- @cfpb/cfgov-backends 

## Checklist

* [X] Changes are limited to a single goal (no scope creep)
* [X] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [X] Passes all existing automated tests
* [ ] New functions include new tests
[...]

PR #2275, opened 2016-08-04
Adding Jobs List Molecule

sebworks:
We currently don't have a molecule which matches the current Career page requirements, so I've created a jobs list molecule. This molecule will be used on `/about-us/careers/` page, within the `Sidebar-Breakout` organism.  

## Additions

- Adding Jobs List Molecule

## Testing

- N/A

## Review

- @chosak 
- @richaagarwal 
- @virginiacc 
- @ajbush 

## Screenshots

screen shot 2016-08-04 at 9 37 59 am

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #2341, opened 2016-08-24
Dynamic Wagtail blocks for career pages

chosak:
This PR adds some new Wagtail block types to allow content creators to add current job openings to pages, either in list or table form.

To maintain current (non-Wagtail) behavior, the [careers landing page](http://www.consumerfinance.gov/about-us/careers/) (a Sublanding Page) can be given a new `JobListingList` element. The [current openings page](http://www.consumerfinance.gov/about-us/careers/current-openings/) (a Browse Page) can be given a new `JobListingTable` element.

## Additions

- New `jobmanager.models.JobListingList` and `jobmanager.models.JobListingTable` Wagtail block types for dynamic display of currently open jobs.
- New `v1.atomic_elements.organisms.ModelBlock`, `.ModelList`, and `.ModelTable` Wagtail generic block types as supporting elements for the above, and for potential future use in other apps.
- New tests around all new components.
- Pulled forward the very useful [`IntegerBlock`](http://docs.wagtail.io/en/v1.6/topics/streamfield.html#integerblock) from Wagtail 1.6, until we upgrade.

## Changes

- Moved @sebworks `cfgov/jinja2/v1/_includes/molecules/jobs-list.html` template from #2275 to `cfgov/jinja2/v1/_includes/organisms/job-listing-list.html` to be more consistent with other files -- please advise if it'd be better to structure otherwise.
- Minor change to `v1.util.util. get_secondary_nav_items` logic that was preventing  `BrowsePage` preview.

## Testing

- Testing this code requires that you have some `JobListingPage` pages in your Wagtail DB. These can be created manually or using the conversion script in #2340.

## Review

- @cfpb/cfgov-backends @sebworks @schaferjh 

#### Sublanding pages have a new "Job listing list" option
image
#### Specifying options for the jobs list
image
#### Jobs list with no openings
image
#### Jobs list with openings
image
#### Browse pages have a new "Job listing table" option
image
#### Specifying options for the jobs table
image
#### Jobs table with no openings
image
#### Jobs table with openings
image

## Notes

- This PR is slightly complex mainly because Wagtail does not have a simple way to display dynamic content (e.g. content that gets determined at page view time, not at edit time). Specifically, we want to be able to render jobs lists based on the attributes of the pages in the database. We also want to be able to do this from multiple pages (as opposed to a single parent > filterable children approach).
- Hopefully breaking it out into the generic `ModelBlock`, `ModelList`, and `ModelTable` widgets will be useful for other projects in future, as this should make it easy to do something like "create a block that shows all Foo items in the DB". See the tests for various examples of this.

## Todos

- @sebworks the jobs list table doesn't highlight rows as in [the current page](http://www.consumerfinance.gov/about-us/careers/current-openings/), even if I set the new `row-links` property. Any idea why?
- @richaagarwal @sebworks I'm not sure if it would be ideal to try to move to the new `TableBlock` work you did. Perhaps this should be additional work in future?

## Checklist


cfgov/cfgov/test.py
```diff
@@ -28,3 +29,15 @@ def setup_wagtail_root(self):
 
         module = importlib.import_module(migration)
         module.initial_data(apps, None)
+
+
+class HtmlMixin(object):
```
chosak: Created this useful mixin for use in unit tests where you are comparing generated HTML.

cfgov/v1/tests/atomic_elements/organisms/test_organisms.py
```diff
@@ -1,15 +1,14 @@
-from django.test import TestCase
```
chosak: The only changes to this file are whitespace/PEP8 things.

cfgov/v1/util/util.py
```diff
@@ -68,6 +68,9 @@ def get_secondary_nav_items(request, current_page):
     else:
         page = get_appropriate_page_version(request, current_page)
 
+    if not page:
```
chosak: This logic change is necessary because when previewing a `BrowsePage` the value of `page` is always `None`.

richaagarwal: Oh, interesting, do we know why that's the case?

chosak: I don't know whether this has ever worked to preview new BrowsePages, although it may work when previewing edited ones. In the code there's a reference to `current_page.get_parent()` which will always be `None` when previewing a new page.

schaferjh: Total interloper so ignore if this is irrelevant: just recently (seems to be an impact of the latest Wagtail update) we regained the ability to preview new pages before they had been shared/saved. For awhile that was a minor annoyance.

sebworks:
> I'm not sure if it would be ideal to try to move to the new TableBlock work you did. Perhaps this should be additional work in future?

Addressed in https://github.com/cfpb/cfgov-refresh/pull/2344

sebworks:
Does this PR require migrations?

chosak:
@sebworks it shouldn't require migrations because it doesn't alter any existing pages.

Regarding #2344, there will still need additional work to move over to `TableBlock` because I use `Table` internally [here](https://github.com/cfpb/cfgov-refresh/pull/2341/files#diff-dcf2a8703e945442ba8591d07085a645R199) to render the table dynamically.

sebworks:
> Regarding #2344, there will still need additional work to move over to TableBlock because I use Table internally here to render the table dynamically.

Yep, totally dude.

sebworks:
@chosak , would it be possible to add some help text explaining that the data for the block is dynamic. This would help content editors know that there is no data to enter.

sebworks:
should the jobs table automatically link to the job detail page?

chosak:
@sebworks thanks, the help text is a good suggestion. It's tricky to add help text to the entire block but I can add it to each sub-block. What do you think of these, added in c443605:

image
image


As far as the table linking, right now I generate something like this:

image

I'd like the whole row to link (and highlight) [like the current page](http://www.consumerfinance.gov/about-us/careers/current-openings/) but I'm not quite sure what I'm missing. I have a checkbox for `row-links` -- is this not the correct setting?

sebworks:
Sometimes I'm getting a `maximum recursion depth exceeded while calling a Python` error when previewing a `Browse` page with a Job listing table. I'm not sure it's related to this PR though.

chosak:
@sebworks hmm, interesting, I haven't seen that when previewing before. Can you share more of the full stack trace?

sebworks:
👍

sebworks:
> @sebworks hmm, interesting, I haven't seen that when previewing before. Can you share more of the full stack trace?

This was a local issue that we resolved in chat.

chosak:
Added one additional change in eaa4979 that moves the `JobListingList` into the sidebar of `SublandingPage` instead of in the content section, to be consistent with how this appears [today](http://www.consumerfinance.gov/about-us/careers/).

sebworks:
> Added one additional change in eaa4979 that moves the JobListingList into the sidebar of SublandingPage instead of in the content section, to be consistent with how this appears today.

@chosak thanks, 💯

richaagarwal:
@chosak Overall this looks really solid 👍  If you could add some light documentation to a few of the classes and functions that aren't self explanatory, that would be awesome

chosak:
@richaagarwal thanks for the review. I added docstrings to the new abstract block types in 713a2f5.

richaagarwal:
Thanks!  👍

PR #2342, opened 2016-08-25
Alphabetize author names by default

richaagarwal:
Previously, author names for pages would get displayed in a seemingly random order (as they are entered in as tags, which are not orderable).  This PR ensures they are displayed in alphabetical order (by last name) by default. 

## Testing 
- To test, you can view this page (if you have prod data locally) [http://localhost:8000/about-us/blog/eregulations-updated-regulations-x-c-and-more/](http://localhost:8000/about-us/blog/eregulations-updated-regulations-x-c-and-more/ ) and notice that on this branch, it displays "Andrea Edmonds , Adam Scott and Dan Sokolov" in the correct order, but on master / prod it displays out of order. http://www.consumerfinance.gov/about-us/blog/eregulations-updated-regulations-x-c-and-more/
- You can also click through to the preview page [http://localhost:8000/about-us/blog/?page=3#o-filterable-list-controls-1](http://localhost:8000/about-us/blog/?page=3#o-filterable-list-controls-1) to verify it displays correctly there too.
- If you do not have prod data locally, you can test by creating a page and adding authors to it. You can also just run `tox -e fast` :)

## Notes 
As part of this PR, I also did some clean-up, as I noticed two functions weren't being used at all. 

## Assignees
- @cfpb/cfgov-backends 
- REQUEST: Please assign yourself if you are planning to review, so that others are aware. Thanks!


cfgov/v1/models/base.py
```diff
@@ -131,6 +131,14 @@ class CFGOVPage(Page):
         ObjectList(settings_panels, heading='Configuration'),
     ])
 
+    def get_authors(self):
```
richaagarwal: I realize this function isn't totally necessary but it was a design decision, in case we ever want to add sort_by parameters or another default.

cfgov/v1/tests/test_author_names.py
```diff
@@ -0,0 +1,25 @@
+from django.test import TestCase
+from v1.models.base import CFGOVPage
+
+class TestAuthorNames(TestCase):
+
+    def test_names_get_sorted_alphabetically(self):
```
richaagarwal: It is intentional that this tests `get_authors` and not `alphabetize_authors` as `get_authors` is the function called by the templates. If we ever decide to go with another sort order, this test will fail, which I think is fine.

chosak:
Looks great! :+1:

Minor question: do you think that `v1/tests/test_author_names.py` could go instead under `v1/tests/models/` to keep module organization in sync between code and test code?

cfgov/v1/models/base.py
```diff
@@ -131,6 +131,14 @@ class CFGOVPage(Page):
...
+        """ Returns a sorted list of authors. Default is alphabetical """
+        return self.alphabetize_authors()
+
+    def alphabetize_authors(self):
+        """ Alphabetize authors of this page by last name """ 
+        return sorted(self.authors.names(), key=lambda x: x.split(" ")[-1])
```
vickumar1981: i think space is the default separator for split.

cfgov/v1/tests/test_author_names.py
```diff
@@ -0,0 +1,25 @@
...
+        page = CFGOVPage()
+        page.authors.add('Ross Karchner','Richa Agarwal', 'Andy Chosak', 'Will Barton')
+        expected_result = ['Richa Agarwal', 'Will Barton', 'Andy Chosak', 'Ross Karchner']
+        self.assertEquals(page.get_authors(), expected_result)
+    
+    def test_no_authors(self):
+        page = CFGOVPage()
+        self.assertEquals(page.alphabetize_authors(), [])
```
vickumar1981: i understand that get_authors() calls alphabetize_authors() anyway, but shouldn't this call the get_authors() wrapper like you did on line 10?  that way if you change the implementation of how you alphabetize things, you won't have to change this line in your tests.  same on line 20.

cfgov/v1/models/base.py
chosak: Interesting. Apparently the behavior [isn't quite identical](https://docs.python.org/2/library/stdtypes.html#str.split). Just using `x.split()` would correctly handle a case where someone accidentally puts an extra space at the end:

```py
>>> 'Donald Duck '.split()[-1]
'Duck'
>>> 'Donald Duck '.split(' ')[-1]
''
```

Although, it appears that Taggit correctly strips any extra spaces (try entering `" Donald  Duck "` into the authors list and you'll see it correctly converts it to `"Donald Duck"`.

vickumar1981: yeah, that's a good point.  so perhaps use a trim() before the split.

also, i don't think we're accounting for if the last names are the same.  i.e., i believe there should be a sort ordering - last name, then first name.

vickumar1981: another edge case where i can see this logic wouldn't hold up correctly is the case where there is no space.

```
'Duck'.split()[-1]
'k'
```

That would sort on 'k', not 'Duck', which i'm not sure is the intended behavior.

richaagarwal: Thanks @vickumar1981 -  I'll update it to use `x.split()`. 

We can expect the names in a certain format, so I'm constraining the edge cases a little bit.  I'm not sure how important it is to order by first name when last names are identical, but that would be another specification; that's not an edge case. If the first names were also identical, then would we sort on middle name if there is one?  I guess I feel like we could overengineer these cases that aren't that common @schaferjh Want to chime in here?

richaagarwal:
@chosak Yes, I can reorganize the test file to be under `models`. Thanks for pointing that out!

cfgov/v1/tests/test_author_names.py
richaagarwal: @vickumar1981 Yeah, I went back and forth on this a few times.  I wanted a function to reflect that things do in fact get alphabetized when calling `get_authors()`, but otherwise the unit tests really are for `alphabetize_authors()` which has the logic in it.  Perhaps it will make more sense to have a test (in addition to these) that simply asserts `alphabetize_authors()` is called when `get_authors()` is?

cfgov/v1/models/base.py
schaferjh: I don't think it's necessary to get too fine grained about the edge cases. Ordering by first name when last names are identical sounds good. But more than that feels unnecessary at this point.

richaagarwal:
I added support for cases where the last names are the same, per @vickumar1981's suggestion, as well as addressed some other comments. Ready for a final review!

willbarton:
Tested and working locally for me!

PR #2343, opened 2016-08-25
Downgrade djangorestframework

rosskarchner:
I think this upgrade was erroneous (but understandable, the way our requirements have been split among multiple repo's), but I want to make sure @grapesmoker didn't have a specific reason for upgrading it.

## Changes

- change base.txt to reference djangorestframework==2.4.8

## Review

- @cfpb/cfgov-backends

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -20,7 +20,7 @@ django-taggit==0.18.3
 django-tinymce==2.2.0
 django-treebeard==3.0
 django>=1.8,<=1.8.99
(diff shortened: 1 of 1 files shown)


grapesmoker:
Yeah, this might have just been my bad. I thought I had the right version but perhaps what happened is that I accidentally upgraded it before running pip freeze.

grapesmoker:
👍

rosskarchner:
Thanks for confirming!

PR #2344, opened 2016-08-25
Introducing TableBlocks! (again!)

sebworks:
## Overview 
(from @richaagarwal)
- I'm editing this to reflect the entirety of the work done on introducing TableBlocks since we had to revert a previous PR (https://github.com/cfpb/cfgov-refresh/pull/2295). That PR was reverted since it did not include necessary front-end work that @sebworks has introduced here (thank you!). Also shout-out to @chosak for pairing with me on some tricky back-end issues & adding other improvements here.
- This PR encompasses the front-end and back-end work needed to integrate Wagtail's TableBlocks into our use of tables. Our old `Table` organism will no longer be used, however the field cannot be removed until the migrations in this PR have run on production. Once this is merged in, I can create a clean-up task for that. 
- There is a lot here. I tried to squash some commits, however, due to some weirdness with how reverting PRs work, git really really wants to erase the work I did to reintroduce it, so I feel comfortable saying this is the best we can do.

### Testing the editor (this is new!)
- Create a Browse page.
- Add a table to the page.
- Configure the table with the options of your choosing.
- Verify that the table has the options you chose earlier. 

### Testing that migrations did what they're supposed to 
(this was theoretically reviewed in https://github.com/cfpb/cfgov-refresh/pull/2295)
- Grab a prod dump
- Run migrations, ~~if you have the most recent prod dump you will need to run `python cfgov/manage.py migrate --fake jobmanager 0002` before `python cfgov/manage.py migrate` (unrelated to this PR, but FYI)~~
- Check the console for names of pages that were updated
- Go to one of those pages & see that the table data is still intact

## Review

- @chosak 
- @rosskarchner 
- @contolini  
- @schaferjh 

## Screenshots
screen shot 2016-08-25 at 1 12 30 pm
screen shot 2016-08-25 at 11 58 27 am

## Notes

- I didn't add all the options that were available in the design manual. If we still wish to do so then I can do that.

- There are some scenarios where the options are related (first row is header vs first column is header), it might make sense to use radio buttons in those place. 

- At some point I will open a PR against Wagtail to fix issues with the Hasontable bridge.

## Checklist


cfgov/v1/wagtail_hooks.py
```diff
@@ -6,7 +6,7 @@
 from django.conf import settings
 from django.http import Http404
 from django.contrib.auth.models import Permission
-from django.utils.html import escape
+from django.utils.html import escape, format_html, format_html_join
```
chosak: Can you please remove unused import `format_html`?

cfgov/v1/wagtail_hooks.py
```diff
@@ -49,6 +49,17 @@ def share(page, is_sharing, is_live):
     page.save()
 
 
+@hooks.register('insert_editor_js')
```
chosak: Can you please use an indentation level of 4 spaces here to be consistent with the rest of the code and [PEP8](https://www.python.org/dev/peps/pep-0008/#indentation)?

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -141,6 +149,32 @@ class Meta:
         template = '_includes/organisms/table.html'
         label = ' '
 
+class AtomicTableInput(TableInput):
+    _id = 'table-block'
+    def render(self, name, value, attrs=None):
+        attrs.update({'id': self._id})
+        # Calling the grandparents render method and bypassing TableInputs,
+        # in order to control how we render the form.
```
chosak: I notice that inheriting from `TableInput` as you do doesn't actually add much since you're overloading both `render` and `render_js_init`. Does it make sense or simplify to just inherit directly from `WidgetFromScript` and `forms.HiddenInput` as `TableInput` does [here](https://github.com/torchbox/wagtail/blob/v1.5.3/wagtail/contrib/table_block/blocks.py#L14)?

chosak:
I notice that you copied `table-block.js` and `table-input.html` from Wagtail `master` as opposed to using a specific tag like [1.5.3](https://github.com/torchbox/wagtail/tree/v1.5.3/wagtail/contrib/table_block). For use in keeping this up to date in future, could you add least add the commit ID to the headers in those files? That way we'd be able to do a diff in the Wagtail code to see if/when there are additional changes we'd want to pull in.

chosak:
I notice that setting the first column as header does correctly use `<th>` instead of `<td>` but the styling is the same. Is this expected?

Do we need anything additional to support link highlighting as in [the careers table](http://www.consumerfinance.gov/about-us/careers/current-openings/)? Is this what is referred to as the "browse-links" option?

chosak:
Besides my minor comments, :+1:. This is awesome.

schaferjh:
@sebworks could you clarify what you mean by "I didn't add all the options that were available in the design manual. If we still wish to do so then I can do that." ? What aspect of our DM table styling are missing?

sebworks:
> @sebworks could you clarify what you mean by "I didn't add all the options that were available in the design manual. If we still wish to do so then I can do that." ? What aspect of our DM table styling are missing?

I've ommited the following table modifications ( We can add these at any time because they don't require a migration ) : 
`.table_cell__right-align`, `table__sortable`, `table__entry-header-on-small`,

schaferjh:
> I've ommited the following table modifications ( We can add these at any time because they don't require a migration ) : 
.table_cell__right-align, table__sortable, table__entry-header-on-small.

I think it sounds ok to omit these for now, but just to make sure that I understand the implications:
- right-align is the option to right-align content in a given column
- sortable is the option to sort table by a given column
- entry header on small --> what is this?

cfgov/v1/atomic_elements/organisms.py
sebworks: If I do that I would have to create an `init` method to assign the `table_options` value.

cfgov/v1/wagtail_hooks.py
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2344/commits/362e1236b35477c0095fd8ff5955e2e504c4c342.

cfgov/v1/wagtail_hooks.py
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2344/commits/362e1236b35477c0095fd8ff5955e2e504c4c342

sebworks:
> entry header on small --> what is this?

Just means we want to use the first column as a header instead of the data-label on mobile.

sebworks:
ok, i've added the capability to edit the table using the RichTextEditor. It's hackier than a Netflix horror movie so please test thoroughly.

screen shot 2016-08-31 at 9 08 29 am

willbarton:
> hackier than a Netflix horror movie

T-shirt?

chosak:
I observed two minor issues with the RichTextEditor:

1. If you click to edit a cell, then choose the "link" icon to select another Wagtail Page, but then click the "X" in the upper right to cancel selecting, you're left with a disabled editor and no way to get back to your page without reloading.

2. If you leave a cell blank, it gets rendered as "None" - with the old `Table` it would get left blank:
 image
 image

richaagarwal:
I observed the same thing as @chosak, but it happens any time I exit out of an option.

sebworks:
> I observed the same thing as @chosak, but it happens any time I exit out of an option.

Working on a fix

chosak:
I pushed some changes that modify `TableBlock` to use the new `ModelTable` internally instead of `Table`. This fixes the unit tests that broke Travis above, and adds a whole bunch of new unit tests.

richaagarwal:
@chosak Did you mean to push those to this PR?

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -264,31 +264,58 @@ def make_created_value(self, instance, value):
     fields = None
     field_headers = None
 
-    row_links = blocks.BooleanBlock(
+    first_row_is_table_header = blocks.BooleanBlock(
```
richaagarwal: I'm confused as to why this is necessary

chosak:
@richaagarwal I did. I think what happened is this:

1. We had a `Table` element that used a `table.html` template.
1. I created a `ModelTable` organism that uses a `Table` internally, and has unit tests that verifies the way that it renders.
1. We added the Wagtail `TableBlock` element that has its own template.
1. In this PR, @sebworks created a new `AtomicTable` organism that kind of derives from `TableBlock`, and also modifies `table.html`.
 1. This broke my unit tests, because the `Table` no longer rendered properly.
 1. @sebworks won't this break any legacy pages that use `Table`? Or is the assumption that these will all have already been modified?
1. My changes update `ModelTable` to use `AtomicTable` internally (which it would need to do anyway at some point).

cfgov/v1/atomic_elements/organisms.py
chosak: @richaagarwal I agree this is somewhat suboptimal. The way that `ModelTable` works is that it kind of wraps a Wagtail table block (formerly our `Table`, with this PR our `AtomicTable`). It would be nice to be able to use inheritance to avoid this, and I would like to revisit in future to see if this is possible.

For now, what these elements do is add the various `AtomicTable` options to the `ModelTable`:

image

It'd be better not to have this duplication, but without adding these options there'd be no way to set them on the table.

richaagarwal: @chosak Ah, thanks for the context!

richaagarwal:
I went ahead and merged my branch tableblock-v2 in, since this PR won't work without it.

richaagarwal:
Adding the hold label to this since the html editor needs to be fixed before this is ready for further review.

sebworks:
@chosak, @richaagarwal, please review when you get a chance.

richaagarwal:
@sebworks Will review now!

richaagarwal:
@sebworks I am unable to run migrations on this branch.  Did you mean to change the reference on `0007_add_tableblocks_to_pages` ?  Also, could you rename any migrations you've added to be descriptive?

richaagarwal:
@sebworks I'm going to try to clean up the migrations here.

cfgov/cfgov/settings/base.py
```diff
@@ -197,7 +198,8 @@
 # Used to include directories not traditionally found,
 # app-specific 'static' directories.
 STATICFILES_DIRS = [
-    PROJECT_ROOT.child('static_built')
+    PROJECT_ROOT.child('static_built'),
+    PROJECT_ROOT.child('templates', 'wagtailadmin')
```
sebworks: > Why add the wagtail admin templates to the static dirs?

We are adding Javascript and CSS to extend the Wagtail admin. Wasn't sure if there was a more appropriate place to place the assets.

richaagarwal:
@rosskarchner What are your mixed feelings about hardcoding the Django version? I thought we wanted to start pinning exact versions of requirements anyway?

@contolini or @virginiacc could one of you review the front-end work here?

willbarton:
@richaagarwal We shouldn't have to worry about Django 1.8.x introducing any breaking changes, and if we pin, we have to start being more aware of bug/security updates to 1.8. Using `django>=1.8,<=1.8.99` would mean we're at most two weeks behind any new 1.8.x releases with our current deployment cycle.

richaagarwal:
@willbarton That makes sense; locally I had actually been on an older version than I pinned (which wasn't working with my changes) & so @chosak and I thought it would make sense to match what's in production.  I am certainly not opposed to upgrading it even more, but I do think it'd be nice to have a way of the dev environment matching the prod environment.

chosak:
@willbarton just to add more context, @richaagarwal and I ran into an issue where some code worked with 1.8.14 but not with 1.8.8, due to minor changes between those revisions (something to do with the last bullet [here](https://docs.djangoproject.com/en/1.10/releases/1.8.10/#bugfixes) and how we do migrations with Wagtail pages).

Could we change this to `django>=1.8.14,<=1.8.99`?

rosskarchner:
> @rosskarchner What are your mixed feelings about hardcoding the Django version? I thought we wanted to start pinning exact versions of requirements anyway?

Basically what @willbarton-- I could see both sides of it (automatically getting security updates is nice, but it's at least philosophically opposed to our earlier push to Pin All the Things)

chosak:
Due to the new migrations in PR #2386 which was just merged, the migrations in this PR need to be renamed. I'll rename them and fix the merge conflict.

willbarton:
@chosak I don't feel too strongly one way or the other, I think there are pros and cons to pinning a range and an exact version.

> I do think it'd be nice to have a way for the dev environment to match the prod environment.

But this sells me on doing it this way for now. No matter what, this is absolutely not a blocker to getting this in.

cfgov/jinja2/v1/_includes/organisms/table.html
```diff
@@ -12,35 +12,40 @@
...
-                        <td data-label="{{ thead }}">
-                            {{ parse_links(cell.value) | safe }}
-                        </td>
-                    {% else %}
-                        <td data-label="{{ thead }}">
-                            {{ cell | safe }}
-                        </td>
-                    {% endif %}
+            {% for cell in row or [] %}
+                {% if cell is not none %}
```
contolini: This fixes @chosak's [comment](https://github.com/cfpb/cfgov-refresh/pull/2344#issuecomment-243822904) but now if you leave a cell empty, it's removed and the cells to the right are shifted to the left. See screenshots below. Would it be possible to insert `&nbsp;` or some sort of invisible placeholder to preserve empty cells?

screen shot 2016-09-21 at 4 54 45 pm

screen shot 2016-09-21 at 4 54 21 pm

contolini: In those screenshots: s/second row/second column/

cfgov/jinja2/v1/_includes/organisms/table.html
```diff
@@ -12,35 +12,40 @@
...
+    {% set row_index = 0 %}
+    {% if value.first_row_is_table_header and value.data | list | length > 0 %}
+    {% set row_index = 1 %}
     <thead>
         <tr>
-            {% for thead in value.headers %}
-                <th>{{ thead }}</th>
+            {% for column in value.data[0] %}
+                {% if cell is not none %}
+                <th>
```
contolini: Add `scope="col"` to these `th` tags. It'll [help screenreaders](http://webaim.org/techniques/tables/data#scope).

cfgov/jinja2/v1/_includes/organisms/table.html
```diff
@@ -12,35 +12,40 @@
...
-                        </td>
-                    {% else %}
-                        <td data-label="{{ thead }}">
-                            {{ cell | safe }}
-                        </td>
-                    {% endif %}
+            {% for cell in row or [] %}
+                {% if cell is not none %}
+                {% set tag = 'th' if value.first_col_is_header and loop.first else 'td' %}
+                <{{tag}} data-label="{{ value.data[0][loop.index0] | striptags }}">
```
contolini: Also, if they want the first column as a header (`value.first_col_is_header`), add `scope="row"` to these `th` tags.

contolini:
Killer work! 💯

chosak:
Assigning this to myself to fix careers table-related test failures.

cfgov/jinja2/v1/_includes/organisms/table.html
```diff
@@ -12,35 +12,44 @@
...
-                        </td>
-                    {% else %}
-                        <td data-label="{{ thead }}">
-                            {{ cell | safe }}
-                        </td>
-                    {% endif %}
+            {% for cell in row or [] %}
+                {% set tag = 'th scope="row"' if value.first_col_is_header and loop.first else 'td' %}
+                {% if value.data[0][loop.index0] is not none %}
+                   {% set data_label = 'data-label="{0}"'.format(value.data[0][loop.index0] | striptags | trim) %}
```
virginiacc: Should the data-label only be added if `value.first_row_is_table_header`? Otherwise, the first row is used as a header and also displayed when the table stacks on mobile:

screen shot 2016-09-22 at 2 47 30 pm
screen shot 2016-09-22 at 2 47 45 pm

virginiacc: Also, it looks like the data labels are displayed even if the table doesn't stack on mobile:

screen shot 2016-09-22 at 3 07 19 pm

screen shot 2016-09-22 at 3 07 29 pm

because of [this rule in enhancements/tables.less](https://github.com/cfpb/cfgov-refresh/blob/1db1f16329a030de487d7e3d61b038fb0a56b4a2/cfgov/unprocessed/css/enhancements/tables.less#L134-L142).

sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2344/commits/64648278d5c1c164a872c0cbb94beec3bbb6ff5b.

sebworks: > Should the data-label only be added if value.first_row_is_table_header? Otherwise, the first row is used as a header and also displayed when the table stacks on mobile:

I've modified the code to only add the `data-labels` when the first row is present and column header isn't selected.

chosak:
After final review, this looks good to me. Tested on a recent production data dump and tables still look good.

PR #2345, opened 2016-08-26
FIX: Properly convert hyperlink data to new table rows

richaagarwal:
This should've been a part of the TableBlocks PR https://github.com/cfpb/cfgov-refresh/pull/2295 - an oversight on my part.  

## Test

- To see the problem, visit http://build.consumerfinance.gov/data-research/credit-card-data/college-credit-card-agreement/ and notice the rows are missing data.  
- To see the fix, re-import prod data (`./refresh-data.sh production.sql`) & run migrations (`python cfgov/manage.py migrate --fake-initial`) & navigate to [http://localhost:8000/data-research/credit-card-data/college-credit-card-agreement/](http://localhost:8000/data-research/credit-card-data/college-credit-card-agreement/ ) to verify the correct data is there.

## Note
- I'm updating the migration itself as opposed to adding a new one since this hasn't been run on prod yet, and we don't really want to run the previous migration on prod data if we can help it. If you have strong objections to that, discuss! 

## Assignee
- @cfpb/cfgov-backends 
- @sebworks

### DIFF
--- a/cfgov/v1/migrations/0006_update_tables.py
+++ b/cfgov/v1/migrations/0006_update_tables.py
@@ -1,15 +1,10 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
+from __future__ import print_function
+from itertools import chain
 
-from django.db import migrations, models
-import v1.models.snippets
-import wagtail.contrib.table_block.blocks
-import wagtail.wagtailcore.fields
-import wagtail.wagtailcore.blocks
-import wagtail.wagtailsnippets.blocks
-import wagtail.wagtailimages.blocks
-from wagtail.wagtailcore.models import Site
-from django.conf import settings
+from django.db import migrations
+
 from v1.models.learn_page import LearnPage, DocumentDetailPage
 from v1.models.sublanding_page import SublandingPage
 from v1.models.sublanding_filterable_page import SublandingFilterablePage
@@ -18,25 +13,43 @@ from v1.models.blog_page import BlogPage
 from v1.models.browse_filterable_page import BrowseFilterablePage
 from v1.models.demo import DemoPage
 from v1.tests.wagtail_pages.helpers import publish_changes
-from v1.models.base import CFGOVPage
-from itertools import chain
-
-def update_content_tables(wagtail_page):
-    for item in wagtail_page.content.stream_data:
-        if item['type'] == 'table':
-            item['type'] = 'table_block'
-            item['value'] = create_tableblock_data(old_table=item['value'])
-            print "Creating a new TableBlock on page %s" %(wagtail_page.title)
 
 
-def update_full_width_text_tables(wagtail_page):
-    for item in wagtail_page.content.stream_data:
-        if item['type'] == 'full_width_text':
-            for sub_item in item['value']:
-                if sub_item['type'] == 'table':
-                    sub_item['type'] = 'table_block'
-                    sub_item['value'] = create_tableblock_data(old_table=sub_item['value'])
-                    print "Creating a new TableBlock on page %s" %(wagtail_page.title)
+def update_table_items(items, wagtail_page):
+    for item in list(filter(
+        lambda item: item['type'] == 'table',
+        items
+    )):
+        item['type'] = 'table_block'
+        item['value'] = create_tableblock_data(old_table=item['value'])
+        print(
+            "Updating Table to a TableBlock on page %s of page type %s"
+            % (wagtail_page.title, wagtail_page.__class__.__name__)
+        )
+
+
+def update_tables_in_content_field(wagtail_page):
+    update_table_items(
+        items=wagtail_page.content.stream_data,
+        wagtail_page=wagtail_page
+    )
+
+
+def update_tables_in_full_width_text_organisms(wagtail_page):
+    for item in list(filter(
+            lambda item: item['type'] == 'full_width_text',
+            wagtail_page.content.stream_data
+    )):
+        sub_items = item['value']
+        update_table_items(items=sub_items, wagtail_page=wagtail_page)
+
+
+def convert_hyperlink_obj_to_text(value):
+    """ Taken from cfgov/jinja2/v1/_includes/atoms/hyperlink.html """
+    return """<a class='jump-link' href='%s'>%s</a>""" % (
+        value['url'],
+        value['text']
+    )
 
 
 def create_tableblock_data(old_table):
@@ -49,21 +62,25 @@ def create_tableblock_data(old_table):
         for row_list in old_table['rows']:
             new_row = []
             for row in row_list:
+                if row['type'] == 'hyperlink':
+                    row['value'] = convert_hyperlink_obj_to_text(row['value'])
                 new_row.append(row['value'])
             table['data'].append(new_row)
     return table
 
 
-def create_tableblocks_for_every_table(apps,schema_editor):
+def create_tableblocks_for_every_table(apps, schema_editor):
+    print("Updating tables in content field")
     for p in chain(
-        BrowsePage.objects.all(), 
-        SublandingPage.objects.all(), 
+        BrowsePage.objects.all(),
+        SublandingPage.objects.all(),
         LearnPage.objects.all(),
         DocumentDetailPage.objects.all(),
     ):
-        update_content_tables(wagtail_page=p)
+        update_tables_in_content_field(wagtail_page=p)
         publish_changes(child=p)
 
+    print("Updating tables in full width text organisms")
     for p in chain(
         BlogPage.objects.all(),
         BrowseFilterablePage.objects.all(),
@@ -73,9 +90,8 @@ def create_tableblocks_for_every_table(apps,schema_editor):
         SublandingFilterablePage.objects.all(),
         SublandingPage.objects.all(),
     ):
-        update_full_width_text_tables(wagtail_page=p)
+        update_tables_in_full_width_text_organisms(wagtail_page=p)
         publish_changes(child=p)
-
 
 
 class Migration(migrations.Migration):


chosak:
If I understand what this migration is doing, it's equivalent to actually entering this in the Wagtail editor:

image

Which, unfortunately, just changes to this when you click out of the cell:

image

But then does actually put the right content in the page:

image

Will content editors who want to create tables using TableBlock have to actually enter the HTML tags for links when editing?

chosak:
Recognizing that editing of HTML is suboptimal but not a blocker so 👍

cfgov/v1/migrations/0006_update_tables.py
```diff
@@ -49,6 +52,8 @@ def create_tableblock_data(old_table):
         for row_list in old_table['rows']:
             new_row = []
             for row in row_list:
+                if row['type'] == 'hyperlink':
```
vickumar1981: i think this can be fewer lines with a list comprehension.

cfgov/v1/migrations/0006_update_tables.py
```diff
@@ -26,7 +26,7 @@ def update_content_tables(wagtail_page):
         if item['type'] == 'table':
             item['type'] = 'table_block'
             item['value'] = create_tableblock_data(old_table=item['value'])
-            print "Creating a new TableBlock on page %s" %(wagtail_page.title)
+            print "Creating a new TableBlock on page %s of page type %s" %(wagtail_page.title, wagtail_page.__class__.__name__)
```
vickumar1981: is there a reason not to use logger here?

cfgov/v1/migrations/0006_update_tables.py
```diff
@@ -36,8 +36,11 @@ def update_full_width_text_tables(wagtail_page):
...
                     sub_item['type'] = 'table_block'
                     sub_item['value'] = create_tableblock_data(old_table=sub_item['value'])
-                    print "Creating a new TableBlock on page %s" %(wagtail_page.title)
+                    print "Creating a new TableBlock on page %s of page type %s" %(wagtail_page.title, wagtail_page.__class__.__name__)
 
+def convert_hyperlink_obj_to_text(value):
+    """ Taken from cfgov/jinja2/v1/_includes/atoms/hyperlink.html """
+    return """<a class='jump-link' href='%s'>%s</a>""" %(value['url'], value['text'])
 
 def create_tableblock_data(old_table):
```
vccabral: return {
     u'data': [
          [
              convert_hyperlink_obj_to_text(row['value'])
              for row in row_list 
              if row['type'] == 'hyperlink'
          ]
          for row_list in old_table['rows']
     ]
}

vccabral: @richaagarwal I would suggest the comprehension use since you are only doing a transform.

cfgov/v1/migrations/0006_update_tables.py
```diff
@@ -26,7 +26,7 @@ def update_content_tables(wagtail_page):
         if item['type'] == 'table':
             item['type'] = 'table_block'
             item['value'] = create_tableblock_data(old_table=item['value'])
-            print "Creating a new TableBlock on page %s" %(wagtail_page.title)
+            print "Creating a new TableBlock on page %s of page type %s" %(wagtail_page.title, wagtail_page.__class__.__name__)
 
 
 def update_full_width_text_tables(wagtail_page):
```
vccabral: I know this isn't in your scope for the PR but who ever left this code left it super nested. Might be the moment to pull out the loops/conditions into methods.

cfgov/v1/migrations/0006_update_tables.py
willbarton: That and `print ""` isn't python3-compatible either. If we're not using logging, I'd prefer to see us importing `print_function` from `__future__` so we don't introduce the need for future changes.

richaagarwal: @vickumar1981 Mostly b/c we're not using a logger anywhere else, but I agree it's better to log than to print so I'll update it! Thanks.

cfgov/v1/migrations/0006_update_tables.py
richaagarwal: @vccabral Ha, I wrote the original file as well so it can be in scope if we want it to be.  I did try to break out things into functions when it made semantic sense to, but I agree it's still too nested. Could you give an example of what you're thinking would resolve this?

cfgov/v1/migrations/0006_update_tables.py
richaagarwal: @vccabral This doesn't look that much cleaner to me, it's slightly less readable, & it doesn't take into account the headers logic.  What would be the benefit of doing it this way?

vccabral: Great points. I would move the header logic into it's own method for readability and each of the nested comprehensions should be in their own method too.

vccabral: Comprehensions are considered more readable and more pythonic. They have lower overhead and are optmized...etc.

```
my_list = []
for item in list2:
    if cond(item):
       my_list.append(operand(item))
```
becomes

```
my_list = [operand(item) for item in list2 if cond(item)] # this reads like english
```

cfgov/v1/migrations/0006_update_tables.py
vccabral: Have each nested layer call a method so the for loops only go 1 deep. You can also move the if conditions to the right of the for loop on the same line.

richaagarwal:
I addressed most of the comments; ready for a final review.  Please review this time with an eye for data integrity as opposed to style - this is a migration that will be run once, does not need to be maintained, and will eventually be squashed. 

Note that I ended up keeping the print statements as setting up logging would've been way out of scope for this PR, but is definitely something we should do separately!

chosak:
A few things:

---

I didn't test this against the full set of production pages - I'm guessing that might be possible by pulling down a prod data dump at migration `v1.0005`? Having to do that is somewhat suboptimal, because at best it would just be checking that the code runs without failing, and then possibly doing some eyeballing of (a subset of?) existing tables to see if things look okay.

It'd be nice if there were some unit tests of this code that first set up tables, then applied the conversion code, then checked to see that the generated HTML is the same (I did some HTML comparison in [the dynamic job blocks code](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jobmanager/tests/models/test_blocks.py) and the Wagtail blocks code also has [some simple tests that check HTML](https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailcore/tests/test_blocks.py#L69)).

Having unit tests would allow for checking of the full set of table contents (text, hyperlink, text blob, rich text blob) and check things like unicode characters or empty fields. I know this is just a one-off migration but unless we're confident that this won't break any data in production it might give some additional assurance.

---

In order to test this locally, I had to do the following steps:


1. Drop my existing database and create an empty one.

 ```sh
$ ./drop-db.sh
$ ./create-mysql-db.sh
```

2. Temporaily remove migrations v1 0006, 0007.

 ```sh
$ rm cfgov/v1/migrations/0006* cfgov/v1/migrations/0007*
```

3. Apply all migrations through v1 0005

 ```sh
$ ./initial-data.sh
```

4. Start local server and create some pages in Wagtail that have `Table` blocks.

 ```sh
$ ./runserver.sh
```

5. Restore and apply migration 0006 (the modified version from this PR).

 ```sh
$ git checkout -- cfgov/v1/migrations/0006_update_tables.py
$ python cfgov/manage.py migrate v1 0006
```

6. Visually inspect the Wagtail pages to ensure that the `TableBlock` content looks the same.

Is this the best way to test the conversion locally?

---

More concretely, I noticed a visual difference in how multi-line TextBlobs are rendered between `Table` and `TableBlock` (paging @sebworks).

Here's what I entered in the Wagtail editor using a `Table`:

image

Here's what it looks like when the `Table` is rendered:

image

Here's the `Table` source code:

image

Here's what it looks like after being converted to a `TableBlock`:

image

Here's the `TableBlock` source code:

image

Is this a case where custom logic is required, as in the hyperlink?

---

Lastly, even though this is a one-off migration, it might be useful to try to clean up some of the more straightforward style issues, if only for readability. By straightforward I mean the issues around spacing and unused imports, not whether to use a for loop or list comprehension. [Here's](https://github.cfpb.gov/gist/chosak/54f7b5b1d9519f24b3e3ad419a8bfc77) a `flake8` dump of the new migration file, if you think it might be useful.

**Comment by richaagarwal** [comment]
@chosak Thanks for your detailed comments!

- To test locally, it's easiest to use a prod dump. Recent dumps are a lot faster than before. The migration in question, `0006_update_tables.py` that is already in master is not going to be in prod until this PR gets merged in and also gets deployed, so we do not need to worry about that. 
You might run something like
`./refresh-data.sh production.sql` and then `python cfgov/manage.py migrate --fake-initial`. 

- I think that there could be a lot of variations to the data in the old Table type, but for purposes of migrating, we should just focus on the ones that exist and make sure those work. To that end, I'm not actually sure how useful unit tests would be.  I've done a visual gut-check of the pages that get updated. Here is the output from the migration:

```
Updating tables in content field
Updating Table to a TableBlock on page College Credit Card Agreements of page type BrowsePage
Updating Table to a TableBlock on page Small Business Review Panels of page type BrowsePage
Updating Table to a TableBlock on page test to delete of page type BrowsePage
Updating Table to a TableBlock on page Adjudication Proceeding Rules of page type BrowsePage
Updating Table to a TableBlock on page Warning Letters of page type BrowsePage
Updating Table to a TableBlock on page Serving Your Country of page type SublandingPage
Updating Table to a TableBlock on page Civil Penalty Fund allocation schedule of page type LearnPage
Updating Table to a TableBlock on page Integrity Advance, LLC and James R. Carnes of page type DocumentDetailPage
Updating tables in full width text organisms
Updating Table to a TableBlock on page Testing options for Prepaid - LJ+KD of page type BlogPage
Updating Table to a TableBlock on page Title XIV Rules: Mortgage Origination of page type BrowsePage
Updating Table to a TableBlock on page Title XIV Rules: Mortgage Servicing of page type BrowsePage
Updating Table to a TableBlock on page Payments by Case of page type BrowsePage
Updating Table to a TableBlock on page System of Records Notices of page type BrowsePage
Updating Table to a TableBlock on page CFO update for the third quarter of fiscal year 2012 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the third quarter of fiscal year 2012 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the third quarter of fiscal year 2012 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the third quarter of fiscal year 2012 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the third quarter of fiscal year 2012 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the second quarter of fiscal year 2012 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the second quarter of fiscal year 2012 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the second quarter of fiscal year 2012 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the first quarter of fiscal year 2012 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the first quarter of fiscal year 2012 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the first quarter of fiscal year 2012 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the fourth quarter of fiscal year 2011 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the fourth quarter of fiscal year 2011 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the fourth quarter of fiscal year 2011 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the fourth quarter of fiscal year 2011 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the third quarter of fiscal year 2011 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the third quarter of fiscal year 2011 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the third quarter of fiscal year 2011 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the second quarter of fiscal year 2011 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the second quarter of fiscal year 2011 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the second quarter of fiscal year 2011 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the first quarter of fiscal year 2011 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the first quarter of fiscal year 2011 of page type LearnPage
Updating Table to a TableBlock on page CFO update for the first quarter of fiscal year 2011 of page type LearnPage
```

Once this is merged in, perhaps either @schaferjh or @sarahsimpson09 could do a review of the pages above. There are only about 13 unique ones.

richaagarwal:
Btw, so everyone is on the same page, this work will not go live until @sebworks work (on fixing up the admin front-end so that content editors can continue to have a rich text editor inside a Table block) is complete and merged in.  We will roll back this existing migration from master if necessary to deploy before it is all ready.

sebworks:
@chosak, yeah it looks like that's what the `linkbreaksbr` [filter](https://github.com/torchbox/wagtail/blob/8ed30060fdd3e91bee608658e3741037b5fbff8e/wagtail/contrib/table_block/templates/table_block/blocks/table.html#L9 ) is doing in the template. Not sure how to handle that for legacy vs new tables.

schaferjh:
@richaagarwal yes, I am happy to do a visual check of the necessary pages once this is merged. If you could just help remind me when the time comes. 😄

richaagarwal:
@chosak Thanks for linking to the flake8 file!  I went ahead and installed the tool locally and worked through all the errors. Super helpful!

richaagarwal:
Closing this, see #2367 instead.

PR #2346, opened 2016-08-26
Updates gulp-mocha/mocha to version `3.x.x`

anselmbradford:
Updates to next major version of mocha, which drops older node support.

## Changes

- Updated mocha npm module to version `3.0.2` from `2.4.5`.
- Updated gulp-mocha npm module to version `3.0.1` from `2.2.0`.
- Requires gulp-mocha module directly instead of going through plugins since new version had an issue with the plugins syntax (this is probably preferable anyway so that the module name shows up in searches).

## Testing

- `./frontend.sh` 

## Review

- @contolini 
- @sebworks 
- @virginiacc


contolini:
👍

PR #2314, opened 2016-08-15
New Wagtail page type for Job Listings

chosak:
This PR implements a new Wagtail page type and associated panel types to allow for creation and editing of job listings through the Wagtail admin site.

Job listings follow this data model (all fields are required unless otherwise specified):

- Title
- Description
- Open/close dates
- Salary min/max
- Division
- Grade(s), one or more
- Region(s), one or more
- Application link(s), one or more of these two types:
  - USAJobs link, consisting of:
    - USAJobs announcement number
    - USAJobs URL
    - USAJobs applicant type
  - Email link, consisting of:
    - Email address
    - Email label (heading)
    - Email description (optional)

## Additions

- New Wagtail page type `jobmanager.models.pages.JobListingPage` for a single job listing.
- New Panel types for the various related models that need to be part of a job listing:
  - `jobmanager.models.panels.EmailApplicationLink` for email application links.
  - `jobmanager.models.panels.USAJobsApplicationLink` for USAJOBS application links.
  - `jobmanager.models.panels.GradePanel` for job grades.
  - `jobmanager.modes.panels.RegionPanel` for job regions.
- New section of the Wagtail admin for creating and editing all data related to job listings (see screenshots below).

## Changes

- Added `wagtail.contrib.modeladmin` to the list of `INSTALLED_APPS` to be able to use Wagtail's [ModelAdmin](http://docs.wagtail.io/en/v1.5.3/reference/contrib/modeladmin.html).
- Various edits to the job listing page template created in #2296 to support data model.
- Refactored `jobmanager.models` as a package with submodules as opposed to a single module file. Moved "legacy" Django models to `jobmanager.models.django`.
- ~~Change to page sharing logic to support use of ForeignKeys in Wagtail pages. See comment below for additional detail; this should be reviewed.~~ (Reverted this after modifying how some of the foreign key models were referenced and it no longer seems to be needed.)

## Testing

- Run `tox` or `tox -e fast` to verify that all unit tests pass, including new ones.

These work best if you have job data already loaded in your DB, ideally from a production data dump of the `jobmanager` models as a fixture, then loaded via `manage.py loaddata jobs.json`:

- Explore the various job-related pieces of data in the Wagtail admin:
 1. In the Wagtail admin, look at the "Job listings" menu in the vertical nav bar.
 2. Select, for example, "Regions", and note that you can create or edit relevant job-related data.
- Using the Wagtail admin, create a new job listing page locally:
[...]

PR #2347, opened 2016-08-26
Add sidebar to Wagtail JobListingPage

chosak:
This PR adds the job listing page sidebar to the new Wagtail `JobListingPage`, copying it from [the legacy Django template](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/about-us/careers/_single.html#L132).

## Changes

- Added hardcoded sidebar content to `cfgov/jinja2/v1/job-description-page/index.html`.
- Added a bit of logic to the template to handle the (unlikely) case that there are no specified regions or grades for a job. 

## Testing

- Create a new `JobListingPage` in the Wagtail admin, and preview. Compare the generated sidebar with one on the live site that uses the Django template, for example [this page](http://www.consumerfinance.gov/about-us/careers/enforcement-economist-cn-6060/).

## Review

- @sebworks 

## Screenshots

Page with no grades or regions specified (again, this is unlikely, but better to show something than just have a blank):
image

## Todos

- Per @schaferjh for now we are going to hardcode this content, but in future it would be nice to have it default to something but be configurable through the standard Wagtail sidebar editor.

## Checklist


sebworks:
👍

schaferjh:
> Added a bit of logic to the template to handle the (unlikely) case that there are no specified regions or grades for a job.

* Did you mean type/division rather than region? The screenshot doesn't include the region listing. 
* "All divisions" seems like a fine fall back for no division/office
* The dummy data for expiration date and grade strikes me as potentially more confusing/harmful than helpful. Could those two fields display "- -" ?
* Actually, could expiration date and grade be required fields in the admin? So the cms throws an error if someone tries to publish without those?

chosak:
@schaferjh Sorry for the confusion; I did mean region.

I believe our data model specifies that each job "should" have 1 or more regions and 1 or more grades, but the Wagtail editor interface does not enforce this, and lets you select 0.

This PR includes a change where if no regions are specified, I just remove that line from the job header. Otherwise you'd just see "REGION: " with nothing after it.

Division and expiration date are currently required field, as are minimum/maximum salary. So these will always appear.

I mentioned enforcing region/grade as a TODO in [the PR where these pages were created](https://github.com/cfpb/cfgov-refresh/pull/2314)). This is a bit tricky because it's page-level validation as opposed to field-level.

schaferjh:
ok, well for now, in the required fields, if nothing is selected for expiration date and max/min salary they should get a - - rather than placeholder information. I suppose it's fine for the actual grade to be blank, and "All divisions" is fine for divisions.

chosak:
Sorry, I'm not sure if that requires additional changes (or if that's in scope for this PR). Currently in the Wagtail admin the editor for those fields looks like this:

image

Note the stars for required fields, so editors can't even save a page without specifying a close date and salary range.

I created a new backlog task for enforcing at least one Region or Grade; the change here was only meant to prevent a blank element on the page until that is done.

schaferjh:
ok, I think this was all a miscommunication. I assumed all the things in your screenshot were the result of items being left blank in the admin. I was skimming too quickly and not clearly understanding what the screenshot was showing. Your original solution of no grade and no region if those are empty is fine.

New backlog task to enforce at least one Region or Grade sounds great. thanks.

Re: the focus of this PR, I assume the hardcoded sidebar looks just like the sidebar live on the site today? i.e. here: http://www.consumerfinance.gov/about-us/careers/enforcement-economist-cn-6060/

If so then 👍

chosak:
@schaferjh yep, the hardcoded sidebar matches the legacy Django page sidebar exactly. Thanks for the review.

PR #2349, opened 2016-08-26
Modify default Wagtail page settings panel order

chosak:
To make editing Wagtail pages easier for content managers, this PR reorders items in the page configuration tab to match the following specified order:

 * Settings, 
 * Categories, 
 * Tags, 
 * Page preview fields (expanded by default)
 * Authors, 
 * Relevant dates, 
 * Scheduled publishing
 * Language

## Additions

- New tests to enforce that all pages that inherit from `CFGOVPage` comply with this order.

## Changes

- Modified order of fields on `CFGOVPage` and `AbstractFilterPage` to comply with new order.

## Testing

- Create a new `DocumentDetailPage` or edit an existing one in the Wagtail editor and notice that the field order has changed, and that page preview fields should be expanded by default.

## Review

- @cfpb/cfgov-backends @schaferjh 

## Checklist

### DIFF
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -117,11 +117,11 @@ class CFGOVPage(Page):
 
     settings_panels = [
         MultiFieldPanel(Page.promote_panels, 'Settings'),
+        InlinePanel('categories', label="Categories", max_num=2),
         FieldPanel('tags', 'Tags'),
         FieldPanel('authors', 'Authors'),
-        FieldPanel('language', 'language'),
-        InlinePanel('categories', label="Categories", max_num=2),
         MultiFieldPanel(Page.settings_panels, 'Scheduled Publishing'),
+        FieldPanel('language', 'language'),
     ]
 
     # Tab handler interface guide because it must be repeated for each subclass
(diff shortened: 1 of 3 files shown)


schaferjh:
@chosak could I get a screenshot of the new configuration tab on DocDetail? 

Also, does this re-ordering extend to other page types? We'd like it to be consistent across the board.

chosak:
@schaferjh yes, the ordering applies to all of our page types. If a page has only a subset of the fields, then that subset is still ordered in the same way.

Here's the `DocumentDetailPage` configuration tab, for a new page:
image
image

Scotchester:
Can we make the Categories panel open by default, too? Seems strange that it's the 2nd priority in terms of order, and yet the only one closed by default.

chosak:
@Scotchester the Categories panel is slightly different because it's a Wagtail `InlinePanel` (where you add one or more categories) vs. just a `MultiFieldPanel` that might be expanded or collapsed.

I see the behavior on a new page where you see this when there are no categories:
image

And then this once you've added one:
image

The view where Categories are expanded but not selected looks like this:
image

But in this case, you have to select one, otherwise page validation fails. Is categories supposed to be a required field for all page types? If not, then it seems like the collapsed, no categories added, state is appropriate.

Scotchester:
@chosak Ahhhhh, I see. That makes sense.

schaferjh:
It would be nice to have categories open by default, but no, not every page *must* have a category. So, if category panel open = this page must have a category, then I guess we'll go for collapsed by default and just be happy it's further up the page!

👍

PR #2350, opened 2016-08-26
Updates reference to glob-all from glob

anselmbradford:
`glob` was being required while it was not imported in `package.json` so that reference should be updated to the imported `glob-all` module.

## Changes

- Updates reference to glob-all from glob


sebworks:
👍

PR #2351, opened 2016-08-28
Hide unpublished jobs in jobs list/table

chosak:
The new Wagtail jobs list/table blocks introduced in #2341 show all jobs that are currently open, but fail to hide any `JobListingPage` pages that aren't yet published to the live site (those that are in draft or shared mode). This PR introduces logic to filter these out.

## Changes

- Additional filter that enforces `live=True` in the `OpenJobListingsMixin` that the table/list blocks use to generate their `QuerySet` of `JobListingPage` objects.

## Testing

- Run `tox -e fast` and see new tests in `jobmanager.tests.models.test_blocks`.
- To test the `JobListingTable`:
 1. Run a local server and create a new `BrowsePage`, then add a `JobListingTable`.
 1. Create a new `JobListingPage` under the `CFGOV` root, and select "Save as Draft" (make sure your job open/close dates make it currently open).
 1. Note when you refresh your `BrowsePage` that the job does not show up in the table. 
 1. Now go back to your job page and select "Share on Content".
 1. Note that the job still does not appear when the table is refreshed.
 1. Now go back to your job page and select "Publish to WWW".
 1. You should see your new job appear in the table.
- To test the `JobListingList`:
 1. Repeat the above steps but with a `SublandingPage` with a `JobListingList` (in the sidebar).

## Review

- @richaagarwal @cfpb/cfgov-backends 

## Checklist


richaagarwal:
Ah I love that this PR is mostly tests.   And sorry I didn't catch this in my earlier review. 

My main question/concern is, should we be doing this filtering at the `ModelList` level? I imagine we wouldn't want this to accidentally happen with any other lists introduced.

cfgov/jobmanager/tests/models/test_blocks.py
```diff
@@ -10,13 +11,15 @@
...
 
 
-def make_job_listing_page(title, close_date, grades=[], regions=[]):
+def make_job_listing_page(title, close_date=None, grades=[], regions=[],
+                          **kwargs):
     page = mommy.prepare(
         JobListingPage,
         title=title,
-        close_date=close_date,
+        close_date=close_date or timezone.now().date(),
```
vccabral: What value does this have at the end of this line? 

close_date or [value here] <= this looks like a boolean output with a date. Doesn't make sense to me.

richaagarwal: @vccabral It's `close_date` if it's not `None`, otherwise it's the value of `timezone.now().date()`.  It would only be assigned a boolean if one of those variables returned a boolean.

chosak: @vccabral right, like @richaagarwal says it's a useful shorthand to check whether a value for the `close_date` argument was passed in.

Saying this:

```py
close_date or timezone.now().date()
```

is just a shorter way of doing this check:

```py
if close_date is None:
    close_date = timezone.now().date()
```

All non-`None` `datetime` objects are truthy when treated as a boolean value.

chosak:
@richaagarwal that's a good point. `ModelList` is for any model, not necessarily those that are page types. I suppose I could make an intermediate `PageModelBlock` or something that adds this filter at that level, so it'd really be a `PublishedOnlyModelBlock` and then `JobsListBlock` and `JobsListTable` would inherit from that. Might be a few too many abstract classes though until there's a real need.

I'd probably prefer to leave it as-is until there's a second use case for doing filtering of pages, but am happy to make that change if you feel like it's unclear or risky somehow.

richaagarwal:
This either needs to get merged in before we deploy, or as a hotfix if after, right?

richaagarwal:
👍

willbarton:
:+1:

PR #2353, opened 2016-08-29
Just a little docker

rosskarchner:
Provides dockerfiles and docker-compose configuration for spinning up a self-contained development environment. If we decide this is the future, then users won't need to worry about locally running MySQL, elasticsearch, PDFReactor, .env files, and friends in order to work on the site. 

There are also some new "batteries included" that have never been a part of our standard development environment, but should prove useful:
- phpMyAdmin, which provides a web-based gui for inspecting and managing the database.
- fakeS3, which lets us simulate Amazon S3 while working locally. Django is configured to use this for media and image uploads.

I have a few more goals for this before it is merged, but feel free to kick it around, make it better, and contribute ideas here!

Todo:
- [ ] Instructions for running tests. Bonus: Have Travis use Docker.
- [ ] Have a process /  instructions for working on external apps/ sheer sites 
- [ ] Integrate instructions into the 'installation' doc
- [ ] provide a way to grab missing media from S3
### first time

Get the docker bits:

`brew install docker docker-compose docker-machine`

Create a VM for running the docker daemon:

`docker-machine create default --driver virtualbox --virtualbox-cpu-count "2"  --virtualbox-memory "3072"`

(There are [a lot of options available](https://docs.docker.com/machine/reference/create/#/accessing-driver-specific-flags-in-the-help-text)-- these might not be the ideal settings. In particular, [more cpu's isn't necessarily faster](https://github.com/docker/machine/issues/3420#issuecomment-223152559))

`./docker/map-ports-to-localhost.sh`

(this maps a number of useful ports to localhost)

You need a license in order for pdfreactor to work. Download this text file from our github enterprise (cfgov/platform), and drop in into your cfgov-refresh folder: /CFGOV/platform/blob/master/pdfreactor_key.txt
### every time

`eval $(docker-compose env)`
(this connects the local docker utilities to the daemon running in the virtual machine we created above)

`docker-compose up`

This uses the configuration in docker-compose.yml to build and spin up a set of containers. This takes about 15 minutes on my machine, the first time. Subsequent launches will be MUCH faster.

When it's done you should be able to check out:

http://localhost:8000/  (django)
http://localhost:3000/ (gulp watch/browsersync)
http://localhost:9423/ (PDFReactor Demo)
http://localhost:8080/phpMyAdmin/   (phpMyAdmin. log in with any name and any password, but neither can be blank)

Changes to python files or static assets should take affect immediately. Changes to python requirements or package.json will require either:

`docker-compose build django` or `docker-compose build gulp`

If you would like to interact with django on the command line, in another terminal tab, run:
- `eval $(docker-compose env)`
- `docker-compose run django django-admin [subcommand]`, where [subcommand] is any [django management command](https://docs.djangoproject.com/en/1.8/ref/django-admin/).

Hitting control-c will stop all of the running containers.
### importing a database:

Drop the file you want to load into the 'dbupload' folder. Log into phpMyAdmin, click the 'import' tab, and check the 'select from the webserver upload directory' option.

screen shot 2016-08-29 at 3 15 48 pm

Files can be uncompressed SQL, or can be zipped or gzipped.
### taking a break

You can spin down the Docker vm with:

`docker-machine stop` You will then need to `docker-machine start` to start back up again.

### Satellite apps

- Any apps checked out in the develop-apps directory will be `pip install -e`'ed when the Django container starts. This is helpful for apps that don't have a front-end build (like [django-hud](https://github.com/cfpb/django-hud)), but for apps like AskCFPB and regulations-site, you're left to figure out the front-end on your own (using a local copy of Node, etc).
- There is not yet a strategy here for sheer sites.

## Review
- @user
## 
## 
## Checklist
- [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
- [ ] New functions include new tests
- [ ] Placeholder code is flagged
- [ ] Visually tested in supported browsers and devices
- [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)


willbarton:
> If we decide this is the future, then users won't need to worry about locally running MySQL, elasticsearch, PDFReactor and friends in order to work on the site.

❤️

rosskarchner:
I just added a bit about using a PDFReactor license.

higs4281:
> eval $(docker-compose env)

I think this should be 'eval $(docker-machine env)'

higs4281:
> If you would like to interact with django on the command line, in another terminal tab, run:

This doesn't work for me unless I also call eval in the new tab:
```
eval $(docker-machine env default)
docker-compose run django django-admin dbshell
```

rosskarchner:
@higs4281 thanks for those-- fixed!

rosskarchner:
closing this for now, though I plan to re-open before 2017

rosskarchner:
It LIVES.  I'm going to recommend we merge this soon. It shouldn't break any existing workflows or processes, and provides a simpler (and more portable) experience for developing on cfgov-refresh. I removed the changes that used docker in TravisCI (for now)

The big missing feature, is a natural way to work on satellite apps, but I no longer feel like that should hold this up.

rosskarchner:
I just realized PDFReactor needs some work

cfgov/cal/views.py
```diff
@@ -11,19 +11,18 @@
...
-if six.PY2:
-    try:
-        sys.path.append(os.environ.get('PDFREACTOR_LIB'))
-        from PDFreactor import *
-    except ImportError:
-       PDFreactor = None
+try:
+    from PDFreactor import *
+except ImportError:
+    PDFreactor = None
```
richaagarwal: some day I will understand why this is necessary

rosskarchner: basically, the PDFReactor python lib isn't distributed via pip and friends, it's bundled with PDFReactor itself. This is a way to make cfgov-refresh work without having PDFReactor installed.

PR #2359, opened 2016-08-30
Line length feature flag

contolini:
This is an attempt to move some CSS behind a Wagtail feature flag. See https://[GHE]/CFGOV/platform/issues/237 for line length background.

## Additions

- Feature flag gulp task.
- Feature flag CSS directory to put relevant stylesheets in.
- Feature flag conditional at the bottom of `base.html`.

## Testing

- Pull down branch.
- `gulp styles`
- Log into the wagtail admin and add a `NEW_LINE_LENGTH` feature flag.

screen shot 2016-08-30 at 4 28 53 pm

When it's enabled (most) lines should wrap after `41.875rem`. For example, try visiting http://localhost:8000/office-civil-rights/eeo-policy-and-reports/ before and after toggling the `NEW_LINE_LENGTH` checkbox.

## Notes
- After the feature has been reviewed by stakeholders we'll remove this flag and merge the CSS into the appropriate stylesheet.

## Review

- @rosskarchner 
- @virginiacc 
- @sebworks 

## Checklist


rosskarchner:
me right now: http://www.gettyimages.com/license/465117348

sebworks:
👍 , looks great. Should we set the `max-width` on `.blocks` as well?

cfgov/unprocessed/css/feature-flags/line-length.less
```diff
@@ -0,0 +1,11 @@
+// Experimenting with a capped line length across all website content.
+.content_main {
+    h3,
+    h4,
+    h5,
+    h6,
+    p,
+    li {
```
Scotchester: add `dt` and `dd` to the list

contolini:
> 👍 , looks great. Should we set the max-width on .blocks as well?

Ehh I'll let @Scotchester or @anselmbradford answer that one.

KimberlyMunoz:
> Should we set the max-width on .blocks as well?

My instinct is no, because we might get some unexpected behavior on blocks with backgrounds and borders if applied directly to the block. Also content main covers almost all the big blocks of text unless there's a block in the sidebar, navigation, hero homepage or footer, so it doesn't seem necessary.

Scotchester:
I agree with @KimberlyMunoz's answer.

sebworks:
> I agree with @KimberlyMunoz's answer.

👍

PR #235, opened 2015-01-14
Allow for proxy when sending pass/fail status to saucelabs

dpford:
Previously, when using a proxy, the status wouldn't be sent to sauce labs.  Now it will check for proxy via the HTTP_PROXY environment variable and set it before sending the pass/fail status.

PR #2375, opened 2016-09-07
Support for special characters in tags!!!!!!!!!

richaagarwal:
## Context 
- See GHE #235 
- Previously, if a special character was a part of a tag (e.g. "Consumers' response"), the front-end would throw an uncaught JS error and multiselect would not work! Oh no.
- Now, this is fixed.

## Manually test (if you want to)
- Pull down branch & run `gulp scripts`
- Add a tag to a page under `data-research/research-reports` with special characters (such as an apostrophe) 
- Go to [http://localhost:8000/data-research/research-reports/](http://localhost:8000/data-research/research-reports/) and see that the multiselect in "Topics" in "Filter reports" still works, and displays the tag you entered correctly.
- Go to other pages and click on links from the topic tags or author tags and confirm things are working as expected.

## TODO
- Will update CHANGELOG once this is ready to get merged in

## Credits
- @contolini and I paired on this
- Thanks to @kurtrwall and his work here https://github.com/cfpb/cfgov-refresh/pull/1746 that guided this.


cfgov/unprocessed/js/modules/util/dom-manipulators.js
```diff
@@ -14,11 +14,13 @@ function create( tag, options ) {
   for ( var i in options ) {
```
richaagarwal: @sebworks Do the changes to this file look sane to you?

sebworks: 👍

contolini:
This is some high quality code. 👍 💯

cfgov/v1/__init__.py
```diff
@@ -186,10 +186,10 @@ def related_metadata_tags(context, page):
     # From an ancestor, get the form ids then use the first id since the
     # filterable list on the page will probably have the first id on the page.
     id, filter_page = get_filter_data(page)
-    for tag in page.specific.tags.names():
-        tag_link = {'text': tag, 'url': ''}
+    for tag in page.specific.tags.all():
+        tag_link = {'text': tag.name, 'url': ''}
         if id is not None and filter_page is not None:
-            param = '?filter' + str(id) + '_topics=' + tag
+            param = '?filter' + str(id) + '_topics=' + tag.slug
```
vickumar1981: not sure what values are allowed for tag.slug, but you might want to urlencode this.

```
from urllib import urlencode
params = {"filter": id, "_topics": tag.slug
param = "?%s" % urlencode(params)
```

cfgov/v1/models/base.py
```diff
@@ -149,7 +149,7 @@ def generate_view_more_url(self, request):
         available_tags = [tag[0] for name, tags in form.fields['topics'].choices for tag in tags]
         tags = []
         index = util.get_form_id(activity_log)
-        for tag in self.tags.names():
+        for tag in self.tags.slugs():
             if tag in available_tags:
```
vickumar1981: this if can be moved up a line: `for tag in self.tags.slugs if tag in available_tags:`

cfgov/v1/models/base.py
```diff
@@ -149,7 +149,7 @@ def generate_view_more_url(self, request):
...
                 tags.append('filter%s_topics=' % index + urllib.quote_plus(tag))
```
vickumar1981: this funtionality of appending the proper url parameters to the tag, imo, belongs somewhere else, i.e. in the urls.py.  if this eventually calls a view, it should be placed in the urlpatterns.

richaagarwal: Please only review changed code, thanks.

cfgov/v1/__init__.py
willbarton: @vickumar1981 [slugs](https://github.com/alex/django-taggit/blob/master/taggit/models.py#L25') are [already url-safe](https://docs.djangoproject.com/en/1.10/ref/models/fields/#slugfield).

That said, this might be cleaner with Python 3-compatible string formatting, something like:

```
param = '?filter{id}_topics={slug}'.format(id=str(id), slug=tag.slug).
```

But that's not a necessary change. I would very much like to avoid introducing any more old-style string formatting though.

cfgov/v1/forms.py
```diff
@@ -112,25 +111,33 @@ def __init__(self, *args, **kwargs):
         self.set_topics(parent, page_ids, hostname)
         self.set_authors(parent, page_ids, hostname)
 
+
+    def prepare_options(self, arr):
+        """ Returns an ordered list of tuples of the format ('tag-slug-name', 'Tag Display Name') """
+        arr = Counter(arr).most_common() # Order by most to least common
```
higs4281: nice use of Counter.

higs4281:
👍

PR #2376, opened 2016-09-07
Updates shrink-wrap

anselmbradford:
For some reason shrinkwrap was outdated. I don't get why, but it was :(

## Changes

- Re-generated shrinkwrap because ./frontend.sh was failing.

## Testing

- `./frontend.sh` should pass.

## Review

- @contolini 
- @sebworks 
- @virginiacc

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -943,19 +943,7 @@
     "bin-build": {
       "version": "2.2.0",
       "from": "bin-build@>=2.0.0 <3.0.0",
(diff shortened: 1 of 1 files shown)


sebworks:
👍

sebworks:
Why doesn't the error get caught by Travis?

anselmbradford:
Got me ¯\\_(ツ)_/¯ 
Possibly I needed to run prune locally, but still, re-running `npm shrinkwrap --dev` shouldn't have updated anything, but it did.

PR #254, opened 2015-02-17
Df1462 calendar icon

anselmbradford:
- Adds initial calendar icon CSS that looks like:

screen shot 2015-02-17 at 3 40 02 pm

- Removes excess whitespace from Gruntfile.

PR #2377, opened 2016-09-07
Change name and icon for 'Related links'

higs4281:
This does not alter functionality, only the name and icon for a sidebar link 
in the Wagtail admin to satisfy GHE #254.

## Additions

- related-content templates to replace related-links versions for clarity

## Changes

- Updated sidefoot labels for consistent capitalization

## Testing
- `tox -e fast` tests should pass.
- sidefoot label when editing a Wagtail page should be "Related content" and the "grip" icon should show.

## Review
@richaagarwal @chosak @contolini @virginiacc 

* [x] CHANGELOG updated


contolini:
@higs4281 Where is the grip icon supposed to show? Can you take a quick screenshot?

higs4281:
The grip icon should show up above the "Related content" label in a the sidebar tab while editing a page (this is /admin/pages/457/edit/):

grip

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Updated es5-shim npm module to version `4.5.9` from `4.5.7`.
 - Updated gulp-rename npm module to version `1.2.2` from `1.1.0`.
 - Replaces deprecated gulp-cssmin with gulp-clean-css.
+- Updated admin page sidefoot label and icon to read 'Related content'
```
Scotchester: Could probably throw in a "Related links" here to indicate _which_ label and icon have been updated ;)

cfgov/jinja2/v1/_includes/molecules/related-content.html
```diff
@@ -1,21 +1,21 @@
...
     Description:
 
-    Creates related links markup when given:
+    Creates related content markup when given:
 
     value:               Object defined from a StreamField block.
 
-    value.heading:       The heading of the related links inset.
-                         Default is "Related Links".
+    value.heading:       The heading of the related content inset.
```
Scotchester: Could be worth noting that this option is optional.

cfgov/jinja2/v1/_includes/related-content.html
```diff
@@ -24,7 +24,7 @@
 <div class="related-links">
     <h2 class="header-slug">
         <span class="header-slug_inner">
-            Related links
+            Related content
```
Scotchester: I'm not entirely sure how this file is used, but everything else in it is pretty link-specific, so I think I would revert this change.

Scotchester: And now that I've scrolled down to event.html and newsroom/index.html, I see how this is used. It's a separate concern from the "Related content" change, so I would also revert the name of this file, and revert the changes in those two files that call it.

higs4281:
@schaferjh Is this what you were after?

cfgov/v1/atomic_elements/molecules.py
```diff
@@ -12,10 +12,14 @@ def isRequired(field_name):
 
 
 class HalfWidthLinkBlob(blocks.StructBlock):
+    h_text = ('A list of icon names can be obtained at: '
```
Scotchester: Minor: I think I like what you're doing here, but I feel like the variable name should be more specific. Perhaps `icon_help_text` or `icon_help`?

cfgov/jinja2/v1/_includes/related-content.html
higs4281: The module can be used to deliver more than links, which I think is why the issue was created.  
Happy to revert template name. Any other votes?

schaferjh:
👍  thanks @higs4281

cfgov/jinja2/v1/_includes/related-content.html
Scotchester: Yes, I understand. I'm referring _only_ to this file. The file above, the one in `molecules/`, is the one that needs to be updated for the Wagtail module. _This_ file seems to only be called directly by some templates, like Events and Newsroom.

richaagarwal:
I agree with @Scotchester 's comments and otherwise it looks good!

One thing for us to think about (but not meant to be a blocker) - the GHE issue refers to updating the display name; there isn't really a reason to update the rest of the code (since it remains technically accurate), but curious if people feel variable names should match exactly what we display.

cfgov/v1/atomic_elements/molecules.py
chosak: A bit late to this, but I'd agree that if this is going to be pulled out as a separate variable it should be given a more precise name, e.g. `sub_heading_icon_help_text`. That being said, is the reason to pull this out solely for formatting purposes? Doing it this way means there's now a `h_text` class variable on the class itself which doesn't really serve much purpose.

A PEP8-compliant way of doing this might be something like:

```py
    sub_heading_icon = blocks.CharBlock(
        required=False,
        label='Sub heading icon',
        help_text=(
             'A list of icon names can be obtained at: '
             'https://cfpb.github.io/capital-framework/components/cf-icons/. '
             'Example: linkedin-square, facebook-square, etc.'
        )
    )
```

higs4281:
@richaagarwal @Scotchester Adjustments made.

higs4281:
@chosak so much better. thx.

Scotchester:
:+1:

chosak:
👍

Scotchester:
Just realized that this didn't include a migration. Does it need one?

chosak:
@Scotchester yes, it looks like it:

```sh
$ python cfgov/manage.py makemigrations v1
Migrations for 'v1':
  0008_auto_20160908_1247.py:
    - Alter field content on browsepage
    - Alter field molecules on demopage
    - Alter field header on homepage
    - Alter field content on landingpage
    - Alter field content on sublandingpage
```

It's a little bit hard to tell exactly what the changes are from looking at the migration that gets generated. Are these all pages/fields that would be touched by this PR?

Unfortunately because of the way Wagtail stores block configuration in the database, any time you do anything as simple as changing help text, it creates a migration.

Scotchester:
Yes, it does look like the changes in this PR touch each of those areas.

chosak:
@Scotchester sorry, I should have caught this if I had looked more closely when reviewing this PR. Do you want to create a follow-up PR that just has that migration?

PR #2378, opened 2016-09-07
Data migrations for site root and careers pages

chosak:
This PR adds data migrations that create a site root, set up the careers page tree ("About Us", "Careers", etc.) and convert Django jobs into Wagtail `JobListingPage` pages.

Here's the rationale:

1. We want to use a Django data migration to convert existing `jobmanager.models.django.Job` model instances to Wagtail pages of type `jobmanager.models.pages.JobListingPage`.
1. To do this, we need a Wagtail parent page to attach these new child pages to.
1. It makes sense for this page to be the "Careers" root, and we need these pages to be created anyway, so let's add another Django data migration to make these.
1. All new pages need to be added to a site root, and it'd be good if this was consistent with how we set things up in production.
1. Let's make another Django data migration that properly sets up the Wagtail sites and site root in a way that is consistent with production.

If the code changes here seem a bit overly complex, it's because it's [not really possible](http://stackoverflow.com/a/33807804) to properly use Wagtail `Page` methods from within a Django migration. I had to cheat in one place, hopefully in a way that isn't problematic. But, most of the effort here was figuring out what is actually needed to programmatically add pages from within a migration.

Hopefully this PR allows:

1. Local databases and unit tests will have a site root that better matches what we have in production.
1. We can run this PR against the production code and the only side effect will be creation of (empty) careers pages (to be filled in by content editors manually) and migration of existing jobs (within the last 6 months) to the new Wagtail page format.
1. If this works, we can copy this model for future encoding of Wagtail content as data migrations.

## Additions

- New migration `v1.migrations.0008_site_root_data` that sets up site root as `CFGov` and also creates regular and content sites. This migration depends explicitly on the `DJANGO_STAGING_HOSTNAME` environment variable that we document in our `.env_SAMPLE` and appear to set in our various deploy JSON configs. It optionally allows for use of `DJANGO_HTTP_PORT` and defaults to port 80.
- New migration `jobmanager.migrations.0007_create_careers_pages` that sets up "About Us" (if it doesn't exist) along with the tree of Careers pages:
 - `/careers` (SublandingPage)
 - `/careers/working-at-cfpb` (BrowsePage)
 - `/careers/application-process` (BrowsePage)
 - `/careers/students-and-graduates` (BrowsePage)
 - `/careers/current-openings` (BrowsePage)
- New migration `jobmanager.migrations.0008_migrate_job_pages` that migrates existing Django `jobmanager.models.django.Job` objects to `jobmanager.models.pages.JobListingPage` pages.

This PR is a replacement for closed #2340.

## Removals

- Removed Events pages from `initial_data` script as they don't seem to be used by any tests and it's not clear why they are in there. Please let me know if there's a reason they need to exist there.

## Changes

- Added the new `v1.migrations.0008_site_root_data` migration to our Python unit test setup so that all unit tests can depend on having a `cfgov` root. Modified a few existing tests who relied on a site root named `home-page`.

## Testing

- Run unit tests both with database migrations and (`tox`) and without (`tox -e fast`).
- Testing with a local database:
 - Create a new database with `./drop-db.sh; ./create-mysql-db.sh; ./initial-data.sh`.
 - Log into the Wagtail admin and note that you have a `CFGov` root with both a `localhost` and `content.localhost` `Site` running on port 8000.
 - Also note that you have a set of (empty) careers pages.
- Testing with a production database:
 - Download a recent production data dump.
 - Browse around the Wagtail admin and note how the root is set up ("Live") and how there's an "About Us" page with no "Careers" page underneath it.
 - Run migrations against your local database (it should run these new `v1` and `jobmanager` migrations).
 - Note that a "Careers" page has been added with several pages underneath, including a whole bunch of recent job postings. All new pages should be in "Draft" state.
 - Try hitting [`/about-us/`](http://localhost:8000/about-us/careers/) and notice that pre-existing "About Us" content is not affected by this migration.
 - Try hitting [`/about-us/careers/`](http://localhost:8000/about-us/careers/careers) and [`/about-us/careers/current-openings`](http://localhost:8000/about-us/careers/current-openings) and note how these pages are still served from Django even though the pages exist in the Wagtail tree. This is desired behavior.

## Review

- @richaagarwal 

## Notes

- All migrations are wrapped in `@transaction.atomic` to prevent a database error from leaving the DB in an unpredictable state.

## Todos

- Add changes to CHANGELOG.

## Checklist


cfgov/v1/util/migrations.py
```diff
@@ -0,0 +1,59 @@
...
+from django.db import transaction
+from treebeard.mp_tree import MP_Node
+
+
+def get_page(apps, slug):
+    base_page_cls = apps.get_model('wagtailcore', 'Page')
+    return base_page_cls.objects.get(slug=slug)
+
+
+def get_free_path(apps, parent_page):
```
chosak: This terribleness is mainly due to the fact that you [can't really](http://stackoverflow.com/a/33807804) just do `parent_page.add_child` as we might like. So instead of doing that, you have to create the model instance in the DB with the proper fields, one of which is this `path` value. Pages in Wagtail have paths like `00010003000L` (note use of non-hexadecimal character) so this function just chooses a valid path for each new page being added.

willbarton: This sent me down the rabbit hole of treebeard (which I've never used before) — it's pretty neat, but I agree, I don't see a better way to do this right now.

:+1: for including it in a place we can re-use it. This seems like something we could need to do more often as we start using more data migrations.

cfgov/v1/util/migrations.py
```diff
@@ -0,0 +1,59 @@
...
+    offset = 1
+    base_page_cls = apps.get_model('wagtailcore', 'Page')
+
+    while True:
```
vccabral: This while loop can be iterative.

vccabral: And I recommend removing the try catch. 

mode.filter(x=y).exists()

cfgov/v1/util/migrations.py
```diff
@@ -0,0 +1,59 @@
...
+
+@transaction.atomic
+def get_or_create_page(apps, page_cls_app, page_cls_name, title, slug,
+                       parent_page, live=False, shared=False, **kwargs):
+    try:
+        return get_page(apps, slug)
+    except ObjectDoesNotExist:
+        pass
+
+    ContentType = apps.get_model('contenttypes.ContentType')
```
vccabral: You can grab these once in the main migraiton script and pass them in so this line is not executed multiple times. Not that important.

cfgov/v1/util/migrations.py
vccabral: are there mixed spaces and tabs?

vccabral: Any of these methods that are super valuable (read might be used for a later migration) should be pulled out to common libs.

cfgov/jobmanager/tests/test_job_converter.py
```diff
@@ -0,0 +1,53 @@
+import importlib
+
+from decimal import Decimal
+from django.test import TestCase
+from model_mommy import mommy
+from wagtail.wagtailcore.models import Page
+
+from jobmanager.models.django import Job
+
+
```
vccabral: great job testing.

cfgov/jobmanager/migrations/0008_migrate_job_pages.py
```diff
@@ -0,0 +1,125 @@
...
+
+        return [
+            panel_cls(job_listing=page, grade=grade)
+            for grade in job.grades.all()
+        ]
+
+    def make_region_panels(self, page, job):
+        panel_cls = self.apps.get_model('jobmanager', 'RegionPanel')
+
+        return [
```
vccabral: 👍 comprehensions are bomb.

cfgov/jobmanager/migrations/0008_migrate_job_pages.py
```diff
@@ -0,0 +1,125 @@
...
+    def make_grade_panels(self, page, job):
+        panel_cls = self.apps.get_model('jobmanager', 'GradePanel')
+
+        return [
+            panel_cls(job_listing=page, grade=grade)
+            for grade in job.grades.all()
+        ]
+
+    def make_region_panels(self, page, job):
+        panel_cls = self.apps.get_model('jobmanager', 'RegionPanel')
```
vccabral: maybe set this up  in the init.

cfgov/jobmanager/migrations/0008_migrate_job_pages.py
```diff
@@ -0,0 +1,125 @@
...
+            link_cls(
+                job_listing=page,
+                announcement_number=at.announcement_number,
+                url=at.usajobs_url,
+                applicant_type=at.application_type,
+            ) for at in applicant_types
+        ]
+
+    def make_grade_panels(self, page, job):
+        panel_cls = self.apps.get_model('jobmanager', 'GradePanel')
```
vccabral: maybe set this up  in the init.

cfgov/jobmanager/migrations/0008_migrate_job_pages.py
```diff
@@ -0,0 +1,125 @@
...
+                job_listing=page,
+                address=at.announcement_email,
+                label=at.application_type.applicant_type,
+                description=at.application_type.description
+            ) for at in job.jobapplicanttype_set.filter(is_usajobs=False)
+        ]
+
+    def make_usajobs_link_panels(self, page, job):
+        applicant_types = job.jobapplicanttype_set.filter(is_usajobs=True)
+        if not all(at.announcement_number for at in applicant_types):
```
vccabral: beautiful use of all

cfgov/v1/util/migrations.py
chosak: @vccabral thanks for the feedback.

I'm not sure how to make that while loop iterative, considering it has to increase until it finds a free number. You could do something like this:

```py
import itertools

for i in itertools.count():
    # do stuff and return if i is free
```

but my sense is that this is less clear, especially to people who aren't familiar with the somewhat obscure [`itertools.count`](https://docs.python.org/2/library/itertools.html#itertools.count).

I don't believe there are mixed spaces and tabs - what makes you think so?

The use of this:

```py
try:
    Model.objects.get(foo=bar)
except Model.DoesNotExist:
    # do something
```

instead of

```py
if not Model.objects.filter(foo=bar).exists():
```

seems somewhat of a style choice to me. I think the first way reads a little clearer. I've also seen the Django codebase use the `try/except` style, for example [here](https://github.com/django/django/blob/1.8.13/django/views/generic/detail.py#L51) or [here](https://github.com/django/django/blob/4f336f66523001b009ab038b10848508fd208b3b/tests/select_related/tests.py#L25).

Regarding pulling out to a common lib, that's why I put these methods in `v1.util.migrations` instead of the migration file itself.

cfgov/v1/util/migrations.py
chosak: Thanks for the suggestion. That didn't seem worth it to me considering the number of arguments that this function already takes, plus the fact that this is a migration of a handful of pages so the efficiency increase wouldn't be significant.

richaagarwal:
👍  Thanks for the thoughtful, explanatory PR!

richaagarwal:
Could you update the CHANGELOG before this gets merged in?

cfgov/v1/util/migrations.py
vccabral: 👍

cfgov/v1/util/migrations.py
vccabral: For spaces, I was looking at the whole file, not the lines you added. The rest is just preference.

PR #2380, opened 2016-09-08
Enforce unique user emails when creating/editing users in Wagtail

chosak:
We want to avoid having users share the same email address. This PR prevents Wagtail editors from creating users with an email that's already in use. It also prevents them from modifying a user's email to one that's already in use by somebody else.

This PR does not prevent use of the Django shell to set duplicate emails, nor does it create a DB constraint that would prevent this (there might be reasonable dev use cases to support something like this, for example easy local testing). It also doesn't do anything to modify any pre-existing users that may already share the same email.

## Additions

- New custom `v1.auth_forms.UserEditForm` that has custom email validation to prevent duplicates.

## Changes

- Our custom `v1.auth_forms.UserCreationForm` now has custom email validation to prevent duplicates.

## Testing

- Run `tox` or `tox -e fast` to run all unit tests. New tests for this are in `v1.tests.test_auth_forms`.
- Testing with a local database:
 - Go to [the Wagtail User editor](http://localhost:8000/admin/users/) and select "Add a user".
 - Fill out all of the information, including a unique email address. Select "Roles" and then "Add user" to save.
 - Try repeating the process with a second user using the same email address but other unique information. Validation should fail and prevent you from saving the second user.
 - Try repeating the process with a second unique email (this should work), and then go back and edit this new second user by clicking on its username in the user list.
 - Try changing the second user email to the first unique email address you used. Validation should fail and prevent you from updating the second user.

## Review

- @grapesmoker 

## Screenshots

Can't add a user with a duplicate email address:

image

Can't edit a user to set a duplicate email address:

image

## Checklist

### DIFF
--- a/cfgov/v1/tests/test_auth_forms.py
+++ b/cfgov/v1/tests/test_auth_forms.py
@@ -1,8 +1,8 @@
 from django.contrib.auth.models import User
+from django.test import TestCase
 from mock import patch
-from unittest import TestCase
 
-from v1.auth_forms import UserCreationForm
+from v1.auth_forms import UserCreationForm, UserEditForm
 
 
 @patch('v1.auth_forms.send_password_reset_email')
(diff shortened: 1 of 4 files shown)


cfgov/v1/tests/test_auth_forms.py
```diff
@@ -34,3 +34,79 @@ def test_save_without_commit_doesnt_send_email(self, send_email):
...
         send_email.assert_not_called()
+
+    def test_duplicate_email_fails_validation(self, send_email):
+        User.objects.create(username='foo', email=self.email)
+        form = UserCreationForm(self.userdata)
+        self.assertFalse(form.is_valid())
+        self.assertTrue(form.errors['email'])
+
+
+class UserEditFormTestCase(TestCase):
```
grapesmoker: Just a stylistic question: would it not be better to factor out the common stuff like the `data` dict into setUp?

chosak: Good idea, thanks. Addressed in 317e5e2.

richaagarwal:
👍

richaagarwal:
Oops, actually, I just tested this locally and it let me create a user with the same e-mail address as one that exists already.

richaagarwal:
So this also doesn't prevent a duplicate e-mail address being entered in from the django admin panel - is that an oversight or intentional?  I *think* that's still where we primarily add new users, @rosskarchner do you know which one we use?

rosskarchner:
Users get created in both places (for instance, some users will only ever be "Django" users, and never have a need for wagtail)

rosskarchner:
Maybe a signal could be registered, that throws an exception if the email address is a duplicate?

chosak:
@rosskarchner I considered that but thought there might be some valid use cases for using the same email address, for example when developers are testing locally. It seemed a bit overkill to me to absolutely prevent someone from e.g. creating multiple superusers with the same email address using `manage.py createsuperuser`, or even using the Python shell to create users manually. If we want to enforce that, I can create another PR that does that.

As the title of this PR says: "Enforce unique user emails when creating/editing users in Wagtail", so this was purposefully only preventing this activity in Wagtail.

richaagarwal:
@chosak 
In response to " It seemed a bit overkill to me to absolutely prevent someone from e.g. creating multiple superusers with the same email address using manage.py createsuperuser, or even using the Python shell to create users manually. If we want to enforce that, I can create another PR that does that." -  I agree we don't necessarily need to enforce it locally, but the django admin panel is actually separate from that (sorry if this was confusing!).  This is what I was referring to: http://localhost:8000/django-admin/

I'm fine with this being merged in if we want to separate that work, but I thought it might be related in terms of how this code is architected.

rosskarchner:
sounds good!

chosak:
Great. I'll merge this and add a subsequent task.

PR #2381, opened 2016-09-08
Faster Travis through removal of non-critical steps

chosak:
The current Travis process includes several steps that are either not critical for testing or are just redundant. This PR streamlines the process by removing these extra steps, lowering build time from 10-11 minutes to more like 5-6 minutes.

Steps removed include:

1. Installation of virtualenv, virtualenvwrapper, and autoenv. These shouldn't be necessary as Travis already sets up a virtual environment for the build, and tox sets up its own virtualenv as well when running the Python tests.
1. Running of `setup.sh` which in turns called `backend.sh` to run a local database and install all backend dependencies. Because tox uses its own test database, and the JS tests don't depend on the database anymore, this step is unnecessary.
1. Running of `manage.py sheer_index` to load content from Wordpress into the DB. This content is no longer depended on by any tests.
1. Running of `manage.py migrate` to migrate the database. The test DB created by tox already has all migrations applied.

## Changes

- Travis steps optimized as described above.

## Testing

- Look at the Travis logs and note the runtime as compared to other open PRs.

## Review

- @willbarton @cfpb/cfgov-backends 

## Notes

- This PR assumes that we don't want to test `sheer_index` as part of our Travis build. If this is a critical part of the system that we want regular testing for, we should discuss whether Travis is the best place for it or if it makes more sense somewhere else.
- In the past there were browser tests that relied on a database with site content and this might be a good goal to have on Travis in the future.

## Checklist


chosak:
For the record, adding this section to the `.travis.yml` section didn't actually speed anything up (both builds between 9-10 minutes). Will need to look more deeply into the log to understand what's happening.

vccabral:
@chosak what is this doing? It looks cool, want to know.

chosak:
I originally opened this PR to look into caching, but @higs4281 beat me to the punch (for pip) in #2404. 

It is possible to cache the `node_modules` directory if we want (although this might require changing `frontend.sh` or calling it differently to avoid its `clean` step). From some research I've seen other repos that [decided not to do this](https://github.com/jquery/jquery/issues/2541) as it might cause problems with old versions. I don't think this is as big of a problem with Python as long as we pin all of our requirements. If requirements are pinned for our node packages, is this something we could reasonably do, or are there risks? @contolini

richaagarwal:
I thought we don't even use `sheer_index` anymore, right @rosskarchner ?

richaagarwal:
👍

contolini:
@chosak Caching `node_modules` sounds great to me! We shrinkwrap npm dependencies so it should work fine.

chosak:
@contolini going to merge this for now to limit scope and leave caching of `node_modules` for a future task.

PR #2383, opened 2016-09-08
Adds migration 0008 for 'Related content' label

higs4281:
PR 2377 involved changes that generated this migration.

## Additions

- v1 migration 0008

## Review

- @Scotchester @chosak 


chosak:
:+1: with two minor (non-blocking) thoughts:

1. Is the additional `CHANGELOG` message necessary? We already have a separate entry for #2377 and this is more of an oversight that should have gone into that one. We also don't tend to list migrations there.
2. @richaagarwal had a good suggestion that we try to name our migrations something descriptive so that it's easy to tell what they do by looking at their filenames. So instead of `0008_auto_20160908_1401.py`, maybe something like `0008_rename_related_links.py`.

chosak:
👍

PR #2385, opened 2016-09-08
HOLD: Refactor context formation for pages

kurtrwall:
# Holding in favor of breaking this down
## Continuation of #2301
This refactors the way that context is set for pages. This is a rather large PR since it was combined with the above linked PR.

## The benefits of this PR:
- Decouple form logic with page type. We no longer need to create a separate page just because we want to include a form on it.
- Abstracts code away from CFGOVPage.
- Utilizes wagtail's hook functionality for inclusion
- Sets the stage for other modules coming

## What this does:
- moves "filterable" and js file context logic from util module into a handlers module
- handlers are registered in `CFGOVPage.get_context`, which is called for every page load, via `wagtail_hooks.py`
- refactors both the filterable and page js logic a bit
- provides wider test coverage

## Additions

- More tests!
- `wagtail_hooks.py` registered functions for `cfgov_context_handlers` hook to manipulate page context
- handlers module to house them
- `QueryFormMixin` to abstract the weird query stuff we're doing in the filterable form classes

## Removals

- test files where the tests are now handled elsewhere (nothing was removed that wasn't replaced)

## Changes

- `FilterableListMixin` into a handler class
- `FilterControls` is now `FilterableListControls` 
- `has_active_filters` moved to util module
- all context logic for filterable and JS handling moved from the page class to handlers
- `FilterableListControls` moved `page_type` from within `categories` StructBlock to at the base level.
- `NumberBlock` was refactored a bit for filter controls module

## Testing

- `cfgov/manage.py migrate`
- `tox`
- Make sure you go to the blog, newsroom, activity log, events archive, any browse filterable, and any sublanding filterable page. Test various filtering methods to make sure they all still work.
- Make sure JS handling still works by visiting or creating a page that has expandables on it. The expandable should work.

## Review

- @cfpb/cfgov-backends 
- @Scotchester 

## Notes

- This work is necessary to build the components necessary for Design Ops
- The Data & Research (Dada) team will be working off of a branch with these commits in order to move forward.

## Todos

- Since we no longer need them, we should remove the [Sublanding/Browse]FilterablePage classes, along with their subclasses. I felt that it would be a bit too extreme a change to include in the PR. Let me know if we should just include it.

## Checklist


chosak:
There's a lot to review here, but to start with, it would be useful to understand a little more about the rationale behind these changes. Is there an associated GHE user story or task that this addresses?

Could this PR be broken down into smaller, more easily reviewable pieces?

One risk of a large change like this with so many files is that it might break existing functionality. In the testing instructions you suggest

 > Test various filtering methods to make sure they all still work.

Is this a good opportunity to create unit tests that check this?

kurtrwall:
> Is there an associated GHE user story or task that this addresses?

Not directly, but these are factors [GHE]/CFGOV/data-research/issues/168 , [GHE]/CFGOV/data-research/issues/164, as well as design ops work but i can't find the issues specifically.

> Could this PR be broken down into smaller, more easily reviewable pieces?

Originally (#2301), it was.

> Is this a good opportunity to create unit tests that check this?

I think this definitely raises the issue that we don't have enough tests surrounding the actual results of filtering, but these include the relevant tests for the changes this PR introduces.

cfgov/v1/forms.py
```diff
@@ -71,7 +76,60 @@ def clean(self):
...
+        return self.queries.get(self.query_type)
+
+    # Generates a query by iterating over the zipped collection of
+    # tuples.
+    def _generate_query(self):
+        final_query = Q()
+        if self.is_bound:
+            q_tuples = zip(self._get_query_strings(), self.declared_fields)
+            for query, field_name in q_tuples:
+                field = self.cleaned_data.get(field_name, None)
```
vickumar1981: don't need None here.

cfgov/v1/forms.py
```diff
@@ -71,7 +76,60 @@ def clean(self):
...
+                'authors__name__in',     # authors
+            ],
+        })
+
+
+    # Returns a list of query strings to associate for each field, ordered by
+    # the field declaration for the form. Note: THEY MUST BE ORDERED IN THE
+    # SAME WAY AS THEY ARE DECLARED IN THE FORM DEFINITION.
+    def _get_query_strings(self):
+        return self.queries.get(self.query_type)
```
vickumar1981: might want to do the `zip` here and call the function `_get_query_fields`.

cfgov/v1/forms.py
```diff
@@ -175,71 +236,42 @@ def set_field_html_name(self, field, new_name):
...
-            'start_dt__gte',         # from_date
-            'end_dt__lte',           # to_date
-            'categories__name__in',  # categories
-            'tags__slug__in',        # topics
-            'authors__slug__in',     # authors
-        ]
-
+    def __init__(self, *args, **kwargs):
+        super(EventArchiveFilterForm, self).__init__(*args, **kwargs)
+        self.query_type = 'event-archive'
```
vickumar1981: i know theres only 2 "query types", but if this list does grow, or just to make it easier for future people to now what the available options are for this, you could move it to an enum class - that lists what the possible values might be.

cfgov/v1/migrations/0008_filter_controls_move_page_type.py
```diff
@@ -0,0 +1,76 @@
...
+    for model_name in ['SublandingFilterablePage', 'BrowseFilterablePage']:
+        model = apps.get_model('v1.%s' % model_name)
+        for page in model.objects.all():
+            update_page(page)
+            revisions = get_revisions_to_update(page, Revision)
+            for revision in revisions.values():
+                update_revision(revision)
+
+def update_page(page):
+    for i, block in enumerate(page.content.stream_data):
```
vickumar1981: doesn't look like `i` is used, which means you could just iterate `for block in ...`

cfgov/v1/wagtail_hooks.py
```diff
@@ -126,3 +130,36 @@ def expand_db_attributes(attrs, for_editor):
...
+    handler.process()
+
+
+@hooks.register('cfgov_context_handlers')
+def register_filterablelist_handlers(*args):
+    base_handler = Handler(*args)
+    def get_filter_blocks():
+        block_tuples = []
+        blocks_dict = base_handler.get_streamfield_blocks()
+        for blocks_list in blocks_dict.values():
```
vickumar1981: use a comprehension here.

cfgov/v1/util/util.py
```diff
@@ -143,3 +142,11 @@ def all_valid_destinations_for_request(request):
                             valid_destination_for_request(request, pair[1])]
 
     return valid_destinations
+
+
+def has_active_filters(request, index):
+    for key, param in request.GET.items():
```
vickumar1981: use a comprehension here.  also, i think the if blocks can be combined using an and.

cfgov/v1/util/util.py
```diff
@@ -27,11 +27,10 @@ def get_unique_id(prefix='', suffix=''):
 }
 
 def get_form_id(page):
-    form_ids = page.get_filter_ids()
-    if form_ids:
-        return form_ids[0]
-    else:
-        return 0
+    for i, block in enumerate(page.content):
```
vickumar1981: comprehension here.  i is not used, no need for enumerate.

cfgov/v1/models/handlers/filterable_list_handlers.py
```diff
@@ -0,0 +1,200 @@
...
+        get_blog, only_blog = self._if_and_only_blog(categories)
+        newsroom_q = self._get_newsroom_query(form) if not only_blog else Q()
+        blog_q = self._get_blog_query(form) if get_blog else Q()
+        return newsroom_q | blog_q
+
+    @staticmethod
+    def _if_and_only_blog(categories):
+        get_blog = True
+        only_blog = False
+        if categories:
```
vickumar1981: this if block can be simplified.

cfgov/v1/migrations/0008_filter_controls_move_page_type.py
kurtrwall: whoops, i was doing a delete that i ended up leaving out. thanks for reminding me to take it off.

cfgov/v1/migrations/0008_filter_controls_move_page_type.py
```diff
@@ -0,0 +1,76 @@
...
+    return revisions_to_update
+
+def update_revision(revision):
+    revision_content = json.loads(revision.content_json)
+    data = json.loads(revision_content['content'])
+    for block in data:
+        if block['type'] == 'filter_controls':
+            # import pdb; pdb.set_trace()
+            block['value']['page_type'] = block['value']['categories']['page_type']
+            # del block['value']['categories']['page_type']
```
kurtrwall: Remove.

PR #2387, opened 2016-09-12
fixed a mistake in get_browserfilterable_posts and wrote a test for it

grapesmoker:
The get_browsefilterable_posts function had a mistake in it that would cause an uncaught exception if one of the intermediate calls to get_page_set returned None. That was fixed. Also, tests were written for get_browsefilterable_posts and get_appropriate_descendants functions.

## Additions

* test for get_appropriate_descendants
* test for get_browsefilterable_posts

## Changes

* in get_browsefilterable_posts, the filtered_controls dict accumulates the results of get_page_set via append so that it doesn't throw an error if that result is None.

## Testing

* two tests written as above

## Review

@richaagarwal @chosak @willbarton

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Author names are now displayed in alphabetical order by last name, falls back on first name if necessary
 - Ability to output sharing links within an Image and Text 50/50 Group module
 - Google Optimize code on `find-a-housing-counselor` page
+- Added a test for get_browsefilterable_posts function of the sublanding page 
 
 ### Changed
 - Special characters no longer break the multiselect in the filter form
--- a/cfgov/v1/models/sublanding_page.py
+++ b/cfgov/v1/models/sublanding_page.py
@@ -65,13 +65,15 @@ class SublandingPage(CFGOVPage):
 
     def get_browsefilterable_posts(self, request, limit):
         filter_pages = [p.specific for p in self.get_appropriate_descendants(request.site.hostname)
-                        if 'FilterablePage' in p.specific_class.__name__ and 'archive' not in p.title.lower()]
+                        if 'FilterablePage' in p.specific_class.__name__
+                        and 'archive' not in p.title.lower()]
 
         posts_tuple_list = [(str(util.get_form_id(page)), post)
                             for page in filter_pages
-                            for post in page.get_page_set(page.get_form_class()(parent=page,
-                                                                                hostname=request.site.hostname),
-                                                          request.site.hostname)]
+                            for post in page.get_page_set(
+                                page.get_form_class()(parent=page,
+                                                      hostname=request.site.hostname),
+                                request.site.hostname)]
 
         posts = sorted(posts_tuple_list, key=lambda p: p[1].date_published, reverse=True)[:limit]
         return posts
--- a/cfgov/v1/models/sublanding_page.py
+++ b/cfgov/v1/models/sublanding_page.py
@@ -1,6 +1,3 @@
-from itertools import chain
-from collections import defaultdict
-
 from wagtail.wagtailcore.fields import StreamField
 from wagtail.wagtailadmin.edit_handlers import TabbedInterface, ObjectList, \
     StreamFieldPanel
--- a/cfgov/v1/models/sublanding_page.py
+++ b/cfgov/v1/models/sublanding_page.py
@@ -75,10 +75,8 @@ class SublandingPage(CFGOVPage):
                 filtered_controls.update({id: []})
             form_class = page.get_form_class()
             posts = page.get_page_set(form_class(parent=page, hostname=request.site.hostname), request.site.hostname)
-            if filtered_controls[id]:
-                filtered_controls[id] += posts
-            else:
-                filtered_controls[id] = posts
-        posts_tuple_list = [(id, post) for id, posts in filtered_controls.iteritems() for post in posts]
+            filtered_controls[id].append(posts)
+
+        posts_tuple_list = [(id, post) for id, posts in filtered_controls.iteritems() for post in chain(*posts)]
         posts = sorted(posts_tuple_list, key=lambda p: p[1].date_published, reverse=True)[:limit]
         return posts
--- a/cfgov/v1/models/sublanding_page.py
+++ b/cfgov/v1/models/sublanding_page.py
@@ -1,4 +1,5 @@
 from itertools import chain
+from collections import defaultdict
 
 from wagtail.wagtailcore.fields import StreamField
 from wagtail.wagtailadmin.edit_handlers import TabbedInterface, ObjectList, \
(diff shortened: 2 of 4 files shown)


cfgov/v1/tests/models/test_sublanding_page.py
```diff
@@ -0,0 +1,89 @@
...
+                                         [atomic.filter_controls], True)
+
+        helpers.save_new_page(self.post1, self.sublanding_page)
+        helpers.save_new_page(self.post2, self.sublanding_page)
+
+        # manually set the publication date of the posts to ensure consistent
+        # order of retrieval in test situations, otherwise the `date_published`
+        # can vary due to commit order
+
+        self.child1_of_post1 = AbstractFilterPage(title='child 1 of post 1', date_published=dt.datetime(2016, 9, 1))
```
higs4281: minor: could use dt.date(2016, 9, 1)
the model uses a DateField

cfgov/v1/models/sublanding_page.py
```diff
@@ -75,10 +75,8 @@ def get_browsefilterable_posts(self, request, limit):
                 filtered_controls.update({id: []})
             form_class = page.get_form_class()
             posts = page.get_page_set(form_class(parent=page, hostname=request.site.hostname), request.site.hostname)
-            if filtered_controls[id]:
-                filtered_controls[id] += posts
-            else:
-                filtered_controls[id] = posts
-        posts_tuple_list = [(id, post) for id, posts in filtered_controls.iteritems() for post in posts]
+            filtered_controls[id].append(posts)
+
```
richaagarwal: Won't this throw a KeyError if it doesn't exist?

richaagarwal: Oh, nevermind, just saw that's taken care of in line 74.

cfgov/v1/models/sublanding_page.py
```diff
@@ -75,10 +75,8 @@ def get_browsefilterable_posts(self, request, limit):
                 filtered_controls.update({id: []})
             form_class = page.get_form_class()
             posts = page.get_page_set(form_class(parent=page, hostname=request.site.hostname), request.site.hostname)
```
richaagarwal: posts could still return `None`, right?  The later call to `chain(*posts)` still results in an error if it is `None`.

cfgov/v1/models/sublanding_page.py
```diff
@@ -75,10 +75,8 @@ def get_browsefilterable_posts(self, request, limit):
...
-            if filtered_controls[id]:
-                filtered_controls[id] += posts
-            else:
-                filtered_controls[id] = posts
-        posts_tuple_list = [(id, post) for id, posts in filtered_controls.iteritems() for post in posts]
+            filtered_controls[id].append(posts)
+
+        posts_tuple_list = [(id, post) for id, posts in filtered_controls.iteritems() for post in chain(*posts)]
```
richaagarwal: I know this isn't your doing, but I found it really confusing that `posts` here is different than what is referenced above.  Might be an opportunity to update it, unless I'm the only one this is confusing to.

richaagarwal:
👍

cfgov/v1/models/sublanding_page.py
chosak: This is a good use case for [`defaultdict`](https://docs.python.org/2/library/collections.html#collections.defaultdict), which lets you avoid the check:

```py
filtered_controls = defaultdict(list)
for page in filter_pages:
    ...
    filtered_controls[id].append(posts)
```

cfgov/v1/models/sublanding_page.py
chosak: Agreed. The use of a dictionary here and then using `chain` to iterate over a list of lists seems a bit overly complicated. Could this logic be made a bit more straightforward with something like this?

```py
posts = []

for page in filter_pages:
    page_id = str(util.get_form_id(page))

    form_class = page.get_form_class()
    page_posts = page.get_page_set(
        form_class(parent=page, hostname=request.site.hostname),
        request.site.hostname
    )

    posts.extend((page_id, page_post) for page_post in page_posts)

posts_by_most_recent = sorted(posts, key=lambda p: p[1].date_published, reverse=True)

return posts_by_most_recent[:limit]
```

cfgov/v1/tests/models/test_sublanding_page.py
```diff
@@ -0,0 +1,89 @@
...
+        # this one only has a filter_controls
+        self.post2.content = StreamValue(self.post1.content.stream_block,
+                                         [atomic.filter_controls], True)
+
+        helpers.save_new_page(self.post1, self.sublanding_page)
+        helpers.save_new_page(self.post2, self.sublanding_page)
+
+        # manually set the publication date of the posts to ensure consistent
+        # order of retrieval in test situations, otherwise the `date_published`
+        # can vary due to commit order
```
chosak: Curious about this comment. Is it the order that varies? Or just the fact that the dates will be the same because these are all published at the same time?

chosak:
:+1: This looks good to me. One minor request, can you run `flake8` or something similar against `test_sublanding_page.py` to clean up the formatting?

cfgov/v1/tests/models/test_sublanding_page.py
grapesmoker: Sure, I'll change this.

cfgov/v1/models/sublanding_page.py
grapesmoker: Sure. I made the smallest change consistent with the behavior as I understood it, but I can certainly pretty it up.

cfgov/v1/models/sublanding_page.py
grapesmoker: Good point.

cfgov/v1/tests/models/test_sublanding_page.py
grapesmoker: What I found was that when I ran the test several times in a row, I would get different results, i.e. occasionally the test would fail when previously it had succeeded (without anything being changed). When I looked at the order of the posts returned from get_browsefilterable_posts, I discovered that it would vary. It looks like the cause of this is that if you do not supply the date_published argument, it will be set to the current date, and when that happens the retrieval order will depend on, presumably, the order of commit. One way of making sure this doesn't become an issue for actual posts is to change the DateField that @higs4281 was referring to into a DateTimeField, which would save us from relying on commit order for posts that were published on the same date. For testing purposes, I just supplied a date manually so that the retrieval order would be consistent.

grapesmoker:
Are we scrupulously observing line length? It annoys me that flake complains about a line that is 80 characters in length when splitting that line in two makes for more difficult reading than just keeping it as is. Also in some cases there is a visual indent that flake will complain about if you remove it in order to reduce line length. If people have really strong opinions about this I'll do it, but honestly I would much rather overrun the line length by a few characters and keep the visual coherence of the calls intact than try and satisfy flake.

richaagarwal:
@grapesmoker Yeah, I think those are good reasons for ignoring the line length suggestion.  AFAIK, we're treating it as a suggestion as opposed to a requirement.

chosak:
There's also some previous line length discussion [here](https://github.com/cfpb/back-end/issues/14#issuecomment-211378971). I wouldn't consider indentation a blocker, but for lines like [this](https://github.com/cfpb/cfgov-refresh/pull/2387/files#diff-2ba27f15d5ce948b3f9483f62581c805R70) consider that it can be difficult to look at the code when an editor is divided horizontally into two buffers on small (laptop) screens:

image

grapesmoker:
I tried to cut down the line lengths as much as possible, mostly in the comments and in that one generator expression that @chosak posted a picture of. I get that line length can be an issue with an IDE like emacs (which I use too) but sometimes there just isn't a better aesthetic alternative than to run a bit over. Hopefully the changes I've made help.

chosak:
@grapesmoker agreed, and thanks. :+1:

higs4281:
👍

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Author names are now displayed in alphabetical order by last name, falls back on first name if necessary
 - Ability to output sharing links within an Image and Text 50/50 Group module
 - Google Optimize code on `find-a-housing-counselor` page
+- Added a test for get_browsefilterable_posts function of the sublanding page
```
richaagarwal: @grapesmoker Already merged this in, but for future reference - I don't think you need to add tests to the `CHANGELOG`; in this case I asked you to update it because you fixed something, so that'd go under the "Fixed" section.

grapesmoker: Oh, my bad. I didn't quite understand what you meant. Sorry, I can push a change to `master` to fix it.

richaagarwal: Oh no worries!  Sure, feel free to.

richaagarwal: If it's helpful we can go over what the `CHANGELOG` is meant to capture at a high-level in scrum or something like that.

PR #2388, opened 2016-09-12
Set up logging locally

richaagarwal:
So that we don't have to use print statements.  I updated `archive_events` as an example, although I don't actually think we use it anymore. 

Run `python cfgov/manage.py runscript archive_events` to see the logging in action.

@cfpb/cfgov-backends

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 ## Unreleased Changes
 
 ### Added
+- Logging configuration to `local.py`
 - Author names are now displayed in alphabetical order by last name, falls back on first name if necessary
 - Ability to output sharing links within an Image and Text 50/50 Group module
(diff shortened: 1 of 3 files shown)


rosskarchner:
Thanks for figuring this out! 👍 

I would have expected to see the "No events to archive found...." message, but that's probably unrelated to your work here.

grapesmoker:
Where does the log information get logged to?

rosskarchner:
@grapesmoker the console-- which makes sense for local users, at least.

Scotchester:
🚨 n00b alert 🚨 

If it's just logging to the console, why is it better than using `print` statements?

rosskarchner:
In deployment, logging can be configured to go elsewhere

Scotchester:
Ahah.

rosskarchner:
... and then, it's a settings change, the call to `logger.info` serves both purposes

chosak:
:+1:

PR #2389, opened 2016-09-12
Feature flag decorators

chosak:
We want to be able to support conditional URLs and views based on feature flag state. The first use case for this is the careers pages as they transition from Django to Wagtail. Implementing feature flag decorators will support this use cases and others in future.

This PR lets you decorate views like this:

```py
from flags.decorators import flag_required

@flag_required('FLAG_NAME')
def my_view(request, *args, **kwargs):
    ...
```

By default, views will return 404 if the flag is not set. You can also specify a fallback view, for example to Wagtail:

```py
from transition_utilities.conditional_urls import wagtail_fail_through

@flag_required('FLAG_NAME', fallback_view=wagtail_fail_through)
def my_view(request, *args, **kwargs):
    ...
```

Class-based views can also be decorated this way, or can inherit from the `flags.views.FlaggedViewMixin` mixin to enable this behavior (see tests for examples).

This PR contains no careers-specific code, but is the foundation for putting the careers pages behind a flag.

## Additions

- Lots of new unit tests in `flags.tests`.
- Better related names for `Site` flag-related foreign keys (`flags` and `flag_states`).
- New `flags_enabled` method for checking multiple feature flags at once.

## Changes

- We now enforce a unique pair of (`Site`, flag state) at the database level.
- Feature flag functions (like `flag_enabled`) now require an initial `request` parameter which is an `HttpRequest` object (see "Notes" below).
- Updated documentation.

## Testing

- Run `tox flags.tests` or `tox -e flags.tests`.
- Regenerate the documentation by running `mkdocs serve` (after doing `pip install -r requirements/manual.txt`.
- Using a new database, start your local server and visit the home page. Note the beta site banner. Go to the flag editor [in the Wagtail admin](http://localhost:8000/admin/flags/1/), uncheck the `BETA_NOTICE` checkbox, and select "Save Flags". Reload the home page and note that the site banner is gone.

## Review

- @rosskarchner @cfpb/cfgov-backends 

## Screenshots

Feature flag docs:
image

## Notes

- I modified the behavior of `flag_enabled` and friends to take an initial `request` parameter. The existing code was assuming that the local thread had an `HttpRequest` that was set by the `sheerlike.middleware.GlobalRequestMiddleware`. This caused difficulty when testing, though, and also seemed a bit too magical to me versus just passing the request object in directly. I believe that we should always have access to the request when rendering templates or template blocks; if not, we should fix that. Let me know if you think that's not a reasonable change; one option would be to default `request=None` and check the local thread if one isn't passed in, but this seems error-prone to me.

## Checklist


willbarton:
This looks awesome!

@chosak my only thought before having tested locally is it would be nice to have what you wrote in the PR description in docs/feature-flags.md to document `@flag_required`.

chosak:
@willbarton thanks for the suggestion, implemented in 	0648d6f. Here's how the documentation looks with that change:

image

rosskarchner:
This looks good, generally-- though I have mixed feelings about expecting the 'request' to be globally available, since that's not the Django default. We implemented it to be able to support templates written for Sheer, which DID have a globally available request object, (which I think comes free with Flask)

chosak:
@rosskarchner when you say globally available, do you mean in the template context? The existing code assumes that the local thread object has had the request object stuck on it, using the middleware [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/sheerlike/middleware.py#L9). This seemed a bit hidden and bug-prone to me. Since we need the request no matter what (to get the site), it seemed to me to be a choice between "make sure we have the request where we need it" or "assume that the request has been made available in some other code disconnected from here".

rosskarchner:
@chosak yeah, in the template context. The inability to do it truly globally, is what has lead to hacks like this: https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/urls.py#L256

chosak:
@rosskarchner interesting. We [do have](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/base.py#L117) the [`RequestContext`](https://docs.djangoproject.com/en/1.8/ref/templates/api/#subclassing-context-requestcontext) context processor specified. I thought that always made the `request` object available, at least for Django templates.

This doesn't help us for Wagtail, unfortunately ([here's](https://groups.google.com/forum/#!topic/wagtail/3BkKoKqTqeM) someone that encountered a similar issue). @kurtrwall did previously setup [a custom context function](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/__init__.py#L139) that updates the Wagtail block context before it is rendered. This approach might already give `request` to Wagtail blocks, although one downside is that it has to be called manually by templates when blocks are included.

rosskarchner:
ah-- interesting. I wonder if there's a supported way of using django.template.context_processors.request with Jinja2

chosak:
@rosskarchner using context processors with Jinja2 [is going to be supported](https://docs.djangoproject.com/en/dev/releases/1.11/#templates) in  Django 1.11.

The documentation for the Jinja2 processor does include a note ([as of 1.9](https://docs.djangoproject.com/en/1.9/topics/templates/#django.template.backends.jinja2.Jinja2)) that the context includes the `request` object if rendered with one. [This doc fix](https://code.djangoproject.com/ticket/25804) seems to indicate that it already existed in 1.8 but just wasn't documented. So we should have it, if rendering with `render()`.

rosskarchner:
Hmm, OK-- maybe we need to experiment with not trying to inject it into the context ourselves.

rosskarchner:
👍

willbarton:
This looks great @chosak!

One final thing, can you update the CHANGELOG before merging?

chosak:
@rosskarchner another thing I just noticed - Wagtail 1.6 introduces [a new mechanism](http://docs.wagtail.io/en/v1.6/releases/1.6.html#include-block-tag-for-improved-streamfield-template-inclusion) fore rendering StreamFields, which passes the page context. @kurtrwall this may also be relevant to your interests.

PR #2394, opened 2016-09-14
Fixed a few different bugs with form validation

jimmynotjim:
The email signup wasn't running the input through validation as well as not using the API to return a notice instead of loading a new page on submission. Fixed the API issue by updating the elem query to a class name and fixed the validation by removing unneccesary returns.

## Fixes

- Fixed an issue w/ the email signup not utilizing the Gov Delivery JSON view for instant error/success feedback.
- Fixed an issue w/ the form validation skipping any validation types other than `required`.

## Testing

- navigate to `http://localhost:3000/about-us/blog/` and enter a malformed email in the subscribe form. If you're using Chrome, it'll automatically flag any mistake other than `string@string` so try that. 

## Review

- @KimberlyMunoz 
- @sebworks 

## Screenshots

screen shot 2016-09-14 at 9 18 52 am

## Notes

- The email signup organism should eventually be merged with the the form organism and the heading moved within the form. The email signup specific styles would then be moved to a 'o-form__email-signup` modifier. The modifier put in place here is only the first step, nothing was changed to the styles.
- Validation and notifications should eventually be updated to use the Atomic Component framework @sebworks has created, but that's far out of scope here.

## Checklist


sebworks:
@jimmynotjim, @KimberlyMunoz, how would you guys feel removing the client side validation and just relying upon the browser?

jimmynotjim:
@sebworks that's already how this works. The browsers implement theirs before the form is actually submitted, so this is a backup for cases where validation isn't supported (which is everything except for Chrome at the moment).

sebworks:
So I'm suggesting just rely upon the validation that comes with setting the input type as `email`. It's pretty well supported: http://www.wufoo.com/html5/types/1-email.html. It won't catch cases such as `foo@example` but that's ok; that can be handled by the server. We can go a step further and add a `pattern` [attribute](http://caniuse.com/#search=pattern. http://stackoverflow.com/questions/5601647/html5-email-input-pattern-attribute):  ( with regex that matches that of the one on the server ) but I don't think it's necessary.

We should start moving towards relying upon what the browsers support natively, and removing as much Javascript as possible.

jimmynotjim:
@sebworks I totally agree, but this is more than just email inputs, this is also for "required" fields including checkboxes and radios and is necessary for other work that D&R will be doing. It also fixes the response from gov delivery so that you're not forced into a new window.

jimmynotjim:
@sebworks also enter any string into Safari or Firefox on `http://www.consumerfinance.gov/about-us/blog/` and you'll see it submits without any validation at all. Until other browsers catch up to Chrome, I still think we need the email bit.

screen shot 2016-09-15 at 11 31 21 am

sebworks:
> Until other browsers catch up to Chrome, I still think we need the email bit.

👍 .  Merge at will.

jimmynotjim:
@sebworks thanks for the review, hopefully we can refactor this with your framework soon.

PR #2396, opened 2016-09-14
Refactor page JS inclusion via Wagtail hooks

kurtrwall:
This changes the inclusion of page/streamfield media object into the jinja2 context. It just pulls the code from the CFGOVPage and puts it into a handler class for cleaner code.

## Additions

- Handler and JSHandler classes

## Changes

- Move JS handling logic into JSHandler class from CFGOVPage

## Testing

- `tox`

## Review

- @chosak 
- @Scotchester 
- @cfpb/cfgov-backends 

## TODO 

- Add tests for JSHandler.set_block_js method.
#
## Checklist

### DIFF
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -365,62 +365,17 @@ class CFGOVPage(Page):
         user_perms = CFGOVUserPagePermissionsProxy(user)
         return user_perms.for_page(self)
 
-    class Meta:
-        app_label = 'v1'
-
     def parent(self):
         parent = self.get_ancestors(inclusive=False).reverse()[0].specific
         return parent
 
     # To be overriden if page type requires JS files every time
     # 'template' is used as the key for front-end consistency
-    def add_page_js(self, js):
-        js['template'] = []
+    def get_page_js(self):
+        return []
 
-    # Retrieves the stream values on a page from it's Streamfield
-    def _get_streamfield_blocks(self):
-        lst = [value for key, value in vars(self).iteritems()
-               if type(value) is StreamValue]
-        return list(chain(*lst))
-
-    # Gets the JS from the Streamfield data
-    def _add_streamfield_js(self, js):
-        # Create a dictionary with keys ordered organisms, molecules, then atoms
-        for child in self._get_streamfield_blocks():
-            self._add_block_js(child.block, js)
-
-    # Recursively search the blocks and classes for declared Media.js
-    def _add_block_js(self, block, js):
-        self._assign_js(block, js)
-        if issubclass(type(block), blocks.StructBlock):
-            for child in block.child_blocks.values():
-                self._add_block_js(child, js)
-        elif issubclass(type(block), blocks.ListBlock):
-            self._add_block_js(block.child_block, js)
-
-    # Assign the Media js to the dictionary appropriately
-    def _assign_js(self, obj, js):
-        try:
-            if hasattr(obj.Media, 'js'):
-                for key in js.keys():
-                    if obj.__module__.endswith(key):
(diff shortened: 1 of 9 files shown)


cfgov/v1/models/base.py
```diff
@@ -225,6 +226,12 @@ def get_next_appropriate_siblings(self, hostname, inclusive=False):
     def get_prev_appropriate_siblings(self, hostname, inclusive=False):
         return self.get_appropriate_siblings(hostname=hostname, inclusive=inclusive).filter(path__lte=self.path).order_by('-path')
 
+    def get_context(self, request, *args, **kwargs):
```
chosak: This is a great approach. I really like the use of Wagtail hooks to customize the context.

cfgov/v1/wagtail_hooks.py
```diff
@@ -126,3 +127,9 @@ def expand_db_attributes(attrs, for_editor):
 @hooks.register('register_rich_text_link_handler')
 def register_cfgov_link_handler():
     return ('page', CFGovLinkHandler)
+
+
+@hooks.register('cfgov_context_handlers')
+def register_js_handler(*args):
```
chosak: It might be more clear and helpful to others looking to make handlers like this if you specified the args explicitly, e.g.

```py
def register_js_handler(page, request, context):
```

Otherwise you have to go look at the `JSHandler` class to figure out what this is doing, and even then you have to go look at the base class.

cfgov/v1/handlers.py
```diff
@@ -0,0 +1,20 @@
...
+class Handler(object):
+    def __init__(self, page, request, context, *args, **kwargs):
+        self.page = page
+        self.request = request
+        self.context = context
+
+    def process(self):
+        raise NotImplementedError
+
+    # Retrieves the stream values on a page from it's Streamfield
```
chosak: Nitpicky grammar: should be "its".

chosak:
Generally, I like this approach. One question, though. What if there are other types of Wagtail blocks that are not defined under the `v1.atomic_elements` tree? For example, `jobmanager` [has some of these now](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jobmanager/models/blocks.py), and other apps may follow the same pattern of putting app-specific blocks outside of `v1`. Won't that cause problems with [this line](https://github.com/cfpb/cfgov-refresh/pull/2396/files#diff-adb6929cc356648f322d906c46a76ec1R6) and [this line](https://github.com/cfpb/cfgov-refresh/pull/2396/files#diff-adb6929cc356648f322d906c46a76ec1R25) if they have their own custom JS to be included?

kurtrwall:
@chosak thanks for all your helpful comments.

> What if there are other types of Wagtail blocks that are not defined under the v1.atomic_elements tree?

This could probably be refactored to include those cases. I'll make some updates for this and the rest of the comments here.

willbarton:
@kurtrwall What is the status of this PR now that #2411 has been merged?

kurtrwall:
I'm currently catching this one up. Now that I have you though, I'm wondering what your thoughts are for the comment @chosak made just above.

I'm working on and enum class that should be updated for any JS that should be included. Here's an example:

```python
class JSEnum:
    atoms     = []
    blocks    = []
    molecules = []
    organisms = [
        'BaseExpandable',
        'Expandable',
        'ExpandableGroup',
        'FilterControls',
    ]
```

chosak:
I'm going to close this pull request for now, in light of the recent [Wagtail upgrade PR](https://github.com/cfpb/cfgov-refresh/pull/2476). There might be other, easier ways to do this in the new version.

PR #2401, opened 2016-09-15
3.8.2 backports

rosskarchner:
Changelog updates, and copy changes from @Scotchester

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cfgov-refresh",
-  "version": "3.6.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cfgov-refresh",
-  "version": "3.6.0",
+  "version": "3.8.0",
```
Scotchester: This should be 3.8.2 now, right?

rosskarchner: yeah,  standby

PR #2404, opened 2016-09-16
Try parallel testing and caching

higs4281:
In hopes of speeding up test runs, this takes a shot at running parallel
tests, caching pip installs between runs and combining the coverage
reports. If this doesn't help, we could try caching specific directories.

This also fixes branch specifiers for the build and code climate badges, 
which were still pointed at the  flapjack branch.

## Changes

- New travis configs
- README badges

- @chosak @sebworks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 
+cache: pip
+
 python:
   - "2.7"
 
@@ -12,12 +14,12 @@ install:
   # Project configuration requirements.
   - pip install virtualenv virtualenvwrapper
   - pip install autoenv
+  - pip install coveralls
   - export CXX=clang++
   - curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.31.0/install.sh | bash
   - nvm install 5.5.0
   - nvm use 5.5.0
   - npm install -g gulp
-  - pip install coveralls
 
   # Project initialization.
   - chmod +x ./setup.sh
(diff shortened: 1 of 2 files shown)


higs4281:
So this experiment mostly failed. Parallel tests, at least as Travis handles them for our suite of gulp and tox tests, become a demolition derby of processes, each racing to see who can run into a missing dependency first.

I did add a command to combine coverage results into one figure, but the search for parallel tests with separate results continues.

rosskarchner:
From my reading of [this](https://docs.travis-ci.com/user/speeding-up-the-build/)It seems like the way to do this would be to have a (sorry @willbarton @grapesmoker ) shell script that runs one set of tests or the other based on an environment variable, and then have two 'env' entries in the travis.yml, like:

```yaml

env:
- RUNTEST=frontend
- RUNTEST=backend
```

Those will each kick off entirely seperate travis builds

higs4281:
@rosskarchner thanks -- I'll pursue that.

.travis.yml
```diff
@@ -3,6 +3,8 @@ language: python
 python:
   - "2.7"
 
+cache: pip
+
```
willbarton: I wonder if it's worth including the `node_modules` directory as @chosak did in #2381.

higs4281: Since it didn't speed anything up, I would leave it in 2381 until we know why.

PR #2406, opened 2016-09-19
Add jobmanager app to CFGOVRouter

chosak:
Models for the `jobmanager` app should go to/from the `default` database on the content server, so that they can be edited and shared/published properly in the Wagtail editor.

## Additions

- New unit tests for the `CFGOVRouter` database router.

## Changes

- Adds the `jobmanager` app to the list of apps covered by `CFGOVRouter`.

## Testing

- Testing this properly locally is difficult because we don't have a way to reproduce the multi-database setup. There are new unit tests (`tox`) that test the router itself. This change mirrors the existing structure which works for other apps.

## Review

- @rosskarchner @willbarton 

## Checklist

### DIFF
--- a/cfgov/v1/db_router.py
+++ b/cfgov/v1/db_router.py
@@ -6,7 +6,7 @@ cfgov_apps = [
     'v1',
     'flags',
     'taggit',
-    'jobmanager',
+    'jobmanager'
 ]
 
 
--- a/cfgov/v1/db_router.py
+++ b/cfgov/v1/db_router.py
@@ -6,6 +6,7 @@ cfgov_apps = [
     'v1',
     'flags',
     'taggit',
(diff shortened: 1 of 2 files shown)


rosskarchner:
Nothing to disagree with here-- good job with the test!

@schaferjh @willbarton worth noting: in the time between this code is deployed, and the feature flag is flipped to enable Wagtail careers, those users will no longer need to use Picard to push changes from content to prod: they will be working on a single database.

schaferjh:
> those users will no longer need to use Picard to push changes from content to prod: they will be working on a single database.

@rosskarchner just to make sure I'm clear on this, who are "those users"?

rosskarchner:
Users who currently edit and publish jobs.

PR #2407, opened 2016-09-19
Towards forever-caching static assets.

rosskarchner:
To reduce server load and bandwidth costs, and increase performance for users, we should never make users download the same file more than once.

Part of doing **that**, is making sure each *version* of a file has a unique URL. Django can do that for us with [ManifestFileStaticStorage](https://docs.djangoproject.com/en/1.8/ref/contrib/staticfiles/#manifeststaticfilesstorage)-- but in order for that to work, we need to avail ourselves of Django's 'static' tag, and stop referring to files at absolute paths.

This PR enables ManifestFileStaticStorage, and changes templates and other static assets not to depend on absolute paths.

Once this is in place, we can make changes to our web server to send the appropriate far-future cache expiration headers.

## Additions

- in production, use [ManifestFileStaticStorage](https://docs.djangoproject.com/en/1.8/ref/contrib/staticfiles/#manifeststaticfilesstorage) which writes every static file to a unique URL

## Changes

- In Jinja2 templates, replace references to files in (for example) /static/foo with `{{ static('foo') }}`
- In django templates, do the equivalent `{% static 'foo' %}`
- In javascript and CSS, change absolute references to relative references.

## Testing

after checking this out, add a line to local.py like:

 `STATIC_URL = '/foobar/'`

'foobar' could be anything.

Browse the site, and observe that everything still works. Watch the console or view source, and observe that static files are being pulled from 'foobar' (or whatever) instead of /static/. 

## Review

- @willbarton 
- @cfpb/cfgov-frontends 

## Todos

- we'll need to apply similar these changes to our other django apps, and sheer sites, before this gets deployed
- what to do about Wordpress?

## Checklist

…function

to resolve paths in templates, and relative references in Javascript


vccabral:
I am super happy your team is taking on this challenge. There are also apps inside of CFPB that suffer from this problem.

1. Are these going to be addressed in this PR eventually a different PR. 
2. There is a mix of relative URLs and static references. 
3. What are you thoughts on css/js versus template reference. Any idea on how to enforce consistency now that you are making this decision.

rosskarchner:
Apps that are on the outside today will remain so (at least, as far as this PR is concerned), so they will have their own PR's. 

I think 2 is what I addressed here, and I'm not sure what 3 means.

vccabral:
css/less/sass files might reference static content like images. What is your standard for reffering to the location of those files. 

1. If you reference them in a relative manner that could break when files are migrated.
2. If you inject the STATIC url into the preprocessor you will have to use an environment variable.
3. If you setup a base tag it might solve our issues but that is a bit hacky.

rosskarchner:
I went with option 1-- but you're right, all of the answers are sort of awkward and annoying

cfgov/jinja2/v1/_includes/organisms/_vars-mega-menu.html
```diff
@@ -27,7 +27,7 @@
         'url':  '/about-us/the-bureau/',
     },
     'image': {
-        'src': '/static/img/fmc-about-us-540x300.jpg',
+        'src': 'img/fmc-about-us-540x300.jpg',
```
higs4281: should these take a `static()` call?

cfgov/jinja2/v1/_includes/organisms/_vars-mega-menu.html
```diff
@@ -39,7 +39,7 @@
         'url':  '/policy-compliance/guidance/implementation-guidance/',
     },
     'image': {
-        'src': '/static/img/fmc-poly-com-540x300.jpg',
+        'src': 'img/fmc-poly-com-540x300.jpg',
```
higs4281: not static()?

cfgov/jinja2/v1/index.html
```diff
@@ -19,8 +19,8 @@
             'body':    'We are the Consumer Financial Protection Bureau,
                         a U.S. government agency that makes sure banks, lenders,
                         and other financial companies treat you fairly.',
-            'sm_img':  '/static/img/home-hero_sm.jpg',
-            'lg_img':  '/static/img/home-hero_lg.jpg'
+            'sm_img':  'img/home-hero_sm.jpg',
```
higs4281: static()?

cfgov/jinja2/v1/index.html
```diff
@@ -48,7 +48,7 @@ <h2 class="header-slug">
                         {{ info_unit( {
                             'modifier': 'home',
                             'image': {
-                                'url': '/static/img/home-complaint_150.png',
+                                'url': 'img/home-complaint_150.png',
```
higs4281: static()?

cfgov/jinja2/v1/index.html
```diff
@@ -71,7 +71,7 @@ <h2 class="header-slug">
                         {{ info_unit( {
                             'modifier': 'home',
                             'image': {
-                                'url': '/static/img/home-answers_150.png',
+                                'url': 'img/home-answers_150.png',
```
higs4281: static()

cfgov/jinja2/v1/index.html
```diff
@@ -90,7 +90,7 @@ <h2 class="header-slug">
                         {{ info_unit( {
                             'modifier': 'home',
                             'image': {
-                                'url': '/static/img/home-goals_150.png',
+                                'url': 'img/home-goals_150.png',
```
higs4281: static()

cfgov/jinja2/v1/index.html
rosskarchner: I handled it in the macro that gets called with these values (I think)

cfgov/jinja2/v1/_includes/organisms/_vars-mega-menu.html
rosskarchner: It's handled in the macro: https://github.com/cfpb/cfgov-refresh/pull/2407/files#diff-f2ca90745a1b3ee1fd53aa8189a37e09R49

jonathanwcrane:
While I am ignorant of many of the technical aspects of this PR, I am highly supportive, and delighted that you're doing this.

cfgov/jinja2/v1/_includes/organisms/_vars-mega-menu.html
higs4281: right you are. 
👍

rosskarchner:
@higs4281 did you see many instances of NOT using the static tag in the other apps?

willbarton:
I think we want some FEWD eyes on this, right?

@contolini @virginiacc @sebworks

higs4281:
@rosskarchner 
> did you see many instances of NOT using the static tag?

I'll keep looking, but I think css is more of an issue.

CSS refs like `../img/us-flag_44x26.png` aren't resolving

contolini:
👍  I can't find any broken images.

sebworks:
💯

higs4281:
late, but in my local setup i did find missing assets on wagtail pages that call for s3 images, such as 

- /consumer-tools/auto-loans/  
    The page calls http://www.consumerfinance.gov/f/_images./.. instead of http://s3.amazonaws.com/files.consumerfinance.gov/f/_images/...

PR #2408, opened 2016-09-20
added tests for various sharing-related template tags

grapesmoker:
A test for template tags related to sharing.

## Testing

See above.

## Review

@chosak @richaagarwal @higs4281 @willbarton 

## Checklist


cfgov/v1/tests/test_templatetags_share.py
```diff
@@ -0,0 +1,113 @@
...
+
+        self.page.live = False
+        self.page.shared = True
+        self.page.save()
+
+        result = share.get_page_state_url(None, self.page)
+        self.assertEqual(result, self.page.url.replace('http://localhost',
+                                                       'http://content.localhost'))
+
+    def test_get_page_state_url_live(self):
```
richaagarwal: Could you add comments describing each of these tests or otherwise make the test itself read a little bit more descriptively/  For instance, referencing `self.page.url` in the assertion is a little confusing, if that's where our expectation should be, shouldn't we just say 'Live' or whatever there?

cfgov/v1/tests/test_templatetags_share.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def test_get_page_state_url_staged(self):
+
+        self.page.live = False
+        self.page.shared = True
+        self.page.save()
+
+        result = share.get_page_state_url(None, self.page)
+        self.assertEqual(result, self.page.url.replace('http://localhost',
+                                                       'http://content.localhost'))
```
richaagarwal: My preference here would be to be explicit about what the assertion is equal to rather than rely too much on what is happening in `setUp`. I think it would make this more readable and eliminate the need for `self.page.url.replace(...)`

grapesmoker:
I have to say that I don't understand why the test is failing above. It succeeds on my local install. I traced the path of the code down into the wagtail internals and it does what I expect; if this is different on the current server, then it must be because of a difference in the internals.

richaagarwal:
@grapesmoker Yes, it works for me locally too. I left a few comments, if those can be addressed I think this is ready to be merged in (pending whatever fixing needs to happen in travis, unrelated to your PR)

cfgov/v1/tests/test_templatetags_share.py
grapesmoker: So the problem here is that the URL that is automatically generated for the page gets a number appended to it that seems to depend on some internals (order of object saved? not sure). Because of that it doesn't seem like it's possible to compare `page.url` directly to a hard-coded result; the expectation here is that the result will be `page.url` with the server replaced by the staging server, so that's what I'm checking for. I will add commentary explaining this.

cfgov/v1/tests/test_templatetags_share.py
grapesmoker: I'll add commentary.

chosak:
This test always failed for me locally when running `tox`, although not `tox -e fast`. The reason the test fails is because pages will only have a null URL if they aren't aligned to a site root, meaning that their `url_path`, e.g. `path/to/page/` doesn't match any of the site root URLs, e.g. `cfgov`.

@richaagarwal @grapesmoker do these tests pass for you when running the full suite with `tox`?

richaagarwal:
@chosak Sorry, didn't try `tox` alone before. Doesn't work for me either.

chosak:
@richaagarwal thanks for checking.

I think the way to get that test to work is to first delete all of the existing `Site` objects. That change will need to be contained to a single test, so that it doesn't break everything else. The best way to do this might be to change this test case from a `unittest.TestCase` to a `django.test.TestCase`, which isolates each test's changes to the DB (this would also remove the need for the specific `tearDown` method here which deletes the user).

@grapesmoker I'm happy to pair or try to make that change if you want.

grapesmoker:
Ok, so it looks like this works now. Sorry for being bad at Django tests :(

richaagarwal:
👍

PR #2411, opened 2016-09-21
Add AbstractFormBlock for handling forms in Wagtail pages

kurtrwall:
This adds a block class to be used for modules that have a form. The class automatically handles getting the form into the context of the page. Developers should only have to follow steps from the docs to get the form in the page context.

## Additions

- AbstractFormBlock
- `CFGOVPage.serve_ajax()` to handle ajax requests.
- util module
 - `load_class` that takes in a path to a class and returns the class
 - `get_streamfields` returns a dictionary of a page's streamfield names and values
- `form_module_handlers` function to add form module's context to the page's context

## Changes

- `CFGOVPage.serve` overridden to call `serve_ajax` for ajax requests.
- `CFGOVPage.get_context` overridden to call registered `cfgovpage-context-handlers` functions

## Testing

- `tox`

## Review

- @cfpb/cfgov-backends 
- @hillaryj 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Created new `WAGTAIL_CAREERS` feature flag to toggle from Django to Wagtail careers pages.
 - Production settings now use ManifestStaticFilesStorage
 - Added a 'run_travis.sh' script to enable separate JS and Python test coverage reporting 
+- AbstractFormBlock
 
 ### Changed
 - Refactored heroes to support the new "bleeding" format.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Optimized Travis build by removing unnecessary steps.
 - Refactored heroes to support the new "bleeding" format.
 - `CFGOVPage.get_context()` now uses wagtail hooks to call functions registered with the hook name `cfgovpage_context_handlers`
-- `CFGOVPage.serve()` calls `CFGOVPage.serve_ajax()` to handle AJAX requests
+- `CFGOVPage.serve()` calls `CFGOVPage.serve_post()` to handle POST requests
 
 ### Removed
 - `max-height` styling on info unit images
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,13 +19,17 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Created new `WAGTAIL_CAREERS` feature flag to toggle from Django to Wagtail careers pages.
 - Production settings now use ManifestStaticFilesStorage
 - Added a 'run_travis.sh' script to enable separate JS and Python test coverage reporting 
-- AbstractFormBlock
+- AbstractFormBlock to be used as a base class for modules requiring Django Forms
+- wagtail_hooks.py function `form_module_handlers` used by `CFGOVPage.get_context()`
 
 ### Changed
 - Refactored heroes to support the new "bleeding" format.
 - In templates, ALL static file references now use Django's `static` tag/function
 - In CSS/Less, references to other assets are now relative
 - Optimized Travis build by removing unnecessary steps.
+- Refactored heroes to support the new "bleeding" format.
+- `CFGOVPage.get_context()` now uses wagtail hooks to call functions registered with the hook name `cfgovpage_context_handlers`
+- `CFGOVPage.serve()` calls `CFGOVPage.serve_ajax()` to handle AJAX requests
 
 ### Removed
 - `max-height` styling on info unit images
--- a/cfgov/v1/blocks.py
+++ b/cfgov/v1/blocks.py
@@ -1,5 +1,5 @@
 from wagtail.wagtailcore import blocks
-from v1.util import util
+from django.utils.module_loading import import_string
 
 
 class AbstractFormBlock(blocks.StructBlock):
@@ -16,7 +16,7 @@ class AbstractFormBlock(blocks.StructBlock):
         if not handler_path:
             raise AttributeError(
                 'You must set a handler attribute on the Meta class.')
-        return util.load_class(handler_path)
+        return import_string(handler_path)
 
     def is_submitted(self, request, sfname, index):
         form_id = 'form-%s-%d' % (sfname, index)
--- a/cfgov/v1/blocks.py
+++ b/cfgov/v1/blocks.py
@@ -1,0 +1,32 @@
+from wagtail.wagtailcore import blocks
+from v1.util import util
+
+
+class AbstractFormBlock(blocks.StructBlock):
+    """
+    Block class to be subclassed for blocks that involve form handling.
+    """
+    def get_result(self, page, request, value, is_submitted):
+        handler_class = self.get_handler_class()
+        handler = handler_class(page, request, block_value=value)
+        return handler.process(is_submitted)
+
+    def get_handler_class(self):
+        handler_path = self.meta.handler
+        if not handler_path:
+            raise AttributeError(
+                'You must set a handler attribute on the Meta class.')
+        return util.load_class(handler_path)
+
+    def is_submitted(self, request, sfname, index):
+        form_id = 'form-%s-%d' % (sfname, index)
+        if request.method.lower() == self.meta.method.lower():
+            query_dict = getattr(request, self.meta.method.upper())
+            return form_id in query_dict.get('form_id', '')
+        return False
+
+    class Meta:
+        # This should be a dotted path to the handler class for the block.
+        handler = None
+        method = 'POST'
+        icon = 'form'
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -7,7 +7,7 @@ from django.core.urlresolvers import reverse
 from django.db import models
 from django.db.models import Q
 from django.db.models.signals import pre_delete
-from django.http import Http404
+from django.http import Http404, JsonResponse
 from django.utils.translation import ugettext_lazy as _
 from django.utils import timezone
 from django.dispatch import receiver
@@ -15,7 +15,7 @@ from django.contrib.auth.models import User
 
 from wagtail.wagtailimages.models import Image, AbstractImage, AbstractRendition
 from wagtail.wagtailadmin.edit_handlers import StreamFieldPanel
-from wagtail.wagtailcore import blocks
+from wagtail.wagtailcore import blocks, hooks
 from wagtail.wagtailcore.blocks.stream_block import StreamValue
 from wagtail.wagtailcore.fields import StreamField
 from wagtail.wagtailcore.templatetags.wagtailcore_tags import slugurl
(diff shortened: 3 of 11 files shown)


cfgov/v1/blocks.py
```diff
@@ -0,0 +1,32 @@
...
+        handler_class = self.get_handler_class()
+        handler = handler_class(page, request, block_value=value)
+        return handler.process(is_submitted)
+
+    def get_handler_class(self):
+        handler_path = self.meta.handler
+        if not handler_path:
+            raise AttributeError(
+                'You must set a handler attribute on the Meta class.')
+        return util.load_class(handler_path)
```
chosak: Could you use the built-in [`django.utils.module_loading.import_string`](https://docs.djangoproject.com/en/1.8/ref/utils/#django.utils.module_loading.import_string) here instead?

```py
>>> from django.utils.module_loading import import_string
>>> LandingPage = import_string('v1.models.landing_page.LandingPage')
>>> LandingPage
<class 'v1.models.landing_page.LandingPage'>
```

cfgov/v1/util/util.py
```diff
@@ -143,3 +144,32 @@ def all_valid_destinations_for_request(request):
                             valid_destination_for_request(request, pair[1])]
 
     return valid_destinations
+
+
+def load_class(full_class_string):
```
chosak: See above; this is useful, but a duplicate of existing Django functionality in [`django.utils.module_loading.import_string`](https://docs.djangoproject.com/en/1.8/ref/utils/#django.utils.module_loading.import_string).

cfgov/v1/blocks.py
kurtrwall: I wasn't aware of this! Thanks I'll use it for sure.

cfgov/v1/util/util.py
kurtrwall: I've replaced it as suggested. Thanks!

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 
 ### Added
 - Created new `WAGTAIL_CAREERS` feature flag to toggle from Django to Wagtail careers pages.
+- AbstractFormBlock
```
willbarton: Can we say just a teensy bit more about this, or link to the doc?

cfgov/v1/tests/test_wagtail_hooks.py
```diff
@@ -217,4 +218,78 @@ def test_not_publishing_a_page(self, mock_publish, mock_settings):
...
+    @mock.patch('v1.wagtail_hooks.util.get_streamfields')
+    def test_sets_context(self, mock_getstreamfields, mock_hasattr):
+        mock_hasattr.return_value = True
+        child = mock.Mock()
+        mock_getstreamfields.items.return_value = [('name', [child])]
+        form_module_handlers(self.page, self.request, self.context)
+        assert 'form_modules' in self.context
+
+    @mock.patch('v1.wagtail_hooks.util.get_streamfields')
+    def test_sets_context(self, mock_getstreamfields):
```
willbarton: This test has the same name as the one above.

tox.ini
```diff
@@ -3,7 +3,7 @@ envlist = py27,py34
 skipsdist = True
 
 [testenv]
-recreate = True
+recreate = False
```
willbarton: I think there was a reason @chosak left it as `True` in the main testenv when he made the "fast" env below, wasn't there?

chosak:
What about browsers without JavaScript enabled?

tox.ini
chosak: Yes, the default tox environment should recreate the virtualenv each time, so this should be left as `recreate=True`. You can use `tox -e fast` if you don't want to recreate (but also skip most migrations).

cfgov/v1/models/base.py
```diff
@@ -225,6 +225,43 @@ def get_next_appropriate_siblings(self, hostname, inclusive=False):
...
+    def serve_ajax(self, request):
+        """
+        Attempts to retreive form_id from the POST request and returns a JSON
+        response.
+
+        If form_id is found, it returns the response from the block method
+        retrieval.
+
+        If form_id is not found it returns an error response.
+        """
```
chosak: I think I understand what this is trying to do, although this seems to introduce certain hidden assumptions, namely that Wagtail page submissions via AJAX will always include a `form_id` parameter. This means in future we couldn't support other arbitrary AJAX submissions.

Also, should this return an `HTTP 400 Bad Request` instead of 200 to be consistent with RESTful status codes?

cfgov/v1/tests/models/test_base.py
```diff
@@ -0,0 +1,97 @@
+import mock
+
+from django.http import JsonResponse
+from django.test import TestCase
+from django.test.client import RequestFactory
+
+from ...models.base import CFGOVPage
```
chosak: Nitpicky style thing, but I prefer to use `v1.models.base` here, because if the structure of this file were to ever change (say, in another subdirectory), this would break. It's also a little harder to parse visually because you have to think about what module this is referring to.

cfgov/v1/tests/models/test_base.py
```diff
@@ -0,0 +1,97 @@
...
+        mock_super().serve.assert_called_with(self.request)
+
+    @mock.patch('__builtin__.super')
+    @mock.patch('v1.models.base.CFGOVPage.serve_ajax')
+    def test_serve_calls_serve_ajax_on_ajax_request(self, mock_serve_ajax, mock_super):
+        self.request = self.factory.post('/', HTTP_X_REQUESTED_WITH='XMLHttpRequest')
+        self.page.serve(self.request)
+        mock_serve_ajax.assert_called_with(self.request)
+
+    def test_serve_ajax_returns_failed_JSON_response_for_no_form_id(self):
```
chosak: See above; this means that the only AJAX requests supported must have a form ID.

cfgov/v1/util/util.py
```diff
@@ -1,3 +1,4 @@
+import importlib
```
chosak: Not needed due to your recent revision removing the custom import function.

cfgov/v1/util/util.py
```diff
@@ -143,3 +144,14 @@ def all_valid_destinations_for_request(request):
                             valid_destination_for_request(request, pair[1])]
 
     return valid_destinations
+
+
+def get_streamfields(page):
+    """
+    Retrieves the stream values on a page from it's Streamfield
```
chosak: Nitpicky grammar, should be "its".

More seriously, what if you have multiple Streamfields on a page, with identically-named fields? I don't know if this exists right now in our page classes but it doesn't seem impossible, e.g.

```py
class PageWithMultipleStreamfields(Page):
    section1 = StreamField([
        ('introduction', molecules.TextIntroduction()),
        ('content', blocks.RichTextBlock()),
    ])
    section2 = StreamField([
        ('introduction', molecules.TextIntroduction()),
        ('content', blocks.RichTextBlock()),
    ])
```
This method would combine multiple blocks into a single dictionary value.

cfgov/v1/wagtail_hooks.py
```diff
@@ -4,15 +4,15 @@
 from django.utils import timezone
 
 from django.conf import settings
-from django.http import Http404
+from django.http import Http404, JsonResponse
```
chosak: Import not needed here.

docs/forms-in-wagtail.md
```diff
@@ -0,0 +1,44 @@
+# Including forms into Wagtail page context
```
chosak: I like this documentation although I might leave out the first two paragraphs which are, although perhaps fairly critical of Wagtail's form support, more internal details of our development process and less appropriate for developers trying to use the code.

tox.ini
kurtrwall: Whoops, this was a mistake. Thanks for catching.

cfgov/v1/models/base.py
kurtrwall: I agree with the status code comment, I'll add that one in. The idea behind the form_id is that a content editor could add multiple forms on a page, so we'd need a way figure out which form should receive the data in the request. 

Since it would be quite difficult to guarantee that a Wagtail page won't have a form in the content, I decided that a form_id attribute in the request was a necessity. I'm open to other suggestions or conclusions though. 

This also doesn't leave out the idea of adding more logic to handle AJAX requests that don't have the form_id attribute, it's just not something we do at the moment.

cfgov/v1/tests/models/test_base.py
kurtrwall: Right away, sir.

cfgov/v1/tests/test_wagtail_hooks.py
kurtrwall: Womp.

cfgov/v1/util/util.py
kurtrwall: _sigh_

cfgov/v1/util/util.py
kurtrwall: This method actually uses the name of the Streamfield on the page as a key in the dictionary, so this wouldn't happen. The dictionary for this example page would be:
```python
{
    'section1': [block, block, block, ...],
    'section2': [block, block, block, ...]
}
```

docs/forms-in-wagtail.md
kurtrwall: > more internal details of our development process and less appropriate for developers trying to use the code

I think I understand where you're coming from. However, is our documentation supposed to be restrictively about code use? The first two paragraphs serve as an explanation for its use and purpose, not to criticize Wagtail. 

I find it wholly better to introduce something (especially something that changes a lot) with more explanation than just "here's how you use it", because it lets others in on the problem it is attempting to solve. That way down the road if the problem goes away or a better solution can be had, it would be easier to refactor since anyone can track down how and why it's there in the first place.

I'm completely open to scrapping this in favor of brevity, I just think giving something context is a generally good thing.

kurtrwall:
>  seems to only support AJAX requests when submitting forms

Non-ajax requests are definitely supported! Try it out!

> It then presumably relies on some frontend code to handle failure states or marking up fields that fail backend validation

On AJAX requests, this is true. I was under the assumption that frontend validation would be exactly how we should handle these types of requests.

> What's the advantage of this approach vs. just POSTing to a separate Django view

There were a couple of problems that I ran into with the approach.

1. The view doesn't know about the page, which meant I needed to pass the page's ID and do a DB lookup for it. Not a huge deal, just seemed a bit silly since we should already know what page we're coming from.
2. There wasn't a satisfying way to insert an invalidated form into the context of the page when the page would be rerendered.
3. There would need to be a new endpoint for every module we create.

Those are the ones I just remembered when dealing with how to do it this way.

docs/forms-in-wagtail.md
willbarton: I think it's helpful to have justification, but I might word it something more along these lines:

> Wagtail provides a FormBuilder module, but it cannot be used with subclasses of Page, like our CFGOVPage. This is an alternative approach that will work with our Page subclass.

The second paragraph doesn't really seem to be informative at all.

willbarton:
> > seems to only support AJAX requests when submitting forms

> Non-ajax requests are definitely supported! Try it out!

@kurtrwall perhaps it'd be good to have a test that illustrates support for non-AJAX form submission, because I also don't really see how it could happen from reading the code.

kurtrwall:
Follow the instructions from #2415

kurtrwall:
@willbarton non-AJAX requests are supported because the request would go through the same flow that it normally would without this. The request is passed into the wagtail_hooks.py function to manipulate the context of the page, but that ultimately comes from Wagtail's Page.serve() method. Since we wanted AJAX requests, it was easier to override that serve method to handle them differently, but if it's not AJAX, it remains exactly how it always has been.

docs/forms-in-wagtail.md
kurtrwall: I'll update it with these changes.

kurtrwall:
@willbarton @chosak up for final review!

chosak:
@kurtrwall Fortune favors the bold - looks good to me. Looking forward to reviewing 2415 again once it's been simplified with this. @willbarton if all looks good, please merge at your leisure?

docs/forms-in-wagtail.md
```diff
@@ -0,0 +1,40 @@
...
+
+Wagtail provides a FormBuilder module, but it cannot be used with subclasses of Page, like our CFGOVPage. For our purposes there is `AbstractFormBlock`, a subclass of StructBlock that implements methods to process a request. If a developer wishes to add a module that includes a form, they only need to follow a few steps in order to get it handled properly:
+
+1. Create the form.
```
willbarton: Coming out of our forms-that-store vs forms-that-don't conversation, I think I'd like to see a parenthetical note here or something that the form can be a a Form or ModelForm subclass backed by a model. But that's nothing that should hold this PR up any longer.

Thinking forward, it might even be worth including a Model and ModelForm in the example, even if that's a known Django pattern, just to emphasize how doable it is.

willbarton:
Extra points for having included the CHANGELOG updates from the start. 👍

PR #2415, opened 2016-09-22
Add Data & Research app and Conference Registration form backend

kurtrwall:
Dada is implementing a form to register attendees of conferences for Research. This PR adds the `data_research` app and the modules necessary to stand up this form on a Browse page. For frontend review, this does change some templates around so I wanted you all to be aware.

## Additions

- `data_research` app to project
- ConferenceRegistrationForm, ConferenceRegistrationBlock, ConferenceRegistration model, ConferenceRegistrationHandler, and template
- data_research template root to template engine

## Changes

- Some atoms templates were changed to conform to Conference form needs

## Testing

- Run `cfgov/manage.py migrate`
- Run `tox`
- Create a Browse page and select Conference Registration Form block in the context Streamfield and add the fields. At least a gd code and a session.
- Publish the page.
- Visit the page and fill out the form
 - The first 3 fields are required and email must be properly formatted
- Submit the form
 - Note that the form should return an error because the GD code used is not registered with GD. This is the proper behavior for this circumstance.
 - Add a proper GD code to the page via the admin
 - Republish and resubmit the form. Notice the form should submit properly if all fields were filled out

Other things of note
- Failed submissions should have an error message and the values from the submissions filled out in the form.
- Successful submissions should have a success message and clear the  form

## Review

- @hillaryj 
- @cfpb/cfgov-backends 
- @jimmynotjim 
- @KimberlyMunoz 

### Admin
screen shot 2016-09-22 at 9 52 11 am

### New Form
screen shot 2016-09-22 at 9 49 40 am

### Bad Registration
screen shot 2016-09-22 at 9 53 19 am

### Bad form
screen shot 2016-09-22 at 9 59 31 am

### Success
screen shot 2016-09-22 at 9 55 26 am

## Notes

- This is backend specific. No frontend validation is expected to work at this time. That's a todo below.

## Todos

- Add validation for checkbox selection
- Integrate frontend ajax request handling for form

## Checklist


cfgov/cfgov/settings/base.py
```diff
@@ -123,9 +125,13 @@
     {
         'NAME': 'wagtail-env',
         'BACKEND': 'django.template.backends.jinja2.Jinja2',
-        'DIRS': [V1_TEMPLATE_ROOT, V1_TEMPLATE_ROOT.child('_includes'),
+        'DIRS': [
+            V1_TEMPLATE_ROOT,
+            V1_TEMPLATE_ROOT.child('_includes'),
             V1_TEMPLATE_ROOT.child('_layouts'),
-            PROJECT_ROOT.child('static_built')],
+            DATA_RESEARCH_TEMPLATE_ROOT,
```
chosak: Question: thus far we haven't separated out the templates for different apps, as everything is under `cfgov/jinja2/v1`. For example, the careers pages (even though their backend code is in `jobmanager`) or calendar pages (even though their backend code is in `cal`). I can see arguments for both sides, although I think I lean towards centralizing everything under one directory, if only because satellite app templates will always be including core modules/atomic elements. Centralizing also reduces possibility for name conflict, e.g. the off chance that multiple apps have files at the same path, e.g. `v1/index.html` and `otherapp/index.html`.

I'd be curious how @cfpb/cfgov-frontends would lean on this issue.

cfgov/data_research/handlers.py
```diff
@@ -0,0 +1,82 @@
+import json
+from govdelivery.api import GovDelivery
+from urlparse import urlparse
+
+from django.conf import settings
+from django.contrib import messages
```
chosak: Will this actually work properly on the production site? I know it's in use elsewhere (like in [`cal`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cal/views.py#L7)) but by default this functionality uses [sessions](https://docs.djangoproject.com/en/1.8/topics/http/sessions/), which requires cookies. I thought that cookies can't work properly with Akamai in production. @sebworks @rosskarchner

cfgov/cfgov/settings/base.py
kurtrwall: > Centralizing also reduces possibility for name conflict 

That was an immediate concern that I thought of when doing this, I thought I'd just add this in controversially ;)

I think I lean towards a central location as well, but it should probably be reworked. For example, let's remove the `v1` folder and just have everything live directly under jinja2. This type of change is out of scope for this PR though. I can move it back to status quo if requested.

Scotchester: > For example, let's ~~remove the v1 folder and just have everything live directly under jinja2~~ burn down the entire folder structure and start over.

FTFY

jimmynotjim:
It turns out the checkbox validation isn't quite as simple as we thought it was. Only radios allow you to list them as required and pick one, checkboxes require you to pick them all. This is going to take a bigger update than the scope of this PR and should be a new task specific to checkbox validation.

chosak:
@kurtrwall now that #2411 has been merged, can you resolve the conflicts with master so that we can review this one again?

sebworks:
@jimmynotjim, try `data-required` instead. The filterable list controls validation is due a refactoring but I think it will handle checkboxes.

sebworks:
@kurtrwall @chosak , is it possible to handle modifications in forms without migrations? The TableBlocks module handles it by using javascript to write everything to a hidden field and storing the JSON blob.

kurtrwall:
@sebworks It all depends on how the form is created with the module, but for the most part migrations are not necessary.

cfgov/cfgov/settings/base.py
kurtrwall: For now, I'm putting the template back into the jinja2/v1 folder.

chosak:
@kurtrwall @sebworks raises a good point. If we define forms as Wagtail StructBlocks (e.g. [here](https://github.com/cfpb/cfgov-refresh/pull/2415/files#diff-41c53f8b5fb83c10bec63c74dc949ce7R5)), then every time a form is changed (renaming a field, changing help text, etc) then a migration will be generated for all Page models that include those forms as fields or StreamField options.

cfgov/jinja2/v1/browse-basic/index.html
```diff
@@ -21,7 +21,14 @@
     {%- endfor %}
 
     {% for block in page.content -%}
-        {{ render_block.render(block, loop.index) }}
+        {% if block.block_type == 'conference_registration_form' %}
```
chosak: Is every possible form type going to need their own hardcoded entry in page templates, or can this be made generic?

cfgov/v1/handlers/__init__.py
```diff
@@ -0,0 +1,12 @@
+from itertools import chain
+from wagtail.wagtailcore.blocks.stream_block import StreamValue
```
chosak: These imports are not used.

CHANGELOG.md
```diff
@@ -21,6 +21,10 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Added a 'run_travis.sh' script to enable separate JS and Python test coverage reporting
 - AbstractFormBlock to be used as a base class for modules requiring Django Forms
 - wagtail_hooks.py function `form_module_handlers` used by `CFGOVPage.get_context()`
+- `data_research` app to project
+- ConferenceRegistrationForm, ConferenceRegistrationBlock, ConferenceRegistration model, ConferenceRegistrationHandler, and template
+- data_research template root to template engine
+- CFGOVPage.serve_ajax to handle ajax requests
```
chosak: What change does this CHANGELOG entry refer to?

cfgov/data_research/models.py
```diff
@@ -0,0 +1,11 @@
+from django.db import models
+
+# Used for registering users for a conference
+class ConferenceRegistration(models.Model):
+    # Required entries: name, email, sessions
```
chosak: If you set `blank=True` but don't set `null=True` as well then these will be saved in the database as the empty string. Is there a specific reason this is better than saving them as null?

cfgov/jinja2/v1/browse-basic/index.html
kurtrwall: We could probably add a property to the block class to take care of this.

cfgov/data_research/models.py
kurtrwall: There is no specific reason, no.

kurtrwall:
@chosak that's true for any form that we manipulate via content in a block, yes. We could _in theory_ reuse or assemble our own form field block that could let content editors create form fields whenever they want so there wouldn't be migrations.

CHANGELOG.md
```diff
@@ -21,6 +21,10 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Added a 'run_travis.sh' script to enable separate JS and Python test coverage reporting
 - AbstractFormBlock to be used as a base class for modules requiring Django Forms
 - wagtail_hooks.py function `form_module_handlers` used by `CFGOVPage.get_context()`
+- `data_research` app to project
+- ConferenceRegistrationForm, ConferenceRegistrationBlock, ConferenceRegistration model, ConferenceRegistrationHandler, and template
+- data_research template root to template engine
```
kurtrwall: Yes.

CHANGELOG.md
kurtrwall: `serve_post` used to be named `serve_ajax`

cfgov/jinja2/v1/_includes/conference-registration-form.html
```diff
@@ -0,0 +1,107 @@
...
+
+{% macro render(value, settings, form_id) %}
+{% set form = settings.form %}
+
+<form id="registration-form"
+      method="post"
+      action="."
+      class="o-form js-validate-filters">
+
+    <input type="hidden" name="csrfmiddlewaretoken" value="{{ csrf_token }}">
```
chosak: This was perhaps something that should have been discussed in #2411, but this will not work with Akamai, because the token will be cached and so will be the same for every form submission. I believe this is the reason why `@csrf_exempt` is required [here](https://github.com/cfpb/cfgov-refresh/blob/e0e0f874adb344f5afdd47a6404a7acdb5004478/cfgov/core/views.py#L62).

kurtrwall:
Pushed an additional commit to add handling for the conference capacity. Please let me know if this change belongs in another PR, or if it's okay to stay here.

chosak:
@kurtrwall what's left to get this PR merged, and then add subsequent smaller changes (e.g. the conference capability) as future PRs? Seems like there might be some migration file changes required due to other PRs that have gone in. Can you fix the conflicts with master?

kurtrwall:
@chosak Absolutely. Doing that now.

kurtrwall:
@chosak I can remove that last one if you want. There's one more PR coming in on custom error messages but the requirements aren't fleshed out completely.

I'll fix migrations and other conflicts right now.

kurtrwall:
@chosak should I make the same template changes that @Scotchester made on the Feedback module [here](https://github.com/cfpb/cfgov-refresh/pull/2418/commits/08fcefb53d2d4d26e4bdbddf7f074774f841a74e)?

chosak:
@kurtrwall hmm, it seems like the block that he added as `setup-feedback-and-render.html` is actually the same block you need for your form, isn't it? (Although there's a hardcoded `feedback` variable name in there.) Could you just reuse that block, and maybe rename the file to something more generic?

kurtrwall:
@chosak I made a generic form module template https://github.com/cfpb/cfgov-refresh/pull/2415/commits/36631e18561791f4bf3f16d66a73f09199c35350

and fixed something small here https://github.com/cfpb/cfgov-refresh/pull/2415/commits/6be9dae558f91c925abe68e2a6157d351157fe09

chosak:
@kurtrwall I think if you've renamed that template snippet file, you'll also need to fix all of the other templates where it was referred to by @Scotchester. I'm hoping that Travis fails because we have unit tests that would break if the template snippet didn't exist!

Scotchester:
@chosak Didn't fail. Probably because that would only throw a 500 if a page with one of those templates is accessed AND (I think) if that page has a feedback form on it, because the `include` is only run if a feedback module is found.

chosak:
@Scotchester yes, that's too bad. It'd be nice if we had unit tests that rendered each page type with each possible field type on them, to protect against this happening. Might take some effort but I'll add a backlog task.

cfgov/jinja2/v1/_includes/article.html
```diff
@@ -51,7 +52,7 @@
             {% if block.block_type == 'content' %}
                 {{ parse_links(block.value) | safe }}
             {% elif block.block_type == 'feedback' %}
-                {% include 'setup-feedback-and-render.html' %}
+            {{- form_block.render(block, 'content', loop.index0) -}}
```
Scotchester: Trivial: mis-nested.

cfgov/jinja2/v1/_includes/article.html
```diff
@@ -24,6 +24,7 @@
 
    ========================================================================== #}
 
+{% import 'form_block.html' as form_block with context %}
 {% import 'organisms/item-introduction.html' as item_introduction with context %}
```
Scotchester: Should all of these imports be using the minus syntax for collapsing whitespace?

cfgov/jinja2/v1/_includes/form_block.html
```diff
@@ -0,0 +1,42 @@
...
+       Call this macro when the block class inherits from AbstractFormBlock to
+       to render it correctly.
+
+        {% import 'form_block.html' as form_block with context %}
+
+       {% for block in page.streamfield %}
+           {{ form_block.render(block, 'streamfield', loop.index0) }}
+       {% endfor %}
+
+   ========================================================================= #}
```
Scotchester: Trivial: the indentation in this docblock is all over the map.

cfgov/jinja2/v1/browse-basic/index.html
```diff
@@ -21,8 +22,8 @@
     {%- endfor %}
 
     {% for block in page.content -%}
-        {% if block.block_type == 'feedback' %}
-            {% include 'setup-feedback-and-render.html' %}
+        {% if block.block_type in ['feedback', 'conference_registration_form'] %}
```
Scotchester: Is there any other way to express this check that doesn't have to be updated every time we add a new type of form block?

cfgov/v1/models/browse_page.py
```diff
@@ -35,6 +36,7 @@ class BrowsePage(CFGOVPage):
         ('table_block', organisms.AtomicTableBlock(table_options={'renderer':'html'})),
         ('job_listing_table', JobListingTable()),
         ('feedback', v1_blocks.Feedback()),
+        ('conference_registration_form', ConferenceRegistrationForm()),
```
Scotchester: Why doesn't this have `v1_blocks.` in front of `ConferenceRegistrationForm()`?

cfgov/jinja2/v1/_includes/atoms/textarea.html
```diff
@@ -15,10 +15,13 @@
    disabled:    Whether the field is disabled. Defaults to false.
 
    ========================================================================== #}
-
```
Scotchester: Trivial: Don't remove this blank line.

cfgov/jinja2/v1/_includes/atoms/submit-button.html
```diff
@@ -1,5 +1,9 @@
+{% macro render(value) -%}
+
 <div class="o-form-actions">
     <input class="btn btn__super form-actions_item"
            type="submit"
            value="{{ value.button_text }}">
 </div>
+
+{%- endmacro %}
```
Scotchester: Is this atom being called anywhere else that will break due to it being put into a macro?

Scotchester: Looks like no. A search of cfgov-refresh reveals no prior references to `submit-button.html`.

cfgov/v1/models/browse_page.py
chosak: I think because `ConferenceRegistrationForm` was created as a block under `data_research.blocks` instead of under `v1.blocks`.

cfgov/jinja2/v1/browse-basic/index.html
kurtrwall: If there's such a thing as `getattr` in jinja2 then maybe. This will eventually be refactored.

PR #2417, opened 2016-09-23
Update akamai_flush to flush all pages

richaagarwal:
This reduces the time for an existing page to get updated from 20-30 minutes to ~5 minutes. An unfortunate side effect is that we have to flush all pages, so we will need to monitor this to see if it has any other performance side effects. This adds an env variable so prod would need to get updated with that (I have created a PR for that).

## Changes

- See `CHANGELOG.md`

## Testing

- Since this is an Akamai change you can't *really* test this locally, but to see that the code is working and logging properly, you can schedule a page to be published a few minutes in the future, and then run `python cfgov/manage.py publish_scheduled_pages` at the appropriate time. However, you will also need to have your .env configured with 
```
#export ENABLE_AKAMAI_CACHE_PURGE=<boolean_enable_purging_akamai_cache>
#export AKAMAI_USER=<akamai_username>
#export AKAMAI_PASSWORD=<akamai_password>
#export AKAMAI_OBJECT_ID=<akamai_object_id>
```

## TODO

- I need to update this so that we only call the flush on existing pages, as a new page would not need to be flushed. I think before this wasn't really an issue since we weren't flushing all, but now we should try to minimize unnecessary calls to Akamai. I can do that as part of this PR or in a separate one.

## Review

- @cfpb/cfgov-backends 

## Checklist


cfgov/v1/wagtail_hooks.py
```diff
@@ -67,15 +71,26 @@ def configure_page_revision(page, is_sharing, is_live):
...
+            'domain': 'production',
+            'objects': [akamai_object_id]
+        }
+        r = requests.post(
+            akamai_url,
+            headers=headers,
+            data=json.dumps(payload),
+            auth=(akamai_user, akamai_pw)
+        )
+        logger.info('Initiated Akamai flush with response %s' % r.text)
```
higs4281: chance to drop the deprecated %s formatting?

cfgov/v1/wagtail_hooks.py
```diff
@@ -1,17 +1,21 @@
...
 from django.conf import settings
 from django.http import Http404
 from django.contrib.auth.models import Permission
 from django.utils.html import escape
 
-from wagtail.wagtailcore.models import Page
-from wagtail.wagtailcore import hooks
+import requests
+import os
+import json
```
chosak: Minor thing, but it'd be nice to clean up the [PEP8 import order](https://www.python.org/dev/peps/pep-0008/#imports) here.

cfgov/v1/wagtail_hooks.py
```diff
@@ -67,15 +71,26 @@ def configure_page_revision(page, is_sharing, is_live):
...
-        url_paths = [page.url_path.replace('cfgov/', '')]
-        if url_paths[0] == '/':
-            is_home_page = True
-        else:
-            is_home_page = False
-
-        akamai_cache_reset(url_paths, invalidate_root=is_home_page, user_email=page.owner.email)
+        akamai_user = os.environ.get('AKAMAI_USER')
+        akamai_pw = os.environ.get('AKAMAI_PASSWORD')
+        akamai_object_id = os.environ.get('AKAMAI_OBJECT_ID')
```
chosak: Do we want this method to fail if these environment variables aren't set? If so we could use e.g. `os.environ['AKAMAI_USER']` which will raise `KeyError` if the specified variable isn't defined. Seems like that might be a nice check against accidentally running this without having those set up.

cfgov/v1/wagtail_hooks.py
```diff
@@ -67,15 +71,26 @@ def configure_page_revision(page, is_sharing, is_live):
...
+        }
+        r = requests.post(
+            akamai_url,
+            headers=headers,
+            data=json.dumps(payload),
+            auth=(akamai_user, akamai_pw)
+        )
+        logger.info('Initiated Akamai flush with response %s' % r.text)
+        return True
+    return False
```
chosak: Another minor thing, but this method might be a little clearer if you do the initial check and return `False` right away if it doesn't pass, instead of having a big block of code before the return.

cfgov/v1/wagtail_hooks.py
```diff
@@ -67,15 +71,26 @@ def configure_page_revision(page, is_sharing, is_live):
...
-        if url_paths[0] == '/':
-            is_home_page = True
-        else:
-            is_home_page = False
-
-        akamai_cache_reset(url_paths, invalidate_root=is_home_page, user_email=page.owner.email)
+        akamai_user = os.environ.get('AKAMAI_USER')
+        akamai_pw = os.environ.get('AKAMAI_PASSWORD')
+        akamai_object_id = os.environ.get('AKAMAI_OBJECT_ID')
+        akamai_url = 'https://api.ccu.akamai.com/ccu/v2/queues/default'
```
chosak: Might be nice to add a link to [the Akamai documentation](https://api.ccu.akamai.com/ccu/v2/docs/) for this endpoint so that in future if things change or break we have some record of where this format is defined.

cfgov/v1/wagtail_hooks.py
```diff
@@ -67,15 +71,26 @@ def configure_page_revision(page, is_sharing, is_live):
...
+            'type': 'cpcode',
+            'domain': 'production',
+            'objects': [akamai_object_id]
+        }
+        r = requests.post(
+            akamai_url,
+            headers=headers,
+            data=json.dumps(payload),
+            auth=(akamai_user, akamai_pw)
+        )
```
chosak: Perhaps this is out of scope for this PR, but it might be nice to check `r.status_code` here to assert that the request was successful (from [the docs](https://api.ccu.akamai.com/ccu/v2/docs/#section_Basics)) it looks like it should return 201.

cfgov/v1/wagtail_hooks.py
```diff
@@ -67,15 +71,26 @@ def configure_page_revision(page, is_sharing, is_live):
 
 def flush_akamai(page, is_live):
     if is_live and settings.ENABLE_AKAMAI_CACHE_PURGE:
```
chosak: So this bug predates your PR and I'm not sure if you want to address it here. But note that in our settings, [we try to read](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/base.py#L503) this `ENABLE_AKAMAI_CACHE_PURGE` from the environment, and in the `.env_SAMPLE` [the placeholder says](https://github.com/cfpb/cfgov-refresh/blob/master/.env_SAMPLE#L38):

```sh
#export ENABLE_AKAMAI_CACHE_PURGE=<boolean_enable_purging_akamai_cache>
```

This is actually tricky, though, because if you define it as `False`, it will be read in as a string, and the above check will pass:

```sh
$ ENABLE_AKAMAI_CACHE_PURGE=False python
>>> import os
>>> if os.environ.get('ENABLE_AKAMAI_CACHE_PURGE'): print('uhoh')
... 
uhoh
```

I'm not sure what the easiest thing to do here would be, maybe just change the `.env_SAMPLE` to remove the "boolean" reference and just say "only set this to enable caching".

rosskarchner: @chosak  Your suggestion (set = enabled) seems correct to me

rosskarchner:
This might be overly fiddly, but what do you think of making these *settings* variables, that default (in base.py) to pulling values from the environment variables. I think that would be more transparent to, say, an experienced Django developer who was new to our project.

Here's a place where we use that sort of pattern: https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/base.py#L354

cfgov/v1/wagtail_hooks.py
vickumar1981: could this also be something which we set in the configuration?

cfgov/v1/wagtail_hooks.py
richaagarwal: @higs4281 Word, will update to use `.format`.  I wish python 2.7 supposed f-strings 😭 https://www.python.org/dev/peps/pep-0498/#can-t-combine-with-u

cfgov/v1/wagtail_hooks.py
richaagarwal: Cool, will update `.env_SAMPLE` to reflect this..

cfgov/v1/wagtail_hooks.py
richaagarwal: I was debating this but thought that, for local development, it might be nice to see how this generally works but w/o having all the config variables set up. So how it works now, is that it would log that it attempted a flush but with the failure response text. As you mention in your other comment, I could integrate the `status_code` into this. Thoughts?

cfgov/v1/wagtail_hooks.py
richaagarwal: @vickumar1981 Do you mean adding these variables to `base/settings.py` ?  @rosskarchner mentioned that as well, which I can do.

cfgov/v1/wagtail_hooks.py
chosak: The case I was trying to avoid was where we think we've flushed, but for some reason (the variables aren't set, or Akamai returns an error) things fail. If all we do is log that error, then we'd never know unless we happened to check the log If the code raised an exception if something isn't set up or things go wrong, it'd be an easier way to tell because the caller or upstream request (Wagtail editor) would see the error happen.

richaagarwal: Ah, that makes a lot of sense. In that case, I'm not sure we should set defaults in `settings.py`, as @rosskarchner suggests in this comment: https://github.com/cfpb/cfgov-refresh/pull/2417#issuecomment-249253315

cfgov/v1/tests/test_wagtail_hooks.py
```diff
@@ -197,24 +196,3 @@ def test_publishing_a_page(self):
         latest_content = json.loads(latest.content_json)
         assert latest_content['shared']
         assert latest_content['live']
-
-
```
richaagarwal: Removing these tests since all they really do is test the logic of `if is_live and settings.ENABLE_AKAMAI_CACHE_PURGE` and not the akamai purge itself. I will try to come up with another way of testing this, particularly when I add the check for if it's a new page.

richaagarwal:
Putting this on hold until I implement the check for whether it's a new page or not.

cfgov/cfgov/settings/base.py
```diff
@@ -501,3 +501,8 @@
 
 # Akamai
 ENABLE_AKAMAI_CACHE_PURGE = os.environ.get('ENABLE_AKAMAI_CACHE_PURGE', False)
+if ENABLE_AKAMAI_CACHE_PURGE:
+    AKAMAI_USER = os.environ['AKAMAI_USER']
```
higs4281: It's reasonable to assume the Akamai vars will be available if `ENABLE_AKAMAI_CACHE_PURGE` is set, but calling os.environ values directly risks a KeyError that would halt the whole show. 
Could we use `os.environ.get()` for each sub-value to scope the risk to just the purge process?

richaagarwal: @higs4281 I tend to agree with you but @chosak brought the point up that if the config variables were not set, this would quietly fail (https://github.com/cfpb/cfgov-refresh/pull/2417#discussion_r80318403). This could be a reason to call `os.environ` directly from the function which I was doing earlier, but there also seems to be a preference for keeping these in `settings/base.py`.

richaagarwal: I *could* update these to `os.environ.get()` here, and then throw an error in `flush_akamai` if they are `None`. A little verbose, but perhaps that's the best compromise?

higs4281: I would vote for that over risking a 500.

cfgov/v1/wagtail_hooks.py
```diff
@@ -30,7 +39,8 @@ def share_the_page(request, page):
 
     share(page, is_sharing, is_live)
     configure_page_revision(page, is_sharing, is_live)
-    flush_akamai(page, is_live)
+    if is_live:
+        flush_akamai(page)
```
chosak: Should we check the return value here and do something if the flush failed? At a minimum a `log.error`?

richaagarwal: @chosak If we attempt a flush, it does get logged (even if it fails). This would return `False` if we shouldn't be flushing at all, but it seemed a little out of scope to add logging to parts of the code I didn't change.

PR #2418, opened 2016-09-27
Add Feedback module

kurtrwall:
Adds the Feedback module to Browse, Browse Filterable, Landing, Sublanding, Sublanding Filterable, Learn, and Document Detail pages.

## Additions

- Add Feedback model and form
- Add blocks.py for new module
- Add django admin registration for Feedback
- Add feedback template
- Add migration file for Feedback table
- Add handlers module in v1 app to handle module response

## Testing

- `cfgov/manage.py migrate`
- `tox`
- Go to one of the referenced page types and create or edit a page.
 - Select a feedback block
 - Publish the page
- Go to the page's URL and click the Submit feedback button with no content. You should receive an error telling you to select a choice.
- Resubmit with a choice and/or optional feedback
- Go to the Django admin
- Go to the Feedbacks link
- Note the list of feedbacks with a filter option on the right with all of the names of the pages

## Review

- @Scotchester 
- @cfpb/cfgov-backends 

## Checklist


cfgov/jinja2/v1/browse-basic/index.html
```diff
@@ -21,7 +21,14 @@
     {%- endfor %}
 
     {% for block in page.content -%}
-        {{ render_block.render(block, loop.index) }}
+        {% if block.block_type == 'feedback' %}
```
chosak: This block of code is repeated in every page template (and similarly in #2415 [here](https://github.com/cfpb/cfgov-refresh/pull/2415/files#diff-45f9e68fb5117558b28c78af907fac50R24)). Can you pull this out somehow to reduce this code duplication, either here or in that other PR?

cfgov/v1/models/base.py
```diff
@@ -539,3 +539,9 @@ class TemporaryLockout(models.Model):
     user = models.ForeignKey(User)
     created = models.DateTimeField(auto_now_add=True)
     expires_at = models.DateTimeField()
+
+
+class Feedback(models.Model):
+    is_helpful = models.BooleanField()
+    comment = models.TextField(blank=True, null=True)
+    page = models.ForeignKey(Page, on_delete=models.SET_NULL, null=True)
```
chosak: Using `models.SET_NULL` here seems better than [`CASCADE`](https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.ForeignKey.on_delete), but still means that if someone sends feedback for a page, and that page is then deleted, then we'll have feedback left in the DB with no way to associate it to a page.

Could you just use the page URL here, instead of a foreign key? That'd record the equivalent information and would be simpler, especially if this feedback is just intended for export to some other format.

If you do keep this as a foreign key, can you add a more useful [`related_name`](https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.ForeignKey.related_name) here, so that you could refer to e.g. `Page.feedback`?

Scotchester:
Exciting 😄 

I did some work on the front-end of this locally yesterday. I have an appointment at 11:30, but after that I will review this and push up my FE contributions.

cfgov/v1/handlers/blocks/feedback.py
```diff
@@ -0,0 +1,51 @@
...
+            feedback.save()
+            return self.success()
+        else:
+            return self.fail(form)
+
+    def success(self):
+        if self.request.is_ajax():
+            return JsonResponse({'result': 'pass'})
+        else:
+            messages.success(self.request, message='Thanks for your feedback!')
```
chosak: Per [comments in #2415](https://github.com/cfpb/cfgov-refresh/pull/2415#discussion_r80100025), using Django messages will likely not work in production because of Akamai.

cfgov/v1/migrations/0015_feedback.py
```diff
@@ -0,0 +1,80 @@
...
+        migrations.CreateModel(
+            name='Feedback',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('is_helpful', models.BooleanField()),
+                ('comment', models.TextField(null=True, blank=True)),
+                ('page', models.ForeignKey(on_delete=django.db.models.deletion.SET_NULL, to='wagtailcore.Page', null=True)),
+            ],
+        ),
+        migrations.AddField(
```
chosak: The following steps fail when running on a new local database:

```sh
$ ./drop-db.sh
$ ./create-mysql-db.sh
$ ./initial-data.sh
```

The reason is documented at GHE Platform issue 323. @Scotchester and I hit this last week, and a temporary workaround is to remove the changes introducted to `v1.DemoPage` (don't add a new Streamfield there, which will remove this step from this migration). Pending a more comprehensive solution to this problem, can you remove the changes to the `DemoPage` class, as they don't seem critical/necessary for this PR?

kurtrwall:
@Scotchester Nice! Looking forward to the push.

cfgov/jinja2/v1/browse-basic/index.html
kurtrwall: I think a full solution for this is out of scope for these PR's. My recommendation for it would be to create a template that would be imported/included that would encompass all of if/else clauses the pages have. 

I can make the change in a different PR, unless it doesn't seem as out of scope as I'm suggesting.

cfgov/v1/models/base.py
kurtrwall: The association as a foreign key makes more sense to me than a page URL, especially considering the possibility for a slug/structure change we'd have to account for.

I'll most definitely create a `related_name` though.

cfgov/v1/handlers/blocks/feedback.py
kurtrwall: I think this is something we can solve in a separate PR that can refactor the generic Handler class these all use.

cfgov/v1/migrations/0015_feedback.py
kurtrwall: Yeah, I'll do that. In reference to "Pending a more comprehensive solution", I have multiple examples of migration files that were deleted in the "big crunch :)" which solves this. I'll find and post to that issue.

kurtrwall: Just made a comment on it GHE/platform/issues/323#issuecomment-89775.

cfgov/v1/admin.py
```diff
@@ -28,7 +28,15 @@ def send_password_reset_email(self, request, queryset):
...
+    if feedback.page:
+        return feedback.page.url
+
+feedback_page_url.short_description = 'Page URL'
+
+
 @admin.register(Feedback)
 class FeedbackAdmin(admin.ModelAdmin):
-    list_display = ('is_helpful', 'comment')
+    list_display = ('is_helpful', 'comment', feedback_page_url)
```
Scotchester: Thanks for adding this; can we also get the datetime of the feedback submission?

kurtrwall: Oh, right. Forgot about that. Adding it in now.

cfgov/jinja2/v1/browse-basic/index.html
richaagarwal: Yeah, I think it'd be nice if the logic was included in a partial. I don't think it's out of scope since the duplication is happening within this PR.

Scotchester:
Hey @kurtrwall, after your latest commit, I get the "Your models have changes that are not yet reflected in a migration, and so won't be applied." message. Running `makemigrations` creates a new migration with `Alter field submitted_on on feedback`. Could you revise your last commit with the full migration update for the `submitted_on` field?

Scotchester:
The weird thing is, this is all that's in the 0016 migration it created:

```python
class Migration(migrations.Migration):

    dependencies = [
        ('v1', '0015_feedback'),
    ]

    operations = [
        migrations.AlterField(
            model_name='feedback',
            name='submitted_on',
            field=models.DateTimeField(auto_now_add=True),
        ),
    ]
```

And it looks like the field is already configured that way by your 0015 migration.

Scotchester:
Something doesn't seem quite right with `submitted_on`. When I test submitting a feedback with your latest commit, I get an OperationalError exception: "(1054, "Unknown column 'submitted_on' in 'field list'")"

Scotchester:
Nevermind... apparently it was me being dumb about migrations. Trying with a fresh database reload, all is well.

cfgov/v1/migrations/0015_feedback.py
```diff
@@ -0,0 +1,79 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name='Feedback',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
+                ('is_helpful', models.BooleanField()),
+                ('comment', models.TextField(null=True, blank=True)),
+                ('page', models.ForeignKey(related_name='feedback', on_delete=django.db.models.deletion.SET_NULL, to='wagtailcore.Page', null=True)),
+                ('submitted_on', models.DateTimeField(default=datetime.datetime(2016, 9, 28, 14, 43, 49, 537049, tzinfo=utc), auto_now_add=True)),
```
chosak: Can you remove this hardcoded default value?

cfgov/jinja2/v1/wagtail/demo/index.html
```diff
@@ -20,6 +20,17 @@
...
+    {% for block in page.blocks %}
+        {% if block.block_type == 'feedback' %}
+            {% import block.block.meta.template as feedback with context %}
+            {% set form_id = 'form-content-' ~ loop.index0 %}
+            {% set settings = form_modules.content[loop.index0] %}
+            {{ feedback.render(block.value, settings, form_id) }}
+        {% else %}
+            {{ render_block.render(block, loop.index) }}
+        {% endif %}
+
```
Scotchester: missing `{% endfor %} here`

cfgov/v1/migrations/0015_feedback.py
kurtrwall: Ugh, sorry about that.

cfgov/jinja2/v1/wagtail/demo/index.html
Scotchester: Fixed in 08fcefb

cfgov/jinja2/v1/browse-basic/index.html
Scotchester: Fixed in 08fcefb

cfgov/jinja2/v1/wagtail/demo/index.html
```diff
@@ -28,7 +28,7 @@
             {{ render_block.render(block, loop.index) }}
         {% endif %}
     {% endfor %}
-    </div>
+
```
Scotchester: Good catch.

PR #2424, opened 2016-09-28
Django templates: use staticfiles instead of plain old 'static'

rosskarchner:
The static tag that you get from `{% load static %}` doesn't respect the value of STATICFILES_STORAGES, which means it can't take advantage of the changes made in #2407.

`{% load static from staticfiles %}` is a drop-in replacement that DOES respect those your configured static storage mechanism.

https://docs.djangoproject.com/en/1.8/ref/contrib/staticfiles/#std:templatetag-staticfiles-static

## Changes

- described above!

## Testing

- check a legacy django page (like /find-a-housing-counselor/ ) before and after the change, and you should notice no changes.

## Review

- @cfpb/back-end-team-admin 

## Checklist

### DIFF
--- a/cfgov/legacy/templates/front/base_nonresponsive.html
+++ b/cfgov/legacy/templates/front/base_nonresponsive.html
@@ -1,4 +1,4 @@
-{% load static %}
+{% load static from staticfiles %}
 {% load global_include %}
 {% get_static_prefix as STATIC_PREFIX %}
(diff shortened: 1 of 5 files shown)


rosskarchner:
This is tiny, and important to get into the 3.10 release tomorrow. I'm gonna go ahead and merge

PR #2434, opened 2016-09-30
Use relative URLs in get_protected_url

chosak:
The template context function `get_protected_url` is used in several places to make sure that pages on the live server don't link to pages on the content server. Currently the logic used here generates absolute URLs (e.g. http://www.consumerfinance.gov/something), which is unnecessary and problematic for conversion to HTTPS. This PR modifies the logic in `get_protected_url` to return relative URLs instead, e.g. [/something/](http://www.consumerfinance.gov/something).

@schaferjh enumerating the logic here for your review:

- Reference to a live page from www returns a relative URL.
- Reference to a live page from content returns a relative URL.
- Reference to a live and shared page from www returns a relative URL.
- Reference to a live and shared page from content returns a relative URL.
- Reference to a shared page from www returns a hash (`#`)
- Reference to a shared page from content returns a relative URL.
- Reference to a draft page from www returns a hash.
- Reference to a draft page from content returns a hash.
- Reference to a null page from www returns a hash.
- Reference to a null page from content returns a hash.

Note that all of the above only apply to places where `get_protected_url` is used in the template; there are many other places and types of links to pages that don't use this method.

## Additions

- New `v1.routing` module that contains some page routing-related logic.
- Moved reference to environment variable `DJANGO_STAGING_HOSTNAME` into setting `STAGING_HOSTNAME`.
- New `v1.middleware.StagingMiddleware` that adds an `is_staging` boolean flag to all incoming requests, thus removing the need for the various checks against `request.site.hostname == os.environ.get('DJANGO_STAGING_HOSTNAME`).

## Changes

- Logic in `get_protected_url` modified to always return relative URLs to pages when possible, or the hash (`#`) when no link should be permitted.

## Testing

- Testing is easiest with a recent production data dump. Browse to [http://localhost:8000/about-us/blog/](http://localhost:8000/about-us/blog/) and notice that all URLs are relative. Browse to http://content.localhost:8000/about-us/blog/, and notice the addition of a draft post and that URLs are also all relative.

## Review

- @cfpb/cfgov-backends  

## Notes

- If people are okay with the use of middleware and `settings.STAGING_HOSTNAME`, there are numerous other places in the code where we check `os.environ.get('DJANGO_STAGING_HOSTNAME) == request.site.hostname` that could also be cleaned up. Those changes could make sense as part of one or more future PRs.
- Apologies to @grapesmoker for modifying this logic and thus breaking and requiring the refactor of his recently-added unit tests for `get_protected_url`.

## Checklist


richaagarwal:
"Reference to a live page from content returns a relative URL." - so this means that the user gets directed to the content version of the page... right?  Would there be no way of sending them to the www page?

chosak:
@richaagarwal thanks for the review - I thought that the current behavior was "if you're on content, and you request a page which is from www (absolute url), we change the request url and send you to the content page". There's an existing test that checks that [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/tests/test_links.py#L82).

That behavior seems more consistent and desirable to me, so that if you're browsing content, you stay on content, and don't unexpectedly get linked over to www. I can definitely modify this if I am misinterpreting the intent. @schaferjh can you comment?

higs4281:
I like how this PR simplifies URL handling by environment, and it catches URLs in content that were getting routed to www. It works for me locally.

We'll have to [adjust StagingMiddleware](https://docs.djangoproject.com/en/1.10/topics/http/middleware/#upgrading-middleware) if/when we graduate to Djano 1.10, but that's a future concern, and not a big one at that.

:+1: for throwing another percent on our testing hearth

schaferjh:
@richaagarwal is correct that currently folks are sometimes sent over to www from content, but that is a bug not a desired behavior.  The test @chosak references was recently written (if I remember correctly) and together with this PR should put us in better position re: linking and our servers!

cfgov/v1/middleware.py
```diff
@@ -0,0 +1,7 @@
+from django.conf import settings
+
+
+class StagingMiddleware(object):
```
willbarton: I love this so much.

willbarton:
CHANGELOG!

willbarton:
Update the CHANGELOG at your discretion, I should say :)

Otherwise, :shipit:

PR #2438, opened 2016-10-03
Remove version number from css banner and package.json

rosskarchner:
Including the version number in built output will lead to files being marked as "changed", even if there aren't actual changes to the code-- busting the cache on every release, instead of just when there are real changes.

This PR stops that from happening. Thank you for the consultation, @sebworks and @contolini!

## Removals

- no more version number in package.json
- no more version number in the 'banner' gulp template.

## Review

- @cfpb/cfgov-frontends 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,12 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 
 ### Removed
 
+## 3.10.3
+
+### Removals
+
+- This removes the current version number from built assets, like main.css
+
 ## 3.10.2
 
 ### Changed
--- a/gulp/config.js
+++ b/gulp/config.js
@@ -20,7 +20,7 @@ module.exports = {
       ' *                     $$\n' +
       ' *                     $$\n' +
       ' *\n' +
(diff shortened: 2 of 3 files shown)


Scotchester:
I may be wrong, but I think we should keep it in `package.json`. It's a required field, and npm will throw warnings on every `npm install` if it's not there.

The update to `config.js` should be all you need to stop the unintentional cache busting.

rosskarchner:
I figured that would be controversial, and officially have no opinion on it ;)

@contolini might disagree, though?

Scotchester:
@rosskarchner Why was 10.0.3 released before this was merged?

contolini:
Is it used for anything? If it's not (and I don't believe it is), I vote to remove it. Otherwise it's just another thing we have to remember to update when we release cf.gov.

@Scotchester w/ npm version 3.10.3 I don't see any warnings when running `npm install`.

rosskarchner:
I think the side affect (not cache busting on simple on releases) was important enough to get out the door tomorrow with the other bandwidth-saving work, but the package.json change seemed to be worth discussion

Scotchester:
If it's not putting up warnings, I'm cool. :shipit:

PR #2446, opened 2016-10-06
Move careers data migrations to scripts

chosak:
Motivated by [GHE]/issues/323, we want to deprecate the use of Django data migrations for creating Wagtail pages. This PR moves functionality from two `jobmanager` data migrations into standalone Python scripts:

- `jobmanager.0007_create_careers_pages` > `scripts.create_careers_pages`

 This script constructs the `About Us` > `Careers` page hierarchy as empty pages in a draft state. These pages already exist in production, so running against a prod DB is a noop.

- `jobmanager.0008_migrate_job_pages` > `scripts.migrate_job_pages`

 This script migrates the last 180 days of existing job postings (Django models) to Wagtail `JobListingPage` pages. New pages are created in draft state as children to the `About Us` > `Careers` > `Current Openings` page. This will need to be run in production to migrate those pages over.

## Additions

- New scripts under `scripts` to perform page creation, plus tests.

## Changes

- Removed functionality from `jobmanager.0007` and `jobmanager.0008` migrations, replaced by comment recording new script locations.

## Testing

- With a new database:
 - `./drop-db.sh` (drop existing database)
 - `./create-mysql-db.sh` (create new, empty database)
 - `./initial-data.sh` (run migrations and set initial page data)
 - `./runserver.sh` (start the Django server)
 - Browse the [Wagtail admin](http://localhost:8000/admin/pages/) and note that there's only a single page.
 - `python cfgov/manage.py runscript create_careers_pages` (create careers page tree)
 - Browse the admin again and note the new empty pages, all created in draft state.
- With a production database dump:
 - `./drop-db.sh` (drop existing database)
 - `./create-mysql-db.sh` (create new, empty database)
 - `./refresh-data.sh /path/to/dump.sql` (load production data)
 - `python cfgov/manage.py migrate` (apply unapplied migrations)
 - `./runserver.sh` (start the Django server)
 - Browse the Wagtail admin and you'll see that the careers page tree already exists.
 - `python cfgov/manage.py runscript migrate_job_pages` (migrate job postings over)
 - Browse the admin again and note the job pages that were created under `Current Openings`, all in draft state.

## Review

- @cfpb/cfgov-backends 

## Notes

- I discovered that the default state for Wagtail `Page` objects is that `live=True`, meaning that if you create a new page programmatically like this, it'll be live by default:

 ```py
 >>> page = MyPageClass(title='Foo', slug='foo')
 >>> root_page.add_child(page)
 ```

 This is not what I would expect. You have to explicitly set `live=False` to create a new page in draft state, apparently:

 ```py
 >>> page = MyPageClass(title='Foo', slug='foo', live=False)
 ```

The relevant Wagtail code is [here](https://github.com/torchbox/wagtail/blob/v1.5.3/wagtail/wagtailcore/models.py#L310); you can see some explanation of this choice by one of the Wagtail devs [here](https://github.com/torchbox/wagtail/issues/1132#issuecomment-88412670).

## Checklist


willbarton:
> python cfgov/manage.py runscript migrate_job_pages (migrate job postings over)

Is this something that'll have to be a manual action post-deploy in production?

chosak:
@willbarton yes, that's correct, although job postings only currently exist in production, so the second script won't apply to those other servers.

willbarton:
@chosak alrighty. This looks good to me — and I'm going to use your script/test as a model for some StreamField modifications I have to make.

:+1:

chosak:
@willbarton thanks for the review.

One clarification since you mention "StreamField modifications": in case it's not clear from the sprawling discussion in the other thread, it is possible (and, I think we came out, preferable) to use Django data migrations to do Wagtail page content updates for things like StreamFields. It's only when you want to actually create new pages that you run into problems and have to do it with standalone scripts.

higs4281:
👍 This code looks good. 
All your steps work for me but the last one, and I think this is a local db problem. 
After all the "Page created ... Page edited" sequences ran, only the 7 dev pages show.
I suspect I hosed my local db setup, so this shouldn't hold up the PR if it's working for others.

willbarton:
@chosak re: StreamFields, alrighty. Sounds like I'll capture that in the docs :)

chosak:
@higs4281 thanks for the review. I think I could have been more clear with the instructions. With an empty local dev database, the result of running both of these scripts will only be the creation of the careers pages tree. No actual job pages will be created, because there aren't any Django jobs in the database to migrate over. That part can only be tested with a production data dump containing Django job objects.

higs4281:
👍 @chosak resurrected mysql and confirmed functionality.

PR #2448, opened 2016-10-07
FilterableListForm refactor

richaagarwal:
## Overview
- This is a pretty giant refactor involving `FilterableListForm` and how we use it, primarily with the goal of making this code cleaner and more testable. 
- I removed the `NewsroomFilterForm` and `ActivityLogFilterForm` entirely in favor of using their parent `FilterableListForm`, and moved out the custom functionality they require into a function called `clean_categories`. This should not change any current functionality. 
## Testing
- Examples of where we use this form:
  - [http://localhost:8000/activity-log/](http://localhost:8000/activity-log/)
  - [http://localhost:8000/about-us/newsroom/](http://localhost:8000/about-us/newsroom/)
  - [http://localhost:8000/policy-compliance/notice-opportunities-comment/open-notices/
    ](http://localhost:8000/policy-compliance/notice-opportunities-comment/open-notices/)
- You could load prod data locally and test that the above pages behave correctly. 
- There are also tests specific to this form now, that can be run by `tox -e fast -- v1.tests.test_filterable_list_form`. 
- I removed a lot of tests that broke due to my changes, not because my changes broke functionality but because functions were removed, renamed, etc. It didn't seem worthwhile to update most of these tests, since they weren't testing anything meaningful. But feel free to chime in if you disagree.
## Notes
- All results from a `FilterableListForm` are always a `AbstractFilterPage` type. You could create a `BrowsePage`, for example, and it would never show up in the search results for the Newsroom or Activity Log. I spoke with @sarahsimpson09 about this & it sounds like this is by design, but I'd like @schaferjh to confirm.
## Review
- There are a few tests I want to add (see TODO) but I think this is ready for a final review regardless.
- @cfpb/cfgov-backends 
## TODO
- Add test for filtering by parent
- Add test with date ranges
- I will update the `CHANGELOG` once it seems this is ready to be merged in


cfgov/v1/util/queries.py
```diff
@@ -0,0 +1,52 @@
+import logging
+import ref
+
+from ..models.learn_page import AbstractFilterPage
```
chosak: Possibly nitpicky style thing, but PEP8 [recommends](https://www.python.org/dev/peps/pep-0008/#imports) absolute imports over relative imports. [This](http://programmers.stackexchange.com/questions/159503/whats-wrong-with-relative-imports-in-python/159505#159505) is also a good explanation of the downsides of relative imports.

cfgov/v1/util/queries.py
```diff
@@ -0,0 +1,52 @@
...
+        aren't categories themselves. Rather, they consist of sub-categories,
+        but since we aren't showing those sub-categories to the end user,
+        selecting the parent category is equivalent to all of them
+        getting checked. The mapping of Blog and Reports to their
+        respective categories exists in cfgov/v1/util/ref.py
+        """
+        categories = self.form.cleaned_data.get('categories', [])
+        if 'blog' in categories:
+            self.form.cleaned_data['categories'].remove('blog')
+            for x in ref.categories[1][1]:
```
chosak: This seems prone to breaking, especially the hardcoded `[8][1]`. I see why it's tricky, because the form gives you back something like `research-reports` and there's no non-hacky way to look up 'Research Report' in the reference without hardcoding it. Is there a reason that we can't change [this line](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/util/ref.py#L146) to `"Research Report"` so that it's at least possible to programmatically look up and map it to its subcategories [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/util/ref.py#L97)?

From a design point of view, I wonder if it might be better to modify modify `FilterableListForm.generate_query` to do this subcategory expansion. There'd need to be some extra logic [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/forms.py#L185) that expanded the category. This makes sense to me as the method named `generate_query` would imply "this method generates a Q object that can be used to look up pages as specified by this form's values". If we're implementing custom logic to expand subcategories (because otherwise the query the form generates actually doesn't work to get, e.g. the blog pages), then should that be part of the form class?

cfgov/v1/util/queries.py
```diff
@@ -0,0 +1,52 @@
...
+                self.form.cleaned_data['categories'].append(x[0].lower())
+        if 'research-reports' in categories:
+            self.form.cleaned_data['categories'].remove('research-reports')
+            for x in ref.categories[8][1]:
+                self.form.cleaned_data['categories'].append(x[0].lower())
+
+    def get_page_set(self):
+        if self.form.__class__.__name__ == 'FilterableListForm':
```
chosak: I don't quite understand why this is the relevant check for whether to include the parent page in the filter or not. Is that just an unfortunate side effect of the design that we only use `FilterableListForm` in that case, but children of `FilterableListForm` otherwise?

Could you instead check whether `page == None`, and only pass it into this class if you want to filter by the parent page? Am I interpreting it right that `self.page` here means "parent page, only return pages that are children of this one" -- if so, would it make sense to name that as `self.parent_page`?

cfgov/v1/util/filterable_list.py
```diff
@@ -50,19 +41,19 @@ def process_forms(self, request, forms):
...
 
-
     def get_forms(self, request):
-        form_class = self.specific.get_form_class()
-        form_specific_filters = self.get_form_specific_filter_data(form_class, request.GET)
-        return [form_class(form_data, parent=self, hostname=request.site.hostname) for form_data in form_specific_filters]
-
+        form_specific_filters = self.get_form_specific_filter_data(request_dict=request.GET)
+        parent = None
+        if self.__class__.__name__ != 'ActivityLogPage' and self.__class__.__name__ != 'NewsroomLandingPage':
```
richaagarwal: @chosak

cfgov/v1/util/queries.py
richaagarwal: I moved this into a function inside `FilterableListForm`.  It's a good idea to update the line to "Research Report" but since that changes what we display, I think we should address that in a separate task/ PR.

richaagarwal:
This is ready for review again 🙏🏽

cfgov/v1/tests/test_filterable_list.py
```diff
@@ -1,107 +1,74 @@
...
-
-    # FilterableListMixin.get_page_set tests
     @mock.patch('v1.models.learn_page.AbstractFilterPage.objects')
-    def test_get_page_set_calls_live_shared(self, mock_cfgovmanager):
-        form = mock.Mock()
-        hostname = 'hostname'
-        self.mixin.get_page_set(form, hostname)
+    def test_base_query_calls_live_shared(self, mock_cfgovmanager):
+        FilterableListForm(parent=None, hostname='hostname').base_query()
         assert mock_cfgovmanager.live_shared.called
```
willbarton: > Perhaps this is a good BEWD topic to discuss and try to standardize?

YES!

richaagarwal: @chosak Good point, Andy.  I updated the tests I added in `test_filterable_list_form`.

cfgov/v1/forms.py
```diff
@@ -102,22 +106,59 @@ class FilterableListForm(forms.Form):
...
 
+    def clean_categories(self):
+        """ This is a (hopefully) temporary solution for dealing w/ the fact
+        that we show Blog and Reports as options for filtering, but
+        aren't categories themselves. Rather, they consist of sub-categories,
+        but since we aren't showing those sub-categories to the end user,
+        selecting the parent category is equivalent to all of them
+        getting checked. The mapping of Blog and Reports to their
+        respective categories exists in cfgov/v1/util/ref.py
+        """
```
willbarton: This is a fantastic docstring, thank you!

I'm curious what the reason is for this, and whether this is something we ought to consider changing/fixing.

richaagarwal:
Thanks to everyone who helped out with this PR and reviewed it. I'm merging, but there are some open questions that would be good to address for the longer-term.

cfgov/v1/forms.py
schaferjh: @richaagarwal We only offer the opportunity to filter by parent category (Blog, News, etc) on the Activity Log, so it's not a "standard" option. The Activity Log page is older flapjack-era code and pre-dates the other filterable lists, so it was probably created with slightly different logic. As things stand, it seems that the "parent category" as you call it, comes from the site section in which the browse page lives and not some other assignment.

Worryingly, since the PR has been merged, the Activity log page on build is missing these parent categories. What needs to happen to fix that?

schaferjh: Actually just realized, this might just be a function of the fact that the filter is pulling in inappropriate browse pages. See my comment below.

schaferjh:
@richaagarwal It looks like this PR has caused some worrying issues on build. Both Newsroom and Activity Log are items seem to be pulling random data-research browse pages in. The Activity Log is also missing the "parent categories" -- but that might just be because the wrong content is being pulled in.
Interestingly, the other filterable list pages I checked on build looked ok at first glance.

Per the note in the PR description:
> All results from a FilterableListForm are always a AbstractFilterPage type. You could create a BrowsePage, for example, and it would never show up in the search results for the Newsroom or Activity Log. I spoke with @sarahsimpson09 about this & it sounds like this is by design, but I'd like @schaferjh to confirm.

I don't understand what this is referring to -- can we chat tomorrow so I can get a bit more context? Thanks!

richaagarwal:
@schaferjh Let's definitely chat tomorrow.  I also peeked at build and it looks all kinds of messed up (different than what I have locally), so will need to sort that out, but perhaps I can do a screenshare and confirm what I have locally is OK.

PR #2450, opened 2016-10-07
Eliminate duplicate category in ItemIntroduction organism

willbarton:
On pages that use the item introduction organism, content managers had to enter the category in both the item introduction and the configuration panel. The category in the item introduction was used to determine whether or not the category should be displayed on the page.

This PR replaces the category in the item introduction organism with a boolean field (show_category) that, when checked, will show the category selected in the page's configuration panel. This "show category" field is checked by default.

## Additions

- Added the `show_category` field on the ItemIntroduction organism
 
## Removals

- Removed the `category` field on the ItemIntroduction organism

## Testing

-   Add or edit a Learn page or Document detail page (children of `AbstractFilterPage`), add a category under "Configuration", toggle the new "Show category" checkbox, preview the page and the category should show up when "Show category" is checked or not depending on the value of the checkbox.
- Run `tox` to run the tests for the migration.

## Review

- @chosak and @kurtrwall for the migrations
- @sarahsimpson09 for the item introduction change
- @cfpb/cfgov-frontends for the template changes
- @cfpb/cfgov-backends 

## Screenshots

3296ce9a-81b4-11e6-9673-1e53e9da4894

## Todos

- [ ] Add documentation for Wagtail and migrations
- [ ] Update CHANGELOG

## Checklist


higs4281:
👍 Works for me locally as advertised, and with multiple categories.   
Two notes:
- Our [Wagtail docs](https://github.cfpb.gov/pages/flapjack/wagtail-documentation/learn_page.html#categories) should be updated to say categories are allowed on Learn pages 
- Unrelated to this PR: Adding an intro item introduction results in two H1 elements without any sectioning. Is this a concern?

cfgov/v1/util/migrations.py
```diff
@@ -57,3 +61,33 @@ def get_or_create_page(apps, page_cls_app, page_cls_name, title, slug,
...
+        set_page_stream_data(page, field_name, new_stream_data)
+
+
+@transaction.atomic
+def migrate_page_types_and_fields(apps, page_types_and_fields, mapper):
+    """ Migrate the fields of a wagtail page type using the given mapper
+        function. page_types_and_fields should be a list of 3-tuples
+        providing ('PageType', 'field_name', 'field type'). """
+    for page_type, field_name, field_type in page_types_and_fields:
+        page_cls = imported_apps.get_model('v1', page_type)
```
chosak: Why does this use `imported_apps` and not `apps`? This could cause problems with future migrations.

willbarton: Primarily because I just adapted the code in [0011_hero_refactor_data.py](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/migrations/0011_hero_refactor_data.py).

I'll modify it to use `apps`.

cfgov/v1/migrations/0021_replace_dup_category_field_data.py
```diff
@@ -0,0 +1,82 @@
...
+        ('AbstractFilterPage', 'header', 'item_introduction'),
+        ('DemoPage', 'header', 'item_introduction'),
+    ]
+    migrate_page_types_and_fields(apps, page_types_and_fields,
+                                  migrate_category_field_backwards)
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('v1', '0021_replace_dup_category_field'),
```
chosak: Did you purposefully start this migration's name `0021_` to match the schema `0021_` migration? This is a nice idea although inconsistent with how we've done it so far. Does it cause any problems with Django if two migrations start with the same number, or is it just an arbitrary identifier?

cfgov/v1/tests/migrations/test_0021_replace_dup_category_field_data.py
```diff
@@ -0,0 +1,92 @@
...
+        migrated = self.migration.migrate_category_field_backwards(mock_revision, data)
+        self.assertEqual(migrated, {'category': 'test-category'})
+
+    @mock.patch('v1.migrations.0021_replace_dup_category_field_data.json')
+    def test_backwards_with_show_category_revision_without_category(
+            self, mock_json):
+        """ Backward migration show_category=True for a revision """
+        mock_revision = mock.Mock(spec=['content_json'])
+        mock_json.loads.return_value = {
+            'categories': []}
```
chosak: This indentation is a little hard to follow and not compliant with [PEP8](https://www.python.org/dev/peps/pep-0008/#indentation) (search for "closing brace/bracket/parenthesis").

cfgov/v1/tests/test_util_migrations.py
```diff
@@ -0,0 +1,192 @@
...
+
+from v1.util.migrations import (
+    migrate_page_types_and_fields,
+    migrate_stream_field,
+    get_stream_data,
+    set_stream_data,
+)
+
+
+class MigrationsUtilTestCase(TestCase):
```
chosak: 👍  for creating unit tests, but these all seem to run the tests against `Mock` objects. Is it possible to write unit tests that actually run these tests against real `Page` and `PageRevision` objects?

docs/wagtail-migrations.md
```diff
@@ -0,0 +1,82 @@
+# Wagtail and Django data migrations
```
chosak: 👍  for writing docs for this! When this is live we should drop a link to our approach in [the Wagtail issue](https://github.com/torchbox/wagtail/issues/1101) you linked to.

cfgov/v1/util/migrations.py
```diff
@@ -57,3 +61,73 @@ def get_or_create_page(apps, page_cls_app, page_cls_name, title, slug,
...
+        set_stream_data(page_or_revision, field_name, new_stream_data)
+
+
+@transaction.atomic
+def migrate_page_types_and_fields(apps, page_types_and_fields, mapper):
+    """ Migrate the fields of a wagtail page type using the given mapper
+        function. page_types_and_fields should be a list of 3-tuples
+        providing ('PageType', 'field_name', 'block type'). """
+    for page_type, field_name, field_type in page_types_and_fields:
+        page_model = apps.get_model('v1', page_type)
```
willbarton: Yup, I meant to do that, actually. 

I was thinking this would be a nice thing to have live in a Wagtail contrib module in our fork of Wagtail... whenever we do that.

cfgov/v1/migrations/0021_replace_dup_category_field_data.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # If the original category field is empty, we take that to mean
+    # show_category should be False.
+    if 'category' not in data or data['category'] == '':
+        data['show_category'] = False
+    else:
+        data['show_category'] = True
+
+    if 'category' in data:
+        del data['category']
```
willbarton: Yes! That's quite elegant, actually.

cfgov/v1/migrations/0021_replace_dup_category_field_data.py
willbarton: It's an arbitrary identifier, the digraph is created based on `dependencies`, IIRC.

I actually did it because I could've sworn the hero data migration was the same number as the schema migration, but it isn't. I'll change that.

cfgov/v1/util/migrations.py
```diff
@@ -57,3 +61,73 @@ def get_or_create_page(apps, page_cls_app, page_cls_name, title, slug,
...
+        # If page_or_revision is a page, this will work
+        field = getattr(page_or_revision, field_name)
+        stream_data = field.stream_data
+    except AttributeError:
+        # Otherwise it will raise an Attribute error and we can assume
+        # page_or_revision is a revision
+        revision_content = json.loads(page_or_revision.content_json)
+        field = revision_content[field_name]
+        stream_data = json.loads(field)
+    return stream_data
```
kurtrwall: It might make more sense to use `isinstance` instead.

cfgov/v1/util/migrations.py
```diff
@@ -57,3 +61,73 @@ def get_or_create_page(apps, page_cls_app, page_cls_name, title, slug,
...
+        setattr(page_or_revision, field_name, stream_value)
+        page_or_revision.save()
+    except AttributeError:
+        # Otherwise it will raise an Attribute error and we can assume
+        # page_or_revision is a revision
+        revision_content = json.loads(page_or_revision.content_json)
+        revision_content[field_name] = json.dumps(stream_data)
+        page_or_revision.content_json = json.dumps(revision_content)
+        page_or_revision.save()
+
```
kurtrwall: Same as [above](https://github.com/cfpb/cfgov-refresh/pull/2450/files#r84109043).

cfgov/v1/util/migrations.py
willbarton: I think I prefer sticking with the `try`/`except`... it seems more Pythonic, unless you (or anyone else) is strongly in favor of using `isinstance`.

That said, this code wouldn't really be very pretty either way.

cfgov/v1/util/migrations.py
vickumar1981: at the least, this needs to be refactored.  the `try...catch` piece of logic is redundant, and should be put in a method that returns to you whether you are dealing with a page or a page revision.

willbarton: `save()` should only be called in a single place, and I can fix that, but I'm not clear that would do much here, Pseudocoding it out, we'd end up with:

```
   if ispage(page_or_revision):
        field = getattr(page_or_revision, field_name)
        stream_block = field.stream_block
        stream_value = StreamValue(stream_block, stream_data, is_lazy=True)
        setattr(page_or_revision, field_name, stream_value)
    else:
        revision_content = json.loads(page_or_revision.content_json)
        revision_content[field_name] = json.dumps(stream_data)
        page_or_revision.content_json = json.dumps(revision_content)
    page_or_revision.save()
```

willbarton: Actually, there is a better way. You *can* get a PageRevision as a Page object:

https://github.com/torchbox/wagtail/blob/stable/1.6.x/wagtail/wagtailcore/models.py#L1430

vickumar1981: I'm thinking something like this:

```
def _check_stream_data(page_or_revision, field_name):
	_is_page = True
    try:
        _field = getattr(page_or_revision, field_name)
    except AttributeError:
    	_field = json.loads(page_or_revision.content_json)
    	_is_page = False
    return _field, _is_page

def set_stream_data(page_or_revision, field_name, stream_data):
	field, is_page = _check_stream_data(page_or_revision, field_name)
    if is_page:
        stream_block = field.stream_block
        stream_value = StreamValue(stream_block, stream_data, is_lazy=True)
        setattr(page_or_revision, field_name, stream_value)
    else:
        field[field_name] = json.dumps(stream_data)
        page_or_revision.content_json = json.dumps(revision_content)
    page_or_revision.save()

def get_stream_data(page_or_revision, field_name):
	field, is_page = _check_stream_data(page_or_revision, field_name)
    if is_page:
        stream_data = field.stream_data
    else:
        stream_data = json.loads(field[field_name])
    return stream_data
```

Even here, I think one way to really think about this, is what if I later in my models added a field called 'content_json' to the page object?  Would this cause a breaking change later on?

cfgov/v1/tests/test_util_migrations.py
willbarton: I want to come back to this because there was some interesting discovery that I made in the process of using an actual Page and PageRevision object to test these functions with, that I hope we can apply in future too.

Because these functions operate on StreamFields, I needed a Page subclass with a StreamField. We clearly have multiple pages like that in `v1.models`, but all of them are more complex than I really needed for testing purposes. Wagtail uses [`wagtail.tests.testapp`](https://github.com/torchbox/wagtail/tree/master/wagtail/tests/testapp) for its tests, and I started digging into it. It includes [StreamPage](https://github.com/torchbox/wagtail/blob/master/wagtail/tests/testapp/models.py#L634) which is a perfect page type for testing these functions: it's a light-weight, direct subclass of Page with an added StreamField.

I want to call attention to this because there are other nice little models in `testapp` too that might be useful to us in other testing scenarios.

willbarton:
Don't merge quite yet — I've got a failure running locally against a database dump.

willbarton:
@chosak I think this is (finally) ready for a final review.

Feedback addressed, tests are passing, and it works as expected with the most recently available database dump.

cfgov/jinja2/v1/_includes/organisms/item-introduction.html
```diff
@@ -34,9 +34,7 @@
 {% set has_authors = page.authors.exists() %}
 
 <div class="o-item-introduction">
-    {% if value.category is string %}
```
chosak: I think there might be a problem with this change, as category URLs don't seem to render properly.

I tried following the instructions above on both a new local database as well as with a production dump. When the category link gets generated on the page, you get something like

http://localhost:8000/testing/#?filter**{{ index }}**_categories=us-supreme-court

(Emphasis added.) On [the example page](http://localhost:8000/policy-compliance/rulemaking/final-rules/2013-real-estate-settlement-procedures-act-regulation-x-and-truth-lending-act-regulation-z-mortgage-servicing-final-rules/) that your screenshot is from above, the category URL is

http://localhost:8000/policy-compliance/rulemaking/final-rules/?filter**{{ index }}**_categories=final-rule

To me it looks like the variable `index` that is getting referred to is never being set anywhere. Maybe it should be `loop.index`? I'm not quite sure how that variable is used.

willbarton:
I split off the StreamField migrations utilities into PR #2481. That PR should be merged before this one.

This PR branch has been rebased as a result of that split. I've also addressed the concerns about a lack of categories and the `show_categories` boolean.

I still want to confirm the behavior this PR introduces with @schaferjh and @sarahsimpson09 before merging.

willbarton:
> I still want to confirm the behavior this PR introduces with @schaferjh and @sarahsimpson09 before merging.

The current `{% else %}` behavior is not necessary. I will remove that then this PR will be ready for final review.

PR #2451, opened 2016-10-07
Add anchor link field and Content with Anchor block for use in data-research site

cfarm:
This is a very limited scope implementation of anchor link fields in the Wagtail admin. It is meant to be used in ~~half-width link blob contexts~~ only for now. **Updated** this is now for use in full-width-text elements only.

User story: as a content editor managing links within the Data and Research site, I want to create and copy anchor links from Wagtail blocks on the Research Conference Papers page, so I can link to those blocks from the Research Conference agenda page.

## Additions

- AnchorLink atom for saving an element id in the db
- AnchorLink is viewable in admin as part of a new "Content with Anchor" module on Full Width Text elements
- it is copy and pasteable
- it is editable by a content editor
- the id is output on `m-full-width-text` elements
- if there is no anchor_ prefix with the unique id, then it generates that and adds it to the text that the editor put in the field
- if there is no text entered in the field, it autogenerates it on save. this ensures that whether the field was entered or not, the front end outputs a unique id to the template

## Testing

Test `AnchorLink` block inputs:

```
tox -e fast v1.tests.test_blocks
```

Test full-width-text organism w/new ContentWithAnchor molecule:

```
tox -e fast v1.tests.atomic_elements
```

Add a full-width-content block to a browse or learn page in the admin to see this in action locally

## Review

- @kurtrwall 

## Screenshots

Admin interface:

screen shot 2016-10-12 at 3 44 58 pm

Front end template output:
screen shot 2016-10-12 at 3 44 13 pm

```

## Checklist

### CODE REVIEW

File: cfgov/v1/util/util.py [hunk 1]
```diff
@@ -33,6 +33,8 @@ def get_form_id(page):
     else:
         return 0
 
+def slugify(string):
```
willbarton: Django has a [django.utils.text.slugify](https://docs.djangoproject.com/en/1.8/ref/utils/) function built-in.

cfgov/v1/atomic_elements/atoms.py
```diff
@@ -154,3 +155,33 @@ def clean(self, data):
             raise ValidationError("ImageBasicUrlAlt validation errors", params=error_dict)
         else:
             return data
+
+class AnchorLink(blocks.StructBlock):
+    link_id = blocks.CharBlock(required=False)
```
kurtrwall: Instead of making this a StructBlock, it should just subclass the CharBlock since this is the only block in it.

kurtrwall: On a more important note, the value of the block is at risk of being changed inadvertently because the clean method will be called on each page save and the `get_unique_id()` will provide a different value since it's randomized at execution.

kurtrwall: Separately, I would also move this block definition outside of `atoms` since it's not part of the atomic design and therefore shouldn't be defined as such. You could put it in `v1/blocks.py`.

cfgov/v1/atomic_elements/atoms.py
```diff
@@ -154,3 +155,33 @@ def clean(self, data):
...
+
+        def format_id(string):
+            words = string.split('_');
+            suffix = ''
+            if string:
+                suffix = '_'
+            if 'anchor' in words:
+                return slugify(string)
+            else:
+                return get_unique_id('anchor_' + slugify(string) + suffix)
```
kurtrwall: If I understand this right, you could make a few adjustments to this, like so:

```python
def format_id(string):
    if 'anchor' in string:
        return slugify(string)
    else:
        suffix = '_' if string else ''
        return get_unique_id('anchor_' + slugify(string) + suffix)
```
What's the point of `suffix`?

chosak:
Would it be possible to remove the 6 auto-generated migration files, and regenerate a single one? Deleting those files and re-running `manage.py makemigrations` should accomplish that, given that none of those were created manually.

Also, we've been trying to name migration files something descriptive, so instead of, e.g. `0015_auto_20161004_1045.py`, it might be helpful to rename the file to something like `0015_half_width_anchor_links.py`.

sebworks:
@cfarm, just for reference, but the `m-info-unit` macro is used for quite a few atomic components.

https://github.com/cfpb/cfgov-refresh/search?utf8=%E2%9C%93&q=info-unit.html

cfgov/v1/atomic_elements/atoms.py
cfarm: I moved the block to `v1/blocks`

for now I will keep this as a StructBlock to see if I can also output a field in the admin that prints the relative URL with the anchor link appended, for easy copy and paste for a content editor.

cfarm:
@sebworks I removed this field from the `info-unit` template - we are using it on `full-width-text` now

cfarm:
@chosak I'll update the migrations once this is in a good place. Thanks for the tip!

cfgov/jinja2/v1/_includes/atoms/anchor-link.html
```diff
@@ -0,0 +1,19 @@
...
+
+   Description:
+
+   Create an ID for use in anchor links.
+   See [GHE]/flapjack/Modules-V1/wiki/Anchor-Link
+
+   value:                  Object defined from a StreamField block.
+
+   value.link_id:         A string consisting of a prefix ('anchor_'), a concatenated random number, and text that a content editor can optionally supply.
+
```
sebworks: Wrap the text here to stay in-line with current standards.

cfgov/jinja2/v1/_includes/organisms/full-width-text-anchor.html
```diff
@@ -0,0 +1,25 @@
...
+   ==========================================================================
+
+   Description:
+
+   Create a full width text organism with an id attribute for use in anchor links.
+   See [GHE]/flapjack/Modules-V1/wiki/Full-Width-Copy
+
+   ========================================================================== #}
+
+{% if value.bound_blocks['anchor_link'] -%}
```
sebworks: We typically use dot notation.

cfgov/v1/blocks.py
```diff
@@ -30,3 +33,44 @@ class Meta:
...
+            if 'anchor' in words:
+                return slugify(string)
+            else:
+                return get_unique_id('anchor_' + slugify(string) + suffix)
+
+        if data:
+            try:
+                data['link_id'] = format_id(data['link_id'])
+            except ValidationError as e:
+                error_dict.update(e.params)
```
sebworks: Are you missing a raise statement?

cfarm:
@kurtrwall @chosak this is updated for your review. still hoping to get this piece merged by the tagging tomorrow - I'll ping you about testing this locally

cfgov/v1/blocks.py
```diff
@@ -1,5 +1,8 @@
+from django.core.exceptions import ValidationError
 from wagtail.wagtailcore import blocks
 from django.utils.module_loading import import_string
+from django.utils.text import slugify
+from .util.util import get_unique_id
```
chosak: Nitpicky style thing: [PEP8](https://www.python.org/dev/peps/pep-0008/#imports) recommends grouping third-party imports together (Django & Wagtail) separated by a space by local imports (the util ones). Within each group we also tend to try to alphabetize the imports (e.g. `from django` before `from wagtail`).

cfgov/v1/blocks.py
```diff
@@ -30,3 +33,39 @@ class Meta:
         handler = None
         method = 'POST'
         icon = 'form'
+
+class AnchorLink(blocks.StructBlock):
+    link_id = blocks.CharBlock(required=False, label='ID for this content block', help_text=(
```
chosak: More nitpicky style: backend [style discussions](https://github.com/cfpb/back-end/issues/14) have roughly converged on trying to use 80-character line widths, and [PEP8](https://www.python.org/dev/peps/pep-0008/) for most other things like indentation.

It can be useful to run a tool like [flake8](https://pypi.python.org/pypi/flake8) (`pip install flake8`) against changed files (`flake8 cfgov/v1/blocks.py`, for example) to check code style.

cfgov/v1/blocks.py
```diff
@@ -30,3 +33,39 @@ class Meta:
...
+            )
+        )
+
+    # def get_url():
+    #     return '/relative/path/'
+
+    # full_url = blocks.CharBlock(required=False, default=get_url(), help_text=(
+    #         'Copy this relative URL to insert anchor links that take the user to this content section.'
+    #         )
+    #     )
```
chosak: Might be nice to remove these commented lines. The PR description can serve as future TODOs.

cfgov/v1/atomic_elements/atoms.py
cfarm: suffix appends an underscore to the string if the user entered some text. for example:

1. user enters 'video' for the anchor field
2. `format_id` adds the 'anchor_' prefix to the beginning of the string 'video', followed by an underscore (from suffix) and appends a unique id 

versus
1. user does not enter an input string for anchor link
2. field generates anchor_34938908908 as an ID when no custom text is entered, otherwise field would have 2 underscores

cfarm:
Fixed style issues and added more tests for the AnchorLink input! Also fixed a case that failed while writing tests :)

cfgov/v1/atomic_elements/atoms.py
kurtrwall: > output a field in the admin that prints the relative URL with the anchor link appended, for easy copy and paste for a content editor.

How does the StructBlock allow you to achieve this? We should chat about what you mean.

cfgov/v1/atomic_elements/atoms.py
kurtrwall: Gotcha, thanks!

Scotchester:
Am I correct in understanding that `anchor_` will be prepended and a string of random hex digits will be appended to whatever the user enters in that field?

cfgov/jinja2/v1/_includes/atoms/anchor-link.html
```diff
@@ -0,0 +1,20 @@
...
```
jimmynotjim: This isn't an element, @kurtrwall noted in chat it should be moved to `_includes/blocks/ `

cfgov/jinja2/v1/_includes/molecules/full-width-text-anchor.html
```diff
@@ -0,0 +1,25 @@
+
+{# ==========================================================================
+
+   Full Width Text with Anchor
```
Scotchester: Why create a new variant module for this instead of just adding it as a standard field on Full Width Text?

cfarm:
Just a note to @Scotchester et al who may be interested in using this feature in Wagtail: this implementation is super limited and will only be used in the Data and Research site for the short term as we look into more sustainable solutions. Possibilities are noted in comments above, including a Wagtail core contribution. For now, be warned that your use of this feature might cause future breakages to your links without our intervention 😱  We'll keep the discussion going in Platform team office hours!

PR #2453, opened 2016-10-11
initial robots.txt

rosskarchner:
It's a robots.txt file!

We will almost certainly iterate on this, but I think this is an OK start. I've only chosen to disallow URL's that produce PDF's, and the "print" versions of Spanish AskCFPB pages (the english versions were removed in the recent redesign).

I've also used the 'sitemap' declaration to point to the activity log RSS feed, since [the sitemap protocol allows it](http://www.sitemaps.org/protocol.html#otherformats). We should still look into a "real" sitemap, but this should allow search engines to pick up new content faster (at least, content that's reflected in the activity feed)

Once merged (and deployed), we'll need to create an alias from /robots.txt to  (site location on server/static/robots.txt

## Additions

- a robots.txt files

## Review

- @willbarton 
- @sarahsimpson09 
- @schaferjh 

## Notes

-

## Todos

-

## Checklist

### DIFF
--- a/cfgov/core/static/robots.txt
+++ b/cfgov/core/static/robots.txt
@@ -1,0 +1,6 @@
+User-agent: *
+Disallow: /save-hud-counselors-list
+Disallow: /about-us/the-bureau/leadership-calendar/pdf
+Disallow: /es/obtener-respuestas/*.imprimir.html
+
+sitemap: http://www.consumerfinance.gov/activity-log/feed/
--- a/cfgov/core/static/robots.txt
+++ b/cfgov/core/static/robots.txt
@@ -2,5 +2,6 @@ User-agent: *
 Disallow: /save-hud-counselors-list
 Disallow: /about-us/the-bureau/leadership-calendar/pdf
 Disallow: /es/obtener-respuestas/*.imprimir.html
+Disallow: /external-site
 
 sitemap: http://www.consumerfinance.gov/activity-log/feed/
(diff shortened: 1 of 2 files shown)


cfgov/core/static/robots.txt
```diff
@@ -0,0 +1,6 @@
+User-agent: *
+Disallow: /save-hud-counselors-list
+Disallow: /about-us/the-bureau/leadership-calendar/pdf
+Disallow: /es/obtener-respuestas/*.imprimir.html
```
Scotchester: Please add: `Disallow: /external-site`

cfgov/core/static/robots.txt
```diff
@@ -0,0 +1,6 @@
...
+
+sitemap: http://www.consumerfinance.gov/activity-log/feed/
```
Scotchester: This URL doesn't point to a sitemap.

Scotchester: Sorry... didn't read the OP. 😳

rosskarchner: the sitemap protocol allows for RSS feeds to stand in: http://www.sitemaps.org/protocol.html#otherformats

cfgov/core/static/robots.txt
rosskarchner: done!

schaferjh:
Sitemap should happen soon... 👍

PR #2454, opened 2016-10-11
enable USE_ETAGS

rosskarchner:
An "etag" is a token that represents the content at a URL at a particular point in time. The idea is, that if a web browser holds on to these etag values, and if a URL is requested again (using the [If-None-Match](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) request header, the browser can make a *conditional* request-- "send me the content at this URL, unless the etag is still XYZ". If the etag is unchanged, instead of responding with "200 OK" and the content, it'll respond with "304 Not Modified" and *not* transmit the content.

Our Akamai configuration is currently not set to validate etags, but I am working to get that fixed shortly.

You can accomplish similar things with the Last Modified date, but that's a little tricker with dynamic content (and Django doesn't provide a similar USE_LASTMODIFIED setting)

## What's next?

One interesting angle with this setting: it does not save any *processing* on the Django side of things, all of the database queries, logic, and rendering must still happen BEFORE the etag comparison is made (otherwise, what would it compare the incoming etag to?). It only saves bandwidth, not computing cycles.

I think we could get some mileage out of caching etags (and maybe even generating last-modified dates while we're at it), so long as we flush this cache at the same time, or do it intelligently when pages change. I'll open up a discussion of this idea in our internal issue tracker.

The Django docs have [some good information on other ways to support conditional processing](https://docs.djangoproject.com/en/1.8/topics/conditional-view-processing/)-- there's a danger in tying etags or modified dates to the underlying page model, though, in that it could create blind spot (like template changes, or parts of a page that are dynamic)

## Additions

- adds the USE_ETAGS configuration value.

## Testing

Use curl to verify that this is working:

```
curl -I http://localhost:8000/ 
HTTP/1.0 200 OK
Date: Tue, 11 Oct 2016 18:36:21 GMT
Server: WSGIServer/0.1 Python/2.7.11
Vary: Cookie
X-Frame-Options: SAMEORIGIN
ETag: "bdc03a6c1c6aba24a39f22738fbfd302"
Content-Type: text/html; charset=utf-8

$ curl -I http://localhost:8000/ --header 'If-None-Match: "bdc03a6c1c6aba24a39f22738fbfd302"'                                     
HTTP/1.0 304 NOT MODIFIED
Date: Tue, 11 Oct 2016 18:36:46 GMT
Server: WSGIServer/0.1 Python/2.7.11
Content-Length: 0
Vary: Cookie
```
The actual value of etag/If-None-Match will probably be different for you.

## Review

- @cfpb/cfgov-backends 

## Notes

-

## Todos

-

## Checklist

https://docs.djangoproject.com/en/1.10/ref/settings/#std:setting-USE_ETAGS

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -16,6 +16,8 @@ SECRET_KEY = os.environ.get('SECRET_KEY', os.urandom(32))
 # Use the django default password hashing
 PASSWORD_HASHERS = global_settings.PASSWORD_HASHERS
 
+# see https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-USE_ETAGS
+USE_ETAGS = True
 
 try:
     import mysql


chosak:
:+1: Nice "lines of code to size of win" ratio!

Agree about the risk of dynamic content. If part of the page is rendered via JS, and that changes, the etag won't change. Using edge-side includes with Akamai might also cause trouble.

higs4281:
👍 slick

schaferjh:
>If part of the page is rendered via JS, and that changes, the etag won't change.

Does this potentially mess with our A/B testing?

rosskarchner:
@chosak @schaferjh It shouldn't cause a problem for JS-powered content, since the JS and HTML are seperate HTTP request

rosskarchner:
@kurtrwall -- this is what I meant by "conditional get" in chat the other week-- the idea was that stuff like this should save enough bandwidth, that we should feel free to try out different caching strategies on specific pages and sections.

chosak:
@rosskarchner re: JS - thanks, that makes sense. It's HTTP requests all the way down.

kurtrwall:
@rosskarchner Thanks for the clarification. In the future, is there something clever we can do with middleware to further optimize?

rosskarchner:
@kurtrwall like I note under "What's Next"-- I think we could come up with a scheme for caching etags/last-modified dates, so that we will often be able to skip going through the whole querying/rendering cycle.

There's also this [decorator-based approach](https://docs.djangoproject.com/en/1.8/topics/conditional-view-processing/) that Django supports, but the implication seems to be storing that information with the model (say, using the modification time of the current revision as the Last-Modified date), which could be dangerous because it ignores other sorts of changes (the template, or other server-side dynamic content on the page)

rosskarchner:
as of this afternoon, the Akamai change is in place

PR #2456, opened 2016-10-12
Upgrade to Wagtail 1.6.3

chosak:
This PR upgrades Wagtail from version 1.5.2 to version 1.6.3.

Relevant Wagtail release notes that describe the changes: [1.6](http://docs.wagtail.io/en/v1.6.3/releases/1.6.html), [1.6.1](http://docs.wagtail.io/en/v1.6.3/releases/1.6.1.html), [1.6.2](http://docs.wagtail.io/en/v1.6.3/releases/1.6.2.html), [1.6.3](http://docs.wagtail.io/en/v1.6.3/releases/1.6.3.html)

## Changes

- Custom `Page` classes [must have an explicit manager defined](http://docs.wagtail.io/en/v1.6.3/releases/1.6.1.html#multi-level-inheritance-and-custom-managers) (3dd7402)
- `wagtail.wagtailadmin.taggable.TagSearchable` is deprecated so migration references [need to be changed](http://docs.wagtail.io/en/v1.6.3/releases/1.6.html#tagsearchable-needs-removing-from-custom-image-document-model-migrations) to `wagtail.wagtailsearch.index.Indexed` (42213ea)
- Custom block `render` methods [now take an optional `context` argument](http://docs.wagtail.io/en/v1.6.3/releases/1.6.html#render-and-render-basic-methods-on-streamfield-blocks-now-accept-a-context-keyword-argument). (fec4dbb)
- `StructBlock` now defines its own `required` property so we cannot use it as a custom class member on our own blocks. (1f0c516)
- Wagtail version upgrade requires these additional package upgrades (46739a6):
 - beautifulsoup4: 4.4.1 -> 4.5.1
 - djangorestframework: 2.4.8 -> 3.1.3

## Testing

- All unit tests pass with `tox` and `gulp test`.
- To test with a recent production data dump:

 ```sh
 $ ./drop-db.sh && ./create-mysql-db.sh
 $ ./refresh-db.sh /path/to/dump.sql
 $ ./runserver.sh
 ```

## Review

- This is a large change so it would be helpful if a few people from @cfpb/cfgov-backends and @cfpb/cfgov-frontends could take a look.
- It's difficult to enumerate a full set of things to check, so general browsing of functionality of a production dump is probably a good approach. If anything is found to be broken, it would be a good idea to first write a (failing) unit test that covers it and then investigate a fix.

## Notes

- The 1.6 release includes a new `{% include_block %}` tag for [improved StreamField template inclusion](http://docs.wagtail.io/en/v1.6.3/releases/1.6.html#include-block-tag-for-improved-streamfield-template-inclusion). This may allow us to remove or simplify some of our custom [`render_stream_child`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/__init__.py#L141) logic, and is worth investigating in future.
- This upgrade also includes a new built-in `IntegerBlock`, which we could move to (in future) instead of our current [custom one](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/atomic_elements/atoms.py#L53). There are also new blocks for `EmailBlock`, `FloatBlock`, `DecimalBlock`, and `RegexBlock`, which might also be useful in the current atomic structure.
- There are new [custom editing interfaces for `StructBlock`](http://docs.wagtail.io/en/v1.6.3/topics/streamfield.html#custom-editing-interfaces-for-structblock) which should allow more control over how our elements appear in the Wagtail editor. 

## Screenshots

Interesting new [Style Guide](http://localhost:8000/admin/styleguide/) section of the admin:

image

## Checklist


chosak:
@higs4281 I note that there are two optional apps that depend on django-rest-framework, both of which are part of the [OAH repo](https://github.com/cfpb/owning-a-home-api). That repo depends on [djangorestframework==2.4.3](https://github.com/cfpb/owning-a-home-api/blob/master/requirements.txt#L3), which gets upgraded in this PR to 3.1.3.

If I install the repo locally, I can hit the [rate explorer page](http://localhost:8000/owning-a-home/explore-rates/) locally, but I get an error that says "We're having trouble connecting to our data". I'm not familiar enough with the OAH code to know whether this might be related to the DRF upgrade. I saw your name in the OAH contributors; would you be able to take a look?

higs4281:
@chosak i'll pursue.

kurtrwall:
Starting a review. I'm in love with this set of changes, and, for better or worse, this could change some custom functionality. 💯

kurtrwall:
There should be a consideration for how the templates have changed over the versions. The first thing I've found is [our template](https://github.com/cfpb/cfgov-refresh/blob/wagtail-upgrade-1.6.3/cfgov/templates/wagtailadmin/pages/listing/_page_title_explore.html) overrides [this one](https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailadmin/templates/wagtailadmin/pages/listing/_page_title_explore.html) that now uses a [hook](http://docs.wagtail.io/en/v1.6.3/reference/hooks.html?highlight=page_listing_buttons#buttons-with-dropdown-lists) to register page actions. There are certainly more changes in the templates to be found, so I'll keep digging. 

That one in particular may not be a blocker for this upgrade, but should be converted at some point soon after.

chosak:
Thanks @kurtrwall, that's a good point. We have an open issue in [GHE]/platform/357 that addresses the need to update custom template overrides. I'd prefer not to include that in this upgrade unless those overrides break functionality somehow, and leave them for a later cleanup.

kurtrwall:
Yeah I agree with that, I just wanted to make it explicitly known that's the policy since I wasn't aware of that issue and the description of the PR doesn't say anything about templates.

cfgov/v1/models/sublanding_filterable_page.py
```diff
@@ -48,10 +49,13 @@ class SublandingFilterablePage(FilterableFeedPageMixin, FilterableListMixin, CFG
 
     template = 'sublanding-page/index.html'
 
+    objects = PageManager()
```
richaagarwal: Can `objects` live on CFGOVPage so this doesn't have to be defined for each page?

richaagarwal: Oh, nevermind - it has its own custom manager. Hrm...

richaagarwal:
Code changes look good. I won't be able to test anything until I get a new laptop, so fwiw, giving this a 👍

cfgov/v1/models/sublanding_filterable_page.py
chosak: Yeah, this is somewhat non-intuitive. [This](http://docs.wagtail.io/en/v1.6.3/releases/1.6.1.html#multi-level-inheritance-and-custom-managers) is the part of Wagtail release notes for 1.6.1 that motivated this. There's a somewhat long discussion on the changes in Django and Wagtail that motivated this change [here](https://github.com/torchbox/wagtail/issues/2933).

sebworks:
👍

cfgov/v1/atomic_elements/atoms.py
```diff
@@ -99,9 +107,13 @@ class ImageBasic(blocks.StructBlock):
...
 
     def __init__(self, required=True):
-        self.required = required
+        self.is_required = required
         super(ImageBasic, self).__init__()
 
+    @property
+    def required(self):
+        return self.is_required
+
```
kurtrwall: What's the point of continuing the `required` property? Instead of replacing its use with `is_required`.

chosak:
Thanks all for the feedback. Given the significance of this upgrade, I'd like to get signoff from @willbarton and @schaferjh before proceeding.

willbarton:
@chosak I don't think it's worth waiting for @schaferjh to return to get this merged. Let's get it in and that gives us until the next freeze to work out any issues in our checkouts and on build

I just have two higher level questions:

- Are there any changes that would affect our wagtail documentation? (I.e. does it need the documentation update tag?)
- Could you create tasks for the first two items you included in your OP notes?

:shipit:

chosak:
@willbarton thanks, created GHE issues 378/379 to track those two issues, and also added that tag as you suggested. Crossing fingers..

higs4281:
@chosak   college-costs updated to handle DRF 3.1.3  
https://github.com/cfpb/college-costs/pull/287

PR #2457, opened 2016-10-12
Removing handlebars for performance purposes

sebworks:
We really aren't utilizing handlebars so I'm removing it. I'm replacing it with a a lightweight function to replace template tokens.

## Removals

- Removing handlerbars from `package.json`.

## Changes

- Modified `cfgov/unprocessed/js/modules/jquery/cf_notifier.js` to remove `handlebars`. 

## Testing

- Visit `http://localhost:8000/about-us/blog/` and enter `foo@foo`.
- Click the `Sign-up` button.
- Verify that the notification error is being displayed.

## Review

- @contolini 
- @virginiacc 

## Screenshots
screen shot 2016-10-12 at 5 38 19 pm

## Notes

- Eventually we will be removing `cfgov/unprocessed/js/modules/jquery/cf_notifier.js`

## Checklist


cfgov/unprocessed/js/modules/jquery/cf_notifier.js
```diff
@@ -51,9 +49,13 @@ var _notifier = {
       state:   settings.state,
       icon:    icon[settings.state]
     };
-    var template = handlebars.compile( settings.template );
+    var template = settings.template.replace( /\{\{\s?(.*?)\s?\}\}/g,
+      function( match, token ) {
+        return data[token];
```
contolini: If this function were to be generalized for use elsewhere it'd be nice to have it return an empty string instead of `undefined` if the token isn't found (i.e. `return data[token] || ''`) buuuut in the current scenario it works A-OK.

sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2457/commits/db93ec3aa5aad85c58412a86cdba58db0bfbe7dc. Currently working on removing this code so I'm going to leave as-is.

PR #2458, opened 2016-10-13
Hotfix: Fix links in TableBlocks

richaagarwal:
### Overview

TableBlocks may reference internal documents or pages. Before we moved over to TableBlocks, these were stored as a document or page ID and a look-up occurred at the time the page rendered. But with the new TableBlocks, the actual href gets stored. In addition, the migration to TableBlocks did not convert these links to store the href. This PR takes care of that.

It would convert a row previously stored as 
``` 
[u'<p><a id="593" linktype="document">129</a></p>', u'<p>07/11/2016</p>', u'<p>Enforcement Counsel\u2019s Objections to Respondents\u2019\nProposed Exhibits and Witnesses (Filed by <i>Consumer Financial Protection\nBureau</i>)</p>']
``` 
to 

``` 
[u'<p><a href="/documents/129_Enforcement_Counsels_Objections_to_Respondents_Proposed_Exhibits_and_Witnesses.pdf" id="593" linktype="document">129</a></p>', u'<p>07/11/2016</p>', u'<p>Enforcement Counsel\u2019s Objections to Respondents\u2019\nProposed Exhibits and Witnesses (Filed by <i>Consumer Financial Protection\nBureau</i>)</p>']
```

Note that we actually do (at some point) want to mimic the functionality we had before, with the look-up occurring at the time the page is rendered, since the document title, etc. can change. 

### Deploying
Assignees: @cfpb/cfgov-backends 
I finished this later than I was expecting to, and don't feel comfortable attempting my first deploy at this hour, so if someone gets to this in the morning and approves the changes, please feel free to do the hotfix on my behalf, as content is waiting on this fix. Once it's deployed, someone just needs to run `python cfgov/manage.py runscript fix_links_in_tableblocks` in prod. Thanks in advance!

### Testing
You can test this locally with a prod dump. 
- `./refresh-data.sh dump.sql`
-  `python cfgov/manage.py migrate`
- `python cfgov/manage.py runscript fix_links_in_tableblocks`

The console should tell you a few pages have been updated. e.g. 
```
Page edited: "Title XIV Rules: Mortgage Origination" id=1287 revision_id=54346
Page published: "Title XIV Rules: Mortgage Origination" id=1287 revision_id=54346
Page edited: "Title XIV Rules: Mortgage Servicing" id=3273 revision_id=54347
Page published: "Title XIV Rules: Mortgage Servicing" id=3273 revision_id=54347
Page edited: "Integrity Advance, LLC and James R. Carnes" id=1424 revision_id=54348
Page published: "Integrity Advance, LLC and James R. Carnes" id=1424 revision_id=54348
```
You can compare one of them to the one in prod to see what's been fixed. e.g. [http://localhost:8000/policy-compliance/guidance/implementation-guidance/title-xiv-mortgage-rules/](http://localhost:8000/policy-compliance/guidance/implementation-guidance/title-xiv-mortgage-rules/) vs. http://www.consumerfinance.gov/policy-compliance/guidance/implementation-guidance/title-xiv-mortgage-rules/ (Click the links to the left in the table).


cfgov/scripts/fix_links_in_tableblocks.py
```diff
@@ -0,0 +1,67 @@
...
+            for row in rows:
+                for idx, item in enumerate(row):
+                    soup = BeautifulSoup(item, 'html.parser')
+                    links = soup.findAll('a')
+                    if links and convert_links(links):
+                        # Set the item to the modified HTML
+                        row[idx] = str(soup)
+                        changes = True
+        if changes:
+            publish_changes(page.specific)
```
chosak: Why does this use `publish_changes` instead of `save_page`?

One risk of [`publish_changes`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/tests/wagtail_pages/helpers.py#L25) is that if a page is only in draft or shared mode, this code would publish it. I don't know if any of the existing `TableBlock` pages would be in that state, but it might be something we should keep an eye on in future.

cfgov/scripts/fix_links_in_tableblocks.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    updated = False
+    for link in links:
+        linktype = link.get('linktype')
+        object_id = link.get('id')
+        if link.get('href'):
+            continue  # Don't update it if the href is already set
+        if linktype == 'document':
+            doc = get_object_or_404(get_document_model(), id=object_id)
+            filepath = '/' + doc.file.name
```
chosak: Converted document links don't resolve properly for me. When I run this script on page [1424](http://localhost:8000/policy-compliance/enforcement/actions/integrity-advance/), the link

`<a id="1033" linktype="document">176</a>`

gets converted to

`<a href="/documents/2015-CFPB-0029_Document_176_09272016.pdf" id="1033" linktype="document">176</a>`

That URL doesn't resolve for me. From looking at the [urlconf](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/urls.py#L33) I think this is missing the ID, e.g. it should be

`<a href="/documents/1033/2015-CFPB-0029_Document_176_09272016.pdf" id="1033" linktype="document">176</a>`

Instead of using `doc.file.name`, could this use `d.url` (with no need for a leading slash)?

```py
>>> from wagtail.wagtaildocs.models import Document
>>> d = Document.objects.get(pk=1033)
>>> d.url
u'/documents/1033/2015-CFPB-0029_Document_176_09272016.pdf'
```

cfgov/scripts/fix_links_in_tableblocks.py
richaagarwal: @chosak Yeah, I think the changes here are relatively contained to a few pages but I appreciate the concern - @sarahsimpson09 is there any risk of these pages being in a draft state in the next day or so?

cfgov/scripts/fix_links_in_tableblocks.py
richaagarwal: Thanks for catching this!  Updating now.

higs4281:
When I run the script after loading the [latest prod data](https://github.cfpb.gov/CFGOV/platform/wiki/Database-downloads) I get a bs4 error:

```
...
  File "/Users/higginsw/Projects/cfgov-refresh/cfgov/scripts/fix_links_in_tableblocks.py", line 60, in run
    soup = BeautifulSoup(item, 'html.parser')
  File "/Users/higginsw/.virtualenvs/fresh/lib/python2.7/site-packages/bs4/__init__.py", line 176, in __init__
    elif len(markup) <= 256:
TypeError: object of type 'NoneType' has no len()
```

richaagarwal:
@higs4281 My work laptop literally just stopped working so I can't look into that.  @chosak Are you seeing a similar issue?

richaagarwal:
@rosskarchner @grapesmoker @willbarton @higs4281  Y'all wanna join the huddle?

chosak:
@higs4281 yes, I get the same error. Above when I did my review I was using a prod dump that was ~2 weeks old, but with the latest one from the link you provide, I get that error.

sarahsimpson09:
@richaagarwal I just got a request to add a document to the table on this page: http://www.consumerfinance.gov/about-us/payments-harmed-consumers/payments-by-case/

This page doesn't seem to be affected the way others are; all docs seem to open just fine. Is it ok if I save this page as a draft/share with the stakeholder while you guys are working on this or should I wait? (They need the doc uploaded by tomorrow).

chosak:
I pushed a change in 602ce3a that properly handles null columns. Running against the most current production data dump, there are only two pages that get touched by the conversion script ([1](http://localhost:8000/policy-compliance/guidance/implementation-guidance/title-xiv-mortgage-rules/), [2](http://localhost:8000/policy-compliance/enforcement/actions/integrity-advance/)). One of these has links to pages, and one has links to documents, and they both appear correct to me after the latest version of the script is run.

@higs4281 since @richaagarwal is offline due to laptop difficulties, would you be able to re-try using the latest version of this PR?

higs4281:
👍 script runs without error now.

richaagarwal:
@sarahsimpson09 Totally fine to proceed with that page (and even publish if need be) - this work is unrelated.

sarahsimpson09:
@richaagarwal @chosak @higs4281 I just added the Hoffman Law Group row to the table on this shared page: http://content.consumerfinance.gov/about-us/payments-harmed-consumers/payments-by-case/. As you can see, the formatting of the link is a little different from the others. It looks like the others are entered as links, while this one is a doc. Ideally they'd all look the same. I'm thinking this might have something to do with the tableblock update?

richaagarwal:
@sarahsimpson09 That page would not be impacted by the fixes we made today. I can't take a look at content at the moment but is this something that was previously formatted correctly?

sebworks:
It looks like the conversion script added the `jump-link` class: https://github.com/cfpb/cfgov-refresh/blob/82ae7842b19e568741877f6955b58b88347dbc74/cfgov/v1/migrations/0013_update_tables_to_tableblocks.py#L57.

richaagarwal:
@sebworks Ah, is that what's causing it?  I added the `jump-link` class to mimic what was in `cfgov/jinja2/v1/_includes/atoms/hyperlink.html`.

sebworks:
@richaagarwal, the those links should just be regular links because they are contained within a table. @schaferjh, please correct me if I'm wrong.

schaferjh:
@sebworks @richaagarwal I'm not sure I follow your conversation perfectly, but I see two odd things on /about-us/payments-harmed-consumers/payments-by-case/:

Keeping in mind that all of the links in this table are document links:

A. On the live site right now none of the links have document icons but they used to. [Link to live page](http://www.consumerfinance.gov/about-us/payments-harmed-consumers/payments-by-case/). [Link to web archive version of page from May](https://web.archive.org/web/20160520203230/http://www.consumerfinance.gov/about-us/payments-harmed-consumers/payments-by-case/).

B. When I look on content to see that there is a difference between the legacy links and the one that @sarahsimpson09 entered yesterday:
screen shot 2016-10-14 at 9 05 36 am
The only thing I can think of is this: In the old table admin there was an option to choose what kind of content you were entering -- links, plain text, rich text, etc. If you selected "links" you got a dialog like this instead of a rich text field:
screen shot 2016-10-14 at 9 08 11 am
With tableblocks and the rich text modification we've made, now we're entering almost all content through a rich text editor. Is it possible that the migration treated the links entered through the links fields differently than links entered via rich text editor?

----

At the moment I'm more concerned about A than B. -- The document icon is how we inform people that they're clicking on a download rather than a hyperlink. Whereas slightly differently formatted links is bothersome but not as much of an informational issue. I know this PR has been merged -- should I make this comment elsewhere for further discussion?

sebworks:
> I know this PR has been merged -- should I make this comment elsewhere for further discussion?

In regards to A, I know what the problem is and will submit a hotfix. The second issue involves the migration script adding the `jump-link` class to table block links during the migration.

PR #2465, opened 2016-10-17
Removes unnecessary mysql dependency when running manage.py collectst…

vickumar1981:
Removes dependency on mysql when running collectstatic command.

## Additions

-  Adds a flag to check if collectstatic is being run in the settings.py

## Removals

-  

## Changes

-  Only attempts to set database to mysql if COLLECTSTATIC flag is False

## Testing

-  Run collectstatic.  Notice that no connection to a mysql database is necessary for this command to execute correctly.

## Review

- @rosskarchner @chosak @willbarton @vccabral @m3brown

## Notes

-

## Todos

-

## Checklist

…atic command.

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -1,4 +1,4 @@
-import os
+import os, sys
 from unipath import Path
 from ..util import admin_emails
 
@@ -18,12 +18,6 @@ PASSWORD_HASHERS = global_settings.PASSWORD_HASHERS
 
 # see https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-USE_ETAGS
 USE_ETAGS = True
-
-try:
-    import mysql
-    MYSQL_ENGINE = 'mysql.connector.django'
-except ImportError:
-    MYSQL_ENGINE = 'django.db.backends.mysql'
 
 # Application definition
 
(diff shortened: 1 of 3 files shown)


chosak:
I'm able to run `manage.py collectstatic` locally even without a local MySQL database. Can you share the error (stack trace) that you get?

See also [this](https://code.djangoproject.com/ticket/17656) related Django ticket that seems to imply that `collectstatic` should be runnable even without a connection to the DB.

vickumar1981:
@chosak   I guess the bigger question pertains to these lines of code:

```
try:
        import mysql
        MYSQL_ENGINE = 'mysql.connector.django'
    except ImportError:
        MYSQL_ENGINE = 'django.db.backends.mysql'
```

do we need both settings?  or is 'django.db.backends.mysql' sufficient now that jenkins is on RH6?

chosak:
@vickumar1981 I'm not familiar with the need for this code.

vickumar1981:
@chosak   collectstatic will work fine, unless your in that `import mysql` branch of the code, at which point, it will need system dependencies for that.  Assuming that is no longer needed, I think I can close this PR, and just set the engine to the django db backend for mysql.

vickumar1981:
@chosak   Assuming that the 1st setting is no longer needed b/c jenkins is RH6, I can close this PR.  The change is now reflected here - https://github.com/cfpb/cfgov-refresh/pull/2466

vccabral:
Let's close this PR then.

vickumar1981:
k.  so this branch succeeds and deploys to build:

cf.gov-frontend-build/249/console
cf.gov-deploy-django/197/console

vickumar1981:
i'd like to try and explain this PR a bit so that everyone is on the same page.

Let's begin by looking at the germane code block:

```
    try:
        import mysql
        MYSQL_ENGINE = 'mysql.connector.django'
    except ImportError:
        MYSQL_ENGINE = 'django.db.backends.mysql'
```

This conditional import is needed b/c we don't have parity among our build, dev, and prod environments.

https://12factor.net/dev-prod-parity

Parity is an important thing to have, but our build and process rests heavily on an assumption that is incorrect given our environments.  Thus, you cannot just set the backend to 'django.db.backends.mysql' or the build/deploy process will fail (see https://github.com/cfpb/cfgov-refresh/pull/2466).

The relevant settings for mysql depend on having two pip modules installed:

for 'mysql.connector.django' => mysql-connector-python-2.1.3
for 'django.db.backends.mysql' => MySQL-python==1.2.5

These pip modules in turn, depend on having the correct mysql client and development libraries installed, b/c python calls into those libraries to establish that connection.

Now, b/c on the collectstatic step, I've switched to using sqllite, which doesn't need any additional dependencies, we can remove the mysql-connector-python package from both the build and deploy steps, and we can remove the fix-mysql flag in the fabric job (see [GHE]CFGOV/website/blob/master/build.sh#L20)

This results in a simpler build/deploy process with fewer dependencies, and @vccabral, though I haven't tested it yet, it should also build correctly on your ansible setup.

If you notice, this code doesn't have that `try..catch` block.  It sets the db setting to 'django.db.backends.mysql', but explicitly when you are not collecting static assets.

rosskarchner:
👍  Thanks @vickumar1981 "use sqlite" is  much more elegant work-around than  "use this funky off-brand MySQL library, and then immediately uninstall the funky library after deploy"

rosskarchner:
be sure to tweak build.sh and the fabfile to neither install nor uninstall mysql.connector

m3brown:
If we neither install nor uninstall mysql.connector, does this mean that the normal `MySQL-python` package will be built on Jenkins and then deployed to the webserver (i.e. just like all the other python package)?

My impression based on past learning was that the MySQL-plugin built on Jenkins didn't work on our prod webservers, is that accurate? If so, won't `MySQL-python` still need to be reinstalled on the current webservers?

vickumar1981:
@rosskarchner @m3brown yep, those are the next steps.  I"ll test out removing the flag and dependencies.

Just to double check, prod is the other environment where the fix-mysql flag is set, correct?

Also, this branch needs to be rebased before merging.

PR #2466, opened 2016-10-17
Setting db to django mysql backend.

vickumar1981:
Sets database to use django mysql backend explicitly.  Removes try...catch block.

## Additions

-

## Removals

-

## Changes

-  Uses django mysql database backend

## Testing

-

## Review

- @rosskarchner @chosak @willbarton @vccabral @m3brown

## Notes

-

## Todos

-

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -19,11 +19,7 @@ PASSWORD_HASHERS = global_settings.PASSWORD_HASHERS
 # see https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-USE_ETAGS
 USE_ETAGS = True
 
-try:
-    import mysql
-    MYSQL_ENGINE = 'mysql.connector.django'
-except ImportError:
-    MYSQL_ENGINE = 'django.db.backends.mysql'
+MYSQL_ENGINE = 'django.db.backends.mysql'
 
 # Application definition
 


vccabral:
Can this go directly into the DATABASE dict?

vccabral:
LGTM except maybe it doesn't need to be a variable at all now.

vickumar1981:
@vccabral   The variable, `MYSQL_ENGINE`, is used in the following places, so it could go directly in the dictionary, but it would create more code than reduce:

https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/local.py#L14
https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/production.py#L40
https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/base.py#L159

vickumar1981:
@vccabral   I suppose I could refactor those lines of code also, to reduce the redundancy.  What do you think?

vccabral:
@vickumar1981 thanks. Your comment address my concern.

rosskarchner:
@vickumar1981 what does this change improve?

I suspect it will break our current deploy, for any environment that uses the `fix_mysql` flag

We could *eliminate* fix_mysql if we build the mysql-python wheel on the destination server, instead of in Jenkins.

vickumar1981:
@rosskarchner  Sorry, should have put the issue on this PR: 

https://jira.demo.cfpb.gov/browse/DEV-2042

vickumar1981:
@rosskarchner If I run 'collectstatic' in sqllite mode, instead of using mysql, would that still build using drama-free and the `fix_mysql` flag?  

B/c that's what this PR attempts to do here: https://github.com/cfpb/cfgov-refresh/pull/2465

I would like to not have a `fix_mysql` flag if it is possible.

vickumar1981:
@rosskarchner   k, so this branch fails the build.

https://zusa.demo.cfpb.gov/view/Cf.gov/job/cf.gov-frontend-build/245/console
http://zusa.demo.cfpb.gov/view/Cf.gov/job/cf.gov-deploy-django/194/console

If you look at the deploy django job it states:

```
out: django.core.exceptions.ImproperlyConfigured: Error loading MySQLdb module: libmysqlclient_r.so.16: cannot open shared object file: No such file or directory
```

I have a few simple observations here before I close this PR:

1)  The cf.gov-deploy-django job while reporting the mysql error, also incorrectly reports that the job has succeeded.  Unless you actually look at the stack trace, you would have no idea there was even an error, meaning it could be easily overlooked.

2)  If you look at this PR, there's a green checkbox that says `All checks have passed`.  I strongly feel it would be a step forward, if we could ensure that this green checkbox also means that we have a high degree of confidence that a build will pass and deploy correctly.  Currently, that is not the case.

willbarton:
The source of the problem seems to be [GHE]/CFGOV/website/blob/master/fabfile.py#L62. I'm not entirely sure I follow what's supposed to happen there.

> 2) If you look at this PR, there's a green checkbox that says All checks have passed. I strongly feel it would be a step forward, if we could ensure that this green checkbox also means that we have a high degree of confidence that a build will pass and deploy correctly. Currently, that is not the case.

Do you have a suggestion of how we could catch a problem with the fabfile/deployment scripts in travis?

vickumar1981:
@willbarton  I'm open to ideas here.  But, I think we could perhaps open an issue about it.

vickumar1981:
please merge https://github.com/cfpb/cfgov-refresh/pull/2465 instead.

PR #2467, opened 2016-10-17
Move site root data migration to script

chosak:
In #2378 the logic that set up the site root and home page in the `initial-data.sh` script was moved to a Django data migration. Motivated by [GHE]/issues/323, we want to deprecate the use of Django data migrations for these types of operations. This PR moves that setup back into `initial-data.sh`.

This partially reverts the changes in #2378; the other half of those changes were previously reverted (and reimplemented as standalone scripts) in #2446.

This PR also tries to simplify existing scripts by removing `setup_sites.py`, which currently is run after a database dump is loaded. The two workflows are:

1. __Working with a new database:__ Run `drop-db.sh && create-mysql-db.sh` to create a new empty database, then `initial-data.sh` to set up initial testing data. This creates an `admin` superuser, with the password defined by environment variable `WAGTAIL_ADMIN_PW`. It creates two sites (default and staging), and a single home page for both of them with slug `cfgov`.
1. __Working with a database dump:__ Run `refresh-data.sh` to load the db dump. This script now also calls the `initial_data.py` script, which sets things up the same as above: admin user and two local sites.

## Removals

- Removes `setup_sites.py` script as its functionality is now covered by `initial_data.py`.

## Changes

- Removed functionality from v1.0009 migration, replaced by comment recording script location where logic was moved.
- `refresh-data.sh` now enables the `admin` user when working with a database dump; previously a production account or use of `manage.py createsuperuser` was necessary.

## Testing

- With a new database:
 
 ```sh
 $ ./drop-db.sh
 $ ./create-mysql-db.sh
 $ ./initial-data.sh
 $ ./runserver.sh
 ```

 Browse to [http://localhost:8000/admin](http://localhost:8000/admin). You should be able to login with username `admin` and password set to the value of the `WAGTAIL_ADMIN_PW` environment variable. Note that there are two sites setup with the port set to the value of `DJANGO_HTTP_PORT` environment variable. Note that the default site has a hostname of `localhost` and the staging site has a hostname with the value of the `DJANGO_STAGING_HOSTNAME` environment variable.

- With a database dump:

 ```sh
 $ ./drop-db.sh
 $ ./create-mysql-db.sh
 $ ./refresh-data.sh /path/to/dump.sql
 $ ./runserver.sh
 ```

 Browse to [http://localhost:8000/admin](http://localhost:8000/admin). You should be able to login with username `admin` and password set to the value of the `WAGTAIL_ADMIN_PW` environment variable. Note that there are two sites setup with the port set to the value of `DJANGO_HTTP_PORT` environment variable. Note that the default site has a hostname of `localhost` and the staging site has a hostname with the value of the `DJANGO_STAGING_HOSTNAME` environment variable.
 
## Review

- @cfpb/cfgov-backends 

## Screenshots

Updated documentation:

image

## Notes

- Because `initial-data.sh` now enables and sets the `admin` superuser password, it should never be run in production. Please let me know if you would prefer if this behavior stayed as is (`refresh-data.sh` doesn't enable `admin` user; you need to use an existing superuser or create a new one when working with local database dumps) - I modified this to make working with dumps a little bit easier.

## Checklist


cfgov/scripts/initial_data.py
```diff
@@ -4,18 +4,72 @@
...
+from django.db import transaction
+from wagtail.wagtailcore.models import Page, Site
 
+from v1.models import HomePage
 
+
+@transaction.atomic
 def run():
-    print('Running script \'scripts.initial_data\' ...')
+    print('Running script initial_data')
```
richaagarwal: I know your changeset didn't introduce this, but would this be a good time to update the print statements to log statements?  An added benefit would be that we'd have a way of checking if it was accidentally run in prod.

refresh-data.sh
```diff
@@ -17,8 +17,8 @@ refresh_data(){
 	./create-mysql-db.sh
 	echo 'Importing refresh db'
 	mysql v1 -u root -p < $refresh_dump_name
-    echo 'Setting up Sites'
-    ./cfgov/manage.py runscript setup_sites
+    echo 'Setting up initial data'
+    ./cfgov/manage.py runscript initial_data
 }
```
richaagarwal: Is there a reason that `setup_sites` has to exist inside `initial_data` ?  Is it possible for both `initial_data` and `refresh-data` to call that?  I see `refresh-data.sh` (might be time to rename it) as the script to load production data, whereas I see `initial_data` as the script to load made-up data - so they're pretty distinct to me.

richaagarwal:
👍

cfgov/scripts/initial_data.py
chosak: Thanks, good suggestion. I've made this change in 2e02050.

willbarton:
This PR fixes the weird breakage that `wagtail.tests.testapp` causes in #2450!

docs/installation.md
```diff
@@ -291,12 +289,12 @@ unzip it, and then run:
 ./refresh-data.sh /path/to/dump.sql
 ```
 
-This will remove the initial Wagtail admin user that was created by
-the `initial-data.sh` script that was called by `backend.sh`.
-If you need to access the Wagtail admin, create a new user with the following:
+The `refresh-data.sh` script will apply the same changes as the `initial-data.sh` script described above (including setting up the `admin` superuser), but will not apply migrations.
```
willbarton: I had to stop and consider *why* refresh-data.sh wouldn't apply migrations and initial-data.sh would. It might be good to expand on this sometime, but that seems out of scope here. I'm just making a note of it so it's out of my head and into the PR record.

chosak: Thanks for the review! Yeah, maybe to simplify it does make sense to include migrations in `refresh-data.sh` as well. The use case I had in mind was one where you want to test the database as is; not how it would be with the most recent version of the code.

PR #2476, opened 2016-10-19
Minor tweaks

rosskarchner:
## Changes

- more complete usage of the static tag in base.html (for the favicon and og:image)
- brings in a template bug fix that was included in 3.11.2
- remove's the CDN'ed jquery, per feedback from @ascott1 

## Review

- @cfpb/cfgov-backends 

## Checklist


kave:
👍

rosskarchner:
thanks, @kave ;)

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -124,13 +123,7 @@
...
     <!--[if IE 9]><script src="{{ static('js/ie/common.ie.js') }}"></script><![endif]-->
 
-    {# TODO: jQuery CDN is provided in the head to satisfy
-             Google Tag Manager (GTM) requirements.
-             Ideally GTM would handle its own dependency management
-             and initializing jQuery could be moved to the footer. #}
-
-    <script src="//ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js"></script>
-    <script>if (!window.jQuery) { document.write('<script src="{{ static('js/jquery.min.js') }}"><\/script>'); }
+    <script src="{{ static('js/jquery.min.js') }}"></script>
```
Scotchester: Have we confirmed with the data team that they no longer require CDN-loaded jQuery in the `head`? And if they do, how do we reconcile that with the directive from @ascott1?

If they don't need it, then we should proceed, but move the jQuery inclusion to the bottom of `body`, right?

cfgov/jinja2/v1/newsroom/press-resources/index.html
```diff
@@ -160,7 +160,7 @@ <h2 class="press-section_title press-photos-bios_title">Photos and bios</h2>
             <section class="contact-person contact-person__director content-l_col-1-2">
                 <div class="media">
                     <img class="media_image media_image__150 contact-person_photo"
-                         src="{{ static('img/director-cordray-round-300x300.jpg')}"
+                         src="{{ static('img/director-cordray-round-300x300.jpg')}}"
```
Scotchester: Trivial: A space before the closing `{{` here would make this typo more obvious in the future :)

sebworks:
I'm not sure why we aren't loading jQuery from the CDN. Most first time users will already have jQuery stored in their cache if we use the CDN.  I feel like we should still use the CDN and work on removing jQuery from the codebase and the Google Analytics code.

Scotchester:
I wouldn't _think_ so, but, that comment in the HTML seems to indicate otherwise.

rosskarchner:
@sebworks wrote:
> I'm not sure why we aren't loading jQuery from the CDN. Most first time users will already have jQuery stored in their cache if we use the CDN.  I feel like we should still use the CDN and work on removing jQuery from the codebase and the Google Analytics code. 

I think the fear is is that a CDN could be compromised, and then a hostile party could inject code into the website. Far fetched? Maybe, but @ascott1 (and security folks) seem to think it's a risk worth avoiding.

sebworks:
> I'm not sure why we aren't loading jQuery from the CDN. Most first time users will already have jQuery stored in their cache if we use the CDN. I feel like we should still use the CDN and work on removing jQuery from the codebase and the Google Analytics code.

Yeah, but what about Google Analytics and Google Tag Manager? They say 70% of the web is using jQuery. I'd venture a significant portion of the group is loading jQuery from the Google CDN.  Their is also a performance hit for all our users once we have switched over. The minified file is 90kb.  I would rather work on removing jQuery from our codebase ( if possible given time constraints ).

richaagarwal:
Curious about this.  I think I agree with @sebworks, but I'm wondering why the only concern with a CDN is the jQuery (if it were to be compromised)?  Not sure I'm following the logic here. If a CDN is compromised, don't we have other things to worry about as well?

rosskarchner:
@richaagarwal in this case, we're talking about "ajax.googleapis.com" -- so the only problem would be jQuery. 

I 1000% agree with @sebworks (that the benefits of the CDN outweigh the risk), I just don't think it's a decision that's in our hands.

rosskarchner:
I'm going to remove the jQuery change from this PR: from the discussion it's clearly not a "minor tweak" ;)

Internal issue 384 has been opened for further discussion.

sebworks:
👍

cfgov/jinja2/v1/_layouts/base.html
rosskarchner: I've removed this change!

cfgov/jinja2/v1/newsroom/press-resources/index.html
rosskarchner: done!

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -74,8 +74,7 @@
         {% block og_article_author %}{% endblock %}
     <!-- End of Open Graph properties -->
 
-    {# TODO: Explicit favicon link needed for testing. Remove for production. #}
-    <link rel="shortcut icon" type="image/x-icon" href="/favicon.ico">
+    <link rel="shortcut icon" type="image/x-icon" href="{{ static('assets/favicon.ico')}}">
```
rosskarchner: Done!

rosskarchner:
it looks like Travis can't talk to Gitub right now--- maybe part of the big DNS outage?

Scotchester:
Ah yes, that could be it.

rosskarchner:
In a build [two hours ago](https://travis-ci.org/cfpb/cfgov-refresh/jobs/169589338):

> fatal: unable to access 'https://github.com/cfpb/cfgov-refresh.git/': Couldn't resolve host 'github.com'

PR #2477, opened 2016-10-19
Upgrade `gulp-sourcemaps` from `1.6.0` to `2.1.1`

anselmbradford:
## Changes

- Upgrade `gulp-sourcemaps` from `1.6.0` to `2.1.1` in `package.json`.
- Rebuild `npm-shrinkwrap.json`.

## Testing

- throw out `node_modules`.
- `npm install`
- `gulp build` should pass.

## Review

- @cfpb/cfgov-frontends


sebworks:
I honestly feel like we should kill source maps. They overhead isn't worth it. We should just provide easier options to run the code without minifying.

sebworks:
👍

anselmbradford:
> I honestly feel like we should kill source maps. They overhead isn't worth it. We should just provide easier options to run the code without minifying.

This is a good idea!

PR #2485, opened 2016-10-25
Removes gulp-load-plugins

anselmbradford:
`gulp-load-plugins` makes it more difficult to use search to find where a particular npm module is used, since the require statement for the modules is omitted and they come through via the `plugins` variable. 
Some tasks were already not using it, so by removing it completely, there's almost the equivalent amount of code and the modules can now be found in the requires section at the top of the gulp tasks.

## Removals

- Removes `gulp-load-plugins`

## Changes

- Alphabetizes require statements in gulp tasks.

## Testing

- `./frontend.sh` should pass.

## Review

- @cfpb/cfgov-frontends 
- @chosak


sebworks:
👍, @anselmbradford please update branch and merge when ready.

PR #2494, opened 2016-10-27
Update CHANGELOG setting so merges are less painful

richaagarwal:
@cfpb/cfgov-backends @cfpb/cfgov-frontends


anselmbradford:
Note that if you add a changelog entry on a branch in the `Unreleased` section and a release happens in the meantime, when rebasing with master the changelog entry will end up in the list of changes for the release, not in the `Unreleased` section where it should be. Just to be aware of since there won't be any change conflicts.

PR #2508, opened 2016-11-02
refactor get_appropriate_page_version into CFGOVPage

grapesmoker:
This is part of an ongoing effort to rid ourselves of global utility functions that should properly be instance methods. This PR refactors get_appropriate_page_version into the root CFGOVPage class and changes the usage of the function to correspond.

## Changes

- get_appropriate_page_version moved from util to CFGOVPage

## Testing

- Tests that previous mocked the page and called get_appropriate_page_version on the mock have been changed to use database fixtures. All tests pass.

## Review

- @rosskarchner @chosak @richaagarwal @willbarton @higs4281 

## Notes

- While working on this, it occurred to me that it's quite weird to use the `live` and `shared` properties in the way that we do here. Specifically, we're examining the `content_json`, which varies between page revisions, but Wagtail's own docs indicate that `live` and `shared` are to be interpreted as *per page* rather than as *per version*. While this technically works, it results in some confusion because when you load a revision, those two properties are overwritten with the page's value, which means that you can't rely on just reading those values from the returned object, but rather have to parse out the `content_json` into a dict and then look those things up. This seems like a bit of an abuse of the framework, and we might want to think about adding properties that achieve the functionality we want without hijacking Wagtail terminology for something that it might not be well-suited for.

## Checklist


cfgov/v1/models/base.py
```diff
@@ -212,12 +212,27 @@ def specific_categories_query(block, parent_slug):
         return {search_type: queryset for search_type, queryset in
                 related.items() if queryset}
 
+    def get_appropriate_page_version(self, hostname):
```
grapesmoker: No problem. I meant to do that but I just forgot. I can change this.

richaagarwal:
👍 Agree about the revisions stuff, perhaps a discussion for our next BEWD huddle?

grapesmoker:
Updated the code to use @chosak's middleware.

cfgov/v1/tests/test_util.py
```diff
@@ -19,57 +23,61 @@ def setUp(self):
...
+        self.revision_3 = self.new_page.revisions.create(content_json=json.dumps(content_json))
+
+        # rev 4
+        content_json['title'] = 'revision 4'
+        content_json['live'] = False
+        content_json['shared'] = False
+        self.revision_4 = self.new_page.revisions.create(content_json=json.dumps(content_json))
+
+    def test_production_returns_first_live_page(self):
+        self.request.is_staging = False
```
grapesmoker: I'm not really sure why this is hard to understand. These objects are attached to `self` because they are carried over from test to test; they are initialized in `setUp` so it's pretty clear what they're for. I prefer the method of factoring out common operations into `setUp` because it lets you see what is the same across all tests. For this particular series of tests, I don't feel super strongly about it and I can change it if people really prefer @chosak's version, but honestly I don't think it really improves readability in any meaningful way.

chosak: I don't feel strongly either so don't let me hold back this PR. This is a nice improvement so thanks for doing this refactoring.

One comment on "they are carried over from test to test" -- there's no guarantee of test order so changes in one test to a class property shouldn't be required for other tests to pass. Each test should only count on the initialization in `setUp`; there shouldn't be anything else getting carried over.

In this case, `self.request` is just a mock object so there's no advantage to creating it in `setUp` versus just making it inline in the test -- it's not particularly hard to understand but my point was just that it's an extra thing to go lookup to figure out how the test is working.

Another example would be [`test_shared_returns_None_for_only_draft_versions`](https://github.com/cfpb/cfgov-refresh/pull/2508/files#diff-d1f37930c781c3ba60f18e1c25884f40R67), which after this PR looks like:

```py
def test_shared_returns_None_for_only_draft_versions(self):
     self.revision_1.delete()
     self.revision_2.delete()
     self.revision_3.delete()

    version = self.new_page.get_appropriate_page_version(self.request)
    self.assertIsNone(version)
```


Let's say this test fails because of some changes and I'm looking at it as someone new to this code. It starts with deletion of three class member variables, which I have no way of understanding without looking through `setUp`. It calls the method under test using another class member `self.request` which isn't defined locally. I really have to dig into to all of these variables to understand what this is supposed to be testing.

Another way to do this might be to have helper methods in the class, so then you could have psuedocode like:

```py
def test_shared_returns_None_for_only_draft_versions(self):
    draft_page = self.create_new_page(live=False, shared=False)
    draft_revision = self.create_page_revision(draft_page, live=False, shared=False)
    self.assertIsNone(page.get_appropriate_page_version(request=Mock())
```

This code is more self-contained and readable and a new developer can tell what it's doing just from looking at the steps: first make a new page, then make a single draft revision, then assert that there are no appropriate page versions for that page.

(Some of those helper functions could be added to the ones we already have in `v1.tests.wagtail_pages.helpers`.)

This might be more a matter of preference so again I don't mean to hold this up; I do find it interesting to compare styles and understand the pros and cons of each.

grapesmoker: > One comment on "they are carried over from test to test" -- there's no guarantee of test order so changes in one test to a class property shouldn't be required for other tests to pass. Each test should only count on the initialization in setUp; there shouldn't be anything else getting carried over.

True, I hope I didn't imply that this was the case but if I did I just want to make it clear that I recognize this.

>Let's say this test fails because of some changes and I'm looking at it as someone new to this code. It starts with deletion of three class member variables, which I have no way of understanding without looking through setUp. It calls the method under test using another class member self.request which isn't defined locally. I really have to dig into to all of these variables to understand what this is supposed to be testing.

Yeah, this isn't great; I could probably change this to just pull up the various revisions and change their `live` and `shared` properties to match the test conditions. I think for a test with more complicated logic your approach would be the more correct one.

grapesmoker:
If there are no objections, can I merge this?

chosak:
:shipit: 👍

cfgov/v1/models/base.py
```diff
@@ -212,12 +212,26 @@ def specific_categories_query(block, parent_slug):
...
+                if page_version['shared']:
+                    return revision.as_page_object()
+
     def get_breadcrumbs(self, request):
         ancestors = self.get_ancestors()
         home_page_children = request.site.root_page.get_children()
         for i, ancestor in enumerate(ancestors):
             if ancestor in home_page_children:
-                return [util.get_appropriate_page_version(request, ancestor) for ancestor in ancestors[i+1:]]
+                return [ancestor.get_appropriate_page_version(request) for ancestor in ancestors[i+1:]]
```
rosskarchner: Seems like we need to call 'specific' here, since get_ancestors can return Page objects (which don't have the method). I included that fix on #2518

PR #2509, opened 2016-11-02
Add flake8 Python linting

willbarton:
This PR adds a flake8 config file, adds flake8 to travis, and cleans up the Python style everything except for migrations and cfgov/v1/tests

There are a few things I want to bring up for discussion based on this:

1. By ignoring v1/tests, [this configuration](https://github.com/cfpb/cfgov-refresh/blob/flake8/.flake8) will let us run flake8 in Travis and ensure that the cleanup in this PR does not regress. However, it has a side-effect, in that anytime one runs `flake8` from the root of cfgov-refresh, it will ignore v1/tests.

2. There is some pretty serious refactoring needed that this PR *does not do*, and which I is out of scope. [This `if` condition is probably the worst example](https://github.com/cfpb/cfgov-refresh/blob/flake8/cfgov/sheerlike/filters.py#L58), but there are other less bad examples as well. 

3. This still doesn't help [`urls.py` very much](https://github.com/cfpb/cfgov-refresh/blob/flake8/cfgov/cfgov/urls.py)

4. I am completely ignoring [our settings files](https://github.com/cfpb/cfgov-refresh/blob/flake8/.flake8#L33). It strikes me that the readability of those might have different concerns than simply Python linting, but I'd welcome some discussion of that.

5. I'd love to include a max-complexity test, but currently there are parts of the cleaned code that fail ([like this](https://github.com/cfpb/cfgov-refresh/blob/flake8/cfgov/core/management/commands/_helpers.py#L90). This goes back to the need for refactoring, above.

So, I'm tagging all the back-ends that are working on cfgov-refresh right now (from platform and data and research) so that we can hash out some of these.

Once we work through this, I'd like to squash these commits when we merge, so there can be only one.

## Additions

- flake8 config file
- flake8 dependencies
- flake8 as a default travis command

## Changes

- General flake8-compliant style cleanup

## Testing

- Run `flake8` in the root of `cfgov-refresh`. There should be no errors.

## Review

- @chosak 
- @richaagarwal 
- @grapesmoker 
- @higs4281 
- @rosskarchner 
- @hillaryj 

## Checklist


willbarton:
I meant to add reasons for why each of those ignored packages is ignored:

- v1: too much potential for conflict. We'll need to handle this one piecemeal.
- jobmanager: most of the flake8 complaints are with the Django models and admin, which we'll be cleaning up anyway.
- data_research: I'm not sure if there's a potential for me to introduce conflicts with work-in-progress here. @hillaryj?

hillaryj:
@willbarton data_research doesn't have any outstanding changes, so flake8 shouldn't introduce conflicts.

willbarton:
@hillaryj yay! I'll take care of it before it does:D

willbarton:
Update: apparently flake8's `--exclude` overrides the configured value, so it's impossible to configure all the other paths and then include `--exclude v1` in the invocation. You have to specify all the exclusions either on the command line or in the config file. So, splitting them up is not really an option.

cfgov/cal/admin.py
```diff
@@ -28,21 +30,23 @@ def start_date(self, obj):
...
+
 class CFPBCalendarImportICSFileAdmin(admin.ModelAdmin):
 
     def changelist_view(self, request, extra_context=None):
         context = {}
         context.update(extra_context or {})
-        context.update({ 'icsfile': None,})
-        context.update({ 'calendar_record': None,})
-        context.update({ 'processed_events': None,})
+        context.update({'icsfile': None, })
```
chosak: Why these changes to add extra spaces after the last comma? That doesn't trigger a warning for me with flake8. See [PEP8 Pet Peeves](https://www.python.org/dev/peps/pep-0008/#pet-peeves).

cfgov/cal/admin.py
```diff
@@ -51,28 +55,29 @@ def changelist_view(self, request, extra_context=None):
...
-                if(file_type != "text"):
-                    error.append( (['There was an error with the File'],["Incorrect File Type"]) )
-                    context.update({ 'error': error,})
-                    return super(CFPBCalendarImportICSFileAdmin,self).changelist_view(request,context)
+                if (file_type != "text"):
+                    error.append((['There was an error with the File'],
+                                  ["Incorrect File Type"]))
+                    context.update({'error': error, })
+                    return super(CFPBCalendarImportICSFileAdmin,
+                                 self).changelist_view(request, context)
```
chosak: This might be one of the stylistic personal preferences that you mention, but this feels hard to read versus the alternate

```py
return super(
    CFPBCalendarImportICSFileAdmin,
    self
).changelist_view(request, context)
```

Alternatively, because there are a few places in this function that use `super(CFPBCalendarImportICSFileAdmin, self).changelist_view`, could this code be cleaned up instead using something like:

```py
base_view = super(CFPBCalendarImportICSFileAdmin, self).changelist_view
...
base_view(request, context)
```

cfgov/cal/event.py
```diff
@@ -1,10 +1,12 @@
-from cal.models import CFPBCalendar, CFPBCalendarEvent
-from datetime import date, datetime
+from cal.models import CFPBCalendarEvent
+from datetime import datetime
+import re
```
chosak: Nitpicky, and not a PEP8 thing I suppose, but it'd be nice to follow [PEP8 import guidelines](https://www.python.org/dev/peps/pep-0008/#imports):

> Imports should be grouped in the following order:
 1. standard library imports
 2. related third party imports
 3. local application/library specific imports

 You should put a blank line between each group of imports.

So, alternatively:

```py
import re

from datetime import datetime

from cal.models import CFPBCalendarEvent
```

tox.ini
```diff
@@ -23,6 +23,7 @@ changedir = {toxinidir}/cfgov
 commands =
     coverage erase
     coverage run --source='.' manage.py test {posargs}
+    flake8
```
chosak: Minor request: could we add the `flake8` command to `run_travis.sh` instead of as part of tox? Linting seems different to me than unit testing. The frontend code has them set up as separate gulp tasks.

In addition, when iterating on code, I find it useful to only test the specific code I'm working on, for example repeated invocations of `tox -e fast path.to.my.tests`. This is a quick way to do test-driven development. The change here runs flake8 on the _entire_ codebase even if tox is being invoked on a single file.

willbarton: That is a very good idea! I generally use tox the same way, and that hadn't occurred to me.

cfgov/cal/admin.py
willbarton: Weird, I get:

```
cfgov/cal/admin.py:43:40: E231 missing whitespace after ','
```

We can also eliminate the comma entirely (or ignore the error).

chosak: Oops, you're right. I made a mistake in how I tested this. My personal preference here would be to eliminate the trailing comma where the closing parenthesis immediately follows. So this is still useful:

```py
data = {
    'foo': 1,
    'bar': 2,
}
```

because then you can add/remove elements without changing extra lines. But in this case:

```py
data = {'foo': 1, 'bar': 2, }
```

the trailing comma isn't really serving any purpose.

higs4281:
count me in team (...)

question for @chosak:
datetime is part of the standard library, so why separate it from the `re` import?

chosak:
@higs4281 I like the [Django style guide](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#imports) for imports, which is slightly stricter (emphasis added):

> Put imports in these groups: future, standard library, third-party libraries, other Django components, local Django component, try/excepts. Sort lines in each group alphabetically by the full module name. **Place all import module statements before from module import objects in each section.**

higs4281:
@chosak sold!

cfgov/cal/event.py
willbarton: So, flake8-import-order doesn't support Django's particular import styles. I've cleaned up all the imports though in 37df319 (for the all the apps in the project that I'm touching so far). But I'd still love a good way to check imports on an ongoing basis.

willbarton:
Bringing this PR back up to date, @chosak and @higs4281. I've addressed the imports as they exist now (but I'd like to find a solution to checking them with flake8), and I've added the jobmanager. The only Python code in `cfogv/` that I haven't cleaned up yet is `v1`.

I'm inclined to get this in, if it looks good to you all, and then we can address a cleanup of `v1`. This will give us automated linting of all the Python that's *not* in `v1` already.

higs4281:
👍 monumental!

.flake8
```diff
@@ -0,0 +1,42 @@
...
+    cfgov/es_mappings,
+    cfgov/jinja2,
+    cfgov/static_built,
+    cfgov/templates,
+    cfgov/unprocessed,
+
+    # Generated migration files will throw errors. We need to find a way
+    # to exclude django-generated migrations while including
+    # manually-written migrations.
+    cfgov/*/migrations/*,
```
chosak: Does the second `*` here mean "exactly one directory", or could it include multiple levels? If the latter, then we might have some non-migration code that matches this pattern. I'm thinking mainly of tests, like `cfgov/v1/tests/migrations/`, but we also have `cfgov/v1/util/migrations.py` which one could imagine being a package down the road. Does this file support regexes, e.g. something like `*cfgov/*/migrations/\d{4}_*.py`?

cfgov/cal/admin.py
```diff
@@ -28,21 +32,23 @@ def start_date(self, obj):
...
-        context.update({ 'calendar_record': None,})
-        context.update({ 'processed_events': None,})
+        context.update({'icsfile': None})
+        context.update({'calendar_record': None})
+        context.update({'processed_events': None})
 
         leadership = CFPBCalendar.objects.all().order_by("title")
 
-        context.update({ 'leadership': leadership,})
+        context.update({'leadership': leadership, })
```
chosak: Why the extra comma and trailing space on this line? Also, it seems like this line and the various ones above could be combined to a single `context.update({ dict with lots of entries })`. Maybe that crosses the line from PEP8 to more general cleanup.

cfgov/cal/admin.py
```diff
@@ -51,28 +57,31 @@ def changelist_view(self, request, extra_context=None):
             if icsfile:
                 file_type = icsfile.content_type.split('/')[0]
 
-                if(file_type != "text"):
-                    error.append( (['There was an error with the File'],["Incorrect File Type"]) )
-                    context.update({ 'error': error,})
-                    return super(CFPBCalendarImportICSFileAdmin,self).changelist_view(request,context)
+                if (file_type != "text"):
```
chosak: Not PEP8, and maybe out of scope for this PR, but it'd be nice to standardize on use of single quotes and change double quotes where we see them. See the line that starts with `error.append` that adds two strings that use different quotes!

cfgov/cal/admin.py
```diff
@@ -51,28 +57,31 @@ def changelist_view(self, request, extra_context=None):
...
+                context.update({'error': error, })
+                return super(CFPBCalendarImportICSFileAdmin,
+                             self).changelist_view(request, context)
 
             leader_id = request.POST['drpLeader']
 
             calendar_record = CFPBCalendar.objects.get(pk=int(leader_id))
 
-            context.update({ 'calendar_record': calendar_record,})
+            context.update({'calendar_record': calendar_record, })
```
chosak: Same comment on trailing comma/space here.

cfgov/cal/event.py
```diff
@@ -60,40 +63,47 @@ def _set_value(self, i):
...
 
-    # make sure uid is unqiue - this is do to recurring events sharing the same uid
+    # make sure uid is unqiue - this is do to recurring events sharing
+    # the same uid
     def _create_unique_id(self):
 
-        temp = str(self.e.uid)+'@'+str(self.e.dtstart).replace(" ","").replace(":","").replace("-","")
+        temp = (str(self.e.uid)
+                + '@'
+                + re.sub(r'[ -:]', '', str(self.e.dtstart)))
```
chosak: :+1: for use of regex here.

cfgov/cal/models.py
```diff
@@ -29,7 +31,7 @@ def day(self):
         return self.dtstart.date()
 
     def save(self):
-        if self.uid == None or self.uid == "":
+        if self.uid is None or self.uid == "":
```
chosak: A more concise way of doing this might be to just check `if not self.uid`.

cfgov/core/admin.py
```diff
@@ -1,3 +1,3 @@
-from django.contrib import admin
+# from django.contrib import admin
```
chosak: Any reason this entire file can't just be deleted?

cfgov/core/management/commands/_helpers.py
```diff
@@ -125,7 +131,7 @@ def authenticate_user(self):
         if user:
             if not user.is_active or not user.is_superuser:
                 raise PermissionDenied('This user does not have permissions to'
-                                       + 'perform this operation.')
+                                       'perform this operation.')
```
chosak: Not related to your change, but isn't there a missing space between these two phrases?

cfgov/core/models.py
```diff
@@ -1,3 +1,3 @@
-from django.db import models
+# from django.db import models
```
chosak: Can this file be deleted?

cfgov/cal/admin.py
willbarton: So much agreement.

cfgov/cal/admin.py
willbarton: Yeah, I agree these should be a single context.update. I purposefully didn't do that with this PR though.

.flake8
willbarton: Hmm, good point, I am going to play around with it a little bit. That regex didn't work, unfortunately.

cfgov/cal/event.py
willbarton: So apparently I did a bit of refactoring.

cfgov/scripts/archive_events.py
```diff
@@ -24,12 +29,14 @@ def run():
                         if event.end_dt < timezone.now():
                             if event.can_move_to(archived_events):
                                 event.move(archived_events, pos='last-child')
-                                loggger.info(event.title + ' Event .....archived')
+                                logger.info(event.title +
+                                            ' Event .....archived')
```
higs4281: would be nice to use 3-period ellipses rather than 3, 4, and 5

cfgov/cfgov/urls.py
```diff
@@ -39,60 +37,101 @@
...
+    url(r'^owning-a-home/closing-disclosure/',
+        include(oah.urls_for_prefix('closing-disclosure'))),
+    url(r'^owning-a-home/explore-rates/',
+        include(oah.urls_for_prefix('explore-rates'))),
+    url(r'^owning-a-home/loan-estimate/',
+        include(oah.urls_for_prefix('loan-estimate'))),
+
+    url(r'^owning-a-home/loan-options/',
+        include(oah.urls_for_prefix('loan-options'))),
+    url(r'^owning-a-home/loan-options/FHA-loans/',
```
willbarton: Yes we should! GHE/platform/issues/297

cfgov/scripts/archive_events.py
willbarton: So nice.

cfgov/v1/models/browse_page.py
```diff
@@ -68,7 +69,7 @@ def add_page_js(self, js):
         js['template'] += ['secondary-navigation.js']
 
     def full_width_serif(self):
-        return true
+        return True
```
higs4281: future bug: squashed!

higs4281:
I was initially worried about leaving cfgov/core without any models.py file, because it has management commands. But that was a pre-1.7 problem, and the commands run without error on this branch, as do migration operations. So forward ho!

cfgov/v1/models/browse_page.py
richaagarwal: lol that function isn't even called anywhere 😒

willbarton: 300px-paris_tuileries_garden_facepalm_statue

richaagarwal:
Do you/ will you need help resolving all the conflicts?  Also, what would it take to get this to a ready state so that there isn't an ongoing conflict mess?

willbarton:
@richaagarwal nah, I'll take care of the conflicts. I'd like to see that we're okay merging it before I do though, because otherwise I'll just have to do it again :)

FWIW, I've been merging master in with `-s theirs` so that style choices in this PR don't override code changes that've been made elsewhere.

cfgov/scripts/archive_events.py
willbarton: Ask and ye shall receive…

willbarton:
@richaagarwal @chosak @higs4281 @rosskarchner Alright, This is once again up-to-date post-machete. What can we do to get this merged and have some automatic linting of Python PRs?

willbarton:
giphy

PR #2517, opened 2016-11-04
Conference registration form fixes

chosak:
This PR (worked on with @hillaryj) makes ~~two minor~~ three changes to the conference registration form block:

1. Fixes a logic bug when checking registration capacity for a given set of GovDelivery codes.
2. Adds useful help text to the form's appearance in the Wagtail editor to let content editors know that certain fields (non-customizable ones like email) aren't expected to appear before Preview/Publish.
3. Modifies form to use only a single GovDelivery code for signup, instead of supporting multiple codes.

## Additions

- Added help text to the conference registration form's heading element.

## Changes

- Fixed conference registration form capacity check.
- Updated `ConferenceRegistration` model to use only a single GovDelivery code.
- Disables most logging during backend unit tests.

## Testing

- Run unit tests with `tox -e fast data_research.tests.test_handlers`.
- In the Wagtail admin, create a new BrowsePage and add a Conference Registration Form. Add a GovDelivery code and verify that the page can be previewed.
- With proper local GovDelivery credentials in your environment (`GOVDELIVERY_BASE_URL=https://stage-api.govdelivery.com/`, `GOVDELIVERY_ACCOUNT_CODE`, `GOVDELIVERY_USER`, and `GOVDELIVERY_PASSWORD`), try creating a form with an invalid GovDelivery code and verify that it can't be submitted.
- Try creating a form with a valid code and verify that it can be submitted.

## Review

- @cfpb/cfgov-backends @hillaryj 

## Screenshots

New help text that appears at the top of the form:

image

## Notes

- Unfortunately there's no way to add general help text to the top of a block, so we attached it to the first (heading) element.

## Todos

- ~~Per [this comment](https://github.com/cfpb/cfgov-refresh/compare/reg-form-fix-capacity-check?expand=1#diff-2ab90947aaf17de9ab23cdbff150638fR36), currently the way codes are checked against current registrations is fragile (a simple string contains against a JSON-encoded list of codes). This will cause problems if codes are not unique enough or are substrings of each other.~~ This is no longer an issue as of 1d60d67.

## Checklist


chosak:
I'm happy to see that our automated test for missing migrations worked, while at the same time so unhappy that modifying Wagtail block help text requires a new migration.

cfgov/data_research/handlers.py
```diff
@@ -31,10 +31,14 @@ def process(self, is_submitted):
...
-            self.block_value['codes']
-        )
+        query = Q()
+
+        # TODO: Let's perhaps use a less-dangerous way to compare
+        # codes than a straight string-compare against a JSON list
+        # Alternatively (less-safe), make sure codes are strictly
+        # very-unique and not subsets of each other.
+        for code in self.block_value['codes']:
+            query &= Q(codes__contains=code)
```
willbarton: I'm curious to hear more about this... I don't know that I'm fully aware of what the return JSON looks like.

I'd recommend sticking something on either our or data & research's backlog.

chosak:
I've made some updates to this PR to clean up the code a bit and added some testing steps above.

One change I made is to add some logging around GovDelivery signup to help debug form failure messages. Log messages will never contain PII. On success, the log will include a message like

```
subscribing to GovDelivery
Starting new HTTPS connection (1): stage-api.govdelivery.com
```

On failure, the log includes a stack trace like

```
error subscribing to GovDelivery
  File "/cfgov-refresh/cfgov/data_research/handlers.py", line 77, in subscribe
    subscription_response.raise_for_status()
  File "/cfgov-refresh/lib/python2.7/site-packages/requests/models.py", line 851, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
HTTPError: 422 Client Error: Unprocessable Entity
```

This includes the API error from GovDelivery (the above case is if an invalid code was given) but not any of the submission details.

@willbarton are you okay with this additional logging?

willbarton:
I am ok with that additional logging!

cfgov/data_research/blocks.py
```diff
@@ -5,11 +5,21 @@
...
+    sessions = blocks.ListBlock(
+        blocks.CharBlock(label='Session'),
+        label='Sessions attending'
+    )
+    capacity = atoms.IntegerBlock(help_text=(
+        'Enter an integer that will be the conference attendance limit.'
+    ))
+    at_capacity_message = organisms.FullWidthText(help_text=(
+        'Enter a message that will be shown when the event is at capacity'
+    ))
```
willbarton: 👍 for style cleanup.

sebworks:
Is this interim architecture, or is this how we are generally going to handle forms?

chosak:
@sebworks I hope interim. Do you have any suggestions?

sebworks:
>  Do you have any suggestions?

Was just thinking that we should have a more general, flexible, and reusable approach to forms. Seems like we will end up with a lot of code for similar forms.

chosak:
@sebworks agreed. For example I'd like to understand better why [this template](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/conference-registration-form.html#L36) needs to explicitly iterate over all of the fields rather than using Django default form rendering in templates. That'd at least reduce the boilerplate per form.

PR #2523, opened 2016-11-09
Refactor Activity Snippets and sort by latest 

richaagarwal:
## Overview
Per GHE 458: 
Currently we show "Latest Activities" on the About Us page, and they're in no particular order -- http://www.consumerfinance.gov/about-us/
We should sort them by latest date.

Because much of this logic existed in a template, I decided to do some heavy refactoring as part of this fix. There's a lot here so please take your time reviewing and ask questions if anything looks questionable.

## Test
To test, notice this behavior locally before pulling down the branch, and after, at [http://localhost:8000/about-us/](http://localhost:8000/about-us/
)

### DIFF
--- a/cfgov/jinja2/v1/_includes/macros/activity-snippets.html
+++ b/cfgov/jinja2/v1/_includes/macros/activity-snippets.html
@@ -25,7 +25,7 @@
    ========================================================================== #}
 
 {% macro render(tag,
-                activity_types=['posts', 'op-ed', 'newsroom', 'press release', 'speech', 'testimony'],
+                activity_types=['blog', 'op-ed', 'newsroom', 'press release', 'speech', 'testimony'],
                 include_date_flag=false,
                 number_columns=1
                )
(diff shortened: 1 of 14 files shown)


cfgov/v1/util/categories.py
```diff
@@ -0,0 +1,25 @@
...
+        return None
+    subcategories_dict = dict(ref.categories)
+    for unicorn in ['blog', 'newsroom', 'research-reports']:
+        if unicorn in selected_categories:
+            if unicorn == 'research-reports':
+                unicorn = 'Research Report'
+            for category in subcategories_dict[unicorn.title()]:
+                selected_categories.append(category[0].lower())
+    logger.info('Filtering by categories {}'.format(selected_categories))
+    return selected_categories
```
chosak: It took me a minute looking at this code and at [this line](https://github.com/cfpb/cfgov-refresh/pull/2523/files#diff-a20999810035f4df9b089b5ce85e66b6R115) to see that this function both modifies the list passed in and then also returns it. Do you mean to do the former, or only the latter?

cfgov/v1/routing.py
```diff
@@ -46,3 +53,32 @@ def get_page_relative_url(page, site):
...
+@register.assignment_tag(takes_context=True)
+def get_page_state_url(context, page):
+    page = Page.objects.get(id=page.id).specific
+    if not page.live and not page.shared:
+        return None
+    url = page.url
+    if url is None:  # If page is not aligned to a site root return None
+        return None
+    page_hostname = urlparse(url).hostname
+    staging_hostname = os.environ.get('DJANGO_STAGING_HOSTNAME')
```
chosak: Could this use `settings.STAGING_HOSTNAME` to avoid checking the environment?

cfgov/jinja2/v1/_includes/macros/activity-snippets.html
```diff
@@ -72,67 +72,29 @@
...
-        {% set header = 'Reports' %}
-        {% set categories = ['consumer-complaint', 'super-highlight', 'data-point', 'industry-markets', 'consumer-edu-empower', 'to-congress'] %}
-        {% set feed = cfgovpage_objects.live_shared(request.site.hostname).filter(categories__name__in=categories)[:quantity] %}
-    {% else %}
-        {% set icon = category_icon.render('blog') %}
-        {% set header = 'Feed' %}
-        {% set categories = ['at-the-cfpb', 'policy_compliance', 'data-research-reports', 'info-for-consumers'] %}
-        {% set feed = cfgovpage_objects.live_shared(request.site.hostname).filter(categories__name__in=categories)[:quantity] %}
-    {% endif %}
+    {% set feed = get_latest_activities(activity_type, request.site.hostname) %}
```
chosak: 💯

cfgov/v1/templatetags/share.py
```diff
@@ -1,10 +1,7 @@
-import os
-from urlparse import urlparse
-
 from django import template
-from wagtail.wagtailcore.models import Page
 
 from v1.models import CFGOVUserPagePermissionsProxy
+from jinja2 import contextfunction
```
richaagarwal: Ah thanks, yeah, I was moving around things (even though I *thought* I ran flake8 afterward) but I'll do a final clean-up

cfgov/v1/util/categories.py
richaagarwal: Yeah good catch. So right now I'm doing both.  Originally, it just modified the list and didn't return anything.  But now I need it to return the categories for the way I'm using it in this PR.  Would it be better to create a new list, return that, and then assign it in `forms.py` ? The only thing with that is I'd be copying over one list to another, which seems sort of unnecessary. But I don't know what the convention is for this sort of thing.

chosak: I think you'd have to copy it over, unfortunately, since you're using the `get` method of the dictionary instead of `[]`:

```py
>>> def function_that_alters_a_list(my_list):
...    my_list.append('d')
...    return my_list
...
>>> my_dict = {'my_list': ['a', 'b', 'c']}
>>> my_dict.get('my_list')
['a', 'b', 'c']
>>> function_that_alters_a_list(my_dict.get('my_list'))
['a', 'b', 'c', 'd']
>>> my_dict.get('my_list')
['a', 'b', 'c']
# The function didn't alter the original dictionary.
>>> my_dict = {'my_list': ['a', 'b', 'c']}
>>> function_that_alters_a_list(my_dict['my_list'])
['a', 'b', 'c', 'd']
>>> my_dict.get('my_list')
['a', 'b', 'c', 'd']
# The function alters the original dictionary.

richaagarwal: Meant to ask this sooner, but is it bad to both modify and return a list?   (Or is what you're saying here that the modifying isn't actually working?)

chosak: No, It's not necessarily bad as long as it's clear to the caller what the side effects of the function are. I rescind my comment above - I was mistaken that somehow calling `dict.get` and passing that into this method meant that the list wouldn't get modified. The code as written is :+1:.

richaagarwal:
This is ready for review again!  Also, once this is on build, I'd like for you to take a look @schaferjh

schaferjh:
@richaagarwal sounds good -- would you just move it into "acceptance testing" on our board when it's merged? thx!

cfgov/v1/routing.py
rosskarchner: Didn't we add a function for that recently?

rosskarchner: Ah-- nm, it's a method that we stick on to the request via middleware, not useful here

richaagarwal:
I don't know why Github isn't displaying previous comments made (including one I made to say I addressed it), so just posting again to say I've addressed the SETTINGS issue and decided to merge.

PR #2527, opened 2016-11-10
Add feedback variants that record referring page

higs4281:
Owning a Home needs two feedback variants that can reside on their own pages. 
- One is a feedback page that can be referred to from other pages that captures the referrer and allows free-form feedback.
- The second captures referrer, solicits owning-a-home feedback and asks for more user information, including email.

## Additions

- Add two forms to handle variants.

## Changes
- Adapt the Feedback model, form handler and template to the variations.

## Testing

- Fire up locally, run `migrate --fake` 
- Create a feedback page under Owning a Home using the slug 'feedback' and delete the default type for 'was it helpful' and 'radio intro'
- check out http://localhost:8000/owning-a-home/feedback/
- Try entering feedback with no comment
- Try entering feedback with a comment. The success redirect will 404 because page.url points to www, but if you arrow back to the previous page you should see a success message.

## Review

- @cfpb/cfgov-backends @cfpb/cfgov-frontends 

## Todos
- This reduces repeated model, handler and template files but makes the admin more complicated because of unused fields in certain cases.
- Adding more tests but wanted to get eyes on this approach.
- OAH team is deciding on content to add to the sidebar at /owning-a-home/feedback/


cfgov/jinja2/v1/sublanding-page/index.html
```diff
@@ -28,7 +28,7 @@
                         block__flush-sides">
                 {{ featured_content.render(block.value) }}
             </div>
-        {% elif block.block_type == 'feedback' %}
+        {% elif block.block_type == 'feedback' or block.block_type == 'referred_feedback' %}
```
chosak: If we could somehow use class inheritance to unify these feedback blocks, then this could be made simpler and we wouldn't have to enumerate each block type here.

higs4281:
> Could the existing one be a subclass of this new one, but just adding the is_helpful flag?
@chosak I didn't think about subclassing back from the new. That seems more promising than extending the current class.

cfgov/v1/migrations/0024_create_referred_feedback_model_20161110_1839.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
```
chosak: Minor comment: since this migration filename has the index (`0024`) and the custom name, the date and timestamp isn't needed.

sebworks:
👍

cfgov/jinja2/v1/_includes/blocks/feedback.html
```diff
@@ -57,29 +59,115 @@
...
+    </fieldset>
+
+        {% if value.optional_intro %}
+    <fieldset class="o-form-input-group">
         <div class="form-l_col
-                    form-l_col-1
-                    form-l_col__flush-bottom">
-            <button class="btn" type="submit">{{ value.button_text }}</button>
+                    form-l_col-1">
+            <h3>
```
sebworks: It might make sense to wrap these in `if` statements so they won't be included in the template if not provided.

higs4281: @sebworks 
I wrapped sections based on the 3 form flavors in `if` statements, so sections can be turned on/off based on admin choices.

The comment box is common to all, so I don't think it needs to be wrapped

higs4281:
@chosak I simplified model, block and template includes, somewhat at the expense of admin complexity. Admin help_text mitigates that a bit, but it's still not optimal.

cfgov/v1/handlers/blocks/feedback.py
```diff
@@ -14,7 +16,12 @@ def __init__(self, page, request, block_value):
 
     def process(self, is_submitted):
         if is_submitted:
-            form = FeedbackForm(self.request.POST)
+            if 'help-us-improve' in self.page.url:
```
chosak: Is there some way to avoid having to hardcode these URL checks? Is using `self.block_value` an option?

cfgov/v1/models/base.py
```diff
@@ -549,19 +569,24 @@ def too_many_attempts(self, value, timestamp):
...
+    submitted_on = models.DateTimeField(auto_now_add=True)
     page = models.ForeignKey(
         Page,
         related_name='feedback',
         null=True,
         on_delete=models.SET_NULL,
     )
-    submitted_on = models.DateTimeField(auto_now_add=True)
+    comment = models.TextField(blank=True, null=True)
+    referrer = models.CharField(max_length=255, blank=True, null=True)
```
chosak: What purpose does the `referrer` field serve that the `page` field doesn't? Is it for pages that aren't in Wagtail?

FWIW when this was first being implemented [I thought](https://github.com/cfpb/cfgov-refresh/pull/2418#discussion_r80705807) it might be better to just use the page URL or slug for `page` instead of the foreign key to a Wagtail `Page` object. I would be okay with removing that `page` field altogether and always just storing the referrer. Do we know anything about what the requirements are for storing where feedback comes from on those legacy pages?

cfgov/v1/forms.py
```diff
@@ -254,10 +271,38 @@ def get_query_strings(self):
...
         model = Feedback
         fields = ['is_helpful', 'comment']
 
     def __init__(self, *args, **kwargs):
         super(FeedbackForm, self).__init__(*args, **kwargs)
         self.fields['is_helpful'].required = True
+
+
+class ReferredFeedbackForm(forms.ModelForm):
+    """For feedback modules that need to capture the referring page"""
```
higs4281: wouldn't hurt.

cfgov/v1/models/base.py
higs4281: referrer is intended for a stand-alone feedback page that could be linked to from any number of other pages. in those cases, the product owners want to know which page the user came from.

higs4281:
@chosak hard-coded url values addressed

chosak:
:+1: This looks good to me and works locally. Over the long term it would be nice to make this more flexible (configurable radio button answers) and consider how best to support a wider variety of question combinations. Additionally, because this does store email addresses in the database so we should make sure that all appropriate policies wrt PII are followed.

PR #2536, opened 2016-11-15
Exclude reports from showing up on the Newsroom Landing Page

richaagarwal:
## Overview

This is needed because my previous Filterable refactoring (https://github.com/cfpb/cfgov-refresh/pull/2448) unintentionally** removed this functionality.

- **We need to have some way of documenting or capturing more clearly in the code the very custom functionality of the Activity Log and Newsroom pages.  The fact that the Newsroom should not include in its results anything from research reports is not obvious, and I only found this out by @schaferjh pointing it out. I tried to make it clearer in the code I'm adding here, but there are other things specific to these pages (e.g. only include results from AbstractFilterPages) that isn't immediately obvious to someone looking at the page or looking at the code.  @chosak and I have discussed that it might be better to configure this functionality in the Wagtail editor itself, which would require further refactoring.  I think that's probably the way to go, but we might need to have a BEWD discussion about the best approach.

## Other notes
- Instead of passing around `filter_parent` as I did in my earlier refactor, I opted for passing around `base_query` as that allows us to pass from the `NewsroomLandingPage` definition directly the fact that it needs to exclude research reports' subcategories.  It looks slightly messier in the `get_browsefilterable_posts` function because the logic for filtering by parent page is now redundant (defined there as well as in `FilterableListMixin`) but I think that redundancy is a vestige of the fact that `get_browse_filterable_posts` is defined in the wrong class.
- Not updating the `CHANGELOG` since together with the filterable refactor, this does not change functionality
- I included some minor flake8 clean-up as well as updated relative imports to absolute imports where I saw fit
- Removed validation on topic tags specifically, so that we can pass in whatever in the "View More" URL and not worry about it breaking things

## Testing
- Confirm that no reports show up on Newsroom, [http://localhost:8000/about-us/newsroom/](http://localhost:8000/about-us/newsroom/), but that they do show up on the Activity Log [http://localhost:8000/activity-log/](http://localhost:8000/activity-log/)

## Review
@cfpb/cfgov-backends 

## TODO
- Add a test for `generate_view_more_url` given that I updated it and it does not have any tests (and actually broke it without realizing it earlier - thanks @rosskarchner for catching that before deploying)

### DIFF
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -161,20 +161,16 @@ class CFGOVPage(Page):
         return sorted(author_names, key=lambda x: x.name.split()[-1])
 
     def generate_view_more_url(self, request):
-        from ..forms import FilterableListForm
         activity_log = CFGOVPage.objects.get(slug='activity-log').specific
-        form = FilterableListForm(parent=activity_log, hostname=request.site.hostname)
-        available_tags = [tag[0] for name, tags in form.fields['topics'].choices for tag in tags]
         tags = []
         index = activity_log.form_id()
         for tag in self.tags.slugs():
-            if tag in available_tags:
-                tags.append('filter%s_topics=' % index + urllib.quote_plus(tag))
+            tags.append('filter%s_topics=' % index + urllib.quote_plus(tag))
         tags = '&'.join(tags)
         return get_protected_url({'request': request}, activity_log) + '?' + tags
 
     def related_posts(self, block, hostname):
-        from . import AbstractFilterPage
+        from v1.models.learn_page import AbstractFilterPage
         related = {}
         query = models.Q(('tags__name__in', self.tags.names()))
         search_types = [
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -499,7 +499,6 @@ def rendition_delete(sender, instance, **kwargs):
     instance.file.delete(False)
 
 
-<<<<<<< 891c51e596fe6ee61f11aac5e81092b875df1c99
 # keep encrypted passwords around to ensure that user does not re-use
 # any of the previous 10
 class PasswordHistoryItem(models.Model):
@@ -507,15 +506,6 @@ class PasswordHistoryItem(models.Model):
     created = models.DateTimeField(auto_now_add=True)
     expires_at = models.DateTimeField()  # password becomes invalid at...
     locked_until = models.DateTimeField()  # password cannot be changed until
-=======
-# keep encrypted passwords around to ensure that user does not re-use any of the
-# previous 10
-class PasswordHistoryItem(models.Model):
-    user = models.ForeignKey(User)
-    created = models.DateTimeField(auto_now_add=True)
-    expires_at = models.DateTimeField()   # password becomes invalid at...
-    locked_until = models.DateTimeField()  # password can not be changed until...
->>>>>>> flake8 updates and use urlencode instead
     encrypted_password = models.CharField(_('password'), max_length=128)
 
     class Meta:
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -1,54 +1,43 @@
-import os
-import json
-import urllib
-from itertools import chain
 from collections import OrderedDict
+from itertools import chain
+import json
+import os
+from urllib import urlencode
 
+from django.contrib.auth.models import User
 from django.db import models
 from django.db.models import Q
 from django.db.models.signals import pre_delete
+from django.dispatch import receiver
 from django.http import (
     Http404,
     JsonResponse,
     HttpResponseBadRequest,
     HttpResponse
 )
-from django.template.response import TemplateResponse
-from django.utils.translation import ugettext_lazy as _
-from django.utils import timezone
-from django.dispatch import receiver
-from django.contrib.auth.models import User
 
-from wagtail.wagtailimages.models import (
-    Image, AbstractImage, AbstractRendition
-)
-from wagtail.wagtailadmin.edit_handlers import StreamFieldPanel
-from wagtail.wagtailcore import blocks, hooks
-from wagtail.wagtailcore.blocks.stream_block import StreamValue
-from wagtail.wagtailcore.fields import StreamField
-from wagtail.wagtailcore.models import (
-    Orderable,
-    Page,
-    PageManager,
-    PagePermissionTester,
-    PageQuerySet,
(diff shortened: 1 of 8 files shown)


rosskarchner:
I ended up reverting this in the 4.1 branch-- it was causing errors on some pages (like events)

```
[Tue Nov 15 22:48:20 2016] [error]     {{ related_posts.render(posts, block, half_width) }}
[Tue Nov 15 22:48:20 2016] [error]   File "/srv/cfgov/versions/124/cfgov/jinja2/v1/_includes/molecules/related-posts.html", line 56, in template
[Tue Nov 15 22:48:20 2016] [error]     {{ _view_more() }}
[Tue Nov 15 22:48:20 2016] [error]   File "/srv/cfgov/versions/124/cfgov/jinja2/v1/_includes/molecules/related-posts.html", line 119, in template
[Tue Nov 15 22:48:20 2016] [error]     href="{{ page.generate_view_more_url(request) }}">
[Tue Nov 15 22:48:20 2016] [error]   File "/srv/cfgov/versions/124/cfgov/v1/models/base.py", line 150, in generate_view_more_url
[Tue Nov 15 22:48:20 2016] [error]     form = FilterableListForm(parent=activity_log, hostname=request.site.hostname)
[Tue Nov 15 22:48:20 2016] [error]   File "/srv/cfgov/versions/124/cfgov/v1/forms.py", line 110, in __init__
[Tue Nov 15 22:48:20 2016] [error]     self.base_query = kwargs.pop('base_query')
[Tue Nov 15 22:48:20 2016] [error] KeyError: 'base_query'
```

cfgov/v1/tests/test_filterable_list.py
```diff
@@ -11,11 +11,6 @@ class TestFilterableListMixin(TestCase):
     def setUp(self):
         self.mixin = FilterableListMixin()
         self.factory = RequestFactory()
-
-    @mock.patch('v1.models.learn_page.AbstractFilterPage.objects')
-    def test_base_query_calls_live_shared(self, mock_cfgovmanager):
-        FilterableListForm(parent=None, hostname='hostname').base_query()
-        assert mock_cfgovmanager.live_shared.called
```
richaagarwal: Didn't seem worth it to update this test so I just removed it.

richaagarwal:
Thanks @rosskarchner for attempting to get this in and catching the error! I've fixed it now, but it could stand for a more thorough review.  In theory, this should get in ASAP, but I have a feeling no one (who cares) will notice we're showing reports on the Newsroom page. That said, once it does get approved and merged, can we still do a hotfix & deploy?

rosskarchner:
Yeah, of course-- just open a seperate PR to the 4.1 branch (and test it there as well)

cfgov/v1/models/base.py
```diff
@@ -145,20 +145,16 @@ def alphabetize_authors(self):
...
-        form = FilterableListForm(parent=activity_log, hostname=request.site.hostname)
-        available_tags = [tag[0] for name, tags in form.fields['topics'].choices for tag in tags]
         tags = []
         index = activity_log.form_id()
         for tag in self.tags.slugs():
-            if tag in available_tags:
-                tags.append('filter%s_topics=' % index + urllib.quote_plus(tag))
+            tags.append('filter%s_topics=' % index + urllib.quote_plus(tag))
         tags = '&'.join(tags)
         return get_protected_url({'request': request}, activity_log) + '?' + tags
```
chosak: This code could be made a little bit safer (no manual string combine or use of `&`) and simpler by using `urllib.urlencode`:

```py
>>> from urllib import urlencode
>>> urlencode([('filter%s_topics' % index, tag) for tag in self.tags.slugs()])
'filter0_topics=a&filter0_topics=b&filter0_topics=c&filter0_topics=d'
```

cfgov/v1/models/sublanding_page.py
```diff
@@ -72,13 +76,16 @@ class SublandingPage(CFGOVPage):
...
     def get_browsefilterable_posts(self, request, limit):
-        filter_pages = [p.specific for p in self.get_appropriate_descendants(request.site.hostname)
+        hostname = request.site.hostname
+        filter_pages = [p.specific for p in self.get_appropriate_descendants(hostname)
                         if 'FilterablePage' in p.specific_class.__name__
                         and 'archive' not in p.title.lower()]
         posts_tuple_list = []
         for page in filter_pages:
+            base_query = AbstractFilterPage.objects.live_shared(hostname).filter(CFGOVPage.objects.child_of_q(page))
+            logger.info('Filtering by parent {}'.format(page))
```
chosak: Use `logger.debug` instead? Not sure what conventions are around `info` vs `debug`.

cfgov/v1/models/browse_filterable_page.py
```diff
@@ -5,11 +5,13 @@
...
-from ..feeds import FilterableFeedPageMixin
-from ..util.filterable_list import FilterableListMixin
-from .. import blocks as v1_blocks
+from v1.models.base import CFGOVPage
+from v1.atomic_elements import molecules, organisms
+from v1.feeds import FilterableFeedPageMixin
+from v1.util.filterable_list import FilterableListMixin
+from v1 import blocks as v1_blocks
+from v1.util import ref
+from v1.models.learn_page import AbstractFilterPage
```
chosak: How would you feel about alphabetizing these imports?

cfgov/v1/models/sublanding_page.py
richaagarwal: 👍

cfgov/v1/models/browse_filterable_page.py
richaagarwal: 👍

cfgov/v1/models/base.py
richaagarwal: 👍

cfgov/v1/models/sublanding_page.py
richaagarwal: Actually, I changed it back to `logger.info`, I tried updating the settings in `local.py` so that it would show debug, but then it was showing all this other stuff, and I don't feel like mucking with that in this PR.  Definitely something we should configure better later though.

richaagarwal:
@chosak I've addressed all your comments, feel free to take a look again in the morning. Thanks for all your feedback!

chosak:
:+1: This looks good to me. I agree that this functionality is sufficiently complicated and hard to modify that it'd be good to spend some time on a larger refactor.

PR #2544, opened 2016-11-18
Feedback style fixes

higs4281:
Finishing up styling and message interaction on new feedback modules to serve Owning a Home. 
This PR meshes with work @virginiacc is doing on the owning-a-home repo, which forwards non-wagtail feedback users to a wagtail feedback page. 

## Additions

- Contact_advisory richtext block to get long verbiage out of template.

## Changes

- Typos and style tweaks

## Testing

- Requires having owning-a-home running alongside cfgov-refresh locally. 
- Check out feedback forms on various pages:
  - http://localhost:8000/owning-a-home/
  - http://localhost:8000/owning-a-home/feedback/
  - http://localhost:8000/owning-a-home/help-us-improve/
  - http://localhost:8000/owning-a-home/process/prepare/

## Review

- @sebworks @chosak (block/migration changes) 

## Checklist


sonnakim:
@sebworks @chosak would you guys mind reviewing this PR sometime today? Sorry to rush y'all, we have some dependencies!

chosak:
@sonnakim sure, sorry for the delay.

sonnakim:
@chosak no worries whatsoever - sorry to rush you!

cfgov/jinja2/v1/_includes/blocks/feedback.html
```diff
@@ -11,11 +11,16 @@
...
       class="o-form
              block
              block__bg
              block__border
              block__padded-top">
+{% endif %}
+<div class="form-l">
     <input type="hidden" name="csrfmiddlewaretoken" value="{{ csrf_token }}">
     <input type="hidden" name="form_id" value="{{ form_id }}">
     <input type="hidden" name="referrer" value="{{ request.META.HTTP_REFERER }}">
```
chosak: If you navigate to this page directly, and there's no referrer, then the form element looks like this:

image

To avoid this, can [this line](https://github.com/cfpb/cfgov-refresh/pull/2544/files#diff-0acfcc6db8ff7440bb3c2b303a8e3a4bL21) be changed to something like:

```html
<input type="hidden" name="referrer" value="{{ request.META.get('HTTP_REFERER') or '' }}">
```

cfgov/jinja2/v1/_includes/blocks/feedback.html
```diff
@@ -58,14 +63,17 @@
             </div>
         </div>
     </fieldset>
+    {% else %}
+    <input type="hidden"
+           name="is_helpful"
+           value="{{ request.GET.get('is_helpful', '') }}">
```
chosak: Instead of passing the request query parameter into the page like this, does it make sense to check the input (either through some logic here, or in a helper method) and force a value of `1` or `0`?

cfgov/v1/blocks.py
```diff
@@ -109,7 +109,20 @@ class Feedback(AbstractFormBlock):
...
+    contact_advisory = blocks.RichTextBlock(
+        required=False,
+        default=('If you provide your contact information, '
+                 'we may contact you to clarify your question. '
+                 'This information is collected in accordance with the '
+                 '<a href="/owning-a-home/help-us-improve/'
+                 'paperwork-reduction-act-statement/">'
+                 'Paperwork Reduction Act Statement</a> and the '
+                 '<a href="/owning-a-home/help-us-improve/'
+                 'privacy-act-statement/">Privacy Act Statement</a>.')
```
chosak: It would be nice not to have to hardcode these links here. Since this is a RichTextBlock, can editors internal links to these actual Wagtail pages? Actually, can we just remove this default text entirely and just require content editors to add it? How many pages does this apply to?

cfgov/v1/handlers/blocks/feedback.py
```diff
@@ -59,22 +59,37 @@ def get_response(self, form):
...
-                message = 'Something went wrong. Please try again.'
-                messages.error(self.request, message)
+            msg = 'Something went wrong. Please try again.'
 
-        return {'form': form}
+        if self.request.is_ajax():
+            return JsonResponse({'result': 'fail', 'message': msg})
+        else:
+            messages.error(self.request, msg)
+            return {'form': form}
```
chosak: If you come to the feedback form page with `?is_helpful=1` set, and then POST the form without filling out a message, this logic shows you the page again with an error message (e.g. "You must enter a comment") but then you lose the query string. Is there some way of setting the hidden field on the form directly here?

cfgov/v1/handlers/blocks/feedback.py
```diff
@@ -59,22 +59,37 @@ def get_response(self, form):
...
+                )
+            else:
+                return JsonResponse(
+                    {'result': 'pass',
+                     'heading': '',
+                     'message': 'Thanks for your feedback!'}
+                )
         else:
             messages.success(self.request, message='Thanks for your feedback!')
             return HttpResponseRedirect(self.page.url)
```
chosak: If this used `request.path` instead then you wouldn't run into issues with content redirecting to www, etc.

cfgov/v1/handlers/blocks/feedback.py
chosak: If you come to the feedback page with `?is_helpful=1` set, and then try to submit the form without entering a message, it re-renders the page but now you've lost the query string (because it POSTS back to the same URL). Is there some way of setting the field value directly on the form?

cfgov/v1/handlers/blocks/feedback.py
```diff
@@ -49,7 +49,7 @@ def get_response(self, form):
         if form.is_valid():
             feedback = form.save(commit=False)
             is_helpful = self.request.POST.get('is_helpful', None)
-            if is_helpful is not None:
+            if is_helpful is not None and is_helpful != '':
```
chosak: You can cause a 500 by coming to the page with an invalid query string, e.g. `?is_helpful=text`.

What about using something like this instead:

```py
try:
    feedback.is_helpful = bool(int(is_helpful))
except ValueError:
    pass
```

cfgov/v1/handlers/blocks/feedback.py
```diff
@@ -59,22 +59,37 @@ def get_response(self, form):
 
     def success(self):
         if self.request.is_ajax():
-            return JsonResponse({'result': 'pass'})
+            if get_feedback_type(self.block_value) == 'suggestion':
+                return JsonResponse(
+                    {'result': 'pass',
+                     'heading': 'Thank you!',
+                     'message': "Be sure to also sign up for our email list "
```
chosak: Would it be possible to have this success message be configurable on the block, instead of hardcoding it here? This makes the feedback block OAH-specific.

cfgov/jinja2/v1/_includes/blocks/feedback.html
higs4281: We want None as a possible value, because the model stores it that way (NullBooleanField) to handle feedback modules with no is_helpful question. So I removed the '' default so that it will pass None instead if the query string is not present.

cfgov/v1/blocks.py
higs4281: I had included it as a convenience, to use the cleared language as a starting point that can be overridden at will in the admin, but we can remove the default text and let an editor enter it.

cfgov/v1/handlers/blocks/feedback.py
higs4281: another good idea.

cfgov/v1/handlers/blocks/feedback.py
higs4281: The radio-button suggestion format is indeed OAH-specific, and all other feedback forms should get the default success message. I think this could be part of a radio-button framework discussed earlier that is out of scope for this effort.

cfgov/v1/handlers/blocks/feedback.py
higs4281: Grazie.

cfgov/v1/handlers/blocks/feedback.py
higs4281: the best path may be to handle this error browser-side (which we've discussed with @virginiacc), because the full Django round-trip gets too convoluted -- both the query string and the referrer values pose chicken-and-egg problems. For now I think we can live with losing the `is_helpful` value if someone tries to submit an empty comment. `Is_helpful` is already recorded in analytics on the referring page, and it was considered a nice-to-have on the comment page.

chosak:
:+1: Thanks for addressing my feedback @higs4281. I agree that a more configurable radio button approach would be a good future effort if/when additional feedback forms are needed elsewhere on the site.

PR #2554, opened 2016-11-21
Making the 25 / 75  image a link on the Careers page

sebworks:
Making the 25 / 75  image a link on the Careers page

## Changes

- Added logic in Jinja to make the 25 / 75 image a link, when the url ends in `about-us/careers` and a link is passed.

## Testing

-  Visit the `http:/localhost:8000/about-us/careers/` page and click on the 25 / 75 links.
- You should taken to the correct respective links.

## Review

- @chosak 
- @virginiacc 

## Notes

- This doesn't handle scenarios in which query-string params are passed. 

## Todos

-

## Checklist


cfgov/jinja2/v1/_includes/molecules/info-unit.html
```diff
@@ -49,8 +49,13 @@
          change to a boolean parameter. #}
 <div class="m-info-unit
             {{ 'm-info-unit__' ~ value.modifier if value.modifier else '' }}">
-
     {% if value.image %}
+        {# TODO: This code was added for the sake of A / B testing. #}
+        {% set use_image_link = request.url.endswith( '/about-us/careers' )
```
chosak: This doesn't work if [`APPEND_SLASH`](https://docs.djangoproject.com/en/1.8/ref/settings/#append-slash) is set to `True`, which is how we have it configured (the page is at `/about-us/careers/`). This didn't work locally for me without making that change.

Scotchester:
Instinctually, I'm not a big fan of this approach. It doesn't really scale.

Did you consider adding an option to the ImageText2575Group module in Wagtail to enable image linking?

sebworks:
yeah, we are discussing it now. Just trying to determine if it's worth a migration.

Scotchester:
> yeah, we are discussing it now. Just trying to determine if it's worth a migration.

In my opinion, to keep page-specific logic out of module templates, absolutely yes. But perhaps I am not fully cognizant of the downsides of creating a migration.

cfgov/jinja2/v1/_includes/molecules/info-unit.html
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2554/commits/ac7daa93d602e820edb43584a948b9d28f7b8839

sebworks:
@Scotchester, @chosak, I've updated the code to add the boolean flag to the molecule.

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Fixed broken static assets on Technology and Innovation Fellowship page.
 - Updated the analytics code to send events on form submission.
 - Fixed issue surrounding event venue not displaying on event page.
+- Making the 25 / 75 image a hyperlink on the `/about-us/careers/` page.
```
jimmynotjim: This should get updated to match the latest boolean option code.

jimmynotjim:
I'm with Scott, thanks for making this an editor option rather than section/page specific.

sebworks:
@Scotchester, can you take a look at this again.

cfgov/v1/atomic_elements/molecules.py
```diff
@@ -51,6 +51,7 @@ class ImageText2575(blocks.StructBlock):
     image = atoms.ImageBasic()
     links = blocks.ListBlock(atoms.Hyperlink(), required=False)
     has_rule = blocks.BooleanBlock(required=False)
+    should_link_image = blocks.BooleanBlock(default=False, required=False)
```
Scotchester: This is probably worthy of a more clear label and/or help text.

sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2554/commits/1b253fecb1e55cb4b92f5d2d407fd05149b64e3c

Scotchester:
One final question: could we put the option on the whole group, so you only have to remember to check one box? Or is there a scenario we can envision where we'd want some images in a group to be linked but not all? Perhaps @schaferjh would like to weigh in on that point?

Scotchester:
(I should note that I did just test this and it worked well.)

cfgov/v1/atomic_elements/molecules.py
```diff
@@ -51,6 +51,10 @@ class ImageText2575(blocks.StructBlock):
     image = atoms.ImageBasic()
     links = blocks.ListBlock(atoms.Hyperlink(), required=False)
     has_rule = blocks.BooleanBlock(required=False)
+    should_link_image = blocks.BooleanBlock(default=False,
+                                            required=False,
+                                            help_text='Boolean indicating whether the image should be wrapped in a link tag.'
```
Scotchester: Sorry to keep nitpicking this, but I think it should probably mention where it's getting the link from. How about something like, "Check this to link the image to the URL of the first link in the list, if there is a link." Or, if we move the option to the group, "Check this to link all images to the URL of the first link in their unit's list, if there is a link."

schaferjh:
Re: @Scotchester 
> One final question: could we put the option on the whole group, so you only have to remember to check one box? Or is there a scenario we can envision where we'd want some images in a group to be linked but not all? Perhaps @schaferjh would like to weigh in on that point?

For now, I agree that putting the option on the whole group makes sense. I don't think we'd want to have a group where some images are clickable and some aren't. Once we see how this trial goes, we may want to reevaluate.

sebworks:
@Scotchester, I've made the changes. I've updated the molecule context value in the template, which I don't like, but it should suffice for now.

Scotchester:
Works for me locally! You will need to resolve the CHANGELOG conflict and a new migration conflict.

PR #2557, opened 2016-11-22
Add Wagtail video player module

Scotchester:
Per the discussion in [this thread](https://github.com/cfpb/cfgov-refresh/pull/2321#issuecomment-241115051), we want to add a standalone video player module to Wagtail.

As it stands now, this is the absolute MVP for doing this. All it does is create a bare-bones "wrapper" module in Wagtail, whose template just calls the existing video player macro. This ensures that we're not interfering with anything else that uses the existing video player macro (namely, Event pages and the FCM).

I think this is what we want, because the video player needs to be called from within other modules, like the FCM, so it can't _just_ be a Wagtail module. But I labeled this a WIP because I wasn't sure if we should take the time now to do any other refactoring or issue solving with the video player. (For example, [the close button overlaps the video when used within an FCM](http://www.consumerfinance.gov/prepaid-rule/).)

Please give me your thoughts on if you think this PR is good to go as-is, or should take on other updates.

## Additions

- Wagtail module for placing a video player on the page.

## Testing

1. Pull down this branch
2. `runserver`
3. Create/edit a Learn page and add a Video Player to the content area

## Review

- @cfpb/cfgov-frontends 
- @cfpb/cfgov-backends 
- @schaferjh 
- @jimmynotjim 

## Screenshots

screen shot 2016-11-22 at 17 01 31

## Notes

- The module is currently only made available on Learn pages. Should it be available elsewhere? @schaferjh 

## Todos

- [x] Add the migration
- [x] Minor styling fix: dead space under the video preview image
- [ ] Update the change log

## Checklist


schaferjh:
> The module is currently only made available on Learn pages. Should it be available elsewhere? 

No, for now let's stick with keeping this module on Learn pages.

cfgov/v1/atomic_elements/molecules.py
```diff
@@ -291,4 +291,45 @@ class ContentWithAnchor(blocks.StructBlock):
 
     class Meta:
         icon = 'edit'
-        template = '_includes/molecules/full-width-text-anchor.html'
\ No newline at end of file
+        template = '_includes/molecules/full-width-text-anchor.html'
+
+
+# class VideoPlayer(blocks.RegexBlock):
```
jimmynotjim: Should this commented block be removed?

cfgov/v1/atomic_elements/molecules.py
```diff
@@ -291,4 +291,45 @@ class ContentWithAnchor(blocks.StructBlock):
...
+#
+#     class Meta:
+#         icon = 'media'
+#         template = '_includes/macros/video-player.html'
+
+class VideoPlayer(blocks.StructBlock):
+    video_url = blocks.RegexBlock(
+        label='YouTube URL',
+        required=True,
+        # help_text="You must use the embed URL "
```
jimmynotjim: This would have been nice to have, I didn't know the format till I got the invalid message. Can we make it happen?

jimmynotjim:
Tested locally and the player was easy to add and worked as expected. Only a few comments above.

cfgov/v1/atomic_elements/molecules.py
Scotchester: Sadly, no. If we want regex validation of this field, it doesn't support `help_text` :\ Pretty dumb. I'll clean up the comment.

cfgov/v1/atomic_elements/molecules.py
Scotchester: Whoops, yes. Thanks, I'll clean that up.

sebworks:
👍 , was thinking it might make more sense for it to be an organism due to it's complexity.

Scotchester:
@sebworks Agreed. Will make that change.

cfgov/v1/atomic_elements/molecules.py
Scotchester: Fixed in 2b803c72dfe22a7521f1e789d3bdda8bd0aa4c75

Scotchester:
All comments and todos addressed, with the exception of the CHANGELOG, which I haven't updated because the one in `master` has not been updated to reflect the 4.2 release. What's the process for doing that, @rosskarchner @willbarton?

Scotchester:
@richaagarwal Did you still want to review this? I plan to merge Monday morning.

Scotchester:
Hallelujah!

PR #2558, opened 2016-11-23
Limit Newsroom content to Blog/Newsroom posts

chosak:
This PR corrects an issue where `NewsroomLandingPage` pages (such as [the Newsroom page](http://www.consumerfinance.gov/about-us/newsroom/)) were showing posts that didn't belong there.

This change makes it so that `NewsroomLandingPage` pages will always show all Wagtail pages that inherit from `AbstractFilterablePage` (e.g. `BlogPage`, `LearnPage`, `DocumentDetailPage`, `EventPage`, and `NewsroomPage`) *and* are tagged with at least one category from the Blog or Newsroom sets defined [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/util/ref.py#L51)).

## Additions

- Unit tests for `v1.models.base.CFGOVPageQuerySet` and its `live_shared` functionality.

## Changes

- Corrected Newsroom page behavior.
- Replaced one use of `os.environ.get('DJANGO_STAGING_HOSTNAME')` with `settings.STAGING_HOSTNAME`.

## Testing

- This is most easily tested with a production data dump. Run a local server and navigate to [http://localhost:8000/about-us/newsroom/](http://localhost:8000/about-us/newsroom/). You should notice that all links lead either back to the (filtered) Newsroom page or to the (filtered) Blog index page. All posts should be labeled with either "Blog", "Op-Ed", "Press Release", "Speech", or "Testimony".

## Review

- @schaferjh @richaagarwal @cfpb/cfgov-backends 

## Checklist


schaferjh:
👍

cfgov/v1/models/browse_filterable_page.py
```diff
@@ -68,7 +72,19 @@ class NewsroomLandingPage(BrowseFilterablePage):
...
-        """ Modify the base query for Newsroom to exclude research reports """
-        reports = [x[0] for x in dict(ref.categories)['Research Report']]
-        return AbstractFilterPage.objects.live_shared(hostname).exclude(categories__name__in=reports)
+    @classmethod
+    def eligible_categories(cls):
+        categories = dict(ref.categories)
+        return sorted(itertools.chain(*(
+            dict(categories[category]).keys()
+            for category in ('Blog', 'Newsroom')
+        )))
```
richaagarwal: Don't these categories ('Blog' and 'Newsroom') need to get mapped to their subcategories as is done in `clean_categories` here? https://github.com/cfpb/cfgov-refresh/blob/7c6e15e1eaa5fa61d4e46d0163f5595e21b6c609/cfgov/v1/forms.py#L132

chosak: @richaagarwal Thanks for reviewing asking this, I figured you'd have a better sense of what this logic is supposed to do.

The logic in this PR tries to say: "`NewsroomLandingPage` pages only ever show posts that are tagged with the categories listed under "Blog" and "Newsroom" in the lists at [`v1.util.ref.categories`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/util/ref.py#L51)."

The logic you linked to is slightly different (I think) -- it says: "When someone submits a form, where the choice of topics comes from the list at [`v1.util.ref.page_types`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/util/ref.py#L27), we have to convert the topics from that list to their subcategories in `v1.util.ref.categories`."

For me at least there's some confusion about the difference between "topics" and "categories" and "page types". Why are there separate lists for `page_types` and `categories` at all? Could [the set of category choices in `FilterableListForm`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/forms.py#L118) just use the set of `categories`, not `page_types`?

richaagarwal:
👍  I think we can do something similar for the ActivityLogPage, & then continue to think about how to make this configurable in the editor instead.

PR #2563, opened 2016-11-29
Limit Activity Log content to blog, newsroom, and reports pages

richaagarwal:
See GHE 452

Similar to Andy's work in https://github.com/cfpb/cfgov-refresh/pull/2558.  As of this change, ActivityLog and NewsroomLandingPage are identical except ActivityLog has "Reports" as well.

If possible, this should go in as a hotfix to 4.2.0 and be part of tonight's deploy.

To test: go to http://www.consumerfinance.gov/activity-log/ and [http://localhost:8000/activity-log/](http://localhost:8000/activity-log/) and notice that locally, we are now excluding categories like "Official Guidance" and "Final Rule" which should've never been displayed.

### DIFF
--- a/cfgov/v1/models/sublanding_filterable_page.py
+++ b/cfgov/v1/models/sublanding_filterable_page.py
@@ -1,3 +1,5 @@
+import itertools
+
 from wagtail.wagtailadmin.edit_handlers import StreamFieldPanel
 from wagtail.wagtailcore.fields import StreamField
 from wagtail.wagtailadmin.edit_handlers import TabbedInterface, ObjectList
@@ -7,7 +9,8 @@ from v1.models.base import CFGOVPage
 from v1.models.learn_page import AbstractFilterPage
 from v1 import blocks as v1_blocks
 from v1.atomic_elements import molecules, organisms
-from ..feeds import FilterableFeedPageMixin
+from v1.feeds import FilterableFeedPageMixin
+from v1.util import ref
 from v1.util.filterable_list import FilterableListMixin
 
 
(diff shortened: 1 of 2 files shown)


cfgov/v1/models/sublanding_filterable_page.py
```diff
@@ -46,8 +49,22 @@ class ActivityLogPage(SublandingFilterablePage):
...
 
-    def base_query(self, hostname):
-        return AbstractFilterPage.objects.live_shared(hostname)
+    @classmethod
+    def eligible_categories(cls):
+        categories = dict(ref.categories)
+        return sorted(itertools.chain(*(
+            dict(categories[category]).keys()
+            for category in ('Blog', 'Newsroom', 'Research Report')
+        )))
```
schaferjh: Since I'm not set up to test locally, I'm going to verify the obvious, since I saw there was some convo re this on #2558 -- these "parent" categories map to all the specific subcategories?

rosskarchner:
Once this is in, please open up a new PR that applies these changes to the 4.2 branch

cfgov/v1/models/sublanding_filterable_page.py
chosak: The subcategory mappings are listed [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/util/ref.py#L51).

chosak:
:+1: It'd be nice if we didn't have to duplicate the logic from #2558 but this seems good for the immediate need.

cfgov/v1/models/sublanding_filterable_page.py
schaferjh: Great, thanks!

schaferjh:
👍  LGTM!

richaagarwal:
@chosak - yeah, I thought about that, but I think they *happen* to be similar right now, and that may change pretty soon as these pages change anyway.

PR #2567, opened 2016-11-30
Customize social media share link content 

virginiacc:
As part of the OAH updates, we want to individually customize the content for each of the social media share types on OAH Wagtail pages. This PR adds the ability to specify separate content in the Wagtail social media module for a page's email, LinkedIn, and Twitter links.
       
## Additions

- New fields for email, LinkedIn, and Twitter content

## Changes

- Use new content values in generating links in `social-media.html`

## Review
@schaferjh 
@higs4281 
@cfpb/cfgov-frontends  
@cfpb/cfgov-backends 

## Testing

- Create a page and add a social media block in its sidebar. 
- Add custom content for email, LinkedIn, and Twitter links.
- Preview page and verify that links work as expected.

## Screenshots

screen shot 2016-11-30 at 8 58 40 am

screen shot 2016-11-29 at 9 21 57 pm

screen shot 2016-11-29 at 9 21 17 pm

screen shot 2016-11-29 at 9 20 35 pm

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
* [ ] Reviewers requested with the [Assignee tool](https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/) :arrow_right:

PR #2572, opened 2016-12-02
Tamper-proof URL redirector

rosskarchner:
We had two problems with our external-site redirector:

- in certain circumstances, it could be tricked into allowing (and executing) javascript
- It could be hijacked to create .gov URL's that pointed to arbitrary sites, like [as described here](http://krebsonsecurity.com/2016/03/spammers-abusing-trust-in-us-gov-domains/)

With this PR, we validate that that URL's are valid HTTP or HTTPS URL's, and make sure they are either explicitly whitelisted (the whitelist is stored in django settings) OR cryptographically signed using our [SECRET_KEY](https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-SECRET_KEY).

Links that are hard-coded into templates should use the whitelist approach. Links that are embedded in wagtail content (including social media sharing links) are now automatically signed.

A signed redirect looks like this:

/external-site/?ext_url=http%3A%2F%2Fsettlement-claims.com%2Fautolending%2FEligibilityForm.html&signature=vWnjkpKCAf0eRAaiF0Eq5lEtlxQ

The signature and URL are a pair-- if either of them is altered, then the link is no longer valid. An attacker who wanted to generate valid signatures would need our SECRET_KEY (and if an attacker has our SECRET_KEY, we've probably got bigger problems...)

Per advice from @schaferjh, invalid signatures (or, URL's that aren't whitelisted) will return 404's.

## Additions

- core.utils functions for building signed URL's (and exposed in Jinja2 templates

## Changes

- links generated in Jinja2 code must be wrapped in the signed_redirect function
- the code that had been altering external links to use the redirector, now generates signed links.

## Testing

- Look for links to external content (blog posts about settlements are a good bet)
- try out the social media sharing links
- try tampering with signed URL's and validate that it doesn't work.

## Checklist


schaferjh:
Question re: this action:
> Links that are embedded in wagtail content (including social media sharing links) are now automatically signed.

Is this only happening for non .gov external links entered in the wagtail admin, or all external links? I suppose it's okay for it to happen for all, it just doesn't necessarily _need_ too. (And the gov exclusion really cuts down the number of links that need this treatment.) 

Re: content on invalid redirect page -- what steps has a user taken to end up here? Or would they only see this page as a result of an error on our part?

rosskarchner:
It plugs into the same logic that had been inserting the redirects before, so it does seem to exclude other.gov links, etc (using [this regex](https://github.com/cfpb/cfgov-refresh/pull/2572/files#diff-ac27c5919c46548eada7a7388e726c59R237))

Someone could end up on that page if:

- they've somehow saved or bookmarked, or been sent an invalid or no-longer-valid /external-site/ URL
- or if a search engine has indexed it

rosskarchner:
Now that it's merged, I'll work to incorporate the changes from #2567

rosskarchner:
This is ready to review again

schaferjh:
@rosskarchner Coming back to the content for the error page --

The use case I care about the most is a user being sent to an invalid or no-longer-valid /external-site/ URL. Bookmarking/saving one of these external redirects seems very unlikely, as does (hopefully) the search indexing. Would it be possible to show users the standard 404 error page?

rosskarchner:
@schaferjh Yeah-- we could certainly do that!

rosskarchner:
That solves the FEWD issues too.

schaferjh:
Let's do it, then!

rosskarchner:
@schaferjh that change is in

cfgov/core/tests/test_utils.py
```diff
@@ -51,6 +55,27 @@ def test_django_tag(self):
         self.render_and_check_sha(template)
 
 
+class URLBuildingTest(TestCase):
+    def test_appned_query_args(self):
```
rosskarchner: fixed!

richaagarwal:
👍 

At some point it might be good to apply the whitelisting to *all* URLs, which would require separating out the logic for external URLs. For now this seems fine since someone would be hardcoding those anyway.

richaagarwal:
Oh also please update the CHANGELOG :)

rosskarchner:
changelog update incoming!

cfgov/core/tests/test_forms.py
```diff
@@ -0,0 +1,47 @@
...
+
+from core.utils import sign_url
+from core.forms import ExternalURLForm
+
+
+class TestExternalURLForm(TestCase):
+    def setUp(self):
+        # insulate the test from future whitelist changes
+        self.preserved_whitelist = settings.EXTERNAL_URL_WHITELIST
+        settings.EXTERNAL_URL_WHITELIST = (r'^https:\/\/facebook\.com\/cfpb$',)
```
chosak: Could also use [`@modify_settings`](https://docs.djangoproject.com/en/1.8/topics/testing/tools/#django.test.override_settings) decorator on the test class here, avoiding the need for separate `setUp`/`tearDown`.

rosskarchner: oh, cool-- thanks!

higs4281:
Locally, I run into 404s or another error pages when I try certain external links:

page | app | external link
:--- | :-- | :------------
about-us/blog/ technology-changing-how-people-access- their-financial-records-share-your-story/ | V1 | Facebook and Twitter links in blog text get signatures but return 404
retirement/before-you-claim/ | satellite Django | social links aren't signed and return 404 after trying `http://localhost:8000/external-site/?ext_url=https://twitter.com/cfpb` which results in a call to `/external-site/null/`
owning-a-home/ | V1 | Signed external URL returns 404
owning-a-home/explore-rates/ | satellite non-Django | social links returns the error page saying "You attempted to reach an external link, but something went wrong."

rosskarchner:
Interesting-- thanks @higs4281

rosskarchner:
@higs4281 I can't replicate the first item. Maybe you need to hardcode SECRET_KEY in your .env file?

rosskarchner:
The social links on retirement DO work for me

higs4281:
Got latest code, ran `.frontend.sh` and checked secret key.
- social links on blog page, `/owning-a-home/`, `/about-us/` and their subpages work for me now.
- social links still 404 on `/retirement/before-you-claim/` 

I suspect this is a local issue.

rosskarchner:
Thanks, @higs4281 -- merging!

I'll then spend some quality time reviewing things on build

PR #2573, opened 2016-12-05
Add a CSV export option to the Feedback admin

higs4281:
Owning-a-Home stakeholders need to see and analyze feedback on site
pages that recently got feedback modules. This allows a content editor
to export slices of feedback (by page, date, etc) as a CSV, emailed to
their CFPB account. Only those with admin credentials can get access.

## Additions

- bits to create a new modeladmin action.

## Testing

- If you test locally, the dev settings will pipe email to the console. I tested email with temporary settings that used a dummy email account as email host.

## Screenshots

admin_export

## Notes

- ~~Just realized our test suite excludes admin. Pursuing that.~~


cfgov/v1/admin.py
```diff
@@ -38,12 +38,63 @@ def feedback_page_title(feedback):
 
 @admin.register(Feedback)
 class FeedbackAdmin(admin.ModelAdmin):
+
+    actions = admin.ModelAdmin.actions + ['export_selection_as_csv']
```
chosak: I don't think you need to explicity add this to `admin.ModelAdmin.actions`; you should be able to just do `actions = ['export_selection_as_csv']`. The [Django code](https://github.com/django/django/blob/1.8.15/django/contrib/admin/options.py#L875) automatically pulls in actions from all levels of inheritance.

mthibos:
@higs4281   Awesome!  This looks like it would accomplish the basic purpose.

Question:  How do you select the records for export?  

From the screenshot, it looks like you are limited to what is already available in DJANGO, which includes: (1) choose a page from the sidebar; (2) choose a pre-defined date range from the sidebar; OR (3) search for a string (which lets you choose a family of pages).  

Those three options are sufficient for basic functionality (I can always export more than I need to, and manipulate in excel).  However, if it's not heaps of work, the functionality would be much improved if you could incorporate:  (a) a way to specify a custom date range beyond what is offered in the sidebar; and (b) a way to specify both a custom date range **AND** a custom family of pages (ie, search result) at the same time.

higs4281:
@mthibos I think the admin can make the slices and range selections you're after.

higs4281:
@chosak I vote `aye` to programmatic access!

PR #1970, opened 2016-04-21
Fixing the missing spacing below full-width images

jimmynotjim:
Classname was just incorrectly set.

## Changes

- Fixed the `full-width-media_image` classname

## Testing

- gulp build and create a blog post with images (or trust the screenshot)

## Review

- @KimberlyMunoz 
- @sebworks 
- @anselmbradford 

## Screenshots

__Before__

screen shot 2016-04-21 at 12 48 03 pm

__After__

screen shot 2016-04-21 at 12 47 53 pm

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #1944, opened 2016-04-20
Created initial full-width-media molecule

jimmynotjim:
Blog posts need to have the ability to include full width images, figures, and block quotes. This is only a start, but it accounts for the full width images content editors need at this momemnt. Figures and block quotes should be addressed post-launch.

## Additions

- Added simple default styles for images within a full-width-media molecule 

## Changes

- Added initial media markup to the full-width-text-group

## Testing

- Add this between L:29 and L:30 in  `organisms/full-width-text.html`:
  ```
      <div class="m-full-width-media">
            <img class="m-full-width-media_image"
                 alt="Mortgage Moves: What kind of mortgage will you get? graphic"
                 src="http://files.consumerfinance.gov/f/201604_cfpb_what-kind-of-mortgage-will-you-get-blog.png">
      </div>
```

  and open a doc detail page like http://localhost:8000/data-research/research-reports/2015-office-minority-and-women-inclusion-annual-report-congress/

## Review

- @kurtw 
- @KimberlyMunoz 
- @sebworks 
- @anselmbradford 

## Screenshots

screen shot 2016-04-20 at 3 01 53 pm

## Notes

- This is a terrible example, but until wagtail blog posts are completed, it's the best I was able to cobble together.

## Todos

- Build the backend doc detail blog posts.

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #2581, opened 2016-12-06
Disable image renditions for GIFs

chosak:
We've experienced problems uploading animated GIFs as Wagtail images. These seem to be caused by problems with the underlying image processing libraries used as part of Wagtail's image renditions logic. This PR bypasses the Wagtail image rendition process for all GIF images, and forces use of the original image file in all template references. 

Normally, anytime the Wagtail [`{% image %}` tag](http://docs.wagtail.io/en/latest/topics/images.html) is used in a template (either Wagtail built-in or our own), some image processing is performed on the image being embedded. This is true even if the "original" format of the image is referred to, i.e. `{% image 'original' %}` (this invokes a file-open-read-write-in-same-format operation). This process can fail on large animated GIF files.

The Wagtail admin makes use of a thumbnail image rendition `{% image max-165x165 %}` in the admin images browser and when selecting an image in a page editor. It also uses [certain predefined](http://docs.wagtail.io/en/latest/topics/images.html#images-embedded-in-rich-text) renditions when including images into a Rich Text block (for example, specifying "full-width" when selecting an image actually sets its width to 800 pixels).

This PR bypasses any actual image processing when GIF files are embedded (technically: image files ending in `.gif`) and instead refers to the original image file while keeping appropriate `width` and `height` parameters in the `` tag. This keeps the appearance of default behavior, although on page views a larger (non-optimized for display size) image is being downloaded.

## Additions

- New `add_images` management command for programmatic adding of images (I created this for testing and thought it might be useful enough to keep around).

## Changes

- Pulled image-related classes out of `v1.models.base` in favor of `v1.models.images` for readability.
- Added mock rendition logic so that GIF files embedded in templates always load the source image.

## Testing

- Create a new local, empty database, and run your local server.
- Navigate to the [image upload page](http://localhost:8000/admin/images/multiple/add/) and upload a large, non-GIF image. After it's been uploaded, navigate back to the [image browse page](http://localhost:8000/admin/images/) and inspect the image element. Note that a thumbnail filter has been applied and that a resized image is being referred to:

image

- Next, repeat the same procedure with a large GIF image. Note that the original image is what gets embedded in the page, with appropriately-set `width` and `height` attributes:

image

- Test the Rich Text block functionality by creating a new BrowsePage and adding a FullWidthText element with a Context block. Within the Context block, select the image icon and select the large PNG file, entering it as "left aligned". Then preview the page and inspect the image element. Note that a `width-500` filter has been applied, and resized image is being embedded in the page:

image

- Repeat the previous step, but with a large GIF image. Note that the original image is embedded, but with `width` and `height` attributes set to 500px:

image

## Notes

- One discussion-worthy choice that I made here is to explicitly fail page renders if a GIF image is passed to a Wagtail revision function that involves actual image processing beyond just a resize. Examples of these would be `fill` (which resizes an image by padding it to fill a certain size) or `format` which converts an image to a different file format. Another option would be to just pass the original image through (thus ignoring the revision request) and maybe log a warning message; I thought an explicit `RuntimeError` (which would cause a 500 in practice) was a stricter defense of this. @schaferjh @willbarton any thoughts?

## Todos

- Should the site documentation be updated with a mention of this change? If so, where in the documentation tree would be appropriate?

## Checklist


schaferjh:
Still absorbing this, but: I keep wondering why, in our recent images+Wagtail convos, the media module is never mentioned. Is this deliberate or not?

**Story time interlude:** While creating our initial Wagtail templates,  the team created a "media" module within Full Width Text (I believe the name is `full-width-media`) and directed admin users not to embed images via the rich text block because there was some kind of issue (formatting or Wagtail-y; memory fails me) around images directly uploaded into the rich text block. @jimmynotjim I see your name on https://github.com/cfpb/cfgov-refresh/pull/1944 and https://github.com/cfpb/cfgov-refresh/pull/1970, do you remember the reasoning?

**Relevant to this PR:** I assume but want to be sure -- the process created by this PR still applies when using the Full Width Text/Media module?

**Tangent:** @chosak given your recent immersion in all things images + Wagtail, is it your opinion that the media module is a formatting thing only and that there's no code concern around embedding images via the rich text block?

chosak:
@schaferjh if you [look at the template](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/full-width-text.html#L29) for the "media" option in Full Width Text, it uses the Wagtail image tag like this:

```html
<img class="m-full-width-media_image" src="{{ media.url }}" alt="{{ media.alt }}">
```

With this PR, using `'original'` like that will result in the actual original file being embedded for GIFs only (for other file types, it'll still invoke a load-and-resave operation). I didn't call out the media module specifically because it uses the same Wagtail image tag as all of the other templates.

In response to your question about embedding images via the RTB, @Scotchester might have a more thorough understanding about the issues (I think there are some if you make the window too narrow).

schaferjh:
> I didn't call out the media module specifically because it uses the same Wagtail image tag as all of the other templates.

👍

Scotchester:
👍 Working great for me. I might do a more thorough perusal of the code tomorrow morning.

> While creating our initial Wagtail templates, the team created a "media" module within Full Width Text (I believe the name is full-width-media) and directed admin users not to embed images via the rich text block because there was some kind of issue (formatting or Wagtail-y; memory fails me) around images directly uploaded into the rich text block. @jimmynotjim I see your name on #1944 and #1970, do you remember the reasoning?

@schaferjh There are a couple issues with using images in rich text fields:

1. Wagtail's HTML output includes the `width` and `height` attributes of the `img` element. This causes unsightly stretching when the available width is less than the 800px or 500px width (or actual size, if smaller) that Wagtail specifies for images in rich text fields. The code @chosak is proposing here actually makes it pretty easy to fix this issue. Since we're already overriding the base `get_rendition`, we can take it a step further and prevent those attributes from being output with specific values.
2. We haven't implemented the `.left` or `.right` classes in CSS, so the left and right alignment functions don't actually do anything to move images to the left or right. This would just be a matter of creating those classes with simple floated image styles applied.

> [I]s it your opinion that the media module is a formatting thing only and that there's no code concern around embedding images via the rich text block?

If those two things were fixed, there would be no issues at all with using images in a rich text field.

chosak:
@Scotchester:

>  Since we're already overriding the base `get_rendition`, we can take it a step further and prevent those attributes from being output with specific values.

Do you mean never output `width` or `height` attributes with image tags? This would break a lot of functionality that this PR otherwise preserves (like image thumbnails). If instead you're suggesting only excluding those attribute when, e.g. `full-width` is used, that might be tricky based on the way [this code](https://github.com/torchbox/wagtail/blob/v1.6.3/wagtail/wagtailimages/formats.py#L92) defines those shortcuts. Only the last part gets passed into the `get_rendition` function that gets overridden, so you'd never be able to disambiguate a "real" width=800 request from the "full width" one.

schaferjh:
Gracias for the explanation/reminder @Scotchester! And now I remember the image stretching thing.

Scotchester:
@chosak Yeah, you're right, it's more complicated than I described. I'm not suggesting trying to do anything else in this PR, but I'd like to have these issues on the backlog to work through in the near future. I'd be happy to contribute to that effort.

The stupid thumbnails is really the big problem, because it's probably the only place where we want to constrain the height of images as well as the width.

chosak:
Thanks @Scotchester. I'm going to fix CHANGELOG conflicts and merge this in.

higs4281:
👍 works for me locally.

Nit in test instructions: 

> Test the Rich Text block functionality by creating a new BrowsePage and adding a FullWidthText element with a Context block.

I think this should be "Content block"

Question: `add_images` is a great addition, but I figure how to get loaded images available in wagtail. They get loaded to the `v1_cfgovimage` table; what would be their path to `wagtailimages_image`?

chosak:
Thanks for the review @higs4281. I believe that since we use a custom image class, images that we upload don't ever get added to the `wagtailimages_image` table (because `CFGOVImage` replaces `Image` completely). It's a little different than in the pages case, where `CFGOVPage` inherits from `Page` using [multi-table inheritance](https://docs.djangoproject.com/en/1.8/topics/db/models/#multi-table-inheritance), so both `wagtailcore_page` and `v1_cfgovpage` exist and are used in the DB (with the latter containing foreign keys back to the former.

jimmynotjim:
Seriously late to this, but another reason for the full-width-media was that the full-width-text was restricted for leading (line length) based on that container and Justin wanted to keep the images actually full width. Any image embeded inside the full-width-text was restricted to the leading settings and messed up the layout. I believe that's fixed now so that it's just limited to the text elems, right @Scotchester? If that, plus the height/width, and the left/right issues were resolved, I believe we could drop the full-width-media module.

PR #2596, opened 2016-12-12
Add missing uniqueness constraint to CFGOVRendition

chosak:
Our use of a custom image rendition class `v1.models.images.CFGOVRendition` neglected to include the uniqueness constraint present on the reference rendition model [here](https://github.com/wagtail/wagtail/blob/v1.6.3/wagtail/wagtailimages/models.py#L516) and in the documentation [here](http://docs.wagtail.io/en/v1.7/advanced_topics/images/custom_image_model.html?highlight=unique_together#custom-image-models):

```py
class Meta:
    unique_together = (
        ('image', 'filter', 'focal_point_key'),
    )
```

This lack of constraint may have enabled the creation of duplicate rendition objects which caused some 500 errors when browsing images today: the rendition creation code [uses](https://github.com/wagtail/wagtail/blob/v1.6.3/wagtail/wagtailimages/models.py#L291) Django `get_or_create`, about which [the docs say](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#django.db.models.query.QuerySet.get_or_create):

> This method is atomic assuming correct usage, correct database configuration, and correct behavior of the underlying database. However, if uniqueness is not enforced at the database level for the kwargs used in a get_or_create call (see unique or unique_together), this method is prone to a race-condition which can result in multiple rows with the same parameters being inserted simultaneously.

Without that `unique_together` constraint, we may have been hitting that race condition.

## Additions

- New `unique_together` constraint on `CFGOVRendition`, as recommended by documentation.

## Testing

- This is a little bit difficult to test interactively, but to isolate a new unit test that checks this logic you can run:
 ```sh
$ tox -e fast v1.tests.models.test_images.CFGOVRenditionTest
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 ### Fixed
 - Fixed bug stopping videos in HTTPS pages.
 - bug that wasn't signing links already coded to /external-site
+- Added missing uniqueness constraint on CFGOVRendition.
 
 
 ## 4.3.1
--- a/cfgov/v1/migrations/0036_cfgovrendition_uniqueness.py
+++ b/cfgov/v1/migrations/0036_cfgovrendition_uniqueness.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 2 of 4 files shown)


richaagarwal:
Looks great, thanks for digging into this!

PR #2604, opened 2016-12-14
Spike - Disable deletion in Wagtail UI

richaagarwal:
## Overview
Per GHE 468, this is a MVP/proposal for dealing with deletions until we come up with a longer-term strategy with how we want to approach permissions & deletions, given Wagtail doesn't support configuring this at the moment. 

- I only disabled this in the UI (both Django admin and Wagtail admins); you can still trigger a delete by adding `/delete` to a page.
- *If* someone does delete a page, which I think we will still continue to do by request, I added a simple logging statement for this so that we know who did it. 

## Next steps
- Next steps are outlined in GHE 471, which includes overriding the `/delete` POST in Wagtail and adding a django management command (I think to be done outside this PR)


cfgov/v1/wagtail_hooks.py
```diff
@@ -43,6 +43,17 @@ def share_the_page(request, page):
         flush_akamai(page)
 
 
+@hooks.register('after_delete_page')
+def log_page_deletion(request, page):
+    logger.warning(
```
willbarton: This looks good, but I'm not sure it'll be [caught wit our production logging settings](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/production.py#L5).

cfgov/v1/wagtail_hooks.py
```diff
@@ -43,6 +43,17 @@ def share_the_page(request, page):
...
+        'User {user} with ID {user_id} deleted page {title} with ID {page_id}'.format(
```
chosak: This will raise a `UnicodeEncodeError` if the title or username have any unicode characters in them. To reproduce locally, create a page named "Unicodë", then go to that page's `/delete` page and try deleting. I believe a fix is to just use a string like `u'User {user} with...'` (note the 'u' prefix).

richaagarwal:
@chosak Good point about the django admin.  I'll see what it takes to disable it there as well. The `after_delete_page` Wagtail hook doesn't apply there, either, so it doesn't get logged :(

cfgov/v1/wagtail_hooks.py
richaagarwal: Debug or higher would go to the console, and this is a warning, so it should end up in our logs... we save the console output somewhere, right?

schaferjh:
I also like this as an immediate fix. The django admin might be a different beast -- it is currently mostly used to maintain AskCFPB and I do not know about the permissions there. It may be that we should tackle that separately from this -- but I leave it up to your investigation @richaagarwal.

richaagarwal:
@schaferjh Do you know how it's used for AskCFPB?  I was going to ask the team if there's a reason to keep "Pages" around at all in the django admin.  It's not hard to disable delete specifically there, either, just a few lines of code in `v1/admin.py`.

schaferjh:
@richaagarwal I do not know the specifics of how it's used there, but before making any changes we should definitely find out more. Our team is not actively involved in the Ask editing process due to legacy stuff, so a good place to start is working with @sarahsimpson09 to reach out to CE and find out how/if they're using delete in their process.

higs4281:
In the Django admin, we have more options. We can choose not to display tables or rows, and the fine-grained permissions (such as permission to delete from a specific table) are honored there.

cfgov/v1/wagtail_hooks.py
```diff
@@ -43,6 +43,17 @@ def share_the_page(request, page):
...
 
+@hooks.register('after_delete_page')
+def log_page_deletion(request, page):
+    logger.warning(
+        u'User {user} with ID {user_id} deleted page {title} with ID {page_id}'.format(
+            user=request.user,
+            user_id=request.user.id,
+            title=page.title,
+            page_id=page.id)
+        )
```
chosak: Thanks, looks great.

cfgov/v1/wagtail_hooks.py
```diff
@@ -46,12 +46,14 @@ def share_the_page(request, page):
...
 def log_page_deletion(request, page):
     logger.warning(
-        u'User {user} with ID {user_id} deleted page {title} with ID {page_id}'.format(
+        u'User {user} with ID {user_id} deleted page {title} with ID {page_id} at URL {url}'.format(
             user=request.user,
             user_id=request.user.id,
             title=page.title,
-            page_id=page.id)
+            page_id=page.id,
+            url=page.url,
```
higs4281: should we capture `page.url_path` instead, to cover the deletion of archived pages, which don't have urls?

richaagarwal: Ah, good point @higs4281. Will update

PR #2532, opened 2016-11-15
Convert Bureau Structure to Wagtail

sebworks:
Convert Bureau Structure to Wagtail

## Additions

- Added CSS file to customize Wagtail Admin for the Bureau Structure.

## Changes

- Modified Jinja, Less, and python files to migrate the bureau structure page from a Sheer page to Wagtail. I left the current Sheer Page in place and just updated the necessary associated files to make it atomic.

## Testing
- Run `gulp build`
- Add a Browse page via Wagtail.
- Add a Bureau Structure block to the page
- Verify that the page previews as expected. 

## Review

- @cfpb/cfgov-backends 
- @Scotchester 

## Notes

- It might make sense to use snippets at some point but it would have to more flexible then just injecting that snippet into a 25 / 75 organism

- The CSS modifications to the Wagtail admin are experimental and will probably need to be adjusted.

## Todos

-

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #2605, opened 2016-12-14
Field blocks with placeholder text

chosak:
This PR is meant to provide a possible alternative to the way block placeholder text was implemented in #2532, specifically [this implementation](https://github.com/cfpb/cfgov-refresh/pull/2532/files#diff-cafe2fcbfc87018e811d6d93daa959f4R120) which requires some duplication of Wagtail code.

Instead, this PR creates a new base class `v1.blocks.PlaceholderFieldBlock` and a specific implementation `PlaceholderCharBlock` which uses HTML substitution post-Wagtail render to apply the placeholder. This is still less than ideal, but until Wagtail supports a way to do this (see [wagtail#2328](https://github.com/wagtail/wagtail/issues/2328)) there might be no better way.

## Additions

- New `v1.blocks.PlaceholderFieldBlock` and `PlaceholderCharBlock` to support field blocks with placeholder text.

## Testing

This PR only creates the block type but doesn't modify any atomic elements or page types to include it. For testing, you can apply this patch:

```diff
diff --git a/cfgov/v1/atomic_elements/organisms.py b/cfgov/v1/atomic_elements/organisms.py
index 514e872..e4b8305 100644
--- a/cfgov/v1/atomic_elements/organisms.py
+++ b/cfgov/v1/atomic_elements/organisms.py
@@ -16,6 +16,7 @@ from wagtail.wagtailsnippets.blocks import SnippetChooserBlock
 from . import atoms, molecules
 from ..util import ref
 from ..models.snippets import Contact as ContactSnippetClass
+from v1.blocks import PlaceholderCharBlock
 
 from jinja2 import Markup
 
@@ -30,7 +31,8 @@ class Well(blocks.StructBlock):
 
 
 class ImageText5050Group(blocks.StructBlock):
-    heading = blocks.CharBlock(icon='title', required=False)
+    heading = PlaceholderCharBlock(icon='title', required=False,
+                                   placeholder='Some "placeholder" text')
 
     sharing = blocks.StructBlock([
         ('shareable', blocks.BooleanBlock(label='Include sharing links?',
```

Run a local server and create a BrowsePage, then add an "Image text 50 50 group"

## Screenshots

Here's what the `PlaceholderCharBlock` looks like. Note that quotation marks are properly supported:
image

Here's the HTML source code that gets generated:
image

## Notes

- @sebworks I'm open to feedback that your initial approach might be more straightforward and less hacky (no reliance on HTML parsing). Let me know what you think.

## Checklist


cfgov/v1/blocks.py
```diff
@@ -117,3 +119,33 @@ class Meta:
...
+    def replace_placeholder(html, placeholder):
+        soup = BeautifulSoup(html, 'html.parser')
+        inputs = soup.findAll('input')
+
+        if 1 != len(inputs):
+            raise ValueError('block must contain a single input tag')
+
+        inputs[0]['placeholder'] = placeholder
+
+        return SafeText(soup)
```
chosak: One additional note here: unfortunately, loading into BS and then re-writing as a string does have the potential to introduce minor whitespace differences between how Wagtail typically renders blocks and how these new ones would do it (no space before the closing `<input />`, for example). I didn't think this was worth the pain of implementing some alternative string replacement using regex that'd still be robust enough to e.g. quotation marks in the placeholder text.

sebworks:
Looks great

cfgov/v1/blocks.py
Scotchester: I don't think this is a major issue. That lack of space won't affect how the thing renders, so no biggie.

chosak:
@Scotchester thanks for the review. It'd be difficult short an upstream PR to Wagtail, because right now they hardcode the placeholder text to be equivalent to the label.

Scotchester:
What a poor decision.

PR #2607, opened 2016-12-16
Fix page status string bug for live=True, shared=False

chosak:
This PR fixes a bug with how pages appear in the Wagtail explorer when they have `live=True` but `shared=False`. This state doesn't normally occur through typical site usage but can happen as part of unit tests or if pages are added through a programmatic import process.

Pages with this state currently appear incorrectly in the explorer:
image
Note the use of solid type and status box, indicating that the page is live, even though it says "DRAFT". The link is also clickable to the live page URL.

This PR also makes a modification to the custom implementation of `CFGOVPage.status_string` to avoid a DB query for the page being looked up (since the object must already be in memory). This should speed up the Wagtail explorer a bit by avoiding a DB query for every single page being displayed on each screen.

## Changes

- Fix so that pages with `live=True`, `shared=False` properly appear as "live" in the Wagtail explorer.

## Testing

- Load a production dump and search for "Story" to see some examples of pages in this state.

## Screenshots

Pages now appear properly like this:
image

## Checklist


cfgov/v1/models/base.py
```diff
@@ -301,15 +301,14 @@ def serve_post(self, request, *args, **kwargs):
...
+        elif self.approved_schedule:
             return _("scheduled")
-        elif page.live and page.shared:
-            if page.has_unpublished_changes:
-                if page.has_unshared_changes:
-                    for revision in page.revisions.order_by('-created_at', '-id'):
+        elif self.live and self.shared:
+            if self.has_unpublished_changes:
+                if self.has_unshared_changes:
+                    for revision in self.revisions.order_by('-created_at', '-id'):
```
richaagarwal: Not related to your PR, but is it possible to get rid of this for loop and always `return _('live + (shared + draft)')` if there are unshared changes?

richaagarwal:
Looks great!

cfgov/v1/models/base.py
chosak: @richaagarwal that might be possible, I'm a bit hesitant to mess with this because I don't quite understand the way that shared state and `self.has_unshared_changes` interact.

PR #2609, opened 2016-12-16
Upgrade to Wagtail 1.7

richaagarwal:
See http://docs.wagtail.io/en/v1.8/releases/1.7.html

- The migrations here are part of their upgrade notes.
- I tested that uploading images still works locally and all tests pass. 
- There was an issue with upgrading to 1.8 so we'll do 1.7 for now


higs4281:
looks as if the signature of the operation.run() method changed:

wagtail/wagtailimages/image_operations.py

1.6:
```
    def run(self, willow, image):
```

1.7:
```
    def run(self, willow, image, env):
```

chosak:
Thanks @higs4281, I think my patch in #2581 needs to be updated to match the new signature. I'll push a commit with that change.

higs4281:
@chosak  Since `env` isn't being used in that model method, that seems right.

PR #2610, opened 2016-12-16
Removed layout.less enhancements that have been moved to Capital Fram…

anselmbradford:
Layout.less enhancements have been moved to https://github.com/cfpb/capital-framework/blob/canary/src/cf-layout/src/cf-layout.less#L133-L139

## Removals

- Removes layout.less enhancements that have been moved to Capital Fram…

## Testing

- `gulp build` should run as expected.
- Not sure where this is used, but after discussion with @jimmynotjim we didn't think the `!important` margin-top was needed, but keep an eye out for that.


contolini:
> Visual regression testing would be lovely for situations like this one.

mmmmmhhhhhhmmmmmmmmm

PR #2603, opened 2016-12-14
Django jobs cleanup

chosak:
This PR cleans up some of the careers-related models, views, and templates that have been migrated into Wagtail.

## Removals

- Removes all Django templates under `cfgov/jinja2/v1/about-us/careers/`.
- Removes Django models `jobmanager.Job` and `jobmanager.JobApplicantType`.
- Removes almost all Django views under `jobmanager.views`.
- Deprecates use of `DJANGO_CAREERS` feature flag to switch between Django and Wagtail careers pages.
- Removed deprecated `scripts.migrate_job_pages` script and its associated tests.

## Changes

- Simplifies some models (`jobmanager.ApplicantType`, `Grade`, `JobCategory`, and `Location`) by removing fields that are no longer used (`slug` and `active`).

## Testing

Load a local server using a recent production data dump, to ensure you have relevant jobs data:

```sh
$ ./drop-db.sh
$ ./create-mysql-db.sh
$ ./refresh-data.sh /path/to/dump.sql
```

Then, apply the new migrations from this PR:

```sh
$ python cfgov/manage.py migrate
```

Now run your local server and browse around the admin and the careers pages to verify that things still work correctly.

## Checklist

* [X] Changes are limited to a single goal (no scope creep)
* [X] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [X] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [X] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
* [X] Reviewers requested with the [Assignee tool](https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/) :arrow_right:

PR #2612, opened 2016-12-19
Remove duplicate job models from Django admin

chosak:
The Django jobs cleanup in #2603 left behind two models that were displayed in both the Django admin and the Wagtail admin, `jobmanager.JobCategory` (division) and `jobmanager.ApplicantType`. To avoid confusion, this PR removes these models from the Django admin, where they currently appear like this:

image

and instead leave them only in the Wagtail admin, where they appear like this:

image

## Removals

- Remove duplicate display of `jobmanager.ApplicantType` and `jobmanager.JobCategory` in the Django admin.

## Changes

- Modified `jobmanager.wagtail_hooks.JobCategoryModelAdmin` to use the TinyMCE library to maintain HTML editability of `JobCategory` blurbs.

## Testing

- Run a local server, and using the main Wagtail menu, select Job Listings, Divisions to create or edit a division. Note ability to enter rich text content. Similarly, selected Job Listings, Applicant types to create or edit an applicant type.

## Screenshots

image

## Notes

- The complexity of changes in `jobmanager.wagtail_hooks` is because Wagtail doesn't provide a simple way to override a single form widget in a model's admin view.
- I didn't update the CHANGELOG here because the general "Removed deprecated Django careers-related models, views, and templates." is already in there from #2603.

## Checklist


richaagarwal:
Could you elaborate on the need for this: Modified jobmanager.wagtail_hooks.JobCategoryModelAdmin to use the TinyMCE library to maintain HTML editability of JobCategory blurbs.

(Mostly my own curiosity -- I'm not sure how these pages were being managed before, but I guess I'm surprised the default Wagtail functionality does not provide sufficient editing capabilities).

chosak:
@richaagarwal the blurb field can contain HTML, and TinyMCE is just a library for doing simple HTML editing. That functionality has been around for a while so I just migrated it over to the Wagtail admin component. I'm not sure how easy it would be to, say, use a RichTextBlock instead.

PR #2616, opened 2016-12-20
Use the base_query to for FilterableListForm page_ids

willbarton:
This PR fixes a problem with filterable list forms where the topics would include all possible topics, not just the topics that apply to the pages being filtered.

We may want to hotfix 4.3 with this once it's merged.

## Changes

- Change the filterable list `page_ids` to come from the constructed `base_query` rather than all `CFGovPage` objects.

## Testing

- Visit a BrowseFilterablePage and attempt to search for a topic that doesn't belong to any of the subpages and notice topics that have no results appear. Apply this fix and attempt the same and notice that topics with no results no longer appear.

## Screenshots

Before, on http://www.consumerfinance.gov/policy-compliance/enforcement/actions/:
image

After:
image

## Checklist

### DIFF
--- a/cfgov/v1/forms.py
+++ b/cfgov/v1/forms.py
@@ -7,8 +7,7 @@ from django.forms.utils import ErrorList
 from django.forms import widgets
 from taggit.models import Tag
 
-from .models.base import CFGOVPage, Feedback
-from .models.learn_page import AbstractFilterPage
+from .models.base import Feedback
 from v1.util.categories import clean_categories
 from v1.util import ref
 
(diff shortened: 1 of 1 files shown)


cfgov/v1/forms.py
```diff
@@ -126,7 +126,7 @@ def __init__(self, *args, **kwargs):
         self.hostname = kwargs.pop('hostname')
         self.base_query = kwargs.pop('base_query')
         super(FilterableListForm, self).__init__(*args, **kwargs)
-        page_ids = CFGOVPage.objects.live_shared(self.hostname).values_list('id', flat=True)
+        page_ids = self.base_query.live_shared(self.hostname).values_list('id', flat=True)
```
chosak: Should this be using the `self.base_query()` method instead of the `self.base_query` member? Why do both of these exist?

richaagarwal: Oh, yeah, I think that needs to be cleaned up. At some point it was defined in that function but now it's always passed in.

richaagarwal:
I'll review this locally -- I feel like there may be some unintended things, but I'll doublecheck.

higs4281:
👍 working well for me!

PR #2620, opened 2016-12-22
Fix unit test so it doesn't try to make request to Akamai

chosak:
The unit test `v1.tests.wagtail_hooks.TestFlushAkamai. test_should_flush_gets_called_when_trying_to_flush` actually tries to make an HTTP request to the server defined in environment variable `AKAMAI_PURGE_URL`, which is set by default to [the actual Akamai URL](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/base.py#L513). This doesn't actually purge Akamai (because the other settings aren't defined), but does try to make a request. This fails the test when that URL isn't accessible, and is causing some other PR Travis tests to fail.

This PR modifies that test to skip the actual call to Akamai and only test that `should_flush` gets called appropriately.

## Changes

- Fix to unit test to avoid making HTTP request to Akamai server.

## Testing

- All unit tests should pass.

## Checklist

### DIFF
--- a/cfgov/v1/tests/test_wagtail_hooks.py
+++ b/cfgov/v1/tests/test_wagtail_hooks.py
@@ -290,15 +290,13 @@ class TestFlushAkamai(TestCase):
         with self.settings(ENABLE_AKAMAI_CACHE_PURGE=True):
             assert should_flush(self.page)
 
-    @mock.patch('v1.wagtail_hooks.should_flush')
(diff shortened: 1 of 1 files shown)


chosak:
@higs4281 thanks, I *think* the test might only fail if external connections are blocked. Maybe Travis made a change recently so you can't run externally connecting scripts from their infrastructure?

PR #2621, opened 2016-12-22
Render arbitrary HTML from a specified source in Browse Pages

richaagarwal:
## Overview
- This is a basic MVP for rendering the charts HTML into Wagtail.  

## Screenshots

- In Wagtail:
screen shot 2016-12-22 at 2 45 44 pm
- How it renders:
screen shot 2016-12-22 at 2 42 24 pm

## TODO
- We will want to update the regex to accept only files from our s3 domain once we have that set up
- As part of the Jenkins job that updates these graph files on s3, it should also flush Akamai 

## Other considerations 
- At some point we may want to store the HTML in the database but I'm of the opinion that our Akamai caching is sufficient
- Decide on best way to handle the whole page if the graph URL does not return 200


cfarm:
:man_dancing: wahhhhhh this rules

cfgov/v1/templatetags/share.py
```diff
@@ -42,3 +42,8 @@ def v1page_permissions(context, page):
         )
 
     return context['user_page_permissions'].for_page(page)
+
+@register.filter
```
richaagarwal: I'd like for this to be somewhere else -- either in the block definition or its own templatetags file. It's also possible this is a filter that already exists? In any case, just wanted to point out that this is temporary

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -560,3 +560,21 @@ class VideoPlayer(blocks.StructBlock):
...
         icon = 'media'
         template = '_includes/organisms/video-player.html'
+
+
+class HTMLBlock(blocks.StructBlock):
+    html_url = blocks.RegexBlock(
+        label='Source URL',
+        default='',
+        required=True,
+        regex=r'^https://.+$',
```
richaagarwal: the regex for this can be updated once we know where the files will live and their subdomain.

sebworks:
This is good but I don't like it because the request process has to wait for a return. I would rather handle it with JS.

richaagarwal:
@sebworks One option @contolini and I were discussing is storing the raw html and then running a cron job (maybe nightly) that updates it from the source. The data actually will only get updated monthly so someone could also just run the job as it gets updated. 

Do you think it'd be better with JS to remove the risk of the page loading but no chart loaded?  Or for some other reason?

sebworks:
@richaagarwal, the cron job sounds like a great ideal.

cfarm:
Works as advertised 👍

contolini:
> @sebworks One option @contolini and I were discussing is storing the raw html and then running a cron job (maybe nightly) that updates it from the source. The data actually will only get updated monthly so someone could also just run the job as it gets updated.

Yeah it makes sense to cache it in the DB and periodically check for updates in the background.

contolini:
> This is good but I don't like it because the request thread has to wait. I would rather handle it with JS.

We *could* XHR it and then we wouldn't have to rely on a cron job. The downside is that charts will take a few seconds to load for users.

sebworks:
> Edit: @cfarm and @hillaryj have to write a jenkins job to periodically update S3 with the charts' code. We might as well use that same job to update wagtail. So I vote for storing the raw html in wagtail.

👍

cfarm:
Yup agreed w/Chris

chosak:
A bit late to this, but does the built-in Wagtail [EmbedBlock](http://docs.wagtail.io/en/v1.7/topics/streamfield.html#embedblock) already support the use case of rendering arbitrary HTML? See their [template](https://github.com/wagtail/wagtail/blob/v1.7/wagtail/wagtailembeds/templates/wagtailembeds/embed_frontend.html) and [model code](https://github.com/wagtail/wagtail/blob/v1.7/wagtail/wagtailembeds/models.py#L16); this might be overkill for our needs but it does some nice things like cache the HTML.

cfgov/v1/templatetags/share.py
```diff
@@ -42,3 +42,8 @@ def v1page_permissions(context, page):
...
+def get_raw_html(url):
+    r = requests.get(url)
+    return r.content
```
chosak: This will return the content of the request even if it fails, e.g. a 500 error message. It might be worth checking the status code or using [`r.raise_for_status`](http://docs.python-requests.org/en/master/user/quickstart/#response-status-codes) to raise an exception if the request fails. This depends somewhat on the desired behavior should the request fail; should this cause the entire page view to fail, or substitute some kind of error text instead? My preference would be for the former.

sebworks:
> A bit late to this, but does the built-in Wagtail EmbedBlock already support the use case of rendering arbitrary HTML? See their template and model code; this might be overkill for our needs but it does some nice things like cache the HTML.

Yep, the EmbedBlock is hot. Nice find Andy.

richaagarwal:
Thanks all for the feedback!  I'll investigate the EmbedBlock, otherwise sounds like the consensus is to store it in the db

willbarton:
@richaagarwal @contolini just a minor thing, but rather than cron, we should run it as a periodic Jenkins job instead. We're trying to eliminate our reliance on cron.

cfgov/v1/tests/atomic_elements/organisms/test_organisms.py
```diff
@@ -216,3 +216,18 @@ def test_item_introduction(self):
...
+            title='Browse Page',
+            slug='browse',
+        )
+        browse_page.content = StreamValue(
+            browse_page.content.stream_block,
+            [atomic.html_block],
+            True
+        )
+        publish_page(child=browse_page)
+        response = self.client.get('/browse/')
```
chosak: Minor thing, but this test will actually make a request to the URL in the `_atomic_helpers.html_block` value. This will fail if you run the tests without an internet connection (and I think we've also seen cases where Travis seems to block some outgoing requests). I'm not really sure of a nice and easy way to avoid this, short of mocking the requests call before running the test.

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -560,3 +561,26 @@ class VideoPlayer(blocks.StructBlock):
...
+        error_messages={
+            'required': 'The HTML URL field is required for rendering raw HTML from a remote source.',
+            'invalid': 'The URL is invalid or not allowed. ',
+        }
+    )
+
+    def render(self, value, context=None):
+        resp = requests.get(value['html_url'])
```
sebworks: Don't forget to add a timeout per: http://docs.python-requests.org/en/master/user/quickstart/#timeouts

cfgov/v1/atomic_elements/organisms.py
cfarm: we have the files deploying to s3 so we know they will be at http://files.consumerfinance.gov/ now - can you go ahead and add this?

willbarton:
We can use https://s3.amazonaws.com/files.consumerfinance.gov to access the bucket via HTTPS.

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -568,7 +568,7 @@ class HTMLBlock(blocks.StructBlock):
         label='Source URL',
         default='',
         required=True,
-        regex=r'^https://.+$',
+        regex=r'^https://s3.amazonaws.com/files.consumerfinance.gov/.+$',
```
contolini: Can we make the S3 part optional so that we can seamlessly transition to `https://files...` in the future? 

I'm thinking: `^https://(s3.amazonaws.com/)?files.consumerfinance.gov/.+$`

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -560,3 +561,26 @@ class VideoPlayer(blocks.StructBlock):
...
         icon = 'media'
         template = '_includes/organisms/video-player.html'
+
+
+class HTMLBlock(blocks.StructBlock):
+    html_url = blocks.RegexBlock(
+        label='Source URL',
+        default='',
+        required=True,
+        regex=r'^https://s3.amazonaws.com/files.consumerfinance.gov/.+$',
```
contolini: Can we make the S3 part optional so that we can seamlessly transition to `https://files.consumerfinance.gov` in the future? 

I'm thinking: `^https://(s3.amazonaws.com/)?files.consumerfinance.gov/.+$`

cfgov/v1/tests/atomic_elements/organisms/test_organisms.py
richaagarwal: Good catch Andy. I'm inclined to merge this for now but maybe can set up a test HTML page as part of the test.

PR #2622, opened 2016-12-22
Remove spokesperson from press page

contolini:
Buhbye. GHE 522.

## Removals

- Spokesperson who is no longer with the Bureau.


sebworks:
ninja

PR #2627, opened 2016-12-27
Fixes for External Redirect

sebworks:
The external redirect button isn't working and there is a jQuery reference.

## Changes

- Added fix for missing button event listener and removed jQuery.

## Testing

- Visit the homepage ( http://localhost:8000/ ).
- Click on the `Proceed to external site` button.

## Checklist

### DIFF
--- a/cfgov/unprocessed/js/modules/ExternalSite.js
+++ b/cfgov/unprocessed/js/modules/ExternalSite.js
@@ -16,7 +16,6 @@ var atomicHelpers = require( '../modules/util/atomic-helpers' );
  * @param {HTMLElement} element DOM Element.
  */
 function ExternalSite( element ) {
-
   var BASE_CLASS = 'external-site_container';
   var TOTAL_DURATION = 5;
   var INTERVAL = 1000;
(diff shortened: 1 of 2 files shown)


cfgov/unprocessed/js/modules/ExternalSite.js
```diff
@@ -24,6 +24,7 @@ function ExternalSite( element ) {
   var _dom = atomicHelpers.checkDom( element, BASE_CLASS );
   var _durationEl = _dom.querySelector( '.external-site_reload-container' );
   var _directEl = _dom.querySelector( '.external-site_proceed-btn' );
+  var _form = _dom.querySelector( 'form#proceed' );
```
anselmbradford: - Could be `_formEl` to fit with the format of the variables immediately preceding it. 

- Also, you could define the form dom reference inside `_gotoUrl()` instead of globally inside the module, since `_gotoUrl()` is the only place it's referenced and since it's referenced only once per page load.

anselmbradford: Though you might want to do the dom lookup before it's needed so that happens while the timer is going. Either should be fine.

PR #2629, opened 2016-12-29
Better filterable list form date validation

chosak:
Entering invalid dates in filterable list form query strings produces a 500 error, e.g.

http://www.consumerfinance.gov/about-us/newsroom/?form-id=0&filter0_from_date=BAD

We have JS-based validation on the frontend that should prevent users from submitting bad dates, but crawlers or URL hacking could produce invalid formats. This PR properly catches invalid dates and displays a "You have entered an invalid date." message.

## Additions

- `FilterableDateField` and related `FilterableFromDateField` and `FilterableToDateField` fields for better handling of invalid dates.

## Changes

- Some flake8 cleanup.

## Testing

- Load a production database and experiment at [http://localhost:8000/about-us/newsroom/](http://localhost:8000/about-us/newsroom/). Try entering valid or invalid dates in the UI, then try adding something like `?form-id=0&filter0_from_date=x` to see the error message.

## Screenshots

After entering an invalid date:
image

## Notes

- @sebworks in your original code change it seems like you were trying to provide the field name by using a `%s` in the error string. I wasn't able to get this to work properly without significant hacking of the form class, and didn't think it was worth pursuing 100% synchronization with the frontend error. Does that seem okay to you?

## Checklist


sebworks:
> @sebworks in your original code change it seems like you were trying to provide the field name by using a %s in the error string. I wasn't able to get this to work properly without significant hacking of the form class, and didn't think it was worth pursuing 100% synchronization with the frontend error. Does that seem okay to you?

@chosak, yeah this is fine.

cfgov/v1/forms.py
```diff
@@ -38,29 +60,37 @@ class FilterableListForm(forms.Form):
...
+    )
 
     def __init__(self, *args, **kwargs):
         self.hostname = kwargs.pop('hostname')
         self.base_query = kwargs.pop('base_query')
         super(FilterableListForm, self).__init__(*args, **kwargs)
-        page_ids = self.base_query.live_shared(self.hostname).values_list('id', flat=True)
+
+        pages = self.base_query.live_shared(self.hostname)
+        page_ids = pages.values_list('id', flat=True)
```
richaagarwal: Much more readable, thanks :)

cfgov/v1/forms.py
```diff
@@ -1,31 +1,53 @@
...
+        kwargs.setdefault('input_formats', self.default_input_formats)
+        kwargs.setdefault('widget', widgets.DateInput(
+            attrs=self.default_widget_attrs
+        ))
+        kwargs.setdefault('error_messages', ERROR_MESSAGES['DATE_ERRORS'])
+        super(FilterableDateField, self).__init__(*args, **kwargs)
+
+
+class FilterableFromDateField(FilterableDateField):
+    def __init__(self, *args, **kwargs):
```
richaagarwal: I can't figure out why these two classes are needed/ distinct since they have the same definition.

sebworks:
@chosak, will this all the date fields to except flexible date formats such at `10/15`?

cfgov/v1/forms.py
chosak: @richaagarwal These two classes add an extra CSS class to the form field widget (`js-filter_range-date__gte` and `js-filter_range-date__lte`, respectively).

chosak:
@sebworks so it would be easy to expand the backend validation options, although I haven't done that in this PR. Actually [the default Django `DateField` validation](https://docs.djangoproject.com/en/1.8/ref/forms/fields/#datefield) already supports three options (`2006-10-25`, `10/25/2006`, `10/25/06`) and we explicitly limit it for some reason. If we wanted to add some flexibility, though, we'd have to also modify the frontend validation code, and I'm not quite sure where that lives.

If expanding date validation choices on the frontend is easy to do, we could collaborate on a separate PR that does that.

sebworks:
@chosak, I think `10/15` and `2016` are currently accepted by the front-end via the date filters. We were using https://github.com/cfpb/cfgov-refresh/pull/2629/files#diff-a20999810035f4df9b089b5ce85e66b6L18 to allow the backend to accept flexible date formats. This is the regex for the front-end: https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/js/modules/util/validators.js#L24.

chosak:
@sebworks thanks for the additional context. I don't think the backend logic as it stands right now handles those cases properly. Using that line you link to:

```py
>>> from sheerlike.templates import get_date_obj
>>> get_date_obj('10/15').isoformat()
'2016-10-15T09:25:53.309628-04:00'
>>> get_date_obj('2016').isoformat()
'2016-12-01T09:24:05.003052-05:00'
```

So `10/15` is converted to 10/15/2016 at the local time, not 10/01/2015 at midnight as we'd like. Similarly, `2016` is converted to 12/01/2016 using the local month and time, not 01/01/2016. You can also verify this on the live site, e.g. [at the Newsroom here](http://www.consumerfinance.gov/about-us/newsroom/?form-id=0&filter0_title=&filter0_from_date=01%2F16&filter0_to_date=02%2F16).

In a457d01 I added backend support for these formats, which should match the frontend validation:

```py
default_input_formats = (
    '%m/%d/%Y',     # 10/25/2016
    '%m/%Y',        # 10/2016
    '%m/%y',        # 10/16
    '%Y',           # 2016
)
```

cfgov/v1/forms.py
rosskarchner: Could FilterableDateField just take that CSS class as a constructor argument?

sebworks:
💯, beautiful stuff.

cfgov/v1/forms.py
chosak: @rosskarchner that's true, although if any other form used this field we'd have to provide the same argument. I was thinking it made sense for this CSS to live as part of the field class, since it's field-specific not really form-specific.

PR #2651, opened 2017-01-05
Update knowledgebase to 2.2.4

rosskarchner:
## Fixes

- A CSS bug on AskCFPB (demonstrated below)
- For questions that exist in the database, attempts to retrieve a non-existent english or spanish answer will result in 404's, not 500's.

Here's an example URL that includes a valid Question ID, but there is no longer an associated english answer: http://www.consumerfinance.gov/askcfpb/145/what-happens-after-i-apply-for-a-loan.html

The CSS fix turns this:
screen shot 2017-01-05 at 11 03 50 am

Into this:
screen shot 2017-01-05 at 11 03 43 am


sebworks:
👍

PR #2656, opened 2017-01-05
WIP Feature: add line chart block to Wagtail

cfarm:
Looking for feedback on this Wagtail code from @cfpb/cfgov-backends! Check out `GHE/CFGOV/data-research/issues/526` for more context.

This WIP adds a new Chart Block to the Wagtail admin, for entering metadata to generate d3 charts for the Consumer Credit Trends pages.

Content entered in the Chart Block fields will replace the metadata that currently powers the charts in this file: https://github.com/cfpb/consumer-credit-trends/blob/master/src/static/js/templates/charts.js

This will allow our product owner to enter and manage the content for each chart.

## Additions

- Atomic chart organism JavaScript, based on JS from [consumer-credit-trends line.js](https://github.com/cfpb/consumer-credit-trends/blob/master/src/static/js/line.js), including dependencies
- In the spirit of iterative updates, this PR includes the Line Chart code only, for easier review.
- `chart.less` adds styles from  [consumer-credit-trends charts.less](https://github.com/cfpb/consumer-credit-trends/blob/master/src/static/css/charts.less), with updated atomic class names
- New Wagtail block fields required to make the JavaScript render each chart. These include what type of chart to load (this can be used later to determine which JS file to call), and the group property, which is used to filter data for many line charts.

## Removals

- `elementID` field from Chart Block. We don't want to rely on an ID for each chart to render any longer, so this is a step towards that. The JS now expects a matching title + class name instead of the ID. This will require manually updated migration.

## Changes

- `chart.html` template class names have been atomicized to match project conventions
- The original helper functions in the  [consumer-credit-trends line.js](https://github.com/cfpb/consumer-credit-trends/blob/master/src/static/js/line.js) assume that no `chartInfo.group` property is defined for charts that don't need it. Because this updated JS grabs the group property from the HTML data attribute, group property will exist for all charts.
- Another note about groups. Right now the JS expects the following properties for certain types of graphs. We can probably do the data munging/filtering on the Jenkins job that @contolini and @hillaryj are working on for these, instead of in the JS:
  - Origination activity, number/volume line charts: no group
  - Origination activity, year over year number/volume bar charts: "Number of loans" or "Dollar Volume"
  - Borrower Risk Profiles, Line charts: Credit score group (Subprime, prime, etc)
  - Income level line charts: income group (low, moderate, etc)
  - Borrower age line charts:  age group (younger than 30, 30-44, etc)

## Testing

- check out and create a "Chart block" element on a browse page to test the fields display on the page.
[...]

PR #2661, opened 2017-01-06
Chart Block

richaagarwal:
## Overview
- This supports ongoing work for the Data & Research team rendering d3 charts on Browse Pages, and is meant as an alternative to recent `HTML Block` work.
- We decided to create a `Chart Block` since the `HTML Block` wouldn't have supported interactive pages, which will be needed in a couple releases. 
- @cfarm has a PR up - https://github.com/cfpb/cfgov-refresh/pull/2656 - that is a WIP in terms of bringing the js into the repo, so this is meant to be a simplified version of that, that we can get out the door. 
- For the time being, the only required field to load a chart is the chart's element ID, which are defined here - https://github.com/cfpb/consumer-credit-trends/blob/master/src/static/js/templates/charts.js. Future work could include having the content editor specify another one of those fields if we didn't want them typing in the ID, and also pulling the title in. 

## TODO
- Add a test for rendering a `ChartBlock` 

## Screenshots
Wagtail Admin looks like:
screen shot 2017-01-06 at 2 00 37 pm

screen shot 2017-01-06 at 2 02 57 pm

And subsequent Browse Page looks like:

screen shot 2017-01-06 at 2 02 44 pm


cfgov/jinja2/v1/_includes/organisms/chart.html
```diff
@@ -0,0 +1,36 @@
+
```
richaagarwal: This mostly taken from @cfarm 's pull request https://github.com/cfpb/cfgov-refresh/pull/2656

PR #2663, opened 2017-01-09
Add a management command for deleting Wagtail pages

richaagarwal:
## Overview
- GHE 506
- Unpublishes & deletes a page given its slug name

## Usage
- `python cfgov/manage.py delete_page --slug page-slug` 

## Testing
- Grab a page's slug locally and type in the above, replacing `page-slug` as appropriate

### DIFF
--- a/cfgov/v1/management/commands/delete_page.py
+++ b/cfgov/v1/management/commands/delete_page.py
@@ -12,13 +12,29 @@ class Command(BaseCommand):
             nargs='+',
             help='The slug for the page you wish to delete'
         )
+        parser.add_argument(
+            '--id',
+            nargs='+',
+            help='The ID for the page you wish to delete'
+        )
 
     def handle(self, *args, **options):
-        slug = options['slug'][0]
+        page_to_delete = None
+        if options['slug']:
+            slug = options['slug'][0]
+            try:
+                page_to_delete = CFGOVPage.objects.get(slug=slug)
+            except Exception as e:
+                self.stderr.write(str(e))
+                raise CommandError('Slug `{}` did not work'.format(slug))
+        elif options['id']:
+            page_id = options['id'][0]
+            page_to_delete = CFGOVPage.objects.get(id=page_id)
+        else:
+            raise CommandError('Must supply a --slug or an --id')
         try:
-            page_to_delete = CFGOVPage.objects.get(slug=slug)
             page_to_delete.unpublish()
             page_to_delete.delete()
         except Exception as e:
             self.stderr.write(str(e))
-            raise CommandError('failed to delete page {}'.format(slug))
+            raise CommandError('Failed to delete page {}'.format(page_to_delete))
--- a/cfgov/v1/management/commands/delete_page.py
+++ b/cfgov/v1/management/commands/delete_page.py
@@ -1,6 +1,8 @@
+from django.core.exceptions import MultipleObjectsReturned
 from django.core.management.base import BaseCommand, CommandError
 
 from v1.models import CFGOVPage
+from v1.wagtail_hooks import flush_akamai
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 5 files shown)


cfgov/v1/management/commands/delete_page.py
```diff
@@ -0,0 +1,24 @@
...
+        parser.add_argument(
+            '--slug',
+            nargs='+',
+            help='The slug for the page you wish to delete'
+        )
+
+    def handle(self, *args, **options):
+        slug = options['slug'][0]
+        try:
+            page_to_delete = CFGOVPage.objects.get(slug=slug)
```
rosskarchner: - seems like this would break if a slug were ever used more than once (say, in different parts of the tree?)

cfgov/v1/management/commands/delete_page.py
```diff
@@ -0,0 +1,24 @@
...
+
+from v1.models import CFGOVPage
+
+
+class Command(BaseCommand):
+    help = 'Deletes a page by its slug name'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--slug',
```
chosak: Slugs [are not unique](https://github.com/wagtail/wagtail/blob/v1.7/wagtail/wagtailcore/models.py#L305) across the site, only within a particular directory. We ran into this with the D&R conference registration script, unfortunately you might have to specify page primary key instead.

cfgov/v1/management/commands/delete_page.py
chosak: This will raise an exception if more than one page exists with the given slug:

```py
>>> from v1.models import CFGOVPage
>>> CFGOVPage.objects.get(slug='agenda')
 ...
MultipleObjectsReturned: get() returned more than one CFGOVPage - it returned 2!
```

cfgov/v1/management/commands/delete_page.py
willbarton: There has to be a way to use the `route()` method on pages so we can use a path and not the primary key.

```
>>> from django.http import HttpRequest
>>> from wagtail.wagtailcore.models import Page
>>> request = HttpRequest()
>>> request.is_staging = True
>>> root = Page.objects.get(slug='cfgov')
>>> path = "/policy-compliance/enforcement/actions/"
>>> path_components = [component for component in path.split('/') if component]
>>> page, args, kwargs = root.route(request, path_components)
>>> page
<BrowseFilterablePage: Enforcement Actions>
```

(Based on [how Wagtail resolves paths](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailcore/views.py#L19-L20))

richaagarwal: Ah, I had no idea slugs were not unique - thanks @rosskarchner and @chosak for pointing that out. What if we could do it either by slug or by ID? For the majority of cases the former would work, and when that fails, the person would just have to go look up the ID for the one they want.

cfgov/v1/management/commands/delete_page.py
```diff
@@ -0,0 +1,40 @@
...
+            help='The ID for the page you wish to delete'
+        )
+
+    def handle(self, *args, **options):
+        page_to_delete = None
+        if options['slug']:
+            slug = options['slug'][0]
+            try:
+                page_to_delete = CFGOVPage.objects.get(slug=slug)
+            except Exception as e:
```
chosak: I would probably prefer to remove the ability to delete by slug because of non-uniqueness, but if you want to leave it in, can you catch [`MultipleObjectsReturned`](https://docs.djangoproject.com/en/1.8/ref/exceptions/#multipleobjectsreturned) here instead of the generic `Exception`, to avoid hiding some other error from being hidden (e.g. the database connection fails)?

cfgov/v1/management/commands/delete_page.py
```diff
@@ -0,0 +1,40 @@
...
+        parser.add_argument(
+            '--id',
+            nargs='+',
+            help='The ID for the page you wish to delete'
+        )
+
+    def handle(self, *args, **options):
+        page_to_delete = None
+        if options['slug']:
+            slug = options['slug'][0]
```
chosak: Instead of using `nargs='+'` and then referring to e.g. `options['slug'][0]` can you just use `nargs=?` to make the arguments optional?

cfgov/v1/management/commands/delete_page.py
```diff
@@ -0,0 +1,40 @@
...
+            except Exception as e:
+                self.stderr.write(str(e))
+                raise CommandError('Slug `{}` did not work'.format(slug))
+        elif options['id']:
+            page_id = options['id'][0]
+            page_to_delete = CFGOVPage.objects.get(id=page_id)
+        else:
+            raise CommandError('Must supply a --slug or an --id')
+        try:
+            page_to_delete.unpublish()
```
chosak: Question: is this for Akamai flushing purposes, or otherwise? If so, would it be better to first delete and then flush, to make sure that all children are deleted before Akamai is triggered?

richaagarwal: Good point -- I had done it for flushing purposes, but I could totally just call `flush_akamai` directly after as you mention.

cfgov/v1/management/commands/delete_page.py
richaagarwal: Yup! Thanks. I might've been guilty of some copypasta there 😁

cfgov/v1/management/commands/delete_page.py
```diff
@@ -0,0 +1,43 @@
...
+            except MultipleObjectsReturned as e:
+                self.stderr.write(str(e))
+                raise CommandError('Slug `{}` did not work'.format(slug))
+        elif page_id:
+            page_to_delete = CFGOVPage.objects.get(id=page_id)
+        else:
+            raise CommandError('Must supply a slug or an id')
+        try:
+            page_to_delete.delete()
+            flush_akamai()
```
richaagarwal: This will flush Akamai a second time within this command, because Wagtail unpublishes before it deletes, and we have a hook to flush when unpublishing. However, since we need to flush the site _after_ we delete, we must call it explicitly. This is not ideal since we're flushing the entire site each time we flush, so we'll need to come back to this at some point.

cfgov/v1/management/commands/delete_page.py
```diff
@@ -0,0 +1,43 @@
...
+        )
+
+    def handle(self, *args, **options):
+        page_to_delete = None
+        slug = options['slug']
+        page_id = options['id']
+        if slug:
+            try:
+                page_to_delete = CFGOVPage.objects.get(slug=slug)
+            except MultipleObjectsReturned as e:
```
willbarton: I think I'd still like to see path resolution sometime in the future, but that's not a blocker right now. Let's see how we end up using it.

PR #2664, opened 2017-01-10
HOLD: Demo alternate sharing approach

chosak:
**Do not merge!** This PR is intended to demonstrate a prototype of [this](https://github.com/cfpb/wagtail-sharing) alternate sharing approach. 

## Additions

- Adds dependence on [cfpb/wagtail-sharing](https://github.com/cfpb/wagtail-sharing) app for sharing of draft Wagtail content.

## Changes

- Use `wagtailsharing` URL router instead of Wagtail default.
- Disable reliance on a staging server hostname (e.g. `content.localhost:8000`) in favor of a hostname that allows for viewing of latest draft content.

## Testing

- Set the `WAGTAILSHARING_HOSTNAME` environment variable before you run your local server:

 ```sh
$ export WAGTAILSHARING_HOSTNAME='content.localhost:8000'
$ ./runserver.sh
 ```

- Run your local server. Create a new (draft only) page, and then also modify an existing published page with an unpublished change.
- Browse to these pages using the standard [http://localhost:8000](http://localhost:8000) hostname. You should get a 404 (not published yet) and see the old content, respectively.
- Browse to these pages using the new [http://content.localhost:8000](http://content.localhost:8000) hostname. You should see the draft content with a warning banner.

## Screenshots

image

## Checklist


richaagarwal:
The `WAGTAILSHARING_HOSTNAME` can be anything, theoretically, right?  I set mine to `share.localhost:8000` because it feels weird to test on `content.localhost:8000` since that's already configured to load shared pages in the code, so even if you don't set the hostname, `content` will show you something.  On `share.localhost:8000` I can't view anything unless it's also published.

richaagarwal:
Why is it that `path_components` here https://github.com/cfpb/wagtail-sharing/blob/master/wagtailsharing/views.py#L61 is empty if the page exists within Wagtail, and is set to the slug name if it doesn't exist?  Is there not a way to just check if the page exists directly?

chosak:
@richaagarwal thanks for the review. Needing to add `WAGTAILSHARING_HOSTNAME` as an additional site with the duplicate home page is something that I should have added to the wagtail-sharing docs; I'll do that.

Regarding `path_components`: this code is (somewhat painfully) hooking into how Wagtail determine "if the page exists directly", which is to parse the request URL path, piece by piece. So, say, the URL is `/foo/bar/baz`. First it'll try routing the path `/foo` under the site root page, which usually means "is there a page with slug `foo` under the root page?" but not always, due to the existence of [RoutablePageMixin](http://docs.wagtail.io/en/v1.8.1/reference/contrib/routablepage.html). Assuming that `/foo` is found, then the process is repeated, looking for `/bar`, and so on.

[The line you linked to](https://github.com/cfpb/wagtail-sharing/blob/master/wagtailsharing/views.py#L61) attempts to determine why an `Http404` is raised for a given path: is it [because the page does not exist](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailcore/models.py#L659) or [because the page is not live](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailcore/models.py#L668)? We want to handle the second case specially, and catch that `Http404` and turn it into a valid page request (by serving the draft version of the page).

Does that help clarify? I can add some comments to that code to better explain what's happening there.

chosak:
@richaagarwal see PR that adds new documentation at [wagtail-sharing#3](https://github.com/cfpb/wagtail-sharing/pull/3).

richaagarwal:
After looking at https://github.com/wagtail/wagtail/blob/c6666c6de5e83bf94d18324858c121e6584ba47d/wagtail/wagtailcore/models.py#L656 I have a better understanding of the need for the logic in `get_requested_page`. Thanks for the hard work on this Andy!

chosak:
@richaagarwal I went back to look at how sharing works because of your comment

> Update: Nevermind! Forgot to add it to "Sites".

I was confused about this but confirmed that you shouldn't need to add the sharing hostname under Sites at all. You do have to make sure that any test page you create exists under the site root of an existing Site, though (e.g. creating under CFGOV not next to it). I added some additional unit tests that check this at [wagtail-sharing#4](https://github.com/cfpb/wagtail-sharing/pull/4), and also verified it works locally with this PR using:

```sh
$ WAGTAILSHARING_HOSTNAME=drafts.localhost:8000 ./runserver.sh 
```

Create a new draft page under CFGOV (e.g. slug "foo") and it should be visible at http://drafts.localhost:8000/foo/ but not http://localhost:8000/foo/. Similarly, make unpublished edits to a page (just save them), and they should only show up at the shared hostname.

chosak:
Based on some feedback to this PR and elsewhere, I'm going to close this and re-open with a slightly different approach that's simpler and will work better for us. For posterity I tagged the version of wagtail-sharing that this PR was written against with version [0.1](https://github.com/cfpb/wagtail-sharing/releases/tag/0.1).

PR #2668, opened 2017-01-10
HOLD: *Another* sharing approach

rosskarchner:
Short version: this removes replaces most of the logic around "shared" content, and instead adds a new app for migrating content between sites in a single wagtail instance.

## getting started

To try it: copy your homepage (probably called 'CFGOV'), and call the new one something like 'www'. Don't check the boxes.

screen shot 2017-01-10 at 4 29 27 pm

configure your sites to use localhost:8000 and content.localhost:8000, and configure 'localhost' to point to the new root page.

screen shot 2017-01-10 at 1 58 33 pm

In settings, select "Workflow Destination Settings"

screen shot 2017-01-10 at 4 31 36 pm

Configure content.localhost:8000 (select the site from the right corner") and set the destination to 'localhost:8000)

screen shot 2017-01-10 at 4 33 14 pm

## Using it

If you've set things up correctly, http://content.localhost:8000/cfpb-ombudsman/ should work, while http://localhost:8000/cfpb-ombudsman/ 404's. Let's fix that.

Go to http://content.localhost:8000/cfpb-ombudsman/  and you should see the wagtail 'user bar' (the little bird in the bottom right corner). If you don't,  go log in at http://content.localhost:8000/admin/ and then come back. Click the bird, and you'll see a new link for "Migrate to localhost:8000 [default]". Click it.

This page tells you what's going to happen.
screen shot 2017-01-10 at 4 49 42 pm

Click "Do it!" to copy the page. This will leave you looking at the wagtail explorer, a level above the page you just created or updated

screen shot 2017-01-10 at 4 55 16 pm

Confirm that http://localhost:8000/cfpb-ombudsman/ now works
## What works

You can use this to publish or update pages from the "content" site to the "www" site, as long as the parent page exists. In other words, you would need to push '/foo'/ before '/foo/bar/'. If you want to play around with blog posts, you'll need to manually copy the 'About Us' page using wagtail admin, since it's a sheer powered page and doesn't expose the userbar.

## What's missing
- Akamai integration is assumed broken here
- RIght now, we refuse to update a page if the source and destination are different page types
- there's no bulk operations, or strategy for migrating the whole site to the new approach.
- haven't investigated how this interacts with scheduled publishing
- Everything is based on paths right now-- it you published /foo/bar, changed it to /foo/baz, and pushed again, both /foo/bar and /foo/baz would exist, which probably isn't what we actually want.

I don't think any of those things would be *hard* to fix, they just didn't seem necessary for proving the concept.

(I have more to say, but have to leave, so I'm creating the PR)


rosskarchner:
closing this!

PR #2669, opened 2017-01-11
Fix for filter categories on the Enforcement actions page. 

sebworks:
Fix for filter categories on the Enforcement Actions page. 

## Changes

- Modified `cfgov/v1/util/ref.py` to fix case issue.

## Testing

- Visit `http://localhost:8000/policy-compliance/enforcement/actions/` and expand the filter.
- Verify that the categories exist.

## Screenshots

screen shot 2017-01-10 at 11 02 51 pm

## Checklist

### DIFF
--- a/cfgov/v1/util/ref.py
+++ b/cfgov/v1/util/ref.py
@@ -62,7 +62,7 @@ categories = [
         ('data-research-reports', 'Data, research & reports'),
         ('info-for-consumers', 'Info for consumers'),
     )),
-    ('Enforcement action', (
+    ('Enforcement Action', (
         ('fed-district-case', 'Federal District Court Case'),
         ('admin-filing', 'Administrative Filing'),
     )),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 
 
 ### Fixed
-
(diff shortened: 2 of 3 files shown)


cfgov/v1/util/ref.py
```diff
@@ -62,7 +62,7 @@
         ('data-research-reports', 'Data, research & reports'),
         ('info-for-consumers', 'Info for consumers'),
     )),
-    ('Enforcement action', (
+    ('Enforcement Action', (
```
willbarton: Wow, this is worryingly fragile.

Scotchester:
Has this always been broken? It's said 'Enforcement action' since last January. And, what is the root cause? I don't see why anything would care what the capitalization is.

Regardless, I have a separate update to the list of categories that needs to go in today (sorry for just remembering 😳 ), so I'd appreciate it if this could get merged ASAP.

sebworks:
>> Has this always been broken? It's said 'Enforcement action' since last January. And, what is the root cause? I don't see why anything would care what the capitalization is.


The issue is https://github.com/cfpb/cfgov-refresh/blob/18abdaaf692ac19af62b729919c0731a807ba45d/cfgov/v1/util/ref.py#L184. We could check them after converting to lowercase.

willbarton:
Case-insensitive checking wouldn't help if the title were to change in some subtle way. But that's not a blocker for this PR.

PR #2673, opened 2017-01-11
Job pages should only have a single region

chosak:
Currently job listings (represented by the `jobmanager.JobListingPage` page type) can be associated with multiple geographic regions. For simplicity and to support some additional region-specific data, we want to change this so that each job is only ever associated with a single region. This PR makes the necessary changes to support that, including a migration that updates existing jobs.

## Additions

- `jobmanager.migrations.0012_jobs_have_one_region` migration converts existing jobs: if jobs have multiple regions, they are automatically assigned to the headquarters region.

## Removals

- Removed `JobListingPage` as an option in the Wagtail sidebar as it doesn't really make sense there and inserting jobs should only happen at a certain place in the tree.

## Changes

- `JobListingPage` now has a specific `region` attribute for one region only.
- Replaced the `jobmanager.Location` model with a more-appropriately named `jobmanager.JobRegion` one.

## Testing

- The best way to test this is with a production database dump:
 ```sh
 $ ./refresh-data.sh /path/to/db.sql
 ```
- Browse through existing job pages and notice that each one has an assigned region, and that regions have been assigned properly.
- Try to create a new `JobListingPage` and notice that the region field is required.
- Preview or view existing job pages to observe how the region is displayed.

## Screenshots

The region field on a job listing now always appears like this:

image

Existing regions can be viewed or edited through the Wagtail sidebar:

image

image

## Checklist


cfgov/jobmanager/migrations/0012_jobs_have_one_region.py
```diff
@@ -0,0 +1,115 @@
...
+            revision.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('jobmanager', '0011_delete_fellowshipupdatelist'),
+    ]
+
+    operations = [
```
chosak: This migration is a bit complicated, so here's what it's doing:

1. Runs `check_prerequisites` to verify two database assumptions: all jobs have at least one region and a region named "HQ" exists.
1. Creates the new region model as `jobmanager.JobRegion`.
1. Copies over existing `jobmanager.Location` model instances into new `jobmanager.JobRegion`s.
1. Adds a new `region` field to the `JobListingPage` page type, making it potentially null.
1. Runs `multiple_regions_to_single_region` to update all existing `JobListingPage` pages, to copy their (potentially multiple) region(s) from `Location` to `JobRegion` instances. This includes not only updating the page model but also all historical revisions. ಠ_ಠ
1. Updates the `JobListingPage` page type to make its `region` field not nullable.
1. Deletes the now-unneeded `RegionPanel` and `Location` models.

willbarton: The way this is structured though I find it very readable. 

Good migration, 💯, would review again.

chosak:
Thanks for the review @willbarton!

PR #2677, opened 2017-01-13
Disable page deletion via POST request

richaagarwal:
## Overview
- GHE 505
- Disables page deletion via `/delete` POST endpoint in Wagtail admin 
- Can still delete via command line as introduced in this PR -- https://github.com/cfpb/cfgov-refresh/pull/2663

## Screenshots
- What you see when you try to delete a page 
screen shot 2017-01-12 at 4 41 17 pm

## Testing
- Go to a page in the Wagtail admin, and add `/delete` to the path
- Confirm you see the exception
- Go back to the page and confirm it's still there

### DIFF
--- a/cfgov/v1/wagtail_hooks.py
+++ b/cfgov/v1/wagtail_hooks.py
@@ -3,6 +3,7 @@ import logging
 import os
 import requests
 
+from django.core.exceptions import PermissionDenied
 from django.conf import settings
 from django.contrib.auth.models import Permission
 from django.core.urlresolvers import reverse
@@ -22,6 +23,11 @@ from v1.util import util
 
 
 logger = logging.getLogger(__name__)
+
+
+@hooks.register('before_delete_page')
+def raise_delete_error(request, page):
+    raise PermissionDenied('Deletion via POST is disabled')
 
 
 @hooks.register('after_create_page')
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ Given the `MAJOR.MINOR.PATCH` pattern, here is how we decide to increment:
 - Updated custom Wagtail admin templates to 1.7 version.
 
 ### Removed
+- Can no longer delete via `/delete` in Wagtail
 
 
 ### Fixed


cfgov/v1/wagtail_hooks.py
```diff
@@ -19,6 +19,11 @@
 logger = logging.getLogger(__name__)
 
 
+@hooks.register('before_delete_page')
+def raise_delete_error(request, page):
+    raise Exception('Deletion via POST is disabled')
```
chosak: Would it be better to raise [`django.core.exceptions.PermissionDenied`](https://docs.djangoproject.com/en/1.8/ref/exceptions/#permissiondenied) here? This triggers a 403 error instead of a 500. It's also the same exception class [that Wagtail raises](https://github.com/wagtail/wagtail/blob/v1.7/wagtail/wagtailadmin/views/pages.py#L491) if someone tries to delete but doesn't have the correct permissions.

richaagarwal: With a 403 error the only output is "403 Forbidden" and it doesn't include the error message -- I don't have a strong opinion, but strictly speaking this isn't really a matter of permissions right?  We just don't want to allow anyone to be able to do it through the UI?

chosak: That's true. From a pragmatic point of view, the advantage of not raising an unhandled exception here is just to avoid treating it like some kind of failure (triggers error email, shows up in stats as a 500 error).

In production that error message won't appear either -- you'll see [our custom 500 page](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/urls.py#L397) which says "500 internal server error". From a user point of view it won't be obvious that the behavior is unsupported versus that some error occurred.

From a more pedantic point of view, 500 Internal Server Error [is supposed to mean](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html)

> The server encountered an unexpected condition which prevented it from fulfilling the request.

That doesn't really apply in this case: we know how to handle the request, we just want to choose to deny it. 403 Forbidden, on the other hand, is supposed to mean

> The server understood the request, but is refusing to fulfill it.

You're right that in both debug and production right now raising `PermissionDenied` will produce an ugly "403 Forbidden" page (this is true of any other views that do this). We could override this with a custom template (much like we have done with our 404 and 500 pages).

From an even _more_ pedantic point of view, it's good to avoid `raise Exception` using the base exception class as it makes it difficult for code higher up the stack to know what happened. From [PEP8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations):

> Design exception hierarchies based on the distinctions that code catching the exceptions is likely to need, rather than the locations where the exceptions are raised. Aim to answer the question "What went wrong?" programmatically, rather than only stating that "A problem occurred"

PR #2678, opened 2017-01-13
Adding Image inset to the full-width molecule

sebworks:

## Additions

- Added `cfgov/jinja2/v1/_includes/molecules/image-inset.html` to render the image inset molecule. Key thing is that we are using the srcset and sizes attribute on the image.  We also use Wagtail to downsize image and create an image for use on mobile. 

- Added  `cfgov/v1/atomic_elements/molecules.py` to render the `Image Inset` molecule within in the `FullTextWidth` organism.  Key thing is the image drop down which will determine the size of the inset.

## Changes

- Modified `cfgov/unprocessed/css/molecules/inset.less` to render the inset images molecule at various breakpoints. Key point is that we never stretch or upscale the image using CSS. We also ensure that the decorative image looks the same as the regular image.

- Modified `cfgov/jinja2/v1/_includes/organisms/full-width-text.html` to ensure the molecule doesn't get wrapped in a block div.

- Modified `cfgov/unprocessed/css/organisms/full-width-text-group.less` to add margins where appropriate at various breakpoints and float modifiers.

## Testing

- Create a `Blog` page in Wagtail.
- Add the `FullTextOrganism` to the `Blog` page.
- Add the `Content` module to the `FullWidthText` organism, and enter some text.
- Add the `Image Inset` organism to the `FullWidthText`.
- Add the `Content` module to the `FullWidthText` organism, and enter some text.
- Experiment with various options within the `Image Inset` interface.
- Preview and verify the image inset works as expected.

## Screenshots

screen shot 2017-01-13 at 3 16 27 am
screen shot 2017-01-25 at 5 10 29 pm

## Notes

- The inset macro is a little complicated; the code can be migrated to Django if people feel it's necessary.

- Might consider adding a bottom border option to Wagtail.

- Might consider adding the ability to select a mobile inset image to Wagtail.

## Todos

-

## Checklist


cfgov/v1/atomic_elements/molecules.py
```diff
@@ -246,6 +246,25 @@ class Meta:
         label = 'Phone'
 
 
+class ImageInset(blocks.StructBlock):
+    heading = blocks.CharBlock(required=False)
+    paragraph = blocks.RichTextBlock(required=False)
+    image = atoms.ImageBasic()
+    is_image_left = blocks.BooleanBlock(required=False,
+                                        label='Image float left')
+    image_width = blocks.ChoiceBlock(choices=ref.inset_image_widths,
```
chosak: Could this be changed to have `choices` refer to a callable, so that changing the options in future won't require a migration? Something like:

```py
# in ref.py
def get_inset_image_widths():
    return inset_image_widths

# in molecules.py
class ImageInset(blocks.StructBlock):
    ...
    image_width = blocks.ChoiceBlock(choices=ref.get_image_inset_widths, ...)
```

cfgov/jinja2/v1/_includes/molecules/image-inset.html
```diff
@@ -0,0 +1,74 @@
...
+
+    value.paragraph:         String for text below the image.
+
+    value.is_image_left:     Boolean indicating whether to float the image
+                             to the left.
+
+    value.image_width:       String representing the image width.
+                             Possible values: (150, 270, 300).
+                             Defaults to 270.
+
```
chosak: Can you add documentation for `is_image_decorative`?

schaferjh:
The screenshots look fine  @nataliafitzgerald have you seen these? Do they look okay?

What does the admin view look like for both of these options?

jimmynotjim:
I thought from a Wagtail perspective insets we're strictly these things:

screen shot 2017-01-13 at 3 46 22 pm

From the screen shot of the director, it seems like this is slightly different. I'd suggest renaming it to avoid confusion between the two components.

cfgov/jinja2/v1/_includes/molecules/image-inset.html
```diff
@@ -0,0 +1,74 @@
...
+{# Image sizes are doubled for retina displays. Images created via Wagtail
+   are never stretched or upscaled and maintain their aspect ratio. #}
+{% set image_inset = image(value.image.upload, 'width-' ~ value.image_width
+                     | int * 2) %}
+{% set image_mobile = image(value.image.upload, 'width-1200') %}
+<div class="m-inset
+            m-inset__image
+            {{ 'm-inset__image__left' if value.is_image_left else '' }}
+            {{ 'm-inset__image__' ~ ( value.image_width
```
jimmynotjim: Same here, the width value should be right on the parent as `m-insset__150`

cfgov/jinja2/v1/_includes/molecules/image-inset.html
```diff
@@ -0,0 +1,74 @@
...
+               if value.image_width else '270' ) }}
+            {{ 'm-inset__image__portrait'
```
jimmynotjim: And here as well as `m-inset__portrait`. I'm actually not quite sure why we need portrait and landscape modifiers. Shouldn't the content just fit within the container and it stretches vertically for whatever it needs?

sebworks:
@jimmynotjim, it should be able to create both.

cfgov/jinja2/v1/_includes/molecules/image-inset.html
```diff
@@ -0,0 +1,74 @@
...
+               if image_inset.height > image_inset.width else 'landscape' }}">
+    {% if value.heading %}
+    <h2 class="header-slug">
+        <span class="header-slug_inner">
+            {{ value.heading }}
+        </span>
+    </h2>
+    {% endif %}
+    <div class="m-inset_image-container">
+        {% if value.is_image_decorative %}
```
jimmynotjim: Should we ever be using decorative images in an inset? It seems it should always be content and not decorative.

cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -46,6 +68,83 @@
         margin-right: unit( @grid_gutter-width / @base-font-size-px, em );
     });
 
+    &__image {
+        float: right;
+
+        div.m-inset_image {
```
jimmynotjim: Why chain to the `div`? If we're keeping decorative images it should be a modifier like `m-inset_image__decorative`

sebworks:
screen shot 2017-01-13 at 5 05 31 pm

cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -46,6 +68,83 @@
...
+            background-size: contain;
+            background-repeat: no-repeat;
+        }
+
+        .m-inset_text {
+            margin-top: unit( 5px / @base-font-size-px, em )
+        }
+
+        .respond-to-max(@bp-xs-max, {
```
jimmynotjim: Minor: Check parens for proper spacing.

cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -46,6 +68,83 @@
...
+            margin-left: 0;
+            margin-right: 0;
+
+            .m-inset_image {
+                .u-flexible-container_inner();
+            }
+
+            img.m-inset_image {
+                height: auto;
```
jimmynotjim: Isn't height already auto for an `img` elem?

cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -46,6 +68,83 @@
...
+                width: auto;
+            }
+
+            div.m-inset_image {
+
+                // Important is necessary to override the inline height value;
+                height: 100% !important;
```
jimmynotjim: Why is there an inline height on a `div`?

cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -46,6 +68,83 @@
...
+                .u-flexible-container();
+            }
+        });
+
+        .respond-to-min(@bp-sm-min, {
+            &__150 {
+                width: 150px;
+            }
+
+            &__270 {
```
jimmynotjim: 270 is the standard width of insets. Do we need this modifier?

cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -46,6 +68,83 @@
...
+                width: 270px;
+            }
+
+            &__300 {
+                width: 300px;
+            }
+        });
+
+        &__left {
```
jimmynotjim: These rules should all be in the organism.

cfgov/jinja2/v1/_includes/molecules/image-inset.html
```diff
@@ -0,0 +1,74 @@
...
+
+    Creates image inset markup when given:
+
+    value:                   Object defined from a StreamField block.
+
+    value.heading:           The heading of the image inset.
+
+    value.image:             A basic image object to insert.
+
+    value.image.upload:      The image object.
```
sebworks: So `value.image.upload` is just a child object of  `value.image`

cfgov/jinja2/v1/_includes/molecules/image-inset.html
sebworks: This macro is just for image insets. Text only insets don't require a macro.

cfgov/jinja2/v1/_includes/molecules/image-inset.html
sebworks: I can see some cases where a decorative image makes sense. @schaferjh, what do you think?

cfgov/jinja2/v1/_includes/molecules/image-inset.html
jimmynotjim: I understand that, but we still shouldn't be including two modifiers in the same class name. Maybe @anselmbradford or @Scotchester have another idea to resolve it.

cfgov/unprocessed/css/molecules/inset.less
sebworks: Height is getting set to `100%` by `.u-flexible-container_inner`.

cfgov/unprocessed/css/molecules/inset.less
sebworks: The inline height is being set for the background image

cfgov/jinja2/v1/_includes/molecules/image-inset.html
```diff
@@ -0,0 +1,74 @@
...
+    value.image.alt:         String for alt text for screenreaders.
+
+    value.paragraph:         String for text below the image.
+
+    value.is_image_left:     Boolean indicating whether to float the image
+                             to the left.
+
+    value.image_width:       String representing the image width.
+                             Possible values: (150, 270, 300).
+                             Defaults to 270.
```
sebworks: This macro is just being used for image insets.

cfgov/jinja2/v1/_includes/molecules/image-inset.html
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2678/commits/98cde738538340d61bc985dcc16f0bdb95da36c9.

cfgov/jinja2/v1/_includes/molecules/image-inset.html
```diff
@@ -0,0 +1,74 @@
...
+
+    value.heading:           The heading of the image inset.
+
+    value.image:             A basic image object to insert.
+
+    value.image.upload:      The image object.
+
+    value.image.alt:         String for alt text for screenreaders.
+
+    value.paragraph:         String for text below the image.
```
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2678/commits/98cde738538340d61bc985dcc16f0bdb95da36c9.

cfgov/jinja2/v1/_includes/molecules/image-inset.html
```diff
@@ -0,0 +1,74 @@
...
+
+    value.is_image_left:     Boolean indicating whether to float the image
```
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2678/commits/98cde738538340d61bc985dcc16f0bdb95da36c9.

cfgov/jinja2/v1/_includes/molecules/image-inset.html
```diff
@@ -0,0 +1,74 @@
...
+   ========================================================================== #}
+
+{# Image sizes are doubled for retina displays. Images created via Wagtail
+   are never stretched or upscaled and maintain their aspect ratio. #}
+{% set image_inset = image(value.image.upload, 'width-' ~ value.image_width
+                     | int * 2) %}
+{% set image_mobile = image(value.image.upload, 'width-1200') %}
+<div class="m-inset
+            m-inset__image
+            {{ 'm-inset__image__left' if value.is_image_left else '' }}
```
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2678/commits/98cde738538340d61bc985dcc16f0bdb95da36c9

cfgov/jinja2/v1/_includes/molecules/image-inset.html
sebworks: Moved to parent in https://github.com/cfpb/cfgov-refresh/pull/2678/commits/98cde738538340d61bc985dcc16f0bdb95da36c9.

cfgov/unprocessed/css/molecules/inset.less
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2678/commits/98cde738538340d61bc985dcc16f0bdb95da36c9.

cfgov/unprocessed/css/molecules/inset.less
sebworks: Removed in https://github.com/cfpb/cfgov-refresh/pull/2678/commits/98cde738538340d61bc985dcc16f0bdb95da36c9.

cfgov/unprocessed/css/molecules/inset.less
sebworks: The rules need to exist on the molecule otherwise it changes the nature of the component.  You shouldn't define key essentials of the component outside the component. Margin is fine, but width and float are essential to what defines the inset component.  You would also be forced to recreate those rules across the rest of the parent components.


I've adjusted the code in https://github.com/cfpb/cfgov-refresh/pull/2678/commits/98cde738538340d61bc985dcc16f0bdb95da36c9 to reflect that thinking.

cfgov/v1/atomic_elements/molecules.py
sebworks: So I've deleted the changes in ref.py and just added the choices directly to the class. These values shouldn't really be changing so I don't think it warranted being added to ref.py.

cfgov/jinja2/v1/_includes/molecules/image-inset.html
sebworks: This is necessary to maintain the proper crop aspect ratio. It stems from https://github.com/cfpb/cfgov-refresh/blob/83c1281bd92868437c5a8b611d686f020d53f2dc/cfgov/unprocessed/css/misc.less#L175.

sebworks:
@jimmynotjim, please take a look when you get a chance. I've addressed the comments and added a boolean flag for the bottom rule.

cfgov/jinja2/v1/_includes/molecules/image-inset.html
schaferjh: Stakeholders are requesting what could be called "decorative" images in order to draw attention to call-out text.

schaferjh: However, I'm not sure what the "decorative" check box accomplishes?

I'm also uncertain about the image width -- is that a drop down with a set of options or a fill-in?

cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -7,9 +7,16 @@
         <div class="o-full-width-text-group">
             <div class="m-full-width-text"></div>
             <div class="m-inset">
-                <div class="m-quote">
+                <h2 class="header-slug">
+                    <span class="header-slug_inner"></span>
+                </h2>
+                <div class="m-inset_image-container">
+                    <div class="m-inset_image"></div>
```
jimmynotjim: this should have `m-inset_image__decorative` now as well right?

cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -7,9 +7,16 @@
...
+                    <img class="m-inset_image" />
                 </div>
+                <div class="m-quote"></div>
             </div>
+            <div class="m-inset_text"></div>
```
jimmynotjim: Should this text be inside the `m-inset`?

cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -19,32 +26,102 @@
...
             .m-inset
-              .m-quote
-              .m-related-links
-              .m-call-to-action
-            .m-full-width-text
+            .m-inset__bottom-rule
+            .m-inset__left
+            .m-inset__portrait
+            .m-inset__150
+            .m-inset__300
```
jimmynotjim: I'd pick one width to showcase here to avoid any confusion.

cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -19,32 +26,102 @@
...
+                .m-inset_image-container
+                    div.m-inset_image__cover
+                    div.m-inset_image__decorative
+                    img.m-inset_image
+                .m-inset_text
+                .m-quote
+                .m-related-links
+                .m-call-to-action
+                .m-image-inset
+                .m-full-width-text
```
jimmynotjim: This should be indented one less, there will never be full width inside an inset.

cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -19,32 +26,102 @@
...
+                    .header-slug_inner
+                .m-inset_image-container
+                    div.m-inset_image__cover
+                    div.m-inset_image__decorative
+                    img.m-inset_image
+                .m-inset_text
+                .m-quote
+                .m-related-links
+                .m-call-to-action
+                .m-image-inset
```
jimmynotjim: Does this exist?

cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -19,32 +26,102 @@
...
+                .m-full-width-text
   tags:
     - cfgov-molecules
   notes:
     - These are held in a full-width-text-group container to be held with
-      various full-width text blocks,  insets and tables as needed.
+      various full-width text blocks, insets and tables as needed.
```
jimmynotjim: Minor, add a comma after `insets` to avoid the wrath of @Scotchester

cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -19,32 +26,102 @@
...
+        &__150 {
+            width: 150px;
+        }
+
+        &__300 {
+            width: 300px;
+        }
+
+        &__image {
+            float: right;
```
jimmynotjim: Aren't all insets floated right by default?

sebworks: The image insets and text insets behave differently at the various breakpoints. We could unify them though.  It's `@bp-sm-min` for text and `@bp-xs-max` for images.

cfgov/unprocessed/css/molecules/inset.less
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2678/commits/74c81849f5ecd2cb7d3e9cc07c71d8dc53489f31.

cfgov/unprocessed/css/molecules/inset.less
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/2678/commits/74c81849f5ecd2cb7d3e9cc07c71d8dc53489f31.

cfgov/unprocessed/css/molecules/inset.less
sebworks: Removed in https://github.com/cfpb/cfgov-refresh/pull/2678/commits/74c81849f5ecd2cb7d3e9cc07c71d8dc53489f31.

cfgov/unprocessed/css/molecules/inset.less
sebworks: yep, fixed in https://github.com/cfpb/cfgov-refresh/pull/2678/commits/74c81849f5ecd2cb7d3e9cc07c71d8dc53489f31.

cfgov/unprocessed/css/molecules/inset.less
sebworks: fixed in https://github.com/cfpb/cfgov-refresh/pull/2678/commits/74c81849f5ecd2cb7d3e9cc07c71d8dc53489f31.

cfgov/unprocessed/css/molecules/inset.less
sebworks: removed in https://github.com/cfpb/cfgov-refresh/pull/2678/commits/74c81849f5ecd2cb7d3e9cc07c71d8dc53489f31.

cfgov/unprocessed/css/molecules/inset.less
jimmynotjim: Ah ok. I agree it'd be nice to have these unified a bit so it's clear what to expect instead of thinking about the type of inset it is.

cfgov/jinja2/v1/_includes/molecules/image-inset.html
jimmynotjim: Decorative in this case means it's not a part of the content, it's a background image set on an empty element. I think the only place we've really ended up using them is with isocons and some heros, but I could he wrong. @Scotchester would probably know best.

Scotchester: @schaferjh On a practical level, what it does is prevent screenreaders from reading out redundant or superfluous text. The vast majority of our illustrations fall into this category. Describing them aloud, or even announcing the presence of an image, is of no benefit to the user of a screenreader.

sebworks: What is the difference between this image (not decorative): 

http://www.consumerfinance.gov/about-us/blog/your-complaint-is-more-than-data-its-your-story/ 

and these isocon images on the careers page (decorative):

http://www.consumerfinance.gov/about-us/careers/.  Is it the accompanying text?

We also have http://www.consumerfinance.gov/your-story/, which has non-decorative icons.

Scotchester: There shouldn't be a difference, in an ideal world. All of these examples _should_ be considered decorative. The inconsistent implementation comes down to workflow constraints or lack of knowledge around this issue.

The image on the blog post _should_ have been decorative, but this is pretty difficult to do correctly in a CMS (i.e., making it a CSS background image). Fortunately this image _does_ do the next best thing, which is having an empty `alt` attribute, which I believe means that most screenreaders will skip over it.

The isocons on http://www.consumerfinance.gov/your-story/ are also decorative, and should have probably been coded as background images, since this page was built by developers.

sebworks: ok thanks @Scotchester. It sounds like we still need the capability to add decorative inset images.

schaferjh: Thanks for the clarification. I'd like to make sure that our content strategists and designers understand the implication here -- As long as we are taking this distinction into account when creating and uploading images, then this seems like a solid option.

I'm going to make a more general comment below.

sebworks:
> Deciding to enter a header is the deciding factor about whether this content displays within an inset design, or free-floating with text around it. That's unexpected.

The text will always wrap the module in the same fashion.  The differences in the appearance are due to optional elements:

- Header
- Body
- Adding a bottom rule 

> As we introduce the decorative image option, we're still asking for alt text as a blanket requirement for the module. That's also unexpected.

There are some options we can explore in this regard. The best option might be to add a touch of interactivity to the Wagtail UI (select if it's a decorative image first, then show the image options).

sebworks:
PR has been updated based on meeting and previous comments. The following changes were made:

- Removed the Heading option
- Made the float checkbox a dropdown
- Unified the breakpoint for the image and text insets.

sebworks:
screen shot 2017-01-25 at 3 51 08 pm

jimmynotjim:
Looks good from a FEWD perspective, thanks for following through.

PR #2680, opened 2017-01-13
Fix window load event

musketeer925:
Hey big fan of the Bureau I saw you guys on reddit awhile back https://www.reddit.com/r/Python/comments/4crahd/the_consumer_finance_protection_bureau_is/

Noticed a bad jQuery practice in your code.


sebworks:
Thanks for the PR @musketeer925.

PR #2682, opened 2017-01-17
Adds reusable text snippet

richaagarwal:
## Overview
- See GHE 255
- Adds a new snippet class to handle reusable text

## Screenshots
- Currently you can add it to a full width text module, and the sidefoot. Can add to other modules as needed.
screen shot 2017-02-23 at 10 58 41 pm

- Check out the search! Thanks @chosak for pointing this functionality out
screen shot 2017-02-23 at 10 58 54 pm

## Try it out
- Check out the branch, run migrations, doesn't matter if you're on a prod database or not
- Create a few reusable snippets from the Wagtail admin
- Create any page and add a full width text module, select "Reusable text", select a snippet
- Check out the published page to confirm it shows the snippet's text


cfgov/jinja2/v1/_includes/organisms/full-width-text.html
```diff
@@ -29,12 +29,14 @@
                 {% set media = image(block.value, 'original') %}
                 <img class="m-full-width-media_image" src="{{ media.url }}" alt="{{ media.alt }}">
             </div>
-        {% elif 'image_inset' in block.block_type %}
-            {{ block | safe }}
-        {% else %}
+        {% elif block.block_type in ['quote', 'cta', 'related-links'] %}
```
richaagarwal: if you have questions about this, ask @sebworks :)

cfgov/jinja2/v1/_includes/organisms/full-width-text.html
```diff
@@ -29,12 +29,14 @@
...
             <div class="m-inset">
                 {{ block | safe }}
             </div>
+        {% elif 'reusable_text' == block.block_type %}
```
richaagarwal: Right now there's no custom template for this snippet, since I didn't create a Wagtail block for it the way there exists `MainContactInfo` for the contact snippet.  This seems like the most straightforward way to implement this, but I'm open to suggestions as ideally we'd move away from having this type of logic in the template.

cfgov/jinja2/v1/_includes/organisms/full-width-text.html
willbarton: @sebworks What is `cta`?

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -17,7 +17,7 @@
 
 from . import atoms, molecules
 from .. import blocks as v1_blocks
-from ..models.snippets import Contact as ContactSnippetClass
+from v1.models.snippets import Contact as ContactSnippetClass, ReusableText
```
willbarton: Nitpick: this import line bugs me with the combined "import as" and straight "import".

cfgov/jinja2/v1/_includes/organisms/full-width-text.html
sebworks: `call to action`

schaferjh:
Yes, we definitely want to use this in the sidebar area as well. Currently the module used is the `Related content` sidefoot module -- could this be modified to be used there as well?

cfgov/jinja2/v1/_includes/organisms/full-width-text.html
chosak: Yes, this logic makes me sad, but I also don't know a simpler better way right now. The work in GHE#378 might enable an alternate approach, but at root it does seem like the issue is that snippets don't know how to render themselves in the same ways that other blocks do. At least with other blocks you can define a template that gets used when the block is rendered.

One possible approach might be to derive a custom class like `ReusableTextChooserBlock` from [`SnippetChooserBlock`](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailsnippets/blocks.py#L9), that has a `class Meta: template = ...`.  That way you could encapsulate the logic of "how to render ReusableText snippets" in the class template itself.

cfgov/v1/atomic_elements/organisms.py
richaagarwal: Got it - would you recommend separating them onto different lines?

richaagarwal:
@schaferjh Do you mean you want it to show up as an option in the sidefoot, or that you'd want it to be a part of `Related Content` itself?

schaferjh:
@richaagarwal Ah -- either way, frankly. Since this PR adds it as a sub-option to `Full width text` I was not sure if it needed to be added into an existing module or if it could stand alone. If it can stand alone, then it can just be another option in the sidefoot.

What does the reusable text input screen look like in Wagtail?

richaagarwal:
There's a screenshot above, and this is how it would appear in the sidefoot area - 
screen shot 2017-02-27 at 8 33 20 am

richaagarwal:
I've added to the sidefoot per @schaferjh's request and I've created a `ReusableTextChooserBlock` per @chosak's suggestion.  Ready for a final review, or I'll be merging later this afternoon. Thanks!

schaferjh:
@richaagarwal thanks re: the sidefoot! 

as for a screenshoot -- I'm looking for a screenshot of what the Wagtail UI looks like when a user is creating a bit of reusable text. Is there a header field and a body field? or just one text field? how does a user initiate creating one of these reusable text items?

cfgov/jinja2/v1/_includes/snippets/reusable_text.html
```diff
@@ -0,0 +1 @@
+{{ value['text'] | safe }}
```
chosak: Nice to see that this is actually a reasonable option! Two minor suggestions:

1. Can we remove the `| safe` here? We put that into some blocks as a workaround to support some inline HTML tags, but I don't think we want to make it a rule in general unless there's an explicit need.
2. Instead of using `value['text']` can we use `value.text` to be more consistent with other organism templates? It also might be nice to have some header documentation [like this](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/email-signup.html#L1).

richaagarwal:
@schaferjh Oh got it! Here it is -- note that the "Title" doesn't get displayed on the webpage, but would be used for internal display/searching. 

screen shot 2017-02-27 at 9 18 33 am

schaferjh:
@richaagarwal thanks. I'm assuming that text field is a rich text field-- is that correct?

If the title is not displayed as part of adding this item to a page, I think we need some kind of hint or instructional text to convey that. @sarahsimpson09 @jordanafyne Any thoughts on how to ensure this distinction is clear?

I suppose the field name could just be changed to "Snippet title (internal only)" or something like that?

sarahsimpson09:
@schaferjh @jordanafyne what if we called it "Internal title"?

richaagarwal:
@schaferjh @sarahsimpson09 I should've asked to begin with -- do we want to display a title?  I had assumed not, but we totally can.

jordanafyne:
@schaferjh @sarahsimpson09  - I like the full "Snippet title (internal only)" if this is just being used as clarification for content editors.

schaferjh:
Ok, let's go with "Snippet title (internal only)" for the field title.

As long as the text field is a rich text field where users can use the heading styles etc, I _think_ we're fine without a display title field @richaagarwal. It's unlike the rest of the sidefoot modules, but I don't know the code reasoning for why other sidefoot modules have seperate fields for headings, body, etc.

richaagarwal:
Sounds good.  I didn't want to tie us down to a title/ body format because I thought we might use reusable text (maybe further down the road) in places like paragraphs, etc. too where we just want to insert something in, in which context a header wouldn't make sense.

PR #2684, opened 2017-01-18
Add resource snippets and snippet list module

Scotchester:
This PR defines the Resource snippet type and creates a Snippet List module to be able to output them in a filtered, compact format.

**We'd very much like to get this in tomorrow's 4.6 code freeze.** Sorry for the short turnaround, but the code has already been fairly thoroughly vetted via pairing with @chosak and @jimmynotjim and the concept by @schaferjh.

## Additions

- Resources snippet definition
- Snippet List module

## Changes

- Fixes Image and Text 25/75 output to not output an empty `h3` if `heading` is not set.
- Simplifies (and also makes more complex) the `.o-table__stack-on-small` code

## Testing

1. Pull branch and check it out
2. `./cfgov/manage.py migrate`
3. `./cfgov/manage.py runserver`
4. Go to the Snippets screen and create a new Resource, making sure to add a related file or link, and giving it a couple tags. **Remember the tag names you gave it.**
5. Open or create a browse page
6. Create a Snippet List module in the main content area. Be sure to select "Resources" for the snippet type and enter tag names to match the tags you gave the snippet you created. Also, create at least one action, choosing the snippet field that corresponds to to the related file or link field you entered on your snippet.
7. Preview the page

## Screenshots

screen shot 2017-01-18 at 13 44 19

screen shot 2017-01-18 at 13 45 30

screen shot 2017-01-18 at 13 46 39

## Notes

- The tag-based filtering mechanism isn't ideal, but it's a working MVP. Ideally, we'd have a hard-coded list of snippet (sub)categories for editors to choose from, but until we get closer to continuous delivery and/or our idea of what tags gets more solidified, having to deploy code to update that list would be extremely painful.

## Todos

- [x] Write tests for the new `filter_by_tags` method on `TaggableSnippetManager`.

## Checklist


cfgov/jinja2/v1/_includes/organisms/snippet-list.html
```diff
@@ -0,0 +1,97 @@
...
+   value.body:                      A description for the snippet list intro.
+
+   value.image:                     An image for the snippet list intro.
+                                    If set, the snippet list intro will
+                                    format as a 25/75 Image and Text module.
+
+   value.snippet_type:              The type of snippet being listed.
+                                    Options defined in ref.py.
+
+
```
jimmynotjim: Minor: Extra break here.

cfgov/v1/models/snippets.py
```diff
@@ -39,6 +49,93 @@ def get_by_title_slug(self, title, slug):
...
+    def __str__(self):
+        return self.title
+
+    @classmethod
+    def get_by_title_slug(self, title, slug):
+        return self.objects.get(hash=hashlib.md5(title + ';;' + slug).hexdigest())
+
+
 @receiver(pre_save, sender=Contact)
 def set_hash(sender, instance, **kwargs):
```
chosak: Perhaps the new `Resource` snippet should be below this receiver so as to keep it close to the `Contact` model it applies to?

cfgov/v1/models/snippets.py
```diff
@@ -39,6 +49,93 @@ def get_by_title_slug(self, title, slug):
...
+            return self.subtype + '-single.html'
+        else:
+            return self.subtype + '-list-item.html'
+
+    def __str__(self):
+        return self.title
+
+    @classmethod
+    def get_by_title_slug(self, title, slug):
+        return self.objects.get(hash=hashlib.md5(title + ';;' + slug).hexdigest())
```
chosak: Is this just a copy of the code on `Contact` above? If so, won't this snippet model also need the save receiver to actually set the hash? I'm curious where this method is used as [it doesn't seem to be called anywhere](https://github.com/cfpb/cfgov-refresh/search?utf8=%E2%9C%93&q=get_by_title_slug).

cfgov/v1/util/ref.py
```diff
@@ -133,6 +133,24 @@
     ('ht', 'Haitian Creole'),
 ]
 
+snippet_types = [
```
chosak: I'm curious, do we even need to hardcode these types at all? Wagtail has a built-in [`wagatilsnippets.models.get_snippets`](https://github.com/wagtail/wagtail/blob/v1.7/wagtail/wagtailsnippets/models.py#L11) method that returns a list of all registered snippet models (those registered with `@register_snippet`).

We could set the [`verbose_name_plural`](https://docs.djangoproject.com/en/1.8/ref/models/options/#verbose-name-plural) on `class Meta` our existing snippet models (or just use the default, since it'll work in this case), and then we don't need this hardcoded list at all.

Similarly, for the fields, those could be defined on the models themselves (as a class member or a method).

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -678,3 +678,36 @@ class Meta:
...
         template = '_includes/organisms/chart.html'
+
+
+class SnippetList(blocks.StructBlock):
+    heading = blocks.CharBlock(required=False)
+    body = blocks.RichTextBlock(required=False)
+    image = atoms.ImageBasic(required=False)
+
+    snippet_type = blocks.ChoiceBlock(
+        choices=ref.snippet_types,
```
chosak: Using a callable here would avoid the generation of migrations should the list of snippets change in future. See related comment below about potentially moving away from a hardcoded list.

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -678,3 +678,36 @@ class Meta:
...
+        required=True
+    )
+    actions = blocks.ListBlock(blocks.StructBlock([
+        ('link_label', blocks.CharBlock(
+            help_text='E.g., "Download" or "Order free prints"'
+        )),
+        ('snippet_field', blocks.ChoiceBlock(
+            choices=ref.snippet_fields,
```
chosak: See comment above about using a callable here instead.

cfgov/unprocessed/css/organisms/snippet-list.less
```diff
@@ -0,0 +1,44 @@
...
+   o-snippet-list
+   ========================================================================= */
+
+.o-snippet-list_list {
+    display: table;
+    width: 100%;
+    margin-top: unit( 30px / @base-font-size-px, em );
+    border-collapse: collapse; // must be set to allow borders on table-rows
+
+    > ul {
```
jimmynotjim: Should probably change this and the other element attributes below to the corresponding  `data-attr` to be consistent with the suggested changes to the cf-tables enhancement.

cfgov/v1/models/snippets.py
Scotchester: Yes it is. I'll just get rid of it.

cfgov/v1/models/snippets.py
Scotchester: Will do.

cfgov/v1/models/snippets.py
```diff
@@ -39,6 +49,93 @@ def get_by_title_slug(self, title, slug):
...
+@register_snippet
+class Resource(ClusterableModel):
+    title = models.CharField(max_length=255)
+    desc = RichTextField(verbose_name='Description', blank=True)
+
+    thumbnail = models.ForeignKey(
+        'v1.CFGOVImage',
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
```
chosak: One more thought: would it be better to use `models.PROTECT` here? That would prevent deletion of any image or document that was still being referred to by a Resource snippet.

cfgov/v1/util/ref.py
Scotchester: Worked out a good solution in further pairing with @chosak! Stay tuned.

cfgov/v1/models/snippets.py
Scotchester: Sounds like it would be better. I just copied this code from elsewhere; should we always be using PROTECT instead of SET_NULL?

cfgov/unprocessed/css/organisms/snippet-list.less
Scotchester: Addressed in 23f9d15.

jimmynotjim:
Looks good from a FEWD perspective.

cfgov/jinja2/v1/_includes/organisms/snippet-list.html
Scotchester: Addressed in ab6a6a0.

cfgov/v1/atomic_elements/organisms.py
Scotchester: Addressed in ab6a6a0.

cfgov/v1/atomic_elements/organisms.py
Scotchester: Addressed in ab6a6a0.

cfgov/v1/models/snippets.py
Scotchester: Addressed in ab6a6a0.

cfgov/v1/models/snippets.py
Scotchester: Addressed in ab6a6a0.

cfgov/v1/util/ref.py
Scotchester: Addressed in ab6a6a0.

Scotchester:
I have added tests for the new function we wrote (`filter_by_tags`)!

cfgov/v1/models/snippets.py
willbarton: @Scotchester I guess the question is, is preventing the deletion of an image or file desirable?

Scotchester: On this particular line, it wouldn't have a very adverse affect, because the thumbnail is a bonus thing that isn't actually rendered right now, and may never be.

On the other hand, If a Document selected for the `related_file` or `alternate_file` fields below was deleted, that would potentially be a more serious breakage of the snippet in question.

Another question about protecting: What is the messaging like when someone tries to delete that Image or Document? Is it smart enough to say something specific like "Hey now, snippet XYZ uses that Document. I can't let you delete that, Dave." If not, it could be maddening wanting to delete a file and not knowing where to go to remove it and un-protect it.

willbarton: Apparently, Wagtail [doesn't handle `ProtectedError`](https://github.com/wagtail/wagtail/issues/1602), so probably a 500.

So, maybe that's not the thing to do after all.

Scotchester: :+1:

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -678,3 +679,47 @@ class Meta:
...
+    image = atoms.ImageBasic(required=False)
+
+    snippet_type = blocks.ChoiceBlock(
+        # choices=ref.snippet_types,
+        choices=[
+            (
+                m.__module__ + '.' + m.__name__,
+                m._meta.verbose_name_plural.capitalize()
+            ) for m in get_snippet_models()
+        ],
```
willbarton: a714e2147fd22069b3f0432852de5857

cfgov/v1/models/snippets.py
```diff
@@ -48,3 +58,93 @@ def set_hash(sender, instance, **kwargs):
...
+    ]
+
+    # Not needed now, but potentially useful if we ever want to render resource
+    # snippets individually.
+    # def template(self, is_single=False):
+    #    # return concatenation of subtype and kind of template
+    #    if is_single:
+    #       return self.subtype + '-single.html'
+    #    else:
+    #        return self.subtype + '-list-item.html'
```
willbarton: Is this something we'd ever want to do? Should we keep the commented code in for that reason alone? `subtype` doesn't appear to be defined on `Resource` or `ClusterableModel`.

Scotchester: I can see it being something we'd want to do, but I can also see this approach not working. And good point about the lack of `subtype` definition. I'll just pull this chunk.

willbarton: Good thing is, it'll be in the git history this way :)

PR #2681, opened 2017-01-14
Add line chart JavaScript to Chart Block

cfarm:
Updates the chart block from PRs #2657 and #2656, for entering metadata to generate d3 charts for the Consumer Credit Trends pages. This PR adds JS for the line charts to render based on the fields from the Wagtail block.

Content entered in the Chart Block fields will replace the metadata that currently powers the charts in this file: https://github.com/cfpb/consumer-credit-trends/blob/master/src/static/js/templates/charts.js

This will allow our product owner to enter and manage the content for each chart.

## Additions

- Atomic chart organism JavaScript, based on JS from [consumer-credit-trends line.js](https://github.com/cfpb/consumer-credit-trends/blob/master/src/static/js/line.js), including dependencies
- In the spirit of iterative updates, this PR includes the Line Chart code only, for easier review. Subsequent PR's should be submitted to add the Bar + Tile Map charts.
- `chart.less` adds styles from  [consumer-credit-trends charts.less](https://github.com/cfpb/consumer-credit-trends/blob/master/src/static/css/charts.less), with updated atomic class names
- New Wagtail block fields required to make the JavaScript render each chart. These include what type of chart to load (this can be used later to determine which JS file to call), and the group property, which is used to filter data for many line charts.

## Removals

- `elementID` field from Chart Block. We don't want to rely on an ID for each chart to render any longer, so this is a step towards that. The JS now expects a matching title + class name instead of the ID. This will require manually updated migration.

## Changes

- `chart.html` template class names have been atomicized to match project conventions
- The original helper functions in the  [consumer-credit-trends line.js](https://github.com/cfpb/consumer-credit-trends/blob/master/src/static/js/line.js) assume that no `chartInfo.group` property is defined for charts that don't need it. Because this updated JS grabs the group property from the HTML data attribute, group property will exist for all charts.
- Another note about groups. Right now the JS expects the following properties for certain types of graphs. We can probably do the data munging/filtering on the Jenkins job that @contolini and @hillaryj are working on for these, instead of in the JS:
  - Origination activity, number/volume line charts: no group
  - Origination activity, year over year number/volume bar charts: "Number of loans" or "Dollar Volume"
  - Borrower Risk Profiles, Line charts: Credit score group (Subprime, prime, etc)
  - Income level line charts: income group (low, moderate, etc)
  - Borrower age line charts:  age group (younger than 30, 30-44, etc)

## Testing

- check out and create a "Chart block" element on a browse page to test the fields display on the page.
[...]

PR #2688, opened 2017-01-20
Chart Block

contolini:
Straight outta Dada here's the latest iteration of the chart block. It includes all the goodness @cfarm added in #2681 with some simplifications.

screen shot 2017-01-20 at 12 04 32 am

I did my best to make it project-agnostic by decoupling it from Consumer Credit Trends' charts:

- I removed most of the JS and Less. It will be replaced by Highcharts stuff that comes from cfpb-chart-builder.
- The data source field now wants a filepath relative to "http://files.consumerfinance.gov/data/" (the "consumer-credit-trends/" directory is no longer hard-coded).
- The only thing chart.js does is import cfpb-chart-builder (from node_modules) and use `atomicHelpers` to initialize all the charts. This means cfpb-chart-builder will have to be compatible with [`instantiateAll()`](https://github.com/cfpb/cfgov-refresh/blob/72103466e97721b414afc3a887ab5233b3ba2e36/cfgov/unprocessed/js/modules/util/atomic-helpers.js#L122-L132).

Notes:

- As a fallback for users with unreliable internet connections, I added a description field that gets output into the charts' containers. If the charts fail to render for any reason users can read a description of each data visualization.
- The charts' titles and footnotes are specific to CCT so they're still in the chart organism (opposed to being generated by cfpb-chart-builder). Styling is in chart.less which imports cfpb-chart-builder's stylesheet which currently doesn't exist but will soon.
- `January 2017` is hardcoded as the release date. We'll make that dynamic in a future release.

Questions:

- @mistergone  @cfarm  Do we need the chart `group` and `report-type` fields in Wagtail? Or can they be determined in cfpb-chart-builder (possibly by analyzing the data source's filename)?

Stuff to do before we merge this PR:

- Publish a v1.x.x version of cfpb-chart-builder once we finish it and ensure it's compliant with `atomicHelpers`.
- After it's published, add a commit that updates this repo's package.json and npm-shrinkwrap.json.


marteki:
> The charts' titles and footnotes are specific to CCT so they're still in the chart organism (opposed to being generated by cfpb-chart-builder).

This is great, because the designers would rather that the presence of titles and footnotes _not_ be specific to CCT. As part of our draft Bureau data standard visualization guidelines, the following things would be required for every chart:
* Title
* Link to data source file
* Description of the visualization

With a note as a strong suggestion, because there will probably usually be further small things to explain about why the data is constructed in some way. (And this note field could change from release to release depending on changes in the data, so it might be better if it was easier to change by non-developers.)

> Do we need the chart group and report-type fields in Wagtail? Or can they be determined in cfpb-chart-builder (possibly by analyzing the data source's filename)?

If we do through cfpb-chart-builder, we'd need to make sure that everyone who ever uses Chart Block ever always puts in a filename that uses the correct file naming structure expected. How would we manage or correct for that? Would we want to?

cfgov/cfgov/settings/base.py
```diff
@@ -533,4 +533,5 @@
 CSP_CONNECT_SRC = ("'self'",
                    '*.tiles.mapbox.com',
                    'bam.nr-data.net',
+                   'raw.githubusercontent.com/cfpb/consumer-credit-trends/',
```
willbarton: Are these CSP edits still necessary?

contolini: Nope! Smart find.

sebworks:
looks good, other than some CSS stuff. Approach is solid.

cfarm:
> @mistergone @cfarm Do we need the chart group and report-type fields in Wagtail? Or can they be determined in cfpb-chart-builder (possibly by analyzing the data source's filename)?

I think we want to automate this if possible, especially since those fields are so specific to the CCT charts. But if there is any need for an interim step for the January data release where this info gets entered via Wagtail, let's keep it. 

@marteki we could add JS to cfgov-refresh that is specific to CCT for this purpose, instead of including it in chart-builder, or it could be a special util module within chart-builder that only gets included for the CCT page templates (would still require Wagtail template updates here). I'm going to add a task to account for this in our GHE.

cfgov/jinja2/v1/_includes/organisms/chart.html
```diff
@@ -12,25 +12,30 @@
...
-  <p class="block__sub block__border-top block short-desc chart_footnote">
-    {% if value.data_source %}
-        <strong>Source:</strong> <a class="icon-link icon-link__download" href="{{ parse_links(value.data_source) | safe }}"> <span class="icon-link_text">Download CSV file</span> </a><br>
+  <p class="m-chart-footnote block__sub block__border-top block short-desc">
+      {% if value.data_source %}
+      <strong>Source:</strong> <a class="icon-link icon-link__download" href="https://s3.amazonaws.com/files.consumerfinance.gov/data/{{ value.data_source }}"> <span class="icon-link_text">Download CSV file</span> </a><br>
       {% endif %}
 
-        <strong>Date published:</strong> automate this with JS<br>
+      <strong>Date published:</strong> January 2017<br>
```
cfarm: We need to update this footnotes content to reflect the latest from GHE: https://ghe/CFGOV/data-research/issues/543

Date published: January 2017
Note: Data from the last six months is not final. The most recent data available in each visualization is for [month year].

cfarm:
:1st_place_medal:

cfgov/jinja2/v1/_includes/organisms/chart.html
contolini: Cool, once we figure out the correct date I'll hardcode it for this release.

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -663,18 +663,51 @@ class Meta:
...
 class ChartBlock(blocks.StructBlock):
-    element_id = blocks.CharBlock(
+    title = blocks.CharBlock(required=True)
+    # todo: make radio buttons
+    chart_type = blocks.ChoiceBlock(choices=[
+        ('bar', 'Bar'),
+        ('line', 'Line'),
+        ('tile_map', 'Tile Map'),
+    ], required=False)
+    # todo: make radio buttons. required because we need it for the CSS to apply colors to each report page chart
```
cfarm: For report_type field, let's make it more generic and call it color scheme. The options should be green, blue, teal, or navy. Then it can be reused for other projects' charts. We'll update the css to use the new data attribute value too

cfgov/jinja2/v1/_includes/organisms/chart.html
richaagarwal: Should "date published" just be a field in Wagtail for now? That way changing the date (even if manually) won't have to go through a deployment next time

contolini:
Due to the format of the CSVs, I think we need the `group` field in Wagtail. This is unfortunate because it is specific to CCT.

I propose we try to abstract it at least a little bit and call the field `Additional metadata`. It can be optional in wagtail and future non-CCT charts can just ignore it. Or if future charts are similar to CCT and need an extra field, they can use it for whatever they please. Thoughts @cfarm @marteki @hillaryj?

cfarm:
What type of form field would `Additional metadata` be? I like the current choices / select field because it makes it easier to set up; but that requires the choices to be hard coded and specific to the CCT charts for now.

contolini:
> What type of form field would Additional metadata be?

Just making stuff up here but: author, external link, a footnote field.

> I like the current choices / select field because it makes it easier to set up; but that requires the choices to be hard coded and specific to the CCT charts for now.

I don't think we can ever remove the field because doing so will break all CCT charts.

cfgov/jinja2/v1/_includes/organisms/chart.html
contolini: @richaagarwal Not a bad idea but we get new data every month so manually updating 100+ chart blocks would suck. Having to redeploy to update the date will also suck but at least we only have to edit it in one place.

Regardless, next sprint we'll solve the problem by using JS to update these dates so for now I'll just leave it be.

contolini:
@cfarm This is good-to-go from my end. Pull it down and try it in wagtail.

requirements/base.txt
```diff
@@ -20,6 +20,7 @@ django-tinymce==2.2.0
 django-treebeard==3.0
 elasticsearch==1.6.0
 feedparser==5.2.1
+flake8==3.0.4
```
willbarton: flake8 shouldn't be here, it is already in [requirements/test.txt](https://github.com/cfpb/cfgov-refresh/blob/master/requirements/test.txt#L6).

contolini: @richaagarwal's fault! *runs away*

richaagarwal:
Looks great - 
screen shot 2017-01-26 at 3 10 11 pm
Gonna merge

cfgov/unprocessed/css/organisms/chart.less
```diff
@@ -0,0 +1,46 @@
...
+    - cfgov-organisms
+*/
+
+// Import external cfpb-chart-builder stylesheet
+@import (less) 'cfpb-chart-builder.css';
+
+.o-chart {
+
+  // Title above charts
+  h3 {
```
sebworks: These styles are empty. If unneeded, they can probably be removed.

contolini: It looks pretty stupid but I ended up leaving them there because next sprint we'll be styling those elements.

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -664,22 +664,48 @@ class Meta:
 
 
 class ChartBlock(blocks.StructBlock):
-    element_id = blocks.CharBlock(
+    title = blocks.CharBlock(required=True)
+    # todo: make radio buttons
+    chart_type = blocks.ChoiceBlock(choices=[
+        ('bar', 'Bar'),
+        ('line', 'Line'),
+        ('tile_map', 'Tile Map'),
```
cfarm: this should be 'map':

```
('map', 'Tile Map'),
```

contolini: Whoopsie. Would it be easier to change it in `cfpb-chart-builder`? https://github.com/cfpb/cfpb-chart-builder/blob/master/src/index.js#L41

PR #2692, opened 2017-01-24
add missing brace

higs4281:
Typo in css reference

- Missing closing brace


sebworks:
good catch

willbarton:
@higs4281 should this be hotfixed into 4.6?

higs4281:
I have a [related fix](https://github.com/cfpb/college-costs/pull/294) that should probably go with it if we hotfixed. Neither are super urgent, but the PFC feedback pages are functional but wonky without it.

PR #2695, opened 2017-01-24
Make this repo distributable as a wheel

rosskarchner:
This allows us to treat this code base as just another python package, and opens up a lot of possibilities as far as distribution and deployment automation.

It turned out to be a *little* more invasive than I expected, so here are some unexpected changes I needed to make:

- static_built, jinja2, and the top-level templates directories had to be moved into apps (mostly v1, though I put the static files in 'cfgov' for reasons that now escape me)
- in order to get the path of the v1 app at runtime, we have to import it, which was causing some trouble because we do a LOT of stuff in `v1/__init__.py`. I've moved all of that code in to v1/template_environment.py
- There were settings and code tweaks to adapt to the changes described above

## make a wheel

run `python setup.py bdist_wheel` will generate a wheel in the 'dist' directory. 

## use a wheel

This is kind of contrived, but, while in your usual virtualenvironment:
 `pip install dist/cfgov-rest-of-wheel-name.whl` and `cd` into some other directory. Then:

```
export DJANGO_SETTINGS_MODULE=cfgov.settings.test
django-admin runserver
```

and you should be able to browse the site!

At this point, you aren't running the code from your cfgov_refresh checkout, you're running the code that was packaged up into the wheel!

When you're done, be sure to `pip uninstall cfgov` to get your virtualenv back to normal.


rosskarchner:
some of those commits in there don't make sense to me.

rosskarchner:
A note about the wheel file names (and the package versions), they reflect the output of:

`git describe --tags`

sebworks:
Nooooooo, we want to move in this direction. The ideal is to have modular code which can be combined with other modules. 
no

snowflakes: https://www.sciencenews.org/sponsored/classifying-geometry-snowflakes-highlights-their-beauty-and-diversity

rosskarchner:
hmmm---  maybe the right model is the recent work on feature flags: spin things out into proper Python and/or Javascript packages. I don't think the thing that pulls all the other bits together (this project) gains much from being modular.

rosskarchner:
(much as Luke's hand was made modular in the image above)

PR #2703, opened 2017-01-25
Use the latest version of Node v6 in Travis builds

ascott1:
This updates the version of Node used in Travis builds to v6. Odd numbered versions of Node are "progress" versions and do not have a long-term support plan, while even numbered versions are considered stable.

#### More info about Node versioning

- https://github.com/nodejs/LTS
- https://nodejs.org/en/blog/community/v5-to-v7/


sebworks:
@ascott1, please update the CHANGELOG.

Issue #2654, opened 2017-01-05
Stop nuking node_modules


contolini:
Our build script deletes the [project's node_modules directory](https://github.com/cfpb/cfgov-refresh/blob/master/frontend.sh#L29) before re-installing everything:

```sh
# Clean project dependencies.
clean() {
  # If the node directory already exists,
  # clear it so we know we're working with a clean
  # slate of the dependencies listed in package.json.
  if [ -d $NODE_DIR ]; then
    echo 'Removing project dependency directories...'
    rm -rf $NODE_DIR
  fi
  echo 'Project dependencies have been removed.'
}
```

This is good for idempotence but bad for efficiency.

```sh
$ time npm install

real	0m23.881s
user	0m23.155s
sys	0m1.486s

$ rm -rf node_modules/
$ time npm install

real	2m32.014s
user	1m39.444s
sys	0m38.497s
```

It doesn't affect local development a whole lot but it adds minutes to our CI jobs and slows down deployments. How would y'all feel about removing it from our build script?

@cfpb/cfgov-frontends @chosak

marcesher:
💯

Scotchester:
I'm certain that we started doing this intentionally. I don't fully remember why, but are we sure that whatever reason we did it isn't still in play?

chosak:
I like the idea of removing the deletion by default. Would it be possible to add either a command-line parameter to `frontend.sh` (and, I suppose, then `setup.sh`) or an environment variable to retain the option to remove that directory? I'm thinking it might be useful to support a cleaning option in the build scripts, and this would avoid having to refer to `$NODE_DIR` outside of this script.

contolini:
> I'm certain that we started doing this intentionally. I don't fully remember why, but are we sure that whatever reason we did it isn't still in play?

I think it was added to reduce JS compilation errors by ensuring we always have fresh copies of the project's node modules. I would rather manually wipe `node_modules` when I run into errors than do it on every install.

> Would it be possible to add either a command-line parameter to frontend.sh (and, I suppose, then setup.sh) or an environment variable to retain the option to remove that directory?

I'm into this. Something like `./setup.sh --clean`.

sebworks:
> I'm into this. Something like ./setup.sh --clean.

👍

Scotchester:
I think that is a good compromise. One thing we should be sure to watch out for is running that proactively when we remove a dependency.

jimmynotjim:
[...]

PR #2705, opened 2017-01-26
Don't always delete $NODE_DIR

willbarton:
This PR addresses #2654 and implements @rosskarchner's suggestion in GHE#563 to checksum `package.json`, store the checksum in `$NODE_DIR/CHECKSUM`, and then only delete $NODE_DIR if the checksums don't match. This should improve the efficiency of our `frontend.sh` script.

Some quick and dirty metrics on my local machine:

Full install of all dependencies (checksum doesn't exist or doesn't match):

```
./frontend.sh  110.17s user 30.05s system 66% cpu 3:29.80 total
```

When no dependencies have changed (checksum exists and matches):

```
./frontend.sh  55.34s user 2.62s system 155% cpu 37.200 total
```

So, this cuts down the time spent installing frontend dependencies by half unless `package.json` has changed.

## Changes

- Checksum package.json to avoid reinstalling dependencies that haven't changed

## Testing

- Run `frontend.sh` once to generate the checksum file, then run it again and observe the message:

  `Project dependencies checksum matches package.json…`
  
  And that node dependencies aren't reinstalled.

## Notes

I want to get this in front of Travis to make sure I'm not making assumptions about the availability of `shasum`. 

## Checklist

### DIFF
--- a/frontend.sh
+++ b/frontend.sh
@@ -20,7 +20,7 @@ init() {
   echo "npm components directory: $NODE_DIR"
 }
 
-# Clean project dependencies if needed
+# Clean project dependencies if needed.
 clean() {
   # If the node directory exists, $NODE_DIR/CHECKSUM exists, and
   # the contents DO NOT match the checksum of package.json, clear
--- a/frontend.sh
+++ b/frontend.sh
@@ -109,13 +109,11 @@ is_installed() {
 }
 
 # Execute requested (or all) functions.
-if [ "$1" == "init" ] || [ "$1" == "build" ]; then
-  if [ "$1" == "init" ]; then
-    init ""
-    clean_and_install
-  else
-    build
-  fi
+if [ "$1" == "init" ]; then
+  init ""
+  clean_and_install
+elif [ "$1" == "build" ]; then
+  build
 else
   init "$1"
   clean_and_install
--- a/frontend.sh
+++ b/frontend.sh
@@ -15,6 +15,7 @@ init() {
   source cli-flag.sh 'Front end' $1
 
   NODE_DIR=node_modules
+  DEP_CHECKSUM=`cat npm-shrinkwrap.json package.json | shasum -a 256`
 
   echo "npm components directory: $NODE_DIR"
 }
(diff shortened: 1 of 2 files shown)


rosskarchner:
is the npm-shrinkwrap file relevant? Are there circumstances where the shrinkwrap changes but not package.json?

willbarton:
@rosskarchner Oh, good point. I should just use the full command you suggested 😁

sebworks:
> Are there circumstances where the shrinkwrap changes but not package.json?

It should never happen but FEWDs do strange things.

frontend.sh
```diff
@@ -15,25 +15,34 @@ init() {
...
+      echo 'Project dependencies have been removed.'
+    else
+      echo 'Project dependencies checksum matches package.json…'
+    fi
   fi
-  echo 'Project dependencies have been removed.'
 }
 
 # Install project dependencies.
 install() {
```
willbarton: @sebworks if we're not deleting the node_modules directory and package.json and npm-shrinkwrap.json haven't changed is there a need to run `install` here at all? That could also reduce the amount of time spent.

So then we check for the existence and validity of the checksum, and if it passes we just build.

Scotchester: Agreed. If `package.json` and `npm-shrinkwrap.json` haven't changed, no need to `install()`

willbarton: Slicing the `install` step off if the checksums match leaves us with this:

```
./frontend.sh  37.33s user 1.92s system 120% cpu 32.684 total
```

sebworks: very nice, very nice indeed.

frontend.sh
```diff
@@ -94,14 +112,12 @@ is_installed() {
 if [ "$1" == "init" ] || [ "$1" == "build" ]; then
   if [ "$1" == "init" ]; then
     init ""
-    clean
-    install
+    clean_and_install
```
sebworks: @willbarton, how would you feel about:

```

if [ "$1" == "init" ]; then
  init ""
  clean_and_install
elif [ "$1" == "build" ]; then
  build
else
  init "$1"
  clean_and_install
  build
fi
```

willbarton: That works for me.

PR #2710, opened 2017-01-27
Fix for Google Map image of Washington DC on /archive-past-events/ page.

sebworks:
Google changed the location image for [Washington](https://maps.googleapis.com/maps/api/staticmap?center=Washington&size=276x155&zoom=12&scale=2). This PR fixes that by adding the the venue state field to the parameter.

## Changes

- Modified [http://localhost:8000/about-us/events/archive-past-events/](http://localhost:8000/about-us/events/archive-past-events/) to add the venue state field to the location parameter. 

## Testing

- Visit `http://localhost:8000/about-us/events/archive-past-events/`.
- Verify that the correct images are showing for the cities.

## Screenshots
screen shot 2017-01-27 at 9 37 26 am

## Checklist


Scotchester:
Why'd you remove the Note? I laughed :grin:

Scotchester:
But, in regards to that, is there an alternative solution that would get this logic out of the template? Combine the city/state field in the database, perhaps?

higs4281:
👍 works for me locally. great catch!

rosskarchner:
We could move all the logic into python, and have the event page object expose a location_image_url or some such.

Or, more generically, it could be a template function or Jinja tag

rosskarchner:
Another random though, is perhaps we should use Mapbox instead of Google. For one thing, we could style the maps:  https://www.mapbox.com/api-documentation/#static

richaagarwal:
Agree with @rosskarchner that this logic should be moved into python, probably as a function on the model, but I think this should get merged in for the time being since it fixes things currently broken on www.

richaagarwal:
Okay I'm currently working on moving this logic into the back-end. Stay tuned!

rosskarchner:
very cool

sebworks:
thanks all. I'm not suggesting there should be absolutely no logic in the templates. I am however suggesting we strike a better balance.  Should business logic be stored in wagtail page model methods, context functions, or jinja templates? I personally feel most business logic should be stored in the model, or context functions when appropriate (i.e, complex logic required in multiple templates).

I think a healthy approach is described here: http://jinja.pocoo.org/docs/2.9/faq/#isn-t-it-a-terrible-idea-to-put-logic-into-templates

richaagarwal:
Eh, I don't really think "boring and stupid" is a great explanation of why templates might contain logic and I'm surprised it's part of their documentation:
> But templates without any logic mean that you have to do all the processing in the code which is boring and stupid.
That said, I think simple if/else conditionals for including css or javascript or whatever other front-end specific things are fine, but conditionals for database queries or ad-hoc calculations should honestly never be in the template.

PR #2721, opened 2017-01-31
Move es5-shim into deps from dev deps

contolini:
Yay.


willbarton:
leonardo-dicaprio-omg

PR #2724, opened 2017-01-31
Fix npm dependencies and shrinkwrap for production deploys

willbarton:
Our gulp tasks currently require some dependencies that are listed in package.json as `devDependencies`. This is a problem when we run a production install, or try to shrinkwrap a production install. This PR adds those dependencies to `dependencies` in package.json and creates a new production shrinkwrap. It also ensures that optional dependencies won't be installed in production. 

This is needed to successfully deploy in production.

## Changes

- Included build dependencies in package.json

## Testing

```
rm -rf node_modules
./frontend.sh production
```

This should succeed.

## Notes

- We should refactor the build gulp tasks to require only the dependencies they need.

## Checklist


sebworks:
@willbarton, are the `devDependencies` already installed on Jenkins?

package.json
```diff
@@ -33,27 +33,28 @@
...
+    "gulp-mocha": "3.0.1",
     "is-reachable": "1.3.0",
+    "gulp-util": "3.0.8"
+  },
+  "devDependencies": {
+    "chai": "3.5.0",
+    "gulp-plumber": "1.1.0",
     "jasmine-reporters": "2.1.1",
     "jasmine-spec-reporter": "2.4.0",
     "jquery": "1.11.3",
```
sebworks: `jquery` should get moved to dependencies. We still require it for a CDN fallback in `base.html`.

package.json
```diff
@@ -33,27 +33,28 @@
...
-    "gulp-util": "3.0.7",
+    "gulp-coveralls": "0.1.4",
+    "gulp-istanbul": "1.1.1",
+    "gulp-mocha": "3.0.1",
     "is-reachable": "1.3.0",
+    "gulp-util": "3.0.8"
+  },
+  "devDependencies": {
+    "chai": "3.5.0",
+    "gulp-plumber": "1.1.0",
```
sebworks: Doesn't look like we still use `gulp-plumber`.

willbarton:
I will also note, I purposefully left a space between the old dependencies and the new ones so that we can easily differentiate when refactoring.

willbarton:
@sebworks the devDependencies are not already installed on Jenkins.

willbarton:
Confirmed the latest change work in production Jenkins as well.

sebworks:
> @sebworks the devDependencies are not already installed on Jenkins in production.

ok, I just wasn't sure if we wanted to run any tests on Jenkins.  The way it's set up now all tests run on Travis.

willbarton:
@sebworks Yeah. We should set it up to run tests on Jenkins, but we can create a job that works the same way as Travis. It would be separate from the build job though.

PR #2734, opened 2017-02-07
Make Resource snippets orderable and update Snippet Lists to sort by order, then title

Scotchester:
Snippet list output has previously been in a random order. This sets the default order to alphabetical by title, and enables the ability to order them manually by using the Resource snippet's `order` field.

## Additions

- Adds `order` field to Resource snippet model

## Changes

- Uses the `order` field to sort the output of a Snippet List module, defaulting to (and breaking ties with) alphabetical by title.
- Updates `help_text` for Resource's `tags` field.

## Testing

1. Pull fork
1. Create some Resource snippets with various titles, if you don't have any yet. Don't give them an order value, but do give them a tag you can use in the Snippet List.
1. Create a Snippet List on a browse page using the name of the tag you gave to the snippets in the above step.
1. View the page and confirm that they are sorting alphabetically by title.
1. Go back and edit some of the snippets to give them a numeric order value.
1. Refresh the page and confirm that the ones you gave an order to appear below the ones with no order, and they appear in numeric order.

## To do

- As I write this, it occurs to me that I should test this with a Snippet List set to use Contacts, to ensure it doesn't break because Contacts don't have an order field.

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/snippet-list.html
+++ b/cfgov/jinja2/v1/_includes/organisms/snippet-list.html
@@ -55,7 +55,7 @@
     } ) }}
 
     {% set snippets = get_snippets(value.snippet_type) %}
-    {% set filtered_snippets = snippets.objects.filter_by_tags(value.tags).order_by('order', 'title') %}
+    {% set filtered_snippets = snippets.objects.filter_by_tags(value.tags) if snippets.objects.filter_by_tags else snippets.objects.all() %}
     {% if filtered_snippets.exists() %}
         <div class="o-table__stack-on-small o-snippet-list_list" data-display-table="table">
             <header data-display-table="thead">
@@ -66,7 +66,7 @@
             {% for snippet in filtered_snippets %}
                 <li data-display-table="row">
                     <div data-display-table="cell">
-                        <div class="o-snippet-list_list-item-title">{{ snippet.title }}</div>
+                        <div class="o-snippet-list_list-item-title">{{ snippet }}</div>
                         {{ snippet.desc if snippet.desc else '' }}
                     </div>
                     <div data-display-table="cell">
--- a/cfgov/jinja2/v1/_includes/organisms/snippet-list.html
+++ b/cfgov/jinja2/v1/_includes/organisms/snippet-list.html
@@ -55,7 +55,7 @@
     } ) }}
 
     {% set snippets = get_snippets(value.snippet_type) %}
-    {% set filtered_snippets = snippets.objects.filter_by_tags(value.tags) %}
+    {% set filtered_snippets = snippets.objects.filter_by_tags(value.tags).order_by('order', 'title') %}
     {% if filtered_snippets.exists() %}
         <div class="o-table__stack-on-small o-snippet-list_list" data-display-table="table">
             <header data-display-table="thead">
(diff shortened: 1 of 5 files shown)


Scotchester:
> As I write this, it occurs to me that I should test this with a Snippet List set to use Contacts, to ensure it doesn't break because Contacts don't have an order field.

Well, it was already broken if you tried to use Contacts. Oops.

screen shot 2017-02-07 at 18 02 30

I'll address it tomorrow.

cfgov/jinja2/v1/_includes/organisms/snippet-list.html
```diff
@@ -55,7 +55,7 @@
     } ) }}
 
     {% set snippets = get_snippets(value.snippet_type) %}
-    {% set filtered_snippets = snippets.objects.filter_by_tags(value.tags) %}
+    {% set filtered_snippets = snippets.objects.filter_by_tags(value.tags).order_by('order', 'title') %}
```
chosak: One thought: instead of putting this into the SnippetList template, you could attach this to the model as its default ordering, e.g.

```py
class Resource(ClusterableModel):
    ...

    class Meta:
        ordering = ('order', 'title')
```

This would make any queries to this model (like `Resource.objects.all()`, `Resource.objects.filter(...)`) automatically use this ordering by default. It'd also solve your problem with `order` not existing on the `Contact` snippet, because you wouldn't need to add it there.

Scotchester:
I definitely like the suggestion to move it to the model and test that! Thanks, @chosak.

Scotchester:
Pushed an update to do the ordering in-model, per @chosak's suggestion, and adds some tests for that. **Not ready for merge until migration is updated**, which will happen after we pair on potentially taking the filtering logic out of the template.

Scotchester:
Putting a hold on this for a few days while @chosak investigates CFGOV/platform/issues/378, which would enable the passing of the returned snippets in the context and simplifying the template logic.

Scotchester:
Hold removed to get this out the door. The work required by platform#378 will take a bit longer than we hoped.

cfgov/jinja2/v1/_includes/organisms/snippet-list.html
```diff
@@ -55,7 +55,7 @@
     } ) }}
 
     {% set snippets = get_snippets(value.snippet_type) %}
-    {% set filtered_snippets = snippets.objects.filter_by_tags(value.tags) %}
+    {% set filtered_snippets = snippets.objects.filter_by_tags(value.tags) if snippets.objects.filter_by_tags else snippets.objects.all() %}
```
jimmynotjim: Can this be broken at or before 80 characters?

chosak: For posterity (and as we discussed), it'd be better if the logic that defines the `snippets` variable was part of the block Python code. Unfortunately this would be a bigger change so this is a good way to implement this for now.

Scotchester: Why yes, it turns out that Jinja statements CAN be broken onto multiple lines! Fixed in  	d7602b5.

Scotchester:
Merging when Travis approves.

PR #2735, opened 2017-02-08
related posts refactor/tests

richaagarwal:
## What is this 
This is a fairly large refactor of Related Posts, motivated by a desire to simplify some of the logic in `def related_posts` (in `v1/models/base.py`) and spiraled into having to update other things, such as `ref.py`. A couple things of note: 
- There is a new field on CFGOVPage -- `related_posts_categories`, derived from a new model `RelatedPostsCategory`. It's the same as `CFGOVPageCategory` in code and behavior. It's meant to be used in place of `specific_categories` on a page with a Related Posts module, but I will follow up with @sarahsimpson09 to see if this is workable.
  - The motivation behind making it a model, as opposed to a list in a Wagtail block, is that it should be easy to update these categories on the back-end, without relying on the publishing flow to see the changes.  Other work in this PR would've necessitated re-publishing many of the pages if not for this migration,  which isn't really an option given that we rely heavily on `date_published` for determining which content to display first or at all.

## How to test
- Pull down the branch, grab the latest production db and import it, run migrations 
- After running migrations, run `python cfgov/manage.py runscript update_categories` to fill in the new `related_posts_categories` with data from `specific_categories`, which we will want to remove eventually
- Check a few pages with related posts modules, e.g. [http://localhost:8000/about-us/blog/what-do-if-youre-wrongfully-billed-medicare-costs/](http://localhost:8000/about-us/blog/what-do-if-youre-wrongfully-billed-medicare-costs/) to confirm they look like what's on www (https://www.consumerfinance.gov/about-us/blog/what-do-if-youre-wrongfully-billed-medicare-costs/)

## TODO
- Update old, failing tests
- Actually write the new tests


cfgov/v1/models/__init__.py
```diff
@@ -1,17 +1,17 @@
...
+from v1.models.landing_page import *
+from v1.models.sublanding_page import *
+from v1.models.browse_page import *
+from v1.models.browse_filterable_page import *
+from v1.models.learn_page import *
+from v1.models.home_page import *
+from v1.models.sublanding_filterable_page import *
+from v1.models.blog_page import *
+from v1.models.newsroom_page import *
+from v1.models.images import *
```
richaagarwal: this was bugging me but is unrelated to related posts

cfgov/v1/models/base.py
```diff
@@ -109,6 +109,10 @@ class CFGOVPage(Page):
     # Panels
     sidefoot_panels = [
         StreamFieldPanel('sidefoot'),
+        InlinePanel(
+            'related_posts_categories',
+            label="Related Posts Categories"
+        )
```
richaagarwal: The placement of this is not ideal, however there is no real alternative since putting it inside the Related Posts module itself is not possible

cfgov/scripts/update_categories.py
```diff
@@ -0,0 +1,52 @@
...
+                    elif cat == 'At the CFPB':
+                        new_cat = RelatedPostsCategory(
+                            page=page,
+                            name='at-the-cfpb'
+                        )
+                        new_cat.save()
+                    elif cat == 'Press Release':
+                        new_cat = RelatedPostsCategory(
+                            page=page,
+                            name='press-release'
```
richaagarwal: Last I checked there were no op-ed, speech, or testimony categories in prod data, but I'll add them in here anyway to be comprehensive.

richaagarwal:
There are a couple things I need to do still, + update tests, but I'd love to get feedback on this approach in the meantime.

cfgov/scripts/update_categories.py
Scotchester: https://www.consumerfinance.gov/about-us/newsroom/?form-id=0&filter0_title=&filter0_categories=op-ed&filter0_categories=speech&filter0_categories=testimony&filter0_from_date=&filter0_to_date=

cfgov/v1/models/__init__.py
Scotchester: :+1: Could we sort them, too?

cfgov/v1/util/ref.py
```diff
@@ -126,6 +111,11 @@
     )),
 ]
 
+related_posts_categories = [
+    categories[2],
+    categories[7]
+]
```
Scotchester: Clever!

Scotchester:
> The motivation behind making it a model, as opposed to a list in a Wagtail block, is that it should be easy to update these categories on the back-end, without relying on the publishing flow to see the changes.

I don't love the idea of disconnecting this field from the Related Posts block, and I don't fully understand the motivation you gave. Could you elaborate a little?

cfgov/scripts/update_categories.py
richaagarwal: @Scotchester For related posts modules

richaagarwal:
@Scotchester Sure - basically, there's no way to reference another model inside the Related Posts Block -- on the face of it, it looks like you can do it in a `ChoiceBlock`, but you'll eventually run into issues with Django not being able to serialize those values (https://docs.djangoproject.com/en/1.8/topics/migrations/#migration-serializing) when it comes time to migrations. As far as I can tell, the `InlinePanel` is the only way to model this relation -- http://docs.wagtail.io/en/v1.9/reference/pages/panels.html#inline-panels, which is not something you can define inside a `StructBlock`.

cfgov/v1/models/__init__.py
richaagarwal: Sure!

richaagarwal:
After talking to @chosak I'm going to do a little bit more exploratory work so that it doesn't have to be outside the Related Posts block, but it will result in a lot more changes since it would involve removing `ref.py` altogether.

richaagarwal:
Closing for now

rosskarchner:
Back in the sheer days, we used Elasticsearch's "more like this" to provide related content. I wonder if we can sidestep this complexity by getting serious about indexing wagtail content in ES, and going that direction.

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-mlt-query.html

Scotchester:
@richaagarwal I don't see an issue with reducing some of `ref.py` but not all of it. Different things in there have been added at different times for different reasons, and are not all related. If you can pull some stuff out of there to aid in your refactor of Related Posts, in a way that doesn't break anything else, but leave the rest of `ref.py` to be addressed later, I don't have a problem with that.

Perhaps you can elaborate on what you mean by "I think that will cause a lot of confusion"?

PR #2736, opened 2017-02-08
Adding Babel to frontend workflow; Upgrading Webpack to version 2.

sebworks:
Adding babel to frontend workflow; Upgrading Webpack to version 2.

## Additions

- Added `.babelrc` as a babel Configuration file. Using the `es2015` and  `stage-2` presets. Presets are just sets of transformations. 

## Changes

- Modified `config/webpack-config.js` to do the following:
   - Use the Babel loader
   - Enable the cache property for performance purposes. 
   - Delete the second chunk entry which seems no longer necessary in Webpack 2. 

 - Modified `gulp/tasks/scripts.js` to pass `webpack-stream` a Webpack param. 

 - Updated `package.json` to do the following:
   - Add the necessary modules to support Babel.
   - Remove `slick-carousel` as it's no longer needed.
   - Move `snyk` as it's not necessary to run on Jenkins.

## Testing

- Run `rm -rf node_modules`
- Run `frontend.sh`.
- Browse the site as expected.
- Run `rm -rf node_modules`
- Run `frontend.sh production`.
- Browse the site as expected.

## Screenshots
Before:

screen shot 2017-02-08 at 12 26 39 pm

After:

screen shot 2017-02-08 at 11 48 54 am

## Notes

- The key performance bottleneck is the `webpack.optimize.UglifyJsPlugin`.  There are alternatives which uglify in [parallel](https://www.npmjs.com/package/webpack-uglify-parallel), we should investigate if they are worth using.  
- This is an interesting article: https://github.com/lcxfs1991/blog/issues/15.

## Todos

- Converting some modules over to ES6.

## Checklist


.babelrc
```diff
@@ -0,0 +1,9 @@
+{
+  "presets": ["es2015", "stage-2"],
```
sebworks: >> Why'd you choose stage 2?

It was just and arbitrary starting point. We can adjust if we find language features we want to use but find missing. The process is quite interesting, as it appears features are moving in / out of stages based on the TC39 process. https://github.com/tc39/proposals.

ascott1:
@sebworks along with this PR, we should add ES6 guidelines to our [JS standards](https://github.com/cfpb/development/blob/master/javascript.md)

.babelrc
contolini: Works for me. 👍

package.json
```diff
@@ -69,6 +71,7 @@
     "psi": "2.0.4",
     "selenium-webdriver": "2.53.3",
     "sinon": "1.17.3",
+    "snyk": "1.25",
```
contolini: `1.25` -> `1.25.0`

sebworks:
> @sebworks along with this PR, we should add ES6 guidelines to our JS standards.

ok, I will work on this.

anselmbradford:
Will try to review this today!

config/webpack-config.js
```diff
@@ -19,17 +19,26 @@ var modernConf = {
...
+  },
   output: {
-    path:     path.join( __dirname, 'js' ),
+    path: path.join( __dirname, 'js' ),
     filename: '[name]'
   },
   plugins: [
-    new webpack.optimize.CommonsChunkPlugin( COMMON_BUNDLE_NAME ),
-    new webpack.optimize.CommonsChunkPlugin( COMMON_BUNDLE_NAME,
-                                             [ COMMON_BUNDLE_NAME ] ),
```
anselmbradford: IIRC the redundant commons chunk calls here were for the scenario where all included scripts included a reference to a module EXCEPT common.js. In that case, since all scripts reference a module, the module's declaration should be moved to common.js automatically by webpack, instead of leaving them redundantly in each script. I tested this behavior in the old setup and the new setup and it still works as expected 👍

anselmbradford:
LGTM 👍 nice work

config/webpack-config.js
sebworks: You are correct but in order for that to work you need to set the `minChunks` property. The default is infinity. You can test this by setting `minChunks: 3`  and running `gulp build`. It should increase the size of  `common.js` and decrease the size of each of individual modules. That decrease is due to the `standardType` being added to `common.js` as it's required by atleast 3 modules. Do you think we should set the minChunks property?

PR #2744, opened 2017-02-13
Use Yarn instead of npm

contolini:
Yarn simplifies aspects of Node dependency management, notably dependency version locking (`npm-shrinkwrap.json` with npm, `yarn.lock` with yarn). Shrinkwrapping has proven painful for us and our CI environments.

## Additions

- Yarn's lock file.

## Removals

- npm's shrinkwrap file.

## Changes

- All npm install commands replaced with yarn.
- Updated Travis' nvm version and told it to use yarn.

## Testing

- Pull down `yarn` branch.
- `./setup.sh`
- Confirm that there are no console errors and the site works locally as expected.


sebworks:
What's the deal on this?

anselmbradford:
Now that package-lock is in, do we still think we'll go a yarn route some day?

sebworks:
@anselmbradford, yeah, I don't think we need it anymore.

anselmbradford:
@contolini

PR #2751, opened 2017-02-15
enable USE_X_FORWARDED_HOST

rosskarchner:
This is basically an extension of #2747

On beta, I've got the X-Forwarded-Proto header working correctly, *mostly*. The (small) problem is that URL's include the port, which seems is kind of sloppy.

For example, the og:url on a blog post looks like:

https://beta.consumerfinance.gov:443/about-us/blog/cfpb-ombudsmans-office-2016-annual-report/

This is because (for whatever reason) when the CDN requests documents from the origin, the Host: header contains the port number, like beta.consumerfinance.gov:443

By enabling USE_X_FORWARDED_HOST, we can configure the CDN to send a X-Forwarded-Host header with our chose value (which will just be beta.consumerfinance.gov and www.consumerfinance.gov), and Django will respect that when generating URL's

The relevant Django logic is here:

https://github.com/django/django/blob/1.8.15/django/http/request.py#L76

## Additions

- Set settings.USE_X_FORWARDED_HOST to True

## Checklist


rosskarchner:
point well taken!

PR #2755, opened 2017-02-21
update latest data month from november to december

cfarm:
Updates hard-coded date for the CCT monthly data sets. The footnote for each graph should say "December 2016" once the latest date goes live (slated for Tuesday 2/21).

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/chart.html
+++ b/cfgov/jinja2/v1/_includes/organisms/chart.html
@@ -36,7 +36,7 @@
       <strong>Date published:</strong> January 2017<br>
 
       {% if value.note %}
-      <strong>Note:</strong> {{ value.note }} The most recent data available in each visualization is for November 2016.
+      <strong>Note:</strong> {{ value.note }} The most recent data available in each visualization is for December 2016.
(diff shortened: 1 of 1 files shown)


cfgov/jinja2/v1/_includes/organisms/chart.html
```diff
@@ -36,7 +36,7 @@
       <strong>Date published:</strong> January 2017<br>
```
marteki: Do we need to change this line to February 2017?

marteki:
👍  after the additional change I committed.

mistergone:
👍

sebworks:
@marteki, does it make sense to make these fields editable in Wagtail?

marteki:
@sebworks This was a short-term fix (that's lasting longer than we expected). In the very near future, cfpb-chart-builder will handle updating these lines, and I believe the fields will be removed.

@cfarm @mistergone @contolini Can you confirm that this is the intent of the front-end devs who've worked on this?

cfarm:
@marteki that's correct - this will be automated instead of edited via Wagtail or code

PR #456, opened 2015-04-06
DF-1436_Privacy_page

sebworks:
Include privacy statement next to email sign-up fields 
## Changes
 
- Updated ' _includes/email-subscribe-form.html' to include privacy act copy.
- Added  'privacy-policy/_vars-privacy-policy.html' to include privacy navigation.
- Added 'privacy-policy/index.html' to add privacy page.

## Notes
- This hasn't been seen or vetted by the design team so will need to show in Design Review for approval.

## Review
 
- @jimmynotjim  
- @anselmbradford 
- @duelj 
- @juidai 

## Preview
 
screen shot 2015-04-06 at 10 03 14 am

screen shot 2015-04-06 at 10 02 59 am

screen shot 2015-04-06 at 10 02 41 am

PR #2780, opened 2017-03-07
First pass at migrating Ask CFPB content

higs4281:
This focuses on simplifying knowledgebase data models and migrating
the content into Wagtail. The parent question-categories that we're using
as portal-page topics are broken out as categories, each having a set
of subcategories, for organizing answers and pages.
The core content bits -- questions and answers in two languages --
are consolidated into Answer objects that carry the English and Spanish
forms of a question, its answer, its slug, and metadata.

## Workflow
- An Answer object is where content editing is done. 
- Wagtail pages are created or updated based on workflow checkboxes for each language. If a box is checked, the relevant page will be updated with a draft status.
- Each language also has its own manual `last edited` field, intended to give us a more useful timestamp to track  editorial updates. Editors choose when an to mark an answer as updated, so minor metadata changes (or, say, a migration) won't pollute these values.
- Pages then follow our standard Wagtail review and approval workflow.
- The original Ask IDs get migrated to `ask_cfpb` as Answer IDs, which can be handy for searching both answers and pages. The answer page titles carry the ID as a prefix with a colon, so searching the Wagtail admin for "1979:" will find just the English and Spanish pages associated with that answer. 

## Additions

- ask_cfpb app

## Removals

- None. Knowledgebase remains a satellite app for now.

## Changes

- `settings/base.py` changed to add ask_cfpb to INSTALLED_APP

## Testing

These steps will get the current data migrated into the new models in Wagtail.
- create new db, load with recent prod data, run `migrate` and fire up runserver
```
./drop-db.sh
./create-mysql-db.sh
mysql -uroot v1 < ~/Downloads/production_django.sql
./runserver.sh
```

- Create a landing page under CFGov with title "Ask CFPB" and publish it to www
- If you want to view live and draft pages, edit your sites so default is localhost/port 8000
- Optionally add a sharing site for the default site
- in another terminal, open shell and migrate the knowledgebase data.

```
python cfgov/manage.py shell
```

```python
from ask_cfpb.migration_scripts.migrate_knowledgebase import migrate_knowledgebase
migrate_knowledgebase()
```

This take a few minutes to migrate the knowledgebase data into the new models and generate 2K+  English and Spanish pages.

You now should see the data tables in the Wagtail admin left sidebar under "Ask CFPB," and the pages should be live under `CFGov > Ask CFPB >` 

## Screenshots

ask_in_wagtail

Try editing an Answer, selecting workflow update box and saving to update a page.

### DIFF
--- a/cfgov/ask_cfpb/migration_scripts/migrate_knowledgebase.py
+++ b/cfgov/ask_cfpb/migration_scripts/migrate_knowledgebase.py
@@ -1,0 +1,279 @@
+import time
+
+from knowledgebase.models import QuestionCategory as QC
+from knowledgebase.models import Question, Audience, UpsellItem, EnglishAnswer
+from ask_cfpb.models import (
+    Answer,
+    # AnswerPage,
+    Category,
+    NextStep,
+    SubCategory)
+from ask_cfpb.models import Audience as ASK_audience
+
+
+def get_kb_statuses(ask_id):
+    q = Question.objects.get(id=ask_id)
+    return {'en': (q.get_english_answer_state() == 'APPROVED'),
+            'es': (q.get_spanish_answer_state() == 'APPROVED')}
+
+
+def create_answer_pages(queryset):
+    counter = 0
+    if queryset:
+        for answer in queryset:
+            kb_statuses = get_kb_statuses(answer.id)
+            for lang in ['en', 'es']:
+                if kb_statuses[lang]:
+                    page = answer.create_or_update_page(language=lang)
+                    revision = page.get_latest_revision()
+                    revision.publish()
+                    counter += 1
+    else:
+        print "No Answer objects found in queryset."
+    print("\nCreated and publishd {} legacy Answer pages".format(counter))
+
+
+def create_pages():
+    create_answer_pages(Answer.objects.all())
+
+
+def migrate_categories():
+    """Move parent QuestionCategories into the new Category model"""
+    topics = QC.objects.filter(parent=None)
+    create_count = 0
+    update_count = 0
+    for qc in topics:
+        cat, cr = Category.objects.get_or_create(
+            slug=qc.slug,
+            slug_es=qc.slug_es,
+            name=qc.name,
+            name_es=qc.name_es,
+            intro=qc.description,
+            intro_es='')
+        if cr:
+            create_count += 1
+        else:
+            update_count += 1
+    print("Found {} topic categories, "
+          "created {} "
+          "and updated {}".format(
+              topics.count(), create_count, update_count))
+
+
+def get_en_answer(question, cats, en_answer):
+    answer, created = Answer.objects.get_or_create(
+        id=question.id,
+        slug=en_answer.slug)
+    for cat in cats:
+        answer.category.add(cat)
+    answer.last_edited = question.updated_at.date()
+    answer.question = en_answer.title
+    return answer
+
+
+def get_es_answer(question, cats, es_answer):
+    answer, created = Answer.objects.get_or_create(
+        id=question.id,
+        slug_es=es_answer.slug)
+    for cat in cats:
+        answer.category.add(cat)
+    answer.created_at = question.created_at
+    answer.last_edited_es = question.updated_at.date()
+    answer.question_es = es_answer.title
+    return answer
+
+
+def fill_out_es_answer(question, answer, es_answer):
+    answer.slug_es = es_answer.slug
+    answer.question_es = es_answer.title
+    answer.answer_es = es_answer.answer
+    answer.last_edited_es = question.updated_at.date()
+    answer.snippet_es = u''
+    return answer
+
+
+def build_answer(question, cats, en_answer=None, es_answer=None):
+    if en_answer:
+        answer_base = get_en_answer(question, cats, en_answer)
+        answer_base.answer = en_answer.answer
+        answer_base.last_edited = question.updated_at.date()
+        answer_base.snippet = en_answer.one_sentence_answer
+        for cat in cats:
+            answer_base.category.add(cat)
+        for qc in question.question_category.exclude(parent=None):
+            answer_base.subcategory.add(
+                SubCategory.objects.get(slug=qc.slug))
+        if es_answer:
+            answer_base = fill_out_es_answer(question, answer_base, es_answer)
+    elif es_answer:  # no English answer to be had
+        answer_base = get_es_answer(question, cats, es_answer)
+        answer_base = fill_out_es_answer(question, answer_base, es_answer)
+    else:  # no answers for this question yet, so we'll save with an ID slug
+        answer_base, created = Answer.objects.get_or_create(
+            id=question.id,
+            slug="answer-{}".format(question.id))
+        for cat in cats:
+            answer_base.category.add(cat)
+    return answer_base
+
+
+def migrate_answer(question):
+    parents = question.question_category.filter(parent=None)
+    cats = Category.objects.filter(
+        slug__in=[parent.slug for parent in parents])
+    answer = build_answer(
+        question,
+        cats,
+        question.get_english_answer(),
+        question.get_spanish_answer())
+    if answer:
+        answer.save()
+
+
+def migrate_questions():
+    queryset = Question.objects.all()
+    print("Migrating {} KB questions ...".format(queryset.count()))
+    counter = 0
+    for question in queryset:
+        counter += 1
+        if counter % 100 == 0:
+            print "{}".format(counter)
+        migrate_answer(question)
+    print "Migrated or updated {} answers.".format(counter)
+
+
+def migrate_audiences():
+    audiences_created = 0
+    audience_relation_count = 0
+    for audience in Audience.objects.all():
+        new_audience, cr = ASK_audience.objects.get_or_create(
+            id=audience.id,
+            name=audience.name)
+        if cr:
+            audiences_created += 1
+    for question in Question.objects.all():
+        ask_cfpb_answer = Answer.objects.get(id=question.id)
+        for audience in question.audiences.all():
+            ask_cfpb_answer.audiences.add(
+                ASK_audience.objects.get(id=audience.id))
+            audience_relation_count += 1
+    print("Found {} Audience objects; needed to migrate {}\n"
+          "Created {} Audience links".format(
+              Audience.objects.count(),
+              audiences_created,
+              audience_relation_count))
+
+
+def migrate_next_steps():
+    """Move knowledgebase UpsellItems to ask_cfpb NextSteps"""
+    upsells_created = 0
+    upsells_updated = 0
+    nextsteps_linked = 0
+    upsells = UpsellItem.objects.all()
+    print("Migrating {} UpsellItems to the NextStep model ...".format(
+        upsells.count()))
+    for upsell in upsells:
+        nextstep, cr = NextStep.objects.get_or_create(
+            id=upsell.id,
+            title=upsell.title)
+        nextstep.show_title = upsell.show_title
+        nextstep.text = upsell.text
+        nextstep.save()
+        if cr:
+            upsells_created += 1
+        else:
+            upsells_updated += 1
+    print("Created {} NextStep objects "
+          "and updated {}".format(upsells_created, upsells_updated))
+    for ea in EnglishAnswer.objects.all():
+        if ea.upsellitem:
+            answer = Answer.objects.get(id=ea.question.id)
+            answer.next_step = NextStep.objects.get(id=ea.upsellitem.id)
+            answer.save()
+            nextsteps_linked += 1
+    print("Created {} NextStep links".format(nextsteps_linked))
+
+
+def migrate_subcategories():
+    """Put QuestionCategories that aren't parents into new SubCategory Model"""
+    subcategories = QC.objects.exclude(parent=None)
+    create_count = 0
+    update_count = 0
+    for qc in subcategories:
+        subcat, cr = SubCategory.objects.get_or_create(
+            id=qc.id,
+            name=qc.name,
+            name_es=qc.name_es,
+            slug=qc.slug,
+            slug_es=qc.slug_es)
+        subcat.featured = qc.featured
+        subcat.weight = qc.weight
+        subcat.description = qc.description
+        subcat.more_info = qc.more_info
+        subcat.parent = Category.objects.get(slug=qc.parent.slug)
+        subcat.save()
+        if cr:
+            create_count += 1
+        else:
+            update_count += 1
+    print("Created {} ASK categories "
+          "and updated {}".format(create_count, update_count))
+
+
+def add_related_categories():
+    update_count = 0
+    for cat in QC.objects.exclude(parent=None):
+        subcategory = SubCategory.objects.get(id=cat.id)
+        for related in cat.related_subcategories.exclude(parent=None):
+            subcategory.related_subcategories.add(
+                SubCategory.objects.get(id=related.id))
+        subcategory.save()
+        update_count += 1
+    print("Updated related categories for {} ASK categories.").format(
+        update_count)
+
+
+def add_featured_questions():
+    update_count = 0
+    for cat in QC.objects.exclude(parent=None):
+        subcategory = SubCategory.objects.get(id=cat.id)
+        for featured in cat.featured_questions.all():
+            subcategory.featured_questions.add(
+                Answer.objects.get(id=featured.id))
+        subcategory.save()
+        update_count += 1
+    print("Updated featured questions for {} ASK categories.").format(
+        update_count)
+
+
+def add_related_questions():
+    update_count = 0
+    print "Adding related_question links ..."
+    for q in Question.objects.exclude(related_questions=None):
+        answer = Answer.objects.get(id=q.id)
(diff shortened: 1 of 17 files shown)


cfgov/flags/migrations/0001_initial.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
```
willbarton: This file should be removed and replaced with [wagtail-flags](https://github.com/cfpb/wagtail-flags).

rosskarchner:
Not sure if this is interesting, but I got some warnings during the migration, like:

```
/Users/karchnerr/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/django/db/backends/mysql/base.py:124: Warning: Incorrect string value: '\xE2\x80\x8B\xE2\x80\x8B...' for column 'answer_es' at row 1
  return self.cursor.execute(query, args)
```

higs4281:
thanks, @rosskarchner 
I haven't seen that, but they look like utf-8 zero-width spaces. 
I should probably .strip() all the entry fields.

higs4281:
@rosskarchner could mysql version be a factor? I have 5.7.13 locally, but servers are 5.1.45.

rosskarchner:
Maybe? I am using 5.1 locally (via docker)

higs4281:
@rosskarchner or this?

### On BUILD

```
mysql> show variables like 'char%';
+--------------------------+--------------------------------------------------------------+
| Variable_name            | Value                                                        |
+--------------------------+--------------------------------------------------------------+
| character_set_client     | utf8                                                         |
| character_set_connection | utf8                                                         |
| character_set_database   | latin1                                                       |
| character_set_filesystem | binary                                                       |
| character_set_results    | utf8                                                         |
| character_set_server     | latin1                                                       |
| character_set_system     | utf8                                                         |
| character_sets_dir       | /usr/local/mysql-5.1.45-linux-x86_64-glibc23/share/charsets/ |
+--------------------------+--------------------------------------------------------------+

mysql> show variables like "%collation%";
+----------------------+-------------------+
| Variable_name        | Value             |
+----------------------+-------------------+
| collation_connection | utf8_general_ci   |
| collation_database   | latin1_swedish_ci |
| collation_server     | latin1_swedish_ci |
+----------------------+-------------------+
```


### On laptop

```
mysql> show variables like 'char%';
+--------------------------+--------------------------------------------------------------------+
| Variable_name            | Value                                                              |
+--------------------------+--------------------------------------------------------------------+
| character_set_client     | utf8                                                               |
| character_set_connection | utf8                                                               |
| character_set_database   | utf8                                                               |
| character_set_filesystem | binary                                                             |
| character_set_results    | utf8                                                               |
| character_set_server     | utf8                                                               |
| character_set_system     | utf8                                                               |
| character_sets_dir       | /Users/higginsw/homebrew/Cellar/mysql/5.7.13/share/mysql/charsets/ |
+--------------------------+--------------------------------------------------------------------+

mysql> show variables like "%collation%";
+----------------------+-----------------+
| Variable_name        | Value           |
+----------------------+-----------------+
| collation_connection | utf8_general_ci |
| collation_database   | utf8_general_ci |
| collation_server     | utf8_general_ci |
+----------------------+-----------------+
```

chosak:
@higs4281 I'd like to do a more thorough review of this PR, but related to the DB comments above, see also GHE#190.

cfgov/ask_cfpb/models/django.py
```diff
@@ -0,0 +1,321 @@
...
+            _question = self.question_es
+            _snippet = self.snippet_es
+            _answer = self.answer_es
+        else:
+            raise Exception('unsupported language: "{}"'.format(language))
+        base_page = get_or_create_page(
+            apps,
+            'ask_cfpb',
+            'AnswerPage',
+            '{}: {}'.format(self.id, _question[:248]),
```
chosak: This change puts the answer number into the visible title, e.g.

image

Just to confirm: is this desired behavior?

cfgov/ask_cfpb/models/django.py
```diff
@@ -0,0 +1,321 @@
...
+            _question = self.question
+            _snippet = self.snippet
+            _answer = self.answer
+        elif language == 'es':
+            _slug = self.slug_es
+            _question = self.question_es
+            _snippet = self.snippet_es
+            _answer = self.answer_es
+        else:
+            raise Exception('unsupported language: "{}"'.format(language))
```
chosak: Maybe use `ValueError` here as a more appropriate specific error type?

cfgov/ask_cfpb/models/pages.py
```diff
@@ -0,0 +1,56 @@
...
+    snippet = RichTextField(
+        blank=True, help_text='Optional answer intro', editable=False)
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+    publish_date = models.DateTimeField(default=timezone.now)
+    answer_base = models.ForeignKey(
+        Answer,
+        blank=True,
+        null=True,
+        on_delete=models.PROTECT)
```
chosak: It might be nice to have a `related_name='answer_pages'` here so that you can do `answer.answer_pages` instead of needing `answer.answerpage_set`.

cfgov/ask_cfpb/migration_scripts/migrate_knowledgebase.py
```diff
@@ -0,0 +1,281 @@
...
+def create_answer_pages(queryset):
+    counter = 0
+    if queryset:
+        for answer in queryset:
+            kb_statuses = get_kb_statuses(answer.id)
+            for lang in ['en', 'es']:
+                if kb_statuses[lang]:
+                    page = answer.create_or_update_page(language=lang)
+                    revision = page.get_latest_revision()
+                    revision.publish()
```
rosskarchner: Interesting thought... Not sure what to do with that

cfgov/ask_cfpb/models/django.py
higs4281: Yes. It's currently visible in the URL. This would be helpful for content editors.

Scotchester: Hmm... I think it's OK for it to be visible in the URL, but having it in the `<title>` seems like a mistake. It's not useful information for the user.

Scotchester: I would actually request that the number DOES remain in the URL, because it makes writing [redirects for Ask CFPB pages](https://github.cfpb.gov/cfpb-infra/ansible-consumer-finance-frontend/blob/master/templates/redirects.conf.j2#L774-L800) (which happens _frequently_ and in great numbers) a lot easier.

cfgov/ask_cfpb/migration_scripts/migrate_knowledgebase.py
higs4281: @chosak  Re creating live pages, we talked about a feature-flag to switch to the new page sets.- @rosskarchner Would it be reasonable to add an optional argument to revision.publish() that skips the flush?

rosskarchner: revision.publish isn't our method, it's via wagtail. We might need figure out another way.

cfgov/ask_cfpb/models/django.py
higs4281: Yes!

cfgov/ask_cfpb/models/django.py
rosskarchner: Just to make sure the original intent isn't lost, in the current code this acts as a sort of "self healing" mechanism. If the title changes, then the slug also changes. Having the numbers in the URL allowed us to catch that and redirect user's to the updated URL.

GHE/CFPB/knowledgebase/blob/master/knowledgebase/views.py#L507

higs4281: Do we introduce non-page URL nodes in the Wagtail tree anywhere else?

cfgov/ask_cfpb/migration_scripts/migrate_knowledgebase.py
chosak: @rosskarchner I think we talked about this once before, but what about adding some kind of throttled Akamai flush logic? We could do this either with a simple cache value (storing the last time a flush was invoked, don't invoke a new flush if it was too recent) or something more complex like a Django model like `AkamaiCacheFlush`. One nice side effect of this would be to have somewhere to store the unique ID we get back from the Akamai API, so we'd be able to more easily track how often we flush, how long flushes take to happen, etc.

In either case we could stick this check [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/wagtail_hooks.py#L146) and skip the flush if we've just done one.

The downside of this approach would be that when updating a whole bunch of pages you would only flush after the first change, when really you'd want to do the last. So maybe a slightly more hacky way might be to remove the flush logic from the web request altogether; you write flush requests to the DB table and there's an async job (Jenkins) that polls that table for outstanding requests periodically.

cc @richaagarwal

cfgov/ask_cfpb/models/django.py
Scotchester: Yes, though it's probably not a recommended practice. The page at [wagtail-URL]/admin/pages/3637/ is not published, but its children are, and they include its slug in their URLs.

I don't think we can have a ton of numbered empty pages in the tree, though. I think it'd have to be some kind of special URL handling :\

rosskarchner: Every feed URL is a sort of "non-page URL", here's how we add those:

https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/feeds.py#L48

I think this case, I think it would be the AskCFPB page itself would override route and/or serve to handle question ID's.

rosskarchner: You might want to find out if the stakeholders actually care about the "self healing" functionality or having the numbers in the URL. If not, then this can still be the basis for catching those URL's and redirecting to the new ones

higs4281: I'll check, but I suspect current functionality will be preferred. Self-healing is my new personal goal.

Scotchester: They may not care, but I care. URLs going dead is bad, and not being able to redirect with the simple numbers will make life painful, because the redirect requests I get typically say "Redirect #123 to #456". Unless we're planning to give them the ability to set up redirects themselves, I'd like to maintain that simplicity.

rosskarchner: Wolverine

rosskarchner: @Scotchester agreed, but it also could be an opportunity to make Ask less "special". Maybe we need a site-wide system for catching slug changes, or marking one page as being superseded by another

cfgov/ask_cfpb/migration_scripts/migrate_knowledgebase.py
higs4281: I don't think this is actually an issue here. The hook only catches "after create" and "after edit_page" actions, which are admin page transactions, and doesn't touch `revision.publish()` itself. 
So the migration publish step, same as Wagtail's `publish_scheduled_pages` management  command used once a minute in our Jenkins `cf.gov-wagtail-publish-pages-cron-jobs` job, doesn't fire the sharing hook.

I tested this by setting a pdb trace before the akamai flush. If I edit and publish a page, the trace steps in. If I set a go-live datetime and run `publish_scheduled_pages` after the scheduled time, the page publishes and no trace is set.

chosak: Excellent, thanks for tracking this down. I withdraw my concern.

cfgov/ask_cfpb/models/django.py
higs4281: I had one idea for a healing URL pattern that doesn't mess with the Wagtail tree.

I pushed it as a [separate branch](https://github.com/cfpb/cfgov-refresh/tree/healing-url).

It depends solely on the ansewer slug, which carries a prefix that identifies the page by ID and language. 

 `ask-cfpb/old-slug-en-1979` will redirect to the English answer's current URL. If the slug does match, it just serves the page.

This is a pattern that would be straightforward to build redirects to. 

#### Caveat
One detail I haven't worked out is handling trailing slashes and avoiding a redirect loop. Currently the redirection magic only happens for bad URLs that have no trailing slash.

rosskarchner: Cool, I'll play with this later

higs4281:
@chosak 
Regarding your live-pages question, I've explored flagged_url, and it doesn't seem a viable option here, because it doesn't work for includes (at least according to the docs), and because we're looking at self-healing URLs that intercept Wagtail page requests. We can set a flag on the healer, but that just lets the requests just fall through to Wagtail. 

Are either of these options worth exploring?
- Exploit the fact that the new pages will (or can) have new parent urls, and thus the two main page branches (English and Spanish) could be controlled via the two parents. The URLs are not going to be identical in any case, and we'll be needing to redirect the old pages to new. The old-to-new pattern is relatively simple.
- Create a batch-publishing utility that can set go-live times on (or publish or unpublish) a parent page and all its descendants.

higs4281:
I'd like to merge in the initial self-healing-URLs code but set aside the deployment strategy issue. The goal of this PR is to get a working model on build for content editors to test-drive, because if this scheme is not viable for content editing, we'll need another approach. Thoughts?

Scotchester:
@higs4281 I'll see if I can test our your new idea later today, Bill, but one thing I should note is that links to Ask CFPB questions are littered throughout our content, so whatever route we go, if it involves creating a new URL format, it will be pretty important to ensure that we can craft a universal redirect rule from the old format to the new.

I'm guessing that your proposed approach would allow this with a rule like

`/askcfpb/###` ➡️ `/ask-cfpb/heal-me-en-####`

Does that seem correct?

chosak:
@higs4281 I'm good with that, considering most of the changes are in a new/separate app that can be easily iterated on. Worst case if we want to rollback it's just a matter of cleaning up the new app and tables.

The one thing I'm wondering is whether it's worth putting [the new `ask_cfpb` app](https://github.com/cfpb/cfgov-refresh/pull/2780/files#diff-ac27c5919c46548eada7a7388e726c59R52) behind a check in settings, so that it's only enabled on build, but not on real production servers. I'm not 100% sure of the easiest way to do that right now, though.

higs4281:
@Scotchester Yes, that pattern would do it, if you don't use a trailing slash (currently).

higs4281:
@chosak optional app?

Scotchester:
I do think the trailing slash issue is a pretty big one, though it wouldn't affect my proposed redirect.

higs4281:
@Scotchester Agreed. Trailing-slash needs to be resolved.

chosak:
@higs4281 but that's not configurable per-server, is it? I think a nice way to do this would be to check the environment to see if we're running on build, and only add to apps if so.

I'll also note (mostly to annoy @willbarton) that if we had code-based feature flags it'd be easier to put apps behind a feature flag.

higs4281:
I in the code-based flag camp, but there's no instant gratification today in annoying @willbarton

higs4281:
### Self-healing note

If you're trying the `healing-url` branch, there are added setup steps now that the Spanish tree is turned on.

In addition to creating an `ask-cfpb` LandingPage parent under CFGov, you'll need two Spanish parents:
- `inicio` LandingPage under CFGov
- `obtener-respuestas` LandingPage under `inicio`

willbarton:
I have lovely thoughts on a hybrid database/code feature flag setup.

willbarton:
squish

rosskarchner:
I don't seem to have the migration script

```sh
(cfgov-refresh) rk@:~/projects/cfgov-refresh$ git log | head
commit cae6134ef9e812cd63761bdfb012728116e24948
Author: higs4281 <higs4281@gmail.com>
Date:   Wed Mar 15 10:05:21 2017 -0400

    squash migrations

commit 625c8cde74d407984f981c57acb00dcf3b06e114
Merge: 4b8f833b8 e7ecf5679
Author: higs4281 <higs4281@gmail.com>
Date:   Tue Mar 14 13:49:25 2017 -0400
(cfgov-refresh) rk@:~/projects/cfgov-refresh$ ls cfgov/scripts/
__init__.py                 http_smoke_test.py
__init__.pyc                initial_data.py
_atomic_helpers.py          static_asset_smoke_test.py
archive_events.py           tests
create_careers_pages.py     update_es_settings.py
fix_links_in_tableblocks.py
```

higs4281:
@ross the script is in ask_cfpb/scripts.

rosskarchner:
ah, thanks-- I was missing the new environment variable

rosskarchner:
I will say again, I still don't love this approach (I think we could lean on Wagtail features more, rather than create a seperate editing experience), but there's no reason not to get it in front of stakeholders.

chosak:
@higs4281 one last thing: do you need to add this new app [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/db_router.py#L1)? Unfortunately we're currently using a whitelist of "apps that share a DB between content and production", as opposed to a blacklist of "apps that need a separate DB".

higs4281:
@chosak great idea. grazie.

PR #2798, opened 2017-03-14
Set a shorter TTL on filterable list pages

richaagarwal:
## Overview
- Currently we flush the entire site because we don't have a way to easily flush filterable pages on changes to one of their objects. Rather than invalidating the entire cache every time we publish or unpublish, it'd be great to simply lower the TTL to something reasonable (I propose 10 minutes in this PR) for pages for which we regularly need to refresh. 
- I think this allows us to move away from flushing the entire site upon publish/unpublish (PR here -- https://github.com/cfpb/cfgov-refresh/pull/2801), bearing in mind that there are other dynamic pages that might still require a manual trigger to flush the entire site (e.g. pages that live outside Wagtail, job listing pages and snippets, though for some of these we can also automate it)
## Related GHE issues
- GHE 704
- GHE 634
- GHE 550

### DIFF
--- a/cfgov/v1/util/filterable_list.py
+++ b/cfgov/v1/util/filterable_list.py
@@ -5,8 +5,7 @@ from django.core.paginator import EmptyPage, PageNotAnInteger, Paginator
 from v1.forms import FilterableListForm
 from v1.models.base import CFGOVPage
 from v1.models.learn_page import AbstractFilterPage
-
-from ..util.util import get_secondary_nav_items
+from v1.util.util import get_secondary_nav_items
 
 logger = logging.getLogger(__name__)
 
@@ -116,3 +115,9 @@ class FilterableListMixin(object):
             return form_ids[0]
         else:
             return 0
+
+    def serve(self, request, *args, **kwargs):
+        """ Modify response header to set a shorter TTL in Akamai """
+        response = super(FilterableListMixin, self).serve(request)
+        response['Edge-Control'] = 'cache-maxage=10m'
+        return response


cfgov/v1/util/filterable_list.py
```diff
@@ -116,3 +115,9 @@ def form_id(self):
             return form_ids[0]
         else:
             return 0
+
+    def serve(self, request):
```
chosak: Doesn't this need to be

```py
def serve(self, request, *args, **kwargs):
```

to match the correct signature [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/models/base.py#L267) and [here](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailcore/models.py)?

cfgov/v1/util/filterable_list.py
```diff
@@ -116,3 +115,9 @@ def form_id(self):
...
+        """ Modify response header to set a shorter TTL in Akamai """
+        response = super(FilterableListMixin, self).serve(request)
+        response['edge-control'] = 'cache-maxage=10m'
```
chosak: According to the Akamai docs (behind login wall) they use capitalization `Edge-Control`. They also say:

>  the edge server will issue an If Modified Since request back to the content provider to check whether the object
content has changed

By itself, this might require that we send the `Last-Modified` header with our pages, which I don't believe we do right now (cc @rosskarchner).

rosskarchner: We've enabled the use of etags as well as Last-modified. It's less than ideal, because we still need to render the document before doing the etag comparison. 

Doing that better, was part of what I was getting at in my last paragraph on GHE/CFGOV/platform/issues/550#issuecomment-106643

We should do revalidation better, but that's a seperate concern.

cfgov/v1/util/filterable_list.py
richaagarwal: Hmmm, I was matching the signature in the example here -- http://docs.wagtail.io/en/v1.9/reference/pages/model_recipes.html

chosak: I think in that example case they don't pass `*args` or `**kwargs` because that particular page class doesn't do anything with those optional arguments, but the base `Page` method does support them. We also might not use them on `FilterableList` pages but in theory we could if we used [RouteablePageMixin](http://docs.wagtail.io/en/v1.9/reference/contrib/routablepage.html) somewhere. It'd be nice to support that option.

cfgov/v1/util/filterable_list.py
```diff
@@ -5,8 +5,7 @@
 from v1.forms import FilterableListForm
 from v1.models.base import CFGOVPage
 from v1.models.learn_page import AbstractFilterPage
-
-from ..util.util import get_secondary_nav_items
+from v1.util.util import get_secondary_nav_items
```
willbarton: ❤️

PR #2801, opened 2017-03-16
Updating Akamai invalidation

richaagarwal:
### Overview
- Entire site will no longer be flushed when publishing/unpublishing, but the specific page will be
- Related: https://github.com/cfpb/cfgov-refresh/pull/2798
### Changes
- Uses Akamai's Fast Purge option (takes less than a minute!)
- Only invalidates the page being modified
- `flush` function now lives on Page model instead of in `wagtail_hooks.py`
- Removed the flush call from `unshare_all_revisions` and `unpublish_all_revisions` since we will be removing that code soon anyway (yay!), as well as `configure_page_and_revision` and instead hooked it up directly to the `page_published` and `page_unpublished` signals for clarity.
- Removed `should_flush` as this used to encapsulate more complex logic, and is now (and has been) a simple settings check

### TODO
- Immediate:
  - [ ] We'll need to ensure we make the appropriate changes to the variables in `ansible-consumer-finance/provisioning/group_vars`, and to the `ansible-consumer-finance-frontend/templates/configs/environment.json.j2` template file. (thanks @willbarton )
  - [ ] Test this on beta 
- Longer-term
  - [ ] Don't call `flush` on pages that are being published for the first time (Wagtail's frontendcache module doesn't take care of this)
  - [ ] `get_akamai_credentials` should probably not live in `wagtail_hooks.py` either, need to find a home for it 
  - [ ] Hook this into Wagtail's [frontendcache module](http://docs.wagtail.io/en/v1.9/reference/contrib/frontendcache.html)

### DIFF
--- a/.env_SAMPLE
+++ b/.env_SAMPLE
@@ -36,9 +36,10 @@ export DJANGO_HTTP_PORT=8000
 #export EXTERNAL_LINK_CSS=<external_links_css_class_name>
 export ALLOW_ADMIN_URL=True
 #export ENABLE_AKAMAI_CACHE_PURGE=<set_to_enable_purging_akamai_cache>
-#export AKAMAI_USER=<akamai_username>
-#export AKAMAI_PASSWORD=<akamai_password>
-#export AKAMAI_OBJECT_ID=<akamai_object_id>
+#export AKAMAI_CLIENT_TOKEN=<akamai_client_token>
+#export AKAMAI_CLIENT_SECRET=<akamai_client_secret>
+#export AKAMAI_ACCESS_TOKEN=<akamai_access_token>
+#export AKAMAI_FAST_PURGE_URL=<akamai_fast_purge_url>
 #export AKAMAI_NOTIFY_EMAIL=<akamai_notify_email>
 #export AKAMAI_HOST=<akamai_hostname>
 
--- a/cfgov/v1/management/commands/delete_page.py
+++ b/cfgov/v1/management/commands/delete_page.py
@@ -2,7 +2,6 @@ from django.core.exceptions import MultipleObjectsReturned
 from django.core.management.base import BaseCommand, CommandError
 
 from v1.models import CFGOVPage
-from v1.wagtail_hooks import flush_akamai
 
 
 class Command(BaseCommand):
@@ -36,7 +35,6 @@ class Command(BaseCommand):
             raise CommandError('Must supply a slug or an id')
         try:
             page_to_delete.delete()
-            flush_akamai()
 
         except Exception as e:
             self.stderr.write(str(e))
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -1,8 +1,5 @@
 import csv
 import json
-import logging
-import requests
-
 from collections import OrderedDict
 from cStringIO import StringIO
 from itertools import chain
@@ -36,9 +33,6 @@ from v1 import get_protected_url
 from v1.atomic_elements import molecules, organisms
 from v1.models.snippets import ReusableText, ReusableTextChooserBlock
 from v1.util import ref
-from v1.wagtail_hooks import get_akamai_credentials
-
-logger = logging.getLogger(__name__)
 
 
 class CFGOVAuthoredPages(TaggedItemBase):
@@ -222,41 +216,6 @@ class CFGOVPage(Page):
         # hit for that type.
         return {search_type: queryset for search_type, queryset in
                 related.items() if queryset}
-
-    def flush(self):
-        if not settings.ENABLE_AKAMAI_CACHE_PURGE:
-            logger.info(
-                'Page {slug} (id = {id}) was not invalidated because '
-                'ENABLE_AKAMAI_CACHE_PURGE is not set'.format(
-                    slug=self.slug,
-                    id=self.id
-                )
-            )
-            return
-        auth, fast_purge_url = get_akamai_credentials()
-        headers = {'content-type': 'application/json'}
-        payload = {
-            'action': 'invalidate',
-            'hostname': settings.WAGTAIL_SITE_NAME,
-            'objects': [
-                self.relative_url(self.get_site())
-            ]
-        }
-        resp = requests.post(
-            fast_purge_url,
-            headers=headers,
-            data=json.dumps(payload),
-            auth=auth
-        )
-        logger.info(
-            'Attempted to invalidate page {slug} (id = {id}), '
-            'got back response {message}'.format(
-                slug=self.slug,
-                id=self.id,
-                message=resp.text
-            )
-        )
-        resp.raise_for_status()
 
     def get_appropriate_page_version(self, request):
         # If we're on the production site, make sure the version of the page
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -1,5 +1,8 @@
 import csv
 import json
+import logging
+import requests
+
 from collections import OrderedDict
 from cStringIO import StringIO
 from itertools import chain
@@ -33,6 +36,9 @@ from v1 import get_protected_url
 from v1.atomic_elements import molecules, organisms
 from v1.models.snippets import ReusableText, ReusableTextChooserBlock
 from v1.util import ref
+from v1.wagtail_hooks import get_akamai_credentials
+
+logger = logging.getLogger(__name__)
 
 
 class CFGOVAuthoredPages(TaggedItemBase):
@@ -216,6 +222,41 @@ class CFGOVPage(Page):
         # hit for that type.
         return {search_type: queryset for search_type, queryset in
                 related.items() if queryset}
+
+    def flush(self):
+        if not settings.ENABLE_AKAMAI_CACHE_PURGE:
+            logger.info(
+                'Page {slug} (id = {id}) was not invalidated because '
+                'ENABLE_AKAMAI_CACHE_PURGE is not set'.format(
+                    slug=self.slug,
+                    id=self.id
+                )
+            )
+            return
+        auth, fast_purge_url = get_akamai_credentials()
+        headers = {'content-type': 'application/json'}
+        payload = {
+            'action': 'invalidate',
+            'hostname': settings.WAGTAIL_SITE_NAME,
+            'objects': [
+                self.relative_url(self.get_site())
+            ]
+        }
+        resp = requests.post(
+            fast_purge_url,
+            headers=headers,
+            data=json.dumps(payload),
+            auth=auth
+        )
+        logger.info(
+            'Attempted to invalidate page {slug} (id = {id}), '
+            'got back response {message}'.format(
+                slug=self.slug,
+                id=self.id,
+                message=resp.text
+            )
+        )
+        resp.raise_for_status()
 
     def get_appropriate_page_version(self, request):
         # If we're on the production site, make sure the version of the page
--- a/cfgov/v1/signals.py
+++ b/cfgov/v1/signals.py
@@ -1,10 +1,12 @@
 import json
 from datetime import timedelta
+import logging
 
 from django.dispatch import Signal
 from django.utils import timezone
 from wagtail.wagtailcore.signals import page_published, page_unpublished
 
+logger = logging.getLogger(__name__)
 page_unshared = Signal(providing_args=['instance'])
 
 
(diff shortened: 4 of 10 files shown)


cfgov/v1/models/base.py
```diff
@@ -217,6 +220,29 @@ def specific_categories_query(block, parent_slug):
         return {search_type: queryset for search_type, queryset in
                 related.items() if queryset}
 
+    def flush(self):
+        if not settings.ENABLE_AKAMAI_CACHE_PURGE:
+            return False
+        auth, fast_purge_url = get_akamai_credentials()
+        page_relative_path = self.url_path.replace("cfgov/", "")
```
richaagarwal: this is a little funky but I really didn't want to pass in the `request` just to get the `site`

cfgov/v1/signals.py
```diff
@@ -53,25 +53,26 @@ def update_all_revisions(instance, attr):
```
richaagarwal: I removed the flush call from `unshare_all_revisions` and `unpublish_all_revisions` since we will be removing this code soon anyway (yay!) and decided to hook it up directly to the `page_published` and `page_unpublished` signals.

cfgov/v1/signals.py
```diff
@@ -53,25 +53,26 @@ def update_all_revisions(instance, attr):
 
 
 def unshare_all_revisions(sender, **kwargs):
-    from v1.wagtail_hooks import flush_akamai
     update_all_revisions(kwargs['instance'], 'shared')
-    flush_akamai()
```
richaagarwal: I removed the flush call from unshare_all_revisions and unpublish_all_revisions since we will be removing this code soon anyway (yay!) and decided to hook it up directly to the page_published and page_unpublished signals.

Scotchester:
:heart_eyes_cat:

willbarton:
> Entire site will no longer be flushed when publishing/unpublishing, but the specific page will be

not bad

.env_SAMPLE
```diff
@@ -36,9 +36,10 @@ export DJANGO_HTTP_PORT=8000
 #export EXTERNAL_LINK_CSS=<external_links_css_class_name>
 export ALLOW_ADMIN_URL=True
 #export ENABLE_AKAMAI_CACHE_PURGE=<set_to_enable_purging_akamai_cache>
-#export AKAMAI_USER=<akamai_username>
-#export AKAMAI_PASSWORD=<akamai_password>
-#export AKAMAI_OBJECT_ID=<akamai_object_id>
+#export AKAMAI_CLIENT_TOKEN=<akamai_client_token>
+#export AKAMAI_CLIENT_SECRET=<akamai_client_secret>
+#export AKAMAI_ACCESS_TOKEN=<akamai_access_token>
+#export AKAMAI_FAST_PURGE_URL=<akamai_fast_purge_url>
```
willbarton: We'll need to ensure we make the appropriate changes to the variables in `ansible-consumer-finance/provisioning/group_vars`, and to the `ansible-consumer-finance-frontend/templates/configs/environment.json.j2` template file.

cfgov/v1/models/base.py
```diff
@@ -217,6 +220,29 @@ def specific_categories_query(block, parent_slug):
...
+        }
+        resp = requests.post(
+            fast_purge_url,
+            headers=headers,
+            data=json.dumps(payload),
+            auth=auth
+        )
+        if resp.status_code == 201:
+            return True
+        return False
```
willbarton: Would it be worth logging anything here about the failure to flush? Is there anything in a non-`201` response that we could log that would help in debugging?

cfgov/v1/wagtail_hooks.py
```diff
@@ -128,46 +129,21 @@ def configure_page_revision(page, is_sharing, is_live):
 
 
 def get_akamai_credentials():
```
willbarton: You're right, this should probably live somewhere else. I'm not sure where to suggest, right off.

cfgov/v1/management/commands/delete_page.py
```diff
@@ -36,7 +35,7 @@ def handle(self, *args, **options):
             raise CommandError('Must supply a slug or an id')
         try:
             page_to_delete.delete()
-            flush_akamai()
+            page_to_delete.flush()
```
chosak: Can you call `.flush()` after you've called `delete()`?

cfgov/v1/models/base.py
chosak: Because you're using `requests` you could do `resp.raise_for_status()` which will raise `requests.exceptions.HTTPError` for a 4XX or 5xx response. It does seem like we'd want to log this somehow -- is there any public-facing action that would cause this call to happen, or is it only invoked by Wagtail users, for whom the display of an error page caused by exception might be okay?

cfgov/v1/signals.py
```diff
@@ -53,25 +55,33 @@ def update_all_revisions(instance, attr):
...
+
+
+def flush_page(sender, **kwargs):
+    try:
+        kwargs['instance'].flush()
+    except AttributeError as e:
+        # This will happen when setting up the site locally
+        # (e.g. when running `tox`) since it creates a non-CFGOVPage Page
+        # which does not have `flush` defined
+        logger.info(e)
```
chosak: Alternate shortcut:

```py
try:
    kwargs['instance'].flush()
except AttributeError:
    logger.exception('some message here')
```

This will automatically [include the stack trace](https://docs.python.org/2/library/logging.html#logging.Logger.exception) in the message.

That being said, is there an advantage in logging here, and not just ignoring?

cfgov/v1/models/base.py
chosak: Is there some alternate approach we could use here, like `self.relative_url(self.get_site())`? It'd be nice not to have to hardcode this.

cfgov/v1/models/base.py
```diff
@@ -217,6 +220,29 @@ def specific_categories_query(block, parent_slug):
...
 
+    def flush(self):
+        if not settings.ENABLE_AKAMAI_CACHE_PURGE:
+            return False
+        auth, fast_purge_url = get_akamai_credentials()
+        page_relative_path = self.url_path.replace("cfgov/", "")
+        headers = {'content-type': 'application/json'}
+        payload = {
+            'action': 'invalidate',
+            'hostname': settings.WAGTAIL_SITE_NAME,
```
chosak: Will this work with beta.consumerfinance.gov?

cfgov/v1/management/commands/delete_page.py
richaagarwal: Ah, thanks for pointing this out!  Yeah, it looks like it might not work, but it does since the page reference is still intact, and you're just passing along the path to Akamai.  I can't think of a better way to do this, but am totally open to suggestions.

cfgov/v1/models/base.py
richaagarwal: Ooooh I didn't know about `get_site()`. Totally.

richaagarwal: Hmm. Actually, I get a lot of `Site matching query does not exist.` errors with this approach.  It doesn't seem like `self.get_site()` consistently works, unless I broke something else while testing.

chosak: It should work with any pages that are published under a site root. See [docs](http://docs.wagtail.io/en/v1.9/reference/pages/model_reference.html#wagtail.wagtailcore.models.Page.get_site) and [source code](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailcore/models.py#L853). It's supposed to return null if the site doesn't exist, so I'm curious where your errors are coming from. It seems reasonable to me to skip the cache step for pages that aren't published, e.g.

```py
site = self.get_site()

if not site:
    return

# do something with self.relative_url(site) ...
```

richaagarwal: Okay it works again, must've been something with my environment.  This gets called after a page if published, so we can't skip flushing new pages that way.  I can add that to my TODO list but the complex logic might not be worth it.

.env_SAMPLE
richaagarwal: Cool, I'll create a PR for those soon, since those should go in before this gets merged.

cfgov/v1/management/commands/delete_page.py
richaagarwal: Actually, I'm not sure why this was ever here.  We don't need to flush after deleting if we're flushing after unpublishing, since going from unpublished to delete makes no difference for what we're displaying on the live site.

cfgov/v1/signals.py
richaagarwal: Oh, I like that idea (about checking if `flush` exists)!

cfgov/v1/models/base.py
richaagarwal: Yes, meant to add some kind of logging here.  If we were to raise an exception, it would only be displayed to Wagtail users, but I worry this would be confusing, since it'd happen upon publish/unpublish.  While it's good for us to know the flush isn't working, having the page error out on an action like that would make it seem that the page did not successfully publish/unpublish. ... Although, I guess in practice, it didn't if we can't flush the site. So maybe that's not that bad?  Curious what @sarahsimpson09 and Jordana think.

cfgov/v1/models/base.py
```diff
@@ -217,6 +223,41 @@ def specific_categories_query(block, parent_slug):
...
+            ]
+        }
+        resp = requests.post(
+            fast_purge_url,
+            headers=headers,
+            data=json.dumps(payload),
+            auth=auth
+        )
+        logger.info(
+            'Attempted to invalidate page {slug} (id = {id}), '
```
richaagarwal: Okay I added a bit of logging here and above.  On a failure (that I happen to be getting now because of some config thing) this would look something like:
```
Attempted to invalidate page faqs (id = 993), got back response {"supportId": "17PY148XXXXX", "title": "unauthorized arl", "httpStatus": 403, "detail": "http://consumerfinance.gov/foia-requests/faqs/", "describedBy": "https://api.ccu.akamai.com/ccu/v2/errors/unauthorized-arl"}
```
I have it set to log this regardless of if the request is successful or not, since it might be useful in the beginning to log what we are actually successfully flushing.  And the page will error out so we'll know immediately if something's not working.

richaagarwal:
I commented on or addressed most of the feedback left here.  I will make some minor updates to support this for beta, but need to get it working with the credentials I have (which are for prod) first, which I haven't been able to yet. I think it's a setting on Akamai's end that hasn't been fully enabled, but TBD.  In the meantime, keep the feedback coming!

richaagarwal:
> I know you'd looked into the Wagtail frontendcache module a bit; did you consider somehow extending it with a new Akamai backend?

@chosak I hadn't considered it, but that's a great idea.  What do you think about this as more of a MVP and that could be the next iteration?

cfgov/v1/wagtail_hooks.py
richaagarwal: Previously, `flush_akamai` lived here, so I just added it here too, but now it makes less sense.  I'm not going to worry about it too much for now; I anticipate this code will get shuffled around a bit, particularly if we look into extending Wagtail's frontendcache module with an Akamai backend as @chosak suggested.

cfgov/v1/models/base.py
```diff
@@ -217,6 +223,41 @@ def specific_categories_query(block, parent_slug):
         return {search_type: queryset for search_type, queryset in
                 related.items() if queryset}
 
+    def flush(self):
+        if not settings.ENABLE_AKAMAI_CACHE_PURGE:
+            logger.info(
+                'Page {slug} (id = {id}) was not invalidated because '
```
chosak: Minor thing: because this file doesn't have `from __future__ import unicode_literals` you should use unicode strings (`u'Page {slug}...'`) in case slugs contain unicode characters. Similarly at the other logging statement [below](https://github.com/cfpb/cfgov-refresh/pull/2801/files#diff-8e8a3e28d46ee12516bea4f5b50bf5a4R252).

chosak:
@richaagarwal regarding frontendcache: sure, I think this could be a good MVP. It might be worth looking into that approach since it might be less code overall, but it doesn't have to be a blocker for moving ahead with caching improvements. I do think it might be useful for other Wagtail/Akamai users!

.env_SAMPLE
richaagarwal: Done!

cfgov/v1/models/base.py
richaagarwal: If we go the absolute URL route, then we won't need this

PR #2803, opened 2017-03-17
Uppercase DEPLOY_ENVIRONMENT var to match ansible

higs4281:
Syncing with Ansible change.


chosak:
@higs4281 thanks; I did some minor refactoring so that we only ever check the environment once. There are other places in the code that I've been looking forward to using this as a setting as well.

It occurs to me that the pattern of checking deploy environment against a string `'build'` is a bit sub-optimal. It might be worth thinking about having some sort of enumerated set. Ideally it'd be better to just embed this into feature flags somehow, so that we can do code-based flags that are defined like "feature on for environment(s) X, Y, Z".

higs4281:
> it'd be better to just embed this into feature flags somehow

👍

PR #2796, opened 2017-03-14
Hack for sub navigation mobile items being hidden

sebworks:
Removing code so the mobile sub navigations items show on mobile. This is a temporary fix that shouldn't ever be necessary.  This will also break existing logic surrounding which sub nav items show based on the requested URL / item slug.

## Screenshots
- 
screen shot 2017-03-14 at 10 47 47 am

 

## Notes

-

## Todos

-

## Checklist

* [ ] PR has an informative and human-readable title
* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated
* [ ] Reviewers requested with the [Assignee tool](https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/) :arrow_right:

PR #2804, opened 2017-03-17
Fix secondary navigation page highlighting

chosak:
This PR fixes a bug with how pages in sidebar navigation were not being highlighted after a transition from HTTP to HTTPS. Previously the logic to determine if a page was selected used

```py
request.url.startswith(item.url)
```
which was prone to problems if the scheme of the two requests differed (HTTP vs HTTPS, which could happen due to load balancers, etc). This change adds a new `item.active` parameter which indicates whether or not the page should be highlighted.

## Notes

## Changes

- Fix to `v1.util.get_secondary_nav_items` that properly indicates which sidebar item should be highlighted.

## Testing

- Using a production data dump, visit
  - http://localhost:8000/about-us/the-bureau/leadership-calendar/
  - http://localhost:8000/about-us/events/archive-past-events/

## Screenshots

image

## Notes

- This code is fairly convoluted which makes it difficult to test efficiently. The sharing logic is due to be removed soon so this will be a quick fix for now.

## Todos

- Clean up this function once sharing logic is removed.

## Checklist


Scotchester:
Unfortuantely, this is not fixing the issue for me when running on https://localhost.

I don't know if I can debug why your changes aren't working as intended, but why not fix the root cause by making the generated `href`s in the sidenav be root-relative URLs with no protocol?

chosak:
@Scotchester thanks very much for testing this; you're right that it wasn't working properly. I've pushed another fix in a200305 - would you mind testing again?

I agree it'd be better to have those sidebar pages be root-relative URLs -- frankly, the reason I didn't touch that logic is because it currently uses things like `get_appropriate_page_version` and `get_page_state_url`, both of which should go away once our custom sharing code is removed. The code currently (at least tries to) properly handle "shared" pages, and doing that with relative URLs is a bit more tricky. So if it's okay with you, I'd rather keep this band-aid and look forward to a more thorough cleanup in future.

schaferjh:
@chosak thanks for this -- can you confirm (just because I don't see it called out in the overview above) that this solves the related bug of this section nav not being viewable on mobile? (and thus negating the need for @sebworks #2796?)

chosak:
@schaferjh thanks for the reminder about that -- mobile still looks broken. I'm not super familiar with that code so let me take a quick look at that part of it.

Scotchester:
After pairing with @chosak, I've tested the latest commits and this is working perfect for me locally. :+1:

chosak:
Thanks @Scotchester! @schaferjh I think we have a working solution for both desktop and mobile browsing. As a nice side effect of our work we've fixed the sidebar links so they are now root-relative.

schaferjh:
Exciting! 👍

sebworks:
thanks @chosak @Scotchester

PR #2805, opened 2017-03-17
Use settings.DEPLOY_ENVIRONMENT in error email subject

chosak:
Instead of using the environment variable `EMAIL_SUBJECT_PREFIX`, which we don't seem to reliably set, use the `DEPLOY_ENVIRONMENT` setting (which comes from the `DEPLOY_ENVIRONMENT` variable) instead. If this setting isn't set, fallback to the default (`"[Django] "`).

## Changes

- Deprecate use of `EMAIL_SUBJECT_PREFIX` for error email subjects in favor of `DEPLOY_ENVIRONMENT`.

## Checklist


higs4281:
Great idea.

One very minor thought: A lowercase subject prefix might look odd. `.title()`?

cfgov/cfgov/settings/base.py
```diff
@@ -376,7 +376,10 @@
 
 # Email
 ADMINS = admin_emails(os.environ.get('ADMIN_EMAILS'))
-EMAIL_SUBJECT_PREFIX = os.environ.get('EMAIL_SUBJECT_PREFIX')
+
+if DEPLOY_ENVIRONMENT:
+    EMAIL_SUBJECT_PREFIX = u'[{}] '.format(DEPLOY_ENVIRONMENT)
+
```
richaagarwal: is there supposed to be an `else` here to set it to a default, or is that happening somewhere else?

willbarton: [The default is just `[Django]`](https://docs.djangoproject.com/en/1.8/ref/settings/#email-subject-prefix), which is less than ideal but no worse than we have now. I think that's acceptable.

chosak:
@higs4281 thanks, that's a great suggestion. Addressed in 4ce0728.

PR #2807, opened 2017-03-18
Add a redirect_id field to allow editors to redirect AnswerPages

higs4281:
This adds a `redirect_id` field to AnswerPages so that editors can control the redirection of answers instead of filing issues requesting code updates to our global redirections.

If an AnswerPage is redirected in this way, its status in the admin will show "REDIRECTED."

## Testing
To test locally, migrate the model change and then load the knowledgebase data if you don't have it. If you don't want to run the giant KB migration, you could create two dummy AnswerPages.
You'll need to have the knowledgebase app installed inside cfgov-refresh. 

```
```

Then edit an AnswerPage, add an Answer ID for a different answer and publish.
Now if you click on the redirected page's "LIVE" link, it should render the other page. 
Also, the page's status box in the admin should now read "REDIRECTED"

redirected

## Additions

- `redirect_id` field
- status handler for AnswerPage
- view code to handle redirect

## Checklist


willbarton:
This is also hero's work.

I still want to play with this locally before reviewing properly (especially given the set of Ask CFPB redirects we have on this sprint!), but I'm thinking about our `redirects.conf` and wondering if you've given any thought to how we might able to migrate all of those old redirects. And "no" is a perfectly valid answer to that question 😀 .

higs4281:
> wondering if you've given any thought to how we might able to migrate all of those old redirects.

It appears that the redirected answers have been deleted, but the old questions were left in the db. If we don't mind having ghost pages in the Wagtail tree, we could generate blank answer pages with the old question IDs that would perform the redirect. That would let us drop the old conf lines instead of having to rewrite them.

willbarton:
Ah. I think it'd probably be better to keep the Wagtail tree clean than to clean up the conf file.

schaferjh:
>  (especially given the set of Ask CFPB redirects we have on this sprint!)
@willbarton Does this work impact the redirects we need to get done now? I though it would apply only once this content is being maintained from Wagtail, yes?

willbarton:
@schaferjh correct, this only applies after the content is migrated to Wagtail so it doesn't help us with the current redirects. I was just dreaming of a future of fewer lovingly hand-crafted Apache redirects.

schaferjh:
@willbarton indeed. ok, thanks, just making sure I'm tracking correctly!

Scotchester:
> Ah. I think it'd probably be better to keep the Wagtail tree clean than to clean up the conf file.

But @willbarton, if we accept this change and encourage its use, we'll be in the same situation a year or two from now, with a lot of pages in Wagtail with the status of "REDIRECTED".

willbarton:
@Scotchester presumably it'd be slightly more useful though, in that they wouldn't be empty ghosts, but would still contain their historical information, no?

Scotchester:
@willbarton We could, at the very least, display the questions for the ghost pages. And this would make it possible to query for all the redirects in a single location, versus having to go back and forth between the conf file and Wagtail.

higs4281:
If we get this onto build, it would be pretty easy to migrate the conf redirects into build as redirected pages, to get a look at the ghost-page approach.

cfgov/ask_cfpb/models/pages.py
```diff
@@ -56,3 +62,23 @@ def __str__(self):
             return '{}: {}'.format(self.answer_base.id, self.title)
         else:
             return self.title
+
+    @property
+    def status_string(self):
```
chosak: Can I make a request that we handle `redirect_id` specially but then call `super(AnswerPage, self).status_string` if not redirected? This will remove duplicate logic about scheduled/drafted changes.

cfgov/ask_cfpb/views.py
```diff
@@ -6,6 +6,12 @@
 def view_answer(request, slug, language, answer_id):
     answer_page = get_object_or_404(
         AnswerPage, language=language, answer_base__id=answer_id)
+    if answer_page.redirect_id:
+        new_page = get_object_or_404(
+            AnswerPage,
+            language=language,
+            answer_base__id=answer_page.redirect_id)
+        return redirect(new_page.url)
     if "{}-{}-{}".format(slug, language, answer_id) != answer_page.slug:
```
chosak: Minor thing in Python2 - if slug has unicode character this will raise an exception unless you mark the string with `u"{}-{}-{}"` or include `from __future__ import unicode_literals`.

higs4281:
Great suggestions, @chosak 
Addressed in 43f9722.
.

higs4281:
> more than just an IntegerField

I had the same thought, and initially used a foreign key field, but the default admin field for that isn't friendly, and I wanted to avoid the model complexity until we demo the idea.

PR #2813, opened 2017-03-22
Testing for exposed admin urls

willbarton:
We've been talking about admin URLs, and ensuring that we don't expose anything we don't want to expose. We have an `ALLOW_ADMIN_URL` setting that we use to turn admin URLs on and off on a per-Django-instance basis, but there may be situations where we want to enable admin URLs internally but not externally for a single Django instance. Those URLs might be blacklisted between the Django instance and the public, however. For this reason, we need to verify that we always know exactly which admin URLs are enabled by `ALLOW_ADMIN_URL`. This PR adds a test that introspects our URLs and verifies that we're not exposing anything that we don't know about.

This test will work with our optional apps when they're included by using `tox -e optional`. That is where the real benefit of this test comes from — satellite apps could add URLs that use `ALLOW_ADMIN_URL` and we might not be aware of it otherwise.
 
## Additions

- Test for `ALLOW_ADMIN_URL` urls 

## Checklist


cfgov/cfgov/tests/test_urls.py
```diff
@@ -0,0 +1,71 @@
+try:
```
richaagarwal: When/why would this fail?

cfgov/cfgov/tests/test_urls.py
```diff
@@ -0,0 +1,71 @@
...
+    urls = []
+    for p in urlpatterns:
+        if isinstance(p, RegexURLPattern) or hasattr(p, '_get_callback'):
+            urls.append(base + p.regex.pattern)
+        elif (isinstance(p, RegexURLResolver) or
+              hasattr(p, 'url_patterns') or
+              hasattr(p, '_get_url_patterns')):
+            try:
+                patterns = p.url_patterns
+            except ImportError:
```
richaagarwal: Why would this throw an `ImportError` ?

cfgov/cfgov/tests/test_urls.py
```diff
@@ -0,0 +1,71 @@
...
+    '^logout',
+    '^tasks/',
+    '^csp-report/',
+    '^password/',
+    '^d/admin/',
+]
+
+
+# Based on django_extensions's show_urls command.
+def get_urlpatterns(urlpatterns, base=''):
```
richaagarwal: Can you add a docstring here? It's not clear what this function is doing

cfgov/cfgov/tests/test_urls.py
willbarton: This is for Django 1.10+ compatibility. `RegexURLPattern` and `RegexURLResolver` moved to `django.urls` in 1.10.

cfgov/cfgov/tests/test_urls.py
willbarton: I'm not sure — the code this is based on catches an `ImportError` there. Removing it doesn't throw an `ImportError` with cfgov-refresh or any of its optional apps. Removing it.

willbarton:
@richaagarwal good comments on late-night-written code. I've renamed that function to a much more descriptive `extract_regexes_from_urlpatterns`, because that's what it actually does.

cfgov/cfgov/tests/test_urls.py
```diff
@@ -0,0 +1,70 @@
...
+    regexes = []
+    for p in urlpatterns:
+        if isinstance(p, RegexURLPattern) or hasattr(p, '_get_callback'):
+            regexes.append(base + p.regex.pattern)
+        elif (isinstance(p, RegexURLResolver) or
+              hasattr(p, 'url_patterns') or
+              hasattr(p, '_get_url_patterns')):
+            patterns = p.url_patterns
+            regexes.extend(extract_regexes_from_urlpatterns(
+                p atterns, base + p.regex.pattern))
```
richaagarwal: typo here causing test to fail

PR #2820, opened 2017-03-27
Add HTML editor and tip styling button for Ask CFPB Answers

virginiacc:
Adds buttons to the RichTextField editor on Ask admin pages so content editors can continue to create tips/warnings in Ask CFPB answers. 

## Additions

- HTML editor in Ask admin RichTextFields
- Tip styling button for Ask admin RichTextFields
- CSS to style tips in RichTextFields on Ask admin pages
- CSS to style tips on Ask answer pages

## Testing

- Check out branch 
- Set up Ask CFPB per instructions in #2780 
- Run gulp
- Navigate to [an Ask CFPB question](http://localhost:8000/admin/ask_cfpb/answer/edit/2061/) in the admin
  - Click on the `Edit HTML` button on the Answer field to bring up the html editor. Verify that changes made to the HTML are preserved when you save the answer and [preview the associated question page](http://localhost:8000/admin/pages/6950/edit/)
  - Create a test tip header & paragraph in the Answer RichTextField. Highlight the tip and click the tip styling button. Verify that the tip is wrapped in an `<aside class="answer-module>` & appears with the correct styling in the RichTextField and on the associated question page.
- Navigate to a non-Ask page in the admin and verify that the tip and html editor buttons do not appear on RichTextFields

#### HTML editing:

Editor button
screen shot 2017-03-27 at 2 57 06 pm  

  
Editor view
screen shot 2017-03-27 at 2 56 36 pm
  

#### Tip button
screen shot 2017-03-27 at 2 43 25 pm

  
screen shot 2017-03-27 at 2 44 01 pm

## Notes

- Eventually we want to change the way tips are created so editors don't need to apply classes to them, but for now this replicates existing functionality
- We may only need one of the buttons; this PR allows content editors to assess them both on build

## Checklist


sebworks:
If we go this route we are committing to couple of things:

- Allowing the content folks to edit the HTML directly. We typically build the patterns outside the editor.
- Adding more code to modify a deprecated Halo editor. It looks like it Wagtail will begin moving to Daft.js in Wagtail 1.10 ( https://github.com/wagtail/wagtail/projects/1)


I like the overall the direction of the PR . If the content folks are good with it, then I then I am.

chosak:
This is a nice implementation! I do share some of @sebworks' concerns, though.

As usual when we talk about HTML editing I point out that you can do nasty stuff like this:

image

image

I also worry that if/when we upgrade to Wagtail 1.9/1.10 we won't have sufficient testing to detect if this breaks. Is there some way to write automated tests against this functionality?

Also, minor thing, but is it possible to change the tooltip on the Tip button to be more descriptive (maybe [here](https://github.com/cfpb/cfgov-refresh/pull/2820/files#diff-a9f847cd3d7da876341bbe9ce217f574R21))?

image

cfgov/v1/static/js/admin/html_editor.js
```diff
@@ -0,0 +1,79 @@
+(function() {
+  (function($) {
+    return $.widget('IKS.editHtmlButton', {
+      options: {
+        uuid: '',
+        editable: null
+      },
+      populateToolbar: function(toolbar) {
+        if (window.location.href.indexOf('ask_cfpb') > -1) {
```
chosak: I see, this is a nice way to limit, although it'd be really nice if this was somehow a little more straightforward or could be configured via some kind of variable passed into the JS.

virginiacc:
@sebworks @chosak I think we share your concerns about HTML editing and are hoping this will be a short term solution to allow users to work with tips. In the long run, since we need to use RichTextFields for the Ask answers, we are looking at alternatives like creating a separate field for the tip content.  

Also, we're including both buttons so content editors can test them on build. If the tips styling button works well enough, we might be able to remove the html editor right away.


@chosak Wagtail does enforce an HTML element whitelist for RichTextFields when the page is saved, so we should have some protection against things like script tags.

Great point about the Tip button tooltip -- I just pushed an update:

tooltip
  
The tip button icon probably isn't very useful either -- for expediency I just chose one that was already available. If the button's functionality tests well with the content editors, I can find a more suitable icon.

sebworks:
After speaking with the content team I have some additional concerns:

- Copy / pasting from word directly and introducing some unwanted MSO characters.
- Entering in content that creates a security hole 
- Entering content in violation of the Design manual guidelines
- Entering malformed markup with looks correct on modern browsers but breaks on older browsers.

Why not just create a modal workflow allows them to just enter the content.

willbarton:
@sebworks Since this is intended to be temporary, scoped to Ask CFPB, and Wagtail does element whitelisting, I don't think those concerns should be blockers.

higs4281:
@virginia I added tip markup cleaning to the migration script, which might reduce the need for an HTML editor. However, imperfections in the migration and other markup oddities will make the editor useful for migration troubleshooting.

@sebworks I share those concerns, but pasting from Word (or from the web) is an issue regardless of the presence of an HTML editor, and in those cases the editor makes troubleshooting more efficient.

cfgov/ask_cfpb/scripts/migrate_knowledgebase.py
```diff
@@ -35,6 +36,57 @@
...
+                    i = wrapper.contents.index(':')
+                    del(wrapper.contents[i])
+    return unwrap_soup(soup)
+
+
+def fix_tips(answer_text):
+    try:
+        clean1 = convert_divs_to_asides(answer_text)
+        clean2 = clean_orphan_tips(clean1)
+    except Exception:
```
willbarton: I realize this script is intended to be run once on a known dataset, but `Exception` here feels overly broad. What can `clean_orphan_tips` raise that we'd want to catch? Will `convert_divs_to_asides` raise anything? It seems like we expect an exception to indicate that there are no tips, is that correct?

cfgov/ask_cfpb/wagtail_hooks.py
```diff
@@ -82,3 +87,42 @@ class MyModelAdminGroup(ModelAdminGroup):
...
+        settings.STATIC_URL +
+        'css/question_tips.css">')
+
+
+def whitelister_element_rules():
+    return {
+        'aside': attribute_rule({'class': True}),
+    }
+
+if settings.DEPLOY_ENVIRONMENT == 'build':
```
willbarton: I'm increasingly of the opinion that the use of `DEPLOY_ENVIRONMENT` for this is a bit abusive. Hopefully we can replace these checks with feature flags [soon](https://github.com/cfpb/wagtail-flags/pull/7#issuecomment-290088716).

sebworks:
ok, thanks @chosak @higs4281.  I would just make sure you coordinate with the content and the design team to make sure they understand the ramifications of the changes.

cfgov/ask_cfpb/scripts/migrate_knowledgebase.py
higs4281: There are so many flavors of tip markup (which is a powerful argument for ditching the HTML editor) that I didn't want to halt the migration for any edge cases, but I'll try to narrow down the exception.

willbarton: Does that mean that there are cases where the tips cause an exception and get through as part of the answer text? Is that acceptable for the purposes of the migration?

virginiacc:
@higs4281 Thanks for adding the tip cleanup!

@sebworks I share your concerns about allowing html editing. I think the complexity in this comes from creating the formatted tips inside a RichTextField, but we're not yet at a place to change to another approach like separate fields. We'll work with the content editors to test the plugins on build. 

I also like your idea of a modal workflow -- I'll think about that if we continue working with tips inside the RichTextFields.

cfgov/ask_cfpb/scripts/migrate_knowledgebase.py
higs4281: My fears were unfounded. Tips get cleaned with no exceptions.

cfgov/v1/static/js/admin/html_editor.js
virginiacc: I agree -- I'll look at passing the URL as an option to the plugin.

PR #2827, opened 2017-03-30
Remove custom sharing logic

richaagarwal:
(GHE 716) - Building on the work @chosak did to decustomize Wagtail sharing, this PR is a first attempt at removing much of the no-longer-needed custom sharing logic. 

It also fixes a bug with copying pages (GHE 705) and hopefully fixes other bugs we don't know about yet. 

Though remaining tests pass, I'd appreciate an eye on the ones I modified/ removed to make sure I'm not accidentally hiding any issues from those changes.

Note - this PR doesn't remove database fields. I propose that we do that in a follow-up PR, (once we're sufficiently happy with these changes), along with a few other clean-up items discussed below.


higs4281:
We need a word for this. I propose richete.

I have four lines to add, [here](https://github.com/cfpb/cfgov-refresh/blob/cleanup-sharing/cfgov/ask_cfpb/models/django.py#L274-L275) and [here](https://github.com/cfpb/cfgov-refresh/blob/cleanup-sharing/cfgov/ask_cfpb/models/django.py#L278-L279), from the ask_cfpb Answer model. Be happy to push that, but I think I trust leaving the richete in your hands.

cfgov/v1/models/base.py
```diff
@@ -58,15 +53,6 @@ def shared_q(self):
     def shared(self):
```
chosak: Can `shared` and `shared_q` be cleaned up? And, if so, can this entire custom `CFGOVPageQuerySet` class be removed?

cfgov/v1/routing.py
```diff
@@ -54,7 +54,7 @@ def get_protected_url(context, page):
     if page:
```
chosak: I'm wondering if this entire `get_protected_url` method can be removed. Maybe better left for a future PR so as not to make this one any larger than it already is. But is there any more use case for this?

cfgov/v1/util/util.py
```diff
@@ -79,9 +79,9 @@ def get_secondary_nav_items(request, current_page):
     nav_items = []
     for sibling in pages:
         if page.id == sibling.id:
-            sibling = page.get_appropriate_page_version(request)
```
chosak: Is this `if/else` logic even needed anymore? In the first case (`page.id == sibling.id`) you're just assigning a page to itself. In the second case (`else`) you're doing the same thing.

cfgov/v1/models/base.py
richaagarwal: @chosak I started down that path, and it turned out to be a bit of a rabbit hole.  Should be able to clean it up, but I'd prefer to do that in v2 of clean-up.

cfgov/v1/routing.py
richaagarwal: Yeah, it doesn't seem like we need it but we do return `#` if it's not live or doesn't exist, not sure if we still need that around. Agree we can deal with this in a separate PR.

cfgov/v1/util/util.py
richaagarwal: Oops, I can remove the `else` since `sibling` doesn't need to get assigned to `sibling` obviously.   It does seem as if the if/else block isn't doing anything, but I didn't actually change that aspect of it.  If `page` and `sibling` are the same thing in the first case, then `page.get_appropriate_page_version` and `sibling.get_appropriate_page_version` should've returned the same thing, right?  Not opposed to removing it but just curious why it's like this at all.

richaagarwal:
@higs4281 Thanks for pointing that out!

cfgov/v1/models/base.py
richaagarwal: Or, happy to pair on it ... maybe I just don't know how the PageManager is supposed to work.

richaagarwal:
@chosak Good idea about the initial data script. I updated that as suggested, but didn't bother updating the documentation since we can still use that `DJANGO_STAGING_HOSTNAME` variable for now.

cfgov/v1/models/base.py
chosak: Interesting, I'm curious what made it a rabbit hole. These methods don't seem to be used anywhere. I don't think we need a custom `PageQuerySet` anymore without them.

chosak:
Any reason why `CFGOVPage.shared` and `CFGOVPage.has_unshared_changes` can't be removed as well? (This will require a migration.)

cfgov/v1/util/util.py
chosak: Hm, you're right, it does seem like those would have been the same. I really don't know.

cfgov/v1/models/base.py
richaagarwal: Updated to call `PageQuerySet` directly, thanks!

PR #2836, opened 2017-04-07
Enable linking of info unit headings

Scotchester:
Multiple stakeholders have expressed a desire to be able to put links on info unit headings (i.e., headings of Image and Text 50/50 or 25/75 modules). This PR enables that by modifying the existing "Should link image" checkbox to be "Link image and heading", which will link both the image and heading of items in that group when checked.

## Additions

- A tiny bit of Less to style the heading links with black as the default color, a la blog post previews.

## Changes

- `should_link_image` field on Image and Text groups becomes `link_image_and_heading`
- Info unit template now links the heading when that option comes in as `true`.

## Testing

1. Pull branch
2. `./cfgov/manage.py migrate`
3. `gulp styles`
4. `./runserver.sh`
5. Open or create a page with an Image & Text 25/75 or 50/50 Group module
   - Ensure that at least one of the module's children has a link defined on it
6. Check the "Link image and heading" box on that module and preview the page
7. Observe that both image and heading link to the same URL as the first link on the items

## Notes

- See GHE issue 584
- @schaferjh agreed outside of the issue that combining the image and heading link options into a single checkbox was preferable to having two separate options

## Checklist


cfgov/jinja2/v1/_includes/molecules/info-unit.html
```diff
@@ -8,56 +8,57 @@
...
    ========================================================================== #}
 
 {% macro info_unit(value) %}
 {# TODO: If value.modifier is just to switch between two layouts,
          change to a boolean parameter. #}
+{% set link_image_and_heading = value.link_image_and_heading
+                                and value.links
+                                and value.links[0]
+                                and value.links[0].url %}
+
```
richaagarwal: This is interesting.  Ideally this logic would exist in a python function on the molecule's class, but I know this is how it was before so it doesn't need to be a blocker.

sebworks: You typically wouldn't need to this (Jinja would handle it correctly).  It's only here because we are passing in an imaginary [links or null](https://github.com/cfpb/cfgov-refresh/blob/d5d0ca239e777bb4b1d0a54788ebe20ca13b5796/cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html#L259).

PR #2837, opened 2017-04-10
Switch to using Django's date filter instead of sheerlike

willbarton:
A side effect [our recent effort to remove the custom sharing code](https://github.com/cfpb/cfgov-refresh/pull/2827/commits/26e53bdd4ad786436e4457d54243323099263d85#diff-8e8a3e28d46ee12516bea4f5b50bf5a4L220) is that event starting times were showing up in UTC:

image

This is because the `get_appropriate_page_version()` method that was removed hide a quirk of timezone handling: By always getting the latest page revision from JSON and then converting it to a Wagtail Page, it was converting the UTC datetime to the configured `settings.TIME_ZONE`. When that was removed, we're left with UTC. 

Additionally, our `date` filter in Jinja2 templates was not Django's `date` filter, but our own [custom date filter from sheerlike](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/sheerlike/templates.py#L19). The sheerlike `get_date_string()` doesn't do any special timezone handling, and doesn't check `settings.TIME_ZONE` but relies on a timezone being passed in.

This PR replaces the sheerlike `date` in the Jinja2 environment with Django's `date`. Because of this, the `timezone` parameter will be ignored.

This is a quick fix for the upcoming live event. We *must* go back and revisit our date/timezone handling in templates more generally.

With this change we have the desired outcome on the event:

image

## Changes

- Replace sheerlike's `date` filter with Django's `date` filter.

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/macros/util/format/datetime.html
+++ b/cfgov/jinja2/v1/_includes/macros/util/format/datetime.html
@@ -19,11 +19,7 @@
    ========================================================================== #}
 
 {% macro format_time(datetime, timezone=none) %}
-    {% if timezone %}
-        {{- datetime | date('%I:%M %p %Z', timezone) -}}
-    {% else %}
-        {{- datetime | date('%I:%M %p') -}}
-    {% endif %}
+    {{- datetime | date('%I:%M %p %Z') -}}
 {% endmacro %}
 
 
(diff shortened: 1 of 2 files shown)


willbarton:
Closing this because I have been very silly. 

graham_chapman_colonel

PR #2838, opened 2017-04-10
Refactor datetime formatting to use Django's localtime

willbarton:
TLDR: This PR replaces adds Django's `localtime` template tag/filter to our Jinja2 environment and then refactors the `time.render()` and `datetime.format_time()` macros to use `localtime` for timezone handling and changes everywhere that `time.render()` was called with a separate timezone. This should maintain the existing expectations of rendering.

Longer:

A side effect [our recent effort to remove the custom sharing code](https://github.com/cfpb/cfgov-refresh/pull/2827/commits/26e53bdd4ad786436e4457d54243323099263d85#diff-8e8a3e28d46ee12516bea4f5b50bf5a4L220) is that event starting times were showing up in UTC:

image

This is because the `get_appropriate_page_version()` method that was removed hide a quirk of timezone handling: By always getting the latest page revision from JSON and then converting it to a Wagtail Page, it was converting the UTC datetime to the configured `settings.TIME_ZONE`. When that was removed, we're left with UTC. 

Additionally, our `date` filter in Jinja2 templates was not Django's `date` filter, but our own [custom date filter from sheerlike](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/sheerlike/templates.py#L19). The sheerlike `get_date_string()` doesn't do any special timezone handling, and doesn't check `settings.TIME_ZONE` but relies on a timezone being passed in.

With this change we have the desired outcome on the event:

image

## Changes

- Replace sheerlike's `date` filter with Django's `date` filter.

## Checklist

### DIFF
--- a/cfgov/v1/__init__.py
+++ b/cfgov/v1/__init__.py
@@ -4,18 +4,23 @@ import os
 import re
 from urlparse import urlparse, parse_qs
 
-from django.contrib.staticfiles.storage import staticfiles_storage
-from django.template.defaultfilters import pluralize, slugify, linebreaksbr
-from django.utils.module_loading import import_string
-from django.contrib import messages
 from django.conf import settings
+from django.contrib import messages
+from django.contrib.staticfiles.storage import staticfiles_storage
 from django.core.urlresolvers import reverse
+from django.template.defaultfilters import (
+    linebreaksbr,
+    pluralize,
+    slugify,
+)
+from django.utils.module_loading import import_string
+from django.utils.timezone import template_localtime
 
 from jinja2 import contextfunction, Markup
 from sheerlike import environment as sheerlike_environment
 from compressor.contrib.jinja2ext import CompressorExtension
 from flags.template_functions import flag_enabled, flag_disabled
-from .util.util import get_unique_id
+from v1.util.util import get_unique_id
 
 from wagtail.wagtailcore.rich_text import expand_db_html, RichText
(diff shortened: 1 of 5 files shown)


cfgov/v1/__init__.py
```diff
@@ -58,10 +63,12 @@ def environment(**options):
         'get_protected_url': get_protected_url,
         'get_latest_activities': get_latest_activities,
         'get_snippets': get_snippets,
+        'localtime': template_localtime,
     })
 
     env.filters.update({
         'linebreaksbr': linebreaksbr,
+        'localtime': template_localtime,
```
richaagarwal: can we just call `template_localtime` directly so we don't have to do an extra look up here?

richaagarwal: sorry assuming we have to register it, can we call it the same thing, e.g. `template_localtime: template_localtime`

willbarton: What do you mean?

willbarton: [Django includes it in Django templates as `localtime`, not `template_localtime`](https://docs.djangoproject.com/en/1.10/topics/i18n/timezones/#localtime). Here I'm including it in the Jinja2 environment with the same name Django uses in its templates.

richaagarwal: Ah okay, interesting!  Sounds good then.

PR #2845, opened 2017-04-13
Add page models to implement category-based pages

higs4281:
To replicate the Ask CFPB landing page and category-driven sub-pages, we
add the following:
- A page type for the landing page that allows full-width presentation
- A page type for category pages
- A `featured` field on the Answer model so editors can check which
  answers to feature in landing-page cards
- A `featured` filter for answers in the admin

This is prep work for front-end surgeons who will recreate knowledgebase pages as living Wagtail pages.

## Testing

To see this locally, you need to do the migration dance

```
./drop-db.sh
./create-mysql-db.sh
mysql -uroot v1 < ~/Downloads/production_django.sql
```

The new category pages are just stubs, but they're starting to look like the real thing:

category_page

This code will only appear on build.

### DIFF
--- a/cfgov/ask_cfpb/models/django.py
+++ b/cfgov/ask_cfpb/models/django.py
@@ -163,13 +163,6 @@ class Answer(models.Model):
     def spanish_page(self):
         return self.answer_pages.filter(language='es').first()
 
-    @property
-    def available_subcategories(self):
-        subcats = []
-        for parent in self.category.all():
-            subcats += list(parent.subcategories.all())
-        return subcats
-
     panels = [
         MultiFieldPanel([
             FieldRowPanel([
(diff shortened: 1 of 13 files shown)


cfgov/ask_cfpb/models/django.py
```diff
@@ -148,10 +156,18 @@ def __str__(self):
...
+    @property
+    def spanish_page(self):
+        return self.answer_pages.filter(language='es').first()
+
+    @property
     def available_subcategories(self):
         subcats = []
         for parent in self.category.all():
-            subcats += list(parent.subcategory_set.all())
+            subcats += list(parent.subcategories.all())
```
willbarton: I think you could simplify this with something like:

```
subcats = SubCategory.objects.filter(parent__in=self.category.all())
```

Edit: Actually, just re-use `available_subcategory_qs` below.

higs4281: That was some silly duplication there. Thanks!

PR #2854, opened 2017-04-19
Support wagtail-flags 2.0.0 API changes

willbarton:
Wagtail-Flags 2.0 is released and this PR updates the cfgov-refresh usage of Wagtail-Flags to be compatible. 

[I've also updated the documentation to be more cfgov-refresh specific in how we use flags.](https://github.com/cfpb/cfgov-refresh/blob/490deb2a88fc15d2fff3781da31ae482927fd08d/docs/feature-flags.md)

## Additions

- Added our existing feature flags in `settings/base.py`

## Changes

- Updated all Wagtail-Flags API calls as appropriate to the changes
- Pinned Wagtail-Flags to 2.0.0

## Testing

- In your cfgov-refresh virtual environment, `pip install git+https://github.com/cfpb/wagtail-flags@great-refactor`
- Run `python cfgov/manage.py migrate flags` to migrate flag tables
- Experiment with turning flags off and on using the new conditions described in cfpb/wagtail-flags#9!

## Checklist


willbarton:
Note: Travis won't pass until wagtail-flags 2.0.0 is released. I did that on purpose.

willbarton:
@higs4281 I have thoughts on Travis and testing: I think we should `@override_settings(FLAGS={...})` in our tests to enable/disable a flag where appropriate.

With respect to app code deploying, yup. All the migrations will go to all the places. I had some thoughts about a mechanism for flagging migrations, but that could get really, really ugly. So I think my preference is just to presume all migrations go everywhere.

I think we might want to do something different for the definition of the `WAGTAIL_ASK_CFPB` flag — my thought was to be less disruptive with env setups and make it depend on `DEPLOY_ENVIRONMENT` for now, but perhaps there are better conditions we can use for it.

cfgov/cfgov/urls.py
```diff
@@ -412,7 +429,8 @@
         url(r'^(?i)obtener-respuestas/api/autocomplete-(?P<language>es)/$',
             ask_autocomplete, name='ask-autocomplete-es'),
     ]
-    urlpatterns += ask_patterns
+
+urlpatterns += ask_patterns
```
higs4281: so ask_patterns gets defined, but the urls remain empty (or inactive?) until the flag kicks in?

willbarton: @higs4281 That's not quite right, and you've spotted a flaw here. For the shared URLs between knowledgebase and ask_cfpb we'll have to have a `flagged_url` with a fallback.

willbarton: @higs4281 currently none of these URLs overlap between the knowledgebase and ask_cfpb apps, so this code should work. However, if there are going to be URLs that overlap, we should talk about how to make that work.

In all my local testing, the flag works as expected, the knowledgebase URLs are available when `WAGTAIL_ASK_CFPB` is disabled, and the ask_cfpb URLs are available when `WAGTAIL_ASK_CFPB` is enabled.

willbarton:
@higs4281 since there's still traffic in your PR, and CCDB5 is waiting to use these changes I'm inclined to merge this. 

**Note**: I backed out the change to `INSTALLED_APPS`, so `ask_cfpb` being installed still depends on the `DEPLOY_ENVIRONMENT` variable. **Migrations will not apply to beta or prod.** So we don't have to worry about your squashing in #2913.

willbarton:
@higs4281 so as not to disrupt Ask CFPB work, I've backed out all my URL changes. Let's pair on flagging that when you get back from PyCon!

PR #2858, opened 2017-04-20
Add management commands for Akamai flushes

richaagarwal:
## Overview
- We want to be able to invoke flushes to Akamai (both fast purge, and entire site) via Jenkins; it makes sense to have these be management commands rather than defined in the Jenkins job itself 
- We want to use token-based credentials (rather than Ross's credentials) for purging the entire site, so this adds a new env var for that endpoint. Ansible PRs are up. 
- GHE 586

### DIFF
--- a/cfgov/v1/management/commands/invalidate_all_pages.py
+++ b/cfgov/v1/management/commands/invalidate_all_pages.py
@@ -1,38 +1,11 @@
-import json
-import logging
-import os
-import requests
-
-from akamai.edgegrid import EdgeGridAuth
-
+from django.conf import settings
 from django.core.management.base import BaseCommand
 
-logger = logging.getLogger(__name__)
+from v1.models.akamai_backend import AkamaiBackend
 
 
 class Command(BaseCommand):
+    help = 'Invalidates entire cfgov site'
+
     def handle(self, *args, **options):
-        auth = EdgeGridAuth(
-            client_token=os.environ.get('AKAMAI_CLIENT_TOKEN'),
-            client_secret=os.environ.get('AKAMAI_CLIENT_SECRET'),
-            access_token=os.environ.get('AKAMAI_ACCESS_TOKEN')
-        )
-        headers = {'content-type': 'application/json'}
-        payload = {
-            'action': 'invalidate',
-            'type': 'cpcode',
-            'domain': 'production',
-            'objects': [os.environ.get('AKAMAI_OBJECT_ID')]
-        }
-        resp = requests.post(
-            os.environ.get('AKAMAI_PURGE_ALL_URL'),
-            headers=headers,
-            data=json.dumps(payload),
-            auth=auth
-        )
-        logger.info(
-            'Initiated Akamai flush with response {text}'.format(
-                text=resp.text
-            )
-        )
-        resp.raise_for_status()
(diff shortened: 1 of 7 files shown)


cfgov/v1/management/commands/invalidate_single_page.py
```diff
@@ -0,0 +1,23 @@
...
+        )
+
+    def handle(self, *args, **options):
+        page_url = options['url']
+        if page_url:
+            purge_url_from_cache(page_url)
+        else:
+            raise CommandError('Must supply a page URL')
+
+        purge_url_from_cache(page_url)
```
chosak: Why is this being called a second time here?

cfgov/v1/management/commands/invalidate_single_page.py
```diff
@@ -0,0 +1,23 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--url',
+            nargs='?',
+            help='The full URL for the page you wish to invalidate'
+        )
+
+    def handle(self, *args, **options):
+        page_url = options['url']
+        if page_url:
```
chosak: Instead of defining this as an optional argument (`nargs='?'`) and then checking to see if it's set, how about doing:

```py
def add_arguments(self, parser):
    parser.add_argument('url', help='The full URL for the page you wish to invalidate')

def handle(self, *args, **options):
    page_url = options['url']
```

This will enforce that a URL is passed as an argument to the command, without requiring the `--url`. Optionally you could use `required=True` instead of `nargs='?'` if you prefer to keep the `--url` usage.

cfgov/v1/management/commands/invalidate_all_pages.py
```diff
@@ -0,0 +1,38 @@
...
+from akamai.edgegrid import EdgeGridAuth
+
+from django.core.management.base import BaseCommand
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        auth = EdgeGridAuth(
```
chosak: Does it make sense to try to somehow unify this with your [`v1.models.akamai_backend.AkamaiBackend`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/models/akamai_backend.py#L11) class, since it shares some of the same parameters/usage? Like maybe adding a `purge_all` method to that class? Not a big deal, but might be nice to avoid any duplication.

cfgov/v1/management/commands/invalidate_single_page.py
richaagarwal: That was a mistake, thanks for catching it!

cfgov/v1/management/commands/invalidate_single_page.py
richaagarwal: Addressed in https://github.com/cfpb/cfgov-refresh/pull/2858/commits/18d28b2615f06c60a68cc42bcf31b73dd60b0841. I kept the `--url` usage because then it's guaranteed to fail if someone doesn't know how to use it, as opposed to someone passing in some other parameter and it failing at the Akamai step.

chosak: I like that, thanks.

PR #2860, opened 2017-04-20
Removing padding and width from CTA

sebworks:
Removing padding and width from CTA

## Changes

- Modified `cfgov/unprocessed/css/molecules/call-to-action.less` to remove padding and width from the Call to Action molecule. This brings it back inline with our Atomic philosophy. 

## Testing

- Add a Call To Action to a Sub-landing page
- Preview and verify that it has no top / bottom margin or padding.

## Screenshots
- 
screen shot 2017-04-20 at 2 39 57 pm

## Notes

- There is essentially no margin on the CTA within the Full Width Text organism.  You will have to control spacing using the Rich Text Editor.
- Wrapping everything in a block is a mistake, if you aren't going to allow editors to control spacing for the block from Wagtail. 

## Todos

-

## Checklist

### DIFF
--- a/cfgov/unprocessed/css/molecules/call-to-action.less
+++ b/cfgov/unprocessed/css/molecules/call-to-action.less
@@ -23,11 +23,6 @@
 */
 
 .m-call-to-action {
-    padding-bottom: unit( @grid_gutter-width / @base-font-size-px, em ) * 2;
-
-    .respond-to-range(@bp-sm-min, @bp-sm-max, {
-        .grid_column( 6 );
-    });
 
     .btn {
         text-align: center;


Scotchester:
I do not like the idea of encouraging/requiring editors to put blank paragraphs in a rich text field to get adequate margin above the CTA button.

anselmbradford:
> I do not like the idea of encouraging/requiring editors to put blank paragraphs in a rich text field to get adequate margin above the CTA button.

Agreed, but it's current behavior already :\

Scotchester:
> Agreed, but it's current behavior already :\

What do you mean? Then what's the point of this PR?

jimmynotjim:
I'm confused, we're getting rid of it because "it's not atomic" which I agree with, but instead we're going to leave it up to content editors to manipulate the rich text to do as they wish? Just because folks are already doing that doesn't really justify breaking a spacing pattern. Shouldn't the padding and grid styles be moved to the parent rather than be removed all together?

Scotchester:
NM, Ans, I looked closer at the proposed changes and see what you mean. This isn't affecting the space _before_ a CTA.

But yeah, as @jimmynotjim said, the Full Width Text organism should be determining spacing between its child molecules, right?

sebworks:
The CTA is used within the Full Width Text Organism and outside it.  If it's outside it then it typically will get wrapped in a block. Inside it, I don't think it makes sense to try and control the spacing via CSS or Jinja

I would much rather have them use the Rich Text Editor, as an interim solution. Using CSS or wrapping the elements doesn't give consistency across the pages. You simply can't account for the various combinations of components with CSS or Jinja. You won't end up with consistent spacing.

The ultimate solution is to allow them to control all spacing from Wagtail using the interface. They also need the ability to ensure the spacing they are setting isn't violating a Design Manual guideline. 

The padding and width were added when we weren't wrapping everything in blocks.

jimmynotjim:
@sebworks can you link to what this is fixing currently so I can understand the context a bit more?

Scotchester:
Leaving spacing to people putting empty lines in the rich text field is not a recipe for consistency.

sebworks:
It's really just a bug @jimmynotjim

PR #2862, opened 2017-04-20
Add option to filter Related Posts by ALL tags on host page, rather than ANY ONE of the tags

Scotchester:
The need driving this work is to be able to show only posts about a specific topic that are written in Spanish. Currently, the only way to know if a blog post is in Spanish is that it has a "CFPB en español" tag. Therefore, we need to match both that tag and the topic tag.

## Additions

- Option on Related Posts module to tell it to use `AND`-based filtering
- Logic to `CFGOVPage.related_posts` that looks for said option and filters accordingly

## Testing

_Ensure that you have a DB dump newer than Sept. 13, 2016, or you will have to create a test blog post with topics "CFPB en español" and "Banking"._

1. Pull branch, migrate, runserver
1. Add a related posts module to a page, making sure to check the "Match all topics this page is tagged with" option
1. If not using a database dump newer than Sept. 13, 2016, Create a test blog post with the same topics, and test posts with only one of each of those two topics.
1. Set two topics on the page in its Settings tab: "CFPB en español" and "Banking"
1. View the page and see that the Related Posts module is returning only one post, the one with both of the same topics as the page.

## Screenshots

screen shot 2017-04-20 at 16 46 52

## Notes

- I did not add any tests for this new functionality, due to refactoring discussions happening in GHE/platform#825 

## Checklist


schaferjh:
@ielerol Want to make sure you have eyes on this, especially the Wagtail UI, and for awareness.

Scotchester:
Fair point, @willbarton! I will look into adding tests tomorrow.

cfgov/v1/models/base.py
```diff
@@ -186,7 +186,16 @@ def specific_categories_query(block, parent_slug):
...
+                    ))
+                # Apply similar logic as snippets.py's filter_by_tags method
+                # to enable AND filtering
+                if block.value['and_filtering']:
+                    for tag in self.tags.names() or []:
+                        related[search_type_name] = (
+                            related[search_type_name].filter(tags__name=tag)
+                        )
+                related[search_type_name] = \
+                    related[search_type_name][:block.value['limit']]
```
chosak: It's strange that the `limit` here is a `CharBlock` and not something like an [`IntegerBlock`](http://docs.wagtail.io/en/v1.8.1/topics/streamfield.html#integerblock). This shouldn't validate, but it does:

image

If you do this and try to preview or view the page, it 500s because `[:block.value['limit']]` doesn't work with non-integer text (the reason it works with integer text at all is somewhat surprising as well).

While you're messing with this block, how do you feel about fixing that?

(Also, minor style thing. To make the code a bit more readable, you could use a temporary variable for the query and only assign to `related[search_type_name]` at the end.)

ielerol:
The basic approach looks reasonable to me, I do have some suggestions for making the text a bit clearer. 

For the checkbox label: Match all topic tags
For the helper text: If checked, related posts will only be pulled in if they match ALL topic tags set on this page. Otherwise, related posts can match any one topic tag.

Scotchester:
Test added! I just did the simplest cloning and modification of an existing test I could. Please let me know if there's a better approach to take.

Scotchester:
> While you're messing with this block, how do you feel about fixing that?

Oy... well, I'll have to re-do my migration, but I _suppose_ it's worth it :)

cfgov/v1/models/base.py
Scotchester: Whoops, didn't thread my first reply to this. Anyway, the reason for this is probably because IntegerBlock wasn't added until Wagtail 1.6.

Scotchester:
@chosak and @ielerol, your comments have been addressed in 628e120 and 195d55f.

chosak:
@Scotchester thanks for making that change, but I realize now that I review that this will break existing blocks without a data migration (blocks that currently have a string as their value). So as not to... have to do that, can you please roll that change back? I'm sorry for adding extra work.

Thanks for adding that excellent test.

Scotchester:
@chosak Darn. Updated again.

schaferjh:
@ielerol that text change sounds good to me. @Scotchester this will be good from my POV with Lorelei's suggestions. Thanks!

schaferjh:
@Scotchester Ack! I just thought of something: how are you handling the "View more" link? Currently that link sends the user to the Activity Log to see posts based on the selected tags -- but Activity Log filter, like all the other filters, operates on OR rather than AND. Is that okay for your immediate need?

Scotchester:
I'm going to say that's fine for now :)

>

cfgov/v1/models/base.py
```diff
@@ -181,12 +181,19 @@ def specific_categories_query(block, parent_slug):
...
                         search_query
                     ).distinct().exclude(id=self.id).order_by(
                         '-date_published'
-                    )[:block.value['limit']])
+                    )
+                )
+                # Apply similar logic as snippets.py's filter_by_tags method
+                # to enable AND filtering
+                if block.value['and_filtering']:
+                    for tag in self.tags.names() or []:
```
richaagarwal: Is there a reason you have `or []` here?  If there are no tags, it will return an empty array.

cfgov/v1/tests/models/test_related_posts.py
```diff
@@ -148,6 +149,25 @@ def test_related_posts_blog_limit(self):
         self.assertNotIn('Newsroom', related_posts)
         self.assertNotIn('Events', related_posts)
 
+    def test_related_posts_and_filtering(self):
+        """
+        Tests whether related posts are retrieved if and only if they match ALL
+        of the tags on the calling page.
+        """
+
+        self.block.value['relate_posts'] = True
```
richaagarwal: This test passes regardless of whether or not this is `True` or `False` or if it's set at all. Can we get rid of it?

cfgov/v1/tests/models/test_related_posts.py
```diff
@@ -148,6 +149,25 @@ def test_related_posts_blog_limit(self):
...
+        self.block.value['relate_newsroom'] = True
+        self.block.value['relate_events'] = True
```
richaagarwal: Ditto for `relate_events`.  What is this test actually testing if these parameters don't impact the result of the test?

cfgov/v1/models/base.py
chosak: @richaagarwal I think this might be a copy of what @Scotchester did [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/models/snippets.py#L75), which is similar - in that case, because the `filter_by_tags` was being implemented in a separate method, we added the `or []` to support the case where `None` was passed in. You're right that in this case, it's unnecessary.

cfgov/v1/tests/models/test_related_posts.py
chosak: @Scotchester it might be useful to create another comparable test to this one where `self.block.value['and_filtering']` is set to `False`, to demonstrate that (I think) pages with both `'tag1'` and `'tag2'` tags are retrieved. That'd make it a bit more clear for people trying to understand how the test works.

cfgov/v1/models/base.py
Scotchester: Thanks for catching that! I'll remove it.

cfgov/v1/tests/models/test_related_posts.py
Scotchester: @richaagarwal It was intended to demonstrate that the single-tagged posts found in `posts` and `events` are available, but not retrieved when `and_filtering` is `True`.

I think @chosak's solution of creating a new test explicitly testing the `False` condition for `and_filtering` is a good idea. I could also add more detailed comments, if that would help, @richaagarwal.

Scotchester: @richaagarwal Ah, and now reading your final comment, I see where the disconnect might've been here:

> I'm pretty sure you need to add some tags to the testing scenario and then check that the pages in the result set all share the tag in question. The relate_x fields are additional filters but you don't need them for this test (or rather, it's arbitrary which one you choose. You could add blog posts, events, or newsroom pages with the relevant tag, or any combination thereof.)

The `setUp` [adds tags to child pages in each of those types](https://github.com/cfpb/cfgov-refresh/blob/00709c6f5e15b9589c8f3adc585cce4be9869a48/cfgov/v1/tests/models/test_related_posts.py#L45-L69).

Without messing with the `setUp`, in this test, I need to have `relate_[type]` be `True` for each type, so that I can show that when `and_filtering` is `True`, posts with only one of the tags are returned. In the test I plan to write with `and_filtering = False`, I could set only one of the `relate_[type]`s to `True.

Scotchester:
@richaagarwal @chosak Pushed an update in response to your comments. Thanks for the review!

cfgov/v1/tests/models/test_related_posts.py
richaagarwal: Ah, got it. Yeah I generally prefer the conditions required to understand a test to be in the test itself, but I understand the desire to reuse what's in `setUp`. Looks good, thanks for the explanation!

PR #2872, opened 2017-04-26
Remove obsolete custom link handler

chosak:
This commit removes a custom rich text link handler (registered against the `register_rich_text_link_handler` hook) that is obsolete after the move to [wagtail-sharing](https://github.com/cfpb/wagtail-sharing).

Compare the code being removed against [the core Wagtail `wagtail.wagtailcore.rich_text.PageLinkHander` class](https://github.com/wagtail/wagtail/blob/v1.8.1/wagtail/wagtailcore/rich_text.py#L21); removing this should have no impact on how rich text links are rendered.

Our custom code was doing a `urlsplit(page.url)` which will fail if the page URL is relative or `None`. These cases would never happen pre-wagtail-sharing because there were always multiple Wagtail Sites defined, but now (and in standard Wagtail usage) this cannot be assumed to be the case.

This restores the default Wagtail behavior of inserting `<a href="None">` for links that are not part of any Site root.

## Removals

- Removed custom `v1.wagtail_hooks.CFGovLinkHandler` link handler class.

## Testing

- This defaults to the standard Wagtail behavior, so I didn't think it would be appropriate to add custom unit tests for this behavior here. To test manually, you can create a page with a link to a page not under a site root (e.g. under "Trash").

## Screenshots

image

## Checklist

### DIFF
--- a/cfgov/v1/wagtail_hooks.py
+++ b/cfgov/v1/wagtail_hooks.py
@@ -3,11 +3,9 @@ import logging
 from django.core.exceptions import PermissionDenied
 from django.conf import settings
 from django.core.urlresolvers import reverse
-from django.utils.html import escape, format_html_join
+from django.utils.html import format_html_join
 from wagtail.wagtailadmin.menu import MenuItem
 from wagtail.wagtailcore import hooks
-from wagtail.wagtailcore.models import Page
-from urlparse import urlsplit
 
 from v1.util import util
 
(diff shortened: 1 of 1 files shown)


Scotchester:
Hmm...

So the intent of us duplicating this function was to be able to use relative URLs for internal Wagtail page links?

I think that might still be valuable functionality to maintain. Otherwise, if you're checking out a draft of something on the Sharing site and you click a link to a live page, you're now on the Production site, and your testing experience is not confined to the Sharing site.

Thoughts?

chosak:
@Scotchester as long as content editors use the "internal link" function in the rich text editor to insert links to other pages, these links will be rendered in HTML as relative links (as long as they're part of the same Wagtail Site).

This change was needed because the former setup always had two Wagtail `Site`s and so `page.url` would always be absolute. Now we should only ever have or need one, so `page.url` should be relative. Do you see any examples where this isn't the case?

Scotchester:
Ah, now I remember you explaining that with multiple `Site`s, `page.url` had to be absolute, and that in theory, with only one, it will not be.

So, it does seem like (to be verified once deployed) that this change will result in internal links being absolute?

> For legacy reasons we do have two sites so that Live links on content link to the right place.

I don't follow you, here.

chosak:
@Scotchester I'll follow up with you about this in another forum.

PR #2873, opened 2017-04-26
CFGOVPage.url always returns relative URLs

chosak:
This commit overrides the standard `Page.url` behavior for the `CFGOVPage` class so that it always returns a relative URL. This is to enforce use of relative URLs in links when running Wagtail in a multi-`Site` setup that shares a site root.

See comment in `v1.models.base.CFGOVPage.url` for more details.

## Additions

- Method `CFGOVPage.url` overrides standard `Page.url` method to enforce relative URLs.

## Testing

- See new unit tests in `v1.tests.models.test_base.TestCFGOVPagesAlwaysUseRelativeUrls`.

## Notes

- This is a replacement for the custom link handler that was removed in #2872. In practice this is required because we'd like all of our pages to be linked internally with relative URLs, but default Wagtail behavior is to always use absolute URLs for setups with multiple `Sites` (which we still have). See some additional discussion at GHE#832.

## Checklist


chosak:
Looking at this again I realize that this relies on the default `PageLinkHandler` calling `page.specific.url` instead of `page.url`, which it doesn't do until [v1.10rc1](https://github.com/wagtail/wagtail/blob/v1.10rc1/wagtail/wagtailcore/rich_text.py#L50) ([this Wagtail PR](https://github.com/wagtail/wagtail/pull/3425)).

@willbarton should we be optimistic that we'll upgrade soonish to Wagtail 1.10 and so move to this method of forcing relative URLs? Note that this change is marked as a "Bug fix" in the [release notes](http://docs.wagtail.io/en/latest/releases/1.10.html#bug-fixes) and so perhaps that's an indication that this is a reasonable way to approach this.

If instead we revert back to something like what was removed in #2872, we'll have to remember to update our custom handler override from `page.url` to `page.specific.url` if/when we upgrade to 1.10.

willbarton:
> @willbarton should we be optimistic that we'll upgrade soonish to Wagtail 1.10 and so move to this method of forcing relative URLs? 

Are there any reasons why we wouldn't? Let's try to do the upgrade.

Scotchester:
Do we have a sense of when 1.10 will go GM? We don't want to roll out an RC, do we?

chosak:
@Scotchester [here's](https://groups.google.com/forum/#!topic/wagtail/XSP5yDyZzsw) the announcement of the RC release (warning: Google Groups link). This was from April 20th and suggested a release "in a week or two's time".

willbarton:
@chosak what is the practical effect as we wait for Wagtail 1.10?

chosak:
@willbarton the situation in master right now, post merge of #2872, is that all links on the site to internal pages use absolute URLs. This doesn't break anything, but is perhaps suboptimal. Merging this PR in now won't do anything to change/fix that, until the site is also upgraded to 1.10.

willbarton:
@chosak Ok, that sounds alright for now, we just need to make sure our content folks are aware of that.

chosak:
@Scotchester @willbarton please see #2875 for Wagtail 1.10 upgrade (pending release, as of now it's still RC).

chosak:
Closing this PR in favor of the approach implemented in #2904. See [discussion over there](https://github.com/cfpb/cfgov-refresh/pull/2904#issuecomment-299900138) for comparison of the two approaches.

PR #2875, opened 2017-04-27
Upgrade Wagtail to v1.10.1

chosak:
This PR upgrades Wagtail from v1.8.1 to v1.10rc1 (1.10 release is expected in the next week or so, marking this PR as **hold** until that time).

See release notes for these versions:

http://docs.wagtail.io/en/latest/releases/1.8.2.html
http://docs.wagtail.io/en/latest/releases/1.9.html
http://docs.wagtail.io/en/latest/releases/1.9.1.html
http://docs.wagtail.io/en/latest/releases/1.10.html

Notable changes include:

- Support for Django 1.11 and Python 3.6!
- New ability to [compare revisions](http://docs.wagtail.io/en/latest/releases/1.9.html#revision-comparisons) (see screenshot below).
- New ability to set up [many-to-many relations for pages](http://docs.wagtail.io/en/latest/releases/1.9.html#many-to-many-relations-on-page-models). I haven't looked into this much but the docs suggest this might allow for a new/better way to do page categories @richaagarwal @Scotchester.
- Speaking of categories/tags, it looks like [tags now support spaces in them by default](http://docs.wagtail.io/en/latest/releases/1.10.html#other-features).
- Adding users through the Wagtail admin [no longer sets `is_staff`](http://docs.wagtail.io/en/latest/releases/1.10.html#adding-editing-users-through-wagtail-admin-no-longer-sets-is-staff-flag) @rosskarchner.

Includes these changes required to make our tests pass:

- Removal of custom image rendition signals which are no longer needed as of 1.10.
- Change in use of `wagtailimages.models.Filter` to support change from Django model to regular Python class in 1.9.
- Change in location of `wagtailimages.get_image_model` in 1.9.
- Change in how `HTMLBlock` is rendered due to change in block `get_context` usage in 1.10. Included HTML is always rendered verbatim and so should not need context passed in.

## Changes

- Bump of Wagtail version from `1.8.1` to `1.10rc1`.

## Testing

- Please read the release notes above to see the numerous changes involved in this upgrade. This version bump requires a good amount of testing with real data, so the best way to do that is to use a production dump:

 ```sh
 (cfgov-refresh) $ pip install -r requirements/local.txt
 (cfgov-refresh) $ ./drop-db.sh
 (cfgov-refresh) $ ./create-mysql-db.sh
 (cfgov-refresh) $ ./refresh-data.sh /path/to/dump.sql
 (cfgov-refresh) $ python cfgov/manage.py migrate --fake-initial --noinput
 (cfgov-refresh) $ ./runserver.sh
 ```

## Screenshots

New ability to compare revisions:

image

Example revision comparison:

image

## Checklist

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -41,7 +41,7 @@ suds==0.4
 tinys3==0.1.11
 unipath>=1.1,<=2.0
 vobject==0.9.1
-wagtail==1.10
+wagtail==1.10.1
 wagtail-flags==2.0.3
 wagtail-sharing==0.5
 Wand==0.4.2
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -43,5 +43,6 @@ unipath>=1.1,<=2.0
 vobject==0.9.1
 wagtail==1.8.1
 wagtail-flags==2.0.3
+wagtail==1.10
 wagtail-sharing==0.5
 Wand==0.4.2
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -41,8 +41,7 @@ suds==0.4
 tinys3==0.1.11
 unipath>=1.1,<=2.0
 vobject==0.9.1
-wagtail==1.8.1
-wagtail-flags==2.0.3
 wagtail==1.10
+wagtail-flags==2.0.3
 wagtail-sharing==0.5
 Wand==0.4.2
--- a/cfgov/v1/atomic_elements/organisms.py
+++ b/cfgov/v1/atomic_elements/organisms.py
@@ -5,6 +5,7 @@ from django import forms
 from django.apps import apps
 from django.template.loader import render_to_string
 from django.utils.encoding import smart_text
+from django.utils.safestring import mark_safe
 from django.utils.functional import cached_property
 from functools import partial
 from jinja2 import Markup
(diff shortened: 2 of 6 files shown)


sebworks:
@chosak, what do you think about using this as an opportunity to write some browser tests for the admin?

richaagarwal:
Yeah it'd be great to write tests for any of the areas we're concerned about / asking people to manually test. I'm happy to pair on that so it's not as big of a blocker.

requirements/base.txt
```diff
@@ -41,7 +41,6 @@ suds==0.4
 tinys3==0.1.11
 unipath>=1.1,<=2.0
 vobject==0.9.1
-wagtail==1.8.1
```
chosak: Oh, weird, nice catch. How did I miss updating this and just duplicate it? Minor request: could you move the `wagtail` requirement up here so as to be alphabetized properly?

richaagarwal: @chosak that was actually my bad -- it was left over from a rebase conflict. and yes, i can do that!

chosak: Thanks. Also, is it worth using version [1.10.1](https://pypi.python.org/pypi/wagtail/1.10.1) (latest on PyPI right now) for this PR? [Release notes are here](http://docs.wagtail.io/en/v1.10.1/releases/1.10.1.html), looks like just minor bug fixes.

richaagarwal: Sure! I'll test it locally now

PR #2885, opened 2017-05-01
Certify CF changes for enhancements/base

jimmynotjim:
The body font is now in CFv4 and the overflow is related to the menu, and won't be moved to CF.

Looking for feedback on how to properly note what is and isn't in Capital Framework and whether the menu organism really is the best place for that overflow rule.

## Changes

- Moved the menu related overflow rule to the menu file
- Noted the code already in CF

## Checklist

### DIFF
--- a/cfgov/unprocessed/css/enhancements/base.less
+++ b/cfgov/unprocessed/css/enhancements/base.less
@@ -19,7 +19,7 @@
 */
 
 body {
-    .webfont-regular(); // Moved to CF
+    .webfont-regular(); // Moved to CF, delete freely when migrating to CF4
 }
 
 /* topdoc
--- a/cfgov/unprocessed/css/enhancements/base.less
+++ b/cfgov/unprocessed/css/enhancements/base.less
@@ -19,11 +19,7 @@
 */
 
 body {
-    .webfont-regular();
-
-    .respond-to-max( @bp-sm-max, {
-        overflow-x: hidden;
-    } );
+    .webfont-regular(); // Moved to CF
 }
 
 /* topdoc
--- a/cfgov/unprocessed/css/organisms/mega-menu.less
+++ b/cfgov/unprocessed/css/organisms/mega-menu.less
@@ -18,6 +18,22 @@
 @mobile_trigger_ht__px: 54px;
 
 /* topdoc
+  name: Body Overflow
+  notes:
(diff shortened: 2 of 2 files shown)


cfgov/unprocessed/css/enhancements/base.less
```diff
@@ -19,11 +19,7 @@
 */
 
 body {
-    .webfont-regular();
-
-    .respond-to-max( @bp-sm-max, {
-        overflow-x: hidden;
-    } );
+    .webfont-regular(); // Moved to CF
```
Scotchester: I was confused by the fact that you left this here, even though it had moved to CF. Then I realized that it's in a later version of CF than this project has implemented. Should the comment be more explicit about that?

jimmynotjim: IDK, I was going to try and go through and document what is and isn't in CF so that whoever ends up making the change over knows what they can safely delete without having to go through everything. Does that make sense?

Scotchester: It does. Maybe something like "Moved to CF, delete freely when migrating to CF4", then?

jimmynotjim: 👍

cfgov/unprocessed/css/organisms/mega-menu.less
```diff
@@ -18,6 +18,22 @@
...
+  notes:
+    - "Hides menu overflow on mobile sizes."
+  patterns:
+    - name: Body
+      codenotes:
+        - |
+          body
+*/
+body {
+    .respond-to-max( @bp-sm-max, {
```
sebworks: could we target a lower level DOM element?

sebworks:
Jimmy, would it be possible to fix the underlying issue? I don't think it was ever meant to be a permanent thing, and it violates the idea of isolated components. It might take some time though, so it's up to you.

jimmynotjim:
@sebworks yeah, I'm not trying to refactor the site here, just make sure what should be in Capital Framework makes it there. Since this clearly does not belong in CF, I wanted to get it out of the enhancements files. I can add a note to refactor it later if you want.

sebworks:
@jimmynotjim, merge when ready.

PR #2891, opened 2017-05-03
Properly render reusable text rich text fields

chosak:
This commit fixes how the `ReusableText` snippet class renders its `RichTextField` content. It uses the [builtin Wagtail `|richtext` filter](http://docs.wagtail.io/en/v1.9/topics/writing_templates.html#rich-text-filter) that ensures that rich text content is both rendered as safe HTML as well as with expanded page links.

This addresses issue GHE#856. 

## Changes

- When rendering `ReusableText` snippets, pass their `text` property through the `richtext` filter so that links are properly expanded.

## Testing

- See newly added unit tests that verify this behavior works properly.
- A production page that requires this fix can be viewed locally at [http://localhost:8000/consumer-tools/auto-loans/](http://localhost:8000/consumer-tools/auto-loans/) (see sidebar "About us" text link).

## Screenshots

Fixed snippet link:

image

## Notes

- @richaagarwal @sebworks @Scotchester I haven't been too familiar with [these built-in Wagtail template tags](http://docs.wagtail.io/en/v1.10/topics/writing_templates.html#rich-text-filter) before, have you? Seems like they might be really useful elsewhere.

## Checklist


sebworks:
👍, this works @chosak. I do think we should revisit our use of context functions which rely upon `expand_db_html`. It seems like a very expensive process, but this isn't your problem.

chosak:
@sebworks thanks for the review. Yeah, I was thinking our use of `expand_db_html` and of `parse_links` in various places might be better replaced by this `|richtext` filter that should handle those things automatically.

PR #2892, opened 2017-05-04
Adding margin drop-downs to Call to Action

sebworks:
Adding margin drop-downs to Call to Action

## Changes

- Modified files to add margin drop-downs which allow the user to control margin. 

## Testing

- Add CTA to page
- Select margin options from drop-downs.
- Preview and verify the margin is correct.

## Screenshots

- screen shot 2017-05-03 at 11 34 46 pm

## Notes

-  This is largely experimental and the first attempt bring CF control to Wagtail.

## Todos

-

## Checklist


Scotchester:
Remind me again why we don't want to have Full Width Text control the layout of its children like CTA? Wasn't that essentially the principle we agreed upon? Parents control the layout of their children?

/cc @anselmbradford

sebworks:
>Remind me again why we don't want to have Full Width Text control the layout of its children like CTA? Wasn't that essentially the principle we agreed upon? Parents control the layout of their children?

The CTA is used within and outside the Full Width Text organism. It could be used anywhere, potentially.   If we are wrapping everything in blocks ( at the page level ) then we are essentially assigning margin by default.

sebworks:
You can try and handle it with CSS but it just doesn't scale:

```
.o-sidebar-content {
    border-width: 0;

    > [class^="o-"] + [class^="m-"],
    > [class^="o-"] + [class^="o-"],
    > [class^="m-"] + [class^="o-"],
    > [class^="m-"] + [class^="m-"]{
        margin-top: unit(@grid_gutter-width * 2 / @base-font-size-px, em);
    }

    > [class^="m-"]:last-child, > [class^="o-"]:last-child {
        margin-bottom: unit(@grid_gutter-width * 2 / @base-font-size-px, em);
    }
}
```

Scotchester:
I would much rather us write that CSS, which will gzip very well, than use the CMS in this way.

sebworks:
> I would much rather us write that CSS, which will gzip very well, than use the CMS in this way.

I disagree, we are already serving entirely too much CSS and it forces developers to write CSS for different combinations of components that they didn't anticipate.

Scotchester:
> we are already serving entirely too much CSS

I think this is debatable. It's 40 KB, out of 544 KB total on the homepage. It's 2 (of 50 total) requests. Compare with JavaScript (16 requests, 237 KB), images (21 requests, 120 KB) and fonts (4 requests, 118 KB).

> it forces developers to write CSS for different combinations of components that they didn't anticipate.

What else is there to attempt to anticipate? If our spacing rules are consistent, your clever attribute-selector-based snippet seems to cover it pretty comprehensively. We just have to add more parent elements to the initial selector.

---

Our design guidelines should be solid enough to code a consistent solution and content editors shouldn't have to confront these options and try to figure out what to do. Beyond that... what is the end state, if we start going down this road? Margin selectors on every module? That will make editing in Wagtail severely more cumbersome, especially if 99% of the time people will just be leaving the default selected.

ascott1:
@sebworks I'm curious about what the business requirements are that are driving this change. I tend to agree with @Scotchester's comments here and think we should limit configurations in the CMS when possible, but I'm not sure that I fully understand the issue. Can you both break this down to me like I'm 5 (or just a manager who doesn't write much CSS anymore)?

sebworks:
>Our design guidelines should be solid enough to code a consistent solution and content editors shouldn't have to confront these options and try to figure out what to do. Beyond that... what is the end state, if we start going down this road? Margin selectors on every module? That will make editing in Wagtail severely more cumbersome, especially if 99% of the time people will just be leaving the default selected.

If you use Chrome tools you can see that 80% of the CSS that is served is unused for most pages. In addition, the browser still has to parse all of the CSS you are sending down the wire. It's entirely too much CSS which is why we have implement things like `gulp-bless` to prevent IE9 from choking on our bloat. 

There are combinations of components on a page which you can't anticipate in advance.  They should have the ability to control the components in ways we didn't anticipate in Wagtail. I'm not concerned about overcomplicating the UI ( we can adjust it ).

jimmynotjim:
I recognized @sebworks concerns about our CSS as it is now, but I'm concerned this puts the burden on the CMS and editor, rather than on the codebase. We've gone down that road before with WordPress, and we agreed to solve it as a team with atomic design. If we start abandoning those principles now, I'm worried how that will result in the future, especially with us really just now getting off the ground with atomic design in Capital Framework. 

@Scotchester while gzipping helps incredibly with download, the browser still has to parse over 300kb of unzipped styles. We really, really need to figure this out, but I agree this isn't the place to start. 

Converting the project to CFv4 has to be a high priority if we're going to take performance seriously, otherwise we're just sticking bandaids on what is essentially dead code.

ascott1:
Hey y'all, I don't want us to veer down a performance black hole here. We all agree that reducing the size and number of assets served on cf.gov is a valiant goal and critical to improving the site's performance.

I think the next course of action is for @sebworks @schaferjh and I to discuss if this particular change is something we want to see in the CMS as product owners and then we go from there. How does that sound?

sebworks:
I'm closing this because it's apparent this isn't the direction the team wants to go in at this time. Although, I think it's a huge mistake :).

PR #2901, opened 2017-05-05
Add management command to print or email inactive user list

willbarton:
This PR adds a management command that will either list inactive users (determined by an optional number of days since last login that defaults to 90) and the date of their last login or email that list to a given set of email addresses.

GHE#837

## Additions

- `inactive_users` management command

## Checklist


cfgov/core/management/commands/inactive_users.py
```diff
@@ -0,0 +1,63 @@
...
+        User = get_user_model()
+        inactive_users = User.objects.filter(
+            is_active=True, last_login__lt=last_possible_login)
+
+        if len(inactive_users) == 0:
+            return
+
+        if len(emails) > 0:
+            self.stdout.write('Sending inactive user list to '
+                              '{}\n'.format(','.join(emails)))
```
chosak: This raises a `UnicodeDecodeError` if one of the email addresses contains a unicode character, e.g.

```sh
$ manage.py inactive_users --emails unicodë@domain.com
```

Maybe use `u'strings'` or, even better, `__from__ future import unicode_literals`?

cfgov/core/management/commands/inactive_users.py
```diff
@@ -0,0 +1,63 @@
...
+            self.send_email(emails, period, inactive_users)
+        else:
+            self.stdout.write('Users inactive for {}+ days:\n'.format(period))
+            self.stdout.write(self.format_inactive_users(inactive_users))
+
+    def format_inactive_users(self, inactive_users):
+        inactive_users_str = ''
+        for user in inactive_users:
+            inactive_users_str += '\t{username}: {last_login}\n'.format(
```
chosak: This fails (`AttributeError`) if there's a user that hasn't ever logged in. Here's a unit test that demonstrates:

```py
def test_user_never_logged_in(self):
    user = User(username='foo', last_login=None)
    self.assertEqual(
        Command().format_inactive_users([user]),
        u'foo: None\n'
    )
```

Might also be nice to test the date formatting:

```py
def test_format_last_login(self):
    last_login = datetime(2017, 5, 5, 12, 11, 10)
    user = User(username='foo', last_login=last_login)

    self.assertEqual(
        Command().format_inactive_users([user]),
        '\tfoo: 05/05/2017 12:11 p.m.\n'
    )
```

cfgov/core/management/commands/inactive_users.py
```diff
@@ -0,0 +1,63 @@
...
+                username=user.username,
+                last_login=date_format(user.last_login,
```
chosak: This also fails (`UnicodeDecodeError`) if a user has a unicode username (see above). Here's a unit test that demonstrates:

```py
def test_user_unicode_username(self):
    user = User(username=u'us\xebrname', last_login=None)
    self.assertEqual(
        Command().format_inactive_users([user]),
        u'\tus\xebrname: None\n'
    )

File: cfgov/core/management/commands/inactive_users.py [hunk 4]
```diff
@@ -0,0 +1,63 @@
...
+                                       "SHORT_DATETIME_FORMAT"))
+        return inactive_users_str
+
+    def send_email(self, emails, period, inactive_users):
+        now = date_format(timezone.now(), "SHORT_DATETIME_FORMAT")
+        subject = 'Inactive users as of {}'.format(now)
+        msg = ('The following active users have not logged in for '
+               '{period}+ days:\n'.format(period=period))
+        msg += self.format_inactive_users(inactive_users)
+        return mail.EmailMessage(subject, msg, None, emails)
```
chosak: This doesn't actually send the email message, does it? Doesn't it just create the message object? I believe you still need to call [`EmailMessage.send`](https://docs.djangoproject.com/en/1.8/topics/email/#django.core.mail.EmailMessage) to have the message sent.

cfgov/core/management/commands/inactive_users.py
```diff
@@ -0,0 +1,63 @@
...
+            is_active=True, last_login__lt=last_possible_login)
+
+        if len(inactive_users) == 0:
+            return
+
+        if len(emails) > 0:
+            self.stdout.write('Sending inactive user list to '
+                              '{}\n'.format(','.join(emails)))
+            self.send_email(emails, period, inactive_users)
+        else:
```
chosak: Minor thing, but it might be nice to always have the list of users displayed even if the email message option is chosen. This'll make it easier to review e.g. job logs in Jenkins where emails were sent out.

cfgov/core/tests/test_management_inactive_users.py
```diff
@@ -0,0 +1,77 @@
...
+        self.assertNotIn("user_2", out.getvalue())
+        self.assertNotIn("user_3", out.getvalue())
+
+    @mock.patch('core.management.commands.inactive_users.mail.EmailMessage')
+    def test_sends_email(self, mock_EmailMessage):
+        """ Test that mail.EmailMessage is called with the appropriate
+        list of users """
+        out = StringIO()
+        call_command('inactive_users', emails=['test@example.com'], stdout=out)
+        mock_EmailMessage.assert_called_once()
```
chosak: This unit test fails locally for me (and seemingly on Travis too) with an `AttributeError` due to a missing `Mock.assert_called_once`. It looks like [this is new](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once) as of the Python 3.6 version of `mock`. Perhaps that's not yet in our/any version of the Python 2 port? I notice that right now [we're pinned to](https://github.com/cfpb/cfgov-refresh/blob/master/requirements/test.txt#L8) `mock==1.3.0` even though [version 2.0.0 is available on PyPI](https://pypi.python.org/pypi/mock).

willbarton: Interesting, I've got mock==2.0.0 installed locally for some reason.

cfgov/core/management/commands/inactive_users.py
willbarton: I'm not convinced it's necessary to test Django's date formatting here.

chosak: Fair -- by "test the date formatting" I meant more, test the code in `format_inactive_users` that creates the string. Otherwise that code isn't covered.

cfgov/core/management/commands/inactive_users.py
willbarton: Fixed in 0458bb5.

cfgov/core/management/commands/inactive_users.py
willbarton: cb486b5 changes the user filtering to check for users who've joined before the period cutoff date and either have not logged in since then or have never logged in at all. 

04024b8 covers those two cases for `format_inactive_users`.

cfgov/core/tests/test_management_inactive_users.py
willbarton: I went ahead and bumped our pinned version of mock to 2.0.0 in c3b9eb5.

cfgov/core/management/commands/inactive_users.py
```diff
@@ -0,0 +1,75 @@
...
+    def handle(self, *args, **options):
+        period = options['period']
+        emails = options['emails']
+
+        last_possible_date = timezone.now() - timedelta(days=period)
+
+        User = get_user_model()
+        inactive_users = User.objects.filter(
+            Q(is_active=True),
+            Q(date_joined__lt=last_possible_date),
```
chosak: Minor: I don't think you need the `Q()` around these first two clauses.

cfgov/core/tests/test_management_inactive_users.py
```diff
@@ -0,0 +1,110 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from StringIO import StringIO
+
+from datetime import timedelta
+
+import mock
```
chosak: Nitpicky: conventional import order [seems to be to put this first](https://www.python.org/dev/peps/pep-0008/#imports).

cfgov/core/tests/test_management_inactive_users.py
```diff
@@ -0,0 +1,110 @@
...
+
+from django.contrib.auth import get_user_model
+from django.core.management import call_command
+from django.test import TestCase
+from django.utils import timezone
+from django.utils.formats import date_format
+
+from core.management.commands.inactive_users import Command
```
chosak: Nitpicky: alphabetize imports.

cfgov/core/tests/test_management_inactive_users.py
```diff
@@ -0,0 +1,110 @@
...
+        self.assertNotIn("user_5", self.get_stdout())
+
+    @mock.patch('core.management.commands.inactive_users.mail.EmailMessage')
+    def test_sends_email(self, mock_EmailMessage):
+        """ Test that mail.EmailMessage is called with the appropriate
+        list of users """
+        call_command('inactive_users',
+                     emails=['test@example.com'],
+                     stdout=self.stdout)
+        mock_EmailMessage.return_value.send.assert_called_once()
```
chosak: There's a nicer way to do this without mocks, if you're so inclined. During tests that use the `locmem` backend (which we do, [in our test settings](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/test.py#L10)), Django provides [a test outbox](https://docs.djangoproject.com/en/1.8/topics/testing/tools/#django.core.mail.django.core.mail.outbox) that you can check against without needing to mock anything. See an example test from our repo that uses this [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/tests/test_email.py#L56).

cfgov/core/tests/test_management_inactive_users.py
willbarton: I always waffle on this because for us mock is a third-party import until we're at Python 3 (then it's in `unittest.mock` anyway).

cfgov/core/tests/test_management_inactive_users.py
willbarton: I put the django imports before the local imports... I think I'd prefer to keep it that way, unless you feel strongly otherwise.

cfgov/core/management/commands/inactive_users.py
willbarton: I did it because otherwise I was passing keyword arguments before non-keyword arguments because, logically, this is the order I was thinking about them in. But it occurs to me that doesn't matter for the query itself.

cfgov/core/tests/test_management_inactive_users.py
chosak: My bad, you're right. I withdraw the comment.

PR #2904, opened 2017-05-08
Always use relative internal links in rich text fields

chosak:
This commit restores behavior that was removed in #2872.

When we insert internal Wagtail page links into rich text fields, we need them to always render as relative (not absolute) so as to maintain expected behavior across sharing/production sites. This new handler makes sure all internal Wagtail page links are always rendered as relative.

See [comments in `v1.wagtail_hooks.RelativePageLinkHandler`](https://github.com/cfpb/cfgov-refresh/compare/better-link-handler?expand=1#diff-44507cc860654b805baac659b58d2c53R118) for more details.

## Additions

- New `v1.wagtail_hooks.RelativePageLinkHandler` that forces relative internal links in Wagtail rich text fields.

## Testing

- New unit tests cover the new handler class; to manually test locally, [create a `BrowsePage`](http://localhost:8000/admin/pages/add/v1/browsepage/3/), add a `FullWidthText` organism, insert a "Content" `RichTextBlock`, then insert an internal link to another Wagtail page. You'll see that links render as relative when previewing or viewing the page even if multiple Wagtail `Site`s are defined.

## Notes

- As always, [I should have listened better](https://github.com/cfpb/cfgov-refresh/pull/2872#issuecomment-297458526) to @Scotchester.

## Checklist

### DIFF
--- a/cfgov/v1/tests/test_wagtail_hooks.py
+++ b/cfgov/v1/tests/test_wagtail_hooks.py
@@ -1,13 +1,12 @@
-import json
-
 import mock
+
 from django.test import TestCase
-from django.test.client import RequestFactory
+from wagtail.wagtailcore.models import Site
+from wagtail.tests.testapp.models import SimplePage
 
-from akamai.edgegrid import EdgeGridAuth
-
-from v1.models.browse_page import BrowsePage
-from v1.wagtail_hooks import check_permissions, form_module_handlers
+from v1.wagtail_hooks import (
+    RelativePageLinkHandler, check_permissions, form_module_handlers
+)
 
 
 class TestCheckPermissions(TestCase):
@@ -118,3 +117,49 @@ class TestFormModuleHandlers(TestCase):
         form_module_handlers(self.page, self.request, self.context)
         child.block.is_submitted.assert_called_with(self.request, 'name', 0)
 
+
+class TestRelativePageLinkHandler(TestCase):
+    def setUp(self):
+        self.default_site = Site.objects.get(is_default_site=True)
+
+    def test_nonexisting_page_returns_empty_link(self):
+        self.assertEqual(
+            RelativePageLinkHandler.expand_db_attributes(
+                attrs={'id': 0},
+                for_editor=False
+            ),
+            '<a>',
+        )
+
+    def test_page_not_in_site_root_returns_none_link(self):
+        self.assertEqual(
+            RelativePageLinkHandler.expand_db_attributes(
+                attrs={'id': 1},
+                for_editor=False
+            ),
+            '<a href="None">',
+        )
+
+    def check_page_renders_as_relative_link(self):
+        page = SimplePage(title='title', slug='slug', content='content')
+        self.default_site.root_page.add_child(instance=page)
+
+        self.assertEqual(
+            RelativePageLinkHandler.expand_db_attributes(
+                attrs={'id': page.pk},
+                for_editor=False
+            ),
+            '<a href="/slug/">',
+        )
+
+    def test_single_site_returns_relative_link(self):
+        self.assertEqual(Site.objects.count(), 1)
+        self.check_page_renders_as_relative_link()
+
+    def test_multiple_sites_still_returns_relative_link(self):
+        Site.objects.create(
+            hostname='other',
+            root_page=self.default_site.root_page,
+            is_default_site=False
+        )
+        self.check_page_renders_as_relative_link()
--- a/cfgov/v1/wagtail_hooks.py
+++ b/cfgov/v1/wagtail_hooks.py
@@ -4,7 +4,9 @@ from django.core.exceptions import PermissionDenied
 from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.utils.html import escape, format_html_join
+
 from urlparse import urlsplit
+
 from wagtail.wagtailadmin.menu import MenuItem
 from wagtail.wagtailcore import hooks
 from wagtail.wagtailcore.models import Page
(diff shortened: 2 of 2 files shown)


willbarton:
What are the implications for #2873?

cfgov/v1/tests/test_wagtail_hooks.py
```diff
@@ -118,3 +117,49 @@ def test_calls_child_block_is_submitted(self, mock_getstreamfields, mock_hasattr
...
+            '<a>',
+        )
+
+    def test_page_not_in_site_root_returns_none_link(self):
+        self.assertEqual(
+            RelativePageLinkHandler.expand_db_attributes(
+                attrs={'id': 1},
+                for_editor=False
+            ),
+            '<a href="None">',
```
willbarton: Would it be better to behave like a nonexistent page if the page is outside the site root?

cfgov/v1/wagtail_hooks.py
```diff
@@ -110,3 +113,57 @@ def register_django_admin_menu_item():
...
+                editor_attrs = ''
+
+            page_url = page.specific.url
+
+            if page_url:
+                page_url = urlsplit(page_url).path
+
+            return '<a %shref="%s">' % (editor_attrs, escape(page_url))
+        except Page.DoesNotExist:
+            return "<a>"
```
willbarton: Is it only ever when the page doesn't exist in the site root that `page.specific.url` might evaluate to `False` here?

chosak:
@willbarton good question, thanks for bringing that up. I think #2873 can probably be closed as, for now, this does the same thing. If/when this gets merged in, I'll close that one.

cfgov/v1/tests/test_wagtail_hooks.py
chosak: I kept this behavior [consistent with how Wagtail does it](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailcore/rich_text.py#L50). To clarify, it's not "the site root", it's "any site root" - meaning this only appears if a link is to a page that isn't associated with any `Site` at all (wouldn't even have an absolute URL).

Scotchester:
@chosak Hmm... judging by the changed code, #2873 looks like a more elegant solution. Are you sure you wouldn't rather go that route?

cfgov/v1/wagtail_hooks.py
chosak: Yes, if a page isn't under a site root that's been exposed via a `Site`.

cfgov/v1/tests/test_wagtail_hooks.py
willbarton: Ah, alright. Consistency with Wagtail is good... I just wonder when that would happen.

chosak:
@Scotchester I didn't listen well enough to you last time so I have to be extra thoughtful in my reply now :smile:

I think I prefer this implementation as being somewhat more explicit. This PR is limited to "when rendering page links in rich text fields, always use relative URLs". The motivation for all of these changes has generally been based on how the HTML is generated, so tying this to the rendering code seems more straightforward to me.

Changing `CFGOVPage.url` has more wide-ranging impact: for example when working in the Python shell or in other unit tests, all of a sudden your page object acts differently. It's also slightly more hidden in the possibly unlikely case that we fork page classes in future.

For both approaches, the reason we need this is due to our non-standard use of a separate domain for content editing. In the long run it would be better to move away from that setup. Keeping this customization in the realm of the rendering code might also make it easier to remove if/when we get to that point.

cfgov/v1/tests/test_wagtail_hooks.py
chosak: A recent example that happened to us was when a live page contained a link to a "deleted" page that had been moved to the Trash. As the Trash isn't under a site root, it turned the live page's link into `<a href="None">`.

chosak: A recent example that happened to us was when a live page contained a link to a "deleted" page that had been moved to the Trash. As the Trash isn't under a site root, it turned the live page's link into `<a href="None">`.

Scotchester:
Great answer :+1:

> the reason we need this is due to our non-standard use of a separate domain for content editing

For content _editing_ or for _viewing draft_ content? Do you have an idea for how we'd be able to review draft content without a different domain?

chosak:
I meant _editing_ in particular.

Keeping things somewhat vague as this is public, one approach might be to move editing/admin access to the public domain that we serve pages from while keeping the existing internal sharing domain only for viewing of draft content. This would require replacing current `Site`-based feature flags with some other mechanism (e.g. query string) as well as secure limited exposure of `/admin` on that public domain (cc @willbarton).

If we went this route, Wagtail would only have a single `Site`, and thus all `Page.url` calls would be relative everywhere without needing custom handlers.

Scotchester:
Just so I'm clear: The rendered result should be `<a href="none"> for any page that isn't published and in the current site, correct?

This is what my testing bears out.

cfgov/v1/wagtail_hooks.py
```diff
@@ -3,9 +3,12 @@
 from django.core.exceptions import PermissionDenied
 from django.conf import settings
 from django.core.urlresolvers import reverse
-from django.utils.html import format_html_join
+from django.utils.html import escape, format_html_join
+from urlparse import urlsplit
```
Scotchester: Maybe put blank lines above and below the `urlparse` import to separate the groups?

cfgov/v1/wagtail_hooks.py
```diff
@@ -110,3 +113,57 @@ def register_django_admin_menu_item():
...
+
+    This special page link handler makes it so that any internal Wagtail page
+    links inserted into rich text fields, for example:
+
+        <a linktype="page" id="123">foo</a>
+
+    always get replaced with a relative URL to that page, never an absolute
+    one, like this:
+
+        <a href="/path/to/page">foo</a>
```
Scotchester: The above explanation could maybe be a little clearer about the fact that the first example is the database representation, and the second example is the final rendering to the client.

Also, the phrasing makes it unclear to me if the second example is a relative URL or an absolute URL. Consider including an example of each?

cfgov/v1/wagtail_hooks.py
```diff
@@ -110,3 +113,57 @@ def register_django_admin_menu_item():
...
+
+    In standard Wagtail behavior, this may be an absolute URL if an
```
Scotchester: `may` ➡️ `will`

cfgov/v1/wagtail_hooks.py
```diff
@@ -110,3 +113,57 @@ def register_django_admin_menu_item():
...
+    installation has multiple Wagtail Sites. In our current custom usage we
+    have multiple Wagtail Sites (one for production, one for staging) that
+    share the same root page. So forcing the use of relative URLs would work
+    fine and allow for easier navigation within a single domain.
+
+    This will explicitly break things if users ever wanted to host some
+    additional site that doesn't share the same root page.
+
+    This code is modified from `wagtail.wagtailcore.rich_text.PageLinkHandler`.
+
```
Scotchester: Trivial: You could remove this blank line to match our typical docstring formatting

chosak:
@Scotchester that's correct. Thanks for the thorough review; I've applied your excellent style suggestions in 3202686.

PR #2905, opened 2017-05-08
Store rich text table cells like rich text fields

chosak:
The `v1.atomic_elements.organisms.AtomicTableBlock` block provides a rich text editor in the Wagtail admin, but doesn't currently store its content in the database in the same way that the built-in
`RichTextBlock` block type does. This causes some inconsistency, for example in how internal links are rendered.

This change modifies the way that table content is stored in the database, making sure that it is first cleaned using `DbWhitelister` and then restored using `expand_db_html`, consistently with how Wagtail does it for `RichTextBlock` content. See `wagtail.wagtailcore.richtext` for reference.

This change also renames and slightly refactors the `AtomicTableInput` to `RichTextTableInput` to be more explicit about how it is intended to work. The JavaScript `initAtomicTable` method is similarly renamed to `initRichTextTable`. I did not rename `AtomicTableBlock` to `RichTextTableBlock` as that would have required additional migrations.

## Changes

- Refactoring of `AtomicTableBlock` to store and retrieve database data in the same way that `RichTextBlock` does, for more consistent rendering of cell content.

## Testing

- I've added a few unit tests but unfortunately because this change is mainly only visible through JS interaction in the editor, some manual testing is required. To demonstrate this change, create a `BrowsePage`, add a `TableBlock`, and then insert (for example) a link to an internal Wagtail page. You'll see that in the Wagtail editor this table cell's HTML looks like this:

```html
<a data-linktype="page" data-id="319" data-parent-id="1" href="/">CFGov</a>
```

but when the page is rendered, it looks properly like this:

```html
<a class="" href="/">CFGov</a>
```

This is in contrast to the current behavior where the rendered HTLM also contains those various `data-` fields which should be limited to just the editor.

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/table.html
+++ b/cfgov/jinja2/v1/_includes/organisms/table.html
@@ -39,7 +39,7 @@
 
 {% if value.has_data is not defined or value.has_data %}
 {% macro _render_cell(cell) %}
-    {{ (cell or '&nbsp;') | safe | trim | linebreaksbr }}
+    {{ parse_links(cell or '&nbsp;') | safe | trim | linebreaksbr }}
 {% endmacro %}
 
 <table class="o-table
@@ -98,7 +98,7 @@
                     %}
                 {% endif %}
                 <{{ tag | safe }}{{ attributes | safe }}>
-                    {{ _render_cell(parse_links(cell or '&nbsp;')) }}
+                    {{ _render_cell(cell) }}
                 </{{ tag | safe }}>
             {% endfor %}
         </tr>
--- a/cfgov/v1/atomic_elements/organisms.py
+++ b/cfgov/v1/atomic_elements/organisms.py
@@ -6,9 +6,12 @@ from django.apps import apps
 from django.template.loader import render_to_string
 from django.utils.encoding import smart_text
 from django.utils.functional import cached_property
+from functools import partial
 from jinja2 import Markup
-from wagtail.contrib.table_block.blocks import TableBlock, TableInput
+from wagtail.contrib.table_block.blocks import TableBlock
+from wagtail.utils.widgets import WidgetWithScript
 from wagtail.wagtailcore import blocks
+from wagtail.wagtailcore.rich_text import DbWhitelister, expand_db_html
 from wagtail.wagtaildocs.blocks import DocumentChooserBlock
 from wagtail.wagtailimages import blocks as images_blocks
 from wagtail.wagtailsnippets.blocks import SnippetChooserBlock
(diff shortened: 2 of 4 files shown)


cfgov/v1/atomic_elements/organisms.py
```diff
@@ -281,32 +284,53 @@ class Media:
...
+            data, files, name
+        )
+
+        return self.json_dict_apply(value, DbWhitelister.clean)
+
+    @staticmethod
+    def json_dict_apply(value, callback):
+        value = json.loads(value)
+
+        for row in (value or {}).get('data') or []:
```
willbarton: Couldn't this be `for row in (value or {}).get('data', []):`? I had to read the multiple `or`s twice.

chosak: @willbarton I was anticipating the case where the data looks like this:

```py
value = {'data': None}
```

Then even with your suggestion you'd be evaluating

```py
for row in None:
```

which would raise `TypeError`. Remember that [`dict.get`](https://docs.python.org/2/library/stdtypes.html#dict.get) returns the value if the key is present, even if its value is falsy.

cfgov/jinja2/v1/_includes/organisms/table.html
```diff
@@ -39,7 +39,7 @@
 
 {% if value.has_data is not defined or value.has_data %}
 {% macro _render_cell(cell) %}
-    {{ (cell or '&nbsp;') | safe | trim | linebreaksbr }}
+    {{ parse_links(cell or '&nbsp;') | safe | trim | linebreaksbr }}
```
richaagarwal: What was happening before if cell was `None`?

chosak:
@willbarton @richaagarwal I forgot to add the new unit tests I referenced in this PR; I've added them in 9a58ae4.

cfgov/jinja2/v1/_includes/organisms/table.html
chosak: The same as now; it's replaced by a blank space.

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -281,32 +284,53 @@ class Media:
         js = ['bureau-structure.js']
 
 
-class AtomicTableInput(TableInput):
+class RichTextTableInput(WidgetWithScript, forms.HiddenInput):
```
willbarton: Is there a reason you changed this to not inherit from `TableInput` anymore?

chosak: @willbarton mainly because previously we weren't really using `TableInput` ([there's not much more to it](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/table_block/blocks.py#L14)), and skipping its `render` in favor of its grandparent's anyway. This more explicitly defines our own widget that uses rich text.

PR #2911, opened 2017-05-12
Update Google Analtyics/Tag Manager/Optimize code to latest recommendations

Scotchester:
Our implementation had fallen out of step with [the latest recommendations from Google](https://support.google.com/360suite/optimize/answer/7359264) on using these services all together. This PR rectifies that.

The apparent effect of being out of step is that we are recording far fewer Optimize sessions than we should be (roughly ¼ of all traffic to a page, as measured by Analytics) is getting counted as a completed Optimize session).

**This needs to be deployed as a hotfix ASAP**, because we're starting an important Optimize test on Wednesday, and we'd like a little time to test and see that it resumes performing as expected with this new code.

## Additions

- Page hiding snippet to prevent flicker effect when alternate code is loaded by Optimize
- Snippet to explicitly load Optimize (as opposed to letting Google Tag Manager load it, as before)

## Testing

You can't, practically, other than confirming that the new code outputs correctly when the template is rendered.

## Notes

- [This blog post](http://www.lunametrics.com/blog/2017/03/02/install-google-optimize/#3-how-to-install-the-optimize-snippet) from a group with which @kelleyholden did some training with a few months ago 

## Todos

- Once deployed, wait a day or so for new test data to roll in and confirm that the number of Optimize sessions go back to normal.

## Checklist


cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -122,12 +122,43 @@
     If you come across a script that makes a convincing case to be included in
     the head, then file an issue or PR to discuss including it.
 -->
-    {# Google Tag Manager #}
+
+    <!-- Google Optimize page-hiding snippet -->
+    <style>
```
sebworks: Why is this necessary? Do we know why any of this is necessary outside of a blog post?

sebworks:
Does this decrease the overall performance of the site? How does this tie into the work that @cfarm is doing?

cfgov/jinja2/v1/_layouts/base.html
Scotchester: I'm unclear if your question is specific to the page hiding snippet, or the concept of A/B testing in general. What do you mean by "outside of a blog post"?

Scotchester:
> Does this decrease the overall performance of the site? How does this tie into the work that @cfarm is doing?

I don't _think_ so, because it's not loading any new scripts (the only code it's adding to a page load is the literal code added in this PR). But I don't think there's any real way to find out without testing it in production.

I'm certainly interested in the work @cfarm is planning to do on measuring and improving our performance, but I don't think we can wait for that to make this change. We've promised our stakeholders an A/B test beginning on Wednesday, and we need to be sure we're getting the right data.

cfgov/jinja2/v1/_layouts/base.html
sebworks: We should have some level of understanding about he underlying mechanisms. I'm just curious about what is driving these analytics changes.

Scotchester: I tried to explain it succintly in the PR description, but I can give some more details if you like.

Three weeks ago, we removed the explicit call to Google Tag Manager from the head of You Have the Right: https://[GHE]/CFGOV/you-have-the-right/pull/27

At that time, the number of Google Optimize sessions for the alternate headline test we were running on that page dropped significantly, from roughly 90-95% of the total traffic to that page (as reported by Analytics), to roughly 25% of it.

In trying to find out why, we discovered that our implementation for using all three of these services together (a single Google Tag Manager snippet in our template) is not the recommended implementation by Google (an Analytics snippet that also calls in Optimize, with its standard Analytics pageview event removed, _plus_ the standard GTM snippet).

Adding the page-hiding snippet is unrelated to that particular issue, but it was an additional thing they recommend doing to prevent a flash between rendering the original page content and rendering the test variation.

Does that help clear it up?

sebworks: yeah, just seems like we are going from analytics change to the next analytics change, without having a full understanding about the previous change. Is there going to be a change in another month because we don't understand something about this change?

Scotchester: Honestly, it wouldn't surprise me. It's pretty black box stuff, and Google seems to be changing stuff frequently, especially in regards to Optimize. I'm not sure what we can do better, but maybe @cfarm's research will help us figure it out.

mistergone: @Scotchester is correct about this being fairly "black box" stuff. I'd imagine it would require considerable work to determine exactly why bugs like this occur. That's not to say it's fine that we have no idea exactly what's going on, but at present I have no suggestions for viable alternatives to letting Google Analytics and Tag Manager pull us around by the nose a bit. :slightly_frowning_face:

anselmbradford:
Sounds like [they recommend](https://support.google.com/360suite/optimize/answer/7359264#tagging-best-practices) jQuery come before the Google code, so [this](https://github.com/cfpb/cfgov-refresh/blob/221627246cb97fa02d5dcefa89ae0610cf1cd5d5/cfgov/jinja2/v1/_layouts/base.html#L191) should move above it.

mistergone:
Google Analytics/Tag Manager gets crazier every year. Still, reviewing the support page linked about, it at least makes sense to me. We should maybe figure out a date and time to regroup on this and see if it fixes things and if we can make sure it's not interfering with performance or other functions.

Other than that, 👍

Scotchester:
@anselmbradford That should only be in cases where we need jQuery as part of an Optimize experiment. To my knowledge we've not needed that yet. Can you confirm, @kelleyholden?

kelleyholden:
I dont think we've needed that yet, but i think @virginiacc might know better than me.

anselmbradford:
> @anselmbradford That should only be in cases where we need jQuery as part of an Optimize experiment. To my knowledge we've not needed that yet. Can you confirm, @kelleyholden?

That's correct, it's a recommended best practice, not a requirement. If you're in there you could move it, but doesn't need to hold up the PR.

virginiacc:
@kelleyholden @Scotchester We have been using jQuery in the OAH email signup tests, but it's not something we're likely to do very often.

Scotchester:
I guess I don't see the harm in doing it. It's already in the `head`. I'll make that change and merge.

PR #2913, opened 2017-05-14
Add routable page to deliver es answers by tag

higs4281:
For Spanish only, we need to replicate the current cfgov page that
searches for answers by tag. Since we're not using Elasticsearch
faceting in this way in Wagtail, we need a routable page that can serve
subpages showing Answer search results for a given tag.

This PR also delivers a new Spanish home page, `cfgov/es/`, that will
serve as a parent page for Ask CFPB Spanish content. The parent page
itself is not joining Wagtail yet, because it is deemed out of scope
for the migration project. So the live `/es/` will remain a Wordpress page for now.

To test the new pages, you'll need to do the migration dance:

```
./drop-db.sh
./create-mysql-db.sh
gunzip -c ~/Downloads/production_django.sql.gz | mysql -uroot v1
./runserver.sh
```
Edit your default site so that it points to localhost and port 8000, then you should be able to view Spanish pages and follow their associated tags. Example:

- http://localhost:8000/es/obtener-respuestas/que-es-una-hipoteca-es-99/ should show a Spanish answer with sidebar tags that include 'puntos.' Follow the `puntos` link to find other answers with connections to that tag.
- Our new Spanish landing page  
http://localhost:8000/es/obtener-respuestas/  
can be followed to Spanish category pages:  
http://localhost:8000/es/obtener-respuestas/categoria-adquirir-credito/
- Spanish search is also wagging its tail:  
http://localhost:8000/es/obtener-respuestas/buscar/?q=hipotecas

##### Notes
- This PR squashes Django migrations, so beware if you've pulled in earlier ask_cfpb migrations.
- The big knowledgebase migration ends with an index update, which requires Elasticsearch to be running. But it isn't needed for this PR.
- The tag results page is an unstyled stub.


cfgov/ask_cfpb/scripts/migrate_knowledgebase.py
```diff
@@ -142,9 +142,49 @@ def fix_tips(answer_text):
...
+
+    def get_or_create_es_parent_page():
+        """Create the Spanish /es/ parent placeholder page"""
+        text_intro_stream_value = [
+            {'type': 'text_introduction',
+             'value': {
+                 'heading': 'Placeholder for future Spanish home page',
+                 'links': []}}]
+        # this unserved archived page is squatting on the /es/ slug
+        dinero = CFGOVPage.objects.get(id=51)
```
Scotchester: Actually, _this_ unpublished page in CFGov is squatting on `/es` for the purposes of being able to serve [MAYG Spanish](https://www.consumerfinance.gov/es/el-dinero-mientras-creces/) and [Tax Time Saving Spanish](https://www.consumerfinance.gov/es/opciones-de-ahorro-con-el-reembolso-de-sus-taxes/): [Prod Wagtail domain]/admin/pages/3637/

Scotchester:
Is the thing I commented on not a concern at all?

cfgov/ask_cfpb/scripts/migrate_knowledgebase.py
willbarton: @Scotchester that doesn't look like a problem. It gets the actual `/es` page you're referring to and creates the "obtener-respuestas" child under it.

Scotchester: I'm a little confused by what the need for L163–166 is, then.

cfgov/ask_cfpb/scripts/migrate_knowledgebase.py
```diff
@@ -142,9 +142,49 @@ def fix_tips(answer_text):
...
+        es_parent.has_unpublished_changes = True
+        stream_block = es_parent.specific.header.stream_block
+        es_parent.header = StreamValue(
+            stream_block,
+            text_intro_stream_value,
+            is_lazy=True)
+        revision = es_parent.save_revision()
+        es_parent.save()
+        revision.publish()
+        return es_parent
```
Scotchester: So if I follow this correctly, it's going to modify and then publish my ghost page? For what purpose?

cfgov/ask_cfpb/scripts/migrate_knowledgebase.py
willbarton: We have a page with an 'es' slug that's under "Archived Pages", not "CFGov". Because our `get_or_create_page` function gets a page by slug independent of the parent, it is returning that page, rather than the 'es' page you're referring to or creating an 'es' page in the correct root. That's what @higs4281 is trying to get around with those lines (@higs4281 please correct me if I'm getting that wrong).

Scotchester: I think I see now. Would a better approach be to modify `get_or_create_page` to optionally look under a particular parent? Relying on `id` 51 to be there seems brittle.

willbarton: I think it should be fine.

We should absolutely modify `get_or_create_page`, but I think doing that (and dealing with potential side-effects) is outside the scope of this PR.

cfgov/ask_cfpb/scripts/migrate_knowledgebase.py
higs4281: It's publishing the new /es/ page, not your ghost page.
eventually the new /es/ page will serve cfgov/es.

higs4281:
@Scotchester @willbarton 
The problem is more than just `get_or_create`. The save() and save_revision() functions call a full_clean operation that checks URL uniqueness using the base Wagtail Page model, so if I try to create the new parent with a temporary slug and then switch it to `es`, save() and/or save_revision() raise ValueError for slug duplication.

Here's my workaround: I set the ghost page's slug to 'es-dinero' so that page-building can proceed for the 'es' spanish root and its immediate children. After those operations are done, I restore the slug `es` to the ghost page. 

Everyone's happy. At least until the next time someone wants to create a page with 'es' as the slug.

willbarton:
> Using an /es/ URL node to identify language for individual pages is probably going to be troublesome in Wagtail. We might want to consider other identifiers, such as the language attribute shared by all CFGOVPage subclasses, or a URL node suffix.

Or, you know, [use the l10n tools that we already have](https://docs.djangoproject.com/en/1.11/topics/i18n/translation/)... but that's a well-worn soap box.

willbarton:
@higs4281 going down this rabbit hole, [I don't see anything in base Wagtail's `full_clean` or `save` that checks for global uniqueness of slugs](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailcore/models.py#L406-L486).

cfgov/ask_cfpb/scripts/migrate_knowledgebase.py
willbarton: Are you sure about this? Running the migration script on a production database, the CFGov/es/ is still the page that @Scotchester created, with its children (the spanish Money as You Grow and Tax Time Saving Spanish) and the addition of the Ask CFPB landing page. It's not creating a new one.

The page that @Scotchester is talking about is under CFGov, not Archived Pages

chosak:
@willbarton @higs4281 as the shameful author of [the code responsible for global slug lookup](https://github.com/cfpb/cfgov-refresh/blame/master/cfgov/v1/util/migrations.py#L11), I'd support refactoring that to work properly. I know you don't want to hold this PR up, but would it greatly simplify your work here?

cfgov/ask_cfpb/scripts/migrate_knowledgebase.py
Scotchester: Yeah, that's what I'm talking about. In prod, there is already a child of CFGov named `/es`, parent of https://www.consumerfinance.gov/es/el-dinero-mientras-creces/ and https://www.consumerfinance.gov/es/opciones-de-ahorro-con-el-reembolso-de-sus-taxes/

higs4281:
@chosak This really has been the only snag, and I don't think it's a serious one.  
This migration is essentially a onetime, disposable job, and I'd prefer not to widen scope when we're this close to finishing.

@willbarton Ah -- right you are. I was looking at `clean()`, which is not part of full-clean but *is* the error message I kept running into:

https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailcore/models.py#L434-L437

willbarton:
@higs4281 Hmmm, but `clean()` still appears to check within the parent using [`_slug_is_available()`](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailcore/models.py#L390).

higs4281:
Unfortunately this rabbit hole is holding up feature flags, @virginiacc's massive plate of work and getting our migrations squashed.

willbarton:
@higs4281 can you just address @Scotchester's questions about /es? I originally thought the behavior with a production database (keeping the existing CFGov/es) was expected, but that doesn't match your response above.

I don't think this is holding up feature flags. I'm planning to go ahead and merge that PR so it's not blocking CCDB from using the new flags. I don't think you have to worry about migrations — whether the app is installed still depends on `DEPLOY_ENVIRONMENT=='build'`.

cfgov/ask_cfpb/scripts/migrate_knowledgebase.py
higs4281: Thank you, @Scotchester for bringing this up. 
This was my misunderstanding about the /es/ root as it exists in production. I need to revisit this whole scheme.

higs4281:
I'm holding off on this PR until we're straight about the 'es' root, what it serves and where it lives.

higs4281:
Many thanks to @Scotchester for saving me from my own impatience.

The Ask migration doesn't need to create a ghost Spanish root, because it's already there.

@chosak Can you please review the change to v1/util/migrations.py? I think this is what we'd normally want: to check page existence based on a specific parent class, instead of based on the global wagtail Page class. Does this stray from the original intent?

chosak:
@higs4281 maybe I'm missing something, but grepping the code I don't see anywhere else that `get_or_create_page` is used. I think it may have been used at one point but then maybe that usage was removed? Regardless, if you're the only user, you should feel free to modify as you see fit.

It seems like perhaps you don't need that level of configurability, though, since you're doing something very specific (always just creating all pages under a certain place). FWIW, more recently I've been eschewing creating pages this way (using `objects.create`) in favor of just constructing the page and then using `add_child` on the parent.

higs4281:
@chosak Thanks. `add_child` looks like a better pattern going forward, as noted in source:
https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailcore/models.py#L448-L450

higs4281:
@Scotchester We've stopped harassing the squatters. Any other concerns?

Scotchester:
That's all I had :)

PR #2917, opened 2017-05-16
Create skeleton for CCDB5

JeffreyMFarley:
This is a very, _very_ preliminary pull request.  It was created to elicit feedback from the technical stakeholders.  We are looking for feedback on:

* Methodology
  - [x] Is this how we create a new child url?
  - [x] Is this how we use flags in the application

## Additions

- Skeleton pages for the CCDB5 Single Page Application and REST API

## Removals

- None

## Changes

- `urls.py`
- `base` settings
- ~vars menu~

## Testing

- None yet

## Todos

- ~Determine the real URL and not `platzhalter`. (Picked for it's unique word and easy to replace)~

## Checklist

@sephcoster @amymok


amymok:
Looks good to me, but since I don't have the cfgov-refresh context, I can't comment on those.

amymok:
Adding @AdamZarger

chosak:
@JeffreyMFarley can you please rebase against the `master` branch (no `cfpb:`) so we can see only the changes you're introducing, without the 1,000+ additional commits?

JeffreyMFarley:
@chosak Since this is a new feature, I created a feature branch `ccdb5` and set that as the base

cfgov/cfgov/urls.py
```diff
@@ -279,6 +280,13 @@
         name='cckbyo'),
     # Form csrf token provider for JS form submission
     url(r'^token-provider/', token_provider),
+
+    url(r'^platzhalter/$', TemplateView.as_view(
+        template_name='ccdb5_landing_page.html')),
+    url(r'^platzhalter/api/v1/_suggest', CCDB5.suggest),
+    url(r'^platzhalter/api/v1/export', CCDB5.export),
+    url(r'^platzhalter/api/v1/(?P<id>[0-9]+)$', CCDB5.document),
+    url(r'^platzhalter/api/v1/', CCDB5.search),
```
willbarton: I would recommend feature flagging the URLs along with the mega menu below.

I'd also recommend going ahead and using the existing CCDB landing page URL here, but because it's an `include()` (and our flagged urls can't fall back on includes yet) it'll be a little wonky:

```python

flagged_url('CCDB5_RELEASE ', 
            r'^data-research/consumer-complaints/',
            include_if_app_enabled('complaintdatabase', 'complaintdatabase.urls')),
            state=False,
            fallback=TemplateView.as_view(template_name='ccdb5_landing_page.html'))
```

Which will serve the include if the flag is `False` and will "fall back" to the `TemplateView` if it is `True`.

I am very close to merging #2854 which will provide a [much nicer pattern for feature flagging a set of URLs like the `api/` URLs](https://github.com/cfpb/cfgov-refresh/blob/490deb2a88fc15d2fff3781da31ae482927fd08d/cfgov/cfgov/urls.py#L406-L433).

```python
with flagged_urls(`CCDB5_RELEASE`) as url:
    ccdb5_patterns = [
        url(r'^platzhalter/api/v1/_suggest', CCDB5.suggest),
        url(r'^platzhalter/api/v1/export', CCDB5.export),
        url(r'^platzhalter/api/v1/(?P<id>[0-9]+)$', CCDB5.document),
        url(r'^platzhalter/api/v1/', CCDB5.search)
    ]
urlpatterns += ccdb5_patterns
```

chosak:
@JeffreyMFarley you'll also need to add this new app (`ccd5`) to the list [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/db_router.py#L3) so that the servers on content/www share the same database tables for it. @rosskarchner

JeffreyMFarley:
@chosak I can do that, but we aren't using a database for this app, is it still required?

rosskarchner:
@JeffreyMFarley you're correct, it's only relevant for apps that need the RDBMS

JeffreyMFarley:
New commits use @willbarton Wagtail Flags and re-use the Complaint database URLs

Before
-------
screen shot 2017-05-19 at 10 30 17 am

Set the Flag
--------------
screen shot 2017-05-19 at 10 29 26 am

After
------
screen shot 2017-05-19 at 10 31 03 am

JeffreyMFarley:
@rosskarchner @sephcoster @Scotchester any more comments?  Otherwise, I will merge around 11 EDT

cfgov/cfgov/urls.py
```diff
@@ -289,6 +300,16 @@
...
 ]
 
+with flagged_urls('CCDB5_RELEASE') as _url:
+    apiBase = '^data-research/consumer-complaints/api/v1'
+    ccdb5_patterns = [
+        _url(apiBase + '/_suggest', CCDB5.suggest),
+        _url(apiBase + '/(?P<id>[0-9]+)$', CCDB5.document),
+        _url(apiBase, CCDB5.search)
+    ]
+urlpatterns += ccdb5_patterns
```
Scotchester: This is more of a Django n00b question than a code review comment, but why isn't this stuff part of `urlpatterns` above?

JeffreyMFarley: This is the new format @willbarton communicated to me for loading a block of conditionally-flagged urls.  I don't think the `with` block works inside of an array definition

willbarton: Instead of each of these being its own `flagged_url('CCDB5_RELEASE`, …)` line, this way we can flag a bunch of URLs in the same context. It's an attempt to make multiple URLs flagged with the same flag a little easier to read.

Scotchester: Gotcha; that makes sense. I was just thrown by the separation of the single `flagged_url` and the batch of `flagged_urls`. Is it possible to combine them into `flagged_urls`?

willbarton: Because of the use of a fallback view for `^data-research/consumer-complaints/`… not really right now. It would be nice, and I'll happily take suggestions on how to make it so. 

It's possible with https://github.com/cfpb/wagtail-flags/pull/13 we could use a pair of includes for `^data-research/consumer-complaints/`, and fallback on the current in `with flagged_urls('CCDB5_RELEASE', fallback=include(old_ccdb_urls)) as url`. But this is about the only way to do these right now.

Scotchester: Good enough for me 👍

PR #2920, opened 2017-05-18
Always fall back to CFGOVImage's alt text if the ImageBasic block has no alt value

Scotchester:
In instances where we are using the `ImageBasic` block to choose an image for a module, we were not falling back to the alt text set on the `CFGOVImage` itself in the image manager if the `ImageBasic` had no value in its `alt` field. Content editors were confused by this, expecting that the `CFGOVImage` `alt` would be used if one was not set on the `ImageBasic` block. See: GHE/CFGOV/platform/issues/694

This PR changes the behavior of each of those blocks to meet that expectation.

Thanks to @chosak for pairing with me on this!

## Additions

- A new Jinja2 tag, `image_alt_value(ImageBasic)` that takes an `ImageBasic` block and returns the block's `alt` value, if it exits, or the chosen `CFGOVImage`'s `alt` value, if it exists, or an empty string, if neither exist.
- Tests for that tag

## Changes

- Module templates that render ImageBasic blocks use the new `image_alt_value` tag.
- We now override Wagtail's default alt text of an image filename to be the value of the CFGOVImage's `alt` field. In `images.py`, the first addition overrides this for image insertion in rich text fields, and the second addition overrides this for instances of calling them directly, enabling the simplifications in `full-width-text.html` and `post-preview.html`.

## Testing

1. Pull this branch
1. Run the server
1. Upload an image and put something in its `alt` field
1. Add that image to an FCM, Image Inset, Image & Text, Full Width Media, Post Preview, or Snippet List, without setting the `alt` field on the block
1. View the page and see that it has output the `alt` field from the image itself
1. Test other combinations of setting or not setting both alt fields 

## Notes

- In preparing this work, I noticed that [`posts-paginated.html`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/posts-paginated.html) and [`post-macros.html`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/post-macros.html) no longer seem to be used. Wasn't sure if removing them should be part of this PR or a new one.

## Checklist


cfgov/jinja2/v1/_includes/organisms/full-width-text.html
```diff
@@ -30,7 +30,7 @@
                 {% set media = image(block.value, 'original') %}
                 <img class="m-full-width-media_image"
                      src="{{ media.url }}"
-                     alt="{{ block.value.alt or media.alt }}">
+                     alt="{{ media.alt }}">
```
chosak: Why doesn't this use `image_alt_value`?

cfgov/jinja2/v1/_includes/organisms/post-preview.html
```diff
@@ -105,7 +105,7 @@
             <div class="o-post-preview_image-container">
                 <img class="o-post-preview_image"
                      src="{{ photo.url }}"
-                     alt="{{ post.preview_image.alt if post.preview_image.alt else '' }}">
+                     alt="{{ photo.alt }}">
```
chosak: Why doesn't this use `image_alt_value`?

cfgov/v1/models/images.py
```diff
@@ -84,10 +84,19 @@ def resize(self, size):
         operation.run(mock_image, image=None, env={})
         return mock_image.width, mock_image.height
 
+    @property
```
chosak: I remember us putting this in during pairing; did you leave it here purposefully? I thought we'd decided this didn't actually address the need.

cfgov/cfgov/settings/base.py
```diff
@@ -154,6 +154,7 @@
         'OPTIONS': {
             'environment': 'v1.environment',
             'extensions': [
+                'v1.jinja2tags.images',
```
chosak: :+1: for encapsulating Jinja tags into separate modules instead of dumping into the `v1.environment` list.

cfgov/jinja2/v1/_includes/organisms/full-width-text.html
Scotchester: Because it's not an `ImageBasic` block, so you can't pass the right thing into the tag.

cfgov/jinja2/v1/_includes/organisms/post-preview.html
Scotchester: Because it's not an `ImageBasic` block, so you can't pass the right thing into the tag.

cfgov/v1/models/images.py
Scotchester: I believe it's necessary for the two places you commented above which don't use `ImageBasic`.

chosak: I thought all this did was stick a default value into the editor when you're adding an image embed to a rich text block. The Wagtail user still has the ability to modify this to be whatever they want, or blank. It doesn't actually change what gets called on page render.

Scotchester: Ah, sorry. I thought you were commenting on the addition lower in this file. Yes, you're correct about what it does. It doesn't directly address the root issue being worked on, but I think it's still valuable to set that default value when inserting an image into rich text to the CFGOVImage alt value, rather than its title.

Do you see a downside to leaving it in?

chosak: Nope, that sounds good to me.

cfgov/jinja2/v1/_includes/organisms/post-preview.html
sebworks: We should eliminate that difference. You should always be able to call the context function and get the alt value if it exists, or '' if it doesn't.

sebworks:
I think we should move away from using context functions for things like this.  Many people don't even know that context functions exist and there is no documentation for them.  You should be able to write 
``` {{ photo.alt }} ``` and get the correct alt tag:

This seems like a lot of code and configuration to just be able to get an alt tag.

File: cfgov/jinja2/v1/_includes/organisms/post-preview.html (hunk 2 above)
Comment (Scotchester) [inline cfgov/jinja2/v1/_includes/organisms/post-preview.html]: Hmm... I suppose we could do that if we altered the function to accept either an ImageBasic object or nothing at all. What do you think about doing that, @chosak?

**Comment by chosak** [comment]
@sebworks I agree in principle; the problem is that due to our use of `render_stream_child` instead of Wagtail's build-in `get_context` on blocks, we don't have an easy way to set custom properties on blocks. It would be much, much better if we could define, say, `ImageBasic.get_context` which would add an `alt` into the template context.

Fixing this is a much larger task. I think defining the context function is a reasonable approach without the larger refactor, and at this code has tests and is defined in a nicely encapsulated module.

File: cfgov/jinja2/v1/_includes/organisms/post-preview.html (hunk 2 above)
Comment (chosak) [inline cfgov/jinja2/v1/_includes/organisms/post-preview.html]: @Scotchester is it nothing at all, or is it just the image rendition instance returned from

```html
{% set photo = image(post.preview_image, 'original') %}
```

I feel okay about modifying `image_alt_value` to support that as an alternate input type.

Scotchester: Oh yeah, I guess we'd need to pass it the rendition. I'll try that out.

sebworks:
@chosak, what is preventing us from overriding the `alt` property to return what we want?

chosak:
@sebworks we don't actually have a bound instance of the block in the context, we just have a representation of its value. When we render blocks with `render_stream_child`, the `value` that gets set in the context comes [straight from how Wagtail renders blocks](https://github.com/wagtail/wagtail/blob/v1.8.1/wagtail/wagtailcore/blocks/base.py#L214) based on the value that gets pulled out of the database. If we could override that `get_context` method (which we could if we used `{% include_block %}` instead of `render_stream_child`), we could add or replace properties based on a bound instance of the block.

cfgov/jinja2/v1/_includes/organisms/post-preview.html
Scotchester: @sebworks Updated the function. Thanks for the good suggestion!

PR #2928, opened 2017-05-22
Tox environment cleanup

chosak:
This commit cleans up and simplifies the list of available tox environments in `tox.ini`. There's now a single supported main environment (`py27`) which is run with `tox` whether or not you have the `TOXENV` environment variable set.

Documentation in the `tox.ini` file now describes the following available commands:

- `tox -e fast`: run all project unit tests, use the same virtual environment as standard `tox`, don't recreate the virtualenv each time. Also uses special Django settings that skip unneeded migrations.
- `tox -e acceptance`: use the same virtualenv but run browser tests instead of Python unit tests. Also skips unneeded migrations.
- `tox -e acceptance-fast`: same as `acceptance`, but don't recreate the virtual environment each time.
- `tox -e optional`: using a separate virtual environment (named, e.g. `py27-optional`) install optional public Django apps and run their unit tests as well.
- `tox -e optional-fast`: using the same virtualenv as `optional`, run all unit tests but skip unneeded Django migrations.

This change also moves some Travis environment variables into the testing environment for simplification, to avoid needing a separate Travis environment.

In theory this reorganization should make it easier in future to add support for Python3 while maintaining all of these various environments.

## Changes

- `tox.ini` cleanup.

## Testing

- Try running any of the `tox` commands described above. Note that, for example, after you run `tox` once, you can run any of the `-fast` environments without needing to recreate it.

## Checklist

### DIFF
--- a/run_travis.sh
+++ b/run_travis.sh
@@ -8,7 +8,7 @@ if [ "$RUNTEST" == "frontend" ]; then
     gulp "test:unit"
     gulp "test:coveralls"
 elif [ "$RUNTEST" == "backend" ]; then
-    tox -e travis
+    tox
     flake8
     coveralls
 fi
--- a/tox.ini
+++ b/tox.ini
@@ -1,69 +1,67 @@
 [tox]
-envlist = py27,py34
-skipsdist = True
+envlist=py27
+skipsdist=True
 
 [testenv]
-recreate = True
-# TODO: Pull in environment variables from .env instead of re-defining them.
-setenv =
-    GOVDELIVERY_ACCOUNT_CODE = fake_account_code
-    DJANGO_SETTINGS_MODULE = cfgov.settings.test
-    ES_PORT=9200
-    ES_HOST=localhost
-    SHEER_ELASTICSEARCH_INDEX=content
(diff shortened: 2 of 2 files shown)


sebworks:
why not have  `tox -e acceptance fast` which would be in line with `tox -e fast` ?

sebworks:
would it be possible to use the verbose flag when running `tox`?

chosak:
@sebworks unfortunately [I don't see a way](http://tox.readthedocs.io/en/latest/config.html) to include verbosity config in the `tox.ini` file. You can run with the `-v` flag which adds a bit more detail to the output.

Regarding running `tox -e acceptance fast`, are you looking to somehow combine two configuration sections into a single run? Unfortunately I couldn't figure out a nice way to do that. I was trying to reproduce that by providing `tox -e acceptance-fast` as a configuration section that brings in the important parts of `acceptance` (the commands) and `fast` (the virtualenv behavior).

sebworks:
> Regarding running tox -e acceptance fast, are you looking to somehow combine two configuration sections into a single run?

yeah, but we can come back to that.

PR #2929, opened 2017-05-23
Half and third width link blob group headings fix

contolini:
This is an attempt to fix https://[GHE]/CFGOV/platform/issues/860.

We want 30px between headings in link blob groups and 15px between headings and paragraph text. The tricky part is that headings are optional so there's no guarantee that a h2, h3 or h4 will be present. This PR removes h2's bottom padding, adds 30px to the top of h3s and h4s, and adds 15px to the top of paragraphs.

## Screenshots

screen shot 2017-05-22 at 9 53 57 pm

screen shot 2017-05-22 at 9 57 53 pm

## Checklist

### DIFF
--- a/cfgov/unprocessed/css/organisms/info-unit-group.less
+++ b/cfgov/unprocessed/css/organisms/info-unit-group.less
@@ -50,7 +50,7 @@
     // 30px between headings and 15px between headings and paragraphs.
     // See https://[GHE]/CFGOV/platform/issues/860
     h2 {
-      margin-bottom: 0;
+      .u-mb0;
       + .content-l {
         h3 {
           margin-top: unit( @grid_gutter-width / @size-iii, em );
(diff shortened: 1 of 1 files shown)


sebworks:
@contolini, I just wrap the element in `o-info-unit-group_header` and target that.

jimmynotjim:
@sebworks there isn't always an h2 above the h3 or h4, as seen on the homepage and [here](https://www.consumerfinance.gov/educational-resources/youth-financial-education/) so we had to use the sibling rule. This is not at all atomic, and honestly not very clean, but it's the best we could come up with that doesn't start affecting other situations negatively, hence the inline notes @contolini left. I'd definitely like to see us address this further in the future when we break apart info-units from the layout.

Scotchester:
Per @contolini's comment on CFGOV/platform/issues/860, closing this PR in favor of an "Organism Intro" solution.

PR #2944, opened 2017-05-25
Run acceptance tests on Travis w/headless browser

chosak:
This commit adds the running of browser tests to our TravisCI integration, using the headless mode of Google Chrome. Currently this requires a custom installation of `google-chrome-beta` (version 59)
although this functionality should be coming to the stable release soon.

The various Protractor-related files have been modified to look for a `GOOGLE_CHROME_BINARY` environment variable; if this is set, running the browser tests (using `tox -e acceptance` or `gulp test:acceptance`) will try to invoke that path in headless mode.

Note: this is a suboptimal way to configure that headless mode is desired, but I've punted on a better way pending some ongoing discussions with @sebworks about a larger refactor of Protractor config.

To test locally on MacOS, download the beta version of Google Chrome:

https://www.google.com/chrome/browser/beta.html

Then set the environment variable like:

```
GOOGLE_CHROME_BINARY=/path/to/Google\ Chrome\ Beta\ App.app/Contents/MacOS/Google\ Chrome
```

## Additions

- New acceptance testing matrix entry for Travis, and support for running Protractor with a headless browser.

## Removals

- Removed `gulp test:unit:server` to standardize on use of `tox` for running Python unit tests.

## Testing

- Look at the Travis logs! Follow steps above to reproduce with a local headless browser.

## Notes

- Unfortunately there seems to be an issue with calling `setSize` to set the browser size when using the headless mode with chromeDriver; I think this might be because of the need for `Xvfb` which is due to [this open issue](https://bugs.chromium.org/p/chromedriver/issues/detail?id=1772). See also [this open issue to Travis](https://github.com/travis-ci/travis-ci/issues/7650#issuecomment-302467188).

## Checklist

### DIFF
--- a/test/browser_tests/conf.js
+++ b/test/browser_tests/conf.js
@@ -21,7 +21,6 @@ function _paramIsSet( param ) {
  */
 function _chooseSuite( params ) {
 
-
   var paramsAreNotSet = !_paramIsSet( params.browserName ) &&
                         !_paramIsSet( params.version ) &&
                         !_paramIsSet( params.platform );
@@ -36,7 +35,10 @@ function _chooseSuite( params ) {
   // This will make it so that setting the browser/platform flags
   // won't launch several identical browsers performing the same tests.
   var capabilities = defaultSuites.essential;
-  if ( paramsAreNotSet && useSauceCredentials ) {
+
+  if ( envvars.HEADLESS_CHROME_BINARY ) {
+    capabilities = defaultSuites.headless;
+  } else if ( paramsAreNotSet && useSauceCredentials ) {
     capabilities = defaultSuites.full;
   }
 
@@ -159,7 +161,7 @@ function _copyParameters( params, capabilities ) { // eslint-disable-line comple
 
 var config = {
   baseUrl:       environment.baseUrl,
-   cucumberOpts: {
+  cucumberOpts: {
     require:     'cucumber/step_definitions/*.js',
     tags:        false,
     format:      'pretty',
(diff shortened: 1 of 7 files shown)


sebworks:
> Unfortunately there seems to be an issue with calling setSize to set the browser size when using the headless mode with chromeDriver

You might be able to use the command line flag `--window-size`, like --window-size=1200,3000`.  A list of command line flags is here: http://peter.sh/experiments/chromium-command-line-switches/

sebworks:
lol lol, awww, this is too damn good.

test/browser_tests/conf.js
```diff
@@ -203,17 +205,20 @@ var config = {
...
+    if ( !envvars.HEADLESS_CHROME_BINARY ) {
+      browser.driver.manage().window().setSize(
+        windowWidthPx,
+        windowHeightPx
+      );
+
+      // Set default windowSize parameter equal to the value in settings.js.
+      browser.params.windowWidth = windowWidthPx;
+      browser.params.windowHeight = windowHeightPx;
+      browser.params.windowSize = String( windowWidthPx ) +
```
sebworks: I don't think you need to use `String` here. I would try  = ${windowWidthPx},{windowHeightPx}. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

or just `windowWidthPx + ',' + windowHeight`

chosak: Thanks, that's a good suggestion. I didn't modify this code at all, just modified the indent here to only get called if not running in headless mode.

Using `--window-size` is a good suggestion. In the existing setup, we always pass a window size if not provided (based on a default set in `environment.windowWidthPx` and `environment.windowHeightPx`). I'm not sure what the behavior is if you don't pass a size at all; presumably it opens at the browser default size.

I can definitely see wanting/needing to run tests against particular sizes, to test, say, mobile rendering. So there might be some more experimentation needed here to see if you can set a size as long as it's below whatever the "maximum window size" is on Travis, or if setting any size is unsupported as long as the use of Xvfb is needed.

PR #2946, opened 2017-05-25
Tidy up smoke tests

Scotchester:
Our lists had gotten a little out of date with new canonical URLs and a little out of sync with each other.

## Changes

- Default base now uses HTTPS
- Updated YMYG and Library URLs to use their new `/educational-resources` parent

## Testing

1. `cd cfgov/scripts`
2. `python http_smoke_tests.py --full -v`

## Notes

- If you run this against beta.consumerfinance.gov before the next database refresh from prod (which happens nightly, I think?), it will fail because the new Library Resources page isn't in the Beta DB yet. Should we do anything to protect the smoke tests from failing in cases like that?

## Checklist

### DIFF
--- a/cfgov/scripts/http_smoke_test.py
+++ b/cfgov/scripts/http_smoke_test.py
@@ -30,7 +30,7 @@ parser.add_argument(
 
 FULL = False
 TIMEOUT = 10
-BASE = 'http://www.consumerfinance.gov'
+BASE = 'https://www.consumerfinance.gov'
 
 FULL_RUN = [
     '/',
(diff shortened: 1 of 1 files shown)


Scotchester:
I won't be in tomorrow, so feel free to merge without me, if approved.

cfgov/scripts/http_smoke_test.py
```diff
@@ -102,14 +102,15 @@
 SHORT_RUN = [
     '/',
     '/es/',
-    '/complaint/',
+    # ('/es/obtener-respuestas/buscar'
```
richaagarwal: What're the commented out ones you added for?

Scotchester: I was just trying to get the totality of the list at parity with the `LONG_RUN` list, which is what the original intent seems to have been.

PR #2948, opened 2017-05-26
Cucumber tests

richaagarwal:
I duplicated the tests we had for sublanding filterable page, for browse filterable page. Not sure if there's a better way to represent those. 

I also did some clean-up.

I haven't been able to get the multiselect test working yet so I removed that from this branch.


test/browser_tests/cucumber/features/filterable-page.feature
```diff
@@ -0,0 +1,18 @@
+
+Feature: Filterable
+  As a user of cf.gov
+  I expect to see all results when I haven't selected a filter
+
+  Scenario: Sublanding filterable page
+  	Given I goto a sublanding filterable page
+  	When I do not select a filter on the sublanding filterable page
```
richaagarwal: I added "on the sublanding filterable page" and "on the browse filterable page" to make these steps unique; otherwise it complains.  This isn't really ideal -- I don't think we need this level of detail here, and it seems impractical that every step in a test will be unique across all tests. Any thoughts @sebworks ?

sebworks: yeah, we need to manage state within a scenario, I am working on this. I think think for now just make the step definition /I do not select a filter on the (.*) page/. We have the ability to use regexes in the step definitions.

test/browser_tests/cucumber/features/filterable-page.feature
```diff
@@ -0,0 +1,18 @@
...
+  	Then I should see the first result, sfp child 0
+  	Then I should see the last result, sfp child 9
+  	Then I should see the right number of results on the sublanding filterable page
+
+  Scenario: Browse filterable page
+  	Given I goto a browse filterable page
+  	When I do not select a filter on the browse filterable page
+  	Then I should see the first result, bfp child 0
+  	Then I should see the last result, bfp child 9
```
sebworks: Please don't abbreviate, it just makes it more difficult to to comprehend the statements.

sebworks:
@richaagarwal, we actually need the `shared_objects` directory. We are using these objects as mixins across pages.  So every page could have a filterable list control, video player, ..etc.. You can add them to you page like `Object.assign( this, videoPlayer )`.

PR #2961, opened 2017-05-31
Looser coupling between HUD API and code in cfgov/legacy/views/

rosskarchner:
## Additions

- DJANGO_HUD_API_ENDPOINT setting, which can be populated from the
environment. The default matches what we use in production.
- a useful initial value for development is added to .env_SAMPLE

## Changes

- the new setting is used to connect to the HUD API. The previous code had hardcoded "http://localhost" which doesn't work when developing on localhost, and depended on Django's URL reversing, which wouldn't work in a hypothetical future where the HUD API was running as a standalone service.

## Testing

- After adding something like `export HUD_API_ENDPOINT=http://localhost:8000/hud-api-replace/` to your own .env file, (and re-sourcing it), you should be able to successfully search for zip codes at /find-a-housing-counselor/

## Checklist

### DIFF
--- a/cfgov/legacy/views/housing_counselor.py
+++ b/cfgov/legacy/views/housing_counselor.py
@@ -2,6 +2,7 @@ import os
 import re
 import requests
 import sys
+from urlparse import urljoin
 
 from django.conf import settings
 from django.core.urlresolvers import reverse
@@ -42,12 +43,8 @@ class HousingCounselorView(TemplateView):
 
         Queries a locally running django-hud API.
         """
-        api_path = reverse(
-            'hud_api_replace:index',
-            kwargs={'zipcode': zipcode}
-        )
+        api_url = urljoin(settings.DJANGO_HUD_API_ENDPOINT,zipcode)
 
-        api_url = 'http://localhost' + api_path
         response = requests.get(api_url)
         response.raise_for_status()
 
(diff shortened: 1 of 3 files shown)


cfgov/legacy/views/housing_counselor.py
```diff
@@ -40,14 +41,10 @@ def get_context_data(self, **kwargs):
...
         """Return list of housing counselors closest to a given zipcode.
 
-        Queries a locally running django-hud API.
+        Queries a local or remote django-hud API
         """
-        api_path = reverse(
-            'hud_api_replace:index',
-            kwargs={'zipcode': zipcode}
-        )
+        api_url = urljoin(settings.DJANGO_HUD_API_ENDPOINT,zipcode)
```
chosak: This might fail flake8.

cfgov/legacy/views/housing_counselor.py
```diff
@@ -2,6 +2,7 @@
 import re
 import requests
 import sys
+from urlparse import urljoin
```
chosak: Would you mind alphabetizing this with the other third-party imports below (under `django.views.generic`)?

rosskarchner: Even though it's in the standard library?

chosak: Oh, fair point. I withdraw.

PR #2993, opened 2017-06-06
Updates environment configuration filename

anselmbradford:
## Changes

- Updates `test/browser_tests/environment.js` to `test/browser_tests/environment-test.js` so that there aren't vaguely confusing imports like:

```
var environment = require( './environment.js' );
var envvars = require( '../../config/environment' ).envvars;
```
- Moves nested functions for `onPrepare` and `getMultiCapabilities` into top-level private functions and adds jsdocs.
- Updates variable syntax to use ES6 declarations.
- Fixes trivial jsdoc errors.

## Testing

1. `gulp test:acceptance` should run no probs.

### DIFF
--- a/test/browser_tests/conf.js
+++ b/test/browser_tests/conf.js
@@ -1,13 +1,13 @@
 'use strict';
 
-var environment = require( './environment.js' );
-var envvars = require( '../../config/environment' ).envvars;
-var defaultSuites = require( './default-suites.js' );
+const environmentTest = require( './environment-test' );
+const envvars = require( '../../config/environment' ).envvars;
+const defaultSuites = require( './default-suites.js' );
 
 /**
  * Check whether a parameter value is set.
  * @param {string} param The value of a parameter.
- * @returns {Boolean} Whether a parameter value is undefined or not,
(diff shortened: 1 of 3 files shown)


test/browser_tests/conf.js
```diff
@@ -151,77 +151,88 @@ function _copyParameters( params, capabilities ) { // eslint-disable-line comple
...
+  if ( !envvars.HEADLESS_CHROME_BINARY ) {
+    browser.driver.manage().window().setSize(
+      windowWidthPx,
+      windowHeightPx
+    );
+
+    // Set default windowSize parameter equal to the value in settings.js.
+    browser.params.windowWidth = windowWidthPx;
+    browser.params.windowHeight = windowHeightPx;
+    browser.params.windowSize = String( windowWidthPx ) +
```
sebworks: Not sure you need String because it should automatically get converted to a string.

anselmbradford: I might leave it since it makes it clearer that the instances of `windowWidthPx` and `windowHeightPx` on the lines above it are numbers not strings. I didn't update this, that's verbatim what it was before I moved it.

PR #2998, opened 2017-06-07
Ask CFPB deploy config

higs4281:
Prep for ask_cfpb deployment
See GHE/ConsumerExperience/information-architecture/issues/308

Preps code and a flag for final deployment. 
We're ready to install the app everywhere, but pages won't show up yet.

To do: add flag code to megamenu template to handle the change to an  `/ask-cfpb/` link.

### DIFF
--- a/cfgov/ask_cfpb/views.py
+++ b/cfgov/ask_cfpb/views.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 import json
+import re
 from urlparse import urljoin
 
 from bs4 import BeautifulSoup as bs
@@ -132,3 +133,23 @@ def ask_autocomplete(request, language='en'):
                 'url': result.url}
                for result in sqs[:20]]
     return JsonResponse(results, safe=False)
+
+
+def redirect_ask_search(request):
+    # Get selected facets
+    selected_facets = request.GET.get('selected_facets')
+    if selected_facets is None:
+        raise Http404
+
+    # Get the category
+    category_match = re.match(r'category_exact:(?P<category>[^&]+)',
+                              selected_facets)
+    if category_match is None:
+        raise Http404
+
+    category = category_match.groupdict().get('category')
+    if category is None:
+        raise Http404
+
+    return redirect('/ask-cfpb/category-{category}'.format(
+                    category=category), permanent=True)
--- a/cfgov/ask_cfpb/views.py
+++ b/cfgov/ask_cfpb/views.py
@@ -3,11 +3,12 @@ import json
 from urlparse import urljoin
 
 from bs4 import BeautifulSoup as bs
-from django.shortcuts import get_object_or_404, redirect, render
-from django.http import HttpResponse, Http404, JsonResponse
 from haystack.query import SearchQuerySet
 from haystack.inputs import Clean
 from wagtail.wagtailcore.models import Site
+
+from django.http import HttpResponse, Http404, JsonResponse
+from django.shortcuts import get_object_or_404, redirect, render
 
 from ask_cfpb.models import (
     Answer,
--- a/cfgov/cfgov/urls.py
+++ b/cfgov/cfgov/urls.py
@@ -413,7 +413,7 @@ if settings.DEBUG:
     except ImportError:
         pass
 
-if settings.DEPLOY_ENVIRONMENT != 'build':
+if settings.DEPLOY_ENVIRONMENT not in ['build', 'beta']:
     kb_patterns = [
         url(r'^(?i)askcfpb/',
             include_if_app_enabled(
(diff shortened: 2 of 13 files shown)


cfgov/cfgov/urls.py
```diff
@@ -413,7 +422,7 @@
     except ImportError:
         pass
 
-if settings.DEPLOY_ENVIRONMENT != 'build':
+if settings.DEPLOY_ENVIRONMENT not in ['build', 'beta']:
```
willbarton: I'm going to see if using https://github.com/cfpb/wagtail-flags/pull/13 and an `include()` for old/new Ask URLs removes the need for this check here.

cfgov/cfgov/urls.py
```diff
@@ -413,7 +415,8 @@
     except ImportError:
         pass
 
-if settings.DEPLOY_ENVIRONMENT != 'build':
+
+if settings.DEPLOY_ENVIRONMENT not in ['build', 'beta']:
```
willbarton: Because we add `r'^(?i)askcfpb/'` as [a flagged URL below](https://github.com/cfpb/cfgov-refresh/pull/2998/files#diff-e45111e649c8f0371d5367b7351f1dd8R433), this doesn't actually seem to work. It'll need to be a fallback for that flagged URL.

cfgov/cfgov/urls.py
```diff
@@ -413,53 +415,98 @@
...
+                r'^(?i)askcfpb/',
+                include(ask_cfpb_english_redirects),
+                fallback=include_if_app_enabled(
+                    'knowledgebase', 'knowledgebase.urls')),
+    flagged_url('WAGTAIL_ASK_CFPB',
+                r'^es/obtener-respuestas/',
+                include(ask_cfpb_spanish_patterns),
+                fallback=include_if_app_enabled(
+                    'knowledgebase', 'knowledgebase.babel_urls')),
+]
```
willbarton: So, this is a fairly big change, and it breaks up the new ask URLs but I think it's the only way we can get a clean knowledgebase/ask_cfpb transition with the feature flag for URLs they share. We'll group them together once we retire the feature flag after launch.

What's going on here is this: cfpb/wagtail-flags#13 introduces fallback `include()`s, which means for `askcfpb/` and `es/obtener-respuestas/` we can include our new URLs if the flag is enabled, or fallback on the old knowledgebase URLs if it's disabled. I've put the English and Spanish URLs *that are shared* (the english URLs are just redirects) in separate lists above. This lets us `include()` them and fallback on the `include_if_app_enabled()` for the knowledgebase URLs. The new URLs that we're introducing are below and simply enabled with the `WAGTAIL_ASK_CFPB` flag (that hasn't changed).

For this to work right now (until a new release of Wagtail-Flags, which I'll make as soon as I know this works), you'll have to install it in your cfgov-refresh virtualenv manually:

```
pip install git+https://github.com/cfpb/wagtail-flags.git@fallback-include
```

@higs4281 if you have a chance, it'd be awesome if you could make sure this makes sense.

@virginiacc could you confirm that the URLs that should work do work? Since the redirects are permanent, you'll probably want to use incognito mode to try them, otherwise Chrome will cache them until browsing data is cleared (yes, that is the voice of experience 😉 ).

virginiacc: @willbarton The landing and question page redirects are working perfectly for me, but not the category pages (`askcfpb/search/?selected_facets=category_exact:auto-loans`). Maybe that's a local issue, though?

willbarton: @virginiacc I see the problem. [This line](https://github.com/cfpb/cfgov-refresh/pull/2998/files#diff-e45111e649c8f0371d5367b7351f1dd8R465) tries to include the GET parameters in the URL matching. I'll write a simple view to handle that and do the redirect.

willbarton: @virginiacc Can you give it another try? I think I've fixed that.

cfgov/ask_cfpb/views.py
```diff
@@ -131,3 +133,23 @@ def ask_autocomplete(request, language='en'):
...
+                              selected_facets)
+    if category_match is None:
+        raise Http404
+
+    category = category_match.groupdict().get('category')
+    if category is None:
+        raise Http404
+
+    return redirect('/ask-cfpb/category-{category}'.format(
+                    category=category), permanent=True)
```
willbarton: @higs4281 if you have a chance could you review this? Is `selected_facets:category_exact` the only thing we're interested in with old search?

cfgov/cfgov/urls.py
virginiacc: @willbarton The category page redirect is working for me with existing Ask category urls if I add a trailing slash to `search`

willbarton:
I'm going to close this. I don't think there's a clean way to make these URLs work with flags at the moment.

PR #3003, opened 2017-06-07
Fix date-dependent unit test that could break on occasion

chosak:
The unit test [`v1.tests.test_filterable_list_form.TestFilterableListForm.test_filter_by_tags`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/tests/test_filterable_list_form.py#L40) (introduced in #2448) was relying on [`FilterableListForm.get_page_set`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/forms.py#L102) returning pages in a
particular order. That method returns pages ordered by `date_published` [which is a Django `DateField`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/models/learn_page.py#L48).

Two pages that are published on the same day won't necessarily always return in the same order. This commit changes this test so that it doesn't rely on any particular order.

## Changes

- Refactor of unit test to avoid dependence on publishing order.

## Testing

1. Unit tests continue to pass.

## Checklist

### DIFF
--- a/cfgov/v1/tests/test_filterable_list_form.py
+++ b/cfgov/v1/tests/test_filterable_list_form.py
@@ -48,10 +48,10 @@ class TestFilterableListForm(TestCase):
         publish_page(page2)
         publish_page(page3)
         form = self.setUpFilterableForm(data={'topics': ['foo', 'bar']})
-        page_set = form.get_page_set()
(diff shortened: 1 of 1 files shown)


richaagarwal:
👍🏽

PR #3009, opened 2017-06-09
Update babel and webpack

anselmbradford:
Prepares dependencies for babel-core@7.

## Changes

- Update `babel-core` from `6.22.1` to `6.25.0`, which includes backported issues from babel-core@7.
- Update `babel-preset-es2015` from `6.22.0` to `6.24.1`.
- Update `babel-preset-stage-2` from `6.22.0` to `6.24.1`.
- Update `babel-loader` from `6.2.10` to `7.0.0`, which drops support for older node and webpack.
- Update `webpack` from `2.2.1` to `2.6.1`, which includes bugfixes.

## Testing

1. Throw out `node_modules`
2. Run `./frontend.sh`
3. Site should build fine.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -260,14 +260,14 @@
       "resolved": "https://registry.npmjs.org/babel-code-frame/-/babel-code-frame-6.22.0.tgz"
     },
     "babel-core": {
-      "version": "6.22.1",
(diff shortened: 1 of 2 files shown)


sebworks:
We should write some tests to ensure all the gulp tasks work when upgrading modules.

PR #3023, opened 2017-06-13
Add Ask CFPB redirects to convert query-string-based pages

higs4281:
Current askcfpb category pages are built via query strings, as are raw
answer searches. This adds a view to handle redirects from old category 
query-string URLs to the new category pages, which don't depend on query strings.

## Testing

Check that redirects work for some old URLs, such as:
- http://localhost:8000/es/obtener-respuestas/buscar?selected_facets=category_exact:comprar-una-casa, which should be sent to http://localhost:8000/es/obtener-respuestas/categoria-comprar-una-casa/
- http://localhost:8000/askcfpb/search/?selected_facets=category_exact:bank-accounts-and-services, which should be sent to http://localhost:8000/ask-cfpb/category-bank-accounts-and-services/
- http://localhost:8000/askcfpb/search/?q=tuition&btnSearch=, which should go to http://localhost:8000/ask-cfpb/search/?q=tuition


cfgov/ask_cfpb/views.py
```diff
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 import json
+# import re
```
willbarton: Can we remove the commented import?

cfgov/ask_cfpb/views.py
```diff
@@ -131,3 +134,38 @@ def ask_autocomplete(request, language='en'):
...
+        facets = request.GET.getlist('selected_facets')
+        if not facets:
+            raise Http404
+        # Get the category, if there is one
+        for facet in facets:
+            if cat_string in facet:
+                category = facet.replace(cat_string, '')
+        if not category:
+            raise Http404
+        if language == 'en':
```
willbarton: Since you take a default `language='en'`, would you want to make this condition the default return value too, so it always returns?

```python
if language == 'es':
    return redirect('/es/obtener-respuestas/categoria-{category}'…)

return redirect('/ask-cfpb/category-{category}'…)
```

willbarton: Or maybe raise an error of some kind that an supported language was given.

higs4281: Great idea on returning the default by default.

cfgov/ask_cfpb/views.py
higs4281: Ja!

willbarton:
Sneakily updating your PR title for more changelog-friendliness before you merge 😛

PR #3025, opened 2017-06-13
Remove universal jquery reference

anselmbradford:
GTM tags and variables used across the site have had jquery references removed, so this removes it from the base template. 

## Removals

- Remove universal jquery references.

## Testing

1. `./frontend.sh` and check the site over. Outside apps (OAH, KBYO) should handle jquery themselves, if I understand correctly.

cc @kelleyholden

### DIFF
--- a/cfgov/jinja2/v1/_layouts/base.html
+++ b/cfgov/jinja2/v1/_layouts/base.html
@@ -117,13 +117,6 @@
     the head, then file an issue or PR to discuss including it.
 -->
 
-    {# TODO: jQuery is provided in the head to satisfy
-             Google Tag Manager (GTM) & Optimize requirements.
-             Ideally GTM would handle its own dependency management
-             and initializing jQuery could be moved to the footer.
-             (Ideally, also, we will move away from jQuery entirely.) #}
-    <script src="{{ static('js/jquery.min.js') }}"></script>
-
     {% if flag_enabled('AB_TESTING', request) %}
     <!-- Google Optimize page-hiding snippet -->
     <style>
--- a/config/webpack-config.js
+++ b/config/webpack-config.js
@@ -15,10 +15,6 @@ var JS_ROUTES_PATH = '/js/routes';
 var COMMON_BUNDLE_NAME = 'common.js';
 
 var modernConf = {
-  // jQuery is imported globally in the HTML head section in base.html,
-  // so it needs to be defined here as an external script to ignore for
-  // unmet dependency references.
-  externals: { jquery: 'jQuery' },
   cache: true,
   context: path.join( __dirname, '/../', paths.unprocessed, JS_ROUTES_PATH ),
   entry: scriptsManifest.getDirectoryMap( paths.unprocessed + JS_ROUTES_PATH ),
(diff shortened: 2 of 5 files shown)


jimmynotjim:
I have a task to add sortable tables to Wagtail, but the installed version of Capital Framework [requires jQuery](https://github.com/cfpb/capital-framework/blob/f9adf768ec4e66d201b62a221c44b5c8489863d7/src/cf-tables/src/cf-tables.js#L8). Does the Platform team have a recommendation how to move forward now that jQuery is being removed from the project?

cfarm:
@jimmynotjim ooh, we were making this change with the understand that there were no cfgov-refresh dependencies on jQuery. This is still the case, right - what you're talking about is adding a feature from  Capital Framework to cfgov-refresh that is dependent on jQuery, correct?

I would recommend it be refactored to use vanilla JS and we make sure that future Capital Framework features are not dependent on jQuery, as a it's a major priority for us to remove jQuery from the cfgov-refresh pages. It's a pretty big performance hit right now and removing it will save us ~90kb in page weight on Wagtail pages.

sebworks:
There are some references that I added when doing OAH pop stuff. I'll push to this branch and remove them.

jimmynotjim:
>  what you're talking about is adding a feature from Capital Framework to cfgov-refresh that is dependent on jQuery, correct?

Not exactly. Sortable Tables have been in Capital Framework for a while, similar to Expandables, and I thought they were in use, just not within the Wagtail interface. Removing jQuery entirely may break them where they are used (unless it's one of the places we're leaving it, I'm not even sure where they are in use).

> I would recommend it be refactored to use vanilla JS and we make sure that future Capital Framework features are not dependent on jQuery

The latest version of Capital Framework did remove the jQuery requirement, and both JS dependent components use @sebworks Atomic Component code. I know we're in the middle of making that move, but I don't know what to do about Sortable Tables in the mean time. Guess I'm looking for some guidance here from the team before moving forward.

jimmynotjim:
FOLLOW UP: 

Looks like Sortable Tables have been in [CF since late 2015](https://github.com/cfpb/cf-tables/commits/gh-pages/src/js/cf-tables.js), maybe @mistergone knows where they are currently used.

sebworks:
@jimmynotjim, so the best solution for now is just to use the[ sortable table](https://raw.githubusercontent.com/cfpb/AtomicComponent/master/examples/table-sortable.js) shell and modify it where appropriate to fit into cfgov-refresh.  I think we are already poly-filling for class list so you would just need to setup the events.

anselmbradford:
@jimmynotjim or deliver jquery with sortable tables. We just want to remove the universal jquery install, but individual dependencies could install jquery themselves, if they need it.

anselmbradford:
> i don't mind that for external modules but we shouldn't encourage CFPB modules to create backdoor ways for using jQuery. Otherwise, what is the point of removing it from our repo?

True 👍

jimmynotjim:
Sounds like back-porting the latest cf-tables code into this repo is the way to go for my task.

As far as existing uses, I can't find any references to Sortable Tables in this repo. Unless someone speaks up 👍  from me on 🔥  jQuery.

sebworks:
I'm getting `jQuery` errors from the GTM code on pages with a youtube player.

anselmbradford:
Should be gone now. Updated tags are now published according to @kelleyholden.

PR #3029, opened 2017-06-14
Update build dependencies

anselmbradford:
Updates build process dependencies. Notably, gulp-bless, didn't generate sourcemaps correctly, and the update fixes that. 

## Changes

- Updates `gulp-bless` from `3.2.1` to `4.0.0`.
- Updates `gulp-changed` from `1.3.2` to `3.1.2`.
- Updates `gulp-clean-css` from `3.0.4` to `3.4.2`.
- Updates `gulp-imagemin` from `3.1.1` to `3.3.0`.
- Updates `gulp-sourcemaps` from `2.4.1` to `2.6.0`.
- Updates `gulp-uglify` from `2.0.1` to `3.0.0`.

## Testing

1. `./frontend.sh` should complete and the page should load without errors.


sebworks:
@anselmbradford, does ie9 support source maps?

anselmbradford:
> @anselmbradford, does ie9 support source maps?

Heh, no it doesn't, however, gulp-bless can be used with gulp-sourcemaps, and were it to be it would be spitting out incorrect maps according to [the changelog](https://github.com/BlessCSS/gulp-bless/releases/tag/4.0.0), so probably best to update it. Though you raise a good point, I don't quite see why gulp-bless supports sourcemaps at all when they weren't added to IE before IE11. Possibly just to support what the underlying bless library supports.

sebworks:
👍 , at some point we need to write some smoke tests for our gulp processes.

PR #3035, opened 2017-06-15
Ask cfpb go live

higs4281:
This turns off knowledgebase URLs and turns on the redirects to send old
requests to the new cfgov-refresh ask_cfpb app.

Go-live also requires that the `WAGTAIL_ASK_CFPB` flag be set to True in
the admin after code is deployed, to change megamenu links.

We also pick add redirects to capture requests for legacy facet-built 
audience and tag pages.

The facet redirects give priority to category facets, then audience facets,
then tag facets. This should capture the great majority of old search links.

## Testing
These legacy URLs should get redirected:  
#### Category search
- http://localhost:8000/askcfpb/search/?selected_facets=category_exact%3Abank-accounts-and-services 
#### Audience search
- http://localhost:8000/askcfpb/search/?selected_facets=audience_exact%3AOlder+Americans
#### Tag search
- http://localhost:8000/es/obtener-respuestas/buscar?selected_facets=tag_exact:cobrador%20de%20deudas

### DIFF
--- a/cfgov/ask_cfpb/models/django.py
+++ b/cfgov/ask_cfpb/models/django.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import, unicode_literals
 
-from collections import OrderedDict
+from collections import OrderedDict, Counter
 import HTMLParser
 import json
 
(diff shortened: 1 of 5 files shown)


Scotchester:
The tests pass, which is about all I have time to check this morning. On a _quick_ skim, the logic for the redirect changes looks good :+1:

higs4281:
@willbarton Great idea on the flag. Thanks.

PR #3037, opened 2017-06-15
Migrate cf-core to cfv4

anselmbradford:
Trying this workflow toward migrating to CFv4:
 1. Add CF module classes to cfgov-refresh enhancements, module by module.
 2. Check site against live site.
 3. Once all modules are updated, update `package.json` to CFv4 and remove enhancements. 

## Additions

- cf-core styles moved temporarily to enhancements.

## Changes

- Updates `.webfont-*` to `.u-webfont-*` as needed for CFv4.

## Testing

1. `./frontend.sh` and site should look the same as prod.

## Notes

- Not sure if we want to merge this already or wait till all modules are in. I'll be opening PRs for each module that build on this one.

### DIFF
--- a/cfgov/unprocessed/css/enhancements/base.less
+++ b/cfgov/unprocessed/css/enhancements/base.less
@@ -1,28 +1,592 @@
-/* ==========================================================================
-   cfgov-refresh
-   capital framework enhancements
-   ========================================================================== */
+// BEGIN enhancement for normalize-legacy-addon
 
-/* topdoc
-  name: Body Font
-  family: cf-core
-  notes:
-    - "Set the body to Avenir Regular."
-    - "Hides overflow on mobile sizes."
-  patterns:
-    - name: Body
-      codenotes:
-        - |
-          body
-  tags:
-    - cf-core
-*/
+/*
+ * Corrects list images handled incorrectly in IE 7.
+ */
 
-body {
-    .webfont-regular(); // Moved to CF, delete freely when migrating to CF4
+nav ul,
+nav ol,
+nav ul ul,
+nav ol ol {
+    list-style: none;
+    list-style-image: none;
 }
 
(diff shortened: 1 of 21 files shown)


cfgov/unprocessed/css/pages/ask.less
```diff
@@ -270,11 +270,11 @@
     }
 
     .question_summary {
-        padding: unit( (@grid_gutter-width / 2 ) / @base-font-size-px, em ) 
-                 0 
-                 unit( @grid_gutter-width / @base-font-size-px, em ) 
+        padding: unit( (@grid_gutter-width / 2 ) / @base-font-size-px, em )
+                 0
+                 unit( @grid_gutter-width / @base-font-size-px, em )
                  0;
```
jimmynotjim: Not a comment that needs to be addressed here, but the second `0` isn't necessary.

cfgov/unprocessed/css/enhancements/base.less
```diff
@@ -1,28 +1,592 @@
+// BEGIN enhancement for normalize-legacy-addon
+
+/*
+ * Corrects list images handled incorrectly in IE 7.
+ */
+
+nav ul,
+nav ol,
+nav ul ul,
```
jimmynotjim: You only need this one right? The other three are just excess that should probably be removed to avoid duplication.

jimmynotjim:
Also discussed in chat w/ @anselmbradford that merging will result in some duplication and bloat in the final css files. It may be wise to build on a v4 branch until the npm packages are updated, but it'll also be slower going. Leaving it up to you all to decide the best course.

anselmbradford:
Changed base to `cfv4_migration`, which is branch off master. Once all modules are in, I'll PR that off master, so this PR can go in sooner.

PR #3041, opened 2017-06-15
Remove unreliable meta image tests

contolini:
For reasons we've struggled to discover, the meta image tests fail on
production but work in every other enviroment.

See https://[chat]/cfpb/pl/nscfq5kpmbfg5mxgptcesdgcac

@chosak I'll let you decide if this PR should be merged.


chosak:
For future reference, I believe the reason for the intermittent failures is addressed by #3042.

PR #3036, opened 2017-06-15
Keep uploaded meta image URLs relative

contolini:
In production the `image()` function in templates returns an absolute URL to the image's S3 location. In other environments it returns a relative URL. So prepending `{{ request.scheme }}://{{ request.get_host() }}` doesn't play nicely with production.

## Checklist

* [ ] PR has an informative and human-readable title
* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [development playbook](https://github.com/cfpb/development)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated
* [ ] Reviewers requested with the [Assignee tool](https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/) :arrow_right:

PR #3042, opened 2017-06-15
Don't let SheerlikeContext replace the context request.

chosak:
This fixes some unit tests from #3036 that were breaking intermittently in a puzzling way. It turns out this is caused by how our [`SheerlikeContext`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/sheerlike/__init__.py#L44) class (which is the default context class for Jinja templates) tries to handle and replace the context `request` object.

`SheerlikeContext` assumes that the current request is put on the local thread using [this](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/sheerlike/middleware.py#L26) middleware call, but in unit tests this doesn't always happen. Then when templates get rendered, `SheerlikeContext` tries to grab the local thread's request to stick it back in the context, but (since not every test updates this) it might be hanging around from an old test.

This change modifies how `SheerlikeContext` works so that it won't replace an existing `request` in the context if one is already there.

It also makes a minor fix to the tests in #3036 so that they pass properly, now that it's more clear what was going on. 

## Changes

- Make `SheerlikeContext` more conservative about updating the context's `request` object.

## Testing

1. Observe that all unit tests pass.
2. Locally test a page which is served by Sheerlike, for example [http://localhost:8000/owning-a-home/closing-disclosure/](http://localhost:8000/owning-a-home/closing-disclosure/).

## Checklist

### DIFF
--- a/cfgov/sheerlike/__init__.py
+++ b/cfgov/sheerlike/__init__.py
@@ -51,10 +51,14 @@ class SheerlikeContext(Context):
             parent,
             name,
             blocks)
-        try:
-            self.vars['request'] = get_request()
-        except:
-            pass
+
+        # Don't overwrite an existing request already coming into the context,
+        # for example one provided during Wagtail rendering.
+        if 'request' not in self.vars and 'request' not in self.keys():
+            try:
+                self.vars['request'] = get_request()
+            except:
+                pass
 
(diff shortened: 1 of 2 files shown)


contolini:
:1st_place_medal:

PR #3047, opened 2017-06-16
Rename Next steps to "Related resources"

higs4281:
This aligns with how these items will be referred to in templates.

## Testing
This is what users should see:

admin_related_resources


higs4281:
This is just a label change; merging.

PR #3048, opened 2017-06-16
Migrate cf-buttons to cfv4

anselmbradford:

## Additions

- cf-buttons styles moved temporarily to enhancements.

## Changes

- Updates outdated `btn` classes to `a-btn`.

## Testing

1. `./frontend.sh` and site should look the same as prod.

## Notes

- Pagination disabled button background color is darker than the pagination background. This will have to be revisited when updating the pagination module.

### DIFF
--- a/cfgov/unprocessed/css/molecules/global-banner.less
+++ b/cfgov/unprocessed/css/molecules/global-banner.less
@@ -38,7 +38,9 @@
     }
 
     .o-expandable_target {
-        .btn__secondary();
+        // Taken from .a-btn__secondary.
+        background-color: @btn__secondary-bg;
+        color: @btn__secondary-text;
         width: 100%;
         // Restore standard btn padding, which is removed by .expandable_target.
         padding:
--- a/cfgov/unprocessed/css/skip-nav.less
+++ b/cfgov/unprocessed/css/skip-nav.less
@@ -20,8 +20,13 @@
 }
 
 #skip-nav:focus {
-    .btn;
-    .btn__super;
+    .a-btn;
+    // Taken from .a-btn__super.
+    padding:
+      unit( 11px / 18px, em )
+      unit( 29px / 18px, em );
+    font-size: unit( 18px / @base-font-size-px, em );
+
     position: absolute;
     top: 0;
     outline: 0;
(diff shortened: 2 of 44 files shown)


anselmbradford:
Merging this as it's just a snapshot of a larger PR that'll come through `cfv4_migration` to `master` later on.

cfgov/unprocessed/css/skip-nav.less
```diff
@@ -20,8 +20,13 @@
 }
 
 #skip-nav:focus {
```
jimmynotjim: Would it be better to add the `a-btn a-btn__super` classes to `#skip-nav` than to have them duplicated in the output here?

cfgov/unprocessed/css/molecules/global-banner.less
```diff
@@ -38,7 +38,9 @@
     }
 
     .o-expandable_target {
-        .btn__secondary();
+        // Taken from .a-btn__secondary.
```
jimmynotjim: Same question about duplication here. What's stopping us from including the `secondary` class in the markup?

cfgov/unprocessed/css/skip-nav.less
anselmbradford: The problem here is that it's only a super button on focus, not all the time.

jimmynotjim: It's only a super button on focus because it's hidden the rest of the time. But what's if we made it a super button all the time that's hidden with positioning till focused similar to the slide out menu on small screens?

cfgov/unprocessed/css/molecules/global-banner.less
jimmynotjim: This one doesn't have any limitation, it should definitely be moved to using the default class instead of duplicating the styles.

jimmynotjim:
> Merging this as it's just a snapshot of a larger PR that'll come through cfv4_migration to master later on.

I would avoid doing this. Reviewing that PR later is going to be a huge task as it is, it'll be easier to catch small things in these smaller PRs.

anselmbradford:
> I would avoid doing this. Reviewing that PR later is going to be a huge task as it is, it'll be easier to catch small things in these smaller PRs.

The problem was I couldn't reliably continue work on the main branch for the migration while the PR was open, but feel free to comment on anything in the snapshot PRs. https://github.com/cfpb/cfgov-refresh/pull/3062 https://github.com/cfpb/cfgov-refresh/pull/3052 https://github.com/cfpb/cfgov-refresh/pull/3049

PR #3051, opened 2017-06-16
Instruct Twitter to render large summary cards

contolini:
Twitter will only render large preview cards if you manually add a `summary_large_image` meta tag. This tag is added to a page if its meta_image is both high resolution AND its height to width ratio is approximately 1/2.

See https://dev.twitter.com/cards/types/summary-large-image

## Additions

- `summary_large_image` property to `CFGovImage`

## Testing

1. `tox -e fast v1.tests.models.test_images.CFGOVImageTest`

## Checklist


cfgov/v1/models/images.py
```diff
@@ -87,6 +87,14 @@ def default_alt_text(self):
         # Override Wagtail default of setting alt text to the image title.
         return self.alt
 
+    # If the image is both large and its height-to-width ratio is approximately
+    # 1/2 we instruct the template to render large Twitter cards
+    # See https://dev.twitter.com/cards/types/summary-large-image
+    @property
+    def summary_large_image(self):
```
richaagarwal: Nitpicky thing but maybe naming this something like `should_display_summary_large_image` is more descriptive, since it returns a boolean, not the image itself.

willbarton:
> Instruct Twitter

This imperative brings me more joy than it probably should.

cfgov/v1/models/images.py
contolini: @richaagarwal What about just `is_large`? So that it'd be `page.meta_image.is_large`?

richaagarwal: I guess the only reason I don't like that as much is that, in the definition it's clear the image needs to be both large and have a 2:1 aspect ratio.  If someone were quickly looking at the code, they might be like, "ok cool we display the large image for large images" and then get confused why that's not working.  I don't want to be a blocker on this but I think more verbose is better.

contolini: Works for me. 👍

PR #3054, opened 2017-06-19
Post-launch cleanup for Ask CFPB

higs4281:
This finalizes the retirement of knowledgebase and cleans up code needed
only for the migration.

Changes included:
- removal of knowledgebase from settings and requirements
- cleanup of urls.py to incorporate new URLs in the main pattern
- removal of tox DEPLOY_ENVIRONMENT variable
- retirement of the `WAGTAIL_ASK_CFPB` feature flag
- removal of the `WAGTAIL_ASK_CFPB` conditional code in
    - `jinja2/v1/_includes/organisms/_vars-mega-menu.html`
    -  `jinja2/v1/index.html`
- bump the smoketest timeout value to accommodate a slow
deployment completion. Normally the timeout is never reached, but just in case
the deployment needs a few more seconds to finish, this adds a cushion
so that the whole build doesn't fail.

These no-op cleanups should not have any effect on site.

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/_vars-mega-menu.html
+++ b/cfgov/jinja2/v1/_includes/organisms/_vars-mega-menu.html
@@ -69,7 +69,7 @@
 {% set nav_items = [(
     ('Consumer Tools', '#', [(
         ('Submit a Complaint', '/complaint/', []),
-        ('Ask CFPB', '/ask-cfpb/' if flag_enabled('WAGTAIL_ASK_CFPB', request) else '/askcfpb/', []),
+        ('Ask CFPB', '/ask-cfpb/', []),
         ('Tell Your Story', '/your-story/', []),
         ('Information for Students', '/students/', []),
         ('Information for Servicemembers & Veterans', '/servicemembers/', [])
--- a/cfgov/scripts/http_smoke_test.py
+++ b/cfgov/scripts/http_smoke_test.py
@@ -25,11 +25,16 @@ parser.add_argument(
 parser.add_argument(
     "-v", "--verbose",
     action="store_true",
-    help="set logging level to info to see all message output."
+    help="Set logging level to info to see all message output."
+)
+parser.add_argument(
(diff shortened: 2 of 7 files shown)


cfgov/jinja2/v1/_includes/organisms/_vars-mega-menu.html
```diff
@@ -69,7 +69,7 @@
 {% set nav_items = [(
     ('Consumer Tools', '#', [(
         ('Submit a Complaint', '/complaint/', []),
-        ('Ask CFPB', '/ask-cfpb/' if flag_enabled('WAGTAIL_ASK_CFPB', request) else '/askcfpb/', []),
```
willbarton: Missed this when it was added — I didn't know we could use ternaries in Jinja2 templates!

cfgov/scripts/http_smoke_test.py
```diff
@@ -29,7 +29,7 @@
 )
 
 FULL = False
-TIMEOUT = 10
+TIMEOUT = 30
```
willbarton: It might be worth making this a command-line argument so that the timeout could be passed in via Jenkins. However, not necessary to change in this PR.

requirements/optional-private.txt
```diff
@@ -5,6 +5,5 @@
 git+https://private.repo/CFPB/agreement_database.git@2.2.7#egg=agreement-database
 git+https://private.repo/CFPB/cfgov-selfregistration.git@1.3.1#egg=selfregistration
 git+https://private.repo/CFPB/django-college-cost-comparison.git@v1.2.9#egg=comparisontool
-git+https://private.repo/CFPB/knowledgebase.git@2.4.0#egg=knowledgebase
```
willbarton: dancing bear

One more private repo down!

cfgov/scripts/http_smoke_test.py
higs4281: @willbarton Added.

PR #3064, opened 2017-06-20
Removes unused careers CSS

anselmbradford:
`careers.less` was not being included in `main.less` so does not appear to be used at all.

## Removals

- `careers.less`
- `t-careers_social` CSS class, which was unused.

## Testing

1. An individual job posting should not experience a change in layout after this change, as  `t-careers_social` was adding a `float:right`, but does not appear to be used.

## Notes

- I'm not sure how to load mock jobs locally to test, and the live site doesn't have any job postings currently so I can't test remove  `t-careers_social` there, so we may want to wait till a job is posted unless you'll can test.


jimmynotjim:
Looks good, but I'd wait for @sebworks to answer [you on the class](https://github.com/cfpb/cfgov-refresh/commit/8aef3619ea41ec1e4b25b6187af5d1a9aefc4b61#diff-c574ab9aa2258b4c8c4dff6879ddb04aR39). I don't remember what it was from either.

sebworks:
@jimmynotjim, I'm writing some tests for this PR.

sebworks:
@anselmbradford, @jimmynotjim please check out the margin top I added to the social sharing. 

screen shot 2017-06-21 at 2 37 34 pm

anselmbradford:
Works for me! 👍  Let @jimmynotjim sign off before merging. I was trying to find a way to not need the margin util, but didn't find a way.

jimmynotjim:
Sounds good. 💯

PR #3070, opened 2017-06-21
Migrate slug header to CFv4

anselmbradford:
On the road to CFv4 migration…

## Changes

- Updates `header-slug` to use `m-slug-header` syntax from CFv4.

## Testing

1. `./frontend.sh`
2. Any page that has a header slug should be unchanged (though see note below).

## Notes

- A template for the header slug could be created in the future, as all these instances use the same format.

## Todos

- The old heading didn't have a bottom margin under the heading text, the new one does. Which is correct?

Before:

screen shot 2017-06-21 at 4 44 14 pm

After:

screen shot 2017-06-21 at 4 44 07 pm

### DIFF
--- a/cfgov/jinja2/v1/ask-cfpb/answer-page.html
+++ b/cfgov/jinja2/v1/ask-cfpb/answer-page.html
@@ -30,7 +30,7 @@
 
 
 {% block content_main_modifiers -%}
-    {{ super() }} content__flush-bottom 
+    {{ super() }} content__flush-bottom
 {%- endblock %}
 
 {% block content_main %}
@@ -54,7 +54,7 @@
                 {{ page.snippet | safe }}
             </h3>
         {% endif %}
-        <div data-qa-hook="expandable" 
+        <div data-qa-hook="expandable"
              data-read-more="true"
              class="o-expandable
                     o-expandable__read-more">
@@ -63,7 +63,7 @@
                     <div class="answer-text">
                         {{ page.answer | richtext | safe }}
                     </div>
-                    {% if audiences %} 
+                    {% if audiences %}
                     <div class="u-mt15">
                         {{ tags.render( {'links': audiences}, '', is_wagtail=True) }}
                     </div>
@@ -79,7 +79,7 @@
                     </div>
                     {% endif %}
                 </div>
-            </div>            
+            </div>
             <a class="o-expandable_target jump-link o-expandable_link" title="Expand content">
                 <span class="jump-link_text">
                     <span class="o-expandable_cue o-expandable_cue-open">
@@ -93,7 +93,7 @@
                 </span>
             </a>
         </div>
-        
+
 
     </div>
     {{ ask_search_bar.render( 'left' ) }}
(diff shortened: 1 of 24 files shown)


anselmbradford:
> The old heading didn't have a bottom margin under the heading text, the new one does. Which is correct?

Added via https://github.com/cfpb/capital-framework/blob/canary/src/cf-core/src/cf-base.less#L108

cfgov/jinja2/v1/ask-cfpb/answer-page.html
```diff
@@ -119,20 +119,20 @@ <h3 >
...
-                    <span class="header-slug_inner">
+                <header class="m-slug-header">
+                    <h2 class="a-heading">
                         Related {{category.name}} subjects
-                    </span>
-                </h2>
-                <ul class="list list__unstyled list__links">
+                    </h2>
+                </header>
+                <ul class="m-list m-list__unstyled m-list__links">
```
jimmynotjim: Did you mean for lists to sneak in here?

jimmynotjim:
@anselmbradford Doesn't look like slug headers ever made it into the Design Manual. Any designers know what's correct? I'm assuming the margin is a mistake and should be reset in the nested `a-heading` rules. cc @nataliafitzgerald

jimmynotjim:
Wait, it is sorta documented here https://cfpb.github.io/design-manual/page-components/email-sign-up.html with the screenshot of the email signup.

Scotchester:
I would figure the bottom margin should be the same as for H5s, since it's essentially an H5 derivative.

jimmynotjim:
Tested everywhere I can find the slug headers locally, everything looks good other than the extra margin.

nataliafitzgerald:
@anselmbradford @jimmynotjim 
The "Before" image (above) looks closer to correct but what is the space on that? I suspect that using the standard H5 spacing will look correct, as @Scotchester mentioned. If that is different than the "Before" image show above can you post a screenshot here? The "After" image appears to have too much space. 

So, let's see what the standard H5 spacing would look like.

jimmynotjim:
Figured it out, the spacing is being doubled because the `m-slug-header` is adding it's own bottom margin. I think we can remove that and let the `h5` styles dictate the margin.

cfgov/jinja2/v1/ask-cfpb/answer-page.html
anselmbradford: Oops, nope. Fixed in https://github.com/cfpb/cfgov-refresh/pull/3070/commits/e4fe8e8167d88f551b7d07ca5c008b6e426f7b90

anselmbradford:
> Figured it out, the spacing is being doubled because the m-slug-header is adding it's own bottom margin. I think we can remove that and let the h5 styles dictate the margin.

So just so I'm clear in this PR I'd change:

```
.m-slug-header {
    margin-bottom: unit( 17px / @base-font-size-px, em );
    border-top: 1px solid @slug-header_border__thin;
```

…to…

```
.m-slug-header {
    border-top: 1px solid @slug-header_border__thin;
```

And then PR that change to CF?

jimmynotjim:
@anselmbradford Yes, that seems like the right approach to me.

nataliafitzgerald:
@anselmbradford 
Can you post a screenshot?

anselmbradford:
@nataliafitzgerald 

screen shot 2017-06-22 at 1 05 03 pm

nataliafitzgerald:
@anselmbradford - Slug header spacing looks good!

anselmbradford:
Related to https://github.com/cfpb/capital-framework/pull/516

PR #3071, opened 2017-06-22
Ask CFPB: Add ability to remove tip styling using tip button

virginiacc:
Currently the only ways to remove tip styling in an Ask answer field are to edit the HTML directly or to delete the tip and re-enter its contents. This PR adds the ability to remove tip styling from a selection using the answer field's tip button.

## Changes

- Update `ask_cfpb_tips.js` to remove tip styling instead of adding it again if selection is part of a tip.
- Update the way tip selection is handled to try to correct for partial selections. This should help prevent things like tips inside `<p>` elements

## Testing

1. Check out branch
2. Choose an Ask answer in admin. Highlight a selection and click tip button to add tip styling. You should now be able to remove the styling by clicking the tip button again, either immediately or after selecting some or all of the tip contents.
3. If you highlight a partial selection -- leave off the last few characters of a paragraph -- the selection should (generally) expand to contain the stragglers

## Checklist


virginiacc:
@higs4281 Thanks for catching that issue! The tip removal should persist now without having to open the HTML view. I also made some changes to how the tip styling is applied to try to correct partial selections. This should help prevent things like tip `<aside>`s being nested in `<p>`s instead of the other way around.

cfgov/v1/static/js/admin/ask_cfpb_tips.js
```diff
@@ -11,6 +11,8 @@
       populateToolbar: function (toolbar) {
         if (window.location.href.indexOf('ask_cfpb') > -1) {
           var buttonset,
+          baseClass = "answer-module",
+          textInputClass = 'richtext',
           _this = this;
           buttonset = jQuery("<span class=\"" + this.widgetName + "\"></span>");
```
sebworks: We don't use jquery anymore in cfgov-refresh, unless it's a legacy app.

cfgov/v1/static/js/admin/ask_cfpb_tips.js
```diff
@@ -26,22 +28,94 @@
...
-            elem.appendChild(contents);
-            range.deleteContents();
-            range.insertNode(elem);
-
-            $(range.commonAncestorContainer).find('*').each(function() {
-              if ($(this).is(':empty') || $(this).text() === '') {
-                $(this).remove();
+            var node = sel.baseNode;
+            
+            var tip = $(node).closest( '.' + baseClass )[0];
```
sebworks: same thing here.

cfgov/v1/static/js/admin/ask_cfpb_tips.js
```diff
@@ -26,22 +28,94 @@
...
+              _this.options.editable.setModified();
+              return;
+            }
+
+            // Otherwise, get the selected range. If it starts or ends
+            // with a text node or inline element (a, strong, i), 
+            // expand the selection either to that node's 
+            // start or end bound, or the start or end bound of its parent
+            // if it's not a direct child of the text input div.
+            var range = rangy.getSelection().getRangeAt(0);
```
sebworks: This code gives me pause. We are we trying to do here?

sebworks: Would it be possible to use the https://developer.mozilla.org/en-US/docs/Web/API/TreeWalker?

cfgov/v1/static/js/admin/ask_cfpb_tips.js
virginiacc: This is for the admin, not the public-facing site -- it's a jQuery plugin that adds a button to the Answer rich text field toolbar so content editors can select content to get tip styling.

cfgov/v1/static/js/admin/ask_cfpb_tips.js
virginiacc: I could write this as plain javascript, but since it's part of a jquery plugin it seems acceptable to use jquery convenience methods.

cfgov/v1/static/js/admin/ask_cfpb_tips.js
virginiacc: This code gets a selection of text made by the user, expands it as necessary to encompass full elements, and then wraps it in a containing element so tip styling can be applied.

cfgov/v1/static/js/admin/ask_cfpb_tips.js
sebworks: ok, make sense. I didn't realize it was for the admin. We place ours in https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/templates/wagtailadmin/js/table-block.js

cfgov/v1/static/js/admin/ask_cfpb_tips.js
sebworks: naw, makes sense.

cfgov/v1/static/js/admin/ask_cfpb_tips.js
virginiacc: Ah, okay! Maybe I should move this there...

cfgov/v1/static/js/admin/ask_cfpb_tips.js
```diff
@@ -26,22 +28,94 @@
...
+            // Check if range can be surrounded with a wrapper element.
+            // If not, try to find the least common ancestor of the 
+            // start and end containers and expand the start/end of the
+            // range to encompas it so the range can be surrounded.
+            function getDepth( ancestor, node ) {
+              return node === ancestor 
+                          ? 0 : 
+                          $( node ).parentsUntil( common ).length + 1;
+            }
+
```
sebworks: why not just assign `range.canSurroundContents() ` to a variable and make line 112 and `elseif`?

PR #3077, opened 2017-06-23
Add logging via syslog to production settings

rosskarchner:
This adds a (parallel) logging configuration that utilizes the logging subsystem provided by the OS, instead of having Python and Django manage those files. This should prove more robust, since logging messages through syslog does not require any special privileges. 

If this works well, I think we should remove the 'disk' handler-- but there's no reason they can't live side-by-side until we make that call.

## Additions

- a 'syslog' log handler
- a 'tagged' formatter, that adds some metadata for easier filtering

## Testing

as things are right now, you should be set DJANGO_SETTINGS_MODULE=cfgov.settings.production, and do something that would write to the log. Perhaps in the django shell:

```
>>> import logging      
>>> logger = logging.getLogger(__name__)         
>>> logger.info("hey!") 
```

Unfortunately, on our workstations, these messages will get written to /var/log/system.log, which we do not have permission to read. You should be able to observe the timestamp and size of the file changing, at least 😢 .

## In production.

Without any further config, on linux these messages will be logged to /var/log/messages. There is a seperate PR that filters all messages tagged with "django" to /var/log/django.log

[GHE]/cfpb-infra/ansible-consumer-finance/pull/118

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/production.py
+++ b/cfgov/cfgov/settings/production.py
@@ -6,14 +6,24 @@ from .base import *
 # This avoids permissions problems when logged in users (or CI jobs)
 # can't write to the log file.
 if sys.argv and sys.argv[0] == 'mod_wsgi':
-    default_loggers = ['disk']
+    default_loggers = ['disk', 'syslog']
 else:
-    default_loggers = ['console']
+    default_loggers = ['console', 'syslog']
 
 # Sends an email to developers in the ADMIN_EMAILS list if Debug=False for errors
+#
+# in the formatter, "django: " is an rsyslog tag. This is equivalent to:
+#     logger -t django "my log message"
+# on the server, the tag will be used to route the message to the desired
+# logfile
 LOGGING = {
     'version': 1,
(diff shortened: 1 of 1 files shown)


cfgov/cfgov/settings/production.py
```diff
@@ -1,19 +1,34 @@
...
 if sys.argv and sys.argv[0] == 'mod_wsgi':
-    default_loggers = ['disk']
+    default_loggers = ['disk', 'syslog']
 else:
-    default_loggers = ['console']
+    default_loggers = ['console', 'syslog']
+
+
+# This allows the syslog stuff to work on OS X
+syslog_device = next(l for l in ['/dev/log', '/var/run/syslog'] if exists(l))
```
willbarton: Noting that this makes me uncomfortable for reasons I'm struggling to elucidate.

rosskarchner: too inscrutable? Or you don't like the whole idea of checking which file exists?

PR #3079, opened 2017-06-26
remove disk logging, in favor of syslog

rosskarchner:
The 'disk' logger depends on the (apparently) fragile assumption that permissions won't change on the log file. In particular, it appears whenever the httpd packages is updated in RHEL6, ownership of those files is changed back to 'root'.

On Friday, we added the 'syslog' configuration that appears to be working pretty well, and depends on services provided by the operating system, instead of maintaining write-access to a particular file.

## Removals

- the 'disk' logger and associated logic

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/production.py
+++ b/cfgov/cfgov/settings/production.py
@@ -4,8 +4,6 @@ from .base import *
 from os.path import exists
 
 # log to disk when running in mod_wsgi, otherwise to console
-# This avoids permissions problems when logged in users (or CI jobs)
-# can't write to the log file.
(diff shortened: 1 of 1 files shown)


rosskarchner:
@chosak I'm not sure-- That's how it was configured in the original PR that added logging, so I don't want to make assumptions about the intent.

willbarton:
Original intent was a narrow scope since it was a late-night attempt to get it working. I don't think there's a problem with `''`.

chosak:
@rosskarchner okay, that seems reasonable to limit this scope. I'd anticipate a future PR revisiting that at some point when @higs4281 wonders why Ask logs don't show up 😄

PR #3080, opened 2017-06-26
Add "Data snapshot" Wagtail block 

richaagarwal:
Create a Wagtail block for storing "Data snapshot" values on browse pages. Next steps will include automating the population of certain fields from a Jenkins job.

Broader thing we're trying to do:
Automate data snapshots you see on pages like this: https://www.consumerfinance.gov/data-research/consumer-credit-trends/credit-cards/. Right now they exist in Wagtail as a half width link blob, and the numbers that change every month have to be input by an editor.  Once this work is complete, that will happen automatically.


cfgov/jinja2/v1/_includes/organisms/data_snapshot.html
```diff
@@ -0,0 +1,23 @@
+<html>
+    <h2>Data Snapshot for {{ value.last_updated_date }}</h2>
+    <!-- Here needs wrapping HTML if handling both columns of 50-50 -->
```
richaagarwal: Let's clean up some of the comments in this template @hillaryj

PR #3081, opened 2017-06-26
Add the sortable table options to the Wagtail admin

jimmynotjim:
This is step one of adding sortable tables to Wagtail. It includes the admin options, the front-end markup, and a feature flag to protect the front-end. The front-end won't work until the project has been updated to use cf-tables v4.x. When that work is complete, the feature flag should be removed.

## Additions

- Added sortable code to admin template
- Added sortable code to front-end template
- Added scripting to allow user interaction and data saving
- Added feature flag to protect the rendered template until cf-tables v4.x is implemented

## Changes

- Cleaned up some of the existing code to match our formatting standards.

## Testing

I'm not sure how to test this programmatically, maybe a back-end dev could lend a hand? To test locally, run the site, log in to the admin, and create a table, then:

1. Select the checkbox marked `Sortable table`.
2. Select either `default` or `number` from the sortable options that appear above the table.
3. Save the page, make sure the data has saved and the options appear correctly on reload.
4. Preview the page and verify there are no changes on the rendered template (because they are hidden behind the flag).

## Screenshots

screen shot 2017-06-27 at 9 32 15 am

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/table.html
+++ b/cfgov/jinja2/v1/_includes/organisms/table.html
@@ -43,10 +43,11 @@
 {% endmacro %}
 
 <table class="o-table
-              {{ 'o-table__row-links' if value.row_links else '' }}
-              {{ 'o-table__stack-on-small' if value.is_stacked else '' }}
-              {{ 'table__striped' if value.is_striped else '' }}
-              {{ 'u-w100pct' if value.is_full_width else '' }}">
+              {{- ' o-table__row-links' if value.row_links else '' -}}
+              {{- ' o-table__stack-on-small' if value.is_stacked else '' -}}
+              {{- ' table__striped' if value.is_striped else '' -}}
+              {{- ' u-w100pct' if value.is_full_width else '' -}}
+              {{- ' o-table__sortable' if value.is_sortable else '' -}}">
     {% set row_index = 0 %}
     {% set first_row = value.data[row_index] %}
     {% if value.first_row_is_table_header and value.data | list | length > 0 %}
@@ -57,11 +58,18 @@
         {% if first_row | reject('none') | list | length %}
         <tr>
             {% for cell in value.data[0] %}
-                {% if value.fixed_col_widths %}
-                    {% set width_class = 'class="' + value.column_widths[loop.index0] + '"' %}
+                {% if value.fixed_col_widths and value.column_widths[loop.index0] != '' %}
+                    {% set width_class = ' class="' + value.column_widths[loop.index0] + '"' %}
                 {% endif %}
             <th scope="col"{{ width_class | default }}>
-                {{ _render_cell(cell) }}
+                {% if flag_enabled('SORTABLE_TABLES', true) and value.is_sortable and value.sortable_types[loop.index0] != '' %}
+                    {% set data_sort = ' data-sort_type="' + value.sortable_types[loop.index0] + '"' %}
+                    <button class="sortable"{{ data_sort }}>
+                        {{ _render_cell(cell) }}
+                    </button>
+                {% else %}
+                    {{ _render_cell(cell) }}
+                {% endif %}
             </th>
             {% endfor %}
         </tr>
--- a/cfgov/templates/wagtailadmin/js/table-block.js
+++ b/cfgov/templates/wagtailadmin/js/table-block.js
@@ -33,7 +33,8 @@
                 HEIGHT: 1
             },
 
-            $colWidthSelect: {}
+            $colWidthSelect: {},
+            $sortableSelect: {}
         };
 
         var HandsonTable = {
@@ -121,6 +122,10 @@
                 .find( 'td' ).eq( index - 1 )
                 .after( '<td>' + utilities.$colWidthSelect + '</td>' );
 
+              HandsonTableWagtailBridge.ui.$sortableInput
+                .find( 'td' ).eq( index - 1 )
+                .after( '<td>' + utilities.$sortableSelect + '</td>' );
+
               this.$element.trigger( 'table:change', [this.instance.getData()] );
 
               return this;
@@ -144,6 +149,7 @@
                 var hiddenFieldData;
 
                 utilities.$colWidthSelect = $( this.ui.$inputContainer + ' .column-width-input' ).prop( 'outerHTML' );
+                utilities.$sortableSelect = $( this.ui.$inputContainer + ' .sortable-type-input' ).prop( 'outerHTML' );
 
                 this.options = utilities.assign( {} , options );
                 this.element = document.querySelector( this.ui.$inputContainer );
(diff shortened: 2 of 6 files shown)


cfgov/templates/wagtailadmin/js/table-block.js
```diff
@@ -344,12 +396,14 @@
...
                     column_widths:             this.getColumnWidths(),
+                    sortable_types:            this.getSortableTypes(),
                     first_row_is_table_header: ui.$hasRowHeaderCheckbox.prop( 'checked' ),
                     first_col_is_header:       ui.$hasColHeaderCheckbox.prop( 'checked' ),
                     is_full_width:             ui.$isFullWidthCheckbox.prop( 'checked' ),
                     is_striped:                ui.$isTableStripedCheckbox.prop( 'checked' ),
                     is_stacked:                ui.$isStackedOnMobileCheckbox.prop( 'checked' ),
-                    fixed_col_widths:          ui.$fixedWidthColsCheckbox.prop( 'checked' )
+                    fixed_col_widths:          ui.$fixedWidthColsCheckbox.prop( 'checked' ),
+                    is_sortable:               ui.$isSortableCheckbox.prop( 'checked' )
```
Scotchester: I want to say "Add comma at the end of this line," but if this is JSON, it's not legal, right?

cfgov/templates/wagtailadmin/js/table-block.js
```diff
@@ -344,12 +396,14 @@
...
                 } ) );
```
Scotchester: Should we reorganize this so that `fixed_col_widths` and `column_widths` are together and `is_sortable` and `sortable_types` are together, or is it better to keep the `this.<whatever>` stuff at the top?

cfgov/templates/wagtailadmin/table_input.html
```diff
@@ -69,9 +69,47 @@
...
+<div class="field boolean_field widget-checkbox_input">
+    <label for="{{ attrs.id }}-handsontable-sortable">Sortable table</label>
+    <div class="field-content">
+        <div class="input">
+            <input type="checkbox" id="{{ attrs.id }}-handsontable-sortable" name="handsontable-sortable"/>
+            <span class="help">( Enable sortable table functionality )</span>
+        </div>
+    </div>
+</div>
 <br/>
```
Scotchester: Since we're here, let's get rid of this unnecessary self-closing slash and the ones on all `input`s above.

cfgov/templates/wagtailadmin/table_input.html
```diff
@@ -69,9 +69,47 @@
...
 <div id="{{ attrs.id }}-width_warning" class="table-block-warning">Warning! Your total column width is greater than 100%! Please adjust your column widths to be less than 100% total before continuing.</div>
+
+<table id="{{ attrs.id }}-sortable-input">
```
Scotchester: How about putting this below the Fixed Width Column Options to match the order of the checkboxes?

cfgov/templates/wagtailadmin/table_input.html
```diff
@@ -69,9 +69,47 @@
...
+  <thead>Sortable Table Options</thead>
+  <tr>
+    <td>
+      <select class="sortable-type-input">
+        <option value="">None</option>
+        <option value="string">Default</option>
```
Scotchester: "Default" may be unclear. How about "Alphabetical"?

cfgov/jinja2/v1/_includes/organisms/table.html
```diff
@@ -57,11 +58,18 @@
         {% if first_row | reject('none') | list | length %}
         <tr>
             {% for cell in value.data[0] %}
-                {% if value.fixed_col_widths %}
-                    {% set width_class = 'class="' + value.column_widths[loop.index0] + '"' %}
+                {% if value.fixed_col_widths and value.column_widths[loop.index0] != '' %}
```
jimmynotjim: Making sure it's not adding an empty class.

cfgov/templates/wagtailadmin/js/table-block.js
```diff
@@ -33,7 +33,8 @@
                 HEIGHT: 1
             },
 
-            $colWidthSelect: {}
+            $colWidthSelect: {},
+            $sortableSelect: {}
```
jimmynotjim: I thought we didn't add trailing commas to `js` files, sort of as a carry over from `json` formatting.

cfgov/templates/wagtailadmin/js/table-block.js
jimmynotjim: Yeah, pretty sure we shouldn't do that here.

cfgov/templates/wagtailadmin/js/table-block.js
jimmynotjim: I don't think so, these `ui.` ones are all the settings checkboxes. If anything I'd move the functions to the bottom since they're looking at the tables that are below the settings checkboxes.

cfgov/templates/wagtailadmin/table_input.html
jimmynotjim: Yep. Had it there, forget why I switched it around.

cfgov/templates/wagtailadmin/table_input.html
jimmynotjim: Good call.

cfgov/templates/wagtailadmin/table_input.html
jimmynotjim: Fixed.

cfgov/templates/wagtailadmin/js/table-block.js
```diff
@@ -304,6 +341,21 @@
                        ( tableParent.find( '.input' ).height() * 2 );
             },
 
+            getSortableTypes: function getColumnWidths() {
```
jimmynotjim: Fixed.

cfgov/templates/wagtailadmin/js/table-block.js
```diff
@@ -219,7 +238,9 @@
                 $isTableStripedCheckbox:    id + '-handsontable-striped-rows',
                 $fixedWidthColsCheckbox:    id + '-handsontable-col-fixed',
                 $fixedWidthColInput:        id + '-fixed-width-column-input',
-                $widthWarning:             id + '-width_warning',
+                $isSortableCheckbox:        id + '-handsontable-sortable',
+                $sortableInput:             id + '-sortable-input',
+                $widthWarning:              id + '-width_warning',
```
jimmynotjim: Probably. Fixed.

jimmynotjim:
> One thing we may also need to consider: choosing which column to sort on for the initial load. 

I admit I completely forgot about this option. Any thoughts on what's the best solution? I don't think we want to use checkboxes or radios, but that might be the only option.

Scotchester:
> > One thing we may also need to consider: choosing which column to sort on for the initial load. 
> 
> I admit I completely forgot about this option. Any thoughts on what's the best solution? I don't think we want to use checkboxes or radios, but that might be the only option.

Perhaps it's not necessary, actually. Since data entry is all done right there in the module and not, say, pulled from a database or API, we could just rely on the content editor to input it in the order they want initially.

willbarton:
@sebworks is this a candidate for a feature file describing the interactions?

sebworks:
yeah, @willbarton, this is a prime candidate for acceptance testing.

sebworks:
ok, so this is fine for now. I do think we have pushed the limit in terms of the Wagtail UI for TableBlock, as well as the code that supports it.  That is my fault. 

It's do a refactor at this point, but it doesn't have to be now.

sebworks:
@jimmynotjim, great job on writing the feature. Thanks.

cfgov/templates/wagtailadmin/js/table-block.js
Scotchester: I always do if the language allows it. Does the linter complain if you do it?

PR #3084, opened 2017-06-27
Removes unused templates

anselmbradford:
## Removals

- Removes templates that could not be found as imports in other files across the cfgov-refresh codebase, and also removes associated CSS less files.

## Testing

1. `./frontend.sh` and there should be no change in the site.


jimmynotjim:
screen shot 2017-06-27 at 9 43 13 am

I like your style 💯

sebworks:
I am for getting rid of the filter stuff but the atoms might be useful in the future.

anselmbradford:
> I am for getting rid of the filter stuff but the atoms might be useful in the future.

Yeah, I wasn't sure about those as there are others (such as `input.html`) that is used by `conference-registration-form.html`, which isn't currently used but probably will next year for https://www.consumerfinance.gov/data-research/cfpb-research-conference/. Maybe these should be recreated in the examples in Capital Framework though instead since they are using outdated `form-l` classes.

sebworks:
@anselmbradford, I would just go ahead and merge.

PR #3078, opened 2017-06-23
Migrate pull quote to cfv4

anselmbradford:
## Changes

- Updates `m-quote` and `pull-quote` to Capital Framework v4 `m-pull-quote`.
- Adds "is large" option to Wagtail to include large pull quote style from CF.

## Testing

1. Create a test page in Wagtail.
2. Add a full-width text component.
3. Add a quote to the full-width text.
4. Preview changes to see pull quote.

## Screenshots

Before:
screen shot 2017-06-23 at 6 22 25 pm

After:
screen shot 2017-06-23 at 6 15 53 pm

Wagtail:
screen shot 2017-06-23 at 6 42 52 pm

## Todos

- This probably needs tests for the wagtail changes, but I need guidance on how to add them. Anyone?

PR #3085, opened 2017-06-27
Restore missing migration testing

willbarton:
See discussion in https://github.com/cfpb/cfgov-refresh/pull/3078.

Since we're not running migrations in Travis our missing migration unit test no longer works in Travis. This is an attempt to perform the same test by copying the migration test code into a script that can be run by the `runscript` management command in a tox environment.

The script can be removed and replaced by `makemigrations --check` in Django 1.10+.

The output in Travis should appear like this for failure:
```
Missing migrations found:
Migrations for 'v1':
  0070_cfgovpage_changed.py:
    - Add field changed to cfgovpage
```

## Testing

1. Run `tox -e missing-migrations` and see that everything suceeds.
2. Change a model in a trivial way for testing.
3. Run `tox -e missing-migrations` and see that it fails and prints the expected migration it would've created.
4. Revert your change to the model

## Checklist


tox.ini
```diff
@@ -66,3 +66,18 @@ deps={[testenv:optional]deps}
 envdir={[testenv:optional]envdir}
 recreate={[testenv:fast]recreate}
 setenv={[testenv:fast]setenv}
+
```
chosak: Is there a good reason to add this to `tox.ini` instead of just adding to the `run_travis.sh` script?

tox.ini
```diff
@@ -66,3 +66,18 @@ deps={[testenv:optional]deps}
...
+[testenv:missing-migrations]
+# Invoke with: tox -e missing-migrations
+# Check for changes to Django models that are missing migrations by 
+# running a script that calls makemigrations and examines the result.
+# The missing_migrations script can be removed in favor of 
+# `makemigraitons --check` when we move to Django 1.10+.
```
chosak: Typo: `makemigraitons`.

tox.ini
willbarton: Yeah, because with tox we get the environment with Django and all requirements installed already.

willbarton:
[Successful failure!](https://travis-ci.org/cfpb/cfgov-refresh/jobs/247577516)

tox.ini
chosak: I think that makes sense. The reason I hesitated is because if/when we want to move to using Tox in its more typical usage pattern (e.g. with Django 1.8/1.11 or Python 2/3) then these kind of one-off environments will be a little more tricky to include.

willbarton:
@chosak I debated, but I'm hesitant to delete the unit test since it runs as part of the normal test suite when running locally (I may be making assumptions about how others run the tests locally though) and in Jenkins. It gives one extra place to catch missing migrations. But this could be over-caution.

chosak:
@willbarton oh, right, that makes sense. Hm, in that case, is there any way to somehow unify the overlapping code so that we don't have the same logic defined twice?

willbarton:
@chosak I've refactored both to use the same code (in the missing_migrations script).

cfgov/core/scripts/missing_migrations.py
```diff
@@ -11,17 +11,11 @@
...
-    major, minor = VERSION[0:2]
-    if 1 == major and 10 <= minor:
-        print(
-            'makemigrations --exit deprecated in Django 1.10: '
-            'https://docs.djangoproject.com/en/1.10/ref/django-admin/'
-            '#cmdoption-makemigrations--exit'
-        )
-        sys.exit(1)
+# Django 1.10+ includes makemigrations --check, making this unnecessary
+is_django_110_plus = (1 == VERSION[0]) and (10 <= VERSION[1])
```
chosak: Nitpicky, in Django 2.0 won't this return `False`? Maybe `(1 <= VERSION[0])`?

willbarton: True, it would. Change is made. It's saddening (but realistic) to think we might not be fully onto 1.10 when we're testing against 2.0 😢

PR #3086, opened 2017-06-27
Removes micro-copy

anselmbradford:
## Removals

- Unused templates (redundant to some in https://github.com/cfpb/cfgov-refresh/pull/3084)

## Changes

- Updates references to `micro-copy` to `a-micro-copy` and adds CSS from CF.
- Updates `@micro-copy-text` to `@micro-copy`.

## Testing

Only place micro copy is currently used is on the non-live conference registration wagtail block from data and research.
1. Run `echo 'delete from django_migrations where app = "data_research"' | python cfgov/manage.py dbshell`
1. Run `python cfgov/manage.py migrate`
1. Create a browse page in Wagtail.
2. Add Conference registration form.
3. Preview the page.

## Screenshots

micro-copy
screen shot 2017-06-27 at 11 21 46 am

a-micro-copy
screen shot 2017-06-27 at 11 21 16 am

## Notes

- The text ends up larger because it is hardcoded to 14px in `micro-copy`, but uses ems in `a-micro-copy`.

### DIFF
--- a/cfgov/jinja2/v1/_includes/atoms/input.html
+++ b/cfgov/jinja2/v1/_includes/atoms/input.html
@@ -31,11 +31,11 @@
     <label class="form-label-header" for="{{ id }}">
         {{ value.label }}
         {%- if value.required == false -%}
-            <span class="a-micro-copy">&nbsp;(Optional)</span>
+            <small>&nbsp;(Optional)</small>
         {%- endif -%}
     </label>
     {%- if value.helperText -%}
-        <p class="a-micro-copy" id="{{ ht_id }}">{{ value.helperText | safe }}</p>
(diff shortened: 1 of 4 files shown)


sebworks:
@anselmbradford, do you think this is going to get used in other places? If not, then let's get rid of it and just target that specific block.

anselmbradford:
> do you think this is going to get used in other places? If not, then let's get rid of it and just target that specific block.

I'd be fine with wiping it out completely till it's sorted in CF. I can swap out `<small>` for the text that's using it in the form.

anselmbradford:
Swapped `micro-copy` with `<small>`, looks the same as the `a-micro-copy` screenshot above.

PR #3089, opened 2017-06-28
adding eregs-2.0 to cfgov-refresh

grapesmoker:
This PR adds eregs-2.0 to the cfgov-refresh ecosystem.


rosskarchner:
Argh-- sorry for forgetting, but this probably needs some addition to urls.py, right?

And *that* probably also means a feature flag. 

You can find some 'include_if_app_enabled ' (which checks OPTIONAL_APPS)  and flagged_url examples here: https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/urls.py

The wagtail_flags readme is pretty useful too: https://github.com/cfpb/wagtail-flags

rosskarchner:
@grapesmoker I'm gonna take a look at that stuff I just mentioned

rosskarchner:
I've added a feature flag, and flagged_url, and added django_mysql to requirements.txt

@grapesmoker is going to do some work on compatibility with Django 1.8

willbarton:
Making a note: The feature flag will prevent the URLs from working, but migrations will still be run everywhere the app is included.

I don't know if that's a problem, but if it is the app inclusion can be hidden behind `DEPLOY_ENVIRONMENT` like @higs4281 [used for Ask CFPB](https://github.com/cfpb/cfgov-refresh/blob/4f49dcd9f64c227eaef2ab5344e02517948bcb76/cfgov/cfgov/settings/base.py#L79).

rosskarchner:
Good call-- we probably can't run those migrations, etc until prod is on MySQL 5.7

grapesmoker:
Well, the migrations won't work on anything other than 5.7 because it won't have the JSON field. The compatibility wrt 1.8 will just involve rewriting the parts that create indices to do so manually.

rosskarchner:
As Will's suggested, I've removed it from "OPTIONAL_APPS" and put it behind a DEPLOY_ENVIRONMENT check.

Once you've got a version with the manual index creation working, release it and update optional-public.txt in this branch

rosskarchner:
the build server has been upgraded to MySQL 5.7

grapesmoker:
Ugh, sorry, this is going to cause travis to do its thing all over again but the branch couldn't be merged because of updates to master so I had to pull master first...

cfgov/cfgov/settings/base.py
```diff
@@ -77,6 +77,9 @@
     'jobmanager'
 )
 
+if DEPLOY_ENVIRONMENT == 'build':                                                 
+    INSTALLED_APPS += ('eregs_core')
```
chosak: I think this needs to be `('eregs_core',)` -- without the trailing comma, this is just a string, not a tuple. So it can't be added to the tuple of apps.

rosskarchner: Sorry, that was my fault

PR #3091, opened 2017-06-29
Add a management command for updating "Data Snapshot" values

richaagarwal:
Paired with @hillaryj on this.  This is a follow-up to the work here - https://github.com/cfpb/cfgov-refresh/pull/3080 and will allow our Jenkins job to update these Wagtail values monthly (or as needed). In addition to the mgmt command, there is now an image selector and help text for Data Snapshots (thanks to Hillary). The template will need some FEWD love.

To run:

    python cfgov/manage.py update_data_snapshot_values --snapshot_file=cfgov/v1/tests/fixtures/data_snapshots.json

Note that we are passing in a string representation of an array, rather than the array, for ease of piping in data within Jenkins as well as passing to the command. If there's a better way to do this, please let us know :)  

Also note that running the above will be successful even if you have no data snapshots in Wagtail with the corresponding market keys, but feel free to add some to a browse page if you want to see what it looks like locally.  For every page that does get updated, that will get logged. Can add more descriptive logging if needed/desired.

I'll be adding some tests to this but feel free to review in the interim.


cfgov/v1/management/commands/update_data_snapshot_values.py
```diff
@@ -0,0 +1,67 @@
...
+            snapshots
+        )
+        if snapshot:
+            return snapshot[0]
+        return None
+
+    def handle(self, *args, **options):
+        # Convert market data represented by a string,
+        # to its python representation (a list of dicts)
+        markets = ast.literal_eval(options['markets'])
```
chosak: Why not use JSON here?

cfgov/v1/management/commands/update_data_snapshot_values.py
```diff
@@ -0,0 +1,67 @@
...
+from v1.models.browse_page import BrowsePage
+from v1.tests.wagtail_pages.helpers import publish_changes
+
+
+class Command(BaseCommand):
+    help = 'Monthly updates to data snapshot values'
+
+    def add_arguments(self, parser):
+        """Adds all arguments to be processed."""
+        parser.add_argument(
```
chosak: What about supporting some kind of file or `stdin` input, that way you can run the command using

```sh
$ cat input.json | manage.py command
```

or

```sh
$ manage.py command -i input.json
```

See for example [this suggestion](https://stackoverflow.com/q/7576525). I'm curious why it's better/easier to construct a complex string input on the command line vs. using an input file.

cfgov/v1/management/commands/update_data_snapshot_values.py
```diff
@@ -0,0 +1,67 @@
...
+            nargs='?',
+            help='String representation of a list of markets'
+        )
+
+    def get_data_snapshots(self):
+        """ Gets all data snapshots from browse pages
+        Assumes there is one data snapshot per page
+        """
+        snapshots = []
+        for page in BrowsePage.objects.all():
```
chosak: Somewhat offtopic: I'm a bit wary of this design because accessing `page.specific.content.stream_data['type']` is a bit of a hidden/undocumented Wagtail thing and therefore potentially brittle.

Did you consider using a snippet type for this (like [`Contact`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/models/snippets.py#L47)), or maybe a Django model that can be embedded on a page (like [`GradePanel`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jobmanager/models/panels.py#L57))? That would provide a centralized place for editors to modify the data being embedded on pages, and still let one or more pages easily embed the data in the form of a block.

You also wouldn't need to republish the page once the data model gets updated, as it should automatically update when the underlying data does (caching aside).

cfgov/v1/management/commands/update_data_snapshot_values.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def handle(self, *args, **options):
+        # Convert market data represented by a string,
+        # to its python representation (a list of dicts)
+        markets = ast.literal_eval(options['markets'])
+
+        snapshots = self.get_data_snapshots()
+        for market in markets:
+            snapshot = self.find_data_snapshot(market['market_key'], snapshots)
+            if not snapshot:  # Market may not have been added to Wagtail yet
```
chosak: Failing silently like this might hide problems with the input; consider at least logging something if this happens?

richaagarwal: Sure, I can do that. I mentioned this above -- fwiw it's not really something I consider a failure

cfgov/v1/management/commands/update_data_snapshot_values.py
hillaryj: Since this is running in Jenkins, we were trying to do this without impacting the job's workspace. Storing a temporary file with the data is likely a more-robust approach, so we're planning to switch to something like the above.

cfgov/v1/management/commands/update_data_snapshot_values.py
richaagarwal: @chosak   I totally understand your concern here -- I have a few thoughts.
- I didn't seriously consider using a Django model for this because this is something that feels ripe for a CMS. While I would love for Wagtail to support programmatic updates of fields, I think that's a trade-off we've made (which doesn't feel relevant to this specific PR).
- Snippets is an interesting idea but it would deviate from how we're currently using them (for things that are "reusable" somehow).  I think it'd add a layer of abstraction for the editors trying to find and update text fields on a data snapshot since it's somewhat hidden from view on a page itself.  `ReusableText` is the most recent snippet and we had a lot of conversations around its usage.  I know you're more familiar with the Job Listings and the reasoning behind modeling it in Django, but I would be wary of making a habit of building things for our CMS outside of the CMS.
- I think it's OK that we republish this page, don't see that as a con. 

So, that all said, I agree it's not ideal to have to update data this way.  It is brittle, but if the script were to ever stop working, our tests (as I mentioned I plan to add as part of this PR) would fail.  Additionally, this script that will probably get run only once a month.

chosak: @richaagarwal thanks for the detailed reply.

One other thought was to have a custom page type. I love the idea of doing this in the CMS; it's just that it feels broken somehow to have to do a "look through all pages, find blocks of type X" to get to the blocks we want. Is it the case that there'll be one of these blocks on a specific set of pages? Could those pages just be a new page type (that inherits from `BrowsePage`, with this one addition), and then they get updated directly?

richaagarwal: @chosak Thank *you* for the feedback! I talked with @hillaryj about this suggestion, and I think we feel that this would add a bit of (mental) overhead to people working with these pages (as in remembering if they should be adding a new `BrowsePage` or a new subpage -- though perhaps that's not really a concern since they have to be trained in it one way or the other) as well as pollute the codebase a little. I think it would make sense if the page differed in any way, but if the page type exists solely to speed up this lookup, it feels a bit like overengineering, particularly since the script only runs once a month and isn't slowing anything down. I think we'd like to proceed with this for now but if you feel strongly let's talk more.

cfgov/v1/tests/management/commands/test_update_data_snapshot_values.py
```diff
@@ -0,0 +1,37 @@
...
+            # Adds a AUT market to a browse page
+            browse_page.content = StreamValue(
+                browse_page.content.stream_block,
+                [atomic.data_snapshot],
+                True
+            )
+            publish_page(child=browse_page)
+
+            # Call management command to update values
+            call_command('update_data_snapshot_values', '--snapshot_file=v1/tests/fixtures/data_snapshots.json')
```
richaagarwal: @chosak (and anyone else) -- I'd like to get your feedback on this approach for testing this management command.  Does this seem reasonable/ is there a more canonical way to add test fixtures?

cfgov/v1/tests/management/commands/test_update_data_snapshot_values.py
```diff
@@ -0,0 +1,37 @@
+from django.core.management import call_command
+from django.test import Client, TestCase
+
+from wagtail.wagtailcore.blocks import StreamValue
+
+from scripts import _atomic_helpers as atomic
+from v1.models.browse_page import BrowsePage
+from v1.tests.wagtail_pages.helpers import publish_page
+
+django_client = Client()
```
chosak: This is unnecessary since your test case inherits from `django.test.TestCase`, as it automatically [gets one as `self.client`](https://docs.djangoproject.com/en/1.8/topics/testing/tools/#django.test.SimpleTestCase.client). The `Client` class also maintains some state between requests (like cookies and such) so defining it outside of the tests like this could potentially allow tests to affect each other in some subtle way.

cfgov/v1/tests/management/commands/test_update_data_snapshot_values.py
chosak: @richaagarwal this sees good to me, if I understand what you're asking correctly, which is how to include test files to test management commands with (as opposed to test fixture data [like this](https://docs.djangoproject.com/en/1.8/topics/testing/tools/#django.test.TransactionTestCase.fixtures)).

The only concern I'd have about this is because you're using a relative path, this might not necessarily work if you're not running the test from the `cfgov` directory. In the past I've done some ugliness [like this](https://github.com/cfpb/owning-a-home-api/blob/master/ratechecker/tests/data/test_data.py#L10) that uses a relative path to the current file. Another way might be to use `settings.PROJECT_ROOT`, like [here](https://github.com/cfpb/cfgov-refresh/blob/6e44c2f113d15510ea565ca2493fb6f87bea2a80/cfgov/cfgov/test.py#L67).

cfgov/v1/management/commands/update_data_snapshot_values.py
```diff
@@ -0,0 +1,79 @@
...
+        """ Look up data snapshot by the provided market key
+        Assumes there is one data snapshot per key
+        """
+        snapshot = filter(
+            lambda item: item['market_key'] == market_key,
+            snapshots
+        )
+        if snapshot:
+            return snapshot[0]
+        return None
```
chosak: Nitpicky, but a slightly more readable version of this:

```py
def find_data_snapshot(self, market_key, snapshots):
    for snapshot in snapshots:
        if snapshot['market_key'] == market_key:
            return snapshot
```

The final `return None` is unnecessary, and this makes it a bit more clear that it's the first match that gets returned.

cfgov/v1/management/commands/update_data_snapshot_values.py
chosak: @richaagarwal understood, thanks for the explanation. :+1:

cfgov/v1/management/commands/update_data_snapshot_values.py
chosak: One additional thing: I'm not sure this code would work properly in Python 3, where `filter` returns a generator instead of a list. This would also be an issue on [this line](https://github.com/cfpb/cfgov-refresh/pull/3091/files/e1d5f62d7f2d6c5b0f7f7ff42962b47277598e73#diff-54080eea21e188ef2f4b2d1969ad2233R37) where you check the result of a `filter` in an `if` statement. I know we're not on Python 3 yet but it might be good to get into the habit for the future.

Here's an example difference in behavior. In Python 2 you can evaluate the result of the `filter` as a boolean:

```sh
$ python
Python 2.7.11 (default, Mar  7 2016, 09:55:43) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.72)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> x = [1, 2, 3]
>>> y = filter(lambda z: z > 4, x)
>>> bool(y)
False
>>> type(y)
<type 'list'>
```

but in Python 3 this doesn't work properly because it returns an object of type `filter`.

```
$ python3
Python 3.5.1 (default, Feb 12 2016, 16:08:42) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.72)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> x = [1, 2, 3]
>>> y = filter(lambda z: z > 4, x)
>>> bool(y)
True
>>> type(y)
<class 'filter'>

cfgov/v1/tests/management/commands/test_update_data_snapshot_values.py
hillaryj: I may be missing something, but `django_client` doesn't appear to be used anywhere, so I'm removing it since it has `self.client`.

cfgov/v1/management/commands/update_data_snapshot_values.py
hillaryj: That's a useful link, thanks! I see that Python recommends using list comprehensions instead of `filter` for cross compatibility.

PR #3096, opened 2017-06-30
Ask subcategory prefixes

higs4281:
Ask CFPB: Add prefixes to subcategory labels

To work around the lack of Wagtail widget helpers for many-to-many fields
(such as `filter_horizontal` and `filter_vertical`) we need to make
selections friendlier for Ask CFPB editors.

This PR adds category prefixes to subcategory __str__ representations,
plus some other helper-text improvements and code cleanup.

The [main change](https://github.com/cfpb/cfgov-refresh/blob/00e67ded9d3e57e1ae42a13bc65e16dcc3c5462a/cfgov/ask_cfpb/models/django.py#L231-L232) gives editors a way to more easily see and choose relevant subcategories:

subcat_prefixes

### DIFF
--- a/cfgov/ask_cfpb/models/django.py
+++ b/cfgov/ask_cfpb/models/django.py
@@ -229,22 +229,11 @@ class SubCategory(models.Model):
     ]
 
     def __str__(self):
-        return self.name
+        return "{}: {}".format(self.parent.name, self.name)
 
     class Meta:
         ordering = ['weight']
         verbose_name_plural = "Subcategories"
-
-
-class SubcategorySelectionWidget(forms.CheckboxSelectMultiple):
-
-    def __init__(self, attrs=None, answer=None):
-        super(SubcategorySelectionWidget, self).__init__(attrs, choices=())
-        # choices can be any iterable, but we may need to render this widget
-        # multiple times. Thus, collapse it into a list so it can be consumed
-        # more than once.
-        if answer:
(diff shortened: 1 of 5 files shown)


cfgov/ask_cfpb/wagtail_hooks.py
```diff
@@ -8,6 +8,7 @@
 
 from django.conf import settings
 from django.utils.html import format_html, format_html_join
+# from haystack.query import SearchQuerySet
```
chosak: Remove this if not needed?

cfgov/ask_cfpb/models/django.py
```diff
@@ -183,6 +183,59 @@ class Meta:
...
+        FieldPanel('related_subcategories',
+                   widget=forms.CheckboxSelectMultiple),
+    ]
+
+    def __str__(self):
+        return "{}: {}".format(self.parent.name, self.name)
+
+    class Meta:
+        ordering = ['weight']
+        verbose_name_plural = "Subcategories"
```
chosak: Nitpicky: I'm not sure [this is supposed to be uppercase](https://docs.djangoproject.com/en/1.11/ref/models/options/#verbose-name-plural); I thought its use in the admin should uppercase as necessary.

higs4281:
Thanks for the catches, @chosak !

PR #3102, opened 2017-07-06
Simplify Ask CFPB content export script

higs4281:
After trying to get a UTF-8 CSV to work for staffers using Excel,
we punt and export it in windows-1252. This eliminates the need for a 
separate Spanish export.

We also added an optional path argument for the resulting CSV, to replace the default `/tmp/`.  
The file name will include the current date.

## Testing
If you have a recent local copy of production data, running the following command  
should drop a big dated CSV file in your `/tmp` directory, and it should open cleanly with Excel.  
 ```
python cfgov/manage.py runscript export_ask_data
``` 

## Docs
Our [Ask CFPB documentation](https://github.cfpb.gov/CFPB/hubcap/wiki/Exporting-the-data-to-spreadsheets) has been updated.

### CODE REVIEW

File: cfgov/ask_cfpb/scripts/export_ask_data.py [hunk 1]
```diff
@@ -99,51 +75,38 @@ def assemble_output(spanish_only=False):
...
-                f.write("\t".encode('UTF16').join(
-                    [row[key] for key
-                     in SPANISH_HEADINGS]) + '\n'.encode('UTF16'))
-        else:
-            writer = csvkit.UnicodeWriter(f)
-            writer.writerow(HEADINGS)
-            for row in assemble_output():
-                writer.writerow([row[key] for key in HEADINGS])
+    slug = 'ask-cfpb-{}.csv'.format(timestamp)
+    if path is None:
```
richaagarwal: What do you think about setting the default in the definition, e.g. `def export_questions(path='/tmp')` ?

richaagarwal:
I downloaded the latest production db and ran this command, but I get the following error:
```
  File "/Users/agarwalr/apps/cfgov-refresh/cfgov/ask_cfpb/scripts/export_ask_data.py", line 103, in export_questions
    for row in assemble_output():
  File "/Users/agarwalr/apps/cfgov-refresh/cfgov/ask_cfpb/scripts/export_ask_data.py", line 63, in assemble_output
    output['Topic'] = answer.category.first().name
AttributeError: 'NoneType' object has no attribute 'name'
```

It looks like the data isn't there, somehow? Do I need to run anything else first?

richaagarwal:
Additionally, is there already a unit test for this export? If not, could one be added? :)

higs4281:
@richaagarwal Thank you for the review.
- Yes to tests! The other project scripts were experimental or one-use code, but not this one.
- re default args: I fell into the pattern of avoiding mutable default args after being called out on one by @chosak (my memory could be faulty there). But I have been bitten by them as described [here](http://effbot.org/zone/default-values.htm). In this case i think either way would be fine.
- I'll try to duplicate the missing-category bug. I ran the script for the product owner right after downloading the latest dump and did not get the error.

higs4281:
@richaagarwal Aha! Someone has since created a new answer and has yet to assign a category. Guess I'll have to trap for that.

PR #3106, opened 2017-07-06
Generate housing counselor JSON and PDFs in bulk

chosak:
This commit adds two new Django management commands to the `legacy` app: `hud_generate_json` and `hud_generate_html`. These commands are intended to support bulk generation of JSON and PDF files for the website's ["find a housing counselor"](https://www.consumerfinance.gov/find-a-housing-counselor/) tool.

The `hud_generate_json` command downloads HUD counselor data from [data.hud.gov](https://data.hud.gov/housing_counseling.html), cleans it, and sticks it into an in-memory SQLite database, indexed by location (latitude/longitude). An external reference of zipcode -> lat/lng pairs (downloaded from [the Census.gov website](https://www.census.gov/geo/maps-data/data/gazetteer2016.html)) is used to then lookup the closest counselors for every zipcode in the country. A JSON blob is generated for each zipcode and dumped to a target directory, with filenames like `20001.json`. Command usage is like:

```sh
$ python manage.py hud_generate_json /path/to/zipcode/reference.txt target_dir
```

The `hud_generate_html` command then can be used to process that directory full of JSON files to generate self-contained HTML files to be fed into a PDF encoder. "Self-contained" here means that all CSS is included inline and also that the single CFPB logo contained within is embedded as a data URI. This is to avoid having to make multiple requests when encoding as PDF. A new
`housing_counselor_pdf_selfcontained.html` template defines the HTML used by this command. Filenames are generated like `20001.html`, and command usage is like:

```sh
$ python manage.py hud_generate_html input_dir taget_dir
```

The JSON generated and used by this change varies slightly from that generated by the existing [django-hud](https://github.com/cfpb/django-hud) app. Here, the counselor services and languages are included in the JSON as a JSON array, instead of being converted to a comma-separated list.

The data cleaning and standardization behavior introduced here is primarily copied from the [`load_hud_data` management command](https://github.com/cfpb/django-hud/blob/master/hud_api_replace/management/commands/load_hud_data.py) in the [django-hud](https://github.com/cfpb/django-hud) project.

This code also includes a simple geocoder ~~(currently commented out)~~ that has the capability to add missing latitude/longitude information for counselors that are missing it; this just uses the zipcode centroid to avoid making an external geocoder call. This copies [some legacy behavior](https://github.com/cfpb/django-hud/blob/master/hud_api_replace/management/commands/load_hud_data.py#L160) from the [django-hud](https://github.com/cfpb/django-hud) project. ~~Because of the way we query the HUD API, this is actually never needed, but the code remains should it be useful in future.~~ (I was wrong about this, see comments [below](https://github.com/cfpb/cfgov-refresh/pull/3106#pullrequestreview-49277259)).

## Additions

- New `hud_generate_json` and `hud_generate_html` management commands in the `legacy` app, plus various supporting files.

## Testing

1. New unit tests should provide fairly comprehensive coverage of the new code.
2. Commands can be run using the examples above, using the text file contained within [this census.gov download] as a zipcode reference file.

## Notes

- See sibling PR #3107 for implementation of code that serves this bulk generated data.

## Checklist


cfgov/core/utils.py
```diff
@@ -35,3 +35,15 @@ def extract_answers_from_request(request):
     answers = [(param.split('_')[1], value) for param, value in
                request.POST.items() if param.startswith('questionid')]
     return answers
+
+
+def slice_list(l, n):
```
richaagarwal: Could you add a docstring here describing what this does?  Python has a lot of slicing functions so it'd be good to clarify *how* exactly this slices a list (it's clearer in the tests but not here in the code). I ran it locally with some different inputs hoping that would make it clearer, but unfortunately I'm unable to recreate `list(slice_list([1, 2, 3, 4, 5], 3))` (from your test) because I get a `TypeError` (`'list' object is not callable`).

richaagarwal: I just noticed this function isn't actually called anywhere -- is that intended?

chosak: @richaagarwal nice catch; I'd introduced this when experimenting with multithreaded. I've removed this in 51c98e6.

cfgov/legacy/housing_counselor/generator.py
```diff
@@ -0,0 +1,155 @@
...
+
+def get_db_connection():
+    connection = sqlite3.connect(':memory:')
+    connection.create_function('distance_in_miles', 4, distance_in_miles)
+
+    return connection
+
+
+def fill_db(connection, counselors):
+    def prepare_counselors(counselors):
```
rosskarchner: So it's not a ton: but we need to handle counselors without coordinates, which means geocoding them.

cfgov/legacy/housing_counselor/geocoder.py
```diff
@@ -0,0 +1,42 @@
...
+from six import print_
+
+
+def geocode_counselors(counselors, **kwargs):
+    """Fill in missing latitude/longitude data for housing counselors."""
+    return ZipCodeBasedGeocoder(**kwargs).geocode(counselors)
+
+
+class ZipCodeBasedGeocoder(object):
+    """Fill in missing latitude/longitude data using zipcode locations.
```
rosskarchner: This is *basically* what we do now, but I think it's inadequate, especially since we display results on a map

cfgov/legacy/housing_counselor/generator.py
chosak: @rosskarchner thanks for the review; see the second item at GHE#989. This isn't needed right now because we currently do a geo-based lookup at the HUD API, which only returns counselors that have coordinates (presumably that's how they get retrieved). If we wanted to change to retrieve all counselors, even those without coordinates (which, again, we don't do currently), then, yes, we'd have to geocode those.

cfgov/legacy/housing_counselor/geocoder.py
chosak: That's what I though too, but in practice this doesn't seem to happen right now. I was surprised and spent some time in this code, but it really does seem like [these lines](https://github.com/cfpb/django-hud/blob/master/hud_api_replace/management/commands/load_hud_data.py#L158) aren't ever executed right now (at least with the data I could experiment with). I suspect that because we do a radius-based search it doesn't return any counselors without coordinates.

rosskarchner: After that line you reference (on django-hud), I added a print statement, ran the data loader, and it printed 23 times.

I wonder what's different?

chosak: Ah, darn it. You are right! There are indeed 23 hits for `"agc_ADDR_LATITUDE":"0"`. I was erroneously assuming that the lat/lng would be missing but that's apparently not the case. My bad!

To maintain compatibility with current django-hud behavior, what if I modify this so that it does the same zipcode-based geolocation as the existing functionality? The code is already written so it should just be a one line change plus some additional unit tests.

cfgov/legacy/housing_counselor/generator.py
chosak: See below; I'm wrong about this. Will fix.

cfgov/legacy/housing_counselor/geocoder.py
rosskarchner: That would be fine--  we should maybe open a platform issue that follows https://github.com/cfpb/django-hud/issues/20 . I think it *is* something we need to fix, but we should see what the product owners think.

chosak:
@rosskarchner I've addressed your concerns about counselors with zero coordinates in 96c7477 and b105447. This now does zipcode-based geocoding for 23 counselors in the current dataset. I updated the PR description above.

I also found an example of a live page that's showing a 0-distance counselor using the existing code, see [this search for zipcode 63965](https://www.consumerfinance.gov/find-a-housing-counselor/?zipcode=63965). As you suggest I think that should be a separate issue than this refactor.

cfgov/legacy/housing_counselor/fetcher.py
```diff
@@ -0,0 +1,65 @@
...
+    if not counselors:
+        raise RuntimeError('Could not download HUD counselors')
+
+    print_('Retrieved', len(counselors), 'counselors', flush=True)
+    return counselors
+
+
+def replace_abbreviations(counselors, attribute, url):
+    """Replace attribute abbreviations with names from a given URL."""
+    print_('Downloading counselor', attribute, 'from', url, flush=True)
```
richaagarwal: Could this be logged instead of using `print` ?

cfgov/legacy/templates/hud/housing_counselor_pdf_selfcontained.html
```diff
@@ -0,0 +1,281 @@
+{% load staticfiles %}
```
richaagarwal: I'm no front-end expert but is it possible for the js and css to be split out of this?

cfgov/legacy/management/commands/hud_generate_json.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def load_zipcodes(filename):
+    """Load zipcode location data from Census gazetteer file.
+
+    See https://www.census.gov/geo/maps-data/data/gazetteer2016.html
+
+    Returns a tuple: (zipcode, latitude_degreees, longitude_degrees)
+    """
+    print_('Reading zipcodes from', filename, flush=True)
```
richaagarwal: Ditto with logging

cfgov/legacy/housing_counselor/fetcher.py
chosak: Yeah, that's a good suggestion. I agree that using logging is probably better than using print like this. The reason I use this complicated `print_` to be honest is to get the `flush=True` behavior which isn't built into Python 2.7 yet, and this is future compatible with Python 3. I wanted to be able to flush after every statement so that slow-running jobs on e.g. Jenkins would update the output for review while they're still running.

That being said, the complication with using `logging` here instead would be that we don't want this code to log anything during unit tests, but we do when this code is running as part of a management command. I think because we don't have `LOGGING` defined in our `settings.test` I wouldn't have to do anything special to avoid unit test output, but I would then have to modify how this is run to get output on Jenkins. Currently there this is run using `DJANGO_SETTINGS_MODULE=cfgov.settings.test_nomigrations` to avoid migrations and avoid the need for MySQL. To enable logging I'd then need either another settings file or some kind of environment-variable based toggle.

Now that I write all of this out I think that's probably the better approach. Thanks for suggesting, I'll experiment with implementing that.

cfgov/legacy/templates/hud/housing_counselor_pdf_selfcontained.html
chosak: Yes, it should be, mainly the reason I did this as a single self-contained file is to make it easier for the downstream wkhtmltopdf to process. See also GHE#989 under "PDF self-contained template".

cfgov/legacy/management/commands/hud_generate_json.py
chosak: Thanks, addressed [above](https://github.com/cfpb/cfgov-refresh/pull/3106#discussion_r126949393).

chosak:
@richaagarwal I tried to address the logging concerns in c4c246d.

PR #3116, opened 2017-07-10
Add date fields to ChartBlock (for automation) + DataSnapshot updates

richaagarwal:
## Deploy TODOs: 
- Because this updates templates with a new database field for ChartBlocks, existing pages will not render properly until the management command is run. This can be run immediately after deploying via the Jenkins job (which needs to be updated so this shouldn't get merged until it does)
- If you feel like running/testing this locally, you can run `python cfgov/manage.py update_chart_block_dates --snapshot_file=cfgov/v1/tests/fixtures/data_snapshots.json`. Going to a page like this http://localhost:8000/data-research/consumer-credit-trends/credit-cards/lending-neighborhood-income-level/ should then work (and display the dates based on the test file).

## Overview of changes
- Gave what was previously `data_month` a more descriptive name - it is now called `last_updated_projected_data` 
- `last_updated_projected_data` is now a `DateBlock`. Formatting is taken care of in the template.
- `last_updated_projected_data` and `date_published` added to `ChartBlock` as well as a management command for updating these two (will be called via @hillaryj's Jenkins job). This automates the updating of the two dates used in the text below every ChartBlock, e.g. `May 2017` and `March 2017` from the text `Date published: May 2017
Note: Data from the last six months are not final. The most recent data available in each visualization is for March 2017.` (found here https://www.consumerfinance.gov/data-research/consumer-credit-trends/credit-cards/lending-neighborhood-income-level/)
- Added unit tests for `ChartBlock` since this was missing coverage before
- Updated test for DataSnapshot management command and added test for ChartBlock management command
- Tests all pass but if you're curious, the three relevant tests can be run individually like so
  - `tox -e fast -- v1.tests.atomic_elements.organisms.test_organisms`
  - `tox -e fast -- v1.tests.management.commands.test_update_data_snapshot_values`
  - `tox -e fast -- v1.tests.management.commands.test_update_chart_block_dates`
- Modified the data snapshot template to mimic how it currently looks (as in a half width link blob group). See screenshot below (compare to https://www.consumerfinance.gov/data-research/consumer-credit-trends/student-loans/).
screen shot 2017-07-10 at 3 24 45 pm


hillaryj:
Looking good! PRs are always better with Winston/puppy photos. 😄 

The page crashes (due to `strftime`) on any page that has been migrated but hasn't been updated with a date but no live examples exist that aren't test pages. We may want to handle improperly-formatted or empty dates, although it's a required field.

cfgov/jinja2/v1/_includes/organisms/data_snapshot.html
```diff
@@ -41,14 +43,13 @@
         {{ value.year_over_year_change_text }}</h3>
     </div>
 
-    {% if value.image %}
```
chosak: This assumes that `image` is always present yet the help text [above](https://github.com/cfpb/cfgov-refresh/pull/3116/files#diff-bf1cda2d1a0671feb046fdb53fac3546L21) indicates that it is optional. Also the help text still refers to `value.data_month`.

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -743,6 +743,11 @@ class ChartBlock(blocks.StructBlock):
     description = blocks.CharBlock(
         required=True,
         help_text='Briefly summarize the chart for visually impaired users.')
+
+    last_updated_projected_data = blocks.DateBlock(
+        help_text="Month of latest entry in dataset"
```
chosak: Nitpicky: can you please use single quotes for consistency with most of the other strings in this file?

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -834,11 +839,10 @@ class DataSnapshot(blocks.StructBlock):
         max_length=20,
         help_text="Percentage change, e.g. '5.6% increase' or '7.8% decrease'"
     )
-    data_month = blocks.CharBlock(
-        max_length=15,
-        help_text="Month of latest entry in dataset in 'Month YYYY' format"
-    )
 
+    last_updated_projected_data = blocks.DateBlock(
+        help_text="Month of latest entry in dataset"
```
chosak: Same nitpick as above re: quotes.

cfgov/v1/management/commands/update_data_snapshot_values.py
```diff
@@ -65,7 +67,14 @@ def handle(self, *args, **options):
...
             # Update snapshot fields with the provided values
-            snapshot['data_month'] = market['data_month']
+            last_updated = datetime.strptime(
+                market['data_month'],
+                '%B %Y'
+            )
+            snapshot['last_updated_projected_data'] = datetime.strftime(
+                last_updated,
+                '%Y-%m-%d'
+            )
```
chosak: One minor gotcha here is that the Wagtail `DateBlock` doesn't necessarily always use this format, it's just the default. This works because you don't specify a custom format when you define the `DateBlock` [here](https://github.com/cfpb/cfgov-refresh/pull/3116/files#diff-dcf2a8703e945442ba8591d07085a645R843), but I wonder if it might be better to be more explicit and just refer to [`DEFAULT_DATE_FORMAT`](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailadmin/widgets.py#L25) here. For even more explicitness you could define some storage date format as `DataSnapshot.DATE_STORAGE_FORMAT` and then use that when you define the `DateBlock` like `blocks.DateBlock(format=DATE_STORAGE_FORMAT)` and here when you do the formatting.

richaagarwal:
Thanks for your detailed feedback, Andy!  I will add some more tests.  Regarding:

> As your update_data_snapshot_values management command gets more and more complicated, I wonder how you might feel about putting this parsing/storage logic into the DataSnapshot class itself, like some kind of DataSnapshot.update_from_json or something. That way all of the logic for the snapshot fields would be right next to where the fields are designed.

Unfortunately, since `DataSnapshot` isn't a standard Django class, it's not really easy to manipulate in this way.  For instance, in the tests, it would be difficult to iterate over the DataSnapshot objects (it appears if you call a page's `content` that you'd get a list of all objects, but they all just show up as `wagtail.wagtailcore.blocks.stream_block.StreamChild object` and I can't tell how you'd filter them down, whereas it's slightly more straightforward to do it the way I am now where I filter on `content.stream_data`.  If you have an idea on how to approach this though I'd love to hear!

richaagarwal:
This is ready for review again. There are a few migrations, so I'd like to merge this in soon if there's nothing critical blocking it (& can continue to address feedback after it's merged). Thanks!

hillaryj:
I'm having some odd display issues:

- The language list is a bulleted list instead of the one-line display it normally is
- The test chart isn't displaying but also doesn't give any errors in the console

screen shot 2017-07-14 at 2 24 24 pm

richaagarwal:
@hillaryj Thanks for reviewing! The bulleted list was reported earlier this week and is unrelated.  

Which test chart isn't displaying?  In your screenshot the snapshot shows up as expected, but not sure which test chart you're referring to.

PR #3132, opened 2017-07-13
Include imagemin dependencies

anselmbradford:
In Node7 gulp-imagemin throws `gulp-imagemin: Couldn't load default plugin` warnings when running the build script. The plugins need to be explicitly included in package.json to resolve this issue. 

## Additions

- Adds `imagemin-gifsicle`, `imagemin-jpegtran`, `imagemin-optipng`, and `imagemin-svgo`.

## Changes

- Updates image gulp task syntax to ES6.

## Testing

1. `./frontend.sh` should build.

## Notes

- It isn't necessary to require and include the plugins in `image.js` as they will be sourced by gulp-imagemin by default, but I included them so that it's clearer where they are actually used in the codebase when referencing them in `package.json`.


sebworks:
looks good overall, there is an linter error though. Thanks for the config change.

anselmbradford:
Please re-pull and test. Thanks!

PR #3133, opened 2017-07-14
Remove obsolete library-resources entry from urls.py

Scotchester:
These files were removed in the latest fin-ed-resources release. This was causing a problem for me when running locally, though for some reason is not creating problems on Build.

## Removals

- `library-resources` URL from `urls.py`

## Checklist


chosak:
@Scotchester can you clarify what the desired behavior is here? In the current master, hitting the `/library-resources/` locally returns a 404 if you don't have the SheerSite installed at `../fin-ed-resources`. Right now this returns a 404 on build because `/library-resources/` seems to redirect to `/educational-resources/library-resources/`. But these pages do seem to exist on both beta and production.

How is this intended to work? Where should these pages be getting served from, if at all?

Scotchester:
Beta and production are correct. The page is not served by Wagtail. It's 404ing on build because the database there is way out of date with prod. This PR's purpose is to fix the local errors.

Scotchester:
The errors I was getting locally were not 404s, though. It was happening on running the Django server. Installing the latest fin-ed-resources should replicate that for you, I think.

PR #3135, opened 2017-07-14
Add gtm ignore to JS only links

anselmbradford:
Links that contain `javascript:void(0)` for the `href` will register Google analytics data for a link click, but won't navigate anywhere so should be ignored in analytics data. Adding the `data-gtm_ignore` custom attribute used by the analytics team will ignore clicks on these links for analytics purposes.

## Additions

- `data-gtm_ignore` added to "Paperwork Reduction Act statement" link on find a housing counselor results page.

## Changes

- Removed spaces in email address link that prevented it from working properly.
- Fixed some spacing in the HTML (source code cosmetic only)

## Testing

1. Visit /find-a-housing-counselor/, search for a zip, and check that the Paperwork Reduction Act at the bottom still shows/hides the associated text, but also has a `data-gtm_ignore="true"` attribute on the link.

## Notes

- Part of https://github.cfpb.gov/CFGOV/platform/issues/965

### DIFF
--- a/cfgov/legacy/templates/hud/housing_counselor.html
+++ b/cfgov/legacy/templates/hud/housing_counselor.html
@@ -18,10 +18,12 @@ Find a housing counselor
     <li><a href="/">Home</a></li>
         <li>Find a housing counselor</li>
   </ul>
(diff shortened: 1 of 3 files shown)


anselmbradford:
Holding as we might move the logic for this into GTM and just check the href value instead of requiring attributes on these links.

anselmbradford:
Closing in favor of https://github.com/cfpb/cfgov-refresh/pull/3157 per @mistergone 's great non-linking link should be a button suggestion!

PR #3141, opened 2017-07-17
Implement global browser supported environment variable

anselmbradford:
## Additions

- `getSupportedBrowserList` function in `environment.js` for returning an array of supported browsers in a [browserlist](https://github.com/ai/browserslist) compatible format for use in the script and style build processes.

## Changes

- Changes babel environment to `env` from `es2015` and `stage-2`, which then gets passed browserlist.
- Updates variables to `const.

## Removals

- Removes babel presets from `.babelrc` since it's in the webpack-config.

## Testing

2. `gulp styles:ondemand` should pass.
2. `gulp scripts:ondemand` should pass.
3. Ideally should be checked in the browsers we support and make sure there are no changes.

## Screenshots

With `debug: true` set in the babel loader, this helpful output happens at build, which seems helpful to keep:
screen shot 2017-07-17 at 2 22 38 pm

## Notes

- The `'Edge >= 11'` was not in the original browser list, but is in Mosaic, so added it here for consistency.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,4 @@
 {
-  "presets": ["es2015", "stage-2"],
   "env": {
     // only happens if NODE_ENV is undefined or set to 'development'
     "development": {
--- a/config/webpack-config.js
+++ b/config/webpack-config.js
@@ -4,17 +4,18 @@
 
 'use strict';
 
-var BannerFooterPlugin = require( 'banner-footer-webpack-plugin' );
-var path = require( 'path' );
-var paths = require( '../config/environment' ).paths;
-var scriptsManifest = require( '../gulp/utils/scripts-manifest' );
-var webpack = require( 'webpack' );
+const BannerFooterPlugin = require( 'banner-footer-webpack-plugin' );
+const path = require( 'path' );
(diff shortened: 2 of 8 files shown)


.babelrc
```diff
@@ -1,5 +1,4 @@
 {
-  "presets": ["es2015", "stage-2"],
```
sebworks: Thanks for this.

config/webpack-config.js
```diff
@@ -4,25 +4,34 @@
...
       test: /\.js$/,
       use: [ {
-        loader: 'babel-loader?cacheDirectory=true'
+        loader: 'babel-loader?cacheDirectory=true',
+        options: {
+          presets: [ [ 'env', {
+            targets: {
+              browsers: environment.getSupportedBrowserList()
+            },
+            debug: true
```
sebworks: could this be a flag that you pass in?

anselmbradford: Could be, but since there's already non-optional output generated by the build process it seemed helpful to always display what browsers are being targeted. Would there be a reason to turn it off?

sebworks: good point

PR #3148, opened 2017-07-18
Fix for hero image

sebworks:
Fix for hero image

## Changes

- Modified `cfgov/jinja2/v1/_includes/molecules/hero.html` to remove `image-set` properties.

## Testing

1. Add hero to page
2. Add background image to page.
3. Visit page using Chrome.
4. Verify image has the proper width.

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/molecules/hero.html
+++ b/cfgov/jinja2/v1/_includes/molecules/hero.html
@@ -37,7 +37,9 @@
    ========================================================================== #}
 
 
-{# TODO: This is simple now, but in the future we'll hopefully be able to provide 1x and 2x values as well. We should update this to be an image array delivered by Django in the future. #}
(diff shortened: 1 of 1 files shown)


jimmynotjim:
That's so weird. We tested it thoroughly in the CF Docs site and Sandbox in Chrome, Safari, FF, and IE/Edge. Especially weird because all the options were currently the exact same image since we don't have a way to serve multiple from the backend at the moment. Hopefully they work it out soonish.

Scotchester:
Hmm... Now I'm wondering if what I said was accurate. I can think of
something else that might be going on. Let's discuss later.

jimmynotjim:
Just for future clarity, the issue was we provided a 1x image but telling the browser it's a 2x image, so it downsized it by 50%. When I tested, I tested with a 2x image, not realizing what would happen the other way around.

PR #3169, opened 2017-07-24
Refactor layout to remove bespoke .press-information

anselmbradford:
## Removals

- Removes `.press-information` CSS class.

## Changes

- Replaces `.press-information` class for laying out press resources article footer on newsroom-page template and replaces it with cf-layout `content-l` classes.

## Testing

1. `gulp build`
2. Compare http://localhost:8000/about-us/newsroom/cfpb-issues-proposal-provide-flexibility-certain-mortgage-lenders-collecting-information/ to https://www.consumerfinance.gov/about-us/newsroom/cfpb-issues-proposal-provide-flexibility-certain-mortgage-lenders-collecting-information/

## Screenshots

Before:
screen shot 2017-07-24 at 4 35 40 pm

After:
screen shot 2017-07-24 at 4 35 34 pm

## Notes

- Using the `content-l` classes the width of the column has been reduced (see horizontal line length in before/after screenshots compared to "HDMA" tag). Is this a problem? cc @nataliafitzgerald


sebworks:
@anselmbradford, what is the amount of the width reduction?

anselmbradford:
@sebworks looks like it was 7 columns before and this makes it half-width, so 6 columns? Did I do that right?

sebworks:
yep, looks good.

Scotchester:
I don't see a need to constrain the width at all. It seems fine at full width, to me:

screen shot 2017-07-24 at 17 40 26

anselmbradford:
How about full width for the line and 3/4 width for the text:
screen shot 2017-07-25 at 11 00 33 am

Scotchester:
I'm mainly looking for a way to get rid of the extra markup that using ` content-l` necessitates. It feels wrong to use one when there's no need to lay things out in columns.

A designer (pinging @nataliafitzgerald again, or maybe @ajbush, as one of the original V1 designers) should still weigh in, but if we must make it narrower, how about using one of our `u-w##pct` [utility classes](https://github.com/cfpb/capital-framework/blob/canary/src/cf-core/src/cf-utilities.less#L221-L234)?

nataliafitzgerald:
@Scotchester - If we use the utility class what would that look like?

Scotchester:
> If we use the utility class what would that look like?

Whatever you want, essentially. We have width utilities for every 10%, plus 25%, 33%, 66%, and 75%.

> The full width text option that you proposed looks a little wide for the content (long line length).

It was capping the line length at the standard maximum as most text on the website, 670px. If it looks long, it's probably a side effect of having "press office" be orphaned on the second line.

But if you like 75%, that's fine, I'd just prefer to do it with the `u-w75pct` utility class on the paragraph, instead of the extra `content-l` markup.

nataliafitzgerald:
Ok @Scotchester - Yeah, let's do the 75% approach then. Yeah, it's the orphan that's bugging me.

anselmbradford:
Shouldn't we leave the utility classes to layouts we can't, and will never intend to, make with our layout classes? If we have a pattern, we should be able to find it everywhere it's used the same way. 100% width line and 75% width paragraph can be done with cf-layout. I know it's a little more markup, but not much, just an extra div, and well… the work is already done 😉

Scotchester:
I generally agree with your description of how we should use utility classes, but I would also add that I think a good part of their purpose is to avoid jumping through a lot of extra markup/CSS hoops for a one-off rule.

But I'm not sure I would call this a "pattern". In 15 months in production with this website, we've never had a similar situation. This seems like a one-off we'll never repeat, especially since the reasoning is simply to avoid an undesirable orphan at max width. Were it not for that, we wouldn't need a partial width at all.

As to "I know it's a little more markup, but not much, just an extra div"… c'mon, man ;) 

Current markup in this PR (which could be made more efficient by removing the outer `content-l` and `content-l_col`, but still necessitates a new div and a new wrapper around the link, to make sure it doesn't wrap to the right of the paragraph):

```html
    <aside class="content-l block">
        <div class="content-l_col content-l_col-1">
            <header class="m-slug-header">
                <h2 class="a-heading">
                    Press information
                </h2>
            </header>
            <div class="content-l">
              <p class="content-l_col content-l_col-2-3">
                  If you want to republish the article
                  or have questions about the content,
                  please contact the press office.
              </p>
              <p class="content-l_col content-l_col-1">
                <a class="a-link a-link__jump"
                   href="/about-us/newsroom/press-resources/">
                   Go to press resources page
                </a>
              </p>
            </div>
        </div>
    </aside>
```

Using a utility class:

```html
    <aside class="block">
        <header class="m-slug-header">
            <h2 class="a-heading">
                Press information
            </h2>
        </header>
        <p class="u-w75pct">
            If you want to republish the article
            or have questions about the content,
            please contact the press office.
        </p>
        <a class="a-link a-link__jump"
           href="/about-us/newsroom/press-resources/">
            Go to press resources page
        </a>
    </aside>
```

I will admit that I realized this will have the effect of not having it be full-width on mobile anymore, so that is one benefit to the `content-l` approach.

anselmbradford:
@Scotchester Updated!
screen shot 2017-07-27 at 1 34 27 pm

PR #1005, opened 2015-09-22
Adding Acceptance tests for the Careers pages

sebworks:
Acceptance tests for the `Careers` pages

## Changes
 - Added Acceptance tests for the `Careers` pages. 

## Testing
-  Run `gulp test:acceptance`

## Review
@anselmbradford 
@jimmynotjim 
@KimberlyMunoz

PR #3172, opened 2017-07-25
Ask pagination fix

higs4281:
Ask CFPB: consolidate pagination validation

Five Ask CFPB page types deliver answer sets that are paginated when
more than 20 answers are found. Each page type was individually validating 
pagination requests, introducing inconsistency and failing to make full use of
the Django paginator's validation. This PR introduces a helper function
that all pages use to validate a pagination request.

This also fixes broken pagination on Ask Audience pages, and should stop server errors 
generated by invalid pagination requests that were hitting build.

This addresses GHE issue `#1005`. Hat-tip to @virginiacc, who figured out the problem.

## Additions

- page-validation function
- test for non-integer page numbers in a pagination request

## Testing

Check these pages and play with pagination outside bounds (page=999) or non-integer page numbers. (A5)  
These sample Ask pages have paginated answer lists that can be tested: 
- [Answer search results](http://localhost:8000/ask-cfpb/search/?q=mortgages&btnSearch=)
- [Tag search results](http://localhost:8000/ask-cfpb/search-by-tag/checking_account/)
- [Audience page](http://localhost:8000/ask-cfpb/audience-older-americans/)
- [Category page](http://localhost:8000/ask-cfpb/category-auto-loans/)
- [SubCategory pages](http://localhost:8000/ask-cfpb/category-mortgages/understanding-mortgages/)

### DIFF
--- a/cfgov/ask_cfpb/models/pages.py
+++ b/cfgov/ask_cfpb/models/pages.py
@@ -67,19 +67,22 @@ def get_ask_breadcrumbs(category=None):
     return breadcrumbs
 
 
-def set_page(request, paginator):
+def validate_page_number(request, paginator):
     """
     A utility for parsing a request for a page number,
     handling errant or invalid page numbers
     and returning a valid page number and pagination page object.
     """
-    page_number = request.GET.get('page', 1)
+    raw_page = request.GET.get('page', 1)
     try:
-        page = paginator.page(page_number)
+        page_number = int(raw_page)
+    except ValueError:
+        page_number = 1
+    try:
+        paginator.page(page_number)
     except InvalidPage:
         page_number = 1
-        page = paginator.page(page_number)
-    return (page_number, page)
+    return page_number
 
 
 class AnswerLandingPage(LandingPage):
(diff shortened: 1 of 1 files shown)


cfgov/ask_cfpb/models/pages.py
```diff
@@ -67,6 +67,24 @@ def get_ask_breadcrumbs(category=None):
...
+    catching invalid page numbers and always returning
+    a valid page number, defaulting to 1.
+    """
+    raw_page = request.GET.get('page', 1)
+    try:
+        page_number = int(raw_page)
+    except ValueError:
+        page_number = 1
+    try:
+        paginator.page(page_number)
```
willbarton: Several places below you have this pattern:

```python
page_number = validate_page_number(request, paginator)
page = paginator.page(page_number)
```

It looks like everywhere we call `validate_page_number()` we're also calling `paginator.page()`. Would it make sense to simply have this function return `(page_number, page)`, and perhaps rename it something like `get_valid_page`? I don't know how expensive calls to `paginator.page` might be but that would reduce them, except in the case of an `InvalidPage` where we'd have to call it again with `1`.

higs4281: @willbarton I initially tried a similar pattern, passing the paginator around, but it kept breaking the paginator for subsequent requests. 
I could take another swing at it.

PR #3181, opened 2017-07-27
Add New Info Unit Groups

Scotchester:
## Additions

- A new Info Unit Group Wagtail module
- A new Info Unit Group organism template to render the above
- A new Info Unit molecule template to render Info Units belonging to the above
- Three new heading blocks that get grouped together in Wagtail (using the new `struct-with-block-wrapper-classes.html` `form_template` and the new `heading-blocks.css` for a simpler interface for creating headings at different levels, optionally with icons

## Changes

- Minor updates to the existing `m-info-unit_image` styling that help my new Info Units but don't break anything with existing info units

## Testing

1. Pull branch
2. `gulp styles`
3. `./runserver.sh`
4. Create a page, add an Info Unit Group to the content area, and play around with the various options.
5. Preview the page

## Screenshots

screen shot 2017-07-27 at 14 04 52

localhost-8000-testing-info-unit-groups-

## Notes

- I created brand new templates and left the old Image and Text and Link Blob stuff in place, so that we can get this in production without immediately worrying about migrating the old stuff. We can either migrate them by hand or write a script to do it in another PR.

## Todos

- [x] Finish tidying up the styling, most notably vertical spacing
- [x] Update/write tests
- [x] Take another pass at the templates' docstrings and make sure I didn't miss anything
- [x] Determine if we're using the stroke between 25/75s enough to warrant keeping that option, or if we use it enough that we may as well always use it
- [x] Determine if we can get rid of the option to use button-style links

## Checklist


Scotchester:
Added a todo to figure out what to do about lines between 25/75 items.

sebworks:
@Scotchester, what are chief differences between the existing `info-unit` and this one?

Scotchester:
> what are chief differences between the existing `info-unit` and this one?

Visually, there shouldn't really be any (pending the outcome of the last todo item about rule lines). 

Behind the scenes, just a bit of adjustment to the template, the primary effect of which is that decorative images are now correctly output as divs width a `background-image` if they have no alt text set.

On the Wagtail side, rather than having separate H3 & H4 fields, you can now choose the heading level you want and then optionally add an icon. I forgot to add a screenshot of the Wagtail interface originally; I've just added that to the OP.

sebworks:
This looks awesome

sebworks:
@Scotchester, why no make the icons a dropdown?

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
```diff
@@ -0,0 +1,136 @@
...
+   value.link_image_and_heading: Boolean for whether the image and heading
+                                 should be linked to the value.links[0].url.
+
+   ========================================================================== #}
+
+{%- macro info_unit(value, format, link_image_and_heading) -%}
+
+{# Confirm that not only is link_image_and_heading set, but that this
+   particular info unit has a link in its list. #}
+{%- set link_image_and_heading = link_image_and_heading
```
sebworks: We should move this logic out of the template if possible.

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
```diff
@@ -0,0 +1,136 @@
...
+<div class="m-info-unit
+            {{- ' m-info-unit__inline' if format == '25-75' else '' }}">
+
+{%- if value.image.upload %}
+    {%- set img = image(value.image.upload, 'original') %}
+    {% if link_image_and_heading %}
+        <a href="{{ value.links[0].url }}">
+    {% endif %}
+        {%- if value.image.alt %}
+            <img class="m-info-unit_image
```
sebworks: How would you feel about having an image atom? If you passed a `is_decorative` flag you would get back a `div`, else a an `img`. This way we could reuse the code across macros.

Scotchester:
> why no make the icons a dropdown?

Well, I didn't think about it too hard, to be honest. I just followed the example of the existing H4 icon text field. There are a _lot_ of icons, though. Unless we tried to pare it down to some subset allowed for this use case, a dropdown would be massive.

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
Scotchester: Good idea. I'll see if I can make that a property on the model or something.

cfarm:
How hard is it to make this a feature flag so we can test the change on build before making it available more widely? Platform team folks including myself @anselmbradford @nataliafitzgerald @ielerol can test this.

cfarm:
Alternately: give us a heads up when this is merged and we volunteer to test it heavily on build!

cfarm:
Here are some screenshots illustrating the margin that needs to be changed from 15 to 30px:

screen shot 2017-07-27 at 4 26 14 pm

How it looks with 30px margin above the blobs:
screen shot 2017-07-27 at 4 26 26 pm

cfarm:
Another visual bug - when adding more than 3 items to 33/33/33 layout, the new row doesn't have a margin-top or whitespace. I would expect it to have the same whitespace that a new row in 50/50 layout would have:

screen shot 2017-07-27 at 4 41 32 pm

cfarm:
For each link blob, there is this checkbox option in Wagtail:

"Show links as button:"

screen shot 2017-07-27 at 4 57 38 pm

What's the use case for this? Is there a scenario in which we want content editors to add some links as regular links, and some as buttons, within the same info unit? If not, can we improve the UI so you can make all links a button, otherwise default to regular link?

nataliafitzgerald:
@Scotchester 
First off, great work! This will solve a number of spacing issues we've been having as well as giving users more flexibility in terms of heading sizes and such. Merging the "Image text 50/50 group," "Half width link blob group," "Image text 25/75 group," and "Image text 33/33/33 group" into one mega module makes total sense.

One thing I noticed is that we are adding functionality to the "Image text 25/75 group," and "Image text 33/33/33 group" that didn't exist before: The option to use a button instead of a link. Before, the option to pick between a button and a link was only available for the "Image text 50/50 group," and "Half width link blob group."
screen shot 2017-07-27 at 4 57 51 pm

Have users (our team) asked for this added functionality or had a need for this in their projects? I don't know of any use cases that exist thus far (for any of the modules that are wrapped up into this mega module).

@ielerol - Any thoughts?

nataliafitzgerald:
@Scotchester @sebworks 
In terms of the icons, if a dropdown is too massive, how about a live/clickable link in the interface that takes you to the page where all the icon names are? This is something that is included in the existing interface but the link isn't clickable. We could use a link that will take you to the part of the icon page that is relevant to users that are looking for the names: https://cfpb.github.io/capital-framework/components/cf-icons/#icons

screen shot 2017-07-27 at 5 12 06 pm

ielerol:
in regards to showing links as buttons, our current guidance is to use them sparingly to keep their impact high. I can't think of many cases where it would be desirable to use a whole series of buttons in third-width link blobs, or to add a single button in a group of link blobs that otherwise have text links. In fact I haven't even found any examples where we're using buttons in 50/50s on current pages. The closest I've found is [the CCDB page](https://www.consumerfinance.gov/data-research/consumer-complaints/), which is being replaced with the new public complaint search, and in any case isn't in Wagtail.

I want a little more time to discuss with other designers in case there are examples I'm missing, but for now I'm leaning towards eliminating buttons as an option from this module altogether, and just leaving the full-width CTA as the main way to add buttons to the body of a page.

Scotchester:
Thanks for all the feedback! I will be working on addressing it today.

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
Scotchester: I'm interested in this idea, but worried that it would have to be really convoluted to work for all of our use cases. It seems like something best explored in a future PR.

Scotchester:
> In terms of the icons, if a dropdown is too massive, how about a live/clickable link in the interface that takes you to the page where all the icon names are?

@nataliafitzgerald Thanks for nudging me to dig deeper on that. I found an easy way to enable rendering of links in a field's help text! Fixed in 1c9a814fc1d863afc413591755457aca68ca2468.

Scotchester:
@cfarm Vertical spacing issues have been addressed in 0605626.

Scotchester:
@ielerol @nataliafitzgerald @cfarm 

I would be more than happy to remove the option to show links as buttons. What do we need to do / who else do we need to discuss with in order to be comfortable with that decision?

nataliafitzgerald:
@Scotchester - I think the new process is that we consult with @schaferjh or John Paul when there is a lack of consensus. It seems like there's consensus here but if we want to be safe I'd say we reach out to @schaferjh. I don't exactly know what consensus means in a technical sense (as in how many people need to agree on something to reach this consensus).

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
```diff
@@ -0,0 +1,136 @@
...
+   Create an info-unit molecule when given:
+
+   value:                        Object defined from a StreamField block.
+
+   value.modifier:               A modifier CSS class used to alternate
+                                 the info unit from stacked to inline layouts.
+
+   value.image:                  (Optional) An object containing:
+
+   value.image.is_decorative:    Whether the image is a bg image or not.
```
jimmynotjim: Did we decide we can't rely on `alt` text in determining this?

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
```diff
@@ -0,0 +1,136 @@
...
+                                 containing:
+
+   value.links[i].url:           A string for the URL of the link.
+
+   value.links[i].text:          A string for the text of the link.
+
+   value.is_button:              Boolean for whether the link should be
+                                 rendered as a button. Default is false.
+
+   value.link_image_and_heading: Boolean for whether the image and heading
```
jimmynotjim: I know there's been discussions about making this the default behavior and removing the option, did anything come of that?

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
```diff
@@ -0,0 +1,136 @@
...
+                    {%- endif -%}
+                >
+                    {{ value.heading | safe }}
+                </{{ value.heading_level }}>
+        {%- if link_image_and_heading %}
+            </a>
+        {%- endif -%}
+    {%- endif %}
+
+            {{ parse_links( value.body ) | safe }}
```
jimmynotjim: Is it weird that we allow links in the body and in the link list?

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
```diff
@@ -0,0 +1,136 @@
...
+                                 Default is false.
+
+   value.image.is_square:        Whether the image is square or not.
+                                 Default is false.
+
+   value.image.url:              String for the image URL.
+
+   value.image.alt:              String for image alt text.
+
+   value.heading:                The heading for the info unit.
```
jimmynotjim: Can this become an object like the image?

cfgov/jinja2/v1/_includes/organisms/info-unit-group-2.html
```diff
@@ -0,0 +1,68 @@
...
+
+   ==========================================================================
+
+   Description:
+
+   Create an info unit group organism when given:
+
+   value:                         Object defined from a StreamField block.
+
+   value.heading:                 The heading for the info unit group.
```
jimmynotjim: Same object question here as above.

cfgov/jinja2/v1/_includes/organisms/info-unit-group-2.html
```diff
@@ -0,0 +1,68 @@
...
+
+   value.heading_level:           The heading level (H2, H3, or H4).
+
+   value.heading_icon:            An optional cf-icons icon code that
+                                  triggers the insertion of an icon before
+                                  the heading.
+
+   value.intro:                   An introduction for the info unit group.
+
+   value.link_image_and_heading:  Boolean for whether the image and heading
```
jimmynotjim: Did this get copied here by accident? There's no links tuple in the group, only the individual info units

cfgov/jinja2/v1/_includes/organisms/info-unit-group-2.html
```diff
@@ -0,0 +1,68 @@
...
+        {{ value.heading | safe }}
+    </{{ value.heading_level }}>
+{% endif -%}
+
+{%- if value.intro %}
+    {{ value.intro | safe }}
+{%- endif %}
+
+    <div class="content-l">
+    {% for iu in value.info_units %}
```
jimmynotjim: Can we change `iu` to something clearer? Even `unit` would suffice

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
Scotchester: I think we can, and I am doing so in the template, I just haven't removed this from the docstring yet.

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
Scotchester: Possibly. I've thought about doing that. It would be more DRY if we had a `HeadingBlock` that rolled up the three new Heading fields in one. I'll play with that before finishing this PR.

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
Scotchester: I did make it having this be `True` by default, but I don't think we can remove the option. There are definitely (IMO) circumstances where we don't want it, like on the headings in the Half-Width Link Blob at the bottom of [the Careers page](https://www.consumerfinance.gov/about-us/careers/).

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
Scotchester: We can't disallow links in the body, because it's a rich text field.

cfgov/jinja2/v1/_includes/organisms/info-unit-group-2.html
Scotchester: The option is set at the group/organism level, which then gets passed to the render macro for each individual info unit in that group.

cfgov/jinja2/v1/_includes/organisms/info-unit-group-2.html
Scotchester: Sure thing, will do.

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
Scotchester: In responding to Jimmy, I realized I am pretty sure I can't do that, because it gets passed the value of the setting from the group organism, but it can only take effect if the individual info unit has a link with a URL. If there's a way for the InfoUnit molecule model to know the value of its parent InfoUnitGroup organism model's `link_image_and_heading` field, I'm not sure how to do it. Maybe @chosak or @richaagarwal would know?

ielerol:
I've asked @jimmynotjim and @keelerr if they know of any places where 50/50s are currently using buttons, and they haven't used them in any of their work, and have no objections to removing the option. @schaferjh might know some more context around the original decision to include buttons in that module.

I do think it would be a good idea to let the design team know generally that we want to make this change in case anyone does have a need, I can bring it up.

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
chosak: @Scotchester this might be possible in theory if we used `{% include_block %}` (search for `parent_context` in the Wagtail docs [here](http://docs.wagtail.io/en/v1.11.1/topics/streamfield.html#template-rendering)) but given the existing custom `render_stream_child` logic I don't know how easy that would be, as blocks don't have access to their parents. See backlog GHE#378.

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
Scotchester: Another example would be the "Survey of mortgage originations" link blob on the [Data & Research overview](https://www.consumerfinance.gov/data-research/). It has 3 links, and it's not particularly clear which should be the "lead" one, if there was no way to turn this functionality off.

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
Scotchester: Ah, right. Perhaps we can revisit this when we deprecate `render_stream_child` and move to `include_block`, then.

Scotchester:
I did a full review of just about every page Wagtail page on the site (excluding things like individual blog posts and report pages and such) and couldn't find a single example of a button at the bottom of an Image and Text or Link Blob module.

Scotchester:
Also, during my review, I had the opportunity to observe the usage of the horizontal rule lines between 25/75 items. I didn't take notes or anything, but it felt like they're used about half the time.

In my personal opinion, the spacing of items _without_ the horizontal rules feels a bit tight. I'd like to propose that we **always place a horizontal rule between 25/75 items**. Anyone strongly against doing that?

nataliafitzgerald:
@Scotchester 
Regarding making the horizontal rule line the default for 25/75's I'd want us (designers) to have a chance to take a look at this pattern across the site before deciding whether to move in that direction.

Scotchester:
@nataliafitzgerald Absolutely, please do!

Scotchester:
Quick note on timing before I leave for the day:

This needs to get merged and deployed no later than Tuesday afternoon in order for Rara to have the time they'll need to build the pages they're going to launch on Thursday. Monday morning I will finish the tests.

If the design decisions are not settled by Tuesday morning, I will move forward with what seems to be the majority opinion at the time. Note again that deploying this code will not affect any existing Image and Text or Link Blob modules on the site, so if we need to change course from what happens here, we can do that before migrating the existing modules to this new module.

nataliafitzgerald:
I took a look at the 25/75 pattern across a collection of pages. After looking at a variety of examples, **I do not feel that we should always place the horizontal rule below 25/75 items.** We have a lot of instances across the site where we are currently using the 25/75 without the rule line. Continuing to give designers the flexibility to determine whether a rule line makes sense or not for a specific situation makes the most sense to me (and will make designers feel less locked in). 

## Instances where no rule line makes sense:
There are a lot of examples of places where we're currently using the 25/75 without the rule line. Here are some examples. 
### Set of things: 
Instances where the 25/75 items are part of a group of similar things. For example, on the "Everyone has a story" page where we include a group of stories. These stories are part of a set and there is no need from a visual design perspective to emphasize the visual separation. Including rule lines here would be unnecessary and create undue visual clutter.
screen shot 2017-07-28 at 5 39 54 pm

I have noticed when using the 25/75 pattern (and most of our patterns) that guiding our stakeholders as to word counts can be very beneficial. You can see how word counts helped to create equal units here (so each unit has the same level of importance on the page): 
screen shot 2017-07-28 at 5 32 53 pm

screen shot 2017-07-28 at 6 16 14 pm

### Numbered steps
screen shot 2017-07-28 at 6 11 25 pm

### Single 25/75
screen shot 2017-07-28 at 6 21 17 pm

### Landing page with variety of sections
This example works without the rule because the two blocks both have 4 lines of supporting copy. In general these patterns work best when they have a similar or the same number of lines of text (within a group).   
screen shot 2017-07-28 at 6 22 43 pm
Here are the guidelines from the DM:
screen shot 2017-07-28 at 6 24 40 pm

## Instances where including a rule line makes sense:
### Things that are grouped together but which may benefit from some visual separation 
This example would benefit from word count guidance for stakeholders. The varying lengths of content create spacing that looks uneven and gives some of the units more visual impact/dominance than the others. Additionally the heading "What is a prepaid card" (H2) doesn't fulfill the parent relationship I would expect. If the 25/75's are all recommended steps or things to do then the H2 heading should ideally be written to reinforce the common thread of the set. This would also help users that skim content by reading headings. It is possible that with content parameters and a clearly unifying heading that the horizontal rule would be less necessary here. 
screen shot 2017-07-28 at 6 01 22 pm

This example would also benefit from word count guidance for stakeholders for the same reasons mentioned in the previous example. Additionally, this example doesn't have a unifying heading which makes it a little harder to digest from a content perspective. It leaves me asking, what is the unifying idea for all of these things (25/75's). If they are steps then the heading could have the word steps in it. 
screen shot 2017-07-28 at 6 01 36 pm

In looking through a variety of examples, I notice that there is a point where the content gets so long that the spacing between the units starts to look off. I haven't identified what this breaking point is but talking through it with @Scotchester it's when we exceed the minimum height. That is what causes the spacing to look like it does on this example (look at spacing after the third section before the rule line): 
screen shot 2017-07-28 at 6 01 36 pm

If you change the blocks to a max of 3 lines each, you get the ideal spacing (for using the rule line): 
screen shot 2017-07-28 at 6 30 32 pm

With that being said, once the content is wrangled the rule line isn't absolutely necessary: 
screen shot 2017-07-28 at 6 33 29 pm

nataliafitzgerald:
@Scotchester 
I will be out on Monday -Wednesday of next week but my official position is: 
- Button option should be removed from the 25/75, 50/50, 33/33/33 (we can always revisit if a use case happens to come up in the future)
- Horizontal rule lines should be **optional not mandatory** when using the 25/75's

sebworks:
@Scotchester, let's go with the drop down.

huetingj:
Awesome work on this module, @Scotchester! It looks so great. And thanks for pushing it through this quickly to help team Rara.

With the horizontal rules, I lean toward making them optional for the reasons @nataliaf posted regarding items' relatedness. Rara isn't currently stacking 25/75 modules on portal pages. But more broadly speaking, we're creating dual topic pages (i.e. Mortgages having two large sections— origination and servicing), and we're using a horizontal rule as a means to separate the two large sections on the page. So, I lean toward offering flexibility with rule use in modules with the thought that this might help teams down the road ensure the right points of separation/relation are in place for the whole of the page.

schaferjh:
Overall looking good. I made a strategic comment over in the Hub issue, addressing specific issues about implementation raised here:
- Links as buttons: this was included originally to give the modules flexibility. At the time there was an example on either Press Resources or About Us that needed it, though clearly they are not implemented now. I do not see the button option documented in the DM, so let's remove it.
- Please maintain the rule line as an option.
- Would the mincon drop down be a filter-as-you-type drop? If not, I don't support a drop down for 90 items.

Scotchester:
Pushed some updates that should address all of the remaining issues. @cfarm and @jimmynotjim please re-review.

I need to add a few more tests for the custom validator that I just added to ensure that 25/75s have images, but if you could give this a look tomorrow morning, @chosak, I'd appreciate it!

cfgov/jinja2/v1/_includes/organisms/info-unit-group-2.html
cfarm: +1 this would be a nice change to see.

Scotchester: Did I forget to do this?

Scotchester: Yes I did. Fix coming soon!

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
Scotchester: Done in 2961b13!

cfgov/jinja2/v1/_includes/organisms/info-unit-group-2.html
Scotchester: Fixed in 2961b13.

cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
jimmynotjim: 👍

Scotchester:
Wrapped up the test stuff, and I think this is good to go, pending Travis's final check!

Scotchester:
:tada:

nataliafitzgerald:
Nice work @Scotchester!

PR #1021, opened 2015-09-24
Updated the dep dir page title to include "Acting"

jimmynotjim:
Duplication of https://github.com/cfpb/cfgov-refresh/pull/997 to get it onto beta before the freeze

mr-freeze-evil-plan

PR #1020, opened 2015-09-24
Removed bad tests

jimmynotjim:
Removed bad tests that weren't covering much anyway

## Removals

- Two fragile tests of CF Notifier that weren't well written and will eventually move to a CF module anyway.

## Testing

- run `gulp test:unit:scripts`

## Review

- @KimberlyMunoz 
- @sebworks 
- @anselmbradford 

[Preview this PR without the whitespace changes](?w=0)

## Notes

- They were falsely passing, so they shouldn't have really even been in there. This was the very first set of tests I attempted to write and they were not good.

PR #3182, opened 2017-07-27
Ask CFPB: Fix Answer page previews for admin

higs4281:
Answer pages previewed through the Wagtail admin are not getting routed 
through the wagtail-sharing framework because Answer page requests are 
intercepted by a self-healing-url view outside of Wagtail. This view will now check 
the request to see if it is coming from a sharing site, and if so render a preview.

It turns out this problem also affected page unpublishing, so we can fix that at the same time.

This addresses GHE issues `#1020` and `#1021`

## Additions
- Tests for page.live status and for wagtail-sharing requests.

## Changes
- ask_cfpb's answer_view

## Testing

1. Make sure you have a sharing site set up locally. Then try previewing an Ask CFPB answer page that has a state of `LIVE + DRAFT`
If you have a recent production dump, page en-1969 (How can I get a refund ...) should be in that state near the top of the [admin list view](http://localhost:8000/admin/pages/5702/) for Ask CFPB.

2. Hover over the page listing, click on `MORE` on the right and then select `View sharing link`
You should see the same version of the page that you see when choosing `VIEW DRAFT`, and it should be different from the version you see when choosing `VIEW LIVE`

3. Pick a live answer page and un-publish it. The site page should raise Http404.


hillaryj:
Tested locally and this PR appears to solve the unpublish page issue for Ask CFPB questions!

PR #3189, opened 2017-07-31
Prototype mortgage data loading and API

higs4281:
Start an internal mortgage performance data API for cfgov's `data_research` app.

The goal is to deliver visualizations of the National Mortgage Database of 
county-level mortgage performance data, extrapolated to metro areas and states.

The original data will be updated quarterly, and this PR includes some 
initial work for a data pipeline to automate updates.

## A note on the FIPS object
You may notice that we're using a FIPS object as a metadata store that is loaded into memory on the fly for many operations, with `load_fips_meta()`. This would not be an efficient process in production but has been useful for data exploration, finding data problems and edge cases and getting the prototype running.

Once we get the API hooked up to real visualizations, we'll create tables for FIPS codes, counties and states to take over metadata handling. 

## Testing

If you want to try out the API, you'll need to load the initial dataset.

Let me know if you'd like this file.

Assuming you've pulled the file into your Downloads folder, here's how to load it up:

```
gunzip -c ~/Downloads/mortgagedata.sql.gz | mysql -uroot v1
./runserver.sh
```

Once loaded, You can make requests to the API for json files for rendering charts and maps based on FIPS codes and dates. 

#### Timeseries exmaples: counties, metro areas and states
- http://localhost:8000/data-research/mortgages/api/v1/time-series/12081/
- http://localhost:8000/data-research/mortgages/api/v1/time-series/35840/
- http://localhost:8000/data-research/mortgages/api/v1/time-series/27/

#### Map data: counties, metro areas and states
- http://localhost:8000/data-research/mortgages/api/v1/map-data/counties/2009-01/
- http://localhost:8000/data-research/mortgages/api/v1/map-data/metros/2009-01/
- http://localhost:8000/data-research/mortgages/api/v1/map-data/states/2009-01/


contolini:
For context: With the line chart, a location is selected and we need data for that location over a ten year span

0d416c20-6626-11e7-8c42-0756831d2216

With the map, we need data for all locations but only for the selected month (users can select a single location but that just zooms to that location):

00030ba4-6d42-11e7-95cf-1ee866c8eb47

cfgov/data_research/urls.py
```diff
@@ -0,0 +1,22 @@
...
+        name='data_research_api_mortgage_timeseries_national'),
+    url(r'^map-data/(?P<geo>counties)/(?P<year_month>\d{4}-\d{2})/?$',
+        MapData.as_view(),
+        name='data_research_api_mortgage_county_mapdata'),
+    url(r'^map-data/(?P<geo>metros)/(?P<year_month>\d{4}-\d{2})/?$',
+        MapData.as_view(),
+        name='data_research_api_mortgage_metro_mapdata'),
+    url(r'^map-data/(?P<geo>states)/(?P<year_month>\d{4}-\d{2})/?$',
+        MapData.as_view(),
+        name='data_research_api_mortgage_state_mapdata'),
```
chosak: I'm not sure what you're trying to do here by repeating the same pattern multiple times. I don't think you can define multiple URL patterns with the same `name`, when resolving I believe Django will use the last one defined. And all of these patterns are the same with the exception of the name of the capture - could you just use `(?P<geo>counties|metros|states)` as a single pattern?

cfgov/data_research/models.py
```diff
@@ -11,3 +15,162 @@ class ConferenceRegistration(models.Model):
...
+        help_text=("A comma-separated list of state abbreviations "
+                   "touched by FIPS for included counties."))
+
+    def save(self, **kwargs):
+        load_fips_meta()
+        self.aggregate_data()
+        if self.counties:
+            self.states = ", ".join(
+                sorted(set(
+                    [FIPS.county_fips.get(fips.strip())['state']
```
higs4281: More than minor. Thanks for the catch.

cfgov/data_research/models.py
```diff
@@ -11,3 +15,162 @@ class ConferenceRegistration(models.Model):
...
+
+    def aggregate_data(self):
+        if not self.counties:
+            return
+        count_fields = {
+            'total': 0, 'current': 0, 'thirty': 0,
+            'sixty': 0, 'ninety': 0, 'other': 0}
+        county_fips = [fips.strip() for fips in self.counties.split(',')]
+        county_records = CountyMortgageData.objects.filter(
+            fips__in=county_fips, date=self.date)
```
higs4281: Good eye. But County FIPS are always 5 digits. 
State FIPS are two digits, so they can't be in a state FIPS.

cfgov/data_research/models.py
```diff
@@ -11,3 +15,162 @@ class ConferenceRegistration(models.Model):
...
+    """
+    A model to store aggregate national mortgage performance for a given date,
+    updated quarterly.
+    """
+
+    def save(self, **kwargs):
+        self.aggregate_data()
+        super(NationalMortgageData, self).save(**kwargs)
+
+    def aggregate_data(self):
```
higs4281: I think I'd like to do this optimization after the self.aggregate_data methods are less fluid. The sequence of when self_aggregate_data is called might be changing.

cfgov/data_research/mortgage_utilities/fips_meta.py
```diff
@@ -0,0 +1,257 @@
...
+        FIPS.state_fips = json.loads(f.read())
+
+
+def load_thresholds():
+    """Get data thresholds from database, or fall back to starting defaults."""
+    from data_research.models import MortgageDataConstant
+    threshold_year = 2016
+    threshold_count = 1000
+    starting_year = 2008
+    if MortgageDataConstant.objects.filter(name='threshold_year').exists():
```
higs4281: Much better.

cfgov/data_research/mortgage_utilities/s3_utils.py
```diff
@@ -0,0 +1,51 @@
...
+S3_KEY = os.getenv('AWS_S3_ACCESS_KEY_ID')
+S3_SECRET = os.getenv('AWS_S3_SECRET_ACCESS_KEY')
+BASE_BUCKET = 'files.consumerfinance.gov'
+MORTGAGE_SUB_BUCKET = "data/mortgage-performance"
+# s3_base is "http://files.consumerfinance.gov"
+# files live at
+# http://s3.amazonaws.com/files.consumerfinance.gov/data/mortgage-performance/
+
+
+def read_in_s3_csv(url):
```
higs4281: I will pursue this.

cfgov/data_research/tests/management/commands/test_conference_export.py
```diff
@@ -1,8 +1,13 @@
+from __future__ import unicode_literals
```
higs4281: That wasn't my goal.

test_conference_export.py tests were not running because its folder did not have an `__init__.py` file.

When I added the __init__ file, there was a no-op test and a bit of missing coverage, so I added the unicode_literal import for the new tests. I could convert them to python2 style if that's less invasive.

cfgov/data_research/urls.py
higs4281: The names aren't the same: the geography differs for each.

Maps are the least formulated part of the project, and I wanted to leave map endpoints explicit until we know more about what the options will need to be.

That said, if you think this is just a bad practice, I can refactor the geo keyword to be a regex.

chosak: Oh, you're right - I'm sorry for misreading. It's fine to leave for now, but I'm not certain that multiple URL definitions serve any purpose if the regexes are the same, because incoming URL requests will only ever match one.

cfgov/cfgov/urls.py
```diff
@@ -303,6 +303,10 @@
     # Form csrf token provider for JS form submission
     url(r'^token-provider/', token_provider),
 
+    # data-research-api
+    url(r'^data-research/mortgages/api/v1/',
+        include_if_app_enabled('data_research', 'data_research.urls')),
```
higs4281: A great question. We have had a number of APIs that live in external repos, so I was following the pattern of the API being integrated into the app's URL path. But it might make sense to corral all cfgov APIs under one API path separate from the Wagtail tree.

I think a decision about what that namespace would be is in order before decoupling this API from the data_research app. This would affect CCDB5 as well, I would assume.

cfgov/data_research/tests/management/commands/test_conference_export.py
chosak: No, that makes sense. Thanks for the explanation.

higs4281: Also happy to leave conference code untouched by this PR if that would keep concerns separated.

chosak: If you don't feel strongly, that might be best, if only so that if someone checks the commit log on those files they don't see this PR.

higs4281: detangled.

cfgov/data_research/urls.py
higs4281: I'd collapse to one URL with a regex once I'm more sure of the API needs.

richaagarwal:
Does a function like `load_fips_meta()` get called each time a visualization is rendered? It seems like there's a lot of data loading/prep that happens from csv files -- would it make sense to save these to a database?

higs4281:
> there's a lot of data loading/prep

@richaagarwal There is indeed a lot of metadata prep going on. Our base data set is sparse and needs lots of extending and aggregating. This is the first iteration of the API, and once we clarify what data points we need or don't need, we'll focus on optimizing.

marteki:
@chosak and @richaagarwal and others - what else is needed in order to get approval for this PR?

PR #3192, opened 2017-08-01
Cleanup related metadata

anselmbradford:
While looking at post preview I found some unused markup in related metadata.

## Removals

- `u-mb0` classes that appear to not be needed.
- text block paragraph that was not used.

## Changes

- Changes `half_width` boolean to `is_half_width`.

## Testing

1. Create a test document detail page in wagtail.
2. Add a related metadata to the sidebar.
3. Add the block types date, list, text, and topics.
4. Check the "is half width" option to check that that doesn't break.
5. The page appearance should be unchanged.

## Screenshots

Appearance should be unchanged:
screen shot 2017-08-01 at 5 15 19 pm


cfgov/jinja2/v1/_includes/molecules/related-metadata.html
```diff
@@ -135,13 +140,9 @@ <h3 class="h4">
    ========================================================================== #}
 
 {% macro _text(text) %}
-
-<p class="u-mb0">
```
anselmbradford: The wagtail RTE adds a paragraph, so looking at the generated markup it was looking like:
```
<p class="u-mb0"></p>
<p>Some wagtail entered text</p>
```

This could be changed to a div, but I didn't see how the `u-mb0` was used as removing it made no apparent visual difference.

cfgov/jinja2/v1/_includes/molecules/related-metadata.html
```diff
@@ -135,13 +140,9 @@ <h3 class="h4">
...
     {{ parse_links(text.blob) | safe }}
-</p>
 {% endmacro %}
 
-{# TODO: This function is not used and can be removed. #}
```
anselmbradford: This actual is used. I added a comment about how this is called through the type_lookup array.

anselmbradford: While `_topics` is called, I'm unclear how `{{ _list( related_metadata_tags(page), true ) }}` within is called.

cfgov/jinja2/v1/_includes/molecules/related-metadata.html
```diff
@@ -10,38 +10,43 @@
...
    value:                              Object defined from a StreamField block.
 
-   value.half_width:                   A Boolean indicating whether the
+   value.is_half_width:                A boolean indicating whether the
                                        list items should be at half width.
                                        Defaults to False.
 
-   value.slug:                         TODO: Add data type and description.
+   value.slug:                         A string for the header on all
+                                       related metadata items.
```
Scotchester: Wording here is a little unclear. How about something like "A string of text to use as a header for the entire module"?

cfgov/jinja2/v1/_includes/molecules/related-metadata.html
```diff
@@ -10,38 +10,43 @@
...
 
    value.content:
 
    value.content[i].block_type:        String for a block type,
                                        it should be date, list, text,
                                        or topics.
 
-   value.content[i].value:             TODO: Add data type and description.
+   value.content[i].value:             A date, list. text, or topics markup.
```
Scotchester: Typo: period should be a comma in the list here.

cfgov/jinja2/v1/_includes/molecules/related-metadata.html
```diff
@@ -10,38 +10,43 @@
...
 
-   value.content[i].value.show_topics: Boolean for whether to
+   value.content[i].value.show_topics: A boolean for whether to
```
Scotchester: "to" ➡️ "the"

cfgov/jinja2/v1/_includes/molecules/related-metadata.html
```diff
@@ -10,38 +10,43 @@
...
-   value.half_width:                   A Boolean indicating whether the
+   value.is_half_width:                A boolean indicating whether the
                                        list items should be at half width.
                                        Defaults to False.
 
-   value.slug:                         TODO: Add data type and description.
+   value.slug:                         A string for the header on all
+                                       related metadata items.
 
    value.content:
```
Scotchester: Add a description of this, too? "A list of metadata blocks" or something ought to do it.

cfgov/jinja2/v1/_includes/molecules/related-metadata.html
Scotchester: That seems strange. I'm not sure why the text entered in wagtail wouldn't be inside the `p.u-mb0`. Did you look at the actual source, or just look at it in the Inspector? If the latter, I wonder if the browser interprets an invalid nested paragraph as two consecutive paragraphs. Regardless, the change makes sense if no one has cared about the bottom margin not being zeroed-out to this point.

cfgov/jinja2/v1/_includes/molecules/related-metadata.html
Scotchester: I think `page` is always available to templates. Is that what you mean? Some `page` data must be needed by the `related_metadata_tags` function, which must not be part of the template structure, since it's not imported here.

cfgov/jinja2/v1/_includes/molecules/related-metadata.html
anselmbradford: Oh good point. I had just looked at the inspector.

cfgov/jinja2/v1/_includes/molecules/related-metadata.html
anselmbradford: I just meant the `_topics()` function IS called, so the TODO comment was not accurate. However, it's currently not generating any markup when used in Wagtail, so I guess it's functionally not used.

anselmbradford:
@Scotchester Completed your requested code comment updates.

PR #3203, opened 2017-08-03
Fixes spacing on post preview

anselmbradford:
Fixes https://[GHE]/CFGOV/platform/issues/1016

## Changes

- Adds bottom margin to `o-post-preview_description` (note: this could be removed entirely if the content coming in is guaranteed to be enclosed in a paragraph, but having a container in the template makes it a bit easier to see the structure).
- Removes `m-meta-header_left` class from category, which was floating the category left. This matches the update in the activity log that went in with the CFv4 update.
- Minor spacing updates.
- Changed filterable-list-controls.html `from` command to `import` to match style of other imports in the file.

## Testing

1. `./frontend.sh`
2. visit /about-us/blog/ and compare to the live blog.

## Screenshots

See category at top-left and spacing above/below the author by-line.
screen shot 2017-08-03 at 11 01 36 am


anselmbradford:
Updated this to match what is in https://github.com/cfpb/cfgov-refresh/pull/3205 so that the tags and the byline can be spaced correctly. When the byline was in a paragraph the tags were pushed outside of the paragraph making it so they couldn't be targeted by scoping within `o-post-preview_meta`.

sebworks:
@anselmbradford, how does this add margin bottom to `o-post-preview_description`?

sebworks:
sorry, I hit the wrong button.

sebworks:
looks a little weird on mobile but not sure if you want to break to a newline : 
screen shot 2017-08-04 at 11 46 45 am

nataliafitzgerald:
@sebworks
- Good catch! We'd definitely like the title to break to a new line on mobile. 

@anselmbradford 
- What is the space before and after the pipe character
- What is the space after the title before the rule line? 
- What is the space after the rule line before the top of the image? Can we include an image in the screenshot?
- What is the space after the rule line before the top of the text?

cfarm:
To answer your questions

1. Looks like it is a space character, not added with CSS
2. above rule: same as on live site - looks like 23px in the inspector
3. @nataliafitzgerald it's the same spacing with an image here as what's on the live site, 10px in inspector. here's a screenshot of this code locally:

screen shot 2017-08-04 at 12 51 28 pm

4. after rule above picture/text: 10px in the inspector (same as live site)

cfarm:
For the mobile text line break, @nataliafitzgerald can you create a new task for this on GHE? It can then be addressed in a separate PR, so we can merge this one sooner.

nataliafitzgerald:
@cfarm - Ok, will do.

nataliafitzgerald:
@cfarm - I created an issue here: https://github.cfpb.gov/CFGOV/platform/issues/1047

PR #3206, opened 2017-08-03
Update jQuery load call

anselmbradford:
Fixes one of the jquery deprecated load references in https://[GHE]/CFGOV/platform/issues/1006.

## Changes

- Changes `load(` to `on('load'`

## Testing

1. Put a console log above the line 196 in the changed file. 
2. Log into Wagtail and add a table-block.
3. open the dev console and hard reload the page and see that your console message logs.


sebworks:
I'll admin, I don't understand why this is a thing but 👍.

PR #3209, opened 2017-08-04
WIP: adjust padding for global search input to work with cf-forms update

cfarm:
WIP: not mergeable until [this cf-forms PR](https://github.com/cfpb/capital-framework/pull/553) gets published and the upgrade added here.

## Additions

- adds class + styling to add additional padding to the global search text input to match current design with the new decreased text input padding.

## Changes
- text input border color is now gray-60 in latest hub standards

## Testing

1. `./frontend.sh`
1. test that global search text input height matches the search button

## Notes

-

## Todos

-

## Checklist


cfgov/unprocessed/css/molecules/global-search.less
```diff
@@ -102,6 +102,11 @@
         display: block;
     }
 
+    &_input {
+        // Resets for default a-text-input styles.
```
anselmbradford: How many styles do we have that need to come in via `a-text-input`? If we have to reset styles that may be an indication that it is a different pattern and shouldn't use a-text-input at all.

cfarm: this element inherits the following properties from `a-text-input`:
- appearance
- display
- border
- background
- focus state

the only change is to the padding, to make the height match the search button, conforming to the global search pattern design spec.

PR #3212, opened 2017-08-04
Update `fax` to `is_fax` boolean label

anselmbradford:
## Changes

- Changes `fax` boolean label to `is_fax`, per TODO comment.

## Testing

1. Run `python cfgov/manage.py migrate`.
2. Create a browse page in Wagtail.
3. Add a expandable.
4. Add a Phone block to the expandable.
5. Check that the `Is this number a fax?` checkbox works.

## Screenshots

screen shot 2017-08-04 at 4 38 02 pm

## Notes

- Seems like is_fax should be attached to EACH number, not ALL numbers!
Opened https://github.com/cfpb/cfgov-refresh/issues/3213


anselmbradford:
Examples of migrating blocks (via @chosak):
https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/migrations/0022_replace_dup_category_field_data.py
https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/migrations/0054_new_categories.py
https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/migrations/0011_hero_refactor_data.py

Scotchester:
@chosak The standard migration doesn't handle this? I'm confused.

chosak:
@Scotchester no, because you're not renaming a Django model field, you're renaming a block in a Wagtail StreamField. Recall that StreamField content is stored as JSON in the database, e.g. there's a database column named for the field name (like `content`) that contains a JSON representation of all of the blocks.

Changing the _field_ name would result in a Django migration that changed the DB column name. But just changing the _block_ name doesn't trigger any kind of automatic migration of the DB _content_ that already exists. The links @anselmbradford [posted above](https://github.com/cfpb/cfgov-refresh/pull/3212#issuecomment-320358146) are some examples of manual migrations we've done in the past to address this. See also [this Wagtail issue](https://github.com/wagtail/wagtail/issues/2110) on the subject.

Scotchester:
Ahhh, that makes sense, now. Thanks for the explanation, @chosak.

willbarton:
There's a section of the documentation on migrating stream fields: https://cfpb.github.io/cfgov-refresh/wagtail-migrations/

Scotchester:
@anselmbradford In light of the need to write a data migration for this, is it worth doing anymore?

anselmbradford:
I'm rotating to maintenance & response after next sprint, so thinking I'll re-visit this then. If it gets swamped by other priorities I'll close it.

anselmbradford:
😭

Scotchester:
Did you determine it's not worth the effort to write the migration?

anselmbradford:
> Did you determine it's not worth the effort to write the migration?

Not so much, I was just keeping my word from my Dec 15 comment. https://github.com/cfpb/cfgov-refresh/issues/3213 seems like actually bug-like and semantics here could be incorporated into that, hopefully.

Scotchester:
I guess I just don't see the point of closing it if it's something that we still think is worth doing. But I take your point that it could be rolled into #3213.

PR #3222, opened 2017-08-08
Widen gulp's search path when compiling the chart organism

contolini:
The file structure of our internal chart library will be undergoing some changes in preparation for the release of our Mortgage Performance charts. Instead of having gulp look for the library's Less file in a specific directory, have it search all of /node_modules/cfpb-chart-builder/ because we're rearranging some things in cfpb-chart-builder.

The chart library is imported [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/css/organisms/chart.less#L24).

## Changes

- Gulp's search path for the chart library.

## Testing

1. `gulp styles` should not return any errors and http://127.0.0.1:8000/data-research/consumer-credit-trends/credit-cards/borrower-risk-profiles/ should show a bunch of charts.

## Notes

- See https://github.com/cfpb/cfpb-chart-builder/pull/66


gulp/config.js
```diff
@@ -65,7 +65,7 @@ module.exports = {
         paths.modules + '/cf-pagination/src',
         paths.modules + '/cf-tables/src',
         paths.modules + '/cf-typography/src',
-        paths.modules + '/cfpb-chart-builder/src/static/css',
+        paths.modules + '/cfpb-chart-builder/src/**',
```
Scotchester: Late to the party, but given the format of the paths above, do you need the `/**` here?

contolini: Yeah, the old version of the chart library had its Less in `/cfpb-chart-builder/src/static/css`. The new version has it in `/cfpb-chart-builder/src/static/`. To make cfgov-refresh future-proof I added the `**` which will search all subdirectories inside of `src` because it's possible we'll change the Less' location down the road.

The capital framework component paths above *only* search in `src` and not `src`'s subdirectories.

Scotchester: Ah, gotcha. Thanks!

PR #1945, opened 2016-04-20
New blog newsroom

kurtw:
Introduces the new Blog and Newsroom page types

## Additions

- BlogPage and NewsroomPage 
- Article Subheader

## Changes

- Changes some templates around
- Fixes some bugs
- Learn, DocDetail, Blog, Newsroom pages have an option to add an Article Subheader for blog introductory paragraphs
- Full Width Text now has a Media option for images

## Testing

- `cfgov/manage.py migrate` (make sure you have the local.py database settings configured, if not see Notes)
- Go under the blog page and create a new blog post!
 - Add Article Subheader to the Header streamfield
 - Add the media option in the Content streamfield
 - Add some sidefoot content
- Preview it, Save it, Publish, Check it out on the site!
- Repeat for NewsroomPage

## Review

- @kave 
- @jimmynotjim 
- @richaagarwal 
- @rosskarchner 

## Screenshots
screen shot 2016-04-20 at 5 28 24 pm

## Notes

- local.py settings for database
```
DATABASES = {
    'default': {
        'ENGINE': MYSQL_ENGINE,
        'NAME': os.environ.get('MYSQL_NAME'),
        'USER': os.environ.get('MYSQL_USER'),
        'PASSWORD': os.environ.get('MYSQL_PW', ''),
        'HOST': os.environ.get('MYSQL_HOST', ''),  # empty string == localhost
        'PORT': os.environ.get('MYSQL_PORT', ''),  # empty string == default
        'OPTIONS': {
               "init_command": "SET storage_engine=MYISAM",
        }   
    },
}
```

PR #3226, opened 2017-08-09
Removes unused templates

anselmbradford:
I couldn't find a reference to these two templates. Appears they are unused?

## Removals

- `organisms/blog_introduction.html`
- `organisms/layout-two-columns.html`

## Testing

1. Search for use of `blog_introduction.html` and `layout-two-columns.html` and verify they are unused.


sebworks:
@anselmbradford, this was the PR: https://github.com/cfpb/cfgov-refresh/pull/1945. I would go through it.

anselmbradford:
Looks like it may have been accidentally added as a duplicate of item-introduction.html

PR #3229, opened 2017-08-09
Handle two bylines

anselmbradford:
As noted in review, fix needed for multiple bylines so they appear on the same line.

## Removals

- `'m-meta-header_left'` removed from category slug that still had it.
- `o-post-preview_meta` removed and replaced with `o-post-preview_byline-group`.

## Changes

- Updated post-preview code comments.
- Makes post_preview import statements consistent across the codebase.
- Moves `location_image_url` to within conditional statement so it isn't always imported.
- 

## Testing

1. ./frontend.sh
2. Visit /blog and /newsroom

## Screenshots

Before:
screen shot 2017-08-09 at 2 06 57 pm

After:
screen shot 2017-08-09 at 2 09 06 pm

## Notes

- `'m-meta-header_left'` should probably come back after refactoring meta header to use (likely) flexbox so things can be stacked at mobile sizes (currently they publication date and category are on the same line at mobile).


cfgov/jinja2/v1/_includes/organisms/post-preview.html
```diff
@@ -9,39 +9,58 @@
...
-
-   settings.comments_close_date: A datetime object marking the deadline of the comment period.
-   settings.preview_link_text:   A string with the description text of an external link.
-   settings.preview_link_url:    A string with the url of an external link.
-
-   path:                        A string with the path for the homepage of the pagetype
-                                `/about-us/blog/`, `/about-us/newsroom/`
+   post:                          An object with specific article information.
+
+   post.specific_class.__name__:  String used to check for EventPage post type.
```
sebworks: We shouldn't be doing things like this in the template. Just add a method to the class python class.

anselmbradford: Yeah the arguments coming into this are really hard to follow!

PR #3216, opened 2017-08-04
Update optional form input style

anselmbradford:
Required/optional fields were inconsistently styled. See https://[GHE]/CFGOV/platform/issues/465. This makes them consistent.

## Changes

- Remove asterisk for required fields on conference registration form.
- Makes optional fields have `(optional)` text next to them.

## Testing

1. Create a new browse page in Wagtail.
2. Add the conference registration form.
3. Preview the page and view the optional/required fields.
4. Change https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/conference-registration-form.html#L99 to false to see checkbox optional style.

## Screenshots

Before:

screen shot 2017-08-04 at 5 54 00 pm

After:
screen shot 2017-08-04 at 5 53 17 pm

## Notes

- Checkbox background is messed up, but that was that way outside of this PR.

PR #3235, opened 2017-08-10
Lowercase optional text in feedback component

anselmbradford:
Addendum to https://github.com/cfpb/cfgov-refresh/pull/3216

## Changes

- Converts `(Optional)` to `(optional)` in the feedback form.
- Adds `a-label_helper` class, which should be added to CF to set color of optional text. Currently this doesn't do anything.


anselmbradford:
@jimmynotjim One update: changed span to small:
screen shot 2017-08-10 at 3 48 19 pm

PR #3236, opened 2017-08-10
data_research: Adjust mortgage loading scripts for pipeline use

higs4281:
Our data pipeline, which would run once a quarter, will need to read a file from S3, update our list of sampling dates and rebuild our API's dataset. This PR puts those steps together.

## Testing

You can run the pipeline's update step with these commands:
```
python cfgov/manage.py loaddata mortgage_constants
python cfgov/manage.py runscript load_mortgage_performance_csv --script-args delinquency_county_0317.csv

```

It should take a few minutes, and you'll see object counts go by as it builds 740K+ objects and updates our list of sampling dates.

The script would be run once a quarter.


contolini:
This is outside the scope of this PR but I'm curious if you've thought about how this script gets triggered. Will we have a jenkins job monitoring `data/mortgage-performance/source/` for new files?

higs4281:
> Will we have a jenkins job ...


Yes -- the idea is that we'd monitor for a new source file, then this and other scripts would run to update and deploy the data.

contolini:
👍 

screen shot 2017-08-10 at 5 19 32 pm

PR #3240, opened 2017-08-11
Fixes #877 - Updates date format

### LINKED ISSUE #877: Strip leading zeros on dates?
One place in the [code](https://github.com/cfpb/cfgov-refresh/pull/871/files#diff-9dab62ff32679f541743e255c7d93881L48) stripped the leading zero from the date. We don't do this anymore with https://github.com/cfpb/cfgov-refresh/pull/871, but should we? Should the dates be `NOV 02, 2015` or `NOV 2, 2015`?

anselmbradford:
Fixes https://github.com/cfpb/cfgov-refresh/issues/877

## Changes

- Changes date format to include period and remove leading zeros.

## Testing

1. The following pages should have a date in the format of MMM. DD, YYYY (without leading zero):
 - /policy-compliance/notice-opportunities-comment/open-notices/
 - /blog/
 - /activity-log/
 - /about-us/events/
 - /about-us/careers/

## Screenshots

screen shot 2017-08-11 at 11 43 22 am

screen shot 2017-08-11 at 11 42 44 am

screen shot 2017-08-11 at 11 42 50 am

screen shot 2017-08-11 at 11 43 07 am

screen shot 2017-08-11 at 11 42 57 am


anselmbradford:
Added update to convert jobs table date:
screen shot 2017-08-11 at 11 58 43 am

sarahsimpson09:
@anselmbradford This looks good, but I have one question--what about May? Will the period be removed or will it look like "May. 9, 2017"?

anselmbradford:
Pushed the `May` and `Sept.` fixes:

screen shot 2017-08-11 at 1 43 04 pm

screen shot 2017-08-11 at 1 42 51 pm

anselmbradford:
Fixed the tests.

cfgov/jinja2/v1/_includes/macros/time.html
```diff
@@ -41,3 +41,20 @@
     {% endif %}
 </span>
 {% endmacro %}
+
+{% macro render_date_no_markup(datetime,
```
sebworks: ehhhh, I'm not crazy about it.

anselmbradford: It's awesome! Best code I've ever written! Look at all the brackets!

richaagarwal:
Why is this necessary? :)

anselmbradford:
@richaagarwal See the "dates and date ranges" section in the content strategy doc linked via https://[GHE]/UX/Content-Strategy 
I posted a comment there about clarifying the comment on the blog dates being perceived technical limitations imposed by wagtail.

cfgov/jinja2/v1/_includes/macros/time.html
```diff
@@ -41,3 +41,20 @@
...
+    {% set format_month = '{dt:%b}'.format(dt = datetime) %}
+    {% if format_month == 'May' %}
+        {# May doesn't have a period. #}
+        {{ '{dt:%b} {dt.day}, {dt.year}'.format(dt = datetime) }}
+    {% elif format_month == 'Sep' %}
+        {# September is four letters. #}
+        {{ 'Sept. {dt.day}, {dt.year}'.format(dt = datetime) }}
+    {% else %}
+        {{ '{dt:%b}. {dt.day}, {dt.year}'.format(dt = datetime) }}
+    {% endif %}
```
willbarton: I spent some time thinking about this and looking at possible alternatives. I couldn't find a Python library that has built-in support for formatting months like this (conditionality around when a month is abbreviated and to what character-number).  

At the very least, I don't think this should be in the template itself, it should be a filter or something similar that can be called from any template (particularly if this is the standardized way to refer to dates) and can be properly unittested. 

It would also simplify the logic greatly to [use a mapping like Python's build-in datetime library of month number to abbreviation](https://github.com/python/cpython/blob/master/Lib/datetime.py#L144-L146) instead of ifs/elses.

willbarton: Thinking about it further, a custom wrapper around strftime that either introduces our own format code or handles `%b` differently from the standard library would mean we could actually use date formatting here (instead of dt.month, dt.day, dt.year) and in all the other places in this changeset like we should be doing.

richaagarwal:
@anselmbradford Thanks for the background!  That's interesting.  I didn't see anyone reply to your comment -- was consensus reached elsewhere?  I don't think we should be making these changes unless someone really wants them.  If we do proceed, I agree with @willbarton that a wrapper around `strftime` would make the most sense.

cfgov/jinja2/v1/_includes/macros/time.html
anselmbradford: This all sounds well and good and over my head 😄

anselmbradford:
Closing this astoundingly stale PR, but leaving the associated issue open and updating the issue with context that a wrapper around `strftime` would be likely needed.

PR #3256, opened 2017-08-18
Add script for posting SQS messages as issues to Github

richaagarwal:
- Part of the work @willbarton and I are doing around alerts.  This is a very dumb first iteration of posting the SQS message body as an issue to Github, where the message body would be some type of Jenkins alert. This script will be called from Jenkins. 
- Adds two dependencies - boto3 for interfacing with the API for Amazon SQS, and github3 for interfacing with the Github API
- Example usage 
(note - the below will not work as is, you must replace the strings with the right credentials. If you'd like to test it I can provide the credentials):
`python cfgov/alerts/post_sqs_messages_to_github.py --queue_url='https://sqs.queue-url' --aws_access_key_id='AWS access key' --aws_secret_access_key='AWS secret key' --github_token='Github Token' --github_url='Github Enterprise URL' --github_user='Username' --github_repo='Repo name'`


willbarton:
If this is a management command instead of just a script, we're going to have to setup a virtualenv with all the cfgov-refresh requirements and everything in the Jenkins workspace that'll run the command. 

We also don't need boto and github3.py installed on the production servers, just the Jenkins server, which is what'll happen also with those in the base requirements. 

Can we make it just a script?

cfgov/v1/management/commands/post_sqs_messages_to_github.py
```diff
@@ -0,0 +1,76 @@
...
+        # Receive messages from specified SQS queue
+        response = client.receive_message(
+            QueueUrl=options['queue_url'],
+            MaxNumberOfMessages=10,
+        )
+
+        # Post these messages to specified Github repo
+        for message in response['Messages']:
+            body = message['Body']
+            repo.create_issue(title=body[:20], body=body)
```
willbarton: I love how simple this has ended up being. I wonder if we should rudimentary provide support for customizable issue templates around the message body with format strings or something like that.

requirements/base.txt
```diff
@@ -24,6 +25,7 @@ edgegrid-python==1.0.10
 elasticsearch==2.4.1
 feedparser==5.2.1
 geocoder==1.12.0
+github3.py==0.9.6
```
willbarton: Thinking about this a little bit more and looking at our http-smoke-tests, we currently manually run this in that Jenkins job: 

```shell
pip install requests==2.7.0 beautifulsoup4==4.5.1
```

How about instead of these in base.txt, we add requests, beautifulsoup4, boto3, and github3.py to a `requirements/automations.txt` file that we then use in Jenkins jobs that use scripts in this repo?

cfgov/v1/management/commands/post_sqs_messages_to_github.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+
+        # Receive messages from specified SQS queue
+        response = client.receive_message(
+            QueueUrl=options['queue_url'],
+            MaxNumberOfMessages=10,
+        )
+
+        # Post these messages to specified Github repo
+        for message in response['Messages']:
```
chosak: Minor thing: If something goes wrong in this step (say a GH error of some kind), are the messages lost forever? Perhaps it'd be nice to log them to console so that there's at least a record of what was sent.

requirements/base.txt
richaagarwal: Great idea!

richaagarwal:
@willbarton Yes, I can make it into a script.  @chosak hmmmm, is v1 synonymous with wagtail? I'm happy to move it wherever, but maybe good to clarify directories.

chosak:
@richaagarwal it's sort of grown into a catchall but I would love to put pure Python/Django things in `core` and have things more specific to Wagtail/cf.gov in `v1`.

PR #3257, opened 2017-08-18
Stop crawlers from trying filter combinations

rosskarchner:
This came out of the investigation of crawler errors (GHE 1015). It doesn't really provide us with any benefit for Google, etc to crawl every filter combination. Every page is accessible by paginating through the primary set of results.

## Additions

- add the line `Disallow: *form-id=*` to robots.txt

## Checklist

### DIFF
--- a/cfgov/core/static/robots.txt
+++ b/cfgov/core/static/robots.txt
@@ -4,5 +4,6 @@ Disallow: /about-us/the-bureau/leadership-calendar/pdf
 Disallow: /es/obtener-respuestas/*.imprimir.html
 Disallow: /external-site
 Disallow: /*?success
+Disallow: *form-id=*
 
 sitemap: http://www.consumerfinance.gov/activity-log/feed/


chosak:
Unfortunately the robots.txt standard [doesn't seem to support](http://www.robotstxt.org/robotstxt.html) use of wildcards like this:

> Note also that globbing and regular expression are not supported in either the User-agent or Disallow lines. The '\*' in the User-agent field is a special value meaning "any robot". Specifically, you cannot have lines like "User-agent: *bot*", "Disallow: /tmp/*" or "Disallow: *.gif".

chosak:
Although then again [Wikipedia says](https://en.wikipedia.org/wiki/Robots_exclusion_standard#Universal_.22.2A.22_match):

> The Robot Exclusion Standard does not mention anything about the "*" character in the Disallow: statement. Some crawlers like Googlebot recognize strings containing "*", while MSNbot and Teoma interpret it in different ways. _[citation needed]_

So worth a try?

rosskarchner:
There are definitely some crawlers in the wild that don't support it-- but I think the big ones do?

rosskarchner:
According to https://developers.google.com/search/reference/robots_txt#allow

"Google, Bing, Yahoo, and Ask" support it

chosak:
@rosskarchner nice find! :+1:

PR #3260, opened 2017-08-18
data_research: Create elements for mortgage interactives

higs4281:
This creates a mortgage-chart block and page to satisfy requirements of displaying visualizations for the mortgage performance project on a browse page. 

The chart block created for consumer credit trends, which renders single pre-built charts 
with set parameters, cannot be reasonably extended to serve the needs of the
mortgage visualizations, which offer users many geographic and date choices
for making comparisons. 

This doesn't deliver a working chart block yet -- just the building blocks for rendering the initial two interactive browse pages. So the block html is not yet functional.

## Additions
- MortgagePerformancePage
- MortgageChartBlock organism and template.
- MortgageMetaData model for capturing and delivering metadata related to quarterly updates.
- Pipeline script to export CSV files and store metadata about them.
- API URLs and view for metadata

## Testing
The chart modules aren't ready for testing, but the first two metadata API endpoints below should be functional if you first load metadata with:

```
python cfgov/manage.py loaddata mortgage_metadata.json
```

- [sampling dates](http://localhost:8000/data-research/mortgages/api/v1/metadata/sampling_dates/)
- [metadata for download files](http://localhost:8000/data-research/mortgages/api/v1/metadata/download_files/)
- [county metadata by state](http://localhost:8000/data-research/mortgages/api/v1/metadata/state_county_meta/)
- [metro metadata by state](http://localhost:8000/data-research/mortgages/api/v1/metadata/state_msa_meta/)


marteki:
@rosskarchner @richaagarwal @chosak 

Scrum master question: do any of you have availability to review this? I think @higs4281 will keep adding enhancements and new work to it as long as it's open. 😆

chosak:
@marteki with a threat like that, I have no choice but to respond! I'll take a look.

cfgov/data_research/scripts/export_public_csvs.py
```diff
@@ -40,6 +43,43 @@
...
+    download_meta_file, cr = MortgageMetaData.objects.get_or_create(
+        name='download_files')
+    new_posting = {geo_type:
+                   {'slug': slug,
+                    'url': csv_url,
+                    'size': csv_size,
+                    'thru_date': thru_date,
+                    'pub_date': pub_date}}
+    if not download_meta_file.json_value:
+        download_meta_file.json_value = json.dumps(
```
chosak: See above re: possible use of JSONField. That would eliminate the need for this kind of `json.dumps`/`json.loads` here, and be easier to migrate in future to native JSON fields if we want to do that.

cfgov/v1/models/browse_page.py
```diff
@@ -72,11 +76,31 @@ class BrowsePage(CFGOVPage):
 
     objects = PageManager()
 
+    def get_mortgage_meta(self):
```
chosak: Can this function live on the `MortgageChartBlock` so as not to clutter up `BrowsePage`?

cfgov/v1/models/browse_page.py
```diff
@@ -72,11 +76,31 @@ class BrowsePage(CFGOVPage):
...
 
     def get_context(self, request, *args, **kwargs):
         context = super(BrowsePage, self).get_context(request, *args, **kwargs)
         context.update({'get_secondary_nav_items': get_secondary_nav_items})
+        if 'data-research/mortgage' in request.url:
+            context.update(self.get_mortgage_meta())
+        if '30-89' in request.url:
+            context.update({'delinquency': 'percent_30_60'})
+        elif '90' in request.url:
+            context.update({'delinquency': 'percent_90'})
```
chosak: Is there no way to stick this context on the `MortgageChartBlock` somehow? Otherwise this is going to get called and added for every single `BrowsePage` on the site. I know we currently overload the default Wagtail block context (our `render_stream_child`method), but it'd be nice to be able to encapsulate it there.

Is this context addition limited to only pages that have one of these blocks? Could you instead guard this with some kind of iteration over the page blocks to see if a block is indeed present?

cfgov/data_research/models.py
```diff
@@ -176,3 +179,20 @@ def __str__(self):
...
     class Meta:
         ordering = ['name']
+
+
+class MortgageMetaData(models.Model):
+    """
+    Metadata values, stored as json, to supplement display of mortgage charts.
+    """
+    name = models.CharField(max_length=255)
+    json_value = models.TextField(blank=True)
```
higs4281: I didn't know we could use django-jsonfields without a db that supports it. I'll explore!

cfgov/v1/models/browse_page.py
higs4281: I intended for this context to apply to browse pages whose request follows the `data-research/mortgage` URL path -- which will involve pages other than chart pages. If I scope the `30-89` and `90` lines to `data-research/mortgage', wouldn't that leave other browse pages alone?

higs4281: I was trying to avoid adding a new page type.

chosak: I understand the rationale but worry that doing this kind of URL-based logic is error-prone and goes a bit too far in violating the Wagtail model. This means that, for purposes of unit testing or local development of pages with these blocks, users/developers have to just somehow know the right hardcoded path to use. I know we do this thing elsewhere but it would be better if blocks/pages could somehow be self-describing/encapsulate their own behavior based on some consistent logic.

Is "no new page types" a hard and fast rule? Tagging @Scotchester as I know he has run into this kind of thing before, wanting custom behavior on a page based at least in part on its location in the tree. Is there some other distinguishing factor we could use, i.e. a page category?

higs4281: The goal is to provide context values to data-research/mortgage pages to keep charts and their footnotes, download file links and other parameters in sync with data updates. A MortgagePerformancePage would make sense to me, if there are no objections to adding a page type.

higs4281: If we do want to avoid a new page type, would it be any better to peg the custom context to the page's parent? That would be more reliable than a URL test and would not be dependent on an editor setting the right category when creating a new mortgage-performance page.

chosak: What do you mean by "peg the custom context to the page's parent"? How would it be pegged there?

higs4281: Only add context for children of `mortgage-performance-trends`  
https://github.com/cfpb/cfgov-refresh/blob/add-data-research-page/cfgov/v1/models/browse_page.py#L100-L105

chosak: That feels similarly convoluted to just checking the path of the page itself. I don't want to be a blocker on this (especially since this PR has grown so large) but I do think this is a trend we should avoid to keep things simple and straightforward.

Different page behavior based on location in the tree just seems like a hard pattern to maintain, especially if that logic is spread out throughout the code. Explicitly adding a new page type for different behavior or adding a new block that contributes context to the page feel like more sustainable approaches to me.

higs4281: Agreed. Adding behavior to the block won't help us, because we need the context on pages that will not have the block. So I think a new page type is appropriate.

higs4281: @chosak MortgagePerformancePage ahoy. 690bb14

cfgov/data_research/scripts/export_public_csvs.py
```diff
@@ -40,6 +43,43 @@
 logger = logging.getLogger(__name__)
 
 
+def format_size(bytecount, suffix='B'):
```
higs4281: done

cfgov/data_research/models.py
```diff
@@ -176,3 +179,20 @@ def __str__(self):
...
+    note = models.TextField(blank=True)
+    updated = models.DateField(auto_now=True)
+
+    def __str__(self):
+        return "{}, updated {}".format(self.name, self.updated)
+
+    class Meta:
+        ordering = ['name']
+        verbose_name_plural = "Mortage metadata"
```
higs4281: thanks!

cfgov/data_research/mortgage_utilities/s3_utils.py
```diff
@@ -14,9 +14,9 @@
 S3_SECRET = os.getenv('AWS_S3_SECRET_ACCESS_KEY')
 BASE_BUCKET = 'files.consumerfinance.gov'
 MORTGAGE_SUB_BUCKET = "data/mortgage-performance"
-# s3_base is "http://files.consumerfinance.gov"
-# files live at
-# http://s3.amazonaws.com/files.consumerfinance.gov/data/mortgage-performance/
+S3_DOWNLOADS_BASE_URL = (
+    'http://{}.s3.amazonaws.com/data/mortgage-performance/downloads'.format(
```
higs4281: done

cfgov/data_research/models.py
higs4281: @chosak Rather than add another dependency to this PR, I'd like to defer the use of django-jsonfields to a future PR.

PR #3276, opened 2017-08-25
Refines how we will post SQS messages

richaagarwal:
- Option to post to Mattermost as well
 - Creates a comment on an issue if it already exists

(Note that this posts to Mattermost when the issue is created and not when comments are added, since it seems like that would be a lot of chat messages)


cfgov/alerts/post_sqs_messages.py
```diff
@@ -69,9 +89,35 @@
...
+                # New issue, post to github
+                issue = repo.create_issue(title=title, body=body)
+                # AND post to chat, if credentials provided
+                if args.mattermost_webhook_url and args.mattermost_username:
+                    post_to_chat(
+                        endpoint=args.mattermost_webhook_url,
+                        username=args.mattermost_username,
+                        message=body,
+                        issue_url=issue.html_url,
+                    )
```
willbarton: Will this post to Mattermost with every SQS alert, or just with ones that create new issues? 

I think we should post to Mattermost for every alert... our when-to-send-alerts policy should ensure that we're not spamming the channel.

PR #3294, opened 2017-08-30
Post to chat whenever we post to github

richaagarwal:
Because we're going to end up with one GitHub issue ever for each alert we'll need to send a mattermost notification every time we receive it.


richaagarwal:
Oh good call, yup

PR #3288, opened 2017-08-29
Output field required note in Spanish if page language is Spanish

Scotchester:
We're about to launch our first (I think) page in Spanish with an email signup form. This PR makes it so that the "(required)" note on a Form Field with Button molecule will appear in Spanish as "(se require)" if the page language is set to Spanish in its Configuration tab.

## Additions

- Template code to check whether a required form field is in English or Spanish and output the required note accordingly

## Testing

1. Pull branch
2. Run server
3. Preview a page with an email sign-up module, like http://localhost:8000/owning-a-home/
4. Observe that the "(required)" note is output as expected in English.
5. Edit that page and change its language to Spanish, then preview again.
6. Observe that the "(se require)" note is output as expected in Spanish.

## Screenshots

screen shot 2017-08-29 at 13 27 17

## Notes

- Need to revisit the class-based hiding/showing of alternate language messages being used on the Share and Expandable modules to see if we can check `page.language` there, as well. I feel like it may not have been possible on those templates for some reason, but I will double-check.

## Todos

- Add a test for this logic?

## Checklist

* [x] PR has an informative and human-readable title
* [x] Changes are limited to a single goal (no scope creep)
* [x] Code can be automatically merged (no conflicts)
* [x] Code follows the standards laid out in the [development playbook](https://github.com/cfpb/development)
* [x] Passes all existing automated tests
* [x] New functions include new tests
* [x] New functions are documented (with a description, list of inputs, and expected output)
* [x] Placeholder code is flagged
* [x] Visually tested in supported browsers and devices
* [x] Project documentation has been updated
* [x] Reviewers requested with the [Assignee tool](https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/) :arrow_right:

PR #3295, opened 2017-08-31
Add Django translation capability with Jinja2 support

willbarton:
This PR adds support for Django translations, with Jinja2 template support and supporting documentation, to cfgov-refresh. Currently configured languages are Spanish and the default English. 

There is still considerable work to be done to replace all of our other one-off Spanish checks and translations, but this gives us something to build on for the future.

[Documentation can be previewed in Github here](https://github.com/cfpb/cfgov-refresh/blob/translation/docs/translation.md)

## Additions

- Adds `LANGUAGES` and `LOCALE_PATHS` to settings
- Adds a `JinjaTranslations` class to provide `gettext` and `ngettext` to the Jinja2 environment
- Adds a `makemessages` management command that overrides the Django command to support Jinja2

## Changes

- Activates the configured language on any `CFGOVPage` to support translations that match the page's language
- Cleans up imports in `sheerlike/__init__.py` and `v1/__init__.py`

## Removals

- Removes duplicate `docs/mkdocs.yml` 

## Testing

This PR can be tested using the instructions in #3288.

1. Check out this branch, `translation`.
2. Merge in the #3288 branch, `field-required-msg-en-es`, `git merge field-required-msg-en-es`.
3. Follow the testing instructions in #3288.

## Checklist


cfgov/ask_cfpb/tests/models/test_pages.py
```diff
@@ -719,13 +720,14 @@ def test_status_string(self):
...
+            test_page.status_string.lower(),
+            _("redirected but not live").lower())
         test_page.live = True
         self.assertEqual(
-            test_page.status_string.lower(), "redirected")
+            test_page.status_string.lower(), _("redirected").lower())
         test_page.redirect_to = None
         self.assertEqual(
-            test_page.status_string.lower(), "live")
+            test_page.status_string.lower(), _("live").lower())
```
willbarton: @higs4281 want to make you aware of this change. All of these strings we're testing are wrapped in `_()` in `status_string()`. Wagtail actually includes a spanish translation of `live`. Because there are Spanish pages that `activate('es')` because their `language` is set to `"es"` earlier in the test run, Spanish is ending up being the active language when this test is run. So, this test was failing with `"live" != "ver en vivo"`. This change should ensure that if we ever provide translations for the other strings this test will not fail, no matter what language might be selected for running the tests.

cfgov/cfgov/settings/base.py
```diff
@@ -181,6 +183,14 @@
 
 LANGUAGE_CODE = 'en-us'
 
+LANGUAGES = (
+    ('en', _('English')),
+    ('es', _('Spanish')),
+)
+LOCALE_PATHS = (
```
Scotchester: Trivial: Add a blank line between these two declarations to match the existing pattern in the surrounding code.

cfgov/ask_cfpb/tests/models/test_pages.py
```diff
@@ -11,6 +11,7 @@
 from bs4 import BeautifulSoup as bs
 
 from django.utils import timezone
+from django.utils.translation import ugettext as _
```
Scotchester: Trivial: Move the two utils lines here to the bottom of the django section to fix the alphabetization.

cfgov/v1/__init__.py
```diff
@@ -16,37 +18,53 @@
...
+
 from bs4 import BeautifulSoup, NavigableString
-from processors.processors_common import fix_link
+
+from flags.template_functions import flag_enabled, flag_disabled
+
 from core.utils import signed_redirect, unsigned_redirect
-from v1.routing import get_protected_url
+from processors.processors_common import fix_link
+from sheerlike import environment as sheerlike_environment
```
Scotchester: I'm not following the ordering scheme you're now using for the imports in this file. Did you do part of it and not finish?

cfgov/v1/management/commands/makemessages.py
```diff
@@ -0,0 +1,43 @@
...
+else:
+    from django.utils.translation import template as trans_real
+
+
+class Command(makemessages.Command):
+
+    def handle(self, *args, **options):
+        # Jinja2 uses {% trans %}…{% endtrans %}
+        jinja_block_re = re.compile(r'^\s*trans(?:\s+|$)')
+        jinja_endblock_re = re.compile(r'^\s*endtrans$')
```
Scotchester: Should this be updated to support your addition of the standard `gettext` and `_` methods as things that can be called in Jinja?

cfgov/v1/management/commands/makemessages.py
```diff
@@ -0,0 +1,43 @@
...
+
+        # Django uses {% blocktrans %}…{% endblocktrans %}
+        django_block_re = trans_real.block_re
+        django_endblock_re = trans_real.endblock_re
```
Scotchester: Same question, but for the Django template syntax referenced in the docs you're adding below in this PR: `{% trans "Hello World!" %}`

cfgov/v1/__init__.py
willbarton: This file is somewhat beyond hope, IMHO. I'll see what running `isort` does for it.

cfgov/v1/management/commands/makemessages.py
willbarton: Similar to the above, there's [a distinction in Django's `trans_real` between inlines (`{% trans %}`) and blocks](https://github.com/django/django/blob/stable/1.8.x/django/utils/translation/trans_real.py#L544-L546), so we don't need to override for that.

cfgov/v1/management/commands/makemessages.py
willbarton: Django's `trans_real` already matches [`_()`](https://github.com/django/django/blob/stable/1.8.x/django/utils/translation/trans_real.py#L548). I could add `gettext()`, but I think it'd be better to stick to the `_()` convention. This also doesn't prevent just adding the string to the `.po` file manually. `makemessages` is purely for automating that step.

willbarton: I [added an additional comment along these lines below](https://github.com/cfpb/cfgov-refresh/pull/3295/files#diff-15ca6490a115a0eeaffbf9774721a606R32).

cfgov/v1/management/commands/makemessages.py
willbarton: I [added an additional comment along these lines below](https://github.com/cfpb/cfgov-refresh/pull/3295/files#diff-15ca6490a115a0eeaffbf9774721a606R32).

cfgov/v1/__init__.py
```diff
@@ -1,52 +1,63 @@
...
 from v1.routing import get_protected_url
-
+from v1.util.util import get_unique_id
 
 default_app_config = 'v1.apps.V1AppConfig'
 
 
+class JinjaTranslations:
```
chosak: I know this has a constrained use, but shouldn't this be `class JinjaTranslations(object):` for Python 2 object compatibility?

cfgov/ask_cfpb/tests/models/test_pages.py
higs4281: This test doesn't care about language, and adding the ugettext prefix should be fine.

PR #3298, opened 2017-08-31
Fix browser support list IE support

anselmbradford:
@contolini found that the ie string in the browser support list was not doing anything. This changes the array to support IE9. It also splits the array into two for js and css to support different sets.

## Changes

- Split `getSupportedBrowserList` into two arrays for css and js.
- Change browser list support list to properly include older IE support.

## Testing

1. `gulp build` should pass.
2. IE9 should not look blown up.

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -28,17 +28,37 @@ const envvars = {
 /**
  * @description
  * Browser list for autoprefixer, see https://github.com/ai/browserslist.
+ * Support output lists can be viewed via copy/pasting into http://browserl.ist.
+ * @param {string} codeType - The code to support. Either 'js' or 'css'.
  * @returns {array} List of supported browser strings.
  */
-function getSupportedBrowserList() {
-  return [
-    'last 2 version',
-    'Edge >= 11',
-    'not ie <= 8',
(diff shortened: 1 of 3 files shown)


config/environment.js
```diff
@@ -28,17 +28,37 @@ const envvars = {
 /**
  * @description
  * Browser list for autoprefixer, see https://github.com/ai/browserslist.
+ * Support output lists can be viewed via copy/pasting into http://browserl.ist.
```
contolini: To make our lives a little easier, add the browserlist permalink to the comment: https://goo.gl/AXDSxG

anselmbradford: I didn't because that will totally be a URL that will be overlooked if the browserlist gets updated.

contolini: It's true but it took me a few minutes to figure out how to input values into the site (comma separated list of values *without* the single quotes).

anselmbradford: I'll leave it for now, I'm not keen on obfuscating the domain either, and the full URL is really long. It'll be a minor inconvenience, but better than before where the domain wasn't in the comments at all.

PR #3299, opened 2017-08-31
Remove Info Unit Group from Blog pages, restrict to 25/75 layout on Learn pages

Scotchester:
When Info Unit Groups were added recently, I overlooked the fact that we needed the ability to disallow particular layouts on certain page types. This PR removes them entirely from Blog pages, and restricts the layout options to 25/75 only on a Learn page.

## Additions

- `InfoUnitGroup2575Only`, a subclass of `InfoUnitGroup` that offers only 25/75 as a choice for the `format` field
- CSS to hide the automatically generated select menu for `format` when rendering the Wagtail interface for `InfoUnitGroup2575Only`

## Removals

- `InfoUnitGroup` removed as an available module on Blog pages

## Changes

- `InfoUnitGroup2575Only` traded for the standard `InfoUnitGroup` as an available module on Learn pages
- Some minor cleanup of the existing Heading Block CSS

## Testing

1. Pull branch
2. Migrate
3. Run server
4. Create/edit a Learn page
5. Add an Info Unit Group to the content area
6. Observe that you can not select a format other than 25/75
7. Create/edit a Blog page
8. Observe that you cannot add an Info Unit Group.

## Screenshots

This is what you see when adding an Info Unit Group to a Learn page:

screen shot 2017-08-07 at 12 46 00

## Checklist


Scotchester:
Context for this change can be found in https://[GHE]/CFPB/hubcap/issues/55

sebworks:
@Scotchester, are we just creating the new type to restrict the choices or a 25/75?

Scotchester:
> @Scotchester, are we just creating the new type to restrict the choices or a 25/75?

That is correct.

sebworks:
@Scotchester, would it be possible to handle this with just validation or in a dynamic fashion? It feels strange to create a type just to restrict it's usage on particular page types.

Scotchester:
> would it be possible to handle this with just validation or in a dynamic fashion? It feels strange to create a type just to restrict it's usage on particular page types.

I feel ya, @sebworks.

We decided that restricting it via validation was an unacceptable user experience. Andy and I paired on this for an hour or two and we couldn't come up with anything more dynamic than this. It's unfortunately not possible to override a block field's `choices` property when adding the block to a page class, so doing the subclassing seemed like the DRY-est way to handle it.

(We did consider a front-end only solution, but figured that keeping this logic on the back-end was preferable.)

sebworks:
@Scotchester, thanks for the reply. I'm not sure this is the right approach because this is a validation issue, which we are handling by creating a new type.  Would it be possible to focus on validation instead?

chosak:
@sebworks I paired for a bunch with @Scotchester on this, and this seems to be the best option. Doing it via validation would mean that the invalid choices would still appear in the Wagtail UX, and users would only see the error when saving the page. Restricting the choices seems like a more straightforward approach, and also more consistent with the atomic design philosophy of having blocks that have consistent self-describing behavior.

sebworks:
@chosak, ok. Im just thinking that we are recreating every variation of the info unit as a specific component type. I just think we should rethink how we are using it and restricting it. Having said that, the PR looks fine for me.

PR #3303, opened 2017-09-05
Refactor data research API

higs4281:
The initial API was exploratory and leaned on a metadata helper class
called FIPS. Now that all data structures are better understood, we can
refactor the API to remove reliance on the FIPS helper, simplify
save and aggregation methods and speed up operations by normalizing metadata.

## Additions
- data model for "non-MSA" areas, which represent counties in a state that are not part of any Metropolitan Statistical Area. Some may call it rural; I call it antimatter.
- new geo models for county, metro area and state to handle metadata.

The refactor implements excellent suggestions from @chosak and @richaagarwal to remove reliance on the FIPS class, simplify save and data aggregation methods and make use of the [django-jsonfield](https://pypi.python.org/pypi/jsonfield) module.

## Testing
To see the API in action:
- Run migrations
  ```
  ```
- Download the data set http://files.consumerfinance.gov.s3.amazonaws.com/data/mortgage-performance/initial_data/mortgagedata_refactor2.sql.gz
- Load it
  ```
  gunzip -c ~/Downloads/mortgagedata_refactor2.sql.gz | mysql -uroot v1
  ```
Check out some endpoints:
    - [county time series data](http://localhost:8000/data-research/mortgages/api/v1/time-series/90/12081/)
    - [metro map data](http://localhost:8000/data-research/mortgages/api/v1/map-data/90/metros/2009-01/)
    - [metadata: sampling dates](http://localhost:8000/data-research/mortgages/api/v1/metadata/sampling_dates/)


cfgov/data_research/models.py
```diff
@@ -21,6 +20,158 @@ class ConferenceRegistration(models.Model):
...
+    counties = JSONField(
+        blank=True, help_text='FIPS list of counties in the state')
+    non_msa_counties = JSONField(
+        blank=True,
+        help_text='FIPS list of counties in the state that are not in an MSA')
+    msas = JSONField(
+        blank=True, help_text='FIPS list of MSAs in the state')
+    non_msa_valid = models.BooleanField(default=False)
+
+    def __unicode__(self):
```
willbarton: In the metadata models above you're using `__str__`. That's more future-friendly. If you're running into unicode issues, you can use [Django's `python_2_unicode_compatible` decorator](https://docs.djangoproject.com/en/1.8/topics/python3/#str-and-unicode-methods). 

I think we should prefer `__str__` and use that decorator for all of the models using `__unicode__` to make our job of becoming Python3-compatible a little easier whenever that happens in the future.

cfgov/data_research/models.py
```diff
@@ -21,6 +20,158 @@ class ConferenceRegistration(models.Model):
...
+            date__year=threshold_year)
+        annual_sum = sum(record.total for record in records)
+        monthly_average = annual_sum * 1.0 / records.count()
+        if monthly_average >= threshold_count:
+            self.valid = True
+        else:
+            self.valid = False
+        self.save()
+
+    def __unicode__(self):
```
willbarton: Same as above, use `__str__` and `@python_2_unicode_compatible`.

cfgov/data_research/models.py
```diff
@@ -21,6 +20,158 @@ class ConferenceRegistration(models.Model):
...
+        annual_sum = sum(record.total for record in records)
+        monthly_average = (
+            annual_sum * 1.0 / records.count() if records.count() > 0 else 0)
+        if monthly_average >= threshold_count:
+            self.valid = True
+        else:
+            self.valid = False
+        self.save()
+
+    def __unicode__(self):
```
willbarton: Same as above, use `__str__` and `@python_2_unicode_compatible`.

higs4281:
@willbarton 
> It's also unclear to me when and where the data load/validation/export scripts are going to end up running, and how often.

Load, validation and export scripts and related aggregations will run roughly once a quarter.  We don't have an exact timetable because our source, the National Mortgage Database, does not schedule its updates.

The intention is to automate the use of loading scripts into a partial pipeline, but that is for a later sprint.

cfgov/data_research/models.py
higs4281: ack! `__str__` was my intention (I wrote model tests of `__str__`) but I slipped into old habits.

higs4281:
@willbarton 
> I feel like there are an awful lot of queries all over the place ...

This is a fair point -- there are lots of queries, thanks to metadata extrapolations -- but from the API's perspective it's not as bad as it seems. 

The views do have a number of potential queries that can run, but generally only one (or two, for MSA/non-MSAs) will actually fire, depending on which geography has been requested. I was hoping to keep things DRY by passing different geos to the same views.

The hairiest queries are model validation and aggregation methods, but they will only run once a quarter, and not on production servers.

That's not to say we won't have API optimizations to make, but at this stage API response isn't noticeable in the app.

PR #3305, opened 2017-09-05
Improve formatting of Github issue created from SQS script

richaagarwal:
- Adds a space to # so it doesn't erroneously link to a Github issue
- Remove the `[Open]` syntax so that the link expands

See GHE 1146 as an example of how this looks like now, and 1148 as an example of how it will look like after these changes.

### DIFF
--- a/cfgov/alerts/post_sqs_messages.py
+++ b/cfgov/alerts/post_sqs_messages.py
@@ -100,7 +100,11 @@ if __name__ == '__main__':
     )
 
     for message in response.get('Messages', {}):
-        body = message.get('Body')
+        body = message.get('Body').replace(
+            '#', '# '  # Avoids erroneous Github issue link
+        ).replace(
+            '[Open]', ''  # We want to expand the link
+        )
         title = body.split(" - ")[0]
         logger.info('Retrieved message {} from SQS'.format(body))
 


cfgov/alerts/post_sqs_messages.py
```diff
@@ -100,7 +100,11 @@ def post_to_chat(endpoint, username, message, issue_url):
     )
 
     for message in response.get('Messages', {}):
-        body = message.get('Body')
+        body = message.get('Body').replace(
+            '#', '# '  # Avoids erroneous Github issue link
+        ).replace(
+            '[Open]', ''  # We want to expand the link
+        )
```
willbarton: These are Very Good Changes, but I wonder if it's worth trying to make these changes in the [Jenkins SQS plugin](https://github.com/cfpb/jenkins-sqs-plugin) instead of here.

Scotchester:
They're good changes, Brent.

cfgov/alerts/post_sqs_messages.py
richaagarwal: Makes sense to modify the source, but in this case the true source is a Mattermost jenkins plugin that we forked & didn't alter too much. I think it'll be a little more straightforward to introduce the changes here.

contolini: I'd make the change here to keep the jenkins plugin as agnostic as possible.

PR #3321, opened 2017-09-08
Refactor `post_sqs_messages` script

richaagarwal:
In addition to cleaning up some of this code, this was necessary for a branch I'm working on which needs to reuse the "post to github" functionality (see https://github.com/cfpb/cfgov-refresh/pull/3322/files#diff-a2de39d3ae75d0257afafb3cb2d587ce)

### DIFF
--- a/cfgov/alerts/github_alert.py
+++ b/cfgov/alerts/github_alert.py
@@ -3,34 +3,39 @@ import os
 import github3
 
 
-class GithubAlert:
+class GithubAlert(object):
     def __init__(self, credentials):
-        token = credentials.get(
+        self.token = credentials.get(
             'token',
             os.environ.get('GITHUB_TOKEN')
         )
-        url = credentials.get(
+        self.url = credentials.get(
             'url',
             os.environ.get('GITHUB_URL')
         )
-        user = credentials.get(
+        self.user = credentials.get(
             'user',
             os.environ.get('GITHUB_USER')
         )
-        repo_name = credentials.get(
+        self.repo_name = credentials.get(
             'repo_name',
             os.environ.get('GITHUB_REPO')
         )
 
-        gh = github3.login(
-            token=token,
-            url=url,
-        )
-
-        self.repo = gh.repository(user, repo_name)
+    def repo(self):
+        try:
+            gh = github3.login(
+                token=self.token,
+                url=self.url,
+            )
+            return gh.repository(self.user, self.repo_name)
+        except:
+            raise ValueError(
+                'Github credentials provided are incorrect'
+            )
 
     def matching_issue(self, title):
-        issues = self.repo.iter_issues(state='all')
+        issues = self.repo().iter_issues(state='all')
         return next((issue for issue in issues if issue.title == title), None)
 
     def post(self, title, body):
@@ -42,7 +47,7 @@ class GithubAlert:
             issue.create_comment(body=body)
         else:
             # New issue, post to github
-            issue = self.repo.create_issue(
+            issue = self.repo().create_issue(
                 title=title,
                 body=body,
                 labels=[
--- a/cfgov/alerts/github_alert.py
+++ b/cfgov/alerts/github_alert.py
@@ -23,16 +23,11 @@ class GithubAlert(object):
         )
 
     def repo(self):
-        try:
-            gh = github3.login(
-                token=self.token,
-                url=self.url,
-            )
-            return gh.repository(self.user, self.repo_name)
-        except:
-            raise ValueError(
-                'Github credentials provided are incorrect'
-            )
+        gh = github3.login(
+            token=self.token,
+            url=self.url,
+        )
+        return gh.repository(self.user, self.repo_name)
 
     def matching_issue(self, title):
         issues = self.repo().iter_issues(state='all')
--- a/cfgov/alerts/github_alert.py
+++ b/cfgov/alerts/github_alert.py
@@ -1,0 +1,53 @@
+import os
+
+import github3
+
+
+class GithubAlert:
+    def __init__(self, credentials):
+        token = credentials.get(
+            'token',
+            os.environ.get('GITHUB_TOKEN')
+        )
+        url = credentials.get(
+            'url',
+            os.environ.get('GITHUB_URL')
+        )
+        user = credentials.get(
+            'user',
+            os.environ.get('GITHUB_USER')
+        )
+        repo_name = credentials.get(
+            'repo_name',
+            os.environ.get('GITHUB_REPO')
+        )
+
+        gh = github3.login(
+            token=token,
+            url=url,
+        )
+
+        self.repo = gh.repository(user, repo_name)
+
+    def matching_issue(self, title):
+        issues = self.repo.iter_issues(state='all')
+        return next((issue for issue in issues if issue.title == title), None)
+
+    def post(self, title, body):
+        issue = self.matching_issue(title)
+        if issue:  # Issue already exists
+            if issue.is_closed():
+                issue.reopen()
+            # add comment to it to document it happened again
+            issue.create_comment(body=body)
+        else:
+            # New issue, post to github
+            issue = self.repo.create_issue(
+                title=title,
+                body=body,
+                labels=[
+                    'Maintenance and Response',
+                    'alert'
+                ],
+            )
+        return issue
--- a/cfgov/alerts/mattermost_alert.py
+++ b/cfgov/alerts/mattermost_alert.py
@@ -19,7 +19,8 @@ class MattermostAlert(object):
             'text': text,
             'username': self.username,
         }
-        requests.post(
+        resp = requests.post(
             self.webhook_url,
             data=json.dumps(payload),
         )
+        resp.raise_for_status()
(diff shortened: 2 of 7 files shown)


cfgov/alerts/github_alert.py
```diff
@@ -0,0 +1,53 @@
+import os
+
+import github3
+
+
+class GithubAlert:
```
chosak: This should be `GithubAlert(object):` for Python 2 compatibility.

cfgov/alerts/github_alert.py
```diff
@@ -0,0 +1,53 @@
...
+    def __init__(self, credentials):
+        token = credentials.get(
+            'token',
+            os.environ.get('GITHUB_TOKEN')
```
chosak: Minor suggestion: worth changing these things to something like this:

```py
token = credentials.get('token') or os.environ['GITHUB_TOKEN']
```

It'd be nice to have these things raise an exception if not defined, if you're falling back to them.

cfgov/alerts/mattermost_alert.py
```diff
@@ -0,0 +1,25 @@
+import json
+import os
+import requests
+
+
+class MattermostAlert:
```
chosak: Same comment as above re: use of `object` for Python 2.

cfgov/alerts/mattermost_alert.py
```diff
@@ -0,0 +1,25 @@
...
+    def __init__(self, credentials):
+        self.username = credentials.get(
+            'username',
+            os.environ.get('MATTERMOST_USERNAME')
```
chosak: Same comment as above re: environment variables.

cfgov/alerts/mattermost_alert.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def post(self, text):
+        payload = {
+            'text': text,
+            'username': self.username,
+        }
+        requests.post(
+            self.webhook_url,
+            data=json.dumps(payload),
+        )
```
chosak: Will this raise an exception if for some reason the POST fails? Otherwise might be nice to catch the response and `raise_for_status`.

cfgov/alerts/post_sqs_messages.py
```diff
@@ -1,9 +1,9 @@
 import argparse
 import boto3
-import json
 import logging
-import github3
-import requests
+
```
chosak: Minor request: add `from __future__ import unicode_literals` so that if the body string contains unicode characters it won't raise an exception when used in string formatting below.

cfgov/alerts/post_sqs_messages.py
```diff
@@ -55,26 +55,32 @@
...
-        'text': text,
-        'username': args.mattermost_username,
-    }
-    requests.post(
-        args.mattermost_webhook_url,
-        data=json.dumps(payload),
+
+def cleanup_jenkins_message(message):
+    return message.replace(
+        '#', '# '  # Avoids erroneous Github issue link
```
chosak: This seems like it could be potentially over greedy - is it possible that error messages might contain this character elsewhere? You could do something a bit more specific like:

```py
>>> import re
>>> msg = 'Error #123 where string contains # character'
>>> re.sub(r'#(\d+)', r'# \1', msg)
'Error # 123 where string contains # character'
```

cfgov/alerts/post_sqs_messages.py
```diff
@@ -55,26 +55,32 @@
...
-    )
-    payload = {
-        'text': text,
-        'username': args.mattermost_username,
-    }
-    requests.post(
-        args.mattermost_webhook_url,
-        data=json.dumps(payload),
+
+def cleanup_jenkins_message(message):
```
willbarton: I still feel like it'd be nice if this happened in the posting-to-sqs instead of the reading-from-sqs, but not necessary to address for the same reasons as before.

In that same line of thought, the `#1234` is going to be a problem regardless of whether the source if Jenkins. Do we want to make that a regex replacement that checks for `r'^#[0-9]+` instead. And we could always expand markdown links. 

Then we could remove `jenkins` from the function name?

richaagarwal: I'll rename it to `cleanup_message` for now. We can refine it in terms of regexes or move it somewhere else in the next iteration.

richaagarwal:
@willbarton or @chosak thanks for the feedback! bulk of updates I made are reflected in these two commits -- if you feel inclined to review again.  otherwise will be merging soon! 

https://github.com/cfpb/cfgov-refresh/pull/3321/commits/c349ac542c51561132ea83df20e700181db7e725
https://github.com/cfpb/cfgov-refresh/pull/3321/commits/b3570d93091b8a4173debd54b2e9099055b0cb5d.

cfgov/alerts/tests/test_github_alert.py
```diff
@@ -0,0 +1,74 @@
...
+from mock import patch
+
+import github3
+
+from alerts.github_alert import GithubAlert
+
+
+class TestGithubAlert(TestCase):
+
+    closed_issue = github3.issues.issue.Issue(
```
chosak: FYI, after our conversation yesterday around mocking I came across [this part](http://github3py.readthedocs.io/en/latest/testing.html#example-testing-the-release-object) of the github3 library's documentation which talks about a recommended way of mocking.

cfgov/alerts/github_alert.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+
+    def repo(self):
+        try:
+            gh = github3.login(
+                token=self.token,
+                url=self.url,
+            )
+            return gh.repository(self.user, self.repo_name)
+        except:
```
chosak: [The Python docs](https://docs.python.org/2/howto/doanddont.html#except) recommend against using bare `except:` clauses like this because they can hide the real cause of errors and also prevent breaking code using e.g. Ctrl-C. Would it be sufficient to catch [`GithubError`](https://github.com/github3py/github3py/blob/master/github3/models.py#L357)?

cfgov/alerts/github_alert.py
```diff
@@ -0,0 +1,58 @@
+import os
+
+import github3
+
+
+class GithubAlert(object):
+    def __init__(self, credentials):
```
chosak: Minor suggestion: what about passing the credentials as kwargs, i.e. `def __init__(self, **credentials)`? This wouldn't require changing any code below, but would clean up usage of this class a bit, allowing you to do `GithubAlert()` instead of `GithubAlert({})`?

Similar comment applies to `MattermostAlert`.

cfgov/alerts/github_alert.py
```diff
@@ -0,0 +1,58 @@
+import os
+
+import github3
+
+
+class GithubAlert(object):
```
chosak: Putting on my @Scotchester hat to make a super-nitpicky observation that this should be `GitHubAlert` even though I always type "Github" every time I refer to the company that makes the product I am typing on right tnow.

cfgov/alerts/post_sqs_messages.py
```diff
@@ -1,9 +1,11 @@
...
+
 import argparse
 import boto3
-import json
 import logging
-import github3
-import requests
+
+from github_alert import GithubAlert
+from mattermost_alert import MattermostAlert
```
willbarton: Could these imports be absolute instead?

chosak: Why wouldn't that work? How is this script invoked? At a minimum wouldn't it be nice to make these imports explicitly relative (e.g. `from .github_alert import GitHubAlert`)?

willbarton: It's invoked directly, `python cfgov/alerts/post_sqs_messages.py`, so the `PYTHONPATH` doesn't include the `alerts` package.

I agree, I'd like to see more explicit imports. I have a soft preference for requiring the `PYTHONPATH` to be set explicitly to include `cfgov/` and then absolute-importing them with `alerts.github_alert`.

chosak: If `alerts` was added as a top-level app to `settings.INSTALLED_APPS`, could this be converted to a management command?

willbarton: But we would then have to have a full installation of Django and all our requirements in the Jenkins workspaces where this (and the New Relic) script runs. That seems like massive overkill.

cfgov/alerts/github_alert.py
richaagarwal: Oh nice, good catch!

richaagarwal: Thinking about this more, I'm going to remove the `try` entirely.  The Github API will throw its own (more specific) errors.

cfgov/alerts/post_sqs_messages.py
richaagarwal: Updating to be absolute

cfgov/alerts/tests/test_github_alert.py
richaagarwal: I had seen that too, but didn't look too closely.  Was there anything specific from that resource that you think should inform the design here?

cfgov/alerts/github_alert.py
richaagarwal: I like this idea and may address it in a separate PR.

PR #3330, opened 2017-09-11
Changes Activity Log to Recent updates

anselmbradford:
Fixes https://[GHE]/CFGOV/platform/issues/1112

## Changes

- Changes Activity Log page title to "Recent updates" (sentence case)
- Changes "Activity Log" in mega menu to "Recent Updates". Within the mega menu, moves "Recent Updates" to the top of the column, above Blog, Newsroom, Events.
- Changes "See all activities" link under "Latest updates" section at the bottom of the Homepage to "See all updates".

## Testing

1. Visit homepage for "See all updates" change.
2. localhost:8000/activity-log/ to see change to page title, and to check mega menu.

## Screenshots

screen shot 2017-09-11 at 10 32 14 am

screen shot 2017-09-11 at 10 32 18 am

screen shot 2017-09-11 at 12 09 35 pm

### DIFF
--- a/cfgov/jinja2/v1/activity-log/index.html
+++ b/cfgov/jinja2/v1/activity-log/index.html
@@ -3,7 +3,7 @@
 {% block content_main %}
 
     <div class="block block__flush-top">
-        <h1 data-qa-hook="main-title">Activity Log</h1>
+        <h1 data-qa-hook="main-title">Recent updates</h1>
         <p class="lead-paragraph" data-qa-hook="main-summary">
             Find the latest CFPB activities and publications here. Use the filters
             below to browse by date, specific topics, or types of posts.
--- a/cfgov/v1/util/ref.py
+++ b/cfgov/v1/util/ref.py
@@ -25,7 +25,7 @@ related_posts_categories = [
 ]
 
 page_types = [
-    ('activity-log', 'Activity Log'),
+    ('activity-log', 'Recent updates'),
     ('administrative-adjudication', 'Administrative adjudication'),
     ('amicus-brief', 'Amicus Brief'),
     ('blog', 'Blog'),
@@ -159,7 +159,7 @@ def related_posts_category_lookup(related_categories):
 
 def page_type_choices():
     new_choices = [
-        ('Activity Log', (
+        ('Recent updates', (
             ('blog', 'Blog'),
             ('op-ed', 'Op-Ed'),
             ('press-release', 'Press Release'),
--- a/cfgov/jinja2/v1/_includes/organisms/_vars-mega-menu.html
+++ b/cfgov/jinja2/v1/_includes/organisms/_vars-mega-menu.html
@@ -298,10 +298,10 @@
 
 {% set about_us_group_two = {
     'nav_items': [
+        {'text': 'Recent Updates', 'url': '/activity-log/'},
         {'text': 'Blog', 'url': '/about-us/blog/'},
         {'text': 'Newsroom', 'url': '/about-us/newsroom/'},
         {'text': 'Events', 'url': '/about-us/events/'},
-        {'text': 'Recent Postings', 'url': '/activity-log/'},
     ]
 } %}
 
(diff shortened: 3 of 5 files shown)


Scotchester:
Should we be changing the URL from `activity-log` to `recent-updates` in conjunction with this?

cfgov/v1/util/ref.py
```diff
@@ -25,7 +25,7 @@
 ]
 
 page_types = [
-    ('activity-log', 'Activity Log'),
+    ('activity-log', 'Recent updates'),
```
Scotchester: I'm not entirely clear on what the slug in this tuple corresponds to, but I think it should change to match the new name?

Also, we have some inconsistent capitalization for the full names in this list. Currently everything is title case except for "Administrative adjudication".

anselmbradford: Any idea where I would test this?

anselmbradford: Maybe @richaagarwal @chosak knows?

nataliafitzgerald:
@anselmbradford 
I believe that @schaferjh said that changing the URL was not necessary for this update since users won't access this page directly. @schaferjh @sarahsimpson09 - Can you confirm?

schaferjh:
Changing the URL isn't necessary, thanks.

nataliafitzgerald:
@anselmbradford 
Can you add a screenshot of the homepage updates?

anselmbradford:
> I believe that @schaferjh said that changing the URL was not necessary for this update since users won't access this page directly.

Yup! URL is not updated.

anselmbradford:
> Can you add a screenshot of the homepage updates?

@nataliafitzgerald Added to original post above!

cfgov/v1/util/ref.py
chosak: This code is unfortunately pretty messy. It relates to page types that can be searched using a Filter Controls block on a filterable page. For example, after this PR, a page can be configured with a Filter Controls module of "Recent updates" page type:

image

When rendered, this page's filter form then contains the various hardcoded categories associated with that page type (under "Category"):

image

This means that any existing pages in the database that have a form configured like this already have the hardcoded slug `activity-log`, and so changing the slug here as @Scotchester suggests would break those pages. A simultaneous data migration would be needed to keep those in sync.

Regarding the second item in the tuple: its main function is for display in the Wagtail admin, so correcting capitalization should be okay. There's also some additional functionality (see [here](https://github.com/cfpb/cfgov-refresh/pull/3330/files#diff-6de80da6a880654cf8a657eea06ef20aL169)) that does some manual updating of page type - category mappings, and these need to be in sync. So capitalizing "Administrative adjudication" would need to happen in both places.

@richaagarwal did some nice investigative work in this area and in the past we've discussed converting all of this over to some kind of Django model-backed store, which would make these kind of changes less brittle.

cfgov/jinja2/v1/activity-log/index.html
```diff
@@ -3,7 +3,7 @@
 {% block content_main %}
 
     <div class="block block__flush-top">
-        <h1 data-qa-hook="main-title">Activity Log</h1>
+        <h1 data-qa-hook="main-title">Recent updates</h1>
```
chosak: Was the change from capital `L` to lowercase `u` intentional?

anselmbradford: Yes, it should be sentence case, see https://[GHE]/CFGOV/platform/issues/1112#issuecomment-126298 for background.

cfgov/v1/util/ref.py
anselmbradford: Great background @chosak! So if I understand correctly the changes I have made are okay to go in?

chosak: Yes, the only reason I didn't approve was because of the outstanding TODO of renaming the existing Activity Log page. As long as that doesn't get dropped, then this is 👍 !

anselmbradford:
> There's a TODO after this PR which is to actually rename the single "Activity Log" page to "Recent Updates".

Updated this to "Recent updates" (to match incoming case changes) in content.consumerfinance.gov. Do I need to update it anywhere else?

chosak:
For consistency this should be updated on build as well.

anselmbradford:
> For consistency this should be updated on build as well.

Done!

PR #3337, opened 2017-09-13
Refactor New Relic alerts, use a local cache file of violations

willbarton:
This PR refactors the New Relic alert violation processing to make it more testable and to handle previously-seen violations differently.

We had been trying to use the New Relic `opened_at` and `duration` values for violations to track when a violation occurred. If we run every 60 seconds, our assumption was that violations that have been open for less than 60 seconds were new. This turns out not to be true — some violations are not available via the API until well after 60 seconds, but their duration still reflects when the condition began — they'd first appear with durations of 180+ seconds.

Instead, this script now takes a `--known_violations_file` in which to store violation ids, one per-line, and violations are added to this file when they are first observed by the script. That file is presumed to persist.

### DIFF
--- a/cfgov/alerts/newrelic_alerts.py
+++ b/cfgov/alerts/newrelic_alerts.py
@@ -63,7 +63,7 @@ class NewRelicAlertViolations(object):
         )
         body = (
             'New Relic {product}, {name}, {label}.'
-            '<a href="{link}">View incidents</a>'
+            'View incidents: {link}'
         ).format(
             product=violation['entity']['product'],
             type=violation['entity']['type'],
--- a/cfgov/alerts/newrelic_alerts.py
+++ b/cfgov/alerts/newrelic_alerts.py
@@ -1,0 +1,75 @@
+import re
+
+import requests
+
+
+class NewRelicAlertViolations(object):
+
+    newrelic_url = 'https://api.newrelic.com/v2/'
+
+    def __init__(self, newrelic_token, policy_filter, account_number,
+                 known_violations=[]):
+        self.newrelic_token = newrelic_token
+        self.policy_filter_re = re.compile(policy_filter)
+        self.account_number = account_number
+        self.known_violations = known_violations
+
+    def get_current_violations(self):
+        """ Check for all violations currently open in New Relic """
+        headers = {'X-Api-Key': self.newrelic_token}
+        violations_url = (self.newrelic_url +
+                          'alerts_violations.json?only_open=true')
+        r = requests.get(violations_url, headers=headers)
+        response_json = r.json()
+
+        # Filter on the policy name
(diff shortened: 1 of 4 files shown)


willbarton:
Adding a hold label to this because it'll require a change to the Jenkins job that invokes it.

richaagarwal:
Is there a reason the `known_violations` code lives in `send_new_relic_messages_to_sqs.py` and not `newrelic_alerts.py`?

cfgov/alerts/send_new_relic_messages_to_sqs.py
```diff
@@ -67,52 +67,35 @@
...
-        '<a href="{link}">View incidents</a>'
-    ).format(
-        product=violation['entity']['product'],
-        type=violation['entity']['type'],
-        label=violation['label'],
-        link=incidents_link
-    )
-    message_body = '{title} - {body}'.format(title=title, body=body)
-    return message_body
+def cache_known_violations(known_violations_filename, known_violations):
```
richaagarwal: It seems like all this code around known violations could live inside `newrelic_alerts.py` instead. Then you could pass an instance of `NewRelicAlertViolations` the filename instead of the violations. Was there a reason for this separation?

richaagarwal:
This looks great & love how verbose it is!  I left a comment about one question I have but don't let that be a blocker.

cfgov/alerts/newrelic_alerts.py
```diff
@@ -0,0 +1,75 @@
...
+        )
+        incidents_link = (
+            'https://alerts.newrelic.com/accounts/'
+            '{account_number}/incidents'
+        ).format(
+            account_number=self.account_number
+        )
+        body = (
+            'New Relic {product}, {name}, {label}.'
+            '<a href="{link}">View incidents</a>'
```
richaagarwal: Can we remove the `<a href></a>` html syntax here?  It doesn't work well in Mattermost.

cfgov/alerts/send_new_relic_messages_to_sqs.py
willbarton: I thought maybe that file management was best left the script itself. That way we could use the same code in `newrelic_alerts` with some other management scheme for `know_violations` in the future. 

That might be overkill though. I kind of prefer it the way I've set it up, but moving the caching to `NewRelicAlertViolations` is certainly something I can do if you think it's necessary.

cfgov/alerts/newrelic_alerts.py
willbarton: Ah, yes we can. Are we stripping Markdown links entirely? The URL is not particularly informative, and only works in Citrix, but at least it gives you a place to start (I still can't find a way to go from API-readable URL to human-readable URL for an individual alert).

PR #3344, opened 2017-09-14
 Run all generated JS through Babel

anselmbradford:
## Changes

- Moves webpack configuration `module` to a constant shared by all webpack configurations so that each uses the same babel/browserlist pipeline.
- Updates webpack and babel npm modules to latest.

## Testing

1. Run `npm install`
2. Run `gulp build`
3. Search the output JS in `static_built/js/` for `const`.

## Notes

- I restructured the webpack configuration slightly to rename `rules` to `loaders` as I didn't find `module.rules` on https://webpack.github.io/docs/configuration.html#module (I guess maybe it worked before for backward compatibility??? Dunno). Also, I updated `use` to `loaders` per that same doc.

### DIFF
--- a/config/webpack-config.js
+++ b/config/webpack-config.js
@@ -16,33 +16,35 @@ const UglifyWebpackPlugin = require( 'uglifyjs-webpack-plugin' );
 const JS_ROUTES_PATH = '/js/routes';
 const COMMON_BUNDLE_NAME = 'common.js';
 
+// Commmon webpack 'module' option used in each configuration.
+// Runs code through Babel and uses global supported browser list.
+const COMMON_MODULE_CONFIG = {
+  loaders: [ {
+    test: /\.js$/,
+    loaders: [ {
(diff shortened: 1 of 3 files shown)


config/webpack-config.js
```diff
@@ -16,33 +16,35 @@ const UglifyWebpackPlugin = require( 'uglifyjs-webpack-plugin' );
...
-            targets: {
-              browsers: environment.getSupportedBrowserList( 'js' )
-            },
-            debug: true
-          } ] ]
-        }
-      } ],
-      exclude: {
-        test: /node_modules/,
-        // TODO: Move this into a config variable so that we can easily add
```
anselmbradford: @contolini I _think_ I addressed your TODO here by moving this into a constant shared by all configs. Is that right?

contolini: Yeah works for me. 👍

PR #3346, opened 2017-09-14
Removes code climate badge

anselmbradford:
The Code Climate badge hasn't been updated in more than a year because the org permissions don't allow Code Climate and until GH allows finer grained control of permissions this doesn't look like it will updated any time soon.

## Removals

- Removes outdated Code Climate badge from Readme :-(


anselmbradford:
> This is the saddest approval.

I know! 😭 😭 😭

PR #3350, opened 2017-09-15
Mortgage performance: Fix ordering of MSA dropdown

higs4281:
This alphabetizes the state MSA dropdown list and adds the non-MSA entry at the
bottom of the list, in case we haven't turned to other solutions.

The ordering in the API call can be checked on build:

[BUILD]/data-research/mortgages/api/v1/metadata/state_msa_meta/


contolini:
It makes sense to outsource the alphabetizing to the back-end. 👍

PR #3353, opened 2017-09-15
Mortgage performance: move notes to page

higs4281:
This is a small optimization to simplify content handling and optimize
page rendering a bit. Data notes belong in the Wagtail page.

## Additions

-Note field to chart and map blocks for editing the "Note" section of footnotes.

## Removals

- Reliance on the constants table for this content.

## Testing

This generates a migration, because it touches the block definitions, so `python cfgov/manage.py migrate' will be in order to test locally.


cfgov/v1/atomic_elements/organisms.py
```diff
@@ -864,6 +864,9 @@ class MortgageChartBlock(blocks.StructBlock):
     description = blocks.CharBlock(
         required=False,
         help_text='Chart summary for visually impaired users.')
+    note = blocks.CharBlock(
+        required=False,
+        help_text='Note on visualization data.')
```
contolini: This help text isn't super descriptive but anyone using this block will know what the note field is for so it seems fine.

PR #3355, opened 2017-09-18
Add insufficient data tooltip to county map

contolini:
Adds the "insufficient data" tooltip to the county view and changes the insufficient data color from `@gray-10` to `@gray-5`.

## Additions

- Tooltip when the map point has a value of `null`.

## Changes

- Update lowest item in the color range from `@gray-10` to `@gray-5`.
- Use the full namespaced CSS classes instead of `.highcharts-color-0` to avoid conflicts with other CFPB charts.

## Testing

1. Pull down this branch and `./setup.sh`
1. Set up the [mortgage performance API](https://github.com/cfpb/cfgov-refresh/pull/3189).
1. Create a [mortgage chart block and page](https://github.com/cfpb/cfgov-refresh/pull/3260).

## Screenshots

screen shot 2017-09-18 at 4 45 07 am

## Checklist

### DIFF
--- a/cfgov/unprocessed/js/organisms/MortgagePerformanceTrends/apps/map.js
+++ b/cfgov/unprocessed/js/organisms/MortgagePerformanceTrends/apps/map.js
@@ -218,6 +218,9 @@ MortgagePerformanceMap.prototype.renderTooltip = function() {
   return ( point, meta ) => {
     var percent;
     var nationalPercent = Math.round( meta.national_average * 1000 ) / 10;
+    if ( point.value === null ) {
+      return "<div class='m-mp-map-tooltip'>Insufficient data for this area</div>";
+    }
     if ( point.value < 0 ) {
       percent = 'Insufficient data';
     } else {
--- a/cfgov/unprocessed/css/organisms/mortgage-performance-trends.less
+++ b/cfgov/unprocessed/css/organisms/mortgage-performance-trends.less
@@ -27,11 +27,26 @@
             transition: fill 500ms, fill-opacity 500ms, stroke-width 0 !important;
         }
         .highcharts-point {
-            stroke: #838588;
+            stroke: @black;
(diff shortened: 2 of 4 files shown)


cfgov/unprocessed/js/organisms/MortgagePerformanceTrends/apps/map.js
```diff
@@ -218,6 +218,9 @@ MortgagePerformanceMap.prototype.renderTooltip = function() {
   return ( point, meta ) => {
     var percent;
     var nationalPercent = Math.round( meta.national_average * 1000 ) / 10;
+    if ( point.value === null ) {
```
contolini: @marteki The logic here isn't very intuitive. Both the county view and the metro view have map locations that need to show an "insufficient data" tooltip. But for metros we want to show the metro name and the national average along with an insufficient data message. For counties we only want to show the insufficient data message.

In metro view, the API returns a list of all metros and insufficient areas are set in cfpb-chart-builder as having a value of `-1` to indicate there's no data but we still need to show the location on the map.

In county view, the API return a list of *only* counties with sufficient data. So cfpb-chart-builder sets every pixel of the map that's not a county to have the value `null`.

This `renderTooltip` function is location agnostic and doesn't know whether we're in county or metro view so it just checks for `null` (which only appears in county view) or `< 0` (which only appears in metro view) and sets tooltips accordingly.

State view has no insufficient data.

In the future if we added a new location type (census tracts or something) the tooltip would work out of the box based on the data passed to it from the API.

marteki: > But for metros we want to show the metro name and the national average along with an insufficient data message.

Confirmed with design that we don't want to show national average unless we have an actual average to compare it to. But changing that is out of the scope of this PR, maybe.

PR #3356, opened 2017-09-18
Force known New Relic violations to be strings when caching

willbarton:
This PR forces `known_violations` to be strings when writing to our cache file. This should fix caching of New Relic alerts.


willbarton:
I'm going to merge this to quiet our alerts a little bit.

PR #3359, opened 2017-09-18
Disable eRegs caches when working locally

chosak:
The satellite eRegs project [regulations-site](https://github.com/cfpb/regulations-site) depends on the definition of three Django caches: `default`, `eregs_longterm_cache`, and `api_cache`. Currently these are defined for any settings file and include use of the local `/tmp` directory. This
can cause confusion when debugging locally, especially when iterating on changes to eRegs content.

This PR moves those cache definitions to `cfgov.settings.production` and instead defines dummy caches (using [`django.core.cache.backends.dummy.DummyCache](https://docs.djangoproject.com/en/1.8/topics/cache/#dummy-caching-for-development)) in `cfgov.settings.local`. This allows the eRegs code to work properly without actually relying on any caches locally, but still contuining to use those caches in production.

## Changes

- Modifies eRegs caching so that it uses memory/disk only in production, and defines dummy caches for use in local development.

## Testing

1. Install eRegs projects (regulations-core and regulations-site) from `requirements/optional-public.txt`).
2. Create a new local database (`./drop-db.sh && ./create-initial-db.sh && ./initial-data.sh`).
3. Load some eRegs content from local JSON, e.g. `cfgov/manage.py regcore_import_reg -s ../regulations-stub/stub -r 1040`.
4. Run a local server with `./runserver.sh` then navigate to a local page, e.g. http://localhost:8000/eregulations/1040-1/2017-14225#1040-1.
5. Make a change to your local JSON, and reload using `regcore_import_reg` again.
6. Reload the page and notice that it reflects your changes.

## Notes

- As the production servers are now backed by Akamai, is there any reason to maintain these eRegs-specific caches?

## Checklist


cfgov/cfgov/settings/local.py
```diff
@@ -45,3 +45,12 @@
...
 MIDDLEWARE_CLASSES += CSP_MIDDLEWARE_CLASSES
+
+# Define caches necessary for eRegs.
+# Disable caching when working locally.
+CACHES = {
+    k: {
+        'BACKEND': 'django.core.cache.backends.dummy.DummyCache',
+        'TIMEOUT': 0,
+    } for k in ('default', 'eregs_longterm_cache', 'api_cache')
+}
```
willbarton: With this in `settings/local.py` and not `settings/base.py`, will this break running the regulations-site and regulations-core unittests with `tox -e optional`?

willbarton:
> As the production servers are now backed by Akamai

Aren't they *fronted* by Akamai? 😜

cfgov/cfgov/settings/local.py
chosak: I thought of that, but `cfgov.settings.test` imports from `cfgov.settings.local`, so we should be okay. More generally, I thought that we were deprecating pursuit of the optional `tox` configs?

willbarton: Oh, good points!

PR #3322, opened 2017-09-08
Post Django errors to Github

richaagarwal:
- Adds a logging handler to intercept any logging statements of level `ERROR`
- Posts them to Github Enterprise as issues, as we are doing with Jenkins failures and New Relic incidents

PR #3367, opened 2017-09-19
Add github3 dependency to base.txt

richaagarwal:
Previously this was only in `scripts.txt` because it was only called from Jenkins, but with the introduction of https://github.com/cfpb/cfgov-refresh/pull/3322 it's needed in cfgov as well.  I didn't catch it because I already had the lib installed locally as well as in `test.txt` :(

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,7 +1,7 @@
 codecov:
   branch: master
   notify:
-    require_ci_to_pass: yes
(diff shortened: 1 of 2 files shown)


richaagarwal:
FYI @willbarton I modified the codecov setting to not require CI to pass, since it seemed like that's the only thing preventing it from running in parallel.

PR #3373, opened 2017-09-19
Fix New Relic alerts known violation caching

willbarton:
Yet another attempt to fix the New Relic violation caching. This PR adjusts the way the cache file is written and the way it is read to ensure correctly formatted data all the way through the process.

The New Relic alert polling job is currently running (successfully now!) from this branch.


willbarton:
This is (finally) successfully caching violations that have been reported. I'm going to merge this, as it is running successfully.

PR #3380, opened 2017-09-20
Tweak Codecov, require PRs to be fully tested

willbarton:
This PR changes our Codecov config to rely on Codecov's "patch" status instead of "project" for reporting frontend and backend and restores simple absolute coverage checking of the "project".

Per the documentation, "patch" [only measures coverage of the changes in a PR, not the entire codebase](https://github.com/codecov/support/wiki/Patch-Status#require-fully-tested-pull-requests). This seems more immediately useful anyway. This way we can also require PRs to be fully tested by requiring that they have 100% patch coverage.

I would recommend we make the frontend and backend "patch" coverage required for PRs, and abandon the idea of tracking PRs by absolute "project" coverage numbers. "Project" has seemed a bit wonky, with coverage decreasing by hundredths of a percent for seemingly no reason whatsoever while "patch" is at 100%.

--

This PR has changed quite a bit since I opened it, so @higs4281 I'm going to dismiss your review.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -7,16 +7,15 @@ comment: off
 
 coverage:
   status:
-    project:
-      frontend:
-        threshold: 0%
-        flags: frontend
-      backend:
-        threshold: 0%
-        flags: backend
     patch:
-      default:
-        threshold: 0%
+      frontend:
+        flags: frontend
+        target: 100%
+        only_pulls: true
+      backend:
+        flags: backend
+        target: 100%
+        only_pulls: true
 
 flags:
   backend:


richaagarwal:
If the CI doesn't pass, those PRs shouldn't be getting merged in though, right?  It's unclear to me how that would be impacting codecov on subsequent PRs, since the base commit should be the last thing that got merged in.

higs4281:
I want to understand how codecov comes up with these coverage reduction numbers.

willbarton:
@richaagarwal you are correct. I have a new working theory that the commit that messed up Codecov, b7bd68f, was a squash-and-merge, and we don't have Travis automatically running on master to submit coverage for commits like that.

willbarton:
I think I'm going to set this aside for now, since we've figured out a squash and merge was our problem. We might revisit it in the future.

PR #3385, opened 2017-09-20
Update the HTML conditionals to match the main page

JeffreyMFarley:
While verifying the new Complaint Search will work in IE9, I noticed that the logic for detecting the IE version is updated in `jinja2/v1/_layouts` but not in the legacy templates

This update (or something similar) is required for IE9 for CCDB5


Scotchester:
@JeffreyMFarley Which template are you matching this to? The `base.html` used by Wagtail pages is still significantly different from the results of this PR: https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_layouts/base.html#L20-L22

We dropped support for IE 6 and 7 a while ago.

(I see another potential bug in the file I linked above, though... seems like IE 9 would match both L21 and L22. I'll check that out.)

Scotchester:
Well, IE9 doesn't seem to choke on it (it uses the first one, so it gets the `lt-ie10` class). But it clearly seems to be a mistake.

I'm not sure we even need .lt-ie10 classes anyway. [This seems to be the only place we use them](https://github.com/cfpb/capital-framework/blob/4cbbaf1baa2144d46a5b0e2033a593eac92847a2/src/cf-layout/src/organisms/featured-content-module.less#L38), and we could probably rework it slightly to avoid that.

sephcoster:
@Scotchester are you suggesting we update this PR with these changes or is this an issue that would need to be changed for both sets of templates, perhaps in a separate PR?

If there are no issue with the current code I'd like to merge and move forward but wanted to confirm we're good to go. Thanks.

Scotchester:
I don't seen any potential issues with the update in this PR, I was just confused by it not actually matching the version in `jinja2/v1/_layouts` as stated in the description.

The other issue is something we should fix at some point, but since it doesn't appear to be actively harmful, we'll get around to it whenever.

PR #3390, opened 2017-09-22
Further whitelist additions to support Google Tag Manager debugger

rosskarchner:
Resolves some CSP errors that fire when you enable the debugger:

screen shot 2017-09-22 at 9 17 32 am

## Additions

- Allow fonts from data: URL's and fonts.google.com
- allow images from gstatic.com
- allow stylesheets from 'fonts.googleapis.com'

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -470,7 +470,7 @@ CSP_STYLE_SRC = (
     'tagmanager.google.com',
     'optimize.google.com',
     'api.mapbox.com',
-    'fonts.googleapis.com')
+    'fonts.googleapis.com',)
 
 # These specify valid image sources
 CSP_IMG_SRC = (
@@ -505,7 +505,7 @@ CSP_FRAME_SRC = (
     'universal.iperceptions.com')
 
 # These specify where we allow fonts to come from
-CSP_FONT_SRC = ("'self'",'data:', 'fast.fonts.net', 'fonts.google.com')
+CSP_FONT_SRC = ("'self'", "data:", "fast.fonts.net", "fonts.google.com")
 
 # These specify hosts we can make (potentially) cross-domain AJAX requests to.
 CSP_CONNECT_SRC = ("'self'",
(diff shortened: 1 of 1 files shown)


cfgov/cfgov/settings/base.py
```diff
@@ -469,12 +469,15 @@
     'fast.fonts.net',
     'tagmanager.google.com',
     'optimize.google.com',
-    'api.mapbox.com')
+    'api.mapbox.com',
+    'fonts.googleapis.com')
```
Scotchester: Trivial: Include trailing comma to avoid modifying this line next time we add a new one :)

Bonus points: alphabetize the list by TLD, then subdomain.

cfgov/cfgov/settings/base.py
```diff
@@ -502,7 +505,7 @@
     'universal.iperceptions.com')
 
 # These specify where we allow fonts to come from
-CSP_FONT_SRC = ("'self'", 'fast.fonts.net')
+CSP_FONT_SRC = ("'self'",'data:', 'fast.fonts.net', 'fonts.google.com')
```
Scotchester: Trivial: missing space between the first two items in this list.

Scotchester: Hmm... Is the first one double-quoted because the string value has to include the literal single quote characters? `'self'`?

If so, would it be better for consistency to double-quote the rest of the values in this list? Does flake8 care about this at all?

cfgov/cfgov/settings/base.py
rosskarchner: I agree that someone ought to alphabetize these some day ;)

cfgov/cfgov/settings/base.py
rosskarchner: Correct on the literal quotes being neccessary for self. I've double-quoted the whole list now

willbarton: > Does flake8 care about this at all? 

We don't lint the settings files currently... for reasons which I think might be obvious ;)

Scotchester: Hah!

PR #3399, opened 2017-09-25
Update adult financial education urls

jimmynotjim:
Adult fin-ed has been updated and released as Wagtail pages. This change
requires removing the old template url and redirecting the old url to
the new one. It also requires updating the navigation vars to avoid
unnecessary redirects.

## Changes

- Updated the urls.py to redirect the old url instead of pointing to the template.
- Updated the mega menu vars to use the new url

## Testing

1. navigate to `http://localhost:8000/adult-financial-education` and it should redirect.
2. Use the main nav to navigate to "Practitioner Resources" and then "Financial Education for Adults" 

## Checklist


cfgov/cfgov/urls.py
```diff
@@ -101,8 +101,10 @@
             permanent=True)),
     url(r'^know-before-you-owe/',
         include(SheerSite('know-before-you-owe').urls)),
-    url(r'^adult-financial-education/$', TemplateView.as_view(
-        template_name='/adult-financial-education/index.html')),
+    url(r'^adult-financial-education/(?P<path>.*)$',
+        RedirectView.as_view(
+            url='/practitioner-resources/adult-financial-education/%(path)s',
+            permanent=True)),
```
chosak: Could we implement this as an Apache redirect instead of here in `urls.py`? This approach might lead to this file getting cluttered up over time.

sebworks:
@jimmynotjim, you might as well remove the templates as well.

cfgov/cfgov/urls.py
willbarton: Alternatively we could implement this as a Wagtail redirect, I believe given the path.

I think no matter what we're going to have clutter, be it here, Apache, or in Wagtail. Apache would probably be the most performant place for it though.

jimmynotjim: I had a laugh about that as well, but I've [moved it to the Apache config](https://github.cfpb.gov/cfpb-infra/ansible-consumer-finance/pull/144) for good measure.

jimmynotjim:
@chosak @sebworks ready for another look.

cfgov/cfgov/urls.py
chosak: ☝️ @jimmynotjim FYI I edited your comment to remove reference to GHE which shouldn't be in public repos.

sebworks:
nice work @jimmynotjim, 👍

PR #3417, opened 2017-09-28
Optimize FWB JS and fix an IE error

jimmynotjim:
The script had trouble keeping the state of the form correctly saved
when returning to the page after submission. To better track each
individual question the script sets up an object and tracks the radio states in
relation to their parent fieldset. In addition I made some optimizations
and broke code up into smaller testable bites where possible.

## Changes

- Added onload checker to fix issue with IE not immediately recognizing inputs as checked when returning to the page after submission
- Modularized the functionality as much as possible to make testing easier
- Grouped radios by parent fieldset to simplify completion checks

## Testing

1. Open `10.0.2.2:8000/consumer-tools/financial-well-being/` in IE 9
2. Verify the submit button is disabled
3. Randomly select a radio from each fieldset group (I just use the left hand one for all of them)
4. Submit
5. Return to page using browser back button, ensure your choices are still selected and the submit button is not disabled.
6. run `gulp test` and ensure everything passes.

## Todos

- ~~We still need to have a discussion around satellite apps in cfgov-refresh.~~
- ~~Needs unit tests (wanted to get some feedback first).~~
- Could probably stand to use promises instead of one function triggering the next.
- ~~ES6-ify all the things.~~

## Checklist

### DIFF
--- a/cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
+++ b/cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
@@ -1,83 +1,172 @@
-/* eslint no-global-assign: "off" */
 'use strict';
 
-var Analytics = require( '../../modules/Analytics' );
+const Analytics = require( '../../modules/Analytics' );
 
+/**
+ * Initialize the questionnaire
+ */
 function init() {
+  const questionsDom = document.querySelectorAll( '#quiz-form fieldset' );
+  const radiosDom = document.querySelectorAll( '#quiz-form [type="radio"]' );
+  const submitDom = document.querySelector( '#submit-quiz' );
+  const questionStates = {};
 
-  var totalQuestions = 12;
-  var quizComplete = false;
-  var questionValues = {};
-  var radioButtons = document.querySelectorAll( '[type="radio"]' );
-  var submitButton = document.querySelector( '#submit-quiz' );
+  /**
+   * Checks the selection status of a group of radio inputs,
+   * belonging to a parent fieldset
+   * @param {HTMLNode} childRadios - An array of DOM elements.
+   * @returns {boolean} The status of the group of inputs
+   */
+  function checkQuestionState( childRadios ) {
+    let radioIsChecked = false;
+    const radiosLength = childRadios.length;
 
-  function allQuestionsCompleted() {
-    return Object.keys( questionValues ).length === totalQuestions;
-  }
-
-  function enableSubmit() {
-    submitButton.title = 'Get your score';
-
-    submitButton.classList.remove( 'a-btn__disabled' );
-  }
-
-  function handleRadio( input ) {
-    if ( input.name && input.checked ) {
-
-      questionValues[input.name] = input.value;
+    for ( let i = 0; i < radiosLength; i++ ) {
+      if ( childRadios[i].checked ) {
+        radioIsChecked = true;
+      }
     }
 
-    quizComplete = allQuestionsCompleted();
+    return radioIsChecked;
+  }
 
-    if ( quizComplete ) {
+  /**
+   * Checks the status of each fieldset to determine if
+   * the user has completed the form
+   * @returns {boolean} The completion status of the form
+   */
+  function checkFormCompletion() {
+    let formCompleted = true;
+
+    Object.keys( questionStates ).forEach( key => {
+      if ( !questionStates[key] ) {
+        formCompleted = false;
+      }
+    } );
+
+    return formCompleted;
+  }
+
+  /**
+   * Event handler to prevent clicking the submit button before
+   * the form is completed
+   * @param {Object} event - The event object for the click event.
+   */
+  function handleDisabledSubmit( event ) {
+    event.preventDefault();
+  }
+
+
+  /**
+   * Sets the disabled state of the submit button
+   */
+  function disableSubmit() {
+    submitDom.disabled = true;
+    submitDom.addEventListener( 'click', handleDisabledSubmit );
+  }
+
+  /**
+   * Sets the enabled state of the submit button
+   */
+  function enableSubmit() {
+    submitDom.title = 'Get your score';
+    submitDom.disabled = false;
+    submitDom.removeEventListener( 'click', handleDisabledSubmit );
+  }
+
+
+  /**
+   * Updates the status of the fieldset in the data store
+   * and checks if the form has been completed
+   * @param {HTMLNode} input - A DOM element
+   */
+  function handleRadio( input ) {
+    if ( input.name && input.checked ) {
+      questionStates[input.name] = true;
+    }
+
+    if ( checkFormCompletion() ) {
       enableSubmit();
     }
   }
 
+  /**
+   * Sends the user interaction to Analytics
+   * @param {string} action - A GTM data attribute
+   * @param {string} label - A GTM data attribute
+   * @param {string} category - A GTM data attribute
+   */
   function sendEvent( action, label, category ) {
-    var eventData = Analytics.getDataLayerOptions( action, label, category );
+    const eventData = Analytics.getDataLayerOptions( action, label, category );
     Analytics.sendEvent( eventData );
   }
 
-  submitButton.addEventListener( 'click', function( event ) {
-    // Confirm that all questions are answered
-    if ( !quizComplete ) {
-      // If not, block the form submission action
-      return event.preventDefault();
+  /**
+   * Determines the state of Analytics and either passes the data-sets
+   * or waits for Analytics to report readiness
+   * @param {HTMLNode} input - A DOM element
+   */
+  function handleRadioAnalytics( input ) {
+    const action = input.getAttribute( 'data-gtm-action' );
+    const label = input.getAttribute( 'data-gtm-label' );
+    const category = input.getAttribute( 'data-gtm-category' );
+
+    if ( Analytics.tagManagerIsLoaded ) {
+      sendEvent( action, label, category );
+    } else {
+      /* istanbul ignore next */
+      Analytics.addEventListener( 'gtmLoaded', sendEvent );
+    }
+  }
+
+  /**
+   * Event handler to watch user interaction on each input
+   */
+  function setUpListeners() {
+    [].forEach.call( radiosDom, function( el ) {
+      el.addEventListener( 'click', function( event ) {
+        const input = event.target;
+
+        handleRadio( input );
+        handleRadioAnalytics( input );
+      } );
+    } );
+  }
+
+  /**
+   * Determines the state of the form and modifies the UI appropriately
+   */
+  function setUpUI() {
+    if ( checkFormCompletion() ) {
+      // Enable submit button initially if the answers are completed
+      enableSubmit();
+    } else {
+      // Otherwise disable submit button initially
+      disableSubmit();
     }
 
-    // Otherise, allow the form submission to go through as normal
-    return null;
-  } );
+    setUpListeners();
+  }
 
-  [].forEach.call( radioButtons, function( el ) {
-    el.addEventListener( 'click', function radioButtonClickHandler( event ) {
-      var input = event.target;
-      handleRadio( input );
+  /**
+   * Initializes a data store from the initial state of each fieldset
+   */
+  function setUpData() {
+    const questionsLength = questionsDom.length;
 
-      var action = input.getAttribute( 'data-gtm-action' );
-      var label = input.getAttribute( 'data-gtm-label' );
-      var category = input.getAttribute( 'data-gtm-category' );
+    for ( let i = 0; i < questionsLength; i++ ) {
+      const questionID = questionsDom[i].id;
+      const childRadios = document.querySelectorAll(
+        '#' + questionID + ' input'
+      );
 
-      if ( Analytics.tagManagerIsLoaded ) {
-        sendEvent( action, label, category );
-      } else {
-        /* istanbul ignore next */
-        Analytics.addEventListener( 'gtmLoaded', sendEvent );
-      }
-    } );
-  } );
+      questionStates[questionID] = checkQuestionState( childRadios );
+    }
 
-  // Disable submit button initially
-  submitButton.classList.add( 'a-btn__disabled' );
-  submitButton.title = 'Please answer all questions to get your score';
+    setUpUI();
+  }
 
-  // Look at all radios on coming back to the page or on reload
-  [].forEach.call( radioButtons, function( el ) {
-    handleRadio( el );
-  } );
-
+  setUpData();
 }
 
 module.exports = { init: init };
-
--- a/cfgov/unprocessed/js/apps/financial-well-being/fwb-results.js
+++ b/cfgov/unprocessed/js/apps/financial-well-being/fwb-results.js
@@ -1,72 +1,112 @@
 /* eslint no-extra-semi: "off" */
 'use strict';
 
-var Analytics = require( '../../modules/Analytics' );
-var Expandable = require( '../../organisms/Expandable' );
+const Analytics = require( '../../modules/Analytics' );
+const Expandable = require( '../../organisms/Expandable' );
 
-var expandableDom = document.querySelectorAll( '.content .o-expandable' );
-var expandable;
+const expandableDom = document.querySelectorAll( '.content .o-expandable' );
+let expandable;
+
 if ( expandableDom ) {
-  for ( var i = 0, len = expandableDom.length; i < len; i++ ) {
+  for ( let i = 0, len = expandableDom.length; i < len; i++ ) {
     expandable = new Expandable( expandableDom[i] );
     expandable.init();
   }
 }
 
+/**
+ * Initialize the results interactions
+ */
 function init() {
 
-  var toggleButtons = document.querySelectorAll(
+  const buttonsDom = document.querySelectorAll(
     '.comparison-chart_toggle-button'
   );
 
+  /**
+   * Changes the visibility of the results by category
+   * based on user input
+   * @param {string} category - The category to display
+   */
   function switchComparisons( category ) {
-    var allCategories = document.querySelectorAll( '.comparison_data-point' );
-    var selectedButtonClass = 'comparison-chart_toggle-button__selected';
+    const allCategories = document.querySelectorAll( '.comparison_data-point' );
+    const showCategory = document.querySelectorAll(
+      '[class^="comparison_data-point ' + category + '"]'
+    );
+    const selectedButton = document.querySelector(
+      '[data-compare-by="' + category + '"]'
+    );
+    const selectedButtonClass = 'comparison-chart_toggle-button__selected';
+    const hiddenClass = 'u-hidden';
 
     // Hide all categories ...
     [].forEach.call( allCategories, function( el ) {
-      el.style.display = 'none';
+      el.classList.add( hiddenClass );
     } );
     // ... and deselect all toggle buttons ...
-    [].forEach.call( toggleButtons, function( el ) {
+    [].forEach.call( buttonsDom, function( el ) {
       el.classList.remove( selectedButtonClass );
     } );
     // ... so that we can show only the right category data ...
-    var showCategory = document.querySelectorAll( '[class^="comparison_data-point ' + category + '"]' );
-
     [].forEach.call( showCategory, function( el ) {
-      el.style.display = 'inline-block';
+      el.classList.remove( hiddenClass );
     } );
     // ... and then highlight the correct button.
-    var selectedButton = document.querySelector( '[data-compare-by="' + category + '"]' );
-
     selectedButton.classList.add( selectedButtonClass );
   }
 
+  /**
+   * Sends the user interaction to Analytics
+   * @param {string} action - A GTM data attribute
+   * @param {string} label - A GTM data attribute
+   * @param {string} category - A GTM data attribute
+   */
   function sendEvent( action, label, category ) {
-    var eventData = Analytics.getDataLayerOptions( action, label, category );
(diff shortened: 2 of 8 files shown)


cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
```diff
@@ -1,83 +1,169 @@
...
+  }
+
+  /**
+   * Event handler to prevent clicking the submit button before
+   * the form is completed
+   * @param {MouseEvent} event The event object for the mousedown event.
+   */
+  function handleDisabledSubmit( event ) {
+    event.preventDefault();
+  }
```
Scotchester: Is it really necessary to break this out into its own function? Is it essential for testing it the right way or something?

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
```diff
@@ -1,83 +1,169 @@
...
-    submitButton.classList.remove( 'a-btn__disabled' );
+    submitButton.removeAttribute( 'disabled' );
+    submitButton.removeEventListener( 'click', handleDisabledSubmit );
   }
 
+
+  /**
+   * Updates the status of the fieldset in the data store
+   * and checks if the form has been completed
+   * @param {HTMLNode} input A dom element
```
Scotchester: Trivial: "DOM".

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
```diff
@@ -1,83 +1,169 @@
...
+
+    return radioIsChecked;
+  }
+
+  /**
+   * Checks the status of each fieldset to determine if
+   * the user has completed the form
+   * @returns {boolean} The completion status of the form
+   */
+  function checkformCompletion() {
```
Scotchester: Should probably have a capital `F`.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
```diff
@@ -1,83 +1,169 @@
...
+      }
+    } );
+
+    return formCompleted;
+  }
+
+  /**
+   * Event handler to prevent clicking the submit button before
+   * the form is completed
+   * @param {MouseEvent} event The event object for the mousedown event.
```
Scotchester: Should this say "click event"?

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
```diff
@@ -1,83 +1,169 @@
...
+    if ( checkformCompletion() ) {
       enableSubmit();
     }
   }
 
+  /**
+   * Sends the user interaction to Analytics
+   * @param {string} action A GTM data-set
+   * @param {string} label A GTM data-set
+   * @param {string} category A GTM data-set
```
Scotchester: "data-set" sounds kinda weird here, since they're all just single strings

cfgov/unprocessed/js/apps/financial-well-being/fwb-results.js
```diff
@@ -1,72 +1,112 @@
...
-var Analytics = require( '../../modules/Analytics' );
-var Expandable = require( '../../organisms/Expandable' );
+const Analytics = require( '../../modules/Analytics' );
+const Expandable = require( '../../organisms/Expandable' );
 
-var expandableDom = document.querySelectorAll( '.content .o-expandable' );
-var expandable;
+const expandableDom = document.querySelectorAll( '.content .o-expandable' );
+let expandable;
 if ( expandableDom ) {
```
Scotchester: Trivial: Add a blank line here.

cfgov/unprocessed/js/apps/financial-well-being/fwb-results.js
```diff
@@ -1,72 +1,112 @@
...
+ */
 function init() {
 
-  var toggleButtons = document.querySelectorAll(
+  const toggleButtons = document.querySelectorAll(
     '.comparison-chart_toggle-button'
   );
 
+  /**
+   * Changes the viewability of the results by category
```
Scotchester: Visibility? :)

cfgov/unprocessed/js/apps/financial-well-being/fwb-results.js
```diff
@@ -1,72 +1,112 @@
...
+    const selectedButton = document.querySelector(
+      '[data-compare-by="' + category + '"]'
+    );
+    const selectedButtonClass = 'comparison-chart_toggle-button__selected';
+    const hiddenClass = 'u-hidden';
 
     // Hide all categories ...
     [].forEach.call( allCategories, function( el ) {
-      el.style.display = 'none';
+      el.classList.add( hiddenClass );
```
Scotchester: Don't we need the `className` fallback for browsers that don't support `classList` (IE9)? (Here and 3 times below).

cfgov/unprocessed/js/apps/financial-well-being/fwb-results.js
```diff
@@ -1,72 +1,112 @@
...
     } );
-  } );
+  }
+
+  /**
+   * Sets up the UI for the default age category
+   */
+  function setUpUI() {
+    switchComparisons( 'age' );
+
```
Scotchester: Trivial: blank line seems unnecessary here.

cfgov/unprocessed/js/routes/consumer-tools/financial-well-being/results/index.js
```diff
@@ -1,5 +1,5 @@
```
Scotchester: Trivial: could remove blank line here.

cfgov/wellbeing/jinja2/wellbeing/results.html
```diff
@@ -113,12 +113,14 @@
             text-align: right;
         }
         .comparison-chart_list dd {
-            display: inline-block;
             width: 65%;
             border-bottom: 1px solid #b4b5b6;
             position: relative;
             bottom: 0.25em;
         }
+        .comparison-chart_list dd:not(.u-hidden) {
```
Scotchester: Any concern about IE8 not supporting `:not`?

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
```diff
@@ -1,83 +1,169 @@
...
+  /**
+   * Checks the selection status of a group of radio inputs,
+   * belonging to a parent fieldset
+   * @param {HTMLNode} questionRadios An array of dom elements.
+   * @returns {boolean} The status of the group of inputs
+   */
+  function checkQuestionState( questionRadios ) {
+    let radioIsChecked = false;
+
+    for ( let i = 0, l = questionRadios.length; i < l; i++ ) {
```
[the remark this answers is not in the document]
jimmynotjim: Yep, for the `i < l` condition

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
jimmynotjim: Yes because we stop listening for it below.

cfgov/unprocessed/js/apps/financial-well-being/fwb-results.js
jimmynotjim: Nope, we have a polyfill already in place.

cfgov/wellbeing/jinja2/wellbeing/results.html
jimmynotjim: IE8 won't get it because it's added by JS and IE8 isn't getting JS anyway.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
jimmynotjim: Isn't data-set just another name for data attr?

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
```diff
@@ -1,83 +1,169 @@
...
 function init() {
+  const questionElems = document.querySelectorAll( '#quiz-form fieldset' );
+  const radioButtons = document.querySelectorAll( '#quiz-form [type="radio"]' );
+  const submitButton = document.querySelector( '#submit-quiz' );
+  const questionStates = {};
+
+  /**
+   * Checks the selection status of a group of radio inputs,
+   * belonging to a parent fieldset
+   * @param {HTMLNode} questionRadios An array of dom elements.
```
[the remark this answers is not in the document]
jimmynotjim: I don't think so. I modeled these after the other JS files we have here.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
```diff
@@ -1,83 +1,169 @@
...
 'use strict';
 
-var Analytics = require( '../../modules/Analytics' );
+const Analytics = require( '../../modules/Analytics' );
 
+/**
+ * Initialize the questionnaire
+ */
 function init() {
+  const questionElems = document.querySelectorAll( '#quiz-form fieldset' );
```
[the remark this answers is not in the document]
jimmynotjim: For clarity because there was also `questionStates`. I actually noticed today that we use `thingDom` everywhere else, so I'm going to update to match that.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
```diff
@@ -1,83 +1,169 @@
...
+  const radioButtons = document.querySelectorAll( '#quiz-form [type="radio"]' );
```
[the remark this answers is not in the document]
jimmynotjim: In case we ever have another form on here. I know it's unlikely but scoping that widely seemed destined for trouble.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
jimmynotjim: Good catch, fixed.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
jimmynotjim: Yeah, that makes more sense.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
jimmynotjim: Fixed.

cfgov/wellbeing/jinja2/wellbeing/results.html
jimmynotjim: Realizing now this still wouldn't work. Added conditional `lt-i89` class to cover it.

cfgov/unprocessed/js/apps/financial-well-being/fwb-results.js
```diff
@@ -1,72 +1,112 @@
...
+    const selectedButtonClass = 'comparison-chart_toggle-button__selected';
+    const hiddenClass = 'u-hidden';
 
     // Hide all categories ...
     [].forEach.call( allCategories, function( el ) {
-      el.style.display = 'none';
+      el.classList.add( hiddenClass );
     } );
     // ... and deselect all toggle buttons ...
-    [].forEach.call( toggleButtons, function( el ) {
```
sebworks: It might be worth int to have a local forEach utility function.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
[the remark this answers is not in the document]
Scotchester: I mean, why not just do `i < questionRadios.length`? I should probably know the answer to this, but I can't remember. Thanks for indulging me :)

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
anselmbradford: I guess technically dataset is all the data- attributes together https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset

Scotchester: [`dataset`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) in JS is the collection of all `data-` attributes on an element, but these are just individual pieces of data.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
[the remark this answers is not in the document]
anselmbradford: Each loop you are re-calculating the length of `questionRadios`. By setting the length to a variable you're storing the length and just referencing it every loop.

cfgov/unprocessed/js/routes/consumer-tools/financial-well-being/results/index.js
anselmbradford: Generally I think we have a newline after 'use strict' and the rest of the code since it's really just boilerplate and not part of the custom code in the module.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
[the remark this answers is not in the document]
sebworks: Please change the  `l = questionRadios.length` to `questionRadiosLength`. It's more verbose but it makes the code easier to read.

sebworks:
great work, LGTM.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
```diff
@@ -1,83 +1,169 @@
...
-    return Object.keys( questionValues ).length === totalQuestions;
+  /**
+   * Checks the status of each fieldset to determine if
+   * the user has completed the form
+   * @returns {boolean} The completion status of the form
+   */
+  function checkFormCompletion() {
+    let formCompleted = true;
+
+    Object.keys( questionStates ).forEach( function( key ) {
```
sebworks: trivial: You could use double arrow here.

cfgov/unprocessed/js/apps/financial-well-being/fwb-results.js
Scotchester: Ah, I see. Looking at the source, it seems like we should update that, since we manually pasted it in and it's now several years out of date. Perhaps we should install it via npm? (It seems poorly updated on the actual npm registry vs. what's in the repo, but we could reference the Git repo.)

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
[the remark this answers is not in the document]
Scotchester: @anselmbradford Ah, right. I think I keep forgetting this because it's obviously a poor implementation decision that this is the case :trollface: (JK; I'm sure there's a reason it's not stupid.)

cfgov/unprocessed/js/routes/consumer-tools/financial-well-being/results/index.js
Scotchester: Heh, I deleted the comment I made before seeing yours appear because I couldn't remember why I made it. I think a later change Jimmy made may have made it appear in the wrong place?

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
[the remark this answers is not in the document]
Scotchester: I'm not a fan of that convention, but if it's what we're already doing...

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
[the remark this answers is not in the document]
Scotchester: It's probably of negligible impact in this case, but just as a question of best practices, shouldn't we discourage taking the performance hit of using more specific selectors than absolutely necessary?

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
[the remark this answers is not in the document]
Scotchester: The JSDoc docs disagree: http://usejsdoc.org/about-getting-started.html#adding-documentation-comments-to-your-code

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
```diff
@@ -1,83 +1,169 @@
...
-    submitButton.title = 'Get your score';
 
-    submitButton.classList.remove( 'a-btn__disabled' );
+  /**
+   * Sets the disabled state of the submit button
+   */
+  function disableSubmit() {
+    submitDom.disabled = true;
+    submitDom.addEventListener( 'click', handleDisabledSubmit );
   }
```
Scotchester: Do we not want to revert the `title` back to the default in this function?

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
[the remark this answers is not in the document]
jimmynotjim: @Scotchester it's not so much implementation decision as that's just how a lookup works. If you were to do a lookup throughout a loop in any language you'd have the same thing happening. Think of it like saving a DOM query to a variable vs looking it up every time you need it.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
jimmynotjim: This is only triggered on page load so there wouldn't be a case where the `title` is changed to the enabled version and then the button is disabled again since the form is all radios and once you've chosen one, you can't then clear an entire `fieldset` without refreshing the page and thus kicking off the form status function anew.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
jimmynotjim: Oh good call.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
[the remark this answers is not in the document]
Scotchester: @jimmynotjim Yes, I get that part :P I was thinking it should somehow be smart enough to cache the non-counter pieces of the expression, but I suppose it's possible you want to change that within the code that the loops executes.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
jimmynotjim: Ah right, thanks.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
Scotchester: Good point.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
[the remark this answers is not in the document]
jimmynotjim: Yes, but we also should be using data attributes for hooks like these. I felt like I'd already modified so much I really needed to stop at some point and leave things for a future PR.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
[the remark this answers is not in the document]
jimmynotjim: I wouldn't want to see it make that kind of assumption, especially if items were added to the object within the loop.

cfgov/unprocessed/js/apps/financial-well-being/fwb-results.js
jimmynotjim: I can get onboard with that.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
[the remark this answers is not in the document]
jimmynotjim: @sebworks are we sure this is easier to read? https://github.com/cfpb/cfgov-refresh/pull/3417/commits/c0b85a048f580f824cc6bbbc3db62af1cb19528a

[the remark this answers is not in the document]
sebworks: lol, it's gigantic, try `qLen` and `rLen`.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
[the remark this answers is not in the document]
jimmynotjim: After speaking to @anselmbradford it looks like we have a mix of usage. Updating to match the JSDoc docs.

cfgov/unprocessed/js/apps/financial-well-being/fwb-results.js
jimmynotjim: I like that. You mind if I push that to a future PR? This one is getting kinda unruly.

cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js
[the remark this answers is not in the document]
jimmynotjim: @sebworks how bout this? https://github.com/cfpb/cfgov-refresh/pull/3417/commits/2b40d3974fbd470a19d152eeb995ca76d1d13428 Even with a shorter variable name, it was still longer than 80 chars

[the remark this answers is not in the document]
sebworks: yep, that is money.

PR #3433, opened 2017-10-02
Mortgage performance: Prep scripts for pipeline use

higs4281:
This is almost entirely a rearrangement of code to serve the sequencing of a data loading
pipeline job. The main addition is a script to generate a CSV that can
be used to bulk-load county data into mysql, which is far faster than
inserting 350K records one at a time via the Django ORM.

## New
- data_research/scripts/generate_county_data_loading_csv  

The scripts have been run locally, and the result is on build.

This is the pipeline sequence, which can be run locally. It takes about 11 minutes of processing time on my laptop, but it does not load the results locally. That's a job Jenkins will handle.
```
cfgov/manage.py runscript generate_county_data_loading_csv
cfgov/manage.py runscript load_mortgage_aggregates
cfgov/manage.py runscript update_county_msa_meta
cfgov/manage.py runscript export_public_csvs
```

If you want to load the results locally, this should do it:
```
cd /tmp && curl -O http://files.consumerfinance.gov.s3.amazonaws.com/data/mortgage-performance/table_loads/latest/mortgagedata.sql.gz
gunzip -c /tmp/mortgagedata.sql.gz | mysql -uroot v1
```


higs4281:
@marteki Sorry about that. I just added a constant to the game to handle our new request to lop off the latest 3 months of data on our end. So this is expected, although I should have updated the instructions.

marteki:
Then a :+1: from me (for what's it worth).

PR #3435, opened 2017-10-02
Condense webpack-build

anselmbradford:
Minor condensing of webpack-build script to remove linter warnings.

## Changes

- Convert some functions to arrow functions in test.
- Reuse path variable.
- Remove unnecessary else statements.

## Testing

1. `gulp test:acceptance --fast --tags=@webpack` should pass.
2. Edit built `common.js` and change a `var` to a `const`
3. Re-run `gulp test:acceptance --fast --tags=@webpack`. It should fail.


test/browser_tests/cucumber/step_definitions/webpack-build.js
```diff
@@ -5,43 +5,39 @@ const expect = chai.expect;
...
+            const isFailure = transpileRegex.test( contents );
+            if ( isFailure ) {
+              reject( directoryMapKeys[i] + ' contains const or ()=>' );
+            }
+          }
+        );
+      }
+
+      // All files searched.
+      resolve();
```
sebworks: This will always resolve before the loop finishes because of `fs.readFile`, which is async.

anselmbradford: Ahh right, good catch. I re-added the `else` statement.  I can swap with `readFileSync` if you think that's preferable. I'm not sure.

PR #3438, opened 2017-10-03
Mortgage performance: Fix how archive list is sliced

higs4281:
Fixes glitch in how archive files are listed.

Would like to merge this for today's product demo.


higs4281:
merging trivial change for demo

PR #3452, opened 2017-10-05
Mortgage performance: load via SQL

higs4281:
Database-agnostic CSVs were attractive for bulk-loading quarterly updates of our source
data, but our mysql setups won't allow it. SQL dumps are our second choice -- almost as fast, but they are specific to our version of mysql and are fussier to create.

This PR switches our pipeline to SQL dumps, while preserving the option of creating CSVs, in case we get to Postgres sometime soon.

The main addition here is the `source_to_dump` script, which takes a source CSV, slices it and 
cleans up various data problems, then dumps it to an SQL file that can be bulk-loaded.  
Bulk-loading avoids slow insertions in our server environments using the Django ORM.

Some of the changed files here are json metadata files that are programmatically updated.  
There's also some code simplification, and new tests.

## Testing

You'll probably need to run `migrate` if you have recent production data loaded:

```
```

Then make sure your metadata is up to date by loading these fixtures:

```
python cfgov/manage.py loaddata mortgage_metadata.json
python cfgov/manage.py loaddata mortgage_constants.json
```

Now you can run the `source_to_dump` script locally, which pulls down the latest source file from S3, then cleans and translates it into an SQL dump. All you provide is the desired through-date for the data set.

For a short test-run, you can enter `2009-01-01` as the through-date, which will only process one year's worth of data and takes about a minute: 

```
python cfgov/manage.py runscript source_to_dump --script-args 2009-01-01
```
Or you could test the full run by providing our current through-date:

```
python cfgov/manage.py runscript source_to_dump --script-args 2017-03-01
```

The full run should take about 4 minutes.

Then you can wipe and reload the county mortgage data table in your local v1 database:

```
mysql -uroot v1 < /tmp/mp_countydata.sql
echo "SELECT COUNT(*) FROM data_research_countymortgagedata;" | mysql -uroot v1
```

If you ran the short version, the count should be 40,673.

For the full monty, the count should be 347,705.

**Note** This PR sat a few days and was overrun by the CSV loading issue, and it has been refactored. If you pulled in the former version, please delete it and get a fresh one because its history changed.


marteki:
When trying to run the `source_to_dump` script locally, I got the following error:

```
data_research.models.DoesNotExist: MortgageDataConstant matching query does not exist.
```

I consulted with @higs4281, and then ran `python cfgov/manage.py loaddata mortgage_constants.json` to grab the updated constants. I then re-ran `setup.py` to make sure that everything got applied properly, and tried again with the `source_to-dump` script. Threw this exception:

```
  File "cfgov/manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/django/core/management/__init__.py", line 346, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/django_extensions/management/email_notifications.py", line 63, in run_from_argv
    super(EmailNotificationCommand, self).run_from_argv(argv)
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/django/core/management/base.py", line 394, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/django_extensions/management/email_notifications.py", line 75, in execute
    super(EmailNotificationCommand, self).execute(*args, **options)
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/django/core/management/base.py", line 445, in execute
    output = self.handle(*args, **options)
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/django_extensions/management/utils.py", line 71, in inner
    ret = func(self, *args, **kwargs)
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/django_extensions/management/commands/runscript.py", line 172, in handle
    run_script(mod, *script_args)
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/django_extensions/management/commands/runscript.py", line 80, in run_script
    mod.run(*script_args)
  File "/Users/reedm/Repositories/cfpb/cfgov-refresh/cfgov/data_research/scripts/source_to_dump.py", line 136, in run
    create_dump(starting_date, through_date)
  File "/Users/reedm/Repositories/cfpb/cfgov-refresh/cfgov/data_research/scripts/source_to_dump.py", line 89, in create_dump
    sampling_date = parser.parse(row.get('date')).date()
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/dateutil/parser.py", line 720, in parse
    return DEFAULTPARSER.parse(timestr, **kwargs)
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/dateutil/parser.py", line 308, in parse
    res = self._parse(timestr, **kwargs)
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/dateutil/parser.py", line 356, in _parse
    l = _timelex.split(timestr)
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/dateutil/parser.py", line 150, in split
    return list(cls(s))
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/dateutil/parser.py", line 147, in next
    return self.__next__()  # Python 2.x support
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/dateutil/parser.py", line 141, in __next__
    token = self.get_token()
  File "/Users/reedm/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/dateutil/parser.py", line 72, in get_token
    nextchar = self.instream.read(1)
AttributeError: 'NoneType' object has no attribute 'read'
```

willbarton:
@higs4281 I'm running into the same problem as @marteki locally.

It looks like `settings.AWS_STORAGE_BUCKET_NAME` doesn't exist, because the `source_url` in `source_to_dump.py:89` is ending up as `http://None.s3.amazonaws.com/data/mortgage-performance/source/latest_county_delinquency.csv`

cfgov/data_research/mortgage_utilities/s3_utils.py
```diff
@@ -16,9 +16,11 @@
 S3_SECRET = os.getenv('AWS_S3_SECRET_ACCESS_KEY')
 BASE_BUCKET = settings.AWS_STORAGE_BUCKET_NAME
```
willbarton: `settings.AWS_STORAGE_BUCKET_NAME` doesn't exist, so `BASE_BUCKET` will be `None`, which will become part of the `PUBLIC_ACCESS_BASE` URL below.

cfgov/data_research/scripts/source_to_dump.py
```diff
@@ -0,0 +1,138 @@
...
+from data_research.models import MortgageDataConstant, County
+from data_research.mortgage_utilities.s3_utils import (
+    read_in_s3_csv, S3_SOURCE_BUCKET, S3_SOURCE_FILE)
+from data_research.mortgage_utilities.fips_meta import validate_fips
+from data_research.mortgage_utilities.sql_utils import (
+    assemble_insertions,
+    DROP_AND_CREATE_STRING,
+    UNLOCK_STRING)
+
+DUMP_SLUG = '/tmp/mp_countydata'
```
willbarton: Are the resulting files here used later? Would it be better to take an argument for a filename to output to, or generate a filename with `tempfile.mkstemp()` and return it?

higs4281: The next step in the pipeline will load this tmp file and then remove it.

willbarton: So could the pipeline provide the filename to the script? Seems like the script should be agnostic about where it puts the data, and that makes the pipeline fully responsible for the location and cleanup of the file, and we wouldn't have to manage the filename in multiple places in the future if needed.

cfgov/data_research/mortgage_utilities/s3_utils.py
higs4281: @willbarton  Thanks for finding the AWS settings bug.

I share your reservations about the handcrafted SQL dump, but I think it's our best current option until we switch to Postgres, which could occur before this pipeline ever runs. All better options are blocked by environmental conditions on servers.

Great suggestion on who defines the dump location. 663b03c adds dump location to the function signatures so that the pipeline gets to decide.

willbarton:
@higs4281 I still have a problem actually running the script. 

```
  File "cfgov/data_research/scripts/source_to_dump.py", line 137, in run
    if args[1] and args[1] != 'csv':
IndexError: tuple index out of range
```

cfgov/data_research/scripts/source_to_dump.py
```diff
@@ -0,0 +1,144 @@
...
+    `--script-args 2017-03-01 /tmp/mp_countydata csv`
+    """
+    dump_slug = DEFAULT_DUMP_SLUG
+    sql = True
+    starting_date = MortgageDataConstant.objects.get(
+        name='starting_date').date_value
+    if args:
+        through_date = parser.parse(args[0]).date()
+        update_through_date_constant(through_date)
+        if args[1] and args[1] != 'csv':
```
willbarton: @higs4281 If I understand the intent correctly (allow for 'csv' to be the second or third positional argument), then you may want to use `if (len(args) > 1 and args[1] != 'csv'):`

higs4281: Ah, thanks!
This now works for me: 
```
python cfgov/manage.py runscript source_to_dump --script-args 2008-12-01 /tmp/mp_countydata
```

You do have to load those fixtures, because prod data is old.

higs4281: @willbarton changed the arg test

willbarton: @higs4281 I still have a problem with not having `settings. AWS_STORAGE_BUCKET_NAME` defined.

willbarton:
@higs4281 I still have a problem with not having `settings. AWS_STORAGE_BUCKET_NAME` defined. What should it be?

The `source_url` in `source_to_dump.py` line 90 still ends up as `u'http://None.s3.amazonaws.com/data/mortgage-performance/source/latest_county_delinquency.csv'`

higs4281:
Oh, sorry. 
```
export AWS_STORAGE_BUCKET_NAME="files.consumerfinance.gov"
```
https://github.com/cfpb/cfgov-refresh/blob/master/.env_SAMPLE#L75

PR #3464, opened 2017-10-10
Lint fixes that require testing coverage

richaagarwal:
See https://github.com/cfpb/cfgov-refresh/pull/3451 for history


jimmynotjim:
@richaagarwal we should probably close this. I think we can safely say these files aren't going to get tested here and we should just wait till someone edits those files and adds tests.

PR #3469, opened 2017-10-11
Support running individual unit tests.

anselmbradford:
When refactoring an individual module it is helpful to be able to run only its associated unit tests. This PR adds support to run an individual unit test spec. Similar to what can be done with acceptance tests.

## Additions

- Adds support for running individual unit tests to `gulp test:unit` command.

## Testing

1. `gulp test:unit` should run all unit tests.
2. `gulp test:unit --specs=modules/o-table-row-links-spec.js` should run an indvidual test.
3. `gulp test:unit --specs=modules/transition/*.js` should run all transition and dependent unit tests.

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -207,7 +207,9 @@ The audit will run against
 [Google's PageSpeed Insights](https://github.com/addyosmani/psi).
 
 
-# Django and Python unit tests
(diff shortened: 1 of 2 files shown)


Scotchester:
Awesome!

PR #3473, opened 2017-10-12
Ask CFPB: Hide redirected pages from Category pages

higs4281:
Redirected pages in Ask CFPB need to remain live in order to do their
job of redirecting requests to another page. But Category and
SubCategory pages were not excluding redirected pages from their
collection of associated pages, resulting in some duplication in the
list of answers offered.

Addresses GHE issue 1292.

### DIFF
--- a/cfgov/ask_cfpb/models/pages.py
+++ b/cfgov/ask_cfpb/models/pages.py
@@ -249,7 +249,7 @@ class AnswerCategoryPage(RoutablePageMixin, CFGOVPage):
         context.update({
             'answers': answers,
             'audiences': audiences,
-            'facet_map': facet_map,
+            'facets': facet_dict,
             'choices': subcats,
             'results_count': answers.count(),
             'get_secondary_nav_items': get_ask_nav_items
@@ -295,10 +295,10 @@ class AnswerCategoryPage(RoutablePageMixin, CFGOVPage):
             self.ask_subcategory = subcat
         else:
             raise Http404
-        answers = self.ask_subcategory.answer_set.order_by(
-            '-pk').values(
-            'id', 'question', 'slug')
         context = self.get_context(request)
+        id_key = str(subcat.pk)
+        answers = context['answers'].filter(
+            pk__in=context['facets']['subcategories'][id_key])
         paginator = Paginator(answers, 20)
         page_number = validate_page_number(request, paginator)
         page = paginator.page(page_number)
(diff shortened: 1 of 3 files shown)


cfgov/ask_cfpb/tests/models/test_pages.py
```diff
@@ -828,21 +828,19 @@ def test_category_page_context_es(self):
...
     def test_category_page_context_no_es(self, mock_es_query):
+        blank_facets = {'answers': {},
+                        'audiences': {},
+                        'subcategories': {}}
         mock_return_value = mock.Mock()
-        mock_return_value.facet_map = json.dumps(
-            {'answers': {},
-             'audiences': {},
-             'subcategories': {}})
+        mock_return_value.facet_map = json.dumps(blank_facets)
```
hillaryj: Should `facet_map` here be `facets`?

cfgov/ask_cfpb/models/pages.py
```diff
@@ -249,7 +249,7 @@ def get_context(self, request, *args, **kwargs):
         context.update({
             'answers': answers,
             'audiences': audiences,
-            'facet_map': facet_map,
+            'facets': facet_dict,
```
hillaryj: Is this change to distinguish it from the `facet_map()` function?

cfgov/ask_cfpb/tests/models/test_pages.py
```diff
@@ -828,21 +828,19 @@ def test_category_page_context_es(self):
...
         mock_site = mock.Mock()
         mock_site.hostname = 'localhost'
         mock_request = HttpRequest()
         mock_request.site = mock_site
         cat_page = self.create_category_page(ask_category=self.category)
         test_context = cat_page.get_context(mock_request)
-        self.assertEqual(
-            test_context['facet_map'],
-            mock_return_value.facet_map)
+        self.assertEqual(test_context['facets'], blank_facets)
```
hillaryj: Is there a test already for checking that redirects don't make it in the list? If not, let's add one!

cfgov/ask_cfpb/tests/models/test_pages.py
higs4281: facet_map represents the json string that the elasticsearch query returns.
we deserialize that to make the `facets` value in context.

cfgov/ask_cfpb/tests/models/test_pages.py
higs4281: good idea!

PR #3486, opened 2017-10-17
Adding y axis label field to ChartBlock 

sebworks:
Adding y axis label field to ChartBlock 

## Changes

- Modified code to add the y_axis_label field to the chart block. This provides content editors the ability to override the default label. 

## Testing

1. Update the `y-axis-label` field via Wagtail.
2. Preview the page and view source. 
3. Verify that the chart block data value has the value entered.

## Notes

- This change corresponds to a change being made in the chart builder repo.

## Todos

-

## Checklist


cfgov/v1/atomic_elements/organisms.py
```diff
@@ -866,6 +866,9 @@ class ChartBlock(blocks.StructBlock):
         required=False,
         help_text='Text to display as a footnote. For example, '
                   '"Data from the last six months are not final."')
+    y_axis_label = blocks.CharBlock(
+        required=False,
+        help_text='Optional y-axis label')
```
marteki: Could we remove the word "optional" from the helper text here?

Our design standards state that [we should "label our [axes] to avoid confusion](http://cfpb.github.io/design-manual/data-visualization/chart-types.html)." I understand why no axis label is good in this one case, but I don't think we want to make it the norm, and the field is already not a required one.

sebworks: so, if you don't provide a label then it will fallback to using the existing logic.  There will always be a y-axis label though, this just provides the capability to override it.

marteki: If there's always going to be a label displayed, then all the more reason for removing the word  "optional." Maybe replacing it with "Custom y-axis label", if the purpose is for people to be able to override and customize the language?

hillaryj: "Custom y-axis label" sounds good!

sebworks: I used the word "optional" because it's already used with the metadata field.

sebworks: @hillaryj, @marteki, so If left the code as-is. Let me know if that works.

hillaryj: I think "Custom" makes a lot more sense given the logic and that it'll have a label even if not overridden. We can take "optional" out of the `metadata` description to make the language more uniform.

sebworks: ok, updated the label in https://github.com/cfpb/cfgov-refresh/pull/3486/commits/ed4f00feaf97c5236abd6686ccdfcd6bfffb0377

PR #3491, opened 2017-10-18
Remove redundant transition CSS

anselmbradford:
## Additions

- Add `cf-atomic-component` version `1.5.1`

## Removals

- Remove redundant transition CSS.

## Testing

1. `./frontend.sh` should pass and mega menu on site should run fine.


sebworks:
@anselmbradford, why add cf-atomic-component if you are deprecating it?

PR #3516, opened 2017-10-25
Add request where available to GitHub error issues

chosak:
The `alerts.logging_handlers.CFGovErrorHandler` logging handler posts logging messages as issues to the GitHub API. Currently this issue gets its title from the log message and its body from the exception attached to the message.

This information can sometimes be insufficient to debug an error, for example when query string parameters cause a problem. These parameters aren't necessarily included in either the log message or the exception stack trace.

Django's default error emails (in `django.utils.log.AdminEmailHandler`) include a representation of the request. This includes the full path being requested, the GET parameters, cookie information, etc. This change modifies `CFGovErrorHandler` to duplicate the content of Django 500 error emails, and use that content as the GH issue body.

Unfortunately due to how the Django 1.8 code is written, this requires some duplication with Django code. As of Django 1.9+ this could probably be rewritten to be a little bit nicer.

A note about privacy concerns: this change does add additional information to the GH issues that are created, but, again, this is consistent with information that was previously in error emails. Django does provide the ability to filter the data (for example, excluding sensitive POST parameters), and this change invokes the code that would do that, if configured. See the Django docs for more on this:

https://docs.djangoproject.com/en/1.8/howto/error-reporting/#filtering-error-reports

Pinging @ascott1 @schaferjh for any comments on privacy considerations.

Finally, this change modifies the existing unit tests for this code, to test the end-to-end `logging.error` functionality instead of the handler itself. This seemed more consistent with how Django tests its own handler and also requires less manual creation of e.g. `LogRecord` objects.

## Changes

- Duplication of Django error email logic to include requests in GitHub error emails.

## Testing

This can be tested with a repository of your own. You'll need a GitHub API token that has issue create permission for your repo. Run the following command in a shell:

```sh
(cfgov-refresh) $ GITHUB_TOKEN=your-api-token GITHUB_URL=gh-url GITHUB_USER=your-username GITHUB_REPO=your-repo DJANGO_SETTINGS_MODULE=cfgov.settings.production cfgov/manage.py shell
>>> import logging
>>> logger = logging.getLogger('django.request')
>>> from django.test import RequestFactory
>>> try:
...    raise ValueError('exception')
... except ValueError:
...    logger.exception('logging', extra={'request': RequestFactory().get('/')})
```

You should get a new issue on your repo that looks like this:

image

## Checklist


willbarton:
I'm pretty skeptical of the usefulness of logging these to GHE period at this point. It just seems to generate an exceptional amount of issue traffic where none of the individual issues that get opened are actionable (the eRegs issues are one example where quite a large number get opened for a single cause). Instead of getting one issue for each actionable problem, we get hundreds.

That's a long way of saying I'm not going to contribute a review because I'm pretty convinced at this point that this is generally not the way we should be handling 500s.

chosak:
@willbarton thanks for your comment, I understand. I'm in strong agreement that there are better ways that we could log errors. Given that this is what is currently enabled, though, it'd be nice if it were more useful as a debugging tool. I'll defer to PM review of the privacy issues, and hopefully we can pursue an alternate approach soon.

schaferjh:
My skim of this is that I think you're handling the situation fine vis a vis privacy, but I defer to @ascott1

ascott1:
I agree with @schaferjh's comment in regards to privacy.

In regards to @willbarton's comment, my views reflect those of @chosak. I agree that there are better ways to log these errors, but while we are doing it this way, it makes sense to make them usable as possible.

willbarton:
Is the information we're looking for available in New Relic?

chosak:
@willbarton unfortunately I don't believe so. New Relic does give you the stack trace, but doesn't let you see the query string part of the request, which is often deterministic of the error. I believe this is due to [high security mode](https://docs.newrelic.com/docs/agents/manage-apm-agents/configuration/high-security-mode).

PR #3519, opened 2017-10-26
Remove mocha jsdom

anselmbradford:
Fixes [GHE]/CFGOV/platform/issues/1455
`mocha-jsdom` is not being maintained with the current version of jsdom. We can run jsdom directly with some tweaks. This should be the preferable option so that we don't create a series of tests that depend on proprietary mocha-jsdom syntax.

## Additions

- `jsdom-global`

## Removals

- `mocha` (I don't _think_ this is needed, it's not referenced?)
- `mocha-jsdom`

## Changes

- Updates `data-set-spec`, since jsdom now supports `dataset` attribute.
- Moved `HTML_SNIPPET` constant to above describe block for consistency.
- Minor indent cleanup.
- Uppercase two constants.
- Wraps tests in a describe block that weren't inside one root describe block.
- Updates `sinon` to version 4.

## Testing

1. `npm install && gulp test:unit`

## Todos

- ~~Some tests were `xit`'d because they relied on form submission. These need to be refactored to not require form submission since jsdom does not support it https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/living/nodes/HTMLFormElement-impl.js#L70 @jimmynotjim had some ideas.~~ Tests are now passing.
- GlobalBanner is unnecessarily complex and requires updates for the latest jsdom that aren't worth it in my opinion. Aiming to resolve this in https://github.cfpb.gov/CFGOV/platform/issues/1899
- In theory this will allow updating of `gulp-mocha`, but at the moment it's not compatible with istanbul, see https://github.com/SBoudrias/gulp-istanbul/issues/115

## Checklist


test/unit_tests/apps/financial-well-being/fwb-questions.js
```diff
@@ -186,8 +190,8 @@ describe( 'fwb-questions', () => {
       expect( formSubmissionStatus ).to.equal( false );
     } );
 
-  it( 'submit button should submit the form ' +
-      'if all the questions are completed before page load.',
+  xit( 'submit button should submit the form ' +
```
sebworks: These aren't testing form submission; they are testing if `event.preventDefault` was [called](https://github.com/cfpb/cfgov-refresh/blob/5b9987de360a49089b54f1b6fc1cd528f0368a02/cfgov/unprocessed/js/apps/financial-well-being/fwb-questions.js#L55) when the click event was fired. https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent.

I think these test are fine.

anselmbradford: They won't pass as is because they trigger a form submission in the DOM, but yeah could be re-written to not.

sebworks: yeah, I would just add in the before statement:

```
    document.querySelector( '#quiz-form' ).addEventListener( 'submit', event => event.preventDefault() )'
```

anselmbradford: Refactored the test, try it now

anselmbradford: For some reason it ignored the submission event in the before, but works in the trigger click event.

anselmbradford:
@sebworks please chuck your local and re-pull. I rebased with master and did some cleanup.

sebworks:
yeah, I think `mocha` is a requirement of `gulp-mocha`, so we should be good.

PR #3523, opened 2017-10-27
Simplify beta banner

anselmbradford:
Beta banner is way over-engineered. This removes significant JS, CSS, and markup. Fixes https://[GHE]/CFGOV/platform/issues/1899

## Removals

- Removes beta banner JS and associated unit test.
- Removes beta banner expandable-specific less.
- Removes beta banner expandable markup.

## Testing

1. Turn the beta banner flag on in wagtail.
2. visit site and observe non-expandable beta banner.

## Screenshots

screen shot 2017-10-27 at 1 49 13 pm


sebworks:
Remove some of the padding top / bottom and this is great.

anselmbradford:
Updated the padding:

screen shot 2017-10-28 at 12 06 32 am

anselmbradford:
This should really just be a notification organism straight from CF, but that can be another PR.

schaferjh:
@anselmbradford This looks good -- before signing off, would you post a screenshot of what this looks like a mobile dimensions? That's my only area of concern with removing the expandable.

anselmbradford:
@schaferjh different sizes below!

screen shot 2017-10-30 at 9 59 49 am
screen shot 2017-10-30 at 10 00 01 am
screen shot 2017-10-30 at 10 00 24 am

schaferjh:
Great -- looks good. :+1:

anselmbradford:
@Scotchester Are you okay with this?

Scotchester:
I'd prefer it be smaller overall, if it's not going to be collapsible anymore, but I agree that it's worth removing the expandable and its attendant complications.

How would front-end coverage possibly be dropping?

anselmbradford:
> How would front-end coverage possibly be dropping?

You have a carton of 12 eggs, 5 of which are broken. Broken percentage is 5/12 = ~42%
You remove one good egg. Broken percentage is now 5/11 = ~45%

Scotchester:
Oy. Another situation where it seems imprudent to require maintaining a coverage percentage on this kind of PR. Fortunately, looks like that requirement is still off temporarily :grin:

anselmbradford:
> I'd prefer it be smaller overall

Comparing it to cf-notifications, I can reduce the gap between the bold text and the paragraph text, that will help a little.

anselmbradford:
@Scotchester

Okay, I replaced the global banner with a cf-notification, which removes more less:

screen shot 2017-10-30 at 10 46 28 am
screen shot 2017-10-30 at 10 46 39 am
screen shot 2017-10-30 at 10 46 48 am

Scotchester:
Margins and paddings look good now, and good to remove yet more Less, but if I want to be real nitpicky, I'd request undoing the `max-width` rule on the paragraph inside (or just make it not a `p` element) so that it remains on one line at the page's max width.

anselmbradford:
It's actually okay as-is. I had screenshot it at a size where it wraps, but at larger sizes it won't:
screen shot 2017-10-30 at 11 05 39 am

Scotchester:
Ah, great! Looks good enough to me, now. (I'd still like it to be smaller, even if that means creating a new pattern, but that can be discussed later.)

jimmynotjim:
> even if that means creating a new pattern

I think we should avoid that, especially in Capital Framework, for non-public facing content. No one outside those testing beta/build will ever see it. If/when it does become a need for the public side, we can revisit.

Scotchester:
Beta _is_ public-facing content. Anyone can access it. It has been used for usability testing and may be used that way again in the future.

jimmynotjim:
That seems like such a small edge case for a completely new component though. The existing notification pattern works and does the job required of it.

Scotchester:
But it's sooooo big 😿 

I'm also not necessarily arguing for it to be in CF. It could be unique to cfgov-refresh. We also have a little banner on Content that could use similar styling.

anselmbradford:
> But it's sooooo big

We could leave out the paragraph. Do we really need two lines? Couldn't it just be:

> This beta site is a work in progress. Our regular site is at consumerfinance.gov.

Scotchester:
That's a question for @schaferjh, but I would certainly approve of that change.

schaferjh:
I appreciate the sentiment but I don't think the amount of text is causing a legitimate problem and would prefer to keep it as is. The current message is a nice explanation for a broad range of audiences since we can't necessarily predict who will stumble on beta (and whether the understand what a "beta" is!). When we do use beta for testing, participants learn very quickly to block out repetitive text like that banner.

PR #3536, opened 2017-10-31
Differentiate GHE issues posted via logging from alerts

willbarton:
This PR changes the `GitHubAlert.post` to take an optional `labels` kwarg, and the logging handler to use that kwarg to assign the `logging` label instead of `alert`. This is because, as we've seen, some logged issues are not singularly actionable, instead they are all related to the same single problem. This is contrary to our alert policy, where we want each individual alert to be actionable. The practical effect of this is that a large number of issues opened by the logging handler obscures other issues when filtering for "alert" (and makes it very difficult, at least for me, to find other alert issues I'm looking for). Thus I am proposing that we open logged issues with a different tag, `logging`. 

## Notes

This PR is currently based on #3516, when that PR is merged I'll rebase to master. @chosak refactored some of this code in that PR, and it seemed like the best starting point.

## Checklist

### DIFF
--- a/cfgov/alerts/github_alert.py
+++ b/cfgov/alerts/github_alert.py
@@ -33,7 +33,7 @@ class GithubAlert(object):
         issues = self.repo().iter_issues(state='all')
         return next((issue for issue in issues if issue.title == title), None)
 
-    def post(self, title, body):
+    def post(self, title, body, labels=['Maintenance and Response', 'alert']):
         # Truncate the title if needed, max is 256 chars
         title = title[:256]
         issue = self.matching_issue(title)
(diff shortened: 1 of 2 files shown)


willbarton:
@hillaryj logged 500s were never intended to be treated as alerts, they don't follow our policy for alerts (GHE/CFGOV/platform/wiki/cf.gov-Alert-Policy), and as it is right now they make filtering on the "alert" tag useless. This is an attempt to relieve that confusion, policy-violation, and practical effect.

So this clarifies that there is one class of alert, and that 500s that get logged to GHE are their own distinct class of thing.

PR #3537, opened 2017-10-31
Adding tox environment for testing asset validation

sebworks:
Adding tox environment for testing asset validation

## Changes

- Modified `frontend.sh` to add a `dry-run` function.

## Testing

1. Run `tox -e validate-assets -v` and hope you don't see `[object object]`.
2. You can run `tail -f ` on the log file to check the dependency build cycle.

## Notes

-

## Todos

-

## Checklist

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -76,9 +76,9 @@ setenv={[testenv:fast]setenv}
 
 [testenv:missing-migrations]
 # Invoke with: tox -e missing-migrations
-# Check for changes to Django models that are missing migrations by 
+# Check for changes to Django models that are missing migrations by
 # running a script that calls makemigrations and examines the result.
-# The missing_migrations script can be removed in favor of 
+# The missing_migrations script can be removed in favor of
 # `makemigrations --check` when we move to Django 1.10+.
 envdir={[testenv]envdir}
 recreate=False
@@ -87,4 +87,23 @@ setenv=
     {[testenv]setenv}
     DJANGO_SETTINGS_MODULE=cfgov.settings.test
 commands=
-    ./manage.py runscript missing_migrations 
+    ./manage.py runscript missing_migrations
+
+[testenv:validate-assets]
+# Invoke with: tox -e validate-assets
+# Ensure all assets are generated without error.
+envdir={[testenv]envdir}
+changedir={toxinidir}
+recreate=False
+deps=
+    -r{toxinidir}/requirements/mysql.txt
+    {[testenv]deps}
+    -r{toxinidir}/requirements/optional-public.txt
+    -r{toxinidir}/requirements/optional-private.txt
+setenv=
+    {[testenv]setenv}
+    DJANGO_SETTINGS_MODULE=cfgov.settings.production
+    DJANGO_STATIC_ROOT={toxinidir}/collectstatic
+commands=
+    {toxinidir}/frontend.sh production
+    {toxinidir}/cfgov/manage.py collectstatic --noinput


frontend.sh
```diff
@@ -93,6 +93,17 @@ build() {
   fi
 }
 
+# Run tasks to build all project assets for production distribution.
+dry_run() {
```
willbarton: I'm not sure `dry_run` is the best descriptor for what this does. `validate_static_files` maybe?

frontend.sh
```diff
@@ -93,6 +93,17 @@ build() {
...
+  set DJANGO_SETTINGS_MODULE=cfgov.settings.production
+  set DJANGO_STATIC_ROOT=$(pwd)/collectstatic
+  pip install -r requirements/optional-public.txt
+  pip install -r requirements/optional-private.txt
```
willbarton: It occurs to me that we *might* want to do this in a tox environment to avoid muddying up someone's working env.

sebworks: yeah, how would you do that outside a testing environment?

willbarton: Create a testing environment in tox.ini :) 

Unfortunately to do so would require `tox` to be installed, which I don't think we can expect to have when we get to `frontend.sh`. 

Happy to help figure this out if you like.

sebworks: ok, I've added a script to tox.ini but I'm now getting a `No module named MySQLdb` error. I'm not sure why MySQLdb would be necessary when running collectstatic. Also, it's hecka slow and the `-vv` flag doesn't work as expected unless we upgrade to tox 2.9.

sebworks: ok, I just added mysql as a dependency. Should be good to go now.

tox.ini
```diff
@@ -87,4 +87,23 @@ setenv=
     {[testenv]setenv}
     DJANGO_SETTINGS_MODULE=cfgov.settings.test
 commands=
-    ./manage.py runscript missing_migrations 
+    ./manage.py runscript missing_migrations
+
+[testenv:validate-assets]
+# Invoke with: tox -e validate-assets
+# Ensure all assets are generated without error.
+envdir={[testenv]envdir}-validate-assets
```
willbarton: I would leave this as `envdir={[testenv]envdir}` like the other envs, then tox doesn't have to set up a brand new one just for this. That setup time is part of why it takes so long.

frontend.sh
willbarton: The tox environment works beautifully. Can we remove this function, or have it call `tox -e validate-assets` (and rename it if we're keeping it)?

tox.ini
```diff
@@ -87,4 +87,23 @@ setenv=
...
+changedir={toxinidir}
+recreate=False
+deps=
+    -r{toxinidir}/requirements/optional-public.txt
+    -r{toxinidir}/requirements/optional-private.txt
+    {[testenv]deps}
+    mysql
```
willbarton: If you add `-r{toxinidir}/requirements/mysql.txt`, you'll be including the `mysql` package we install for local and production.

tox.ini
sebworks: fixed in https://github.com/cfpb/cfgov-refresh/pull/3537/commits/70d709ce9ce0ce058da1cfa9761914817d8341ca

tox.ini
sebworks: fixed in https://github.com/cfpb/cfgov-refresh/pull/3537/commits/70d709ce9ce0ce058da1cfa9761914817d8341ca

frontend.sh
sebworks: fixed in https://github.com/cfpb/cfgov-refresh/pull/3537/commits/70d709ce9ce0ce058da1cfa9761914817d8341ca.

PR #3539, opened 2017-10-31
Move inline JS to external requests

anselmbradford:
To decrease the time to first paint of the page, we should aim to reduce the HTML page size. One step to doing that is removing the inlined JS. (Part of https://[GHE]/CFGOV/platform/issues/1824).

## Removals

- `banner-footer-webpack-plugin` plugin.
 
## Changes

- Removes inlined JS and moves it instead to external JS asset requests.

## Testing

1. `./frontend.sh` and check

http://localhost:8000/about-us/careers/current-openings/ for testing template-level script (single.js)
http://localhost:8000/ask-cfpb/ for testing page-level script (index.js)
http://localhost:8000/about-us/the-bureau/bureau-structure/ for testing on-demand script (on-demand/bureau-structure.js)

**Note: I don't think you need `./cfgov/manage.py collectstatic`, but try that if the scripts 404.

## Screenshots

HTML byte size before:
screen shot 2017-10-31 at 12 39 45 pm

HTML byte size after:
screen shot 2017-10-31 at 12 39 49 pm

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -218,8 +218,6 @@ MEDIA_URL = '/f/'
 #Enabling compression for use in base.html
 COMPRESS_ENABLED = True
 
-COMPRESS_ROOT = MEDIA_ROOT
-
 COMPRESS_JS_FILTERS = []
 
 # List of finder classes that know how to find static files in
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -171,6 +171,7 @@ TEMPLATES = [
     },
 ]
 
+
 WSGI_APPLICATION = 'cfgov.wsgi.application'
 
 # Admin Url Access
@@ -212,6 +213,14 @@ STATIC_ROOT = os.environ.get('DJANGO_STATIC_ROOT', '/var/www/html/static')
 MEDIA_ROOT = os.environ.get('MEDIA_ROOT',
                             os.path.join(PROJECT_ROOT, 'f'))
 MEDIA_URL = '/f/'
+
+
+#Enabling compression for use in base.html
+COMPRESS_ENABLED = True
+
+COMPRESS_ROOT = MEDIA_ROOT
+
+COMPRESS_JS_FILTERS = []
 
 # List of finder classes that know how to find static files in
 # various locations.
--- a/cfgov/jinja2/v1/_layouts/base.html
+++ b/cfgov/jinja2/v1/_layouts/base.html
@@ -243,32 +243,42 @@ height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
 {% block javascript scoped %}
 <!--[if gt IE 8]><!-->
 {# Include site-wide JavaScript. #}
-<script>
-    {% include '/js/routes/common.js' %}
-</script>
+<script src="{{ static('js/routes/common.js') }}"></script>
 
-{# Include template and page-specific JavaScript. #}
-{% macro page_template_js() %}
-    {# Check and include template-level JavaScript. #}
-    {% include '/js/routes/' + request.path.split('/')[1] + '/single.js' ignore missing %}
-
-    {# TODO: Remove Else conditional when this is fully moved to wagtail. #}
-    {# Check and include page-level JavaScript. #}
-    {% if page and page.media %}
-        {% for type, jsfiles in page.media.iteritems() %}
-            {% for js in jsfiles %}
-                {% include '/js/routes/on-demand/' + js ignore missing %}
-            {% endfor %}
-        {% endfor %}
-    {% else %}
-        {% include '/js/routes/' + request.path[1:] + 'index.js' ignore missing %}
-    {% endif %}
+{# Check and include template-level JavaScript. #}
+{% macro template_js() %}
+    {% include 'js/routes/' + request.path.split('/')[1] + '/single.js' ignore missing %}
 {% endmacro %}
-{% set js_source = page_template_js() | trim %}
+{% set js_source = template_js() | trim %}
 {% if js_source | length > 0 %}
-<script>
-    {{ js_source }}
-</script>
+<script src="{{ static('js/routes/' + request.path.split('/')[1] + '/single.js') }}"></script>
+{% endif %}
+
+{# Check and include component-level JavaScript. #}
+{% macro component_ondemand_js(js) %}
+    {% include 'js/routes/on-demand/' + js ignore missing %}
+{% endmacro %}
+
+{% if page and page.media %}
+   {% for type, jsfiles in page.media.iteritems() %}
+       {% for js in jsfiles %}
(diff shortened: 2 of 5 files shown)


cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -241,32 +241,42 @@
...
+{% macro component_ondemand_js(js) %}
+    {% include 'js/routes/on-demand/' + js ignore missing %}
+{% endmacro %}
+
+{% if page and page.media %}
+   {% for type, jsfiles in page.media.iteritems() %}
+       {% for js in jsfiles %}
+           {% set js_source = component_ondemand_js(js) | trim %}
+           {% if js_source | length > 0 %}
+           <script src="{{ static('js/routes/on-demand/' + js) }}"></script>
```
sebworks: I'm not sure this is the wag to go because you potentially could have 10 scripts on a page. How would be feel about generating page-type ( browse, sublanding, etc...) specific bundles.

anselmbradford: Is there a way we could put this loop into a separate jinja template, which aggregates the scripts and inlines them into that template, and then `base.html` makes an external request to the result of that template. Basically, can jinja be used to dynamically generate a JS file?

anselmbradford: Separate page-type bundles would still be delivering a lot of unneeded code, right?

sebworks: > Is there a way we could put this loop into a separate jinja template, which aggregates the scripts and inlines them into that template, and then base.html makes an external request to the result of that template. Basically, can jinja be used to dynamically generate a JS file?

yes, you can use the compression [tags](https://github.com/cfpb/cfgov-refresh/pull/1409/files#diff-89b43ee86bbd4952dc44cf751cfd775cR205) and set the COMPRESS_JS_FILTERS to []. That seems hacky though, @rosskarchner probably knows a better way to do it.

sebworks: >> Separate page-type bundles would still be delivering a lot of unneeded code, right?
  
No, but you would need to generate every possible combination of components for the given page type.

cfarm: rather than lots of bundles that would be downloaded any time you visit a new page type, can we deliver the JS as one bundled file for now and investigate the impact of different optimization and delivery techniques? At least with one bundled file loaded asynchronously you have the ability to cache it for subsequent page views, reducing overall page load times for the user.

(I think this is what you're suggesting with this comment @ans)
> Is there a way we could put this loop into a separate jinja template, which aggregates the scripts and inlines them into that template, and then base.html makes an external request to the result of that template. Basically, can jinja be used to dynamically generate a JS file?

cfarm:
Highly recommend an improvement to this PR to bundle all the component JS into one single file and have that loaded by all pages. This is an MVP improvement over the status quo, in which every page has to re-download and parse the inline JavaScript, without the benefits of caching the JS or loading non-JS-dependent content more quickly.

Subsequent improvements to bundling the JS files can still be implemented and tested later; but for now we want to see the first meaningful paint time decrease, and serving the JS as a separate single file will accomplish that.

jimmynotjim:
@anselmbradford what page are those graphs for? In theory the gains for a page like blog posts would be minimal whereas one with a ton of scripts would be huge. I think your best bet would be to compare a few pages with a control version (aka live now), your individual component files, and @cfarm's suggestion for one component file. There just doesn't seem to be enough information here to determine the best course of action.

cfarm:
Another thought is to A/B test our different bundling options to see what the impact is on Google Analytics metrics for mobile users we are tracking: Avg. Document Interactive Time and Avg Page Load Time.

sebworks:
@anselmbradford, for now,  just enable the compression and get a BEWD to sign-off. We can work on the generating various bundles.

anselmbradford:
@sebworks would that be your work here https://github.com/cfpb/cfgov-refresh/pull/1409, what shouldn't I take from there? E.g. why was it closed?

anselmbradford:
> what page are those graphs for?

It's the homepage.

jimmynotjim:
It's the homepage.

Are we even loading many individual components there? Do we have a page with multiple scripted components, like maybe a video player and expandables or expandables and sortable tables?

sebworks:
>> @sebworks would that be your work here, what shouldn't I take from there? E.g. why was it closed?

I closed it because I wasn't sure if having an external script was better. If you goto Google.com or Facebook.com all their scripts are inline. Let's just test out using the external scripts and see how they perform.

anselmbradford:
Can I get a bewd signoff @rosskarchner @willbarton @richaagarwal @chosak ?

cfgov/cfgov/settings/base.py
```diff
@@ -213,6 +214,14 @@
                             os.path.join(PROJECT_ROOT, 'f'))
 MEDIA_URL = '/f/'
 
+
+#Enabling compression for use in base.html
+COMPRESS_ENABLED = True
+
+COMPRESS_ROOT = MEDIA_ROOT
+
+COMPRESS_JS_FILTERS = []
```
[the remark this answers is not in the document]
sebworks: We we keeping it empty because we don't want the files to be minified again. They have already been minified by the front-end process.

cfgov/cfgov/settings/base.py
```diff
@@ -213,6 +214,14 @@
                             os.path.join(PROJECT_ROOT, 'f'))
 MEDIA_URL = '/f/'
 
+
+#Enabling compression for use in base.html
+COMPRESS_ENABLED = True
+
+COMPRESS_ROOT = MEDIA_ROOT
```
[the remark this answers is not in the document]
anselmbradford: Any idea on this one @sebworks ?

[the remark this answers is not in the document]
sebworks: yep, no longer necessary. Removed in https://github.com/cfpb/cfgov-refresh/pull/3539/commits/c34f463c3a0c57b873d417da48b7155cf049580d

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -243,32 +243,46 @@
...
-</script>
+<script src="{{ static('js/routes/' + request.path.split('/')[1] + '/single.js') }}"></script>
+{% endif %}
+
+{# Check and include component-level JavaScript. #}
+{% macro component_ondemand_js(js) %}
+    {% include 'js/routes/on-demand/' + js ignore missing %}
+{% endmacro %}
+
+{% if page and page.media %}
```
sebworks: The page.media class will always return a dictionary. It's a different bug that we need to fix.

cfgov/cfgov/settings/base.py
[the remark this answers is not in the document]
willbarton: Alrighty.

anselmbradford:
For posterity, the possible JS bundling options @cfarm @sebworks and I explored 11/2:

js architecture

PR #3540, opened 2017-10-31
Add feature flag for recruitment code comment

anselmbradford:
To reduce the first paint time of the page we should reduce the size of the HTML page (see https://[GHE]/CFGOV/platform/issues/1824). Removing unused code comments is a way to do this. When we're not recruiting for a fellowship we don't need to show this code block. This flag could also be used to show other recruitment campaigns across the site.

## Additions

- Adds a feature flag `CFPB_RECRUITING` for showing or hiding the recruitment code comment in the base template's HTML source.

## Testing

1. Visit the site locally and view the source. There should be no recruitment code comment at the top.
2. Log into /admin Wagtail locally and set `CFPB_RECRUITING` flag boolean to true.
3. Re-visit site locally and view the source. There should be a recruitment code comment.

## Screenshots

screen shot 2017-10-31 at 1 17 10 pm

## Notes

I think this can be off in production? We're not hiring are we? @ascott1 Can you confirm?

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -525,10 +525,13 @@ FLAGS = {
         'site': 'beta.consumerfinance.gov',
     },
 
-    # When enabled, Display a "techical issues" banner on /complaintdatabase
+    # When enabled, include a recruitment code comment in the base template.
+    'CFPB_RECRUITING': {},
+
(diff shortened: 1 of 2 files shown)


ascott1:
👍 I like this change.

Fun fact: the original code comment (as mentioned in an Atlantic blog post) is how I wound up here.

PR #3551, opened 2017-11-02
Only include StreamField media for existing blocks

chosak:
This change fixes some logic that was trying to include StreamField child block media. We want pages to include all child block media (block-specific JavaScript files) for any blocks that are on the page, but the existing logic includes any files for any blocks that could possibly be on the page.

This change leverages some logic in [wagtail-inventory](https://github.com/cfpb/wagtail-inventory) that lists the block types on a page, instead of incorrectly pulling in all possible child blocks of a parent block.

A few unit tests had to be changed because using `model_mommy.prepare` on Wagtail page types doesn't properly set the Django contenttypes id, which breaks use of `page.specific`.

This addresses the first issue in GHE#1264.

## Changes

- Fix logic in `v1.models.base._add_streamfield_js` to properly include only required block JS.

## Testing

- Create and publish a new BrowsePage that has an empty Full Width Text organism. Use the developer tools to search page JS for `o-table`. You shouldn't see any hits because there's no table block in the page. Also note no hits for `email-signup`.
- Now add an Email Signup to the sidebar, and republish. You'll see the page source JS now properly includes `email-signup`.

## Checklist

### DIFF
--- a/cfgov/ask_cfpb/tests/models/test_pages.py
+++ b/cfgov/ask_cfpb/tests/models/test_pages.py
@@ -121,9 +121,7 @@ class AnswerModelTestCase(TestCase):
         kwargs.setdefault('depth', self.english_parent_page.depth + 1)
         kwargs.setdefault('slug', 'category-mortgages')
         kwargs.setdefault('title', 'Mortgages')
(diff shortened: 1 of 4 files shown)


sebworks:
this looks good to me, before you merge let me move the Media class.

PR #3554, opened 2017-11-03
Flatten CFGOVPage.media from dict to list

chosak:
This change flattens the `@media` property on `CFGOVPage` to return a simple list of JavaScript filenames instead of a dictionary. The dictionary keys were never used downstream in the template, so maintaining them serves no purpose and makes it harder to test for files by checking for an empty dictionary.

This change also slightly cleans up the class methods around this functionality, converting `CFGOVPage.add_page_js(js)` (which adds to a dictionary) into `CFGOVPage.page_js` (which just returns a list).

Several Ask CFPB classes and tests had to be modified for this; pinging @higs4281 for a quick review.

Note that this change somewhat strengthens the existing requirement that page/block JS filenames have to be globally unique.

## Changes

- Cleanup of page media into simple list.

## Testing

1. See new/changed unit tests.
2. Create a BrowsePage and notice that the secondary navigation JS is included (compressed) in the scripts (search `o-secondary-navigation` in developer tools).
3. Create a HomePage and notice that this page JS is not included in the template or scripts.

## Notes

- This is the issue referred to by @sebworks on #3539 [here](https://github.com/cfpb/cfgov-refresh/pull/3539#discussion_r148601763).

## Checklist

### DIFF
--- a/cfgov/ask_cfpb/models/pages.py
+++ b/cfgov/ask_cfpb/models/pages.py
@@ -178,7 +178,18 @@ class AnswerLandingPage(LandingPage):
         return 'ask-cfpb/landing-page.html'
 
 
-class AnswerCategoryPage(RoutablePageMixin, CFGOVPage):
+class SecondaryNavigationJSMixin(object):
+    """A page mixin that adds navigation JS for English pages."""
+    @property
+    def page_js(self):
+        js = super(SecondaryNavigationJSMixin, self).page_js
+        if self.language == 'en':
+            js += ['secondary-navigation.js']
+        return js
+
+
+class AnswerCategoryPage(RoutablePageMixin, SecondaryNavigationJSMixin,
+                         CFGOVPage):
     """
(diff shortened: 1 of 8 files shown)


cfgov/ask_cfpb/models/pages.py
```diff
@@ -178,7 +178,18 @@ def get_template(self, request):
         return 'ask-cfpb/landing-page.html'
 
 
-class AnswerCategoryPage(RoutablePageMixin, CFGOVPage):
+class SecondaryNavigationJSMixin(object):
+    """A page mixin that adds navigation JS for English pages."""
+    @property
```
sebworks: This is really nice work

sebworks:
Are there any places where you might be able to use the `@cache_property` decorator?

chosak:
> Are there any places where you might be able to use the `@cache_property` decorator?

That's interesting, I hadn't considered this. In practice these methods should only get called once when the page is served. Is there some use case where they might be called multiple times?

PR #3561, opened 2017-11-06
Add New Relic Python agent, pin to current version

acrewdson:
Adding New Relic Python agent (latest version); it is currently being installed via other means, but it would make more sense to have it installed here.


marcesher:
👍

OrlandoSoto:
👍

willbarton:
Should this maybe be [in `production.txt`](https://github.com/cfpb/cfgov-refresh/blob/master/requirements/production.txt) instead of `base.txt`?

hillaryj:
👍  and second @willbarton's comment about restricting scope of where the newrelic agent gets installed.

acrewdson:
@willbarton @hillaryj I don't know enough about the different requirements files, but since we use New Relic on many environments (build/content/beta/prod), `base` seemed safest to me. Note that I just now updated the main PR (have you both had a chance to review it? I have a question there about the whitelist thing there -- https://github.cfpb.gov/cfpb-infra/ansible-consumer-finance/pull/151/files ) so that New Relic does not get enabled locally -- though yes, the Python dep would still be installed as you note. Either way, would `production.txt` not be the right thing if we are needing this dep in build?

acrewdson:
@rosskarchner will do, and will make corresponding changes to the related PRs

PR #3580, opened 2017-11-09
WIP: Restrict pages found to current site

hillaryj:
WIP: Expanding the scope to other similarly structured pages.

The activity feed page was searching the entire database for `ActivityLogPage` objects and found a page in Trash that had been left live. The same was also possible for `NewsroomLandingPage` objects.

This now passes the `Site` object instead of a string `hostname` and restricts found objects to only those that belong to the `Site` being served (`in_site(site)`).

### DIFF
--- a/cfgov/v1/util/filterable_list.py
+++ b/cfgov/v1/util/filterable_list.py
@@ -33,13 +33,6 @@ class FilterableListMixin(object):
                                   self.per_page_limit())
             page = request.GET.get('page')
 
-    def process_form(self, request, form):
-        filter_data = {}
-        if form.is_valid():
-            paginator = Paginator(form.get_page_set(),
-                                  self.per_page_limit())
-            page = request.GET.get('page')
-
             # Get the page number in the request and get the page from the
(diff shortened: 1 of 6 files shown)


cfgov/v1/util/filterable_list.py
```diff
@@ -22,10 +22,17 @@ def get_context(self, request, *args, **kwargs):
         context['filter_data'] = self.process_form(request, form)
         return context
 
-    def base_query(self):
-        return AbstractFilterPage.objects.live().filter(
+    def base_query(self, site):
+        return AbstractFilterPage.objects.live().in_site(site).filter(
             CFGOVPage.objects.child_of_q(self))
 
+    def process_form(self, request, form):
```
chosak: Is this a copy/paste error? This code is duplicated from below.

hillaryj: Sure is - I rebased. Thanks for catching that.

chosak:
Closing this as it has been superseded by the fix in #3750.

Issue #3578, opened 2017-11-09
server-rack.png image used?


anselmbradford:
Short description explaining the high-level reason for the new issue.

## Current behavior

- The site has both [server.png](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/img/server.png) and [server-rack.png](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/img/server-rack.png), but only server.png is referenced in the less files.
 
## Expected behavior

- Only used images are in the site's repo.

PR #3582, opened 2017-11-09
New error images

anselmbradford:
Fixes https://github.com/cfpb/cfgov-refresh/issues/3578

## Removals
 
- server-rack.png

## Changes

- doubles size of server.png and plug.png for retina display.
 
## Testing

1. localhost:8000/500/ and localhost:8000/404 

## Screenshot

screen shot 2017-11-09 at 5 17 04 pm

screen shot 2017-11-09 at 5 17 27 pm


anselmbradford:
Note that size is approximately double the size they were being displayed at as a background image, not the size of the original images.

anselmbradford:
Looking for a dev review of this @cfarm @Scotchester @jimmynotjim

PR #3589, opened 2017-11-15
Refactoring related modules for readability and usage

richaagarwal:
This PR attempts to clean up some of the code around related modules by
- Moving functions onto the CFGOVPage class when possible, instead of as helper functions
- Rewriting `related_posts` so that it's more readable. Does not change functionality 

All tests still pass

### DIFF
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -156,22 +156,6 @@ class CFGOVPage(Page):
         ]
         specific_categories = block.value['specific_categories']
 
-        def get_appropriate_categories(specific_categories, page_type):
-            """ An array of categories is provided from whatever
-            is selected in the admin, however they each correspond to
-            a page type, e.g. newsroom or blog. This function returns
-            only the categories that belong to the page type in question
-            """
-
-            # Convert the provided categories to their slugs
-            category_slugs = ref.related_posts_category_lookup(
-                specific_categories
-            )
-            # Look up the available categories for the page type in question
-            options = [c[0] for c in ref.choices_for_page_type(page_type)]
-
-            return [c for c in category_slugs if c in options]
-
         for slug, block_name, title in search_types:
             get_related_posts = block.value.get(
                 'relate_{}'.format(block_name), None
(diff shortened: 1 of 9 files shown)


cfgov/v1/models/base.py
```diff
@@ -149,73 +149,74 @@ def generate_view_more_url(self, request):
...
-                    AbstractFilterPage.objects.live().filter(
-                        search_query
-                    ).distinct().exclude(id=self.id).order_by(
-                        '-date_published'
-                    )
-                )
+        specific_categories = block.value['specific_categories']
+
+        for slug, should_get_related_posts in search_types:
+            if should_get_related_posts:
```
anselmbradford: Could this be a guard clause? So return immediately if `should_get_related_posts` does not exist. I don't know what that would look like in python.

cfgov/v1/models/base.py
```diff
@@ -149,73 +149,74 @@ def generate_view_more_url(self, request):
...
+        for tag in self.specific.tags.all():
+            tag_link = {'text': tag.name, 'url': ''}
+            if id is not None and filter_page is not None:
+                param = '?filter' + str(id) + '_topics=' + tag.slug
+                tag_link['url'] = relative_url + param
+            tags['links'].append(tag_link)
+        return tags
+
+    def get_filter_data(self):
+        for ancestor in self.get_ancestors().reverse().specific():
```
anselmbradford: Should `self.get_ancestors().reverse().specific()` go in its own variable before the loop? It looks like each method will get called each loop iteration.

anselmbradford:
Passes tox tests. Probably would help to get a bewd review on the actual code. 👍

cfgov/v1/models/base.py
richaagarwal: It's called in the outermost loop so it's only executed once

cfgov/v1/models/base.py
richaagarwal: I changed this in another way, but one option could've been to do `if not should_get_related_posts: continue` and then have all the code below that indented to the left.

PR #3593, opened 2017-11-17
Tweak Docker setup for local development

willbarton:
This PR makes a few tweaks to our (wonderful) docker local development setup.

## Additions

- Adds `eval $(docker-machine env)` to shell.sh and attach.sh in case the user forgets to run it (as I frequently find myself doing)
- Adds `.python_env` to `.gitignore`
- Adds a `docker` argument to `setup.sh` and `runserver.sh` to allow them to still be used for setup and running.
- Adds the ability to run a command directly with `shell.sh`, i.e. `./shell.sh cfgov/manage.py migrate`

## Changes

- Fixes the link to the usage document in the installation instructions
- Fixes the comment formatting in the `.gitignore` file
- Updates the documentation for the above changes.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -115,3 +115,7 @@ coverage.xml
 # IDE Specifc
 ############
 *.idea
+
+# Docker local development #
+############################
+.python_env
--- a/.gitignore
+++ b/.gitignore
@@ -59,7 +59,7 @@ src/vendor
 *.swp
 
 # Python #
-#################
+##########
 *.pyc
 __pycache__/
 *.eggs/
@@ -76,13 +76,13 @@ vendor/
 .floo*
 
 # Documentation #
-##########
+#################
 docs/scripts
 docs/*/slick.html
 docs/*/pdfreactor-fonts.html
 
 # Django #
-#################
+##########
 *.egg-info
 .installed.cfg
 /static/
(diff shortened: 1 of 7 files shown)


rosskarchner:
I agree about .python_env--

The other change would break things for people who *aren't* using Docker machine (say, a user with the official Docker for mac installed, or a Linux user). I think that's worth avoiding? Maybe it's acceptable, though

In the docs, I brought up the possibility of using autoenv for that.

willbarton:
@rosskarchner but aren't shell.sh and attach.sh only useful with docker? How does adding the eval there break non-docker folks?

willbarton:
I've also added a proposed modification to `setup.sh` and `runserver.sh` so that they can be run as they are now, *or* with `docker` as the first argument. 

If `setup.sh` is run as `setup.sh docker` then the `mac-virtualbox-init.sh` script is run. 

If `runserver.sh` is run as runserver.sh docker` then the `docker-compose up` is run. 

@rosskarchner, what do you think of that? That'd make the broad-strokes of the current workflow similar for both approaches.

rosskarchner:
not "Non-Docker", but "Non-Docker-Machine".

Docker Machine is only a necessary piece where you need to run the docker daemon in a VM. The official "Docker for Mac" does some other magic, which uses the native OS X hypervisor. On Linux (and I think, Windows, in some scenarios at least) Docker runs natively and Docker Machine isn't neccessary.

rosskarchner:
I like the runserver and setup changes!

rosskarchner:
Should update the docs, tho ;)

willbarton:
Alrighty, updated the documentation and added the frontend setup to running `setup.sh docker`.

PR #3596, opened 2017-11-21
Spike: Implement FragmentCacheExtension to cache post previews

richaagarwal:
Certain pages, like blog and newsroom, make a ton of queries to generate each post's preview.  On any given day, the majority of these posts have not changed, and yet any time we deploy or invalidate the full Akamai cache for any other reason, we are regenerating all of these post previews.  This PR implements a cache that allows us to cache fragments of templates, and then invalidate them with the `page_published` Wagtail hook. The entirety of the `FragmentCacheExtension` class is copied and pasted from here -http://jinja.pocoo.org/docs/2.10/extensions/.  I couldn't find an existing implementation of that to import from but if you know of any, I'm all ears!

If you'd like to see this in action:
- Pull down this branch and run `python cfgov/manage.py createcachetable post_preview_cache` to first create the database table the cache will be written to. 
- Then run the server with the Django Debug Toolbar enabled: `ENABLE_DEBUG_TOOLBAR=1 ENABLE_POST_PREVIEW_CACHE=1 ./runserver.sh`
- Then, browse to a page like http://localhost:8000/about-us/blog/, and notice the large number of queries in the sidebar. 
- Refresh it, and notice the queries have gone down significantly. 
- Then, in a separate tab, log into the admin and make a change to one of the pages in the results (adding a tag is easiest), publish it, and refresh http://localhost:8000/about-us/blog/ again.  You'll notice the content has refreshed, but the number of queries has gone only slightly up. 

TODO:
- Add tests


chosak:
This is very cool! Some high-level comments:

How does this template tag differ from [the build-in Django `cache` tag](https://docs.djangoproject.com/en/1.8/topics/cache/#template-fragment-caching)? Is it just that Django's uses Django's cache framework whereas this uses Werkzeug's? I'm not that familiar with Werkzeug but I'd be curious about the differences/advantages of using [their memory-based cache](http://werkzeug.pocoo.org/docs/0.12/contrib/cache/#werkzeug.contrib.cache.SimpleCache) instead of Django's - it'd be nice to use Django's built-in if possible.

A minor design thing, it might be nice to implement the cache key logic as a property on the page class (`CFGOVPage.post_preview_cache_key`). That'd avoid the duplication of logic you have [here](https://github.com/cfpb/cfgov-refresh/pull/3596/files#diff-e64c1122508cbcf9a1de90e7879c62f9R66) and [here](https://github.com/cfpb/cfgov-refresh/pull/3596/files#diff-6e6318491f8b96066b45ee5267a71e6fR45).

It'd be nice if the caching were optional, i.e. only enabled in production, or configurable via environment variable (with a fallback to a null cache) to allow for testing without it.

richaagarwal:
Some more context on how I got to this PR:

- Why post previews, and not something like, an individual attribute?  I had actually looked into caching individual attributes before, but there isn't a real path forward there since we don't have any kind of hook (like a post_save method the way you get with vanilla Django models) that would allow us to invalidate them appropriately.  Because the entirety of the logic for a post preview is currently in a template (not necessarily ideal, but what it is), it seemed useful to look into how to cache templates.
- Why not Django's built-in cache tag, as Andy asks above?  Unfortunately, if you're using jinja in your Django app, you don't have direct access to Django's template tags.  There are workarounds & libraries, but they're not well documented and it's unclear that the setup for that would be worth it over what we have here in this PR.  
- There's something similar to the `FragmentCacheExtension` here https://github.com/niwinz/django-jinja/blob/master/django_jinja/builtins/extensions.py#L88, that I looked into importing, without all the set-up for django-jinja, but I could not get it to work with the cache I specified in our settings. It seems like, again, you'd have to configure much more to get this working with another library. 
- As noted above, I do think it'd be nice if we didn't have to duplicate the code from Werkzeug's example here, and if anyone knows of a similar method that we can easily import, I think that'd be ideal. But, in the meantime, I haven't found much that looks promising, and I think if we do find something better we can always swap that code out.

cfgov/cfgov/settings/base.py
```diff
@@ -650,4 +650,4 @@
 
 # If server's clock drifts from NTP by more than specified offset
 # (in seconds), check_clock_drift will fail
-MAX_ALLOWED_TIME_OFFSET = 5
+MAX_ALLOWED_TIME_OFFSET = 5
```
richaagarwal: No idea why this is showing up in the diff.

cfgov/cfgov/settings/production.py
```diff
@@ -98,12 +97,16 @@
...
         'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
         'LOCATION': 'api_cache_memory',
         'TIMEOUT': 3600,
         'OPTIONS': {
             'MAX_ENTRIES': 1000,
         },
+    },
+    'post_preview': {
+        'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
+        'LOCATION': '/tmp/post_preview_cache',
```
richaagarwal: I'm not sure whether this directory needs to be created first or if it will automatically be created.  @rosskarchner ?

cfgov/cfgov/settings/base.py
anselmbradford: Could your editor be set to remove trailing new lines automatically?

cfgov/cfgov/settings/production.py
rosskarchner: I don't think it needs to be created automatically-- we aren't doing that for the other cache directories

richaagarwal:
@rosskarchner Oh, that's a good point about code changing the post previews.  But... how do you feel about handling that as a one-off case?  I guess I'm worried we'll lose the performance benefits if we clear this every time we deploy, since 99.9% of the time it wouldn't impact that content, right?

rosskarchner:
That's a good counter-argument: they probably won't change often.

Making it a checkbox in the pipeline would work

cfgov/v1/signals.py
```diff
@@ -35,3 +38,10 @@ def user_save_callback(sender, **kwargs):
         current_password_history = user.passwordhistoryitem_set.latest()
         if user.password != current_password_history.encrypted_password:
             new_phi(user)
+
+
+def invalidate_post_preview(sender, **kwargs):
+    instance = kwargs['instance']
+    caches['post_preview'].delete(instance.post_preview_cache_key)
+
+page_published.connect(invalidate_post_preview)
```
chosak: This call will only run on a single server -- the one the publish happens on (content) -- and so won't invalidate the file-based caches on any of the other servers. Unless the servers share some central cache, we need a way to invalidate everywhere, otherwise those other servers will continue to serve stale content.

cfgov/cfgov/settings/production.py
```diff
@@ -98,12 +97,16 @@
...
-    'api_cache':{
+    'api_cache': {
         'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
         'LOCATION': 'api_cache_memory',
         'TIMEOUT': 3600,
         'OPTIONS': {
             'MAX_ENTRIES': 1000,
         },
+    },
+    'post_preview': {
```
chosak: Without a `TIMEOUT` setting this will [default to 5 minutes](https://docs.djangoproject.com/en/1.8/topics/cache/#cache-arguments). Unless this is set longer, will we get much benefit?

cfgov/v1/signals.py
richaagarwal: Updated to use a database cache in 574d22f23f47652827a29526a0c697332c2ca5b5

cfgov/cfgov/settings/production.py
richaagarwal: Nice catch! Updated to `None`.

richaagarwal:
I've addressed all feedback so far (most notably, updated the cache storage to store in a database table so that it's shared across all production servers) and plan to merge this soon unless there's any additional feedback. 

@rosskarchner I will add a management command in a later PR to easily clear the cache, that can be called from Jenkins, but I don't think it should be a blocker for now.

cfgov/cfgov/settings/test.py
```diff
@@ -24,3 +24,15 @@
         'WIDGET': 'wagtail.tests.testapp.rich_text.CustomRichTextArea',
     },
 }
+
```
chosak: Since `settings.test` imports all from `settings.local` (including the default DummyCache for all of the caches), what's the reason for redefining the caches like this?

richaagarwal: Oh, oops -- this was my mistake.  I had assumed it imports from base and didn't look at the first line in this file clearly! Will update.

richaagarwal:
Merged and will create the database table in prod.

PR #3599, opened 2017-11-21
Add self-hosted fonts option

anselmbradford:
Adds self-hosted font update from CF.

## Removals

- `static-legacy/css/licensed-fonts.css`

## Changes

- Replaces `gulp-mq-remove` with `postcss-unmq`.
- Replaces `gulp-autoprefixer` with `autoprefixer`.
- Brings in latest CF updates.

## Testing

1. `./frontend.sh production` should pass.
2. Vising the site locally should be pulling the fonts from fonts.net.
3. change `@use-font-cdn: true;` to `false` inside `main.less` and re-build the assets with `gulp clean && gulp build` and the fonts should be trying to pull from localhost.
4. Check in an IE8 VM to ensure the page isn't broken looking.

## Notes

- `gulp-mq-remove` has a dependency `css-parse` that was failing on the new conditional code inside cf-typography for the font-face rule. Both of those modules haven't been updated in years, so I went with replacing them with a postcss option.

## Todos

- Still more work needs to be done in ansible on the backend to switch `@use-font-cdn: true;` to false.


gulp/tasks/styles.js
```diff
@@ -76,13 +77,12 @@ function stylesIE8() {
...
     .on( 'error', handleErrors )
-    .pipe( gulpAutoprefixer( {
-      browsers: BROWSER_LIST.ONLY_IE_8
-    } ) )
-    .pipe( mqr( {
-      width: '75em'
-    } ) )
-    // mqr expands the minified file
+    .pipe( gulpPostcss( [
+        postcssUnmq( {
```
jimmynotjim: Shouldn't this have been in another PR?

anselmbradford: See note. `css-parse`, which is a dependency of `gulp-mq-remove`, couldn't handle https://github.com/cfpb/capital-framework/blob/canary/src/cf-typography/src/licensed-fonts.less for some reason. Since it hasn't been updated in three years my guess is there is some newer syntax less outputs that it doesn't know what to do with. You can replicate this by bumping `cf-typography` to the latest on `master` and running `./frontend.sh production`.

jimmynotjim: 🤦‍♂️ Sorry, I should have paid more attention to the opening comment.

cfarm:
What were `gulp-mq-remove` and `css-pars` doing? How can I test locally to confirm that the `postcss` option does that thing?

cfgov/unprocessed/css/main.less
```diff
@@ -21,7 +21,13 @@
...
 
 // Icon font path
-@cf-icon-path:                  '../fonts';
+@cf-icon-path: '../fonts';
+
+// Webfont variables
+// This is the path for self-hosted fonts.
+@cf-fonts-path: '/static/fonts';
+// Whether or not to override the local path and retrieve fonts from fonts.net.
+@use-font-cdn: true;
```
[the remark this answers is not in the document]
Scotchester: Sorry again, just saw #3601. I was thrown off by your comment in this PR that it needed to be done on the back-end.

PR #3601, opened 2017-11-22
Turn on self-hosted fonts

anselmbradford:
Flips switch for `@use-font-cdn` so production pulls from self-hosted fonts, not fonts.net.

## Changes

- Flips switch for `@use-font-cdn` to false, so that in production fonts are pulled from consumerfinance.gov/static/fonts

## Testing

1. `gulp build`
2. Check the site in localhost.
3. Check JS console to see that fonts are trying to be loaded from localhost.
4. ~~Contact @anselmbradford @Scotchester to get the fonts so you can install them locally.~~ ~~Download them from [GHE]/CFGOV/cfgov-fonts/~~
5. ~~Place font files in /cfgov/static_built/fonts~~
6. cd `/static.in/`
7. `git checkout [GHE]/CFGOV/cfgov-fonts/`
8. Restart the server.
9. Visit the page and see that the fonts load locally.

## Notes

-  ~~You'll need tokeep the fonts handy in a directory other than cfgov-refresh, as `gulp clean` will wipe out /cfgov/static_built/fonts.~~ Not applicable with `static.in`

### DIFF
--- a/cfgov/unprocessed/css/main.less
+++ b/cfgov/unprocessed/css/main.less
@@ -27,7 +27,7 @@
 // This is the path for self-hosted fonts.
 @cf-fonts-path: '/static/fonts';
 // Whether or not to override the local path and retrieve fonts from fonts.net.
-@use-font-cdn: true;
+@use-font-cdn: false;
 
 // Enhancements that are on a migration path to being added to CF.
 // CF Core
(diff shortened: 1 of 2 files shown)


anselmbradford:
Updated the documentation. `mkdocs gh-deploy --clean` should be run after this is merged.

PR #3602, opened 2017-11-24
Sort imports per dev standards

willbarton:
This PR adds an `isort --check-only --diff` call to our Travis script. This lets us lint our Python imports for compliance with PEP8 and [Django's import style](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#imports).

As part of this I've included an isort configuration file that should enforce those styles of imports and I've run `isort --recursive cfgov` over the codebase to adjust our imports to match. 

https://github.com/cfpb/cfgov-refresh/commit/bb97cbb140eae9889107bbe248039611b4e22ed9 has the isort configuration if you'd like to look independent of the fixing-imports changeset.

## Additions

- `isort` configuration
- `isort --check-only --diff` to Travis so that imports will be linted and corrections suggested.

## Changes

- Import orders in all relevant python files to comply with isort rules in `.isort.cfg`

## Testing

1. `git checkout sort-imports`
2. `pip install -r requirements/test.txt`
3. `git checkout master`
4. `isort --check-only --diff --recursive cfgov`
5. Notice lots of and lots of failures
6. `git checkout sort-imports`
7. `isort --check-only --diff --recursive cfgov`
8. Notice no failures!

## Checklist


.isort.cfg
```diff
@@ -0,0 +1,15 @@
...
+lines_after_imports=2
+multi_line_output=3
+skip=.tox,migrations
+not_skip=__init__.py
+use_parentheses=1
+known_django=django,haystack
+known_wagtail=wagtail
+known_future_library=future,six
+known_third_party=mock
+known_first_party=alerts,ask_cfpb,cfgov,core,data_research,jobmanager,search,v1,wellbeing
```
chosak: I worry this might be forgotten the next time someone adds a new project. Not sure there's an easy way to have this automatically count local directories as first party?

.isort.cfg
```diff
@@ -0,0 +1,15 @@
+[settings]
+combine_as_imports=1
+include_trailing_comma=1
```
chosak: Default for `include_trailing_comma` is false, any particular reason you set it? I have a week preference against.

willbarton: Primarily to avoid 2 lines of diffs instead of one if/when we add to import lists.

I have a weak preference for the comma of hope.

.isort.cfg
willbarton: Without this the sorting was a bit weird sometimes, but it'd get 90% there. I can experiment a bit more.

.isort.cfg
chosak: Ah, looking again I see this PR uses [multi line output mode](https://github.com/timothycrosley/isort#multi-line-output-modes) 3, which does this:

```py
from third_party import (
    lib1,
    lib2,
    lib3,
)
```

in which case that would require a two-line diff. I will express a weak preference for the much more concise mode 5 ([which Django uses](https://github.com/django/django/blob/master/setup.cfg#L16)), which does this:

```py
from third_party import (
    lib1, lib2, lib3
)
```

That being said, I defer to the person actually who does the work to open the PR and fix all of the existing code to match a consistent standard :)

willbarton:
@chosak @richaagarwal @rosskarchner @hillaryj I've switched to multi-line import mode 5. I think this is ready for a final review, then I'll  add some quick documentation about running `isort` and then rebase... again.

PR #3603, opened 2017-11-24
Remove misleading form_id from filterable params

richaagarwal:
At some point, we were supporting multiple filter controls on the same page, thus the need to assign them a form ID and ensure each form displays their respective data. However, we do not appear to have any pages with multiple filterable forms on them, and more importantly, if we were to, this doesn't work as intended -- the logic for how we append the form ID to authors vs. tags is in different places and actually results in different numbers (when you have multiple filterable forms).  For tags, we were always grabbing the index of the first filter controls on the page (regardless of if there were others), and for authors, we were grabbing the index of the block it's in directly within the template.  That's just one example -- this code was incredibly complex and supporting a use case we don't use.  This PR simply cleans that up and adjusts the tests as necessary. 

To see the above "bug", you can check out master (not this branch), and create a page with more than one FilterControls block. Or you can edit the blog page and add another FilterControls block.  Publish the page, view it, and scroll down to the second filterable section.  You'll notice the links in each post preview, for authors and for tags, have different numbers in them. Authors will be prepended with `filter2_`, and tags will be preprended with `filter1_`.  The correct behavior for supporting multiple forms should mean they both have `filter2_`.

Rather than fix this bug, I'm removing this partial support entirely, such that the filters are passed around by their name, `topics` instead of `filter0_topics`. The resulting URL after a GET request would look something like this: http://localhost:8000/about-us/blog/?title=&topics=financial-education&from_date=&to_date= instead of http://localhost:8000/about-us/blog/?form-id=1&filter1_title=&filter1_topics=financial-education&filter1_from_date=&filter1_to_date=

By cleaning this up, it will be easier to add test coverage for the impacted functions (something I started trying to do, but then ended up working on this PR instead), and my hope is it will be more obvious to other developers what this code is doing.

### DIFF
--- a/cfgov/jinja2/v1/_includes/article.html
+++ b/cfgov/jinja2/v1/_includes/article.html
@@ -20,7 +20,7 @@
                        '/about-us/blog/'.  Remember to leverage vars.path
                        instead of using the literal string '/about-us/blog/'.
                        Path is used to create the filtered URL:
-                       {{ path }}?filter_tags={{ tag }}
+                       {{ path }}?tags={{ tag }}
 
    ========================================================================== #}
 
--- a/cfgov/jinja2/v1/_includes/macros/category-slug.html
+++ b/cfgov/jinja2/v1/_includes/macros/category-slug.html
@@ -30,7 +30,7 @@
 
    ========================================================================== #}
 
-{% macro render(category, href, classes, use_blog_category=false, index=0) %}
+{% macro render(category, href, classes, use_blog_category=false) %}
     {% import 'macros/category-icon.html' as category_icon %}
 
     {% if href %}
@@ -38,7 +38,7 @@
         {% if use_blog_category %}
             {% set href = href + '?filter_blog_category=' + category | urlencode | replace('%20', '+') %}
         {% else %}
-            {% set href = href + '?filter' + index | string + '_categories=' + category | urlencode | replace('%20', '+') %}
+            {% set href = href + '?filter_categories=' + category | urlencode | replace('%20', '+') %}
         {% endif %}
     {% endif %}
 
--- a/cfgov/jinja2/v1/_includes/macros/category-slug.html
+++ b/cfgov/jinja2/v1/_includes/macros/category-slug.html
@@ -18,7 +18,7 @@
                        Remember to leverage vars.path instead of
                        using the literal string '/about-us/blog/'.
                        Path is used to create the filtered URL:
-                       {{ href }}?filter_category={{ category }}
+                       {{ href }}?category={{ category }}
 
    classes (optional): Space separated list of class names.
 
@@ -36,9 +36,9 @@
     {% if href %}
         {# TODO: Remove use_blog_category parameter when this element becomes atomic. #}
         {% if use_blog_category %}
-            {% set href = href + '?filter_blog_category=' + category | urlencode | replace('%20', '+') %}
+            {% set href = href + '?blog_category=' + category | urlencode | replace('%20', '+') %}
         {% else %}
-            {% set href = href + '?filter_categories=' + category | urlencode | replace('%20', '+') %}
+            {% set href = href + '?categories=' + category | urlencode | replace('%20', '+') %}
         {% endif %}
     {% endif %}
 
--- a/cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
+++ b/cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
@@ -13,22 +13,21 @@
 
    form:     Django form that carries the fields that are to be rendered.
 
-   index:    A unique number given to render the form and its fields with.
 
    ========================================================================== #}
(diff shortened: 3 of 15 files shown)


anselmbradford:
Can we stop people from being able to add two filters to a page? What happens if they do?

chosak:
@richaagarwal thank you for looking into this! This code has long been in the back of my mind as something that deserves a refresh.

One concern about changing this: what about bookmarks/social media links that are out in the wild with the old URL query scheme? I'd support a way to make this a little nicer but would be concerned about breaking any existing links, at least without some kind of redirect.

hillaryj:
As a note, there's a Filterable List refactoring project upcoming.

richaagarwal:
@anselmbradford We don't stop people from doing it.  As mentioned, it currently doesn't work if you attempt this (there are actually a bunch of reasons why, but one of them I detail in my PR description above). With this PR, it still wouldn't work (though a little more obviously, since there wouldn't be a form ID at all).

@chosak Yeah, that makes sense. On the one hand I feel a little bit mixed about all the redirects we have, but I understand the desire to support previous URL structures. What if we had one redirect that just looked for a form-id param, and redirected that to the page without any params, so something like https://www.consumerfinance.gov/about-us/blog/?form-id=1&filter1_title=&filter1_topics=mortgages&filter1_from_date=&filter1_to_date= would redirect to https://www.consumerfinance.gov/about-us/blog/ ?  That feels less like a good middle ground of redirecting them to a page that works, without hardcoding the param structure in too much.

chosak:
@richaagarwal what to do about existing URLs feels like a PM decision to me.

jimmynotjim:
@richaagarwal @chosak that's essentially what happens if the query string doesn't match any of the form input ids. I don't know that the overhead of setting up redirects is worth supporting when most people aren't even going to understand what that query string is when the end result is the same thing.

Ex https://www.consumerfinance.gov/about-us/blog/?form-id=1&title=&topics=mortgages&from_date=&to_date=

schaferjh:
@richaagarwal thanks for bringing this one to my attention. @ascott1 FYI

I'm less concerned about bookmarks or social media linked to a pre-filtered URL and more concerned about those links within our own site. 
- To my nontechnical eye, it looks like the links created by the "related posts" module (i.e. https://www.consumerfinance.gov/activity-log/?filter0_topics=banking&filter0_topics=checking-account&filter0_topics=overdrafts) would be unaffected by this refactor?
- But one like the reference found in the sidebar of https://www.consumerfinance.gov/consumer-tools/getting-an-auto-loan/  to "Read more about auto lending issues for consumers in our auto loan blog posts." would break.

I'm not sure how many instances of this occur, but this would be a situation where redirecting to the blog without filters would be not-so-helpful for users. Helpfully, it could be easier to address than social media or bookmark issues -- Could we use our site crawler to identify where those links are and change them individually to fit the new schema?

ascott1:
>  Could we use our site crawler to identify where those links are and change them individually to fit the new schema?

Overall this looks very good and I much prefer the proposed URL structure, but I want to +1 this question from @schaferjh. Being able to reliably ensure support for any links on our own site is critical.

richaagarwal:
@schaferjh Thanks for your feedback!  It would actually impact the links created by the related posts module -- I removed all instances of where form-id gets used.

It's possible a full redirect (w/ params) would be the easiest way to handle this but I'll check with @sebworks about our site crawler's capability. Thanks!

richaagarwal:
@schaferjh, @ascott1 : @sebworks provided me with this list of pages that have the hardcoded URLs in question. If we want to proceed, we'd just need to stage these edits and then publish after this change goes live. 
```
title-xiv-mortgage-rules
financial-coaching
key-terms
how-to-get-your-credit-reports
what-to-look-for-in-credit-reports
difference-between-credit-scores-and-credit-reports
credit-application-denied-because-credit-report
requesting-credit-reports
dispute-inaccurate-information
not-enough-credit-history-build-credit-record
request-consumer-reports
victim-fraud-identity-theft
build-good-credit-score
credit-repair-scam
sample-letters-dispute-credit-report-information
state-financial-resources-servicemembers
state-financial-resources-servicemembers
getting-an-auto-loan
military-financial-lifecycle
resources-for-older-adults
credit-reports-and-scores
student-loans
adult-financial-education
financial-well-being-survey-data
```

schaferjh:
@richaagarwal ok -- that seems like a manageable number of links to change. Probably most efficient to document that as a task over in GHE.

What about updates to all the related post links -- how will that be handled?

richaagarwal:
@schaferjh Sounds good.  What do you mean by updates to the related post links?  The links that aren't hardcoded are generated through the code, so they'd be updated once this PR goes live.

schaferjh:
> The links that aren't hardcoded are generated through the code, so they'd be updated once this PR goes live.

@richaagarwal ahhhh ok, no worries then. Will just need to make sure the hardcoded links are updated in concert with this PR going live.

PR #3625, opened 2017-11-30
Switch from gulp-changed to gulp-newer in style and script tasks

Scotchester:
This will enable us to check more source files than what is being passed through the pipe, fixing the bug where assets aren't rebuilt if required/imported JS/CSS is updated.

## Removals

- Unused `lib` config

## Changes

- Replaces gulp-changed step with gulp-newer in style and script tasks
- Updates cf-atomic-component and cf-expandable dependencies

## Testing

1. Pull branch
1. `npm install`
1. `gulp clean`

### Testing style tasks

1. Run `gulp styles:modern` to establish a baseline time
1. Make a trivial edit to a file that is imported into `main.less`, such as `breadcrumbs.less`
1. Run `gulp styles:modern` and see a similar time as before
1. Immediately re-run `gulp styles:modern` and see that it finishes in under 100ms!
1. Repeat similarly with other subtasks, as desired

### Testing script tasks

1. Run `gulp scripts:modern` to establish a baseline time
1. Make a trivial edit to a file that is require into `common.js`, such as `modules/focus-target.js`
1. Run `gulp scripts:modern` and see a similar time as before
1. Immediately re-run `gulp scripts:modern` and see that it finishes in under 100ms!
1. Repeat similarly with other subtasks, as desired

## Notes

- By default, gulp-newer will compare the destination file's modified time against:
  - All the CSS and JS in our `cfgov/unprocessed/` folder
  - The `node_modules/` folder (whose timestamp gets updated any time a dependency is updated)
  - JS files in `config/` and `gulp/`

## Todos

- Migrate other gulp-changed usage (in `copy` and `images` scripts) to gulp-newer?

## Checklist


sebworks:
Let's write some test for some of this stuff. It doesn't have to be now and I can certainly help.

package.json
```diff
@@ -21,10 +21,10 @@
     "babel-loader": "7.1.2",
     "babel-preset-env": "1.6.1",
     "browser-sync": "2.18.13",
-    "cf-atomic-component": "1.5.3",
+    "cf-atomic-component": "1.5.4",
```
sebworks: did you intend to bump these?

gulp/tasks/scripts.js
```diff
@@ -21,23 +24,32 @@ const paths = require( '../../config/environment' ).paths;
 const webpack = require( 'webpack' );
 const webpackConfig = require( '../../config/webpack-config.js' );
 const webpackStream = require( 'webpack-stream' );
-const configLegacy = require( '../config.js' ).legacy;
+
+const NEWER_EXTRAS = [
```
sebworks: Can we change the name to something like `ADDITIONAL_SCRIPTS` or `ADDITIONAL_FILES`?

sebworks: Also, we should probably just move this to the config file.

gulp/tasks/styles.js
```diff
@@ -10,23 +10,42 @@ const configStyles = config.styles;
...
 const gulpLess = require( 'gulp-less' );
+const gulpNewer = require( 'gulp-newer' );
 const gulpPostcss = require( 'gulp-postcss' );
 const gulpRename = require( 'gulp-rename' );
 const gulpSourcemaps = require( 'gulp-sourcemaps' );
 const handleErrors = require( '../utils/handle-errors' );
+const paths = require( '../../config/environment' ).paths;
 const postcssUnmq = require( 'postcss-unmq' );
 
+const NEWER_EXTRAS = [
```
sebworks: Same here: move to the config.

sebworks:
If you move the `webpack-config.js` will this catch the change?

sebworks:
LGTM, minor stuff.

sebworks:
@jimmynotjim, I'm just thinking we should have some smoke tests to ensure our build process works as expected. You probably get the most benefit with bumping build dependencies.

package.json
Scotchester: Yeah. I bumped them to check that it was correctly paying attention to `node_modules`, and figured might as well leave them in here.

gulp/tasks/scripts.js
Scotchester: > Can we change the name to something like `ADDITIONAL_SCRIPTS` or `ADDITIONAL_FILES`? I like the connection to the property name in the gulp-newer options, but it could be clearer. How about `EXTRA_FILES_FOR_GULP_NEWER`?

sebworks: No, because if you don't use `gulp_newer` then the variable name no longer makes sense. Especially if you move the variable into a config file.

Scotchester: > Also, we should probably just move this to the config file.

I disagree with this. This variable will only ever be used within this task file. Moving it to config just adds that much more overhead when debugging or just trying to understand this setup in the future.

Scotchester: > No, because if you don't use `gulp-newer` then the variable name no longer makes sense. Especially if you move the variable into a config file.

Good point, but `ADDITIONAL_FILES` isn't clear enough. Additional files for what?

Scotchester:
I'm definitely into the idea of having tests, but no idea where to begin on that. Let's continue that conversation after this PR.

gulp/tasks/scripts.js
sebworks: >> I disagree with this. This variable will only ever be used within this task file. Moving it to config just adds that much more overhead when debugging or just trying to understand this setup in the future.

right, but this would be the only config object in the gulp files and the code isn't reusable. We should stay with the existing convention and move them to the config file.

Scotchester: OK, moved these things to config and renamed them to be tool-agnostic in a3f4425.

PR #3628, opened 2017-12-01
Remove form_id from filterable params

richaagarwal:
See https://github.com/cfpb/cfgov-refresh/pull/3603 for context -- I had merged that PR but then noticed a bug, so had to revert it.  I'm creating a new commit/ PR here (with the fix) since git otherwise doesn't handle reverting reverts smoothly.

Though the change is limited in scope to removing form_id from filterable params, this touches a lot of files that pass it around (and not always in an obvious way, for instance the bug I hadn't noticed before was that the `render` method on `organisms/filterable-list-controls.html` was still being passed a form ID from `newsroom/index.html`, but it's referred to as `flc` in many places and the form ID might be passed around as a `loop.index` or something like that.  Basically, the code in question is brittle and I don't have a lot of confidence in our existing tests which didn't catch this.  I'm open to pushback that we shouldn't merge this until more test coverage is added, but then again our smoke tests caught it, so that might be an argument for bringing smoke tests into our pre-deploy (rather than post-deploy) checks.

A final BEWD (or FEWD) review is appreciated!

### DIFF
--- a/cfgov/jinja2/v1/_includes/macros/category-slug.html
+++ b/cfgov/jinja2/v1/_includes/macros/category-slug.html
@@ -18,27 +18,24 @@
                        Remember to leverage vars.path instead of
                        using the literal string '/about-us/blog/'.
                        Path is used to create the filtered URL:
-                       {{ href }}?filter_category={{ category }}
+                       {{ href }}?category={{ category }}
 
    classes (optional): Space separated list of class names.
 
    use_blog_category (optional): Whether to use the blog category filter or not.
                                  Defaults to false.
 
-   index (optional):             A unique ID for a form filter URL.
-                                 Defaults to 0.
-
    ========================================================================== #}
 
-{% macro render(category, href, classes, use_blog_category=false, index=0) %}
+{% macro render(category, href, classes, use_blog_category=false) %}
     {% import 'macros/category-icon.html' as category_icon %}
 
     {% if href %}
         {# TODO: Remove use_blog_category parameter when this element becomes atomic. #}
         {% if use_blog_category %}
-            {% set href = href + '?filter_blog_category=' + category | urlencode | replace('%20', '+') %}
+            {% set href = href + '?blog_category=' + category | urlencode | replace('%20', '+') %}
         {% else %}
-            {% set href = href + '?filter' + index | string + '_categories=' + category | urlencode | replace('%20', '+') %}
+            {% set href = href + '?categories=' + category | urlencode | replace('%20', '+') %}
         {% endif %}
     {% endif %}
 
--- a/cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
+++ b/cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
@@ -184,9 +184,8 @@
 {% macro render(controls) %}
     <div class="o-filterable-list-controls"
          id="o-filterable-list-controls">
-        {% set form = filter_data.forms.pop(0) %}
-        {% set posts = filter_data.page_sets.pop(0) %}
-        {% set has_active_filters = request.GET %}
+        {% set form = filter_data.form %}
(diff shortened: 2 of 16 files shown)


cfgov/jinja2/v1/_includes/macros/category-slug.html
```diff
@@ -30,15 +30,15 @@
 
    ========================================================================== #}
 
-{% macro render(category, href, classes, use_blog_category=false, index=0) %}
```
sebworks: You can remove the index help info from the template.

cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
```diff
@@ -183,16 +181,16 @@
...
 
-{% macro render(controls, index) %}
+{% macro render(controls) %}
     <div class="o-filterable-list-controls"
-         id="o-filterable-list-controls-{{ index | string }}">
+         id="o-filterable-list-controls">
         {% set form = filter_data.forms.pop(0) %}
         {% set posts = filter_data.page_sets.pop(0) %}
-        {% set has_active_filters = page.has_active_filters(request, index) %}
+        {% set has_active_filters = request.GET %}
```
sebworks: is it possible that there might be query string params not associated with the filters ?

sebworks:
💯, looks good to me. Thanks for taking this on @richaagarwal.

cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
chosak: Agree, this feels like an oversimplification. It's likely there aren't usually other query string parameters but this is not a safe assumption to make.

cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
```diff
@@ -183,16 +181,16 @@
 {% from 'molecules/notification.html' import render as notification with context %}
 {% import 'organisms/post-preview.html' as post_preview with context %}
 
-{% macro render(controls, index) %}
+{% macro render(controls) %}
     <div class="o-filterable-list-controls"
-         id="o-filterable-list-controls-{{ index | string }}">
+         id="o-filterable-list-controls">
         {% set form = filter_data.forms.pop(0) %}
```
sebworks: I'm not sure if this necessary anymore.

richaagarwal: @sebworks Are you saying, you're not sure if it's safe to remove?

cfgov/jinja2/v1/sublanding-page/index.html
```diff
@@ -8,7 +8,7 @@
 {% block hero -%}
     {% for block in page.header -%}
         {% import 'templates/render_block.html' as render_block with context %}
-        {{ render_block.render(block, loop.index) }}
+        {{ render_block.render(block) }}
```
richaagarwal: This is wrong -- will change back, as well as all other updates to form_block and render_block that I incorrectly made.

cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
sebworks: @richaagarwal is there a reason to use `pop`?

sebworks: Spoke with @richaagarwal and code is good.

cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
richaagarwal: This is fixed in https://github.com/cfpb/cfgov-refresh/pull/3628/commits/1bc68c6ea7f779bf6c373d888c24eeebc2f78df6

PR #3635, opened 2017-12-05
POC: Separate out audit gulp task and add Google Lighthouse performance audit

anselmbradford:
Proof of concept for providing a command-line tool for displaying page loading performance metrics.
There is a bit of overlap between Pagespeed Insights and Lighthouse, but the general difference is PSI provides individual factors and Lighthouse provides aggregate factors to show page speed. We're most interested in time to first meaningful paint.

## Additions

- Adds `gulp audit:perf:lighthouse` task.

## Changes

- Moves audit tasks (accessibility and performance) to their own gulp file.

## Testing

1. Pull down this branch and run `npm install`
2. Run `gulp audit:perf:lighthouse`.
2. Existing tasks should work: `gulp audit:a11y`, `gulp audit:pef`, `gulp test`.

## Screenshots

screen shot 2017-12-05 at 1 35 37 am

## Notes

- I wasn't able to combine PSI and Lighthouse together under `gulp audit:perf` because the task would end after PSI ran. I think because it exits the process.

- This configuration compares localhost against www, which is an unfair comparison, however, if localhost is slower that may indicate something seriously wrong. At the moment the "network payloads" metric is always slower on localhost, which has something to do with gzipping on the prod server I'm guessing? So we may want to exclude that from localhost.


anselmbradford:
Closing as this was just a proof of concept.

PR #3640, opened 2017-12-06
Match-all-tags page filtering

higs4281:
Editors helped us discover a combination of page tags and the use of
"match-all-topic-tags" that locked up the version of mysql that we run
on servers. This PR reduces some of the query complexity and, in the case
of "match-all," forces a query evaluation before starting the recursive
tag filtering routine.

I replicated the mysql lockup by editing page #10848 inside a docker VM that 
uses with the same db version used in production. When I checked the 
"match-all-topic-tags" box, the page locked up mysql when it was rendered. 
This PR stopped the lockups for me.

#### Notes
When the final query succeeded, applying "match-all" to all 9 tags, 
the result was zero related pages. All this trouble for a null!

This addresses GHE issue 1805

### DIFF
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -150,8 +150,9 @@ class CFGOVPage(Page):
         from v1.models.learn_page import AbstractFilterPage
 
         def match_all_topic_tags(queryset, tags):
+            """The narrowest search: match every given tag"""
             for tag in tags:
-                queryset = queryset.filter(tags__name=tag)
+                queryset = queryset.filter(tags=tag)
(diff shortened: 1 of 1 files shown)


rosskarchner:
When checkout this branch, open up admin, and restore a problematic version of page 10848, I'm unable to preview or view the published page.

higs4281:
I'll try restoring a revision.

higs4281:
Rats. Restoring a revision brings back the problem for me, too. 
Back to the taggithole.

rosskarchner:
Been there!

PR #3642, opened 2017-12-06
Bootstrap Postgres support, and support database URL's

rosskarchner:
This should provide us with what we need to *start* work on transitioning apps to Postgres.

### In Django:

- Dependency additions to support Postgres, and [database URL's](https://github.com/kennethreitz/dj-database-url)
-  The application now supports a `PG_DATABASE_URL` environment variable, which if set to a valid database URL will configure a [second database](https://docs.djangoproject.com/en/1.8/topics/db/multi-db/) with the label 'postgres'
- Likewise, you can use the `DATABASE_URL` variable configure the 'default' DB. This takes precedence over the "old" variable style.
- If an app is listed in settings.POSTGRES_APPS (and the 'postgres' database is configured), then reads and writes will be routed to the postgres DB instead of the default. 

### In Docker:

- `compose up` will now also start a Postgres server
- The Python container has PG_DATABASE_URL configured to connect to the Postgres server
- a `./pg_loader_shell.sh` script that drops you into a container with [pgloader](https://pgloader.io/) and that can connect to the mysql and postgres containers. This is not meant to endorse or propose pgloader as part of our migration path, but it seems like a useful tool for experimentation.

### What's next?

Once the infrastructure is ready, we should set PG_DATABASE_URL in every relevant environment, and work to add apps to POSTGRES_APPS.

Once the whole site supports postgres, we should remove settings.POSTGRES_APPS  and simply set DATABASE_URL to the address of the postgres server.

## Removals

- DB Router's support for 'replica' databases.  We should revisit replica's once MySQL is phased out.

## Changes

-

## Testing

To get the updated Python bits, you'll need to `docker-compose build`

To check things end-to-end, you should be able to use `cfgov/manage.py dbshell --database postgres` (via shell.sh). You will be prompted for a password, which is an intentional feature of the psql command line tool. If you're using docker-compose, the password will be 'cfgov'

## Notes

-

## Todos

-

## Checklist


rosskarchner:
In theory, once in the pgloader shell you should be able to use:

`pgloader mysql://v1:v1@mysql/v1 psql://cfgov:cfgov@postgres/cfgov` to migrate the entire database. Every time I try it, it crashes, though.

higs4281:
:+1: This is great. I might have one clue about db migration.

We used [jsonfield](https://github.com/dmkoch/django-jsonfield) in the last data_research project, and it doesn't work as-is in Postgres. The fix is to start using the native Postgres JSONField, but that's not available before Django 1.9. 

Once we get to 1.9, the fix is simple. in cfgov/data_research.models, instead of
```
from jsonfield import JSONField
```

we can use

```
from django.contrib.postgres.fields import JSONField
```

rosskarchner:
What do you mean, `doesn't work as-is`? the readme for that project *recommends* the native Postgres field, but doesn't indicate that it's not supported.

rosskarchner:
This approach looks interesting: https://rlskoeser.github.io/2016/03/31/migrating-data-between-databases-with-django/

higs4281:
> What do you mean, doesn't work as-is?

You're right that they don't explicitly say it's not supported, but I crash when trying to use it as-is.

rosskarchner:
Could one of the people who gave a ❤️ or 👍 bless this with a ✅ ?

higs4281:
reviewing now!

PR #3646, opened 2017-12-06
Moves auditing tasks from test to gulp audit task

anselmbradford:
## Changes

- Breaks gulp `test.js` file apart into `test.js` and `audit.js`.

## Testing

1. `gulp audit:a11y`, `gulp audit:perf`, and `gulp test` should work as they did before.

## Notes

- Spun out of https://github.com/cfpb/cfgov-refresh/pull/3635


sebworks:
I'll be honest, I think unless the gulp audit tasks run on Travis or Jenkins they are a waist of time. People just don't run the tasks.

gulp/tasks/audit.js
```diff
@@ -0,0 +1,156 @@
...
+}
+
+/**
+ * Run PageSpeed Insight (PSI) executable.
+ * @param {Object} params - url, options, and tunnel for running PSI.
+ */
+function _runPSI( params ) {
+  gulpUtil.log( 'PSI tests checking URL: http://' + params.url );
+  psi.output( params.url, params.options )
+  .then( () => {
```
sebworks: We should probably start switching from using promises to async / await for this stuff.

PR #3650, opened 2017-12-07
Fix environment variable statement in Docker installation doc

cfarm:
Fixes missing `export` from `.python_env` example

also changes https to http since the provided links at GHE use http. I made these match so you won't be confused whether you're using the correct URL like I was.


rosskarchner:
Good call on the HTTPS-- but python_env is *not* a shell script, and doesn't require export statements.  

See: https://cfpb.github.io/cfgov-refresh/usage/#set-environment-variables

chosak:
@cfarm do you want to update this PR to only include the HTTP -> HTTPS fix, since the `export` isn't needed?

PR #3655, opened 2017-12-08
Add check_if_docker_installed function

anselmbradford:
## Changes

- Adds `check_if_docker_installed` function to tell the user that Docker isn't installed if it isn't.

## Testing

1. Without docker installed `./runserver.sh docker` should output `Docker not installed. See installation.md for installing docker.`


willbarton:
It seems like [`mac-virtualbox-init.sh`](https://github.com/cfpb/cfgov-refresh/blob/master/mac-virtualbox-init.sh) might be a better location for this... though it might also be better if that script were simply named `mac-docker-init.sh` or something of that nature.

@rosskarchner?

rosskarchner:
That makes sense to me-- it'd be a great check before it attempts to run `docker ps`.

I could go either way on the name change.

anselmbradford:
Okay moved the check. I also normalized the indent spacing on 2 spaces (there was a 2, 3, and 4). I'll leave renaming the file to another PR so changes can be seen easier.

PR #3658, opened 2017-12-11
Deprecate file cache for eRegs and default caches

chosak:
This change modifies the two eRegs caches (`eregs_longterm_cache` and `api_cache`) so that they use the passthrough cache instead of a disk-based cache. These two caches are required by the optional [regulations-site](https://github.com/cfpb/regulations-site) app. Unfortunately that app requires that these caches be defined (including with a `TIMEOUT`) parameter and doesn't handle the fallback case where they aren't.

eRegs pages don't need to be specially cached to disk because the typical production deployment of cfgov-refresh has a comprehensive cache sitting in front of it. These caches thus add complexity without adding any benefit.

This change also modifies the `default` cache so that it uses Django's default of in-memory, instead of disk (if you don't define any caches at all, [you don't need to define this in settings](https://docs.djangoproject.com/en/1.8/topics/cache/#local-memory-caching), but since we do other ones, we have to define a default). The way this cache is currently defined means that doing various things (like the built-in Wagtail caching of sites) requires write permission to `/tmp`, which is not always available.

Is there any particular reason we use a disk cache here instead of memory? @rosskarchner

## Changes

- Change eRegs caches to be pass-through and Django default cache to be in-memory.

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/local.py
+++ b/cfgov/cfgov/settings/local.py
@@ -49,6 +49,7 @@ if os.environ.get('ENABLE_DEBUG_TOOLBAR'):
 MIDDLEWARE_CLASSES += CSP_MIDDLEWARE_CLASSES
 
 # Disable caching when working locally.
+# eregs_longterm_cache and api_cache are required by regulations-site.
 CACHES = {
     k: {
         'BACKEND': 'django.core.cache.backends.dummy.DummyCache',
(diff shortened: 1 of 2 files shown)


chosak:
Is there anything else we use the Django default cache for, maybe in the satellite apps?

PR #3663, opened 2017-12-13
Cache activity log

richaagarwal:
Caches the previews on the activity log page: https://www.consumerfinance.gov/activity-log/

This is mostly an example for the documentation on caching I added, but could be merged in. It decreases (locally) the queries from 200 to 117, and the CPU time from 10 seconds to 5.6 seconds.  Since it's only a single page (and not a set of pages in the case of the post previews work), not sure it's all that worthwhile adding this extra code.

### DIFF
--- a/cfgov/jinja2/v1/activity-log/_activity-list.html
+++ b/cfgov/jinja2/v1/activity-log/_activity-list.html
@@ -16,6 +16,7 @@
                 </th>
             </tr>
         {%- for post in posts %}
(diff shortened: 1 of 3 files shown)


hillaryj:
@richaagarwal I agree it's probably not worth the extra code for one page. Should this PR be closed without merging?

richaagarwal:
@hillaryj Yeah, that sounds good - I just wanted this here as an example but it doesn't need to stay open!

PR #3673, opened 2017-12-14
Revert "Deprecate file cache for eRegs and default caches"

rosskarchner:
Reverts cfpb/cfgov-refresh#3658


chosak:
Let's hold off on reverting this for a bit so I can debug the current issues with that PR.

chosak:
For some reason this change seems to be causing the number of Apache processes to spike, maybe because the eRegs requests are taking such a long time. Going to merge this to get things back up and running.

PR #3675, opened 2017-12-14
Remove ESLint ignore file

anselmbradford:
## Removals

- `.eslintignore` file.

## Changes

- Runs autofix linting options on /route/ and class-list scripts.

## Testing

1. `gulp build` should pass.
2. Site should not have any errors in the console in IE9.

## Notes

- I discovered `<!--[if IE 9]>` doesn't trigger in Browser Mode: IE9 Compatibility View, but does in Browser Mode: IE9. Also, our logo is blown up in Browser Mode: IE9. Anyone up on which is the default?

screen shot 2017-12-14 at 5 02 26 pm

### DIFF
--- a/cfgov/unprocessed/js/routes/owning-a-home/common.js
+++ b/cfgov/unprocessed/js/routes/owning-a-home/common.js
@@ -2,20 +2,18 @@
    Common application-wide scripts for owning-a-home.
    ========================================================================== */
 
-var FormSubmit = require( '../../organisms/FormSubmit.js' );
-var validators = require( '../../modules/util/validators' );
-var ratingsForm = require( '../../apps/owning-a-home/ratings-form' );
-var BASE_CLASS = 'o-email-signup';
-var emailSignup = document.body.querySelector( '.' + BASE_CLASS );
+const FormSubmit = require( '../../organisms/FormSubmit.js' );
+const validators = require( '../../modules/util/validators' );
+const ratingsForm = require( '../../apps/owning-a-home/ratings-form' );
+const BASE_CLASS = 'o-email-signup';
+const emailSignup = document.body.querySelector( '.' + BASE_CLASS );
 
 if ( emailSignup ) {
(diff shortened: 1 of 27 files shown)


sebworks:
We should drop JS support for IE9. In the V1 days it would have been dropped once traffic was less than 2% ( before all the do-gooders got involved ).

cfgov/unprocessed/js/routes/owning-a-home/common.js
```diff
@@ -2,20 +2,18 @@
...
 if ( emailSignup ) {
-  var language = document.body.querySelector( '.content' ).lang;
-  var formSubmit = new FormSubmit( emailSignup, BASE_CLASS,
-    { validator: fields => {
-        return validators.email( fields.email, '', { language: language } )
-               .msg;
-      },
-      language: language
+  const language = document.body.querySelector( '.content' ).lang;
+  const formSubmit = new FormSubmit( emailSignup, BASE_CLASS,
```
jimmynotjim: If you end up back in this file, could you clean up this set of params by putting each on it's own line like the eslint4 changes the other day? It's really hard to follow.

anselmbradford: Stuck it in https://github.com/cfpb/cfgov-refresh/pull/3679/commits/8968c7878347bacca9a20a9af60a0487f176f216

Scotchester:
> I discovered `<!--[if IE 9]>` doesn't trigger in Browser Mode: IE9 Compatibility View, but does in Browser Mode: IE9. Also, our logo is blown up in Browser Mode: IE9. Anyone up on which is the default?

Compatibility View is **not** the default, and it should never be triggered by our website, so I'm not worried about the conditional comment not working in Compatibility View.

As for the logo blowing up in non-compat view, I'm not able to replicate that. Any more details on the circumstances under which you saw that, @anselmbradford?

PR #3677, opened 2017-12-14
ES5 shim cleanup

anselmbradford:
The ES5 shim/sham script was checking ES5 support twice in modernizer. Once in the base template and once in the es5-shim module. IE8 shouldn't need the shim because we turn off JS in IE8 (though in theory GTM could be using ES5 features, but if that happens the GTM script should be refactored to not use ES5 features, as they shouldn't require the amount of complexity that needs ES5 features). IE9 supports ES5 features except for strict-mode, which es5-shim does not polyfill anyway, so we don't need to deliver the shim there. IE10+ support es5 features, so shim doesn't need delivery there.

Update: Completely removes es5-shim.

## Changes

- ~~Renames `scriptsEs5Shim` build function `scriptsES5Shim` to match ECMAScript case.~~
- Renames `common.ie.js` to `common.ie9.js` to better convey that that is the only IE this module is delivered to.
- Moves class-list polyfill into `common.ie9.js`, since that's the only place it was used.
- ~~Renames `es5-shim.js` to `es5-shim-sham.js` to better convey what it contains.~~
- ~~Completely turn off delivery of es5-shim to conditional comment supporting IE versions.~~
- ~~Removes redundant es5 feature check inside es5 shim module.~~
- Removes `fnBind` polyfill, since it is only needed for IE8 and BB7. ~~IE8 shouldn't need the polyfill because the JS where it's used is turned off. BB7 should fail the modernizr es5 check and use the bind polyfill inside the es5 shim.~~
- Turns off JS for browsers that don't support es5 features, according to modernizr.

## Removes
- Completely removes es5 shim.

## Testing

1. `gulp build` should pass.
2. No new errors in IE8 (has youtube and GTM errors) and IE9+

## Notes
- Includes https://github.com/cfpb/cfgov-refresh/pull/3675 and #3678 
- You can check browser support for es5 features with finer control with the following logs:

```
    console.log( 'es5array:' + window.Modernizr.es5array );
    console.log( 'es5date:' + window.Modernizr.es5date );
    console.log( 'es5function:' + window.Modernizr.es5function );
    console.log( 'es5object:' + window.Modernizr.es5object );
    console.log( 'strictmode:' + window.Modernizr.strictmode );
    console.log( 'es5string:' + window.Modernizr.es5string );
    console.log( 'es5syntax:' + window.Modernizr.es5syntax );
    console.log( 'es5undefined:' + window.Modernizr.es5undefined );
```

## Todos

- ~~At some point if we want to drop BB7 es5 feature support we could remove es5-shim completely. @ascott1 @cfarm ?~~ Dropping it!


sebworks:
We don''t support Blackberry and we should remove the shim. I think were also using it for Android 4.x as well.  You can't tell me this works on Blackberry because no here can test it.

anselmbradford:
I've been lazy about the iphone update, so still have BB7. We could remove the shim, deploy, and I could report back what happens in prod and then we can decide if we care. That work?

sebworks:
>> I've been lazy about the iphone update, so still have BB7. We could remove the shim, deploy, and I could report back what happens in prod and then we can decide if we care. That work?

No, we should drop support for it because we don't support it. I am seeing 18 BB7 users since July 1, 2017.  The only concern I have is the Android 4.x users.

anselmbradford:
You give a convincing argument! Okay, I've re-worked it to use modernizr to check for es5 features (aside from strict mode—to handle IE9) and if they are absent, disable JS as we do in IE8. I checked in SauceLabs Android 4.4 (lowest I think they have) and ES5 features were supported. Also, contains https://github.com/cfpb/cfgov-refresh/pull/3678

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -165,25 +165,28 @@
     {# Customized Modernizr build that includes html5shiv.
        Built via gulp-modernizer in `scripts.js` task. #}
     <script src="{{ static('js/modernizr.min.js') }}"></script>
+
+    {#
+      Turn off JavaScript for browsers that don't support ECMAScript 5 features
```
jimmynotjim: Should we change the footer script to do the same instead of relying on the IE 8 comment hack? https://github.com/cfpb/cfgov-refresh/pull/3677/files#diff-89b43ee86bbd4952dc44cf751cfd775cR247

anselmbradford: Okay check now

anselmbradford:
@sebworks Per Jimmy's feedback to change the check in the footer to match, I wrapped the footer scripts in an if statement looking for the `no-js` class. This required injecting the script elements instead of `<script>`. For the component JS I realized we don't need to inline and use django compression on it any more with http/2. Is that correct?

sebworks:
@anselmbradford, why write them out, why not just use script tags?

jimmynotjim:
> the problem with this approach is that it's a lot slower for folks who are using JS because you have to wait for it to create and attach the scripts.

How do you feel about moving it to the header and making the load async to speed it up for JS users without blocking page load but also avoid loading any scripts for no JS users?

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -128,11 +128,11 @@
         .optimize-loading { opacity: 0 !important; }
     </style>
     <script>
-        (function(a,s,y,n,c,h,i,d,e){s.className+=' '+y;h.start=1*new Date;
-        h.end=i=function(){s.className=s.className.replace(RegExp(' ?'+y),'')};
-        (a[n]=a[n]||[]).hide=h;setTimeout(function(){i();h.end=null},c);h.timeout=c;
-        })(window,document.documentElement,'optimize-loading','dataLayer',5000,
-        {'GTM-KHB8MB':true});
+      (function(a,s,y,n,c,h,i,d,e){s.className+=' '+y;h.start=1*new Date;
```
jimmynotjim: Is this indentation change intentional? HTML files should still be four space, even if writing JS

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -241,48 +247,58 @@
...
-{% endmacro %}
-{% set js_source = page_js() | trim %}
-{% if js_source | length > 0 %}
-<script src="{{ static('js/routes/' + request.path[1:] + 'index.js') }}"></script>
-{% endif %}
-<!--<![endif]-->
+    {# Check and include component-level JavaScript. #}
+    {% if page and page.media %}
+       {% for js in page.media %}
+         script = document.createElement( 'script' );
```
sebworks: would it make sense to create a small inline function so you don't have to repeat the create / append/ src code. I think that is what they are doing for the GTM stuff.

cfgov/jinja2/v1/_layouts/base.html
anselmbradford: It was intentional. Once you're inside `<script>` it's JS land, no?

cfgov/jinja2/v1/_layouts/base.html
anselmbradford: We have https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/js/modules/util/js-loader.js. We could copy some of that code over.

sebworks: that works, but minimize the code when you add it base.html ( you can just link to the source ).

cfgov/jinja2/v1/_layouts/base.html
jimmynotjim: I don't think so. If we linted it, the linter would only care that the _file_ was html, not what the code written inside of it was.

anselmbradford: Okay I reverted it, but I think a smart HTML linter wouldn't be touching what's inside a `<script>` tag because it's not HTML. If we used something like https://www.npmjs.com/package/eslint-plugin-html#htmlindent, we could control it however we want, but it would still be JS and JS code style I'd think.

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -241,57 +247,51 @@
...
-{% if js_source | length > 0 %}
-<script src="{{ static('js/routes/' + request.path[1:] + 'index.js') }}"></script>
-{% endif %}
-<!--<![endif]-->
+{#
+  Modernizr adds a no-js class the HTML element if ECMAScript 5 features aren't
+  available. Check for that CSS class here and act accordingly.
+#}
+<script>
+    function loadScript(t,e){var n=document.createElement('script');n.type='text/javascript',n.onload=function(){return e?e():null},n.src=t,document.getElementsByTagName('head')[0].appendChild(n)}
```
anselmbradford: This is the unminified script:
```

function loadScript( url, callback ) {
  const script = document.createElement( 'script' );
  script.type = 'text/javascript';

  script.onload = function() {
    if ( callback ) {
      return callback();
    }
    return null;
  };

  script.src = url;
  document.getElementsByTagName( 'head' )[0].appendChild( script );
}
```

cfgov/jinja2/v1/_layouts/base.html
anselmbradford: Added it here https://github.com/cfpb/cfgov-refresh/pull/3677/files#diff-89b43ee86bbd4952dc44cf751cfd775cR256

cfgov/jinja2/v1/_layouts/base.html
anselmbradford: Actually removed the callback onload bit, as none of the scripts use a callback. https://github.com/cfpb/cfgov-refresh/pull/3677/commits/25cf9ad6f58181863147a04378a7982f542d53a2

jimmynotjim:
Builds fine and looks good in my IE8 VM.

anselmbradford:
Thanks for fixing Travis @sebworks !

PR #3678, opened 2017-12-15
No mobile view on IE9

anselmbradford:
There is a bug in IE9 (non-compatibility mode, which I believe is the default?) where the resize event will fire unexpectedly, causing the mobile menu to get stuck showing on load at mobile sizes. I believe this is because IE9 fires the resize event on resize of any elements on the page, not just the whole window. Instead of hacking around that, this PR turns off the mobile responsive view for IE9, like we do for IE8.

## Changes

- Disable mobile responsive view for IE9-.

## Testing

1. `gulp build` and check IE9 at mobile window sizes and observe it's still the desktop menu.

## Screenshots

Before:

screen shot 2017-12-14 at 7 25 27 pm

After:

screen shot 2017-12-14 at 7 25 39 pm

### DIFF
--- a/cfgov/unprocessed/js/organisms/MegaMenu.js
+++ b/cfgov/unprocessed/js/organisms/MegaMenu.js
@@ -166,6 +166,15 @@ function MegaMenu( element ) {
    * suspends or resumes the mobile or desktop menu behaviors.
    */
   function _resizeHandler() {
+
+    /* If in Internet Explorer 9-, don't support mobile responsive view.
+       lt-ie10 class is added by modernizr. */
+    const htmlEl = document.body.parentElement;
+    let onlyDesktop = false;
+    if ( htmlEl.classList.contains( 'lt-ie10' ) ) {
+      onlyDesktop = true;
+    }
+
     if ( breakpointState.isInDesktop() ) {
       _mobileNav.suspend();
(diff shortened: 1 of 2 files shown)


sebworks:
@ascott1, how do you feel about now supporting IE9 with JS? It's usage is now less than IE8, which we don't support with JS.

ascott1:
@sebworks I would be totally open to a discussion around that. I think it warrants its own issue though.

cfgov/unprocessed/js/organisms/MegaMenu.js
```diff
@@ -166,6 +166,15 @@ function MegaMenu( element ) {
...
+
+    /* If in Internet Explorer 9-, don't support mobile responsive view.
+       lt-ie10 class is added by modernizr. */
+    const htmlEl = document.body.parentElement;
+    let onlyDesktop = false;
+    if ( htmlEl.classList.contains( 'lt-ie10' ) ) {
+      onlyDesktop = true;
+    }
+
     if ( breakpointState.isInDesktop() ) {
```
anselmbradford: Oops, forgot to add `|| onlyDesktop`, but it's taken care of in https://github.com/cfpb/cfgov-refresh/pull/3677

PR #3682, opened 2017-12-15
Mattermost alert failures shouldn't be fatal

chosak:
This change modifies how Mattermost alert failures are handled as part of the SQS alert consumption process.

Currently, when an alert is read off of an incoming SQS queue, the process goes like this:

1. Peek at alert on top of queue.
2. Post alert to GH.
3. Post alert to MM.
4. Delete alert from queue.

If Mattermost happens to be down (as happened recently) and an alert comes it, it'll be repeatedly posted over and over to GH, because a failure at step 3 prevents alert cleanup at step 4.

This change catches exceptions from MM posts and logs them, instead of allowing the alert to exit the process.

In practice that means that if this script is run from e.g. Jenkins the Jenkins job will not fail just because MM is down.

## Changes

- Logic in [`alerts.post_sqs_messages.process_sqs_message`](https://github.com/cfpb/cfgov-refresh/blob/75277246d5533b31911e7a010c544c09210f6b66/cfgov/alerts/post_sqs_messages.py#L97) modified to catch and log Mattermost failures.

## Testing

1. See new unit test.

## Checklist

### DIFF
--- a/cfgov/alerts/post_sqs_messages.py
+++ b/cfgov/alerts/post_sqs_messages.py
@@ -87,13 +87,20 @@ def process_sqs_message(message, github_alert, mattermost_alert):
         title=title,
         body=body
     )
+
     if mattermost_alert:
-        mattermost_alert.post(
-            text='Alert: {}. Github issue at {}'.format(
+        try:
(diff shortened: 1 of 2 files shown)


hillaryj:
I definitely appreciate this fix! :)

PR #3685, opened 2017-12-15
Remove logging of page filtering

chosak:
This change modifies the logging in two places where pages get filtered. ~~Right now these filters get logged at the `INFO` level, which means they go to the console and log files by default.~~

~~This fills up the logs with filtering every time a filter is chosen, which seems overly verbose. If we want these to be logged, could we instead log them to `DEBUG`?~~ Per feedback from @richaagarwal below, I'm just removing these logging statements.

## Changes

- ~~Reduce filter logging from INFO to DEBUG.~~ Remove filter logging.

## Checklist

### DIFF
--- a/cfgov/v1/models/sublanding_page.py
+++ b/cfgov/v1/models/sublanding_page.py
@@ -95,7 +95,7 @@ class SublandingPage(CFGOVPage):
                 CFGOVPage.objects.child_of_q(page)
             )
 
-            logger.info('Filtering by parent {}'.format(page))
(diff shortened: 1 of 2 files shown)


richaagarwal:
I thought these had been removed by you (or someone else) in a previous PR -- but maybe that was something else similar?  I'd be fine with removing them -- they were helpful at one point when this functionality was breaking a lot, but probably no longer useful.

chosak:
@richaagarwal that sounds familiar. Thanks for the review, I've removed them in 13dbed3.

PR #3686, opened 2017-12-15
Downgrade imagemin-svgo

anselmbradford:
The update to imagemin-svgo was optimizing the logo in an undesirable way by making it scale twice as large in IE9. @Scotchester is looking at revamping the approach to the logo overall in https://github.com/cfpb/cfgov-refresh/tree/remove_svg_logo

## Changes

- Downgrade imagemin-svgo from 6.x to 5.x

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,13 @@
         "normalize-path": "2.1.1",
         "source-map": "0.5.7",
         "through2": "2.0.3"
+      },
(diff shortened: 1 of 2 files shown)


anselmbradford:
Since this brings it back to the state of that dependency a day ago, I'm comfortable merging with just @Scotchester approval. Too bad everyone else, you snooze, no review for you!

PR #3699, opened 2017-12-20
Defer all scripts to after common.js load

anselmbradford:
It was possible for scripts to load before common.js, leading to an error that webpack's loader was not found.

## Changes

- Makes template, component, and page-level scripts load within a callback after common.js load.

## Testing

1. `gulp build`
2. check blog and see that page has three scripts injected into the head: common, email-signup, and filterable-list-control.


anselmbradford:
Experimented with Steven on setting `n.async=false,` in the dynamic script loader, which appeared to work, but given holes in browser support and confusion about how it was actually working we decided to use the callback approach in this PR for now. The idea is that common.js needs to load before its dependent scripts. With async=false set for all scripts (common, template, component, and page-level), common.js appeared to load first, but performance panel showed what looked like them loading in parallel.

PR #3700, opened 2017-12-20
Use docker for front-end dev

rosskarchner:
This is a first crack at providing some docker/compose infrastructure for front-end development. Some cons and pro's:

### Con:
- It's definitely slower than running node natively on the mac
- there's some indirection involved, direct interaction with node,npm, gulp, etc happens via ./gulp-shell.sh (like ./shell.sh for Python stuff)
### Pro:
- It removes the necessity of having node, etc installed on the mac
- It gives us a way to control what version of Node people are using, without NVM or homebrew tricks
- 'gulp watch' will always be running at the same time as the django server, etc-- so maybe less direct interaction with node/gulp will be necessary?
- It includes a fix for the issue where scripts that source mac-virtualbox-init.sh stop unexpectedly (if this drags on for a while, maybe we can extract that into it's own PR)

There's an interesting (maybe controversial) assumption I baked in: once you've run `setup.sh docker`, both frontend.sh and runserver.sh will use docker (this is controlled by the existence of a file called .USE_DOCKER. To stop using docker, setup.sh can be run again without the `docker` argument, OR the user can just delete the .USE_DOCKER file.

I did *not* yet add an equivalent to .python_env for passing environment variables into the frontend code-- that's probably a necessary thing to add.

I haven't decided whether [this trick](https://github.com/cfpb/cfgov-refresh/blob/docker-gulp/frontend.sh#L17-L20) is good funky or bad funky.

## Testing

To get the full effect, I would start from scratch:

```
rm -rf node_modules
git clean -fX cfgov/static_built/*
eval $(docker-machine env -u)
docker-machine rm default
```
If you're feeling *really* brave, uninstall node on your mac 😄 . 

and then, [presuming you have all the docker bits installed](https://cfpb.github.io/cfgov-refresh/installation/#1-setup-your-docker-environment)

`./setup.sh docker && ./runserver.sh`

You'll want to do [your favorite database loading routine](https://cfpb.github.io/cfgov-refresh/installation/#3-setup-the-database)

Observe that:
- front-end assets have been built
- `gulp-watch`/[browsersync](https://www.browsersync.io/) is accessible on http://localhost:3000
- you can run `./frontend.sh` as before, but it will use docker instead of the mac-native version of node
- before running any other gulp or npm commands, `./gulp-shell.sh`

## Notes

-

## Todos

-

## Checklist


anselmbradford:
> OR the user can just delete the .USE_DOCKER file.

Can this go into and be set in `.env`? Is there some technical reason this needs to be its own file?

rosskarchner:
@anselmbradford I've been trying to avoid the very need of people to manage their shell environment. What you're proposing would mean that setup.sh would have to write something to .env (or require the user to edit it themselves) and re-`source` it (if they aren't using autoenv or something)

OR, we toss the idea of "remembering" that the user want wants to use docker, and expect them to call `frontend.sh docker` (like we currently expect people do do with runserver.sh)

anselmbradford:
Took about 45 minutes to run. The server is running but I get this:
screen shot 2017-12-21 at 12 05 42 pm

anselmbradford:
Says I have unapplied migrations, which I suspect is the issue, but how do I apply those in the docker setup? Stopping the server, running `python cfgov/manage.py migrate`, and re-starting the server doesn't work.

rosskarchner:
@anselmbradford this will probably be helpful: https://cfpb.github.io/cfgov-refresh/installation/#3-setup-the-database

If you aren't using entering ./shell.sh before the manage.py commands, you're probably still interacting with your local MySQL install, not the one inside Docker.

rosskarchner:
the docs should make it clear: run shell.sh in a seperate tab, while the server is running

anselmbradford:
For the record, I was able to update the migrations with @rosskarchner 's comments. Someone else should probably try the setup too.

rosskarchner:
@anselmbradford do you think we need to use a port other than 3000 for gulp watch?

anselmbradford:
> @anselmbradford do you think we need to use a port other than 3000 for gulp watch?

Might not be a bad idea to at least allow setting it in `.env`.

anselmbradford:
@rosskarchner `gulp` watch has been removed, which may simplify some things in here.

hillaryj:
What's the status on this PR? Is it still useful?

rosskarchner:
I think if the gulp watch stuff is removed, then (most of) the pro's and con's listed above would still stand.  Maybe see if the team is interested in *having* this option?

PR #3702, opened 2017-12-20
Stop serving logo as SVG and use srcset and sizes to serve multiple sizes of PNG

Scotchester:
We recently encountered a problem where an upgrade to `imagemin-svgo` changed how our logo SVG was being optimized, making it appear way too large and cropped in all IEs. Aside from potential future regressions of this nature, the value of serving an SVG in the first place was somewhat questionable, because we don't scale the image fluidly (we have two set sizes at which it appears) and the largest PNG we need to serve is only ⅔ the filesize of the SVG. The code needed to fall back to PNGs in browsers that don't support SVG was also somewhat convoluted.

So, this PR converts us to all-PNG logos, all the time. It serves different sizes using the `srcset` and `sizes` attributes, falling back to the trusty `src` attribute in browsers that don't support those.

## Additions

- New PNG logo files at all sizes necessary for our two sizes at various pixel densities
  1. 50px tall for larger screens
     1. `@2x` for high-dpi displays
  2. 34px tall for smaller screens
     1. `@2x`
     2. `@3x` (plus-sized iPhones and iPhone X, Galaxy S4 and S5, 
     3. `@4x` (Galaxy S6 and up, Galaxy Notes 4 and up, LG Gs 3 and up)
- `srcset` and `sizes` code for choosing the correct image size

## Removals

- Code related to serving SVG versions of the logo
  - _Note:_ `logo.svg` is not removed because PDFreactor still uses it in its template

## Testing

1. Pull branch
1. `gulp build`
1. Load up any page
1. Open responsive design mode in dev tools
1. Try a bunch of different device emulations, reloading after choosing each one and ensuring via the Network panel that the correct one is used

## Checklist

* [x] Any *change* in functionality is tested


anselmbradford:
I got in the habit of including the dimensions in the image name in other images, such as https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/img/plug-269x327.png, so that it is (a) more readily apparent at a glance the size of an image and (b) like hashed JS assets a change in image size requires a change in filename.

anselmbradford:
screen shot 2017-12-21 at 11 03 20 am

Scotchester:
> I got in the habit of including the dimensions in the image name in other images, such as https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/img/plug-269x327.png, so that it is (a) more readily apparent at a glance the size of an image and (b) like hashed JS assets a change in image size requires a change in filename.
>
> What do you think about that?

Yeah, I sometimes do that. I think I was thinking "our logo sizes are based on a particular height, that's the number people will be looking for". But it probably wouldn't hurt to add the horizontal dimension, as well. I think I want to stick with the `@#x` convention, though, as it seems to be a widely-accepted practice.

anselmbradford:
>  I think I want to stick with the @#x convention, though, as it seems to be a widely-accepted practice.

Yeah you can do both, just include the dimensions in the middle of the filename.

anselmbradford:
My issue was I had an older version of this branch.

Scotchester:
Images renamed in 5e52277.

anselmbradford:
I actually meant the actual pixel dimensions, so instead of `logo_237x50@2x.png`, I meant`logo_ 474x100@2x.png`, but not a biggie. Also, seems the majority of images are hyphen separated, not underscore, perhaps [due to this](https://css-tricks.com/forums/topic/seo-image-file-name-do-underscores-_-matter/) (which is old so maybe no longer applicable):

> Underscores are seen by search engines generally as “word combiners”, and hyphens are seen as “word separators.” Scott was right, never use underscores for anything.

Scotchester:
Hmm... I think it would be confusing if the multipliers weren't referencing the same base size.

As for the punctuation... I'm not sure why it would matter in an image filename. I'm not sure in what situation it would affect search results. And I think the Google-provided guidance that person is referencing is specific to URLs to pages.

anselmbradford:
> Hmm... I think it would be confusing if the multipliers weren't referencing the same base size.

The multipliers refer to the pixel density, not the actual pixel count, are you thinking that's better because that's the target pixel size you're going for in the markup/CSS?

Scotchester:
Well, they refer to both, because the actual pixel count is in fact that multiplier of the base size.

PR #3703, opened 2017-12-21
Inline modernizr + externalize common.js in error templates

anselmbradford:
## Removes
- Un-utilized `classlist` modernizr test (handled by IE9 class-list polyfill).

## Changes

- Inlines modernizr to remove modernizr request.
- Adds code comments for modernizr settings.
- Removes inline common.js request from error pages and creates a request.
- Wrap some long lines in error templates.

## Testing

1. `gulp build` should pass.
2. Modernizr should appear inline in the document head.
3. Check /404/ and /500/ error pages and see that there's a request for common.js


cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -162,9 +162,12 @@
...
     {# end Google Tag Manager #}
 
-    {# Customized Modernizr build that includes html5shiv.
-       Built via gulp-modernizer in `scripts.js` task. #}
-    <script src="{{ static('js/modernizr.min.js') }}"></script>
+    {# Customized Modernizr build that includes html5printshiv.
+       Built via gulp-modernizer in `scripts.js` task.
+       TODO: Determine if modernizr is actually needed! #}
+    <script>
+        {% include '/js/modernizr.min.js' %}
```
Scotchester: How does it know where to find this file?

Scotchester:
Correction: Backend coverage is what's failing. And I have no idea why that would be the case. Gonna update from `master` and see if if fails on the next run.

Scotchester:
OK, now the back-end tests are failing with a ton of `TemplateNotFound: /js/modernizr.min.js` errors. So it appears my question was significant.

cfgov/jinja2/v1/_layouts/base.html
anselmbradford: Hmm I'm not sure why this would be problematic. It's how we always included common.js https://github.com/cfpb/cfgov-refresh/pull/3539/files#diff-89b43ee86bbd4952dc44cf751cfd775cL247

@rosskarchner Do you know?

sebworks: aren't you missing `static`?

anselmbradford: You don't need static unless you're making a request I thought. See line 248 and 261, we never used static for inlined JS before https://github.com/cfpb/cfgov-refresh/blob/2e125c5119d45f1ed0b72cf86a0de27cd119ac80/cfgov/jinja2/v1/_layouts/base.html#L248-L261

anselmbradford: Figured it out https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/test.py#L43-L53

anselmbradford:
Includes https://github.com/cfpb/cfgov-refresh/pull/3711

PR #3708, opened 2017-12-21
Proof of concept: minify HTML

ascott1:
This addresses GHE/CFGOV/platform/issues/1824 by using the `django-htmlmin` middleware to minify HTML. This shouldn't be merged without, however, without further testing and discussion. In particular, I notice our own @rosskarchner has [an open bug report](https://github.com/cobrateam/django-htmlmin/issues/108) around aggressive minification.

## Additions

- adds the `django-htmlmin` middleware

## To test

To test this locally or in debug mode add the following setting:

```
HTML_MINIFY = True
```


cfarm:
Added myself as a reviewer so I can remember to test the performance impact of this change by deploying to a dev server and measuring the HTML size! I'll post results here (likely will get to this next week).

rosskarchner:
My interpretation of the conversation on that bug report: we need to define EXCLUDE_TAGS_FROM_MINIFYING in django settings, as a list of all of the inline tags we might use.

willbarton:
@rosskarchner good point. That's something I think we could address with the CMS rich text editor, but do we have any way of knowing what inline elements we'd need to whitelist for non-CMS-RTE pages? Would we want to whitelist all possible inline elements? 

WRT the CMS, I don't *think* we expand or override [this list of elements from the Wagtail whitelister, do we?](https://github.com/wagtail/wagtail/blob/stable/1.10.x/wagtail/wagtailcore/whitelist.py#L71-L92)

rosskarchner:
> Would we want to whitelist all possible inline elements?

I think that's where we're headed-- it's either that, or apply some sort of normalization before the minimization routine.

Scotchester:
Let's wait and see what the performance numbers are before we go too far down this road. I have my doubts that it will affect a whole lot, given gzip compression.

ascott1:
> Let's wait and see what the performance numbers are before we go too far down this road. I have my doubts that it will affect a whole lot, given gzip compression.

Using gzip locally this puts the home page at 12.3kb, which is below the [14kb for 1st requests recommendation](https://www.scottkellum.com/2015/07/07/building-tiny-websites). That said, I don't think there is a huge crunch for putting the effort into this and I would be in favor of getting that number down through other means if possible.

cfarm:
This branch is deployed on the dev4 server, I'll be recording some metrics later today (like 5pm) but feel free to check it out and post any you find here if anyone else has time before I get to it.

cfarm:
I can't tell gzipped size for the HTML using Chrome's Dev Tools (anyone else know how to see this?), but it looks like this would give us a significant decrease in the download size, **and** for the very first response (part of "the critical rendering path") which is a huge win for improving perceived performance with time to First meaningful paint, and reducing overall load times.

> The first step in the critical path is the initial round-trip from the server to the browser, and that new TCP/IP connection can include up to 10 TCP packets—about 14KB of data. That’s more than enough to get our markup into the browser so it can determine what other requests need to be made.
https://bocoup.com/blog/performance-under-pressure

rosskarchner:
Using MDN's [list of inline tags](https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements), I've added them all to the EXCLUDE_TAGS_FROM_MINIFYING setting, which at least fixes the issue @willbarton noticed on the mission statement:

> The Consumer Financial Protection Bureau is a 21st century agency that helps consumer finance markets work by making rules more effective, by consistently and fairly enforcing those rules, and by empowering consumers to take more control over their economic lives. For more information, visit consumerfinance.gov.

cfarm:
Should we get a closer QA check of this? I think the performance gains are significant enough that we should try to ship this!

@ascott1 anyone on Platform able to help review this on dev server 04 to make sure there are no major issues with the minification like the over-aggressiveness we saw before?

cfarm:
Note: this is no longer on dev 04, if someone wants to QA this ping me and I'll deploy it to a dev server again.

ascott1:
I'm going to close this since we don't have plans to merge/implement anytime soon, but in the long-term I think that this is an idea worth further exploration.

PR #3711, opened 2017-12-22
Remove inline JS from error templates

anselmbradford:
## Changes

- Removes inline common.js request from error pages and creates a request.
- Wrap some long lines.

## Testing

1. Visit /404/ and /500/ and not there are no errors.


anselmbradford:
Merged into https://github.com/cfpb/cfgov-refresh/pull/3703

PR #3707, opened 2017-12-21
Serve latest draft page for specific pages via a wagtail hook on beta

willbarton:
This PR enables us to serve specific page primary keys specified in settings on beta.consumerfinance.gov. This is intended to be as non-invasive as possible as we come up with a more long-term solution for previewing content.

## Additions

- `SERVE_LATEST_DRAFT_PAGES` added to settings and page pk 1288 configured on beta
- `serve_latest_draft_page` wagtail hook added that consumes `SERVE_LATEST_DRAFT_PAGES`

## Notes

This code is all @chosak's. I'm just an ape swinging from git branch to git branch.

## Checklist

* [ ] PR has an informative and human-readable title
* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [development playbook](https://github.com/cfpb/development)
* [ ] Passes all existing automated tests
* [ ] Any *change* in functionality is tested
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated
* [ ] Reviewers requested with the [Reviewer tool](https://help.github.com/articles/about-pull-request-reviews/) :arrow_right:

PR #3712, opened 2017-12-22
Even simpler approach to flagged drafts

chosak:
This change is a follow up / alternative to #3707 to support exposure of draft content given certain conditions. Instead of relying on a hardcoded set of page primary keys, it utilizes built-in [wagtail-flags](https://github.com/cfpb/wagtail-flags) capability to control draft exposure via feature flag.

So, for example, Wagtail users could create a flag for a particular site with the condition:

`"path matches": "/path/to/page/"`

and that page will be served with its draft content. A more complicated regular expression would support multiple pages like this, e.g.

`"patch matches": "/page1/|/page2/"`

This lets draft serving be controlled through the Wagtail UI, instead of relying on code changes.

## Removals

- Remove hardcoded set of draft page PKs.

## Changes

- Change draft serving logic to rely on built-in feature flag capability.

## Testing

1. Create a feature flag in the UI for flag `SERVE_AS_DRAFT` with the condition `"path matches": "/path/to/your/page/"`. View that page at its published URL and you should see the draft content there.

## Todos

- Create feature flags in the Wagtail UI for any pages that need to be served like this.

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -639,6 +639,9 @@ FLAGS = {
 
     # Teacher's Digital Platform
     'TDP_RELEASE': {},
+
+    # Serve some pages with their draft content.
+    'SERVE_AS_DRAFT': {},
 }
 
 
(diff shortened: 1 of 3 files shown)


willbarton:
As we discussed, I don't think this will work right now because:

- Feature flags are not persisted on beta.consumerfinance.gov
- [*Any* Wagtail-Flags condition can be `True` for a flag to be enabled](https://github.com/cfpb/wagtail-flags/blob/master/flags/settings.py#L59), so they are effectively a boolean *or*. It would be nice to have a way to *and* certain conditions, but that's not currently implemented (I envision a condition that joins other conditions).

So the only way this would work is to define the conditions in settings like this:

```python
 if DEPLOY_ENVIRONMENT == 'beta': 
    FLAGS['SERVE_AS_DRAFT'] = {
        'path matches': '^/page1/',
        'path matches': '^/page2/'
    }
```

(We also wouldn't need the `|` in the path match either.)

That would mean we couldn't actually rely on the Wagtail UI though.

chosak:
Closing this PR per @willbarton's feedback above, in favor of the existing #3707 which has already been merged. Hopefully we can revisit this approach as part of a larger discussion on selectively sharing draft content.

PR #3693, opened 2017-12-19
Give IE9 the no-js experience

anselmbradford:
Fixes https://github.com/cfpb/cfgov-refresh/issues/3687

## Removals

- `common.ie.js` IE9-specific JS.
- IE scripts build task.
- `class-list.js` polyfill.

## Changes

- Merges IE8 and IE9 style build tasks.
- Increment conditional comments by one to ignore JS in IE9.
- Drop `script.readyState` check in dynamic JS loader (for IE9).

## Testing

1. `gulp clean && gulp build`.
2. Check site in IE9 and IE8.

Issue #3687, opened 2017-12-15
Give IE9 no-js treatment


anselmbradford:
https://github.com/cfpb/cfgov-refresh/pull/3678 aimed to disable the busted mobile menu on IE9.
@sebworks pointed out that very few users actually visit the site in IE9 at mobile window sizes, so the extra bit of JS to do the disabling is probably not worth it. We also have two build configurations for IE8 and IE9 to handle a no-js and js state for each. If we merged those configurations we could eliminate a build step, the conditional comments and script for loading a class-list polyfill for IE9, and the JS added in #3678. Based on our analytics @sebworks says it's time and I agree.

@ascott1 @cfpb/cfgov-frontends Can we do it?

Scotchester:
Probably should get a @schaferjh opinion on this, too.

cfarm:
Is there a way to just serve the no-js menu to IE9 without disabling JS for all IE9 users? Can we scope a class name for the mega menu without js, and turn that on for IE9 and below?

cfarm:
Also, @sebworks and @anselmbradford, as the proposers of this change, can you document for us what key pages will be affected that currently support IE9 JS, and the number of IE9 users on those pages in the last year? (Check out key pages list in the Platform wiki on GHE: https://GHE/CFGOV/platform/wiki/Key-pages-for-QA%2C-testing%2C-analytics). 

Perhaps it would also help to have a snapshot of how those numbers have changed over the last couple of years or so.

sebworks:
ok, let me write some stuff up this weekend on how I think the change will affect us.

sebworks:
Currently, less than 1% of all users are visiting the site using IE9. There are currently more visitors using IE8, which has already been transitioned to the no js experience.  The chart below shows the amount of IE users for the last 6 months. 

<img width="404" alt="screen shot 2017-12-18 at 11 08 00 am" src="https://user-images.githubusercontent.com/1696212/34115471-bddcdd54-e3e3-11e7-9c75-bce8b33f9efa.png">

The chart below depicts IE9 users for a two year period starting December 17, 2015.  From 20,000 monthly users to 630.

<img width="1082" alt="screen shot 2017-12-18 at 11 12 44 am" src="https://user-images.githubusercontent.com/1696212/34115719-67d33b64-e3e4-11e7-9a85-499858a65dd3.png">


[...]

PR #3713, opened 2017-12-22
Refactor for script loading in base.html

sebworks:
Refactor for script loading in base.html

## Changes

- Modified `cfgov/jinja2/v1/_layouts/base.html` to use script detailed in https://www.html5rocks.com/en/tutorials/speed/script-loading/. Uses `async=true` for script tags to ensure proper loading order.  This should allow us to take advantage of loading scripts in parallel and executing sequentially.

## Testing

1. Visit http://localhost:8000/about-us/blog and ensure the page works.
2. Visit http://localhost:8000/data-research/consumer-credit-trends/student-loans/borrower-risk-profiles/ and ensure the page works.

## Notes

- We need a write up on manually testing pages. 
- We should add the ability to use the `rel="preload` attribute.
## Todos

-

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_layouts/base.html
+++ b/cfgov/jinja2/v1/_layouts/base.html
@@ -252,10 +252,25 @@ height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
 
 {% block javascript scoped %}
 
+
+{#
+  Modernizr adds a no-js class the HTML element if ECMAScript 5 features aren't
+  available. Check for that CSS class here and act accordingly.
+#}
+
 <script>
-    {# Minified dynamic JavaScript loader that injects a script tag in the head
-       of the page. It also provides a callback parameter. #}
-    function loadScript(t,e){var n=document.createElement('script');n.type='text/javascript',n.onload=function(){return e?e():null},n.src=t,document.getElementsByTagName('head')[0].appendChild(n)}
+  if ( document.body.parentElement.className.indexOf( 'no-js' ) === -1 ) {
+    {# Include site-wide JavaScript. #}
+    var a = [ '{{ static('js/routes/common.js') }}' ];
+    {# Check and include template-level JavaScript. #}
+    {% macro template_js() %}
+       {% include 'js/routes/' + request.path.split('/')[1] + '/single.js' ignore missing %}
+    {% endmacro %}
+    {% set js_source = template_js() | trim %}
+    {% if js_source | length > 0 %}
+      a.push( '{{ static('js/routes/' + request.path.split('/')[1] + '/single.js') }}' );
+    {% endif %}
+
 
     {#
       Check and include component-level JavaScript.
@@ -263,35 +278,11 @@ height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
       unprocessed/js/routes/on-demand/[component].js
       to the page, based on what components were added in Wagtail.
     #}
-    function loadScriptComponent(){
-      {% if page and page.media %}
-         {% for js in page.media %}
-           loadScript('{{ static('js/routes/on-demand/' + js) }}');
-         {% endfor %}
-      {% endif %}
-    }
-
-    {#
-      TODO: Remove the below blocks when all pages
-            using this template are moved to Wagtail.
-            This moves us away from a URL-dependant script inclusion scheme.
-    #}
-
-    {#
-      Check and include template-level JavaScript.
-      Template-level JavaScript applies to all pages under the URL scheme
-      www.consumerfinance.gov/[app-url]/
-      and JavaScript appears in unprocessed/js/routes/[app-url]/single.js.
-    #}
-    function loadScriptTemplate(){
-      {% macro template_js() %}
-         {% include 'js/routes/' + request.path.split('/')[1] + '/single.js' ignore missing %}
-      {% endmacro %}
-      {% set js_source = template_js() | trim %}
-      {% if js_source | length > 0 %}
-        loadScript('{{ static('js/routes/' + request.path.split('/')[1] + '/single.js') }}');
-      {% endif %}
-    }
+    {% if page and page.media %}
+       {% for js in page.media %}
+         a.push( '{{ static('js/routes/on-demand/' + js) }}' );
+       {% endfor %}
+    {% endif %}
 
     {#
       Check and include page-level JavaScript.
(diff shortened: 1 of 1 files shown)


cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -255,56 +255,46 @@
...
+      a.push( '{{ static('js/routes/' + request.path[1:] + 'index.js') }}' );
+    {% endif %}
 
-<script>
-//<![CDATA[
-    var usasearch_config = { siteHandle: 'cfpb' };
-    loadScript('https://search.usa.gov/javascripts/remote.loader.js');
-//]]>
+    {# Minified dynamic JavaScript loader that injects a script tag in the head
+      of the page. It also provides a callback parameter. #}
```
anselmbradford: We don't need the callback anymore in this PR, right?

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -255,56 +255,46 @@
...
+    var b,c,d=[],f=document.scripts[0];function g(){for(var e;d[0]&&'loaded'==d[0].readyState;)e=d.shift(),e.onreadystatechange=null,f.parentNode.insertBefore(e,f)}for(;b=a.shift();)'async'in f?(c=document.createElement('script'),c.async=!1,c.src=b,document.head.appendChild(c)):f.readyState&&(c=document.createElement('script'),d.push(c),c.onreadystatechange=g,c.src=b);
```
anselmbradford: `readyState` can be stripped out.

anselmbradford:
This looks good! We'll want https://github.com/cfpb/cfgov-refresh/pull/3693 to go in first or this will break things in IE9 (no async support, so get the webpackJsonp error).

cfgov/jinja2/v1/_layouts/base.html
anselmbradford: …because it's for IE9 and below, which won't handle the `async` in this PR anyway.

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -255,56 +255,46 @@
...
+  }
+  var usasearch_config = { siteHandle: 'cfpb' };
 </script>
 
+<script async="true" src="https://search.usa.gov/javascripts/remote.loader.js"></script>
```
ascott1: Could this be simplified to:

```html
<script async src="https://search.usa.gov/javascripts/remote.loader.js">
```

anselmbradford: Good catch, it being a boolean it should have the value of `async` or `async="async"`

Scotchester:
This seems to be working fine for me locally, but I don't feel qualified to provide a real review. I read the whole article and looked at the code changes fairly carefully, but I still don't fully understand what's going on here. Why did you not use the article's recommended minified code exactly as given? I'm not seeing any scripts getting loaded with the `async` tag. In what situations would that occur? Also, isn't the article's recommendation predicated on using the `link rel="subresource"` elements?

cfgov/jinja2/v1/_layouts/base.html
Scotchester: If I understand correctly (and it's very very possible I am not understanding correctly), the whole `readyState` bit is to compensate for the fact that IE9 and under don't handle `async`. Check out the comments inside the `while` loop [in the article's final recommendation](https://www.html5rocks.com/en/tutorials/speed/script-loading/#toc-aggressive-optimisation).

cfgov/jinja2/v1/_layouts/base.html
Scotchester: I think that's right. The callback is gone, since we're no longer using `loadScript`. I think this means we can also get rid of `cfgov/unprocessed/js/modules/util/js-loader.js`? Looks like `VideoPlayer.js` may still be using it, though.

cfgov/jinja2/v1/_layouts/base.html
Scotchester: To add: This PR works fine for me as-is in IE9.

anselmbradford: @Scotchester Try reloading a few times.

sebworks: @Scotchester is correct, the readyState code is for ie9. I'll check the ie9 errors.

Scotchester: Ah yeah, after a few reloads, I got a `'webpackJsonp' is undefined` error.

sebworks: The code wasn't actually being executed in IE9; `'async' in firstScript` was returning true for IE9.  Even once that was fixed, it still wasn't guaranteeing the execution order in IE9 using readyState. 

I've removed the defer and readyState code and conditionally set the async flag, if it's supported.

sebworks:
>> Also, isn't the article's recommendation predicated on using the link rel="subresource" elements?

I've modified the code to preoload common.js in https://github.com/cfpb/cfgov-refresh/pull/3713/commits/28dbed02a2691dcb61e524e40d011d2bf0efe7d5. This provides hints to browser engines about which assets are going to be needed very soon after loading.  There is more info here: https://medium.com/reloading/preload-prefetch-and-priorities-in-chrome-776165961bbf

cfgov/jinja2/v1/_layouts/base.html
sebworks: Fixed in https://github.com/cfpb/cfgov-refresh/pull/3713/commits/38259316db4aed1a8dca4c8c4f94865c74542fd2.

sebworks:
@anselmbradford, @Scotchester please take a look when you get a chance.

Scotchester:
Using `rel="preload"` for `common.js` definitely makes a ton of sense, and after reading that Medium post, I think we should probably also be using it for our webfonts.

I'm still having a bit of a hard time wrapping my head around what's going on here, but let me ask this: Does the recent decision made in #3687 impact what we want to do here at all?

sebworks:
@Scotchester, currently, in order to ensure the proper execution order, we load common.js and then load the the other script files. The downside is all the script files don't load in parallel: 

screen shot 2018-01-24 at 6 37 23 am

This change modifies the code so that the scripts load in paralell but still execute sequentially:

screen shot 2018-01-24 at 6 42 04 am

sebworks:
>> Does the recent decision made in #3687 impact what we want to do here at all?

No, I fixed the IE9 issue by always setting `async=false`.

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -113,6 +112,13 @@
 <!--[if gt IE 9]><!--><link rel="stylesheet" href="{{ static('css/main.css') }}"><!--<![endif]-->
 {% endblock css %}
 
+{#
+    ======
+    PRELOADING DYNAMIC ASSETS
+    ======
+#}
+<link rel="preload" src="{{ static('js/routes/common.js') }}" as="script">
```
Scotchester: `src` should be `href` here

sebworks: fixed in https://github.com/cfpb/cfgov-refresh/pull/3713/commits/0dfe88e20bbdbd2b0b6a344db49614419d1d936b

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -252,83 +257,66 @@
...
 
-<script>
-//<![CDATA[
-    var usasearch_config = { siteHandle: 'cfpb' };
-    loadScript('https://search.usa.gov/assets/sayt_loader.js');
-//]]>
+  var usasearch_config = { siteHandle: 'cfpb' };
 </script>
 
+<script async src="https://search.usa.gov/javascripts/remote.loader.js"></script>
```
anselmbradford: Do we need the change shown here https://github.com/cfpb/cfgov-refresh/pull/3791 ?

sebworks: Fixed.

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -252,83 +257,66 @@
...
-      TODO: Remove the below blocks when all pages
-            using this template are moved to Wagtail.
-            This moves us away from a URL-dependant script inclusion scheme.
-    #}
+{#
+  Modernizr adds a no-js class the HTML element if ECMAScript 5 features aren't
+  available. Check for that CSS class here and act accordingly.
+#}
+<script async>
+  if ( document.body.parentElement.className.indexOf( 'no-js' ) === -1 ) {
```
anselmbradford: Would `document.documentElement` or `document.children[0]` work here for the lookup? Would need to check IE9 compatibility.

anselmbradford: Or I guess IE10+

PR #3725, opened 2018-01-05
update front-end dependencies

ascott1:
This updates the minor and patch releases of all explicit front-end dependencies listed in `package.json` as well as declares `marked` as an explicit dependency.


Scotchester:
This seems to only update axe-core, a subdependency of our protractor-accessibility-plugin dev dependency, from 2.5.0 to 2.6.1. What's the reason for doing this without updating protractor-accessibility-plugin?

ascott1:
Ugh, npm. Commit incoming.

ascott1:
@Scotchester @sebworks can you pull this down and review it?

ascott1:
Ha, thanks @jimmynotjim!

PR #3726, opened 2018-01-08
Remove housing couselor and other dependencies on PDFReactor

willbarton:
This PR removes our dependencies on PDFReactor from the Housing Counselor API in a transparent way for end-users. Previously, visiting https://www.consumerfinance.gov/save-hud-counselors-list/?zip=12345 would generate a PDF using PDFReactor instead of serving our pre-generated PDFs from S3. This PR changes the implementation of that endpoint to redirect to the S3 PDF.

It also removes the unused `code.services.PDFGeneratorView` and the `PDFREACTOR_LIB` setting.

## Checklist


cfgov/legacy/views/housing_counselor.py
```diff
@@ -58,74 +73,22 @@ def get_counselors(cls, request, zipcode):
...
-        return self.generate_pdf_from_url(render_url, filename)
-
-    def get_render_url(self, zipcode):
-        html_path = reverse('housing-counselor')
-        return 'http://localhost{}?zipcode={}&pdf'.format(html_path, zipcode)
-
-    def generate_pdf_from_url(self, url, filename):
-        result = self.pdfreactor.renderDocumentFromURL(url)
+        s3_pdf_url = self.s3_pdf_url(zipcode)
+        result = requests.get(s3_pdf_url)
```
willbarton: @rosskarchner @chosak I opted to proxy the S3 request rather than redirect, to avoid breaking any status code checks or cause problems for API-consuming code that doesn't expect a redirect. Since you two probably have the best idea of how the API endpoint might be used, is that a good idea, or do you think a redirect would be sufficient?

chosak: Can we do either of those via Apache, so that Django doesn't even have to be involved? Here if, for example, the `requests.get` returns a 500 for some reason, we should handle it somehow.

I might lean towards using a temporary redirect via Apache config to reduce load on the webservers.

willbarton: How would that work since Apache can't parse query strings and to construct the URL to redirect to we'd need to extract the zipcode from the query string? Or are you proposing a URL in between S3 and this endpoint that we can use?

chosak: Is it possible to use the mod_rewrite module to redirect based on query strings, like [this](https://perishablepress.com/redirect-query-string-htaccess/)?

willbarton: Scratch that last — I think I see how we could do it with mod_rewrite and a RewriteCond that captures the zipcode. Let me play around.

rosskarchner: I think I would vote for redirection of proxying?-- though I'm having trouble finding the exact words to justify it. Proxying just seems unnecessarily heavy.

I guess I can imagine an "API Client" that doesn't handle HTTP redirects (after all, there was that one that couldn't handle the HTTPS switchover...), but I'd almost rather try the redirect *first*, and change it if we get pushback from users (like we did with HTTPS).

As far as doing the redirection in Apache or Python-- I have a weak preference for Apache, though at least in terms of our ability to understand our own code, the Python implementation would almost be more readable than the mod_rewrite incantations.

willbarton: @rosskarchner certainly redirecting is vastly simpler. If you're both comfortable with that, we can try it and if we get pushback we can implement a mod_rewrite proxy.

cfgov/legacy/views/housing_counselor.py
```diff
@@ -58,74 +74,13 @@ def get_counselors(cls, request, zipcode):
...
-        reactor.setAddBookmarks(True)
-
-        # Ensure that reactor can be connected to.
-        try:
-            reactor.renderDocumentFromContent('')
-        except Exception:
-            raise RuntimeError('Cannot connect to local PDFReactor')
-
-        return reactor
+        return redirect(self.s3_pdf_url(zipcode))
```
willbarton: @rosskarchner @chosak for illustration, here's a redirect.

cfgov/legacy/views/housing_counselor.py
chosak: I have a moderately strong preference for implementing as redirect over proxy and using Apache if possible over Django, to reduce Python complexity and load on servers. My ideal would be a world where Django/Wagtail has a single page that serves the UI, and all other requests (JSON/PDF) don't touch Django/Wagtail at all, and only go to S3 or (for legacy purposes) Apache.

cfgov/legacy/views/housing_counselor.py
chosak: This looks great - minor typo "councelor".

rosskarchner:
# 🕺 🕺 🕺 🕺 🕺 🕺 🕺 🕺 🕺 🕺 🕺 🕺 🕺 🕺 🕺 🕺 🕺 🕺 🕺

rosskarchner:
(previous travis failure was on their side)

PR #3731, opened 2018-01-10
Add second email popup for debt collection pages

chosak:
This change modifies the existing JS email popup so that it can appear on a wider set of pages, and also supports multiple simultaneous popups at the same time with different content and target email lists.

It leaves the existing Owning a Home popup on these pages:

- /owning-a-home/
- /owning-a-home/mortgage-estimate/

and adds a new Debt Collection popup on these pages:

- /ask-cfpb/what-is-a-statute-of-limitations-on-a-debt-en-1389/
- /ask-cfpb/what-is-the-best-way-to-negotiate-a-settlement-with-a-debt-collector-en-1447/
- /ask-cfpb/what-should-i-do-when-a-debt-collector-contacts-me-en-1695/
- /consumer-tools/debt-collection/

This new popup will not turn on until February 1st, and can be configured using the `EMAIL_POPUP_DEBT` feature flag. The existing `EMAIL_POPUP` feature flag that was previously being used to control the OaH popup has now been renamed to `EMAIL_POPUP_OAH`, and is enabled by default.

Specific popup content and configuration, including text/image content and the mailing list to subscribe users to, can be configured using the Django template files in

`cfgov/jinja2/v1/_includes/organisms/email-popup/`

for example, `oah.html` and `debt.html`. This use of code-based configuration instead of full integration in the Wagtail editor is deliberate.

A new `email_popup` template tag has been created that works in both Django and Jinja templates. This tag inserts the HTML code for the email popup if the page request is configured for a popup in settings. The necessary JS code for each page must also be setup through the use of `cfgov/unprocessed/js/routes` files.

The existing email popup uses two browser local storage keys to store information about how many times the user has seen the popup, and when they should see it next. These keys are used to throttle how often the popup appears.

The current OaH popup uses the keys `oahPopupCount` and `oahPopupShowNext`. The new debt popup uses the keys `debtPopupCount` and `debtPopupShowNext`. For debugging/testing purposes these keys can be viewed and deleted in Chrome Developer Tools under Application, Storage, Local Storage, and then the domain of the site you are testing against (e.g. localhost).

## Additions

- New email popup for debt collection pages (behind feature flag until 2/1).

## Changes

- Refactor of email popup code to support multiple sets of pages with independent popups.

## Testing

The best way to test this change (besides the unit tests) is to run a local server and visit the pages listed above to see if the popup appears. In order to enable the debt popup before the go-live date of 2/1, you'll need to manually set the `EMAIL_POPUP_DEBT` flag to something via the Wagtail admin, for example a boolean condition of "True".

## Screenshots

image

## Todos

The use of `modal-house.png` for the debt popup [here](https://github.com/cfpb/cfgov-refresh/compare/email-popup#diff-afca900319cd83827ac48ee8eacf7498R4) is a temporary placeholder pending new imagery.

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -620,8 +620,9 @@ FLAGS = {
     # Intended for use with path conditions.
     'AB_TESTING': {},
 
-    # When enabled, should display the email popup.
-    'EMAIL_POPUP': {},
+    # Email popups.
+    'EMAIL_POPUP_OAH': {'boolean': True},
+    'EMAIL_POPUP_DEBT': {'after date': '2018-02-01T00:00'},
 
     # The next version of eRegulations
     'EREGS20': {
@@ -673,3 +674,17 @@ SEARCH_DOT_GOV_ACCESS_KEY = os.environ.get('SEARCH_DOT_GOV_ACCESS_KEY')
 SERVE_LATEST_DRAFT_PAGES = []
 if DEPLOY_ENVIRONMENT == 'beta':
     SERVE_LATEST_DRAFT_PAGES = [1288]
+
+# Email popup configuration. See v1.templatetags.email_popup.
+EMAIL_POPUPS = {
+    'debt': [
+        '/ask-cfpb/what-is-a-statute-of-limitations-on-a-debt-en-1389/',
+        '/ask-cfpb/what-is-the-best-way-to-negotiate-a-settlement-with-a-debt-collector-en-1447/',
+        '/ask-cfpb/what-should-i-do-when-a-debt-collector-contacts-me-en-1695/',
+        '/consumer-tools/debt-collection/',
+    ],
+    'oah': [
+        '/owning-a-home/',
+        '/owning-a-home/mortgage-estimate/',
+    ],
+}
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -676,7 +676,7 @@ if DEPLOY_ENVIRONMENT == 'beta':
     SERVE_LATEST_DRAFT_PAGES = [1288]
 
 # Email popup configuration. See v1.templatetags.email_popup.
-EMAIL_POPUPS = {
+EMAIL_POPUP_URLS = {
     'debt': [
         '/ask-cfpb/what-is-a-statute-of-limitations-on-a-debt-en-1389/',
         '/ask-cfpb/what-is-the-best-way-to-negotiate-a-settlement-with-a-debt-collector-en-1447/',
--- a/cfgov/jinja2/v1/_includes/organisms/email-popup/base.html
+++ b/cfgov/jinja2/v1/_includes/organisms/email-popup/base.html
@@ -41,9 +41,7 @@
                     <p>
                         <a href="{{ privacy_link_url }}"
                            target="_blank"
-                           rel="noopener noreferrer">
-                            {{ privacy_link_text }}
-                        </a>
+                           rel="noopener noreferrer">Privacy Act statement</a>
                         <br/>
                     </p>
                     <input class="a-btn a-btn__full-on-xs" type="submit" value="Sign up">
--- a/cfgov/jinja2/v1/_includes/organisms/email-popup/base.html
+++ b/cfgov/jinja2/v1/_includes/organisms/email-popup/base.html
@@ -6,29 +6,29 @@
 
  Description:
 
- Sidebar popup that fades in upon scrolling on the Owning a Home sublanding
- page at /owning-a-home/.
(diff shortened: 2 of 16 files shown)


cfgov/v1/templatetags/email_popup.py
```diff
@@ -0,0 +1,26 @@
...
+from django.utils.safestring import mark_safe
+
+from flags.state import flag_enabled
+
+
+register = template.Library()
+
+
+@register.simple_tag
+def email_popup(request):
```
sebworks: This is interesting --- I'm not crazy about using context functions but it does provide a cleaner way for handling lists of URLs.

cfgov/cfgov/settings/base.py
```diff
@@ -673,3 +674,17 @@
 SERVE_LATEST_DRAFT_PAGES = []
 if DEPLOY_ENVIRONMENT == 'beta':
     SERVE_LATEST_DRAFT_PAGES = [1288]
+
+# Email popup configuration. See v1.templatetags.email_popup.
```
anselmbradford: Should this be `See v1/templatetags/email_popup.py`?

cfgov/jinja2/v1/_includes/organisms/email-popup/base.html
```diff
@@ -39,17 +39,17 @@ <h2>Buying a home?</h2>
                                required>
                     </div>
                     <p>
-                        <a href="/owning-a-home/privacy-act-statement/"
+                        <a href="{{ privacy_link_url }}"
                            target="_blank"
                            rel="noopener noreferrer">
-                            View Privacy Act statement
+                            {{ privacy_link_text }}
```
anselmbradford: Are we sure it's not a mistake that the "View Privacy Act statement" text differs across apps? Seems like they should/could be the same site-wide.

cfgov/jinja2/v1/_includes/organisms/email-popup/oah.html
```diff
@@ -0,0 +1,8 @@
+{% extends "base.html" %}
+{% set popup_headline = "Buying a home?" %}
+{% set popup_content = "Sign up for email tips and info to help you through the process." %}
+{% set popup_image = "img/modal-house.png" %}
+{% set mailing_list_code = "USCFPB_128" %}
```
anselmbradford: Trivial: it would be nice if the variable order was the same in debt.html and oah.html

cfgov/jinja2/v1/_includes/organisms/email-popup/base.html
chosak: I agree this could be the same. This configurability was based on input from stakeholders; I will double-check.

sebworks:
I don't think you are using `cfgov/unprocessed/js/routes/ask-cfpb/index.js`.

chosak:
@sebworks thanks, addressed in d472cb8.

cfgov/jinja2/v1/_includes/organisms/email-popup/oah.html
chosak: @anselmbradford thanks, addressed in 327e494.

cfgov/unprocessed/css/organisms/email-popup.less
```diff
@@ -67,7 +67,7 @@
     }
 
     .respond-to-max( @bp-xs-max, {
-        background: none;
+        background: none !important;
```
anselmbradford: This is to override the inline background-image? Perhaps make it specifically `background-image: none !important;`

chosak: @anselmbradford I'll defer to your frontend knowledge on this; would that cause any issues because it would leave [these other attributes set](https://github.com/cfpb/cfgov-refresh/pull/3731/files/255f54268d76d6386bda175e2a53170707861a2c#diff-dc1434e393a4b592cabc175ec1226462R83), or are those two sections mutually exclusive?

cfgov/jinja2/v1/_includes/organisms/email-popup/base.html
chosak: Per feedback, we are going to standardize on "Privacy Act statement" everywhere. Updated in dbded65.

cfgov/unprocessed/css/organisms/email-popup.less
anselmbradford: Those _should_ be fine, because they'd only apply to an image set via `background-image`. It's probably fine as-is, but maybe add a code comment above it that says why the `!important` is needed, as important usage is discouraged by linters and such.

chosak: Thanks @anselmbradford, addressed in f7ecede.

PR #3735, opened 2018-01-12
Pass additional text to expandable cue label 

richaagarwal:
See platform issue 2393 for context.

The expandable text we see on filterable list pages, such as https://www.consumerfinance.gov/about-us/blog, will now display "Show filters" and "Hide filters" instead of "Show" and "Hide".  Use of the expandable on non-filterable list pages will remain unaffected, as it is an optional parameter.

Screenshots:
screen shot 2018-01-12 at 1 22 36 pm
screen shot 2018-01-12 at 1 22 30 pm

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/expandable.html
+++ b/cfgov/jinja2/v1/_includes/organisms/expandable.html
@@ -28,7 +28,7 @@
 
    ========================================================================== #}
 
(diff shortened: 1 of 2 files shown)


nataliafitzgerald:
@richaagarwal - Looks great!

PR #3742, opened 2018-01-17
Narrows OAH scripts scope

anselmbradford:
There were two of the same script on the owning-a-home route.

## Removals

- ~~Removes `unprocessed/js/routes/owning-a-home/single.js`~~
- ~~Removes `unprocessed/js/routes/owning-a-home/mortgage-estimate/single.js`~~

## Changes

- Changes `unprocessed/js/routes/owning-a-home/mortgage-estimate/single.js` to `unprocessed/js/routes/owning-a-home/mortgage-estimate/index.js` so it only applies to the `/owning-a-home/mortgage-estimate/` page.
- Changes `unprocessed/js/routes/owning-a-home/single.js` to `unprocessed/js/routes/owning-a-home/index.js` so it only applies to the `/owning-a-home/` page.

## Testing

1. `gulp build`
2. Clear local storage
2. Visit http://localhost:8000/owning-a-home/mortgage-estimate/ and it should have a popup.


anselmbradford:
> Don't we still need owning-a-home/single.js to ensure that the popup also exists on the root OaH page? I believe that's still a requirement, see GHE#1217.

Ahh, right you are. I was confused over what was required here. Looks like we can actually get rid of /owning-a-home/mortgage-estimate

chosak:
@anselmbradford will that work to limit the popup to just those two pages, though, and not any other pages under OaH?

anselmbradford:
> @anselmbradford will that work to limit the popup to just those two pages, though, and not any other pages under OaH?

Ohhhh okay, then we need `/owning-a-home/index.js` and `/owning-a-home/mortgage-estimate/index.js`. `/owning-a-home/single.js` is all pages under the `/owning-a-home/` URL.

`/owning-a-home/index.js` = www.consumerfinance.gov/owning-a-home/ ONLY
`/owning-a-home/mortgage-estimate/index.js` = www.consumerfinance.gov/owning-a-home/mortgage-estimate/ ONLY

PR #3745, opened 2018-01-17
Minor scripts task cleanup

anselmbradford:
## Changes

- Names vinyl-named import variable `vinylNamed`.
- Compacts vinyNamed function to use arrow function.

## Testing

1. `gulp scripts:modern`


sebworks:
can gulp-replace be removed from package.json

anselmbradford:
It's used at https://github.com/cfpb/cfgov-refresh/blob/master/gulp/tasks/scripts.js#L174, so not yet

PR #3746, opened 2018-01-17
Update to gulp 4

anselmbradford:
## Changes

- Updates to gulp 4.
- Reorders default gulp task so it goes build, lint, test, instead of lint, test, and build. For some reason it would throw an error in the original order.
- Replace task array parameter with native `gulp.series` and `gulp.parallel`.
- Minor re-ordering of `scripts.js` tasks.

## Testing

1. `npm install`
2. `gulp`, `gulp build`, `gulp copy`, `gulp images`, `gulp test`, `gulp lint`, and other tasks should pass.


gulp/tasks/watch.js
```diff
@@ -4,8 +4,12 @@
 const gulp = require( 'gulp' );
 const config = require( '../config' );
 
-gulp.task( 'watch', [ 'browsersync' ], function() {
-  gulp.watch( config.scripts.src, [ 'scripts' ] );
-  gulp.watch( config.styles.cwd + '/**/*.less', [ 'styles' ] );
-  gulp.watch( config.images.src, [ 'images' ] );
-} );
+gulp.task( 'watch',
+  gulp.series(
```
Scotchester: Why make this series, here? Doesn't seem like it was before.

gulp/tasks/watch.js
```diff
@@ -4,8 +4,12 @@
...
+    'browsersync', () => {
```
Scotchester: Should the anonymous arrow function here be on the next line?

gulpfile.js
```diff
@@ -8,7 +8,27 @@
...
+
+gulp.task( 'build',
+  gulp.series(
+    gulp.parallel(
+      'styles',
+      'scripts',
+      'images'
+    ),
+    'copy'
+  )
```
Scotchester: The way you've made this a series makes sense to me intuitively, but I'm confused about why it didn't need to be that was before.

gulpfile.js
```diff
@@ -8,7 +8,27 @@
...
+);
+
+gulp.task( 'default',
+  gulp.series(
+    'build',
+    'lint:scripts',
+    'test:unit'
+  )
```
Scotchester: Is there a reason these need to be in series? Also, for what it's worth, I was able to successfully test putting `build` back to the end in my local environment.

package.json
```diff
@@ -36,7 +36,7 @@
     "cfpb-chart-builder": "3.2.1",
     "del": "3.0.0",
     "glob-all": "3.1.0",
-    "gulp": "3.9.1",
+    "gulp": "4.0.0",
```
Scotchester: Any idea why `"gulp": "4.0.0"` would actually install Gulp 4.0.0 when 4.0.0 hasn't been published to [npm](https://www.npmjs.com/package/gulp) yet? 3.9.1 is the latest published as of the time of this comment.

gulpfile.js
sebworks: Yeah, it would be nice if these ran in parallel.

anselmbradford: Updated to parallel!

gulpfile.js
anselmbradford: before it was `'styles', 'scripts', 'images', 'copy'`, so in series for each task. This changes it so the first three run in parallel, which in theory should be more efficient.

gulp/tasks/watch.js
anselmbradford: Makes sense. Moved it to the next line.

package.json
anselmbradford: Well that's a mystery. I have no idea.

gulp/tasks/watch.js
anselmbradford: I think it was? Before it defined that the `browsersync` task should run before the anonymous function when running `gulp watch`, and that's how it's set up with series.

package.json
Scotchester: We figured this out while you were gone. I forget the answer. Something to do with 4.0.0 being a prerelease or something?

gulpfile.js
Scotchester: I thought that previously Gulp ran things in parallel unless told otherwise, hence the previous need for `runSequence` in some places?

anselmbradford: Ah right, you are correct. Thinking back to this, I think I had them all globbed together and it would fail when `copy` was included. We could try fiddling with this task and see if it can be simplified.

Scotchester: Just to be clear: This makes sense as you have it here. I'm just not sure how copy didn't run before build stuff completed before :)

package.json
jimmynotjim: It was published as `next`, which @Scotchester I wonder if that's what we should do with the icons branch??

Scotchester: That's a very interesting thought, @jimmynotjim...

PR #3747, opened 2018-01-19
Ask CFPB: Make sure indexed pages are live

higs4281:
Ask category pages get metadata on related pages from Elasticsearch to
avoid heavy requests, and the metadata was allowing references to
non-live pages. This adds a filter that is invoked at index time,
and adjusts tests to check that facet_map pages are live.

Addresses GHE issue 2417

### DIFF
--- a/cfgov/ask_cfpb/models/django.py
+++ b/cfgov/ask_cfpb/models/django.py
@@ -151,7 +151,9 @@ class Category(models.Model):
 
     @cached_property
     def facet_map(self):
-        answers = self.answer_set.order_by('-pk').select_related()
+        raw_answers = self.answer_set.order_by('-pk').select_related()
+        answers = [
+            answer for answer in raw_answers if answer.english_page.live]
         subcats = self.subcategories.all().select_related()
         audiences = Audience.objects.all()
         container = {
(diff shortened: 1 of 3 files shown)


cfgov/ask_cfpb/models/django.py
```diff
@@ -151,7 +151,9 @@ def top_tags_es(self):
 
     @cached_property
     def facet_map(self):
-        answers = self.answer_set.order_by('-pk').select_related()
+        raw_answers = self.answer_set.order_by('-pk').select_related()
+        answers = [
+            answer for answer in raw_answers if answer.english_page.live]
```
rosskarchner: Would the result be the same if you filtered by 'live' at query-time?

Like  `self.answer_set.filter(live=True).order_by('-pk').select_related()`

higs4281: the live value isn't available in an answer query, because it lives on the english or spanish page. answer.english_page is a property that can be harvested after the query.

higs4281: It's not efficient, but this operation is done during elasticsearch indexing, not during live requests.

PR #3750, opened 2018-01-19
MakeFilterableListMixin respect Wagtail sites

chosak:
This change modifies how the `FilterableListMixin` page mixin class selects and limits the pages that it filters, making sure that it only filters pages that are in the same Wagtail Site as itself.

Currently this mixin filters any live `AbstractFilterPage`, even those that are elsewhere in the page tree. This causes problems if, for example, a live page is moved to the Trash, or to some other Site running in the same Wagtail instance.

With this change, the default behavior of Page classes that inherit from `FilterableListMixin` is to only include pages in their same Site.

This change includes some minor refactoring of how the filterable queryset gets passed around, renaming some variables `filterable_` for clarity and using the page's own `Page.get_site` method for convenience. Some logic around page category filtering was also collected in one place, with new tests added.

The default `FilterableListMixin` behavior continues to filter only direct children of the page, and use 10 results per page. This behavior can now be overridden for inherited classes (like `NewsroomLandingPage`) using some class attributes; this removes some code duplication.

One side effect of this change is that any filterable page that lives outside of any Site will be unable to filter any pages. This seemed like a reasonable decision to me: if a page isn't part of a Site, then it's unclear where it should get its filterable pages from.

This PR is a followup to work started in #3580.

## Testing

See unit test changes; Additionally, using a production dump, try publishing an `BlogPage` under the Trash, tagging it with a Blog tag (like "At the CFPB") and then visiting the newsroom page at http://localhost:8000/activity-log/. You won't see the page there, unlike the current behavior where it will show up. Similarly, you'll be able to visit that page's feed at http://localhost:8000/activity-log/feed, unlike the current behavior which raises a 500.

## Checklist

### DIFF
--- a/cfgov/v1/models/sublanding_page.py
+++ b/cfgov/v1/models/sublanding_page.py
@@ -85,11 +85,11 @@ class SublandingPage(CFGOVPage):
                         and 'archive' not in p.title.lower()]
         posts_list = []
         for page in filter_pages:
-            base_query = AbstractFilterPage.objects.live().filter(
+            eligible_children = AbstractFilterPage.objects.live().filter(
                 CFGOVPage.objects.child_of_q(page)
             )
 
-            form = FilterableListForm(base_query=base_query)
+            form = FilterableListForm(filterable_pages=eligible_children)
             for post in form.get_page_set():
                 posts_list.append(post)
         return sorted(posts_list,
--- a/cfgov/v1/forms.py
+++ b/cfgov/v1/forms.py
@@ -91,19 +91,18 @@ class FilterableListForm(forms.Form):
     )
 
     def __init__(self, *args, **kwargs):
-        self.base_query = kwargs.pop('base_query')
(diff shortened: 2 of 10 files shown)


cfgov/v1/models/sublanding_page.py
```diff
@@ -85,11 +85,11 @@ def get_browsefilterable_posts(self, limit):
                         and 'archive' not in p.title.lower()]
         posts_list = []
         for page in filter_pages:
-            base_query = AbstractFilterPage.objects.live().filter(
+            eligible_children = AbstractFilterPage.objects.live().filter(
```
richaagarwal: Is there a reason this variable is called `eligible_children` as opposed to `filterable_pages`, which it's later passed in as?

chosak: @richaagarwal no reason; I thought it was a bit more explanatory, as these pages are the children of this page that are eligible to be included in the list of posts.

PR #3751, opened 2018-01-23
Do not index filterable list queries unless it's a single topic tag

richaagarwal:
See GHE/platform/issues/2403 for context and https://developers.google.com/search/reference/robots_meta_tag

### DIFF
--- a/cfgov/v1/tests/test_filterable_list.py
+++ b/cfgov/v1/tests/test_filterable_list.py
@@ -3,9 +3,7 @@ from django.test import RequestFactory, TestCase
 import mock
 
 from v1.forms import FilterableListForm
-
-from ..util.filterable_list import FilterableListMixin
-
+from v1.util.filterable_list import FilterableListMixin
 
 class TestFilterableListMixin(TestCase):
     def setUp(self):
@@ -47,3 +45,26 @@ class TestFilterableListMixin(TestCase):
                 return None
 
         self.assertFalse(MockPageInDefaultSite().filterable_pages().exists())
+
+    def test_do_not_index_is_false_by_default(self):
+        assert self.mixin.do_not_index == False
+
+    def test_do_not_index_is_false_if_no_query(self):
+        request_string = ''
+        self.mixin.get_form_data(self.factory.get(request_string).GET)
+        assert self.mixin.do_not_index == False
+
+    def test_do_not_index_is_true_if_query(self):
+        request_string = '/?categories=test1&topic=test2'
+        self.mixin.get_form_data(self.factory.get(request_string).GET)
+        assert self.mixin.do_not_index == True
+
+    def test_do_not_index_is_false_if_query_is_single_topic(self):
+        request_string = '/?topic=test1'
+        self.mixin.get_form_data(self.factory.get(request_string).GET)
+        assert self.mixin.do_not_index == False
+
+    def test_do_not_index_is_true_if_query_is_multiple_topics(self):
+        request_string = '/?topic=test1&topic=test2'
+        self.mixin.get_form_data(self.factory.get(request_string).GET)
+        assert self.mixin.do_not_index == False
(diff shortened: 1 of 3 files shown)


cfarm:
How can I test this out to see that it's working?

richaagarwal:
@cfarm I still need to add tests, but you can go to a filterable list page locally e.g. http://localhost:8000/about-us/blog/?title=&topics=financial-education&topics=mortgages&from_date=&to_date= and confirm that noindex appears (if it should) in the header, using Chrome's network tab. 
screen shot 2018-01-23 at 1 25 05 pm

cfgov/v1/tests/test_fragment_cache_extension.py
```diff
@@ -32,7 +32,7 @@ def test_cache_gets_called_when_visiting_filterable_page(self, add_to_cache):
         )
         publish_page(page)
 
-        # Add a child to that filterable page so that there are related posts results
+        # Add a child to that filterable page so that there are results with a post preview
```
richaagarwal: This is a minor change and I got lazy so I added it to this PR.  I can open a separate PR for it if anyone feels strongly.

cfgov/v1/tests/test_filterable_list.py
```diff
@@ -3,9 +3,7 @@
 import mock
 
 from v1.forms import FilterableListForm
-
-from ..util.filterable_list import FilterableListMixin
-
+from v1.util.filterable_list import FilterableListMixin
```
willbarton: :heart:

cfgov/v1/util/filterable_list.py
```diff
@@ -104,13 +112,18 @@ def get_form_data(self, request_dict):
...
 
     def serve(self, request, *args, **kwargs):
-        """ Modify response header to set a shorter TTL in Akamai """
+        """ Modify response headers """
         response = super(FilterableListMixin, self).serve(request)
+        # Set a shorter TTL in Akamai
         response['Edge-Control'] = 'cache-maxage=10m'
+        # Set noindex for crawlers if needed
+        if self.do_not_index:
+            response['X-Robots-Tag'] = 'noindex'
```
rosskarchner: I didn't know this was an option-- cool!

cfgov/v1/tests/test_filterable_list.py
```diff
@@ -47,3 +45,26 @@ def get_site(self):
                 return None
 
         self.assertFalse(MockPageInDefaultSite().filterable_pages().exists())
+
+    def test_do_not_index_is_false_by_default(self):
+        assert self.mixin.do_not_index == False
```
willbarton: For consistency, would it make sense to use `self.assertFalse()` here and in the methods below?

cfgov/v1/util/filterable_list.py
```diff
@@ -94,6 +97,11 @@ def process_form(self, request, form):
         filter_data['form'] = form
         return filter_data
 
+    def set_do_not_index(self, field, value):
+        """ Do not index on params except topics, or multiple params """
```
willbarton: Minor: I'm not sure I understand the docstring. I understand the method from reading it, set to index on 1-member long `topic` fields only. But I think the docstring could be better.

cfgov/v1/tests/test_filterable_list.py
richaagarwal: @willbarton I was just following convention within this file when I added these tests, but I had the same thought.  It seems like we've got a few different patterns going on in our back-end tests that I'd like us to standardize on at some point, but I feel that's out of scope for this PR.

cfgov/v1/util/filterable_list.py
richaagarwal: How's "Do not index queries unless they consist of a single topic field" ?

PR #3752, opened 2018-01-24
Change agreements loader to use zip files directly

rosskarchner:
This fixes a few open bugs, related to:

- certain characters in the issuer or agreement name were causing agreements to be skipped, silently.
- re-occurring file names would cause agreements to be skipped, silently
- non-ascii characters could break the loader during uploads to S3

## Additions

- You can optionally pass in a `--encoding `argument, in case filenames (within the zip file) are encoded with a charachter set other than the default, windows-1252

## Changes

- the `--path` argument must now be a zip file, not a directory on disk

## Todos

- update the related Jenkins jobs

## Checklist


rosskarchner:
@higs4281 good call

rosskarchner:
@higs4281 done

cfgov/agreements/management/commands/util.py
```diff
@@ -1,64 +1,71 @@
...
+from agreements.models import Issuer
+import boto3
+
 from slugify import slugify
 
 
-def clear_tables():
-    Agreement.objects.all().delete()
+def s3_safe_key(path, prefix=''):
+    key = prefix + path.encode('utf-8')
```
higs4281: I don't think we need to encode the path at this point; AWS says keys should be delivered as unicode, and the uploads work for me without it.
Also, I would remove comma in addition to the other chars.

cfgov/agreements/management/commands/util.py
```diff
@@ -1,64 +1,71 @@
...
 
-    issuer.name = name
-    issuer.slug = slugify(name, to_lower=True)
-    issuer.save()
 
-    return issuer
+def save_agreement(agreements_zip, raw_path, filename_encoding,
+                   outfile, upload=False):
+    uri_hostname = 'http://files.consumerfinance.gov'
+    s3_prefix = '/a/assets/credit-card-agreements/pdf/'
```
higs4281: the `s3_prefix` var shouldn't have a leading slash. S3 uploads wouldn't work for me unless I removed it.

This also requires an adjustment to the `save_agreement` function, to keep the uri in sync; something like

```
    agreement = issuer.agreement_set.create(
        file_name=filename,
        size=int(zipinfo.file_size),
        uri="{}/{}".format(uri_hostname, s3_key),
        description=filename)
```

cfgov/agreements/management/commands/import_agreements.py
```diff
@@ -1,10 +1,12 @@
...
+import fnmatch
 
 from django.core.management.base import BaseCommand
 
-from .util import clear_tables, update_agreement, update_issuer, upload_to_s3
+from agreements.management.commands import util
+from agreements.models import Issuer, Agreement
 
 
 class Command(BaseCommand):
```
higs4281: Could you update the command's docstring to reflect that source is a zip file, not csv?

cfgov/agreements/management/commands/util.py
rosskarchner: oh, interesting-- I wonder why that didn't bite me

higs4281: I was testing from laptop, not a server?

cfgov/agreements/management/commands/import_agreements.py
rosskarchner: Latest commit addresses this

cfgov/agreements/management/commands/util.py
rosskarchner: Latest commit addresses this

cfgov/agreements/management/commands/util.py
rosskarchner: Latest commit addresses this

PR #3754, opened 2018-01-24
Fix unit test: always sort lists to keep ordering stable, avoid potential breakage

acrewdson:
When running the `tox` tests in a Docker container (filesystem behavior differences are my only guess as to why this has not come up before), I noticed that this test would sometimes break because of the ordering of the items in the first list. When we sort the lists, I can no longer observe the problem.


chosak:
@acrewdson thanks, nice catch!

PR #3767, opened 2018-02-01
add mysql create user statement

OrlandoSoto:
Problem: in mysql 5.7.7 and newer, if the user root@127.0.0.1 does not already exist then the script fails. In previous mysql versions the GRANT statement would create the user so this script works correctly in those versions of mysql.  In mysql 5.7.7 and newer, the GRANT statement no longer creates the user so this script fails. 

## Additions

- CREATE USER IF NOT EXISTS

## Removals

- None

## Changes

-

## Testing

1.  in the terminal: ./create-mysql-db.sh

## Notes

- See: https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_auto_create_user 

## Todos

-

## Checklist


create-mysql-db.sh
```diff
@@ -18,15 +18,15 @@ MYSQL=`which mysql`
 if [ -z "$1" ]
 then
   Q1="CREATE DATABASE IF NOT EXISTS $MYSQL_NAME;"
-  Q2="GRANT ALL ON *.* TO '$MYSQL_USER'@'$MYSQL_HOST' IDENTIFIED BY '';"
+  Q2="CREATE USER IF NOT EXISTS '$MYSQL_USER'@'$MYSQL_HOST'; GRANT ALL ON *.* TO '$MYSQL_USER'@'$MYSQL_HOST' IDENTIFIED BY '';"
```
richaagarwal: This is minor, but would it make sense to create a new line for this, so that "GRANT ALL" ends up as "Q3" and so forth.  You'd have to update the last line SQL="${Q1}${Q2}${Q3}${Q4}" but I think that'd be more readable and follow existing convention within this file.

OrlandoSoto: 👍

PR #3748, opened 2018-01-19
Careers: add granular location data

virginiacc:
Add the ability to associate granular location data (states and cities) with job regions and offices so that information can be added to job posting pages

## Additions

- City and state models in jobmanager
- Separate region and office models
- Ability to associate cities and states with regions and one city with an office

## Testing

1. Check out branch
2. Run `python cfgov/manage.py migrate`
3. Run `python cfgov/manage.py runscript create_career_locations`
4. Verify that cities and states can be added to job offices and regions in jobmanager admin
5. Publish a job posting page with a region location and verify that hidden city and state data is present
6. Publish a job posting page with an office location and verify that location data is displayed in page
7.  Publish a job posting page with an office location and remote option checked and verify that location data is displayed in page

### Office admin
screen shot 2018-01-25 at 9 26 24 am

### Region admin
admin

### Job posting with region location and hidden data
screen shot 2018-02-02 at 2 15 07 pm

### Job posting with office location
screen shot 2018-01-25 at 9 36 12 am

### Job posting with office location and remote option
screen shot 2018-01-25 at 9 36 30 am

## Checklist

* [ ] PR has an informative and human-readable title
* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [development playbook](https://github.com/cfpb/development)
* [ ] Passes all existing automated tests
* [ ] Any *change* in functionality is tested
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated
* [ ] Reviewers requested with the [Reviewer tool](https://help.github.com/articles/about-pull-request-reviews/) :arrow_right:

PR #3734, opened 2018-01-12
Proof of concept: Postgres Migrator

rosskarchner:
This script attempts to migrate all available Django models from MySQL into Postgres. I started [here](https://rlskoeser.github.io/2016/03/31/migrating-data-between-databases-with-django/) and troubleshot it until I had something that worked.  This has *not* been tested with any optional_apps yet, just with code that lives in cfgov-refresh.

I won't say it's particularly good software, but perhaps it can be stripped for parts.

~~## A mystery~~

~~Somehow, when this script is run, Documents and Images disappear from MySQL, even though (as far as I can tell) nothing in the code writes (or deletes!) from MySQL. They don't land in Postgres either.~~

## Thoughts on other approaches

- As noted in #3642, I [couldn't get pgloader](https://github.com/cfpb/cfgov-refresh/pull/3642#issuecomment-349729768) to work
- At least locally, Django's `dumpdata` worked fine, but trying to feed that file into `loaddata` kept giving me memory errors.
- Before the holiday's, I was playing with some ideas for a file-based export (to support Postgres migration, and solve some other problems). It's not yet in a working state (I'll need to port over some of the model-ordering work from `populate_postgres`). 

## Additions

- 'populate_postgres' script.

## Changes

- made the EventPage 'youtube_url' validator a little more forgiving (accommodating values already in the database), relevant if you are calling save(using='postgres') on them.
- I made refresh-data.sh *stop* automatically dropping and recreating the database. That behavior creates a situation where you get missing tables (anything not included in the database dump), which is a problem if you are looping through all available models.

## Testing

- I'd start with a fresh MySQL. If you're lazy like me, stop docker-compose (if it's running), and `docker-compose rm mysql`-- then the next time you `docker-compose up` you'll have a fresh MySQL.
- via `shell.sh`, run the migrations against MySQL like you normally would, `cfgov/manage.py migrate`.
- load a database dump with `./refresh-data.sh`
- Run the migrations against Postgres: `DATABASE_URL=postgres://cfgov:cfgov@postgres/cfgov cfgov/manage.py migrate`
- `cfgov/manage.py runscript populate_postgres` This will take a while.
- That script leaves the Wagtail tree in need of repair, easily done with `DATABASE_URL=postgres://cfgov:cfgov@postgres/cfgov cfgov/manage.py fixtree`
- Stop docker compose, and edit .python_env with: `DATABASE_URL=postgres://cfgov:cfgov@postgres/cfgov`
[...]

PR #3770, opened 2018-02-02
Delete and recreate Django migrations

chosak:
This commit deletes and recreates all Django migrations for the various apps in this project. It generates a similar outcome to how built-in Django squashing works, but works better with the large backlog of `RunPython` commands we've used over time.

This should significantly speed up the time it takes to create a new test database (i.e. using `tox`), although, unfortunately, this is still somewhat slow due to the number of models and migrations in third-party apps like Wagtail.

It has the effect of recreating a set of new migrations that describe the current state of all of our models, with these critical side effects:

1. Databases that exist at some migration state before the one at the point of this commit (7fddfdd) will no longer be able to be migrated to this current state, as the intermediate changes will have been lost. This means that those databases will need to be recreated.

    This also means that historical database archives will require a bit more work to resurrect; they'll need to first be migrated to the point just before this gets committed, and then updated to code at or after this point.

2. Any open pull requests that reference or depend on some of the existing migrations will need to be modified to instead refer to these new migration files. (fyi @virginiacc)

For the record, these new migrations were generated using this approach:

1. Remove all existing migration files:

    ```sh
    $ rm -f -v cfgov/*/migrations/0*
    ```

2. Create new migration files from the state of model Python code:

    ```sh
    $ cfgov/manage.py makemigrations --noinput
    ```

    As it happens this creates new initial migrations (`0001_initial`) for all apps, plus some subsequent migrations (`0002_something`) for apps that depend on other apps (for example, `ask_cfpb` has its own initial migration and then some changes that rely on `v1`).

3. Rename the created migration files so that they follow in sequence the migration files that used to exist. For example, currently there are 101 `v1` migrations, so the first new migration is numbered 102.

4. Manually alter all new migration files to indicate that they replace the old migration files.

    This involves adding lines to these files like:

    ```py
    includes = [('app_name', '0002_foo'), ('app_name', '0003_bar'), ...]
    ```

    This tells Django that these new files replace the old files, so that when migrations are run again, it doesn't need to do anything.

    Also manually update any new subsequent migration files so that they properly refer to each other. For example, if an app has two new migrations 102 and 103, the 103 file needs to properly depend on 102.

To apply these new migration files to an existing database, you can simply run:

```sh
$ cfgov/manage.py migrate --noinput
```

You'll see that there are no changes to apply, as the new files should exactly describe the current model state in the same way that the old migrations did.

This change also deletes a few unit tests that were testing functions in migrations that are removed here.

## Notes

This is a big change so a good review would be much appreciated.

## Checklist


chosak:
Adding the "hold" label because these changes need to be recreated since a migration-altering PR has been merged in (#3748). Feedback on the implementation and decision to move ahead is still welcome.

willbarton:
> 1. Databases that exist at some migration state before the one at the point of this commit (7fd3fa5) will no longer be able to be migrated to this current state, as the intermediate changes will have been lost. This means that those databases will need to be recreated.
>
> This also means that historical database archives will require a bit more work to resurrect; they'll need to first be migrated to the point just before this gets committed, and then updated to code at or after this point.

I don't think this is much of a problem. Maybe it's worth creating a git tag/release right before this change is merged (maybe this is the PR that gets us to 6.0)? That way the process in a hypothetical database restore would be to check out that tag, migrate, then checkout master and migrate again, right?

chosak:
> Maybe it's worth creating a git tag/release right before this change is merged

I think that makes sense, at least a minor release and then some documentation changes that include instructions for using it.

> maybe this is the PR that gets us to 6.0

6.0 we need to save for Django/Wagtail/Python version bump. 😄

> That way the process in a hypothetical database restore would be to check out that tag, migrate, then checkout master and migrate again, right?

Yes, that's what I had in mind.

willbarton:
> One question: Without some sort of cleanup on non-wagtail entries, the django_migrations table will be left with lots of orphaned migration references. Isn't it possible that a future migration could match one of the orphan entry names and thus refuse to run? 

@higs4281 I actually think the PostgreSQL migration might solve that problem by itself [if we use the approach in the linked PR instead of trying to munge a compatible SQL file](https://github.com/cfpb/cfgov-refresh/pull/3734).

willbarton:
@higs4281 also, I can't reactji your review post apparently (annoying, GitHub!), but 💯 for trying to old-dump restore method!

chosak:
Thanks all for the review, and apologies that it's taken me so long to come back to this. @higs4281 you raised an excellent point. By reusing the old migration numbers, we'd introduce a small change of conflict down the road.

To avoid that, I've recreated the migrations (and force-pushed them here) so that they continue the existing numbering. So, e.g. if `v1` had migrations 1-101, the new migration is numbered 102, and marked as replacing all that came before. For apps with multiple new migration files, both new files are marked as replacement, so e.g. `ask_cfpb` has new migrations 13 and 14 that are both marked as replacing 1-12.

I've updated the instructions above for future reference should we want to redo this; I played a bit with creating an automated script but it's a bit tricky as you need to edit the contents of the migration files.

If people wouldn't mind re-reviewing this, I'd very much appreciate it. As before, this is something we'll have to update if any new migrations get merged in before this does.

chosak:
Thanks for the re-review @higs4281! The merge commit (a588589) shows no changes to migration files so this PR should still be good to go in. For future reference the commit that this is going in after is 8aa57b0.

PR #3772, opened 2018-02-05
Set ALLOWED_HOSTS in production with an env var

willbarton:
This PR adds `ALLOWED_HOSTS` to the production.py settings file and reads its value from an `ALLOWED_HOSTS` env var that contains a **JSON list** of allowed hosts. If that env var is not set, an attempt at a restricted but sane, functional default (`['.consumerfinance.gov']`) is made.

JSON seemed the lowest-effort and least invasive way to define a list in  an env var, since we use env vars for host-specific configuration in Ansible. There will be a corresponding PR to our Ansible role to define the env var.

Despite the fact that host validation is already done before any traffic reaches our web server, this is being done to put us in compliance with our security baseline. 

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/production.py
+++ b/cfgov/cfgov/settings/production.py
@@ -3,8 +3,10 @@ import os
 import sys
 from os.path import exists
 
-from .base import *
-from .database_mixin import *
+from django.core.exceptions import ImproperlyConfigured
+
+from cfgov.settings.base import *
+from cfgov.settings.database_mixin import *
 
 
 default_loggers = []
@@ -121,4 +123,7 @@ CACHES = {
 try:
     ALLOWED_HOSTS = json.loads(os.getenv('ALLOWED_HOSTS'))
 except (TypeError, ValueError):
-    ALLOWED_HOSTS = ['.consumerfinance.gov']
+    raise ImproperlyConfigured(
+        "Environment variable ALLOWED_HOSTS is either not defined or is "
+        "not valid JSON. Expected a JSON array of allowed hostnames."
+    )
--- a/cfgov/cfgov/settings/production.py
+++ b/cfgov/cfgov/settings/production.py
@@ -1,3 +1,4 @@
+import json
 import os
 import sys
 from os.path import exists
@@ -8,13 +9,13 @@ from .database_mixin import *
 
 default_loggers = []
 
-# Is there a syslog device available? 
+# Is there a syslog device available?
 # selects first of these locations that exist, or None
 syslog_device = next((l for l in ['/dev/log', '/var/run/syslog'] if exists(l)), None)
 
 if syslog_device:
     default_loggers.append('syslog')
-                           
+
 # if not running in mod_wsgi, add console logger
 if not (sys.argv and sys.argv[0] == 'mod_wsgi'):
     default_loggers.append('console')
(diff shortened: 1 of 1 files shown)


cfgov/cfgov/settings/production.py
```diff
@@ -114,3 +115,10 @@
         'TIMEOUT': None,
     }
 }
+
+# ALLOWED_HOSTS should be defined as a JSON list in the ALLOWED_HOSTS
+# environment variable.
+try:
+	ALLOWED_HOSTS = json.loads(os.getenv('ALLOWED_HOSTS'))
```
chosak: I notice these two lines are indented with tabs instead of spaces. It'd be nice to be consistent within a single file. Would you mind changing these to spaces?

cfgov/cfgov/settings/production.py
```diff
@@ -114,3 +115,10 @@
...
+except ValueError:
```
chosak: If the environment variable isn't set, then `json.loads(None)` raises a `TypeError` instead of a `ValueError`. How would you feel about catching both types of exceptions (`except (TypeError, ValueError):`)?

willbarton:
@chosak good catch on the `TypeError`! And for some reason vim suddenly decided I wanted tabs instead of spaces. That upsets me greatly.

willbarton:
Values have been populated in ansible so this is safe to merge once it's reviewed.

cfgov/cfgov/settings/production.py
```diff
@@ -114,3 +115,10 @@
         'TIMEOUT': None,
     }
 }
+
+# ALLOWED_HOSTS should be defined as a JSON list in the ALLOWED_HOSTS
+# environment variable.
+try:
+    ALLOWED_HOSTS = json.loads(os.getenv('ALLOWED_HOSTS'))
+except (TypeError, ValueError):
```
chosak: Looking at this again, I'm wondering about this fallback case. In true "production" deployments, we need to make sure that our configuration (Ansible) sets this environment variable properly. If something goes wrong there and it's not set, then falling back to this wildcard actually seems wrong, as it would both allow hosts we don't want (like beta on www) and prevent hosts we do (like localhost for same-server requests).

Similarly, there are infrequent times when it's useful as a developer to run against these production settings, and the typical `runserver` workflow wouldn't include a valid `localhost` host header in that case.

If we're going to rely on this environment variable, does it make sense to require it to be set, and add it to `.env_SAMPLE` with some reasonable default value (`localhost`)? Then we could add a more verbose error message, something like (just spitballing):

```py
# settings.py
from django.core.exceptions import ImproperlyConfigured

try:
    ALLOWED_HOSTS = json.loads(os.getenv('ALLOWED_HOSTS'))
except (TypeError, ValueError):
    raise ImproperlyConfigured(
        "Use of production settings requires a valid ALLOWED_HOSTS environment "
        "variable containing a JSON list of allowed hostnames."
    )

willbarton: I was purposefully leaving it out of `base.py` and `.env_SAMPLE` (and the `.python_env_SAMPLE` for docker)  because locally we use `DEBUG = True` which ignores `ALLOWED_HOSTS` generally. So in order to test this setting locally you have to modify the settings to change `DEBUG` anyway.

Replacing the default with an `ImproperlyConfigured` exception is a good idea, I think the rationale for doing so makes sense.

willbarton: Done!

PR #3787, opened 2018-02-12
Careers: unpublish JobListing pages if closed on USAJobs

virginiacc:
USAJobs is the source of truth about the status of our job postings. When jobs close early on USAJobs (midnight Eastern), their pages are still live on our site until we manually unpublish them the next working day. This adds a script that can be automated to run after midnight to unpublish job pages if their related jobs have closed on USAJobs (ie, they have been moved to the USAJobs API archive endpoint, or their pages on USAJobs display a closed message instead of an apply button).

## Additions

- Script to unpublish live job pages if related job listings have closed on USAJobs

## Testing

1. Check out branch
2. Update USAJobs environment variables
3. Publish several expired job listing pages
4. Update database or add pages for [currently open jobs](https://www.consumerfinance.gov/about-us/careers/)
5. Run script and verify that pages for closed jobs have been unpublished and pages for open jobs remain live

## Notes

We currently only have access to jobs that are open to the public via the USAJobs archive endpoint, but that should be expanded to include internal/status jobs in an upcoming API release. In the meantime, this includes a temporary fallback check for closed text on the USAJobs job posting pages.

## Checklist


cfgov/scripts/unpublish_closed_jobs.py
```diff
@@ -0,0 +1,62 @@
...
+URL = os.environ.get('USAJOBS_ARCHIVE_URL')
+HEADERS = {
+    'Host': os.environ.get('USAJOBS_HOST'),
+    'User-Agent': os.environ.get('USAJOBS_USER'),
+    'Authorization-Key': os.environ.get('USAJOBS_API_KEY'),
+}
+
+
+def job_page_closed(link):
+    webpage = urllib2.urlopen(link)
```
willbarton: Minor thing, but if you `from urllib2 import urlopen` above and then just use `urlopen` here, it'll make porting to Python 3 a little bit easier when we do (`urlopen` is in `urllib.requests` in Python 3, so we can make `urlopen`'s import conditional).

cfgov/scripts/unpublish_closed_jobs.py
```diff
@@ -0,0 +1,62 @@
...
+    try:
+        control_number = os.path.basename(os.path.normpath(link))
+        response = requests.get(URL, headers=HEADERS, params={
+            'ControlNumber': control_number,
+            'WhoMayApply': 'all'
+        })
+        results = json.loads(response.text)['SearchResult']
+        search_items = results['SearchResultItems']
+        return search_items[0]['MatchedObjectId'] == control_number
+    except:
```
willbarton: It would be nice to specify what exceptions you expect here, and maybe log something on the following line so that there isn't an unexpected failure that can't be diagnosed when the scheduled job runs.

cfgov/scripts/unpublish_closed_jobs.py
virginiacc: This should be updated in https://github.com/cfpb/cfgov-refresh/pull/3787/commits/2b8db04cfab056d2b00745a3d9ed954b3f4ff240

cfgov/scripts/unpublish_closed_jobs.py
virginiacc: I added some specific exception logging in https://github.com/cfpb/cfgov-refresh/pull/3787/commits/8a577f1f4c5d79295c6d8e8f9e9eb2b83b153e94.

chosak: @virginiacc two minor suggestions:

1. You can use `logger.exception` instead of `logger.info` to automatically include the exception in the stack trace, without needing to add it to the message.
2. [PEP8 recommends](https://www.python.org/dev/peps/pep-0008/#programming-recommendations) avoiding bare `except:` clauses (instead use `except Exception:`), because they make it harder to exit with Control-C:
    
    > A bare except: clause will catch SystemExit and KeyboardInterrupt exceptions, making it harder to interrupt a program with Control-C, and can disguise other problems. If you want to catch all exceptions that signal program errors, use except `Exception:` (bare except is equivalent to `except BaseException:`)

virginiacc: @chosak I updated to use `logger.exception` and a single `except Exception:` clause, since the exception data is automatically logged (unless it's preferable to use individual handlers?)

Thanks for the suggestions, @willbarton & @chosak!

cfgov/scripts/unpublish_closed_jobs.py
```diff
@@ -0,0 +1,74 @@
...
+
+def job_page_closed(link):
+    try:
+        webpage = urlopen(link)
+        soup = BeautifulSoup(webpage, 'html.parser')
+        closed_text = soup.find('div', attrs={'class': 'usajobs-joa-closed'})
+        if closed_text:
+            return 'This job announcement has closed' in closed_text.contents
+    except Exception:
+        logger.exception('Check of USAJobs page "{}" failed'.format(link))
```
chosak: @virginiacc one other thought: should this exception (and the one below) cause the job to fail? Currently if these checks fail there'll be a logging statement, but the command (and, presumably, Jenkins job) will succeed unnoticed.

virginiacc: @chosak That's a good question. I think it's possible that one check could fail but others succeed if there are multiple live job postings. Maybe I could restructure to fail the job if none of the checks succeed?

Or could we notify Human Capital of the results of this job? I've been thinking of this as an optimization rather than complete replacement of the current process -- we still need to double-check that jobs have closed on our site in case of API issues -- and reporting the outcome of this script could be the way to do that.

chosak: @virginiacc if the USAJobs site tends to be down intermittently, and we expect this job to fail from time to time, then it might be confusing/annoying to notify HC every time that happens. If this job is running frequently (hourly?) then it might not matter if any one or several jobs fails.

Following standard Python/command line practice a user might expect that when some error happens the command will exit with a non-zero return code. It might be a good idea to keep checking jobs even if a single one fails, although I'd guess most likely if one fails they all would.

The way I might try to set it up is to have the job fail if any checks fail, and then setup the Jenkins job that calls this to notify the appropriate people after something like two failures in a row. That way if there's some one-off hiccup that isn't a big deal, people aren't bothered with nuisance emails, but if there's a continuous failure, they'll know about it and be able to respond.

virginiacc: @chosak The USAJobs API/site seems stable, but I encountered at least one instance in testing where my requests repeatedly timed out for a few minutes, so I wanted to account for that possibility. I updated in https://github.com/cfpb/cfgov-refresh/pull/3787/commits/716a4659ca406a328e589358c9277d773f87ae10 to fail the job if an API or page check results in an exception, and also to target the checks based on applicant type.

Jobs always close at midnight on USAJobs, so I'd been thinking this could run once a night -- but maybe it should re-run at an interval if it fails the first time, and only send notifications if it continues to fail?

chosak: @virginiacc that sounds good.

Given our current Jenkins conventions I believe it's straightforward to send emails on the second failure. I don't know of a way to alter the interval of running based on success/failure of previous jobs, though. What if you had this job run twice nightly, say, midnight and 1AM? In success cases, the 1AM run would have nothing to do (effectively be a no-op), but in failure cases, it'd be both a re-try and allow for sending of that error email.

virginiacc: @chosak That sounds like it would work perfectly!

Issue #3737, opened 2018-01-16
Update to gulp 4


anselmbradford:
Short description explaining the high-level reason for the new issue.

## Current behavior

- Update to latest gulp, like happened in https://github.com/cfpb/capital-framework/pull/737

## Steps to close

1. Move `build.js` and `default.js` to `gulpfile.js`
2. Convert arrays in tasks to `gulp.serial` or `gulp.parallel` or a combination of both.
3. See branch with work [here](https://github.com/cfpb/cfgov-refresh/tree/gulp_4). gulp-sourcemaps has an issue.

PR #3790, opened 2018-02-12
Update to gulp 4

anselmbradford:
Fixes https://github.com/cfpb/cfgov-refresh/issues/3737

**Re-creates https://github.com/cfpb/cfgov-refresh/pull/3746, which needs Jenkins to support gulp 4 before merge. (Reverted in https://github.com/cfpb/cfgov-refresh/pull/3789)**

## Changes

- Updates to gulp 4.
- Reorders default gulp task so it goes build, lint, test, instead of lint, test, and build. For some reason it would throw an error in the original order.
- Replace task array parameter with native `gulp.series` and `gulp.parallel`.
- Minor re-ordering of `scripts.js` and `styles.js` tasks.
- Updates `gulp-notify`, `gulp-header`, and `gulp-less` to latest.

## Testing

1. `npm install`
2. `gulp`, `gulp build`, `gulp copy`, `gulp images`, `gulp test`, `gulp lint`, and other tasks should pass.


anselmbradford:
@chosak Would you be able to briefly outline what needs to happen in Jenkins for this PR to proceed? Maybe we can get it through next sprint.

chosak:
@anselmbradford because those changes would need to be made on to internal deployment code/systems, this probably isn't the right forum for detailed discussion. Briefly, though, I guess that the new version of Gulp needs to be made available to only certain Jenkins jobs while not modifying the version that other Node-using jobs depend on now.

anselmbradford:
@chosak cool, moved to [GHE]/CFGOV/platform/issues/2467

anselmbradford:
@chosak word is we can try this in Jenkins now. Ready for review again.

anselmbradford:
@chosak Try now please!

anselmbradford:
@chosak `gulp` should just build without errors.

PR #3793, opened 2018-02-13
Allow linting and testing with Python 3

willbarton:
This PR allows us to lint against Python 3 and run tests against Python 3. The tests will fail, but they're runnable. 

The intention here is to allow us to run new and individual tests against both Python 2 and 3 until we get around to making the entire codebase Python 3-compatible. You can now run `tox -e fast-py3 some.package.tests.test_my_great_stuff` or (if you really want to) `tox -e py36 some.package.tests.test_my_great_stuff`.

## Additions

- tox environments `py36`, `lint-py3`, and `fast-py3`
- `py2` and `py3` sections in the `tox.ini` to define common reusable `basepython` and `envdir` values for each respective Python version

## Changes

- Make the codebase Python 3-compatible enough for the tests to run successfully by using `six` for incompatible imports

## Testing

- `tox -e lint-py3` to link with python3
- `tox -e fast-py3` to run tests with python3. There will be a series of individual test failures and errors, but the tests will run (the entire tox run won't fail on syntax or imports).

```
FAILED (failures=9, errors=236)
```

# Notes

As we move ahead with Python 3 support, the [`six` documentation is very useful](http://pythonhosted.org/six/) for figuring out how to support both python 2 and 3. While there are friendlier libraries (`future`) out there, `six` is already one of our dependencies, so it allows us to get away with not adding yet another library.

# Todo

- Have Travis run both `lint-py3` and `fast-py3` but not fail the build when they fail, so we're aware of any potential incompatibilities with Python 3 with new code or changes.

## Checklist


willbarton:
I should note, we need to publish a Python 3 dist or sdist of govdelivery to pypi before this will work. I modified requirements/base.txt locally to fetch govdelivery directly from github.

cfgov/agreements/templates/agreements/issuer_select.html
```diff
@@ -1,5 +1,5 @@
 <select data-placeholder="Choose an issuer" class="chzn-select" tabindex="2" id="issuer_select">
 {% for i in issuers %}
-    <option value="{{ i.slug }}" {% ifequal selected i.slug %}selected{% endifequal %}>{{ i.name }}</option>
+    <option value="{{ i.slug }}" {% ifequal selected i.slug %}selected="selected"{% endifequal %}>{{ i.name }}</option>
```
chosak: Why is this change required for this PR?

Scotchester: This change goes against our HTML style guide.

willbarton: Because BeautifulSoup4 does not automatically populate "selected" as the value of the selected attribute, and BeautifulSoup3 is not Python3-compatible.

chosak: Could that be addressed by modifying [this test](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/agreements/tests/test_legacy.py#L233)? It'd be nice to limit this PR to just enable Python3 testing vs. actually making tests work.

What exactly is the issue - is it that the tox environment setup fails if you try to install a requirement that isn't supported by Python 3?

Scotchester: Yeah, modifying the test to simply check for the presence of the `selected` attribute would be preferable.

willbarton: I tried. Checking for `selected=""` does not match the expected `option`.

higs4281: this works for me:

```python
def get_selected_value(select_element):
    for item in select_element.find_all('option'):
        if item.get('selected', None) == '':
            return item.text
```

willbarton: @higs4281 Yeah, [this was the only alternative I could make work](https://github.com/cfpb/cfgov-refresh/pull/3793/commits/94d00b11e2fc9f32cea68634d92bf191e3dd0834#diff-8f93154e53c7a83b3dc584c02d90e339R233).

@chosak @Scotchester I just don't think a list comprehension to make bs4 match bs3's assumptions is the right way to go.

It's a minor thing — I think it's uglier code, but it's far more important to me to be able to run tests with Python 3. 😁

willbarton:
@higs4281 yeah, I had fixed those. Rebase obliterated the changes.

cfgov/agreements/templates/agreements/issuer_select.html
Scotchester: DId you try `find_all('option', selected=True)`? See https://stackoverflow.com/a/39055066/279566

willbarton: @Scotchester I thought I had, but apparently I hadn't. Cheers! That works!

willbarton:
Alrighty, there's now an sdist of govdelivery==1.1, which should suffice for running python3 tests.

PR #3794, opened 2018-02-13
Update display names for categories

richaagarwal:
See GHE/CFGOV/platform/issues/607#issuecomment-104555 for context.

To review, you can pull down the branch and browse to the blog or newsroom page, or other pages with filterable list categories, and confirm that the new categories are displaying. A screenshot of the blog page:
screen shot 2018-02-13 at 1 29 05 pm
Note that now that "Data, research, and reports" is longer, it wraps to the second line.  A separate GHE issue (597) will take care of the indentation on lines like that.


cfgov/v1/util/ref.py
```diff
@@ -7,21 +7,21 @@
...
 ]
 
 related_posts_categories = [
     ('Blog', (
         ('At the CFPB', 'At the CFPB'),
-        ('Policy &amp; Compliance', 'Policy & Compliance'),
-        ('Data, Research &amp; Reports', 'Data, research & reports'),
+        ('Policy &amp; Compliance', 'Policy and compliance'),
+        ('Data, Research &amp; Reports', 'Data, research, and reports'),
         ('Info for Consumers', 'Info for consumers'),
```
Scotchester: The first string in these tuples (the three lines above) should match the second string, right?

cfgov/v1/util/ref.py
```diff
@@ -7,21 +7,21 @@
...
+    ('contract', 'Op-ed'),
     ('double-quote', 'Testimony'),
 ]
 
 related_posts_categories = [
     ('Blog', (
         ('At the CFPB', 'At the CFPB'),
-        ('Policy &amp; Compliance', 'Policy & Compliance'),
-        ('Data, Research &amp; Reports', 'Data, research & reports'),
+        ('Policy &amp; Compliance', 'Policy and compliance'),
```
richaagarwal: Note that we don't change the left side of the tuple here, because that's what's stored in the Wagtail block `specific_categories` for the page, and to update it we must also publish the page, which would cause other ripple effects, e.g. `date_published` on a bunch of pages would get updated and cause how we display recent posts to be out of order.

cfgov/v1/util/ref.py
```diff
@@ -33,7 +33,7 @@
     ('amicus-brief', 'Amicus Brief'),
     ('blog', 'Blog'),
     ('enforcement', 'Enforcement Action'),
-    ('final-rule', 'Final Rule'),
+    ('final-rule', 'Final rule'),
```
Scotchester: Maybe undo this one, since it's not public-facing and it would then be inconsistent with the rest of this list?

cfgov/v1/util/ref.py
```diff
@@ -67,8 +67,8 @@
     )),
     ('Blog', (
         ('at-the-cfpb', 'At the CFPB'),
-        ('policy_compliance', 'Policy & Compliance'),
-        ('data-research-reports', 'Data, research & reports'),
+        ('policy_compliance', 'Policy and compliance'),
```
Scotchester: It was this way before,  but would it be a pain to fix this so it's not the only item in this list with a slug that uses underscores instead of dashes?

cfgov/v1/util/ref.py
```diff
@@ -77,8 +77,8 @@
         ('stipulation-and-consent-order-2', 'Stipulation and consent order'),
     )),
     ('Final Rule', (
```
Scotchester: Should this one get updated, too? The top-level items in this `categories` object are wildly inconsistent.

cfgov/v1/util/ref.py
Scotchester: D'oh, sorry. Didn't see your comment while I was reviewing :)

cfgov/v1/util/ref.py
Scotchester: That is... unfortunate.

cfgov/v1/util/ref.py
richaagarwal: Yes, it would run into the same issue as noted above.

cfgov/v1/util/ref.py
richaagarwal: Hmm I see what you're saying.  It's used when a content editor selects a page type, and you're right that they didn't request changes on the other ones, however given the request I'd rather just change it everywhere I can change it.

richaagarwal: This is how it looks now with the change. As you can see, there are already inconsistencies in terms of casing in the list.  I think I'm OK with leaving this as-is.
screen shot 2018-02-13 at 2 26 28 pm

cfgov/v1/util/ref.py
richaagarwal: Yes. This code is very, very sad but refactoring it is not easy. See also GHE/CFGOV/platform/issues/825

cfgov/v1/util/ref.py
richaagarwal: Actually, regardless of whether they're used for display purposes, it'd need to be in sync with the other ones given that it's used as a look-up table, so making the change here as well. Thanks for the catch.

cfgov/v1/util/ref.py
nataliafitzgerald: @richaagarwal - I think more of those can be changed to sentence case. I am checking with @sarahsimpson09 now.

richaagarwal: @nataliafitzgerald It'd be easier if that came in through a new request since updating this PR requires updating a migration, especially since that specific list is only internally-facing.

richaagarwal: (That said, I'm not really sure why Final rule was included in the first place... so maybe that's what's causing the confusion)

nataliafitzgerald: Yes, if we want to go ahead and merge these changes and then I can update the original issue with any other changes that's fine. "Final rule" is externally facing so I would like to keep that in the set we're correcting. https://www.consumerfinance.gov/policy-compliance/rulemaking/final-rules/

PR #3795, opened 2018-02-14
remove django-reversion dependency

rosskarchner:
I haven't been able to find any remaining code that uses [django-reversion](https://django-reversion.readthedocs.io/en/stable/), and the last Revision in the database has a 2014 timestamp.

~Potentially controversial: I included a data migration that removes Reversion's entries form the contenttypes table (and thus, permissions). It seems like Django's automatic detection of stale content types does *not* extend to apps that have been removed.~

## Removals

- references to django-reversion

## Notes

-

## Todos

-

## Checklist


cfgov/core/migrations/0001_initial.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+from django.contrib.contenttypes.models import ContentType
+
+
+def remove_reversion_ct(apps,schema_editor):
+    ContentType.objects.filter(app_label='reversion').delete()
```
chosak: Can you refer to a model via import like this, in a migration? I think you might need to do something like

```py
ContentType = apps.get_model('contenttypes', 'ContentType')
```

to void errors.

rosskarchner: hmm, it worked for me-- but this is probably cleaner.

rosskarchner:
@chosak I've removed the migration

cfgov/core/migrations/0001_initial.py
chosak: Right, it'll work, but generally best practice in Django migrations is not to import models like that. The problem would be, if, say, in a future Django version the `ContentType` model has some different fields. This migration might run before the one modifying `ContentType`, which means that the state of the model via import from code would differ from the one in the DB. Using `apps.get_model` ensures that you always get the right model version at the time of the migration.

rosskarchner:
I thought there was no such thing as an empty directory in git?

rosskarchner:
Ah, nm, I see what you mean

rosskarchner:
Deleting it seems to be harmless.

PR #3800, opened 2018-02-15
Fail Python unit tests if anything writes to stdout

willbarton:
This PR is a proposal for capturing stdout while unit tests run and then failing the tests if anything writes to it. It does this by capturing all stdout during the test run and asserting that it is empty at the end. It uses `redirect_stdout` from Python 3's `contextlib` and a function that approximates it in Python 2.

One disadvantage to this approach is that you don't know where the text came from necessarily. The test run looks like this:

```
Creating test database for alias 'default'...
applying migration wagtail.wagtailcore.migrations.0002_initial_data
applying migration wagtail.wagtailcore.migrations.0025_collection_initial_data
.
Ran 1 test in 0.113s

OK
  File "manage.py", line 11, in <module>
    execute_from_command_line(sys.argv)

    … (typical traceback) …

AssertionError: there is content in stdout: this is a test

ERROR: InvocationError: '/Users/bartonw/repos/platform/cfgov-refresh/.tox/py27/bin/coverage run --source=. manage.py test cfgov.tests'
_____________________________________________________________________________________________________________________ summary _____________________________________________________________________________________________________________________
ERROR:   fast: commands failed
```

But it's a proposal. I'd love to hear thoughts, and maybe other ideas.

## Checklist

* [x] Any *change* in functionality is tested

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,9 @@ jobs:
   include:
     - stage: run tests
       env: RUNTEST=frontend
-    - env: RUNTEST=backend
+    - env: 
+      - RUNTEST=backend
+      - TEST_RUNNER=cfgov.test.StdoutCapturingTestRunner
       deploy:
         provider: pages
         local_dir: site
--- a/cfgov/cfgov/settings/test.py
+++ b/cfgov/cfgov/settings/test.py
@@ -20,7 +20,7 @@ DATABASES = {'default': TEST_DATABASE}
 
 EMAIL_BACKEND = 'django.core.mail.backends.locmem.EmailBackend'
 
-TEST_RUNNER = 'cfgov.test.TestDataTestRunner'
+TEST_RUNNER = os.environ.get('TEST_RUNNER', 'cfgov.test.TestDataTestRunner')
 
 INSTALLED_APPS += (
     'wagtail.contrib.settings',
--- a/cfgov/cfgov/test.py
+++ b/cfgov/cfgov/test.py
@@ -69,18 +69,6 @@ class PlaceholderJSMixin(object):
 
 
 class TestDataTestRunner(PlaceholderJSMixin, DiscoverRunner):
-    def run_suite(self, suite, **kwargs):
-        stdout = StringIO()
-        with redirect_stdout(stdout):
-            return_value = super(TestDataTestRunner, self).run_suite(
-                suite,
-                **kwargs
-            )
-        assert stdout.getvalue() == '', (
-            'there is content in stdout: {}'.format(stdout.getvalue())
-        )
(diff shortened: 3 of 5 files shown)


willbarton:
The tests will presently fail because of import order. See #3799.

richaagarwal:
Do we have any information on if this slows down the tests or not?  I'm OK with it if it doesn't slow down Travis. But if it does, I don't know if it's worth it, particularly if it doesn't tell us which test has the print statement(s).

willbarton:
@chosak that's a good point. This could be in a subclass of TestDataTestRunner that we only use when running the tests on Travis.

willbarton:
@richaagarwal quick and dirty timing locally (without recreating the venv):

```shell
%> git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
%> time tox -e fast
…
tox -e fast  59.86s user 1.66s system 96% cpu 1:04.01 total
%> git checkout test-print 
Switched to branch 'test-print'
Your branch is up-to-date with 'origin/test-print'.
%> time tox -e fast       
…
tox -e fast  60.69s user 1.83s system 94% cpu 1:06.04 total
```

Seems to be within acceptable margins.

willbarton:
@chosak I broke our an `StdoutCapturingTestRunner` and added a `TEST_RUNNER` env var that we can use to select it. What do you think?

chosak:
@willbarton works as advertised 👍 

How would you feel about making the error message a little more explanatory? "there is content in stdout" might not be super clear to someone new to the repository, as opposed to something like "Python unit tests should avoid writing to stdout".

Also, what's happening with the failing Travis build here? Shouldn't failing Python3 tests not fail the entire build?

willbarton:
@chosak :+1:

I think the failure is my tweaks to the travis.yml file.

willbarton:
> How would you feel about making the error message a little more explanatory? "there is content in stdout" might not be super clear to someone new to the repository, as opposed to something like "Python unit tests should avoid writing to stdout".

Current messaging: `AssertionError: unit tests should avoid writing to stdout`

willbarton:
@higs4281 Just in test running. If you're writing unittests, and they're testing code that writes to stdout, you should probably be capturing stdout to check the output.

For right now, this is only enabled when `TEST_RUNNER=cfgov.test.StdoutCapturingTestRunner`, and this PR enables that for Travis. We'd need to set that variable in other places we'd want to enforce it.

willbarton:
On the documenting in cfpb/development, we should definitely do that. I think I'd feel better seeing how it plays out here first before dictating that "this is what we do!" I'm still thinking of this as an experiment.

PR #3802, opened 2018-02-15
Create app images task

anselmbradford:
Apps coming in from the outside of cfgov-refresh sometimes come with their own images, CSS, and JS. To manage these assets it may be helpful to deliver them in their own namespaced directory from `unprocessed/apps/`. This PR creates an `images:apps` gulp task to handle moving images from that directory to `static_built`. Additionally, it removes the images entry from the gulp config file, since we should only need to use the config file for values that are shared across tasks, which is not the case with images. Otherwise, it would be clearer to directly include the paths in the tasks themselves.

## Additions

- `images:general` and `images:apps` tasks in gulp.

## Removals

- images config in gulp config file.

## Changes

- Moves images for /wellbeing/ app to unprocessed/apps/wellbeing/img
- 

## Testing

1. Run `gulp images` and check that there is a `static_built/apps/wellbeing/img/` directory.
2. http://localhost:8000/consumer-tools/financial-well-being/ should have images.

## Screenshots

screen shot 2018-02-15 at 1 40 53 pm

## Notes

- I included an owning-a-home image, but it won't be used until the rate checker page is moved over to static files.

## TODO

- If this seems reasonable I'll be creating an analogous style and script task that pull from `unprocesses/apps/**/css/*` and `unprocessed/apps/**/js/*`


sebworks:
Thanks @anselmbradford

anselmbradford:
Closed in favor of https://github.com/cfpb/cfgov-refresh/pull/3806

PR #3805, opened 2018-02-16
Ask CFPB: stop UnicodeDecodeError in feedback

higs4281:
This addresses GHE issue 2464.
The [feedback handler](https://github.com/cfpb/cfgov-refresh/blob/24fd67d40f351687ca488abf07201506f0817edc/cfgov/v1/handlers/blocks/feedback.py#L64-L65) was trying too hard to capture unicode characters from the referrer URL.  
The referrer string is already quote-encoded correctly, and can be safely saved as-is.

The [jinja2 guidance](http://jinja.pocoo.org/docs/2.10/api/#unicode) on how it handles unicode was helpful.

## Testing
I couldn't get the error to occur locally or on our servers, and it only happens sporadically,   
so there may be a browser-based condition involved. But by testing 
how the feedback was getting captured locally, it seems clear that the 
original url-quoted referrer string is ascii-safe and should make all parties happy, 
and the feedback entry ends up with a valid referrer URL.


chosak:
@higs4281 I salute the effort to fix this; thank you for picking it up.

It makes sense that this bug might fix issues with the feedback URLs that get stored in the database. But the error in GHE#2464 is a 500 that happens when the feedback block is rendered (I think [this line](https://github.com/cfpb/cfgov-refresh/pull/3805)), which causes errors like

```py
{{- form_block.render(block.value, settings, form_id) -}}
File "/srv/cfgov/versions/20180209143827/current/cfgov/jinja2/v1/_includes/blocks/feedback.html", line 33, in template

UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 81: ordinal not in range(128)
```

In this particular case the request `HTTP_REFERER` is set to "https://www.consumerfinance.gov/es/obtener-respuestas/buscar-por-etiqueta/rescisi\xc3\xb3n/"

So while this PR gets a :+1: for me for being a seemingly reasonable thing to do, I'm not sure it addresses that particular bug. I wonder if for that we need to do something like `force_text` on the referer before we put it in the template.

Unless I'm missing something, and the value of the feedback does make it in the template there somehow?

higs4281:
@chosak 

Ah -- I was thinking the error was occurring upon form submission, not the original form rendering.

So the mystery remains: why would a bytestring sequence like '\xc3\xb3n/' *sometimes* show up in the referrer string?

The handler code gets called to render the form, but it shouldn't have been introducing a bytestring unless it was handling a submission.

higs4281:
@chosak @richaagarwal 
I adjusted the handler to catch the bad referrers before form submission and added a few tests.

The bad encodings seem to come from two sources: outside sources, such as google search, which we can't do much about, and some pathway in cfgov that sends along unquoted unicode.

Unfortunately, the strings that generate errors can't just be quoted as-is, because the handler gets them as a unicode string, but the encodings are actually utf-8. So to be viable URLs, they need to be fixed before they can be properly encoded, quoted and stored.

For internal referrals, which seem to be come sporadically from Spanish tag-search pages (I can't reproduce), we can fix the known mis-encodings that have cropped up so far, and let the others go unfixed but quoted to avoid 500 errors.

I'm comfortable capturing some inexact referral strings for these reasons:
- Almost-exact referral strings are better than 500 errors and no referrals.
- The referrals from outside sources, especially google search, don't need to be exact.
- Internal referrals are only occasionally wrong, and we can fix all the ones we've seen.
- Internal referrals that are still wrong will only be off by a letter or two, and the actual referral page will be easy to figure out.
- No one is actually looking at these referrals at present. We were asked to save them, but it's not clear they will be used anytime soon.

cfgov/v1/handlers/blocks/feedback.py
```diff
@@ -1,6 +1,4 @@
 # -*- coding: utf-8 -*-
-from __future__ import unicode_literals
```
chosak: Will removing this just postpone the need to rewrite this file once we move to Python 3? Is it possible to use bytestrings `b'something'` when we need to refer to non-unicode strings?

cfgov/v1/handlers/blocks/feedback.py
```diff
@@ -43,14 +41,29 @@ def __init__(self,
...
+        return referrer
 
+    def process(self, is_submitted):
         form_cls = FEEDBACK_TYPES[get_feedback_type(self.block_value)]
 
         if is_submitted:
             form = form_cls(self.request.POST)
             return self.get_response(form)
 
+        self.block_value.update({'referrer': self.sanitize_referrer()})
```
chosak: I worry a bit about overloading the block value like this, as it depends on the inner workings of the Wagtail block rendering. Am I reading this right that you're mocking out a `feedback` attribute on the block [even though one doesn't exist](https://github.com/cfpb/cfgov-refresh/blob/ca9053d66e65ab0f66490fab6f08a4b7354354f8/cfgov/v1/blocks.py#L75)?

Is there some other way of passing this to the context, e.g. add it to the 

```py
return {
    'form': form_cls(),
    'referrer': self.sanitize_referrer(),
}
```

and then in your template refer to `settings.referrer`?

cfgov/v1/handlers/blocks/feedback.py
higs4281: Great point. I need to keep my py3 hat on.

cfgov/v1/handlers/blocks/feedback.py
higs4281: Thank you for this suggestion.

PR #3807, opened 2018-02-21
Change on-demand font paths

anselmbradford:
## Changes

- Changes cf-font paths to be absolute in on-demand header and footer CSS.

## Testing

1. `gulp styles:ondemand` and `gulp scripts:ondemand:header` and `gulp scripts:ondemand:footer` should pass.


anselmbradford:
Enable generation of change in https://github.com/cfpb/cfgov-sheer-templates/pull/29. We should merge that and bump the version in owning-a-home and test that it works here before merging this PR.

sebworks:
👍, of course

PR #3773, opened 2018-02-06
Moving owning-a-home/closure-disclosure pages into cfgov-refresh

sebworks:
## Changes

- Modified CSS, HTML, and images to move `owning-a-home/closing-disclosure/` into cfgov-refresh. 

## Testing

1. Run `gulp build`.
2. Visit `https://www.consumerfinance.gov/owning-a-home/closing-disclosure/` and 
`http://localhost:8000/owning-a-home/closing-disclosure/`.
3. Click the expandables on and ensure they expand / contract as expected.
4. Compare localhost to production.
5. Mouse over the highlighted areas on the image and ensure the related expendables receive focus.
6. Click the 'Get definitions` tab and ensure the expandable terms switch, as expected.
7. Click the pagination buttons and ensure the pages fadeIn/Out, as expected.
8. Resize the browsers and ensure the form explainer images resize properly.
9. Scroll the pages and ensure the form explainer images doesn't overlap the ratings well.

## Notes

- The `focus-target.js` code runs on all pages, and adds an event listener to the all anchor links. This causes the page to jump unexpectedly in some situations via Chrome.  We should modify the modules to require the code instead of including it in common.js.

## Todos

- Write unit tests.

## Checklist

* [ ] PR has an informative and human-readable title
* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [development playbook](https://github.com/cfpb/development)
* [ ] Passes all existing automated tests
* [ ] Any *change* in functionality is tested
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated
* [ ] Reviewers requested with the [Reviewer tool](https://help.github.com/articles/about-pull-request-reviews/) :arrow_right:

PR #3817, opened 2018-02-22
Create JavaScript loader block in base template

anselmbradford:
Apps (such as owning-a-home) will likely have their own template that overrides the base template's `javascript` block (see [example](https://github.com/cfpb/cfgov-refresh/pull/3773/files#diff-3d9dc87803dad1af3ead8842353b8a87R20)). The problem with this is if we want to use the JS loader that's in the base template it will need to be re-created in each child template. This PR creates a `javascript_loader` block that adds a javascript loader function to the page before the `javascript` block, so that child templates can choose to only override the `javascript` block and utilize the JS loader in the parent template. 

## Additions

- Creates `javascript_loader` block that contains a JS loader function.

## Testing

1. Blog page should still show common.js, email-signup.js, and filterable-list-controls.js in the JS network tab in dev tools.


sebworks:
The only issue is that you always have to call super because the common.js stuff is in the block.

sebworks:
Also, if you remove the function then the s variable has global scope.

anselmbradford:
@sebworks Okay, I re-added the function. It's in use in https://github.com/cfpb/cfgov-refresh/pull/3818

sebworks:
Ship it

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -256,7 +264,7 @@
   if ( document.body.parentElement.className.indexOf( 'no-js' ) === -1 ) {
     !function(){
     {# Include site-wide JavaScript. #}
-    var a = [ '{{ static('js/routes/common.js') }}' ];
+    var s = [ '{{ static('js/routes/common.js') }}' ];
```
Scotchester: Minor: should be indented one more level.

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -295,11 +303,11 @@
...
     {% set js_source = page_js() | trim %}
     {% if js_source | length > 0 %}
-      a.push( '{{ static('js/routes/' + request.path[1:] + 'index.js') }}' );
+      s.push( '{{ static('js/routes/' + request.path[1:] + 'index.js') }}' );
     {% endif %}
 
-    {# Minified dynamic JavaScript loader that injects a script tag in the head of the page #}
-    for(var b,c,d=[],f=document.scripts[0];b=a.shift();)c=document.createElement("script"),"async"in c?(c.async=!1,c.src=b,document.head.appendChild(c)):(d.push(c),c.onreadystatechange=function(){for(var e;d[0]&&"loaded"==d[0].readyState;)e=d.shift(),e.onreadystatechange=null,f.parentNode.insertBefore(e,f)},c.src=b);
+    {# Pass scripts to JavaScript loader. #}
+    jsl(s);
```
Scotchester: Minor: should be indented one more level.

cfgov/jinja2/v1/_layouts/base.html
anselmbradford: Fixed

cfgov/jinja2/v1/_layouts/base.html
anselmbradford: Fixed

anselmbradford:
Everyone blame @Scotchester 's shoddy review if this doesn't work 😜

PR #3822, opened 2018-02-23
Fix for using import statements with Mocha / Instanbul

sebworks:
Fix for using import statements with Mocha / Instanbul

## Changes

- Modified test script to fix issue with requiring scripts that contain import statements. 

## Testing

1. Run `gulp test:unit`.

## Checklist

### DIFF
--- a/cfgov/unprocessed/js/modules/BreakpointHandler.js
+++ b/cfgov/unprocessed/js/modules/BreakpointHandler.js
@@ -1,5 +1,5 @@
-const getBreakpointState = require( './util/breakpoint-state' ).get;
-const _breakpointsConfig = require( '../config/breakpoints-config' );
+import  { get as getBreakpointState } from  './util/breakpoint-state';
+import  _breakpointsConfig from '../config/breakpoints-config';
 
 /* Used for checking browser capabilities.
    TODO: Check what browsers this is necessary for and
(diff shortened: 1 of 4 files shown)


anselmbradford:
Hm I dunno, it doesn't seem worth an extra package/copy of babel and the processing just so we can use `import` syntax in test cases. Is there something wrong with the `require` statements as they are that necessitates this?

sebworks:
>> m I dunno, it doesn't seem worth an extra package/copy of babel and the processing just so we can use import syntax in test cases. Is there something wrong with the require statements as they are that necessitates this?

Sorry, so the purpose of this PR is so we can use the import / export syntax is regular modules. If you we don't use babel then we can't use import / export syntax in modules with have tests.

anselmbradford:
Well that's a shame. Would it make sense to share the babel config with webpack. Like export https://github.com/cfpb/cfgov-refresh/blob/master/config/webpack-config.js#L16 and use it in the test.js?

sebworks:
>> Would it make sense to share the babel config with webpack? Like export https://github.com/cfpb/cfgov-refresh/blob/master/config/webpack-config.js#L16 and use it in test.js?

Because we are using gulp and not webpack, then it wouldn't make sense.

anselmbradford:
> Because we are using gulp and not webpack, then it wouldn't make sense.

Why does that matter? They're identical?
https://github.com/cfpb/cfgov-refresh/blob/master/config/webpack-config.js#L22-L27

sebworks:
>> Why does that matter? They're identical?

ok, I thought you were referring to line 16. If we use it in another place then we can swap it to using a config.

anselmbradford:
No worries, I was just thinking if the test babel config was out of sync of the production babel config it could potentially falsely pass, but I approved because I'm not sure it's really worth the extra infrastructure considering we are already sharing the browser list.

PR #3826, opened 2018-02-27
Move rate explorer scripts to subdirectories

anselmbradford:
To make migration easier I want to break apart the explore-rates scripts into submodules, which means having them in a subdirectory makes sense. This moves `/owning-a-home/js/explore-rates.js` to `/owning-a-home/js/explore-rates/index.js`

## Changes

- Moves `/owning-a-home/js/explore-rates.js` to `/owning-a-home/js/explore-rates/index.js`
- Moves rate explorer templates from `/owning-a-home/templates/` to `/owning-a-home/templates/explore-rates`.
- Turns highchart theme into a module that exports a `applyThemeTo` function that takes a highcharts instance. This is to fix difficulty with passing the highcharts reference around after bundling.
- Converts `if ( $( '.rate-checker' ).length < 1 ) {` to `if ( document.querySelectorAll( '.rate-checker' ).length === 0 ) {`

## Testing

1. `gulp build` should pass.


cfgov/unprocessed/apps/owning-a-home/js/explore-rates/index.js
```diff
@@ -1,39 +1,65 @@
...
-var Highcharts = require( '../node_modules/highcharts' );
-require( '../node_modules/highcharts/modules/exporting' )( Highcharts );
 
-var formatUSD = require( '../node_modules/format-usd' );
-var unFormatUSD = require( '../node_modules/unformat-usd' );
+// Load and style Highcharts library. https://www.highcharts.com/docs.
+var Highcharts = require( '../../node_modules/highcharts' );
+require( '../../node_modules/highcharts/modules/exporting' )( Highcharts );
+const highchartsTheme = require( './highcharts-theme' );
+highchartsTheme.applyThemeTo( Highcharts );
```
anselmbradford: This replaces line 19

PR #3827, opened 2018-02-27
Fix test paths for code coverage

anselmbradford:
Istanbul code coverage wasn't including /apps/ path in JS unit tests.

## Changes

- Moves unit tests to /js/ directory so that both /js/ and /apps/**/js directories are processed by Istanbul.
- Normalizes name and location of BASE_JS_PATH variable.

## Testing

1. Run `gulp test:unit`
2. Open `unit_test_coverage/lcov-report/index.html` and note that apps/ or js/ prefix spec paths.


Scotchester:
I'm a little bit confused by the restructuring changes here. The only tests within `tests/unit_tests` are JS tests (all of the Python tests live elsewhere, per Django conventions), so the addition of `js` directories seems superfluous to me. Am I missing something about how `test.src` is used?

anselmbradford:
>I'm a little bit confused by the restructuring changes here. The only tests within tests/unit_tests are JS tests (all of the Python tests live elsewhere, per Django conventions), so the addition of js directories seems superfluous to me. Am I missing something about how test.src is used?

In unprocessed we have /js but now have /apps, which may contain js, css, img, etc for specific-app usecases. So we can't assume all tested js is in /js, since some will appear under /apps/[app namespace/js

Scotchester:
> >I'm a little bit confused by the restructuring changes here. The only tests within tests/unit_tests are JS tests (all of the Python tests live elsewhere, per Django conventions), so the addition of js directories seems superfluous to me. Am I missing something about how test.src is used?

In unprocessed we have /js but now have /apps, which may contain js, css, img, etc for specific-app usecases. So we can't assume all tested js is in /js, since some will appear under /apps/[app namespace/js

But those aren't the paths you changed. You changed the locations of the test files, not the files they are testing.

anselmbradford:
> But those aren't the paths you changed. You changed the locations of the test files, not the files they are testing.

I see what you're asking… yeah I was making it so the path `/unit_tests/some/path/widget-spec.js` matches `/unprocessed/some/path/widget.js`

anselmbradford:
Note that without this also the `--specs` flag for unit tests won't follow the structure of the source. So `gulp test:unit --specs=path/for/specs/and/source/will/differ`

test/unit_tests/js/modules/transition/MoveTransition-spec.js
```diff
@@ -1,4 +1,4 @@
-const BASE_JS_PATH = '../../../../cfgov/unprocessed/js/';
+const BASE_JS_PATH = '../../../../../cfgov/unprocessed/js/';
```
sebworks: We should probably make this a global node variable that you can just use in the files.

anselmbradford: Yeah we could probably do something with https://nodejs.org/api/path.html that figures this out.

PR #3830, opened 2018-02-28
owning-a-home/explore-rates/ params tests

anselmbradford:
Moves code out of rate-explorer script for handling rate explorer parameters and DOM lookup interaction.

## Removals

- Removes chai from dom-value test.

## Changes

- Adds check for undefined value to increase code coverage to 100% for dom-value test.

## Testing

1. `gulp test:unit:scripts --specs=apps/owning-a-home/js/explore-rates/`

### DIFF
--- a/test/unit_tests/apps/owning-a-home/js/explore-rates/dom-values-spec.js
+++ b/test/unit_tests/apps/owning-a-home/js/explore-rates/dom-values-spec.js
@@ -4,7 +4,7 @@ const expect = chai.expect;
 
 const HTML_SNIPPET = `
   <input id="credit-score"
-          type="range" min="600" max="840" step="20" value="700">
+         type="range" min="600" max="840" step="20" value="700">
   <select id="arm-type">
       <option value="3-1">3/1</option>
       <option value="5-1">5/1</option>
--- a/test/unit_tests/apps/owning-a-home/js/explore-rates/dom-values-spec.js
+++ b/test/unit_tests/apps/owning-a-home/js/explore-rates/dom-values-spec.js
@@ -10,7 +10,6 @@ const HTML_SNIPPET = `
       <option value="10-1">10/1</option>
   </select>
   <input id="location" type="text" value="AL">
-  <input id="location" type="text" value="AL">
   <input id="test-price" type="text" value="$300,000">
   <input id="house-price" type="text" placeholder="200,000">
 `;
--- a/test/unit_tests/apps/owning-a-home/js/explore-rates/dom-values-spec.js
+++ b/test/unit_tests/apps/owning-a-home/js/explore-rates/dom-values-spec.js
@@ -1,6 +1,4 @@
 const BASE_JS_PATH = '../../../../../../cfgov/unprocessed/apps/owning-a-home/';
-const chai = require( 'chai' );
-const expect = chai.expect;
 
 const HTML_SNIPPET = `
   <input id="credit-score"
@@ -11,6 +9,8 @@ const HTML_SNIPPET = `
       <option value="7-1">7/1</option>
       <option value="10-1">10/1</option>
   </select>
+  <input id="location" type="text" value="AL">
+  <input id="location" type="text" value="AL">
   <input id="test-price" type="text" value="$300,000">
   <input id="house-price" type="text" placeholder="200,000">
 `;
@@ -23,9 +23,11 @@ describe( 'explore-rates/dom-values', () => {
   } );
 
   it( 'should be able to get a value', () => {
-    expect( domValues.getSelection( 'credit-score' ) ).to.equal( 700 );
-    expect( domValues.getSelection( 'arm-type' ) ).to.equal( '3-1' );
-    expect( domValues.getSelection( 'test-price' ) ).to.equal( 300000 );
-    expect( domValues.getSelection( 'house-price' ) ).to.equal( 200000 );
+    expect( domValues.getSelection( 'not-found-elm' ) ).toBeUndefined();
+    expect( domValues.getSelection( 'location' ) ).toEqual( 'AL' );
+    expect( domValues.getSelection( 'credit-score' ) ).toEqual( 700 );
+    expect( domValues.getSelection( 'arm-type' ) ).toEqual( '3-1' );
+    expect( domValues.getSelection( 'test-price' ) ).toEqual( 300000 );
+    expect( domValues.getSelection( 'house-price' ) ).toEqual( 200000 );
   } );
 } );
--- a/test/unit_tests/apps/owning-a-home/js/explore-rates/params-spec.js
+++ b/test/unit_tests/apps/owning-a-home/js/explore-rates/params-spec.js
@@ -4,7 +4,7 @@ const expect = chai.expect;
 
 const HTML_SNIPPET = `
   <input id="credit-score"
-          type="range" min="600" max="840" step="20" value="700">
+         type="range" min="600" max="840" step="20" value="700">
   <input id="down-payment" type="text" placeholder="20,000">
   <input id="house-price" type="text" placeholder="200,000">
   <input id="loan-amount" type="text" placeholder="200,000">


cfgov/unprocessed/apps/owning-a-home/js/explore-rates/dom-values.js
```diff
@@ -0,0 +1,38 @@
...
+ *                         Usually the HTML element's id attribute.
+ * @returns {Object} Hash of element id and its value(s).
+ */
+function getSelection( param ) {
+
+  const elm = document.querySelector( '#' + param );
+  let val;
+
+  if ( !elm ) {
+    return val;
```
jimmynotjim: Why return an `undefined` variable and not just `false`?

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/index.js
```diff
@@ -531,27 +465,27 @@ function checkForJumbo() {
...
+    loanType:   params.getVal( 'loan-type' ),
+    loanAmount: params.getVal( 'loan-amount' )
   } );
   dropdown( 'loan-type' ).enable( norms );
   dropdown( 'loan-type' ).hideHighlight();
   $( '#county-warning' ).addClass( 'u-hidden' );
 
   // if loan is not a jumbo, hide the loan type warning
-  if ( jQuery.inArray( params['loan-type'], jumbos ) < 0 ) {
+  if ( jQuery.inArray( params.getVal( 'loan-type' ), jumbos ) < 0 ) {
```
jimmynotjim: Any idea why it's `jQuery` here and `$` everywhere else?

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/params.js
```diff
@@ -0,0 +1,69 @@
+const domValues = require( './dom-values' );
+
+let UNDEFINED;
+
+// List all the parameters the user can change and set
+// their default values.
+// `verbotenKeys` are as follows:
+// 9 = tab
```
jimmynotjim: Should these be constants like we've done elsewhere (specifically thinking of the multi-select/typeahead thing for filtering options).

test/unit_tests/apps/owning-a-home/js/explore-rates/dom-values-spec.js
```diff
@@ -0,0 +1,43 @@
+const BASE_JS_PATH = '../../../../../../cfgov/unprocessed/apps/owning-a-home/';
```
jimmynotjim: Should this live in the base of `test/unit_tests/apps/owning-a-home` and be required in to each file that needs it?

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/dom-values.js
```diff
@@ -0,0 +1,38 @@
+const unFormatUSD = require( '../../node_modules/unformat-usd' );
+
+/**
+ * Get value(s) of an individual HTML element in the control panel.
+ * @param   {string} param Name of parameter to get.
+ *                         Usually the HTML element's id attribute.
+ * @returns {Object} Hash of element id and its value(s).
+ */
+function getSelection( param ) {
```
jimmynotjim: This feels like the wrong name, it's not just getting a selection, it's returning the value of that selection.

test/unit_tests/apps/owning-a-home/js/explore-rates/dom-values-spec.js
```diff
@@ -0,0 +1,43 @@
...
+  beforeEach( () => {
+    sandbox = sinon.sandbox.create();
+    document.body.innerHTML = HTML_SNIPPET;
+  } );
+
+  afterEach( () => {
+    sandbox.restore();
+  } );
+
+  it( 'should be able to get a value', () => {
```
jimmynotjim: This is definitely better than no tests, but is more like a browser test than a unit test. It also doesn't account for the condition where it exists early with `undefined`

anselmbradford:
Note that istanbul fails on this PR, apparently because istanbul is being sunset in favor of `nyc` and is having a problem with es6 features run through Babel. After discussing with Steven, I'm looking at Jest as an alternative.

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/dom-values.js
anselmbradford: This used to tie into the [params object](https://github.com/cfpb/owning-a-home/blob/master/src/static/js/modules/explore-rates.js#L73) and use https://api.jquery.com/jquery.extend/, which throws out undefined and null values. So I kept that functionality.

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/index.js
anselmbradford: Hm, not sure. That's the original setup. It seems like it's just aliased to `$` https://github.com/cfpb/owning-a-home/blob/master/src/static/js/modules/explore-rates.js#L4

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/dom-values.js
jimmynotjim: Ah gotcha, thanks for the explanation.

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/index.js
jimmynotjim: There's even another use of `$.inArray()` farther down the code base. Mind updating it to conform with the rest?

test/unit_tests/apps/owning-a-home/js/explore-rates/dom-values-spec.js
anselmbradford: We could probably do something with https://nodejs.org/api/path.html, but that should apply to all tests so best in its own PR.

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/dom-values.js
anselmbradford: This is the original name, but it could be updated, what do you think?

test/unit_tests/apps/owning-a-home/js/explore-rates/dom-values-spec.js
anselmbradford: Sorry, where what exists early?

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/index.js
anselmbradford: Removed `jQuery` variable in https://github.com/cfpb/cfgov-refresh/pull/3830/commits/69528e828534f09ee491cc38775bb3697a136043

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/params.js
anselmbradford: The lookup for these uses a numeric value and there's another group of keys that's a bigger group than this, so probably best to leave as is till we're certain this is stable and then refactor.

anselmbradford:
Moved just the JS without tests to https://github.com/cfpb/cfgov-refresh/pull/3843 so that we can isolate the test issue.

anselmbradford:
This is now just the tests, please re-pull

test/unit_tests/apps/owning-a-home/js/explore-rates/dom-values-spec.js
jimmynotjim: This one that was moved to the other PR. https://github.com/cfpb/cfgov-refresh/pull/3830#discussion-diff-171411382R14

jimmynotjim:
Tests are passing and gulp builds. Just address the question about the if condition

test/unit_tests/apps/owning-a-home/js/explore-rates/dom-values-spec.js
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/pull/3830/commits/7e1c421a579f3b77bd8bb593c60747a17fa53cc1

PR #3836, opened 2018-02-28
Remove unused templates

anselmbradford:
Removes two OAH handlebar templates that I can't find references to, even in the original OAH repo.

## Removals

- `loan-form.hbs` and `loan-add-button.hbs`


sebworks:
Yeah, this looks like it was for a loan-comparison feature, which was never went live.

PR #3834, opened 2018-02-28
Cleanup for apps-install script

anselmbradford:
Address @Scotchester comments in https://github.com/cfpb/cfgov-refresh/pull/3831

## Removals
- Removes erroneous `package-lock.json`.

## Changes

- Corrects incorrect code comments in apps-install npm script.
- Wrap npm standard error output in an if statement.

## Testing

1. See testing in https://github.com/cfpb/cfgov-refresh/pull/3831

PR #3842, opened 2018-03-01
Removes unneeded package-lock file

anselmbradford:
## Removals

- Removes unneeded package-lock file from apps directory.


Scotchester:
Aren't you already doing this in #3834?

anselmbradford:
> Aren't you already doing this in #3834?

Yeah, but those are in further review and this file is ending up in other branches so I thought I'd just spin it out into its own PR before it sneaks back in.

PR #3843, opened 2018-03-01
owning-a-home/explore-rates/ params

anselmbradford:
Moves code out of rate-explorer script for handling rate explorer parameters and DOM lookup interaction.

## Additions

- dom-values for pulling values out of rate explorer DOM.
- params modules for setting/getting/updating from DOM values for us in the rate explorer.

## Changes

- Updates params syntax to use a method from the params module.
- Creates params.update method not attached to internal params object.
- Removes jquery from params and dom-values code.

## Testing

2. `gulp build` should pass (run `cp ./cfgov/unprocessed/apps/owning-a-home/example-config.json ./cfgov/unprocessed/apps/owning-a-home/config.json` if needed).
3. With `OAH_EXPLORE_RATES` flag set to true, http://localhost:8000/owning-a-home/explore-rates/ should load without console errors.


cfgov/unprocessed/apps/owning-a-home/js/explore-rates/index.js
```diff
@@ -1,5 +1,4 @@
 var $ = require( '../../node_modules/jquery' );
-var jQuery = $;
```
sebworks: are you sure this isn't necessary?

anselmbradford: I converted the references to it to `$`, per https://github.com/cfpb/cfgov-refresh/pull/3830#pullrequestreview-100261276 .

PR #3844, opened 2018-03-02
Migrating to webpack 4

sebworks:
Migrating to Webpack 4.  You can read about some of the benefits here: https://medium.com/webpack/webpack-4-released-today-6cdb994702d4

## Changes

- Modified scripts to migrate code to Webpack 4. This PR also includes modifications to allow us to run in dev mode, by setting the NODE_ENV environment variable.  This PR allows each app to have it's own webpack-config.js file.

## Testing
1. Run `frontend.sh`.
2. Run `export NODE_ENV=development` at the command line.
3. Run `gulp scripts` and verify that soucemaps are enabled.
4. Browse the site and ensure all the expected pages are working. Check the console for errors. 

## Notes

- This isn't DRY but we shall get there.

## Todos

-

## Checklist


anselmbradford:
We should probably start standardizing on `process.env.NODE_ENV` for setting environments https://webpack.js.org/guides/production/#specify-the-environment (by including it in https://github.com/cfpb/cfgov-refresh/blob/master/config/environment.js#L10). Would it make sense to set the dev build through that envvar here?

sebworks:
>>  Would it make sense to set the dev build through that envvar here?

Yeah, we can do that. You would still need the ability to run in both modes though locally.

anselmbradford:
> Yeah, we can do that. You would still need the ability to run in both modes though, locally.

The reason to run production locally is to check the build process as it will appear on prod, correct?

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/index.js
```diff
@@ -23,43 +23,22 @@ require( '../../node_modules/rangeslider.js' );
 require( './tab' );
 require( '../placeholder-polyfill' );
 
+var TEMPLATE_DIR = '../../templates/explore-rates/';
```
jimmynotjim: Did the webpack update necessitate this change?

jimmynotjim:
Should `gulp watch` run the `--dev` flag by default?

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/index.js
anselmbradford: The webpack handlebar loader `handlebars-loader!` wouldn't work if the path included a variable, but Steven's moves that to the webpack-config, which makes better sense. It's not specifically webpack 4 I presume so could be spun out of here into its own PR if he wants.

sebworks:
>> The reason to run production locally is to check the build process as it will appear on prod, correct?

Yeah

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/index.js
sebworks: Yeah, this was more to show that each app could have it's own webpack config. You definitely want to handle things like this in Webpack but I didn't have to do it here.

sebworks:
>> Should gulp watch run the --dev flag by default?

hmm, the only what I know how to do that is with usuing `exec`.

gulp/tasks/scripts.js
```diff
@@ -33,6 +35,11 @@ const webpackStream = require( 'webpack-stream' );
  * @returns {PassThrough} A source stream.
  */
 function _processScript( localWebpackConfig, src, dest ) {
+  const cmdLineArgs = gulpUtil.env;
```
anselmbradford: Deprecated, should be replaced with https://www.npmjs.com/package/minimist (see https://medium.com/gulpjs/gulp-util-ca3b1f9f9ac5), though I'm unclear how.

config/webpack-config.js
```diff
@@ -118,8 +123,16 @@ const spanishConf = {
   ]
 };
 
+const devConf = {
```
anselmbradford: So the idea here is that `devConf` is the development configuration on top of all other webpack configs, correct?

Another way to do it would be to set this in https://github.com/cfpb/cfgov-refresh/blob/master/config/environment.js#L24 

```js
// environment.js

…
NODE_ENV = process.env.NODE_ENV
…
```

And then at the top of this file set:

```js
// webpack-config.js

const envvars = require( '../config/environment' ).envvars;
…
const env = envvars.NODE_ENV
…
```

And then at the bottom set:

```js
…
const configExports = {
  commonConf,
  devConf,
  externalConf,
  modernConf,
  onDemandHeaderRawConf,
  appsConf,
  spanishConf
};

if ( env === 'development' ) {
  for (let key in configExports) {
    Object.assign( configExports[key], devConf );
  }
}

module.exports = configExports;
```

Then elsewhere if you want to set an environment flag (`--dev` or `--development` or whatever), that flag would set the value of `process.env.NODE_ENV` and the changes would trickle down where ever they are needed, instead of having a dev flag for each command.

anselmbradford:
Documentation for individual app webpack config added in https://github.com/cfpb/cfgov-refresh/pull/3873

config/webpack-config.js
sebworks: ok, I've removed the ability to set the flag when running gulp scripts. You must set the flag using `export NODE_ENV=development` or setting the variable in your `.bash_profile`.

anselmbradford: Thanks! I'm thinking the direction we'll want to go is to enable overriding `NODE_ENV` in cli-flags.sh (not here and now, but in the future), that way the `./frontend production` or whatever would be the central place to flip environments, and then elsewhere we'd only be referencing the value of `NODE_ENV` for conditional code based on environment. Does that sound like it makes sense?

anselmbradford:
Works for me! I get two warnings we'll want to address as some point: `DeprecationWarning: Tapable.plugin is deprecated. Use new API on `.hooks` instead`, and `WARNING in asset size limit` and `WARNING in entrypoint size limit`.

Also, missing `semi` needs fixing for Travis.

PR #3849, opened 2018-03-06
Replace gulpUtil with fancyLog

anselmbradford:
gulp-util is deprecated and log should be migrated to fancy-log per https://medium.com/gulpjs/gulp-util-ca3b1f9f9ac5 

## Changes

- Replace gulp-util.log with fancy-log.
- Replace gulp-util.colors with ansi-colors.

## Testing

1. `npm install`
2. `gulp audit:perf` and `gulp audit:a11y` should log to the console as before (with server running).
3. `gulp test` and `gulp docs` should pass.


sebworks:
Should gulp util be removed from package.json.

anselmbradford:
> Should gulp-util be removed from package.json?

Not yet. This only addresses `gulpUtil.log` (and `gulpUtil.colors`). We have `gulpUtil.noop` and `gulpUtil.env` usage that needs to be migrated with different packages.

PR #3861, opened 2018-03-08
Fix related posts URLs for duplicate slugs

chosak:
The current logic in the Related Posts template is using the built-in Wagtail
[`slugurl`](http://docs.wagtail.io/en/v2.0/topics/writing_templates.html#slugurl) template tag, which returns an undetermined link if there are multiple pages with the same slug anywhere in the site.

This change fixes GHE#2533 which was generating an incorrect link due to two pages having the same slug. It instead uses the built-in `page.url` property which returns the best URL for the page.

## Changes

- Use `page.url` instead of `slugurl(page.slug)` in related posts template.

## Testing

To test, use a production dump and visit http://localhost:8000/data-research/ and observe that the "2018 CFPB Research Conference" link properly points to an event page, not the conference landing page.

## Screenshots

image

## Notes

1. Note that due to the current way we have our production Wagtail Sites configured (we have multiple), `page.url` may return a full absolute URL rather than a relative URL here. I think the downside of this is outweighed by the simplicity of the required code change. Otherwise we'd
have to put in some complex logic like `page.relative_url(page.get_site())` which I would rather not stick in the template.

    If, hopefully, in future we move to a single Wagtail Site in production these links will properly be relative ones.

2. @higs4281 the only other places where `slugurl` is used in this repository is in Ask templates, e.g. [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/ask-cfpb/answer-page-spanish-printable.html#L47) or [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/ask-cfpb/answer-page.html#L135). Is there a chance that any of these places (there are 6 in total, try `git grep slugurl`) would ever be checking against a duplicate slug, and thus need a similar replacement?

## Checklist

* [X] Project documentation has been updated

### DIFF
--- a/cfgov/jinja2/v1/_includes/molecules/related-posts.html
+++ b/cfgov/jinja2/v1/_includes/molecules/related-posts.html
@@ -81,7 +81,7 @@
                m-list__unstyled
                m-list__spaced">
         {% for i in range(limit) %}
-            {% set post_url = slugurl(posts[i].slug) or '' %}
+            {% set post_url = posts[i].url or '' %}
             <li class="m-list_item">
                 <h3 class="h4 u-mb5">
                     <a class="m-list_link"


chosak:
@higs4281 thanks for reviewing. Just to clarify, the risk is that any page across the entire site might have that slug, not just within the same section of Ask. So if, say, the Auto section of CCT [that ends in `/auto-loans/`](https://www.consumerfinance.gov/data-research/consumer-credit-trends/auto-loans/) instead had a slug and ended in `/category-auto-loans/`, there'd be a conflict with Ask's [`/category-auto-loans/` page](https://www.consumerfinance.gov/ask-cfpb/category-auto-loans/), and we'd run into this problem.

richaagarwal:
Would it be worth creating a test that fails before this change, and passes with this change?

chosak:
@richaagarwal thanks for the review;  thought about that and decided against it, since it would essentially be testing that the code was written in a certain way (using `page.url`) and not really testing the functionality of anything. We could potentially have tests for organisms like this that verify that their HTML renders in a certain way, but that seemed out of scope for this minor bug fix.

higs4281:
Late to the party, but you have a good point, @chosak 
slugurl's assumption that the provided slug is unique site-wide means it should only be used in narrow cases, if at all.

PR #3862, opened 2018-03-09
Move test tooling to Jest

anselmbradford:
[Istanbul has been deprecated](https://github.com/gotwarlost/istanbul):

> Deprecation Notice: this version of istanbul is deprecated, we will not be landing pull requests or releasing new versions. But don't worry, the Istanbul 2.0 API is now available and is being actively developed in the new istanbuljs organization.

This is preventing some new tests from being written, such as those in https://github.com/cfpb/cfgov-refresh/pull/3830.

The Istanbul 2.0 API is significantly different. After chatting with @sebworks, instead of migrating istanbul, we could use [jest](https://facebook.github.io/jest/), which largely keeps the same syntax we have, but includes jsdom and istanbul code coverage.

## Additions

- Adds default Babel config to package.json.
- Adds Jest config to package.json.

## Removals

- Removes empty tests (Jest doesn't allow this).
- Removes jsdom from package.json
- Removes gulp-istanbul from package.json
- Removes mocha from package.json

## Changes

- Changes gulp test task to run Jest for JS unit tests.
- Changes `o-table-row-links.js` to set window URL with `window.location.assign(…)` from `window.location = …` so that the location method can be mocked in tests.
- Changes `o-table-row-links-spec.js` to check for calls to `window.location.assign(…)` instead of checks of the URL.
- Removes tests path from the gulp config in favor of retrieving the path through the environment.js.
- 

## Testing

1. Run `./frontend.sh`
2. Run `gulp unit:test` to run all JS unit tests.
3. Run `gulp test:unit --specs=js/modules/` to run a directory of specs.
3. Run `gulp test:unit --specs=js/modules/o-table-row-links-spec.js` to run a single spec.
4. Open `test/unit_test_coverage/lcov-report/index.html` to see code coverage report.

## Todos

- Chai and Sinon can be replaced with Jest too.


test/unit_tests/js/modules/util/js-loader-spec.js
```diff
@@ -4,17 +4,12 @@ const expect = chai.expect;
...
-    } );
-  } );
 
-  after( () => this.jsdom() );
+  /* TODO: Jest doesn't currently expose runscripts: 'dangerously'
+     Find a way to do this. See
+     https://www.npmjs.com/package/@mediaeventservices/jest-environment-jsdom-external-scripts */
 
-  it( 'should invoke the callback method when the script loads', () => {
+  xit( 'should invoke the callback method when the script loads', () => {
```
anselmbradford: Disabled till https://www.npmjs.com/package/@mediaeventservices/jest-environment-jsdom-external-scripts can be figured out.

package.json
```diff
@@ -106,5 +105,29 @@
     "test": "snyk test",
     "postinstall": "node scripts/npm/apps-install"
   },
-  "snyk": true
+  "snyk": true,
+  "babel": {
```
sebworks: lets move the babel and jest stuff to config files.

run_travis.sh
```diff
@@ -6,6 +6,7 @@ set -ex
 echo "running $RUNTEST tests"
 if [ "$RUNTEST" == "frontend" ]; then
     source $HOME/.nvm/nvm.sh
+    npm install --production
```
sebworks: So this file is really for just running processes, we have `install_travis.hs` which runs `frontend.sh test`.

package.json
```diff
@@ -80,15 +78,16 @@
   },
   "devDependencies": {
     "ansi-colors": "1.1.0",
+    "babel-jest": "22.4.1",
```
sebworks: This will need to be moved to `dependencies`.

test/unit_tests/js/modules/util/behavior-spec.js
```diff
@@ -128,28 +117,26 @@ describe( 'behavior', function() {
...
-        behavior.attach( linkDom, 'click', spy );
-        triggerEvent( linkDom, 'click' );
-        expect( spy.called ).to.equal( true );
-        spy.reset();
-        behavior.remove( linkDom, 'click', spy );
-        triggerEvent( linkDom, 'click' );
-        expect( spy.called ).to.equal( false );
-      } );
+    it( 'should remove the event callback for the specific behavior hook', () => {
+      const spy = sinon.spy();
```
sebworks: Lets get rid of Sinon here and throughout the other tests.  We should just switch to `jest.spyOn`.  This should allow us to eliminate Sinon as a depencency.

sebworks:
Nice work on this, @anselmbradford

run_travis.sh
anselmbradford: Do you see what I'm missing [here](https://travis-ci.org/cfpb/cfgov-refresh/jobs/351095910)? Travis was failing on the app dependencies not being installed and this fixed it. Should I put this in `frontend.sh test`?

anselmbradford: Also, how are the Travis scripts called? Are the filenames magic names, or could we name them `travis_install.sh` and `travis_run.sh` so they collect together in the file list? (Or can we combine them?? They're both run when Travis is run, no?)

sebworks: If you check in travis.yml you can see how things work.

package.json
anselmbradford: How come? It's a dependency of `jest`, which is in `devDependencies` too.

run_travis.sh
sebworks: >>> Do you see what I'm missing here?

looking into this, not sure why it's referencing `const unFormatUSD = require( '../../node_modules/unformat-usd' )` in `dom-values.js`

test/unit_tests/js/modules/util/behavior-spec.js
anselmbradford: Yeah I agree, since this is passing in its current state I thought I'd leave that to another PR for review sanity. Chai can be removed too, though we have to change the [matcher syntax](https://facebook.github.io/jest/docs/en/expect.html).

run_travis.sh
sebworks: I would try just removing the `'../../node_modules/`, I'm not sure why it's referenced that way if it's a dependency.

package.json
sebworks: yeah, you are right.

package.json
anselmbradford: Moved to `.babelrc` and `jest.config.json` in https://github.com/cfpb/cfgov-refresh/pull/3862/commits/d54399bb5e5b426c7a93b92d75ea6b4392939a3a. I wanted to put the jest config in /test/ but had trouble with the paths resolving for the coverage source files.

PR #3864, opened 2018-03-12
Remove sinon and chai from unit tests

anselmbradford:
## Removals

- Remove sinon in favor of Jest's built-in mocking.
- Remove chai in favor of Jest's built-in matchers.

## Testing

1. `./frontend.sh`

## Notes

- Chai matchers are used in the acceptance tests so the module isn't removed.


test/unit_tests/js/modules/util/validators-spec.js
```diff
@@ -129,10 +124,10 @@ describe( 'Validators', () => {
       ];
       returnedObject = validators.checkbox( testField, null, fieldset );
 
-      expect( returnedObject ).to.have.property( 'checkbox', false );
-      expect( returnedObject ).to.have.property(
-        'msg', ERROR_MESSAGES.CHECKBOX.REQUIRED.replace( '%s', '2' )
-      );
+      expect( returnedObject['checkbox'] ).toBe( false );
```
anselmbradford: After changing this I realized Jest has a `toHaveProperty` matcher as well, but as far as I can see this works as well and is perhaps clearer that it's checking the value of the property.

PR #3866, opened 2018-03-12
Convert Jest config to JS

anselmbradford:
Jest config can use JS instead of JSON, which arguably simplifies the code (less quotes needed). 

## Changes

- Converts Jest config to JS file from JSON file.
- Removes `globals` and `testURL` settings that didn't appear to be needed for the tests to pass.

## Testing


gulp/config.js
```diff
@@ -33,7 +33,8 @@ module.exports = {
       'config/**/*.js',
       'gulpfile.js',
       'gulp/**/*.js',
-      'scripts/npm/**/*.js'
+      'scripts/npm/**/*.js',
+      'jest.config.js'
```
sebworks: Where does this get used at?

anselmbradford: https://github.com/cfpb/cfgov-refresh/blob/master/gulp/tasks/test.js#L61
I really wish it could go in /test/ but I couldn't figure out the paths, since it needs reference things in both /cfgov/unprocessed/ and /test/ from the root directory.

PR #3868, opened 2018-03-12
Removes jQuery from OAH tabs and adds coverage

anselmbradford:
## Changes

- Removes jquery from OAH rate checker tabs.
- Adds ~~100%~~ 95.24% test coverage for OAH rate checker tabs code.

## Testing

1. `gulp clean && gulp build`
2. http://localhost:8000/owning-a-home/explore-rates/ and interact with the tabs half-way down the page.
3. `gulp test:unit` should pass.

## Screenshots

screen shot 2018-03-12 at 2 54 31 pm

screen shot 2018-03-12 at 2 54 37 pm

## Notes
- There will be a flash of both tab content showing before the JS runs. In this case it doesn't matter because the tabs are below the page fold, but it could be bothersome in another context. However, here I think it's worthwhile to remove the js CSS that's needed to prevent the flash.

## Todos

- These tabs should be re-written in the markup so that when JS is disabled the tab link text is laid out immediately above its contents (this is existing behavior):

screen shot 2018-03-12 at 2 53 42 pm

### DIFF
--- a/cfgov/jinja2/v1/owning-a-home/explore-rates/index.html
+++ b/cfgov/jinja2/v1/owning-a-home/explore-rates/index.html
@@ -368,7 +368,7 @@ home price, down payment, and more can affect mortgage interest rates.
                     </section>
                 </div>
                 <!-- END .result -->
-                <section class="next-steps">
+                <section class="next-steps tabs-layout">
                     <h2><strong>Next steps:</strong> How to get the best interest rate on your mortgage</h2>
                     <!-- put into a container to prevent long line length -->
                     <div class="wrapper">
--- a/cfgov/unprocessed/apps/owning-a-home/js/explore-rates/index.js
+++ b/cfgov/unprocessed/apps/owning-a-home/js/explore-rates/index.js
@@ -1,4 +1,4 @@
-import * as rateExplorer from './rate-checker';
+import * as rateChecker from './rate-checker';
 
 // Do it!
-rateExplorer.init();
+rateChecker.init();
--- a/cfgov/unprocessed/apps/owning-a-home/js/explore-rates/tab.js
+++ b/cfgov/unprocessed/apps/owning-a-home/js/explore-rates/tab.js
@@ -1,20 +1,14 @@
-let tabGroups;
-let tabLinks;
-let tabContents;
-
 /**
  * Initialize the functionality of a group of tabs with content.
  */
 function init() {
-  tabGroups = document.querySelectorAll( '.tabs-layout' );
+  let tabGroups = document.querySelectorAll( '.tabs-layout' );
+  let tabContents;
 
   tabGroups.forEach( tabGroup => {
-    tabLinks = tabGroup.querySelectorAll( '.tab-link' );
     tabContents = tabGroup.querySelectorAll( '.tab-content' );
 
-    tabLinks.forEach( tabLink => {
-      _bindTabLink( tabGroup, tabLink, tabContents );
-    } );
+    _bindTabLink( tabGroup, tabContents );
 
     // Hide all but the first tab.
     for ( let i = 1; i < tabContents.length; i++ ) {
@@ -24,12 +18,11 @@ function init() {
 }
 
 /**
- *
  * @param {HTMLNode} tabGroup - HTML element parent of tabs and tabs-content.
- * @param {HTMLNode} tabLink - HTML element for links in tabs.
+ * @param {HTMLNode} tabContents - HTML element for content in tabs.
  */
-function _bindTabLink( tabGroup, tabLink, tabContents ) {
-  tabLink.addEventListener( 'click', _tabLinkClicked );
+function _bindTabLink( tabGroup, tabContents ) {
+  tabGroup.addEventListener( 'click', _tabLinkClicked );
   const tabs = tabGroup.querySelectorAll( '.tab-list' );
 
   /**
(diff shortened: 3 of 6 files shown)


jimmynotjim:
Any idea why the codecov/patch is less than 100%?

jimmynotjim:
@anselmbradford I'm getting some django errors but I don't think they're related to this PR. Any ideas?

screen shot 2018-03-13 at 5 49 06 pm

anselmbradford:
> Any idea why the codecov/patch is less than 100%?

Yea, because I made tabs a module with an init method that then gets called in the rate checker's main init method, which doesn't have coverage. It'll be covered when that is covered and I'm almost there with it but having trouble with the handlebar templates.

anselmbradford:
@jimmynotjim not sure on that django error. Just pulled this down and tested and it seemed fine. Maybe run `./backend.sh`?

sebworks:
>>> @jimmynotjim not sure on that django error. Just pulled this down and tested and it seemed fine. Maybe run ./backend.sh?

It just means you don't have OAH installed.

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/tab.js
```diff
@@ -1,19 +1,53 @@
...
+}
 
-  evt.preventDefault();
+/**
+ *
+ * @param {HTMLNode} tabGroup - HTML element parent of tabs and tabs-content.
+ * @param {HTMLNode} tabLink - HTML element for links in tabs.
+ */
+function _bindTabLink( tabGroup, tabLink, tabContents ) {
+  tabLink.addEventListener( 'click', _tabLinkClicked );
```
sebworks: Use event delegation here.

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/tab.js
```diff
@@ -1,19 +1,53 @@
...
+   * @param {MouseEvent} evt - Event object for a click on a tab.
+   */
+  function _tabLinkClicked( evt ) {
+    evt.preventDefault();
+
+    const target = evt.target;
+    const tabLi = target.parentNode;
+    const current = tabGroup.querySelector( target.getAttribute( 'href' ) );
+
+    tabs.forEach( tab => tab.classList.remove( 'active-tab' ) );
```
sebworks: Use the dom-tools library to ensure we don't thrash.

sebworks:
We should add this to Capital Framework and Wagtail.

cfgov/jinja2/v1/owning-a-home/explore-rates/index.html
```diff
@@ -392,7 +392,7 @@ <h4 class="rc-comparison-subhead">Interest costs over <span class="loan-years">3
 
           </div><!-- END .result -->
 
-          <section class="next-steps">
+          <section class="next-steps tabs-layout">
```
sebworks: Should we make this atomic?

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/tab.js
anselmbradford: Are we sure this is needed here since it's only two tabs? dom-tools requires a second dom lookup because the method signature for `removeClass` is a selector not a HTMLNode.

cfgov/jinja2/v1/owning-a-home/explore-rates/index.html
anselmbradford: I wouldn't bother at this point. I think the whole tabs bit needs to be re-written. As is the tabs when in a non-JS state doesn't place the tab button text next to its content, but instead groups the tab button text and groups the content. If this goes into CF it'll get atomized and then can replace what's here.

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/tab.js
anselmbradford: Fixed! https://github.com/cfpb/cfgov-refresh/pull/3868/commits/b948eb4e4ccb86c2a8b2601616c518cafefe7bf6

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/index.js
```diff
@@ -21,7 +21,8 @@ const formatTime = require( '../format-timestamp' );
 const params = require( './params' );
 
 require( '../../node_modules/rangeslider.js' );
-require( './tab' );
+const tab = require( './tab' );
```
sebworks: Use import statements here.

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/tab.js
sebworks: So you can pass in a Node or selector like: `DT.removeClass( tab, 'active-tab' ).

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/index.js
anselmbradford: Fixed!

anselmbradford:
@sebworks ready to check again.

sebworks:
looks good, 👍

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/tab.js
anselmbradford: Oh I see, it passes it to `isEl`. Would we want this for all CSS class manipulation, such as `tabLi.classList.add( 'active-tab' )` on the next line? Any way to test whether it's needed? I'll merge this for now to get the tests in and we can adjust.

anselmbradford: Perhaps we should jsdoc the `selector` parameter's type. That's where it's useful 😁

PR #3871, opened 2018-03-13
Tweak Travis config

anselmbradford:
## Removed
- Removed calls to nvm, which was not used.

## Changes

- Rename travis scripts to begin with `travis_` so they collect together in file lists.
- Add code comment around app node_modules.
- Removed quotes around gulp "test".

## Testing

1. Travis should pass in this PR.


sebworks:
LGTM,

PR #3874, opened 2018-03-13
Remove test environment condition from scripts

anselmbradford:
## Removals

- Removes "test" environment from cli-flag.sh workflow.

## Changes

- Makes cli-flag.sh exit when called directly.
- Simplifies conditionals in cli-flag.sh

## Testing

1. `./cli-flag.sh` should output `ERROR: cli-flag.sh should only be called from frontend.sh or backend.sh.`
2. `./frontend.sh` should output:
```
WARNING: '' flag not found, reverting to development environment.
Front end environment: development
```
3. `./frontend.sh production` should output:
```
Front end environment: production
```
4. `./frontend.sh development` should output:
```
Front end environment: development
```
5. `./backend.sh` and `./setup.sh` should behave the same way.

**NOTE: You may want to `ctrl+c` the commands so you don't have to wait for them to fully run once you know it runs through fine the first time.

## Notes

- It seems like we should be defaulting to the production environment, since if something wasn't set properly in production we'd want that to be the default, not development. But I'm not currently seeing an easy way to handle that because we also don't want to always have a flag for local development.


backend.sh
```diff
@@ -29,8 +29,6 @@ install() {
   # Install requirements for Django Server or tox.
   if [ "$cli_flag" = "development" ]; then
     pip install -r ./requirements/local.txt
-  elif [ "$cli_flag" = "test" ]; then
-    pip install -r ./requirements/test.txt
```
sebworks: I'm not sure this would work without the testing dependencies being cached on Travis.

sebworks:
I'll be honest, I think this line of thinking is a mistake.  There are dependencies and processes that you want / need just when running on Travis / Jenkins. Just because`package.json` only allows for `dependencies / devDependencies` doesn't mean that is a paradigm we should follow.  `jasmine-reporters` and `jasmine-spec-reporter` should really only be installed on Jenkins.

I understand that many of the tests you run locally, you run on Travis / Jenkins but that shouldn't always be the case.  There should really be tiered approach to testing which varies based on time, risk, and environment. Eliminating the testing workflow moves us away from the idea.

Even if you don't see the benefits today, we should keep the environments, dependencies, and processes separated.

backend.sh
anselmbradford: How do we check this? The PR passes Travis.

chosak: I don't think this would be necessary for the backend tests, as [the tox config](https://github.com/cfpb/cfgov-refresh/blob/master/tox.ini#L29) also installs this file. So unless there's something in this file that's needed for frontend, it can probably be removed.

willbarton: Yeah, [Travis doesn't use the `backend.sh` script to install test-running dependencies at all](https://github.com/cfpb/cfgov-refresh/blob/master/travis_install.sh#L13). Travis installs `requirements/travis.txt`, which includes the tools necessary to run the tests (like `tox`). Tox then installs`requirements/test.txt`, which includes the things required by the tests (libraries like `mock` and `model_mommy`).

anselmbradford:
@sebworks Thanks for the input. I have a couple questions for understanding the requirements here better:

- Does `backend.sh` not being used in Travis change anything in how we should think about environments?

- Would the `test` environment always be a superset of either `development` or `production`? E.g. should it be the full production environment plus tests, which might help avoid the issues we had of things passing locally and failing in Jenkins?

- Other than the removal of the backend `pip install -r ./requirements/test.txt` this PR doesn't change the existing behavior or make the architecture such that a `test` environment couldn't be added in the future. Would it make sense to add the `test` environment when we need it? For instance, it could be added back into the `if` condition in frontend.sh, but as it currently exists it's identical to `development`.

willbarton:
>  Does backend.sh not being used in Travis change anything in how we should think about environments?

The way our Travis/tox setup is currently is pretty close to how it should be, IMHO. `backend.sh` is a convenience script for humans (which will probably be totally deprecated if we go docker-only when we can use Docker for Mac).

> Other than the removal of the backend pip install -r ./requirements/test.txt this PR doesn't change the existing behavior or make the architecture such that a test environment couldn't be added in the future. Would it make sense to add the test environment when we need it? For instance, it could be added back into the if condition in frontend.sh, but as it currently exists it's identical to development.

I think overall though it's important to remember that `tox`, on the backend, is the tool that manages our backend test environments. We don't need additional setup that duplicates that. So, there's no workflow that's being removed here, from a backend perspective.

backend.sh
sebworks: ok, thanks @willbarton and @chosak.

anselmbradford:
> Would the test environment always be a superset of either development or production? E.g. should it be the full production environment plus tests, which might help avoid the issues we had of things passing locally and failing in Jenkins?

Hmm, Travis/Jenkins don't need to write coverage files, right? This makes me think more that they should be the production environment + testing frameworks.

willbarton:
> Hmm, Travis/Jenkins don't need to write coverage files, right? This makes me think more that they should be the production environment + testing frameworks.

Travis does need to write coverage files since that's how codecov gets its data.

sebworks:
>> Does backend.sh not being used in Travis change anything in how we should think about environments?

No, it doesn't per the BEWDS.


>> Would the test environment always be a superset of either development or production? E.g. should it be the full production environment plus tests, which might help avoid the issues we had of things passing locally and failing in Jenkins?

>> Would the test environment always be a superset of either development or production?

It's tricky because something like BrowserSync will be installed as a `dependency` but only because it runs as a part of our gulp process. You only need it when running in development mode but it still needs to get installed on production.
>> E.g. should it be the full production environment plus tests, which might help avoid the issues we had of things passing locally and failing in Jenkins?

Maybe, we don't run front-end unit or acceptance tests on Jenkins now.  I think the failures are related to visibility ( Jenkins access and understanding ) and differing configurations

sebworks:
@anselmbradford, go ahead and merge this PR. We can always come back around on the differences.

anselmbradford:
> go ahead and merge this PR. We can always come back around on the differences.

@sebworks Okay, will do. I think you are right. I think we do need a test environment, but based on this discussion seems like it should be production + tests (e.g. no BrowserSync). For future reference, we'd want to add `test` back into the if condition, like [here](https://github.com/cfpb/cfgov-refresh/pull/3874/files#diff-345587b57bc55bfdcb8027c7504432b2L48) and change [this](https://github.com/cfpb/cfgov-refresh/pull/3874/files#diff-bf759acc54a937513160a4407f997f21R8) back to `test`.

PR #3879, opened 2018-03-14
Adds template-loader

anselmbradford:
## Changes

- Moves handlebar templates from rate-checker to their own module and adds unit tests.
- Edits handlebar templates to adjust long line lengths and fix errors like `</text>` and missing `</p`>.

## Testing

1. `gulp test:unit --specs=apps/owning-a-home/js/explore-rates/template-loader-spec.js` should pass.

### DIFF
--- a/gulp/tasks/test.js
+++ b/gulp/tasks/test.js
@@ -53,9 +53,15 @@ function testUnitScripts( cb ) {
     fileTestRegex += '.*-spec.js';
   }
 
+  /*
+    The --no-cache flag is needed so the transforms don't cache.
+    If they are cached, preprocessor-handlebars.js can't find handlebars. See
+    https://facebook.github.io/jest/docs/en/troubleshooting.html#caching-issues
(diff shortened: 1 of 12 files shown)


gulp/tasks/test.js
```diff
@@ -53,9 +53,15 @@ function testUnitScripts( cb ) {
     fileTestRegex += '.*-spec.js';
   }
 
+  /*
+    The --no-cache flag is needed so the transforms don't cache.
+    If they are cached, preprocessor-handlebars.js can't find handlebars. See
+    https://facebook.github.io/jest/docs/en/troubleshooting.html#caching-issues
```
anselmbradford: Alternatively to this we could install `handlebars` in the main `package.json`. However, that means two installs of handlebars (in the main package.json and in the oah package.json). I don't see a slow-down without the cache, but ideally we'd address this with `getCacheKey`.

PR #3882, opened 2018-03-15
Fix data-set spec coverage

anselmbradford:
## Changes

- ~~Fixes data-set test, which wasn't properly removing `dataset` property from jsdom document, leading to untested code and reduced coverage.~~ Swaps data-set.js with `elem-dataset` ponyfill.

## Testing

1. `gulp build` should pass.

## Notes

- Data-set module is only used [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/js/modules/VideoPlayer.js#L9) and appears to only be for IE 10.


sebworks:
Lets remove the test and the dependency. Lets just use `https://github.com/awcross/elem-dataset`.

anselmbradford:
> Lets remove the test and the dependency. Lets just use https://github.com/awcross/elem-dataset.

Does that change anything beyond removing the test? The instructions show:
```js
import elemDataset from 'elem-dataset';

const element = document.querySelector('.foo');
const attributes = elemDataset(element);
```
Which looks like we'd still need to import it in the video player?
If that's the case, I'd just leave this as is (with full coverage in this PR) and remove the whole thing when IE10 support is dropped. That way we're not touching package.json.

sebworks:
>> Which looks like we'd still need to import it in the video player?
If that's the case, I'd just leave this as is (with full coverage in this PR) and remove the whole thing when IE10 support is dropped. That way we're not touching package.json.

My thinking is that we should be moving towards removing most of the js utilities which now are supplanted by popular npm modules ( lodash in many cases ).  This allows us to simplify our testing process and reduce the complexity of the repo.

anselmbradford:
Okay swapped it out! @sebworks are you able to check where that change in VideoPlayer is used or direct me to it? I'm not sure how it uses data-set.

sebworks:
thanks @anselmbradford, so you can test the archived events by going to:
 http://localhost:8000/about-us/events/archive-past-events/fall-2017-consumer-advisory-board-meeting-tampa-fla/ 

and ensuring the video player works.

PR #3883, opened 2018-03-15
Remove redundant throttle function

anselmbradford:
~~The throttle function was included in email-popup-helpers.js, but there's an identical throttle module that could be used.~~

## Changes

- Replaces `throttle.js` with `lodash.throttle`

## Testing

1. `gulp build` should pass.
2. Pages such as /consumer-tools/debt-collection should show popup.


sebworks:
I would say lets remove it and go with `https://www.npmjs.com/package/lodash.throttle`.

anselmbradford:
> I say lets remove it and go with https://www.npmjs.com/package/lodash.throttle.

That gives us the whole lodash library in node_modules, even with only `lodash.throttle` in `package.json`, doesn't seem worth it?

screen shot 2018-03-15 at 11 00 00 am

anselmbradford:
Oh wait, nevermind, looks like we're already installing the full library as it's a dep of a bunch of things.

anselmbradford:
Okay, @sebworks, replaced it! You can force showing the popup with https://github.com/cfpb/cfgov-refresh/pull/3883/commits/be7fd7e19962b0e6f87463a402e12d2913f44d96

sebworks:
Looks good, I think it's also used on the `/ask-cfpb/` type ahead.

PR #3894, opened 2018-03-19
Add testing matrix to tox for Django 1.8-2.0, Wagtail 1.10-2.0, and Python 2.7 and 3.6

willbarton:
Per discussion last week with @chosak @higs4281 @hillaryj @richaagarwal, this PR makes significant changes to our `tox.ini` to add the ability to run tests on different versions of our major dependencies, Django, Wagtail, and Python.

`tox` now supports the following env matricies:

- `lint-py{27,36}` to run `flake8` and `isort` on Python source files
- `unittest-py{27,36}-dj{18,110}-wag{110,113}-{fast,slow}`, to run Python unittests
- `unittest-py{36}-dj{20}-wag{20}-{fast,slow}`, to run unittests on Django 2.0+
- `missing-migrations-py{27,36}-dj{18,110}-wag{110,113}`, to check for missing migrations
- `acceptance-py{27,36}-dj{18,110}-wag{110,113}-fast`, to run acceptance tests

In addition, our currently used envs are still supported and use the underlying settings for `unittest-py27-dj18-wag110-fast` as their basis:

- `lint`
- `fast`
- `missing-migrations`
- `acceptance`
- `validate-assets`

Travis will continue running `lint`, `fast`, and `missing-migrations` as it has done.

When you invoke `tox` without a `-e` argument locally, it will run a simple matrix of:

- `lint-py27`
- `lint-py36`
- `unittest-py27-dj18-wag110-fast`
- `unittest-py27-dj18-wag113-fast`
- `unittest-py27-dj111-wag110-fast`
- `unittest-py36-dj111-wag113-fast`

To support this the requirements files are refactored to add the following files:

- `django.txt` is our pinned version of Django
- `wagtail.txt` is our pinned version of Wagtail
- `libraries.txt` includes all our minor dependencies that used to be in `base.txt`

## Checklist

### Browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


willbarton:
**Note**: After the merge of the PR that adds tox-pip-enstensions and `tox_pip_extensions_ext_venv_update=True`, we have dependency conflicts that prevent this from running. I'm trying to work through those now.

willbarton:
FYI: I have resolved the awesome-slugify Unidecode dependency conflict with Wagtail by removing our dependency on awesome-slugify in https://github.com/cfpb/cfgov-refresh/pull/3894/commits/8c982e28c98f9aa0254eadc762ead0d155dabfc0. We were just using it in the `agreements` management commands to generate a lowercase slug. I presume this was a legacy need, but Django 1.8's `slugify` will generate lowercase slugs the same as awesome-slugify, so this doesn't seem necessary anymore.

cfgov/agreements/management/commands/util.py
```diff
@@ -27,7 +28,7 @@ def upload_to_s3(pdf_obj, s3_key):
 
 
 def get_issuer(name):
-    slug = slugify(name, to_lower=True)
+    slug = slugify(name)
```
chosak: Unfortunately, there appears to be a slight incompatibility between awesome-slugify's `slugify(to_lower=True)` and Django's `slugify`. I don't want to post the production values here, but if you run this code against the existing `agreements.Issuer` model instances, you'll find three cases where the slugs don't match:

```py
from slugify import slugify
from django.utils.text import slugify as dj_slugify

for i in Issuer.objects.all():
    if slugify(i.name, to_lower=True) != dj_slugify(i.name):
        print(i.name)
```

Obfuscated, these are banks that have special characters, e.g. `W.B. Bank` and `W&B Bank`. Using awesome-slugify's logic, these both become `w-b-bank`. Using Django's logic, they become `wb-bank`.

It appears this difference is because [Django](https://docs.djangoproject.com/en/1.8/ref/utils/#django.utils.text.slugify)

> Removes characters that aren’t alphanumerics, underscores, or hyphens.

but awesome-slugify doesn't do that.

In practice, what this would mean immediately is that some existing URLs (like https://www.consumerfinance.gov/credit-cards/agreements/issuer/m-t-bank/) would break and be replaced by slightly different ones (https://www.consumerfinance.gov/credit-cards/agreements/issuer/mt-bank/). The agreements would still be discoverable through the search [here](https://www.consumerfinance.gov/credit-cards/agreements/), but if someone happened to bookmark a page, it would break.

Whether this is significant enough to hold this change back is a question I will defer to stakeholders about.

requirements/test.txt
```diff
@@ -1,4 +1,7 @@
--r base.txt
+# Django and Wagtail have to installed seperate from requirements
```
chosak: The necessity of this comment feels a little bit like a sign that we don't have these files set up optimally.

Some of our existing requirements files (`manual.txt`, `mysql.txt`, `scripts.txt`, `travis.txt`) are standalone additions to a base set of requirements necessary to work with this repository. But others (`local.txt`, `deployment.txt`, `test.txt`) include child requirements files within themselves.

How would you feel about moving `test.txt` from the latter group to the former, and directing tox to explicitly install all of the necessary child requirements?

cfgov/agreements/management/commands/util.py
willbarton: Would it make sense to import the awesome-slugify `slugify()` function instead then? This is a conflict we have right now. (I'm not sure why tox-pip-extensions doesn't complain about it... maybe because they're specified in the same requirement file right now?)

requirements/test.txt
willbarton: That's fine by me.

cfgov/agreements/management/commands/util.py
chosak: I've updated my comment above. @higs4281 correctly points out that this won't affect live URLs, but will only affect things the next time the data is loaded quarterly. This change still means that those bank URLs will change slightly upon the next load.

willbarton:
> I do fear for someone who errantly runs `tox'

@higs4281 I know we're going to have errors, but I feel like we should encourage folks to start matrix testing their code. Fixing the Django 1.11 and Wagtail 1.13 errors seems out of scope for this PR. 

Should we make `tox` simply run `lint-py27` and `unittest-py27-dj18-wag110-fast` for now and then add things back as they succeed?

higs4281:
> add things back as we make them succeed?

That would be friendly, but I don't feel strongly either way. It's easy enough to bail from the bare tox command if you want.

I favor going forward with the matrix. Its complexity is worth the price because it moves us closer to modern code, and to a time when we won't need such an involved matrix.

chosak:
I would vote to exclude the Django 1.11 and Wagtail 1.13 environments from the default tox for now if they don't already succeed. The cf.gov-backend-tests job in GHE currently runs vanilla `tox` and so would need to be updated to skip any non-working environments. Generally also developers new to a project or to parts of the code might reasonably expect that `tox` works out of the box.

willbarton:
Alrighty, I've modified the default `tox` run to behave exactly like our bare `tox` does currently and rebased.

Could I get a final 👍 or 👎 on this PR?

--

I would *highly* encourage folks to feel empowered to make their code support future versions of things by running tests they're actively working on selectively with the following envs:

```shell
tox -e lint-py27
    -e lint-py36 \
    -e unittest-py27-dj18-wag110-fast \
    -e unittest-py27-dj18-wag113-fast \
    -e unittest-py27-dj111-wag110-fast \
    -e unittest-py27-dj111-wag113-fast \
    -e unittest-py36-dj18-wag110-fast \
    -e unittest-py36-dj18-wag113-fast \
    -e unittest-py36-dj111-wag110-fast \
    -e unittest-py36-dj111-wag113-fast \
    my.great.test_case
```

I'll try to do what I can to make it possible for the tests to run with `dj111` and `wag113`. Right now we're importing models too early in `v1/__init__.py` so everything breaks, so testing against Django 1.11 isn't even possible. And `wag113` fails on a wagtail-inventory version conflict.

Right now it is possible to run this matrix without error:


```shell
tox -e lint-py27
    -e lint-py36 \
    -e unittest-py27-dj18-wag110-fast \
    -e unittest-py36-dj18-wag110-fast \
    my.great.test_case
```

Which at least ensures Python3-compatibility.

But in general, this will greatly help us start writing future-friendly code and reduce the burden on anyone who has to handle these upgrades in the future. Which is just best practice anyway.

chosak:
:+1: Let's do it. Thanks for restoring `tox`.

> And `wag113` fails on a wagtail-inventory version conflict.

Not sure if you want to include it in this PR, but wagtail-inventory [0.4.1](https://github.com/cfpb/wagtail-inventory/releases/tag/0.4.1) supports Wagtail 1.13.

> Right now it is possible to run this matrix without error

By "without error", do you mean that Python 3.6 tests actually pass, or that `tox` works as it should and reports the various errors?

willbarton:
>> Right now it is possible to run this matrix without error

> By "without error", do you mean that Python 3.6 tests actually pass, or that tox works as it should and reports the various errors?

What I mean is it's possible to run actual test cases — with `dj111` and `wag113`, we currently have errors that cause tox to bail *before* running any test cases. with `py36-dj18-wag110` we can run actual test cases. The test cases *themselves* might (and do, if you run the full suite) fail, but it's possible to run individual tests and have them succeed.

`tox -e unittest-py36-dj18-wag110-fast regulations3k` will succeed on #3909, for example. 

`tox -e unittest-py36-dj18-wag110-fast` will run all the tests, and some tests will error or fail. 

`tox -e unittest-py27-dj18-wag113-fast regulations3k` will not run the regulations3k tests because tox will bail out before running tests on the wagtail-inventory dependency conflict.

willbarton:
> Not sure if you want to include it in this PR, but wagtail-inventory 0.4.1 supports Wagtail 1.13.

Yup, will do. That fixes the `dj18-wag113` envs, but we still have some other dependency conflicts with `dj111`. Fixing those will be another PR I think.

willbarton:
So I think the final combination of envs that it will be possible to run individual tests with are:
```
tox -e lint-py27
    -e lint-py36 \
    -e unittest-py27-dj18-wag110-fast \
    -e unittest-py27-dj18-wag113-fast \
    -e unittest-py36-dj18-wag110-fast \
    -e unittest-py36-dj18-wag113-fast \
    my.great.test_case
```

willbarton:
Next steps for being able to *run* the full matrix:

- [ ] Resolve dependency version conflicts with Django 1.11
- [ ] Resolve the `AppRegistryNotReady` error with Django 1.11 caused by model imports in `v1/__init__.py` and another in-code problems that prevent the test environment from running

PR #3900, opened 2018-03-20
Re-remove ie9-specific checks

anselmbradford:
IE9 script removal changes added in https://github.com/cfpb/cfgov-refresh/pull/3693/files were overridden with the old template when `base-common.html` was added. This re-adds the removal of IE9-specific coding.

### DIFF
--- a/cfgov/jinja2/v1/_layouts/base-common.html
+++ b/cfgov/jinja2/v1/_layouts/base-common.html
@@ -1,3 +1,7 @@
+{#
+    This is the shared template for both the
+    English and Spanish-language consumerfinance.gov.
+#}
 <!DOCTYPE html>
 {% if flag_enabled('CFPB_RECRUITING', request) %}
 <!--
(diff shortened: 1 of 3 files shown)


sebworks:
This looks good. Is there any reason that we are inlining Modernizr?

cfgov/jinja2/v1/_layouts/base-common.html
```diff
@@ -188,27 +191,18 @@
       Turn off JavaScript for browsers that don't support ECMAScript 5 features
       (e.g. Internet Explorer 8)
       by reversing no-js/js CSS class change made by modernizr.
-      The following ECMAScript 5 feature checks come from
+      The ECMAScript 5 feature checks are listed in
       https://github.com/Modernizr/Modernizr/tree/master/feature-detects/es5
     #}
     <script>
         var modernizr = window.Modernizr;
```
sebworks: Wrap this like  so the variables don't have global scrope
```
 <script>
      !function(){
        var modernizr = window.Modernizr;
        if ( !( typeof modernizr !== 'undefined' && modernizr.es5 ) ) {
          var docElement = document.documentElement;
          docElement.className = docElement.className.replace( /(^|\s)js(\s|$)/, '$1no-js$2' );
        }
      }()
    </script>
```

anselmbradford:
> This looks good. Is there any reason that we are inlining Modernizr?

It was suggested in chat to prevent the blocking request since it's before everything else.

cfgov/jinja2/v1/_layouts/base-common.html
anselmbradford: We'll want modernizr in the global scope if we had these checks like `modernizr.es5` elsewhere, but as we don't, this should be good 👍 
Fixed.

sebworks:
Merge when ready.

PR #3902, opened 2018-03-20
Normalizes template titles

anselmbradford:
## Removals

- Removes conditional comments that add `lt-ie7` CSS class, which is not used any more.
- Removes appcache code comment, since appcache is deprecated.

## Changes

- Normalizes template titles to use a pipe between the page name and the bureau name.

## Testing

1. Homepage should have the title `Consumer Financial Protection Bureau`. Other pages should be in the format of e.g. `About us | Consumer Financial Protection Bureau`


cfgov/jinja2/v1/_layouts/base-common.html
```diff
@@ -47,11 +46,8 @@
 
     <title>
         {%- block title -%}
-            {%- if page -%}
-                {{ page.seo_title or page.title }} | Consumer Financial Protection Bureau
-            {%- else -%}
-                Consumer Financial Protection Bureau
-            {%- endif %}
+            {%- if page -%} {{ page.seo_title or page.title }} | {%- endif %}
```
Scotchester: I think you want the `endif` like this to preserve the space after the pipe when it's true, but have no space before "Consumer" when it's false: `{% endif -%}

cfgov/legacy/templates/front/base_print.html
```diff
@@ -80,7 +79,8 @@
   }
 </script>
 <header>
-	<a href="/" class="noStyles"><img id="logo" src="/static/nemo/_/img/logo.png" alt="Consumer Financial Protection Bureau" title="CFPB"></a>
+	<a href="/" class="noStyles"><img id="logo" src="/static/nemo/_/img/logo.png"
+     alt="Consumer Financial Protection Bureau" title="CFPB"></a>
```
Scotchester: Mix of tab and space indentation is causing weirdness here.

cfgov/legacy/templates/front/base_print.html
```diff
@@ -80,7 +79,8 @@
   }
 </script>
 <header>
-	<a href="/" class="noStyles"><img id="logo" src="/static/nemo/_/img/logo.png" alt="Consumer Financial Protection Bureau" title="CFPB"></a>
+	<a href="/" class="noStyles"><img id="logo" src="/static/nemo/_/img/logo.png"
```
Scotchester: The `` tag should probably be started on its own line

cfgov/jinja2/v1/_layouts/base-common.html
```diff
@@ -19,7 +19,6 @@
 -->
 {% endif %}
 
-<!--[if lt IE 7]>      <html lang="es" class="no-js lt-ie9 lt-ie8 lt-ie7"> <![endif]-->
 <!--[if IE 7]>         <html lang="es" class="no-js lt-ie9 lt-ie8"> <![endif]-->
```
Scotchester: While you're here, please fix the `lang` attribute here to have a value of `en`.

cfgov/jinja2/v1/_layouts/base-common.html
anselmbradford: Good catch! Fixed in https://github.com/cfpb/cfgov-refresh/pull/3902/commits/990568f0bc51964c10315a204dc99868de7f4dcc

anselmbradford:
@Scotchester ready for review again!

cfgov/jinja2/v1/_layouts/base-common.html
```diff
@@ -19,8 +19,7 @@
 -->
 {% endif %}
 
-<!--[if lt IE 7]>      <html lang="es" class="no-js lt-ie9 lt-ie8 lt-ie7"> <![endif]-->
-<!--[if IE 7]>         <html lang="es" class="no-js lt-ie9 lt-ie8"> <![endif]-->
+<!--[if IE 7]>         <html lang="en" class="no-js lt-ie9 lt-ie8"> <![endif]-->
```
sebworks: Do we still support IE7?

anselmbradford: The `lt-ie8` class is still in use in several places so I was hesitant to strip this out, however, perhaps those could be stripped out too.

PR #3905, opened 2018-03-21
Completes footer code coverage

anselmbradford:
## Additions

- Adds Footer-spec.js for full 100% Footer organism JS code coverage unit testing.

## Testing

1. `gulp test:unit` should show 100% for Footer.js.


test/unit_tests/js/organisms/Footer-spec.js
```diff
@@ -0,0 +1,22 @@
+const BASE_JS_PATH = '../../../../cfgov/unprocessed/js/';
```
Scotchester: Is it possible for these paths to be root-relative so that we could set `BASE_JS_PATH` once for the entire project?

anselmbradford: I haven't gotten a chance to deeply look into it, but we could probably use a solution in here https://gist.github.com/branneman/8048520. Let me know if any of those strike you as better than another.

PR #3909, opened 2018-03-21
Regulations 3000: Add initial version of RegDown

willbarton:
This PR adds our Markdown extension RegDown. RegDown is in `regulations3k.regdown`, and there is a convenience function, `regulations3k.regdown.regdown(text)` that will process text as Markdown with our RegDown extension.

RegDown currently supports the following extensions to Markdown:

- `{label} paragraph text`: labeled paragraphs, which will generate `<p id="label">`. If no label is given, a paragraph will get an SHA3 hash of its contents as its `id`.

Also added are the following:

- `regulations3k.models.RegDownTextField` as a subclass of `django.db.models.TextField`
- `regulations3k.forms.RegDownField` as a subclass of `django.forms.CharField`
- `regulations3k.forms.RegDownTextarea` as a subclass of `django.forms.Textarea`

These currently only wrap their Django superclasses, but are intended as stubs to let us flesh out admin functionality iteratively.

`regdown` is also made available as a Jinja2 filter, so that `{{ "Hello, world" | regdown }}` can be used in templates. Our `regulations3k/jinja2/regulations3k/base.html` now includes an example of this.

I have used the tox setup in #3894 to ensure Python3 compatibility. I cannot currently test against Django 1.11 with that setup because the `v1` package imports incorrectly in Django 1.11.

## Checklist


contolini:
Thinking long term: Would it make sense to create a standalone repo for RegDown and package it as a [third party extension](https://github.com/Python-Markdown/markdown/wiki/Third-Party-Extensions)? That repo could also explain the spec.

willbarton:
@contolini that's not a bad idea, and one I've had. I'm not sure it makes sense *quite yet*, but I've tried to keep the RegDown code self-contained so that we can do that in the future.

willbarton:
I think I've found a problem... feeding `regdown()` any other block elements (like lists) get completely clobbered. I *think* returning `True` here (like the [Markdown `ParagraphProcessor`](https://github.com/Python-Markdown/markdown/blob/master/markdown/blockprocessors.py#L541-L542)) might be *a bit* too greedy.

cfgov/regulations3k/regdown.py
```diff
@@ -18,9 +18,8 @@ class RegulationsExtension(Extension):
 
     def extendMarkdown(self, md, md_globals):
         md.registerExtension(self)
-        md.parser.blockprocessors.add('labeledparagraph',
-                                      LabeledParagraphProcessor(md.parser),
-                                      '_begin')
+        md.parser.blockprocessors['paragraph'] = \
+            LabeledParagraphProcessor(md.parser)
```
willbarton: This change should fix the greediness of our `LabeledParagraphProcessor` by simply replacing the existing `paragraph` processor while maintaining its location in the `OrderedDict` that is `blockprocessors`.

PR #3911, opened 2018-03-21
Fix linter warnings

anselmbradford:
## Changes

- Fixes some JS linter warnings.

### DIFF
--- a/cfgov/unprocessed/js/modules/BreakpointHandler.js
+++ b/cfgov/unprocessed/js/modules/BreakpointHandler.js
@@ -18,15 +18,15 @@ import { get as getBreakpointState } from './util/breakpoint-state';
  * }
  */
 function BreakpointHandler( opts ) {
-  const hasRequiredArgs =
-  Boolean( opts && opts.breakpoint && opts.enter && opts.leave );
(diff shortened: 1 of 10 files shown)


sebworks:
let's turn off the linter warning on Travis.  I think you just want to see the errors on Travis.

anselmbradford:
> let's turn off the linter warning on Travis. I think you just want to see the errors on Travis.

Sounds good, but where would we do that? ESLint relies on .eslintrc on a per-directory basis, since there's already one in the home directory, we'd need the Travis config to edit that one somehow, or add a config to package.json, but only for Travis.

sebworks:
So gulp scripts ` gulp:scripts --quiet` runs just the errors. We could do it here: https://github.com/cfpb/cfgov-refresh/blob/3bac3b585331d48890a9b980586480cee1c0baed/gulp/tasks/lint.js#L18

jimmynotjim:
Have we ever looked into having travis measure the increase or decrease in warnings? If they're not being run in travis I fear they're just going to keep getting worse.

Scotchester:
I like the idea of making Travis fail if the number of warnings goes up.

PR #3912, opened 2018-03-21
Adjusts tests to be approximate

anselmbradford:
## Changes

- Changes tests that rely on run-speed to use `toBeCloseTo` instead of `toBe`

## Testing

1. `gulp test:unit --specs=js/modules/util/email-popup-helpers-spec.js` should pass.


anselmbradford:
Also removes brittle test that causes https://travis-ci.org/cfpb/cfgov-refresh/jobs/356468035

PR #3913, opened 2018-03-21
Increase decimal range to make toBeCloseTo less brittle

anselmbradford:
Compared values are timestamps that were divided by 100 so toBeCloseTo can be used to make them less brittle since we only care if they are way off. Default for toBeCloseTo is to use two decimal places, so the timestamps actually need to be divided by 10000 to create four decimal places that'll drop the last two digits. (Alternately, the signature of `toBeCloseTo` could be passed `0` as the second parameter, but this allows us to keep the default behavior.)

## Changes

- Increase decimal range to make toBeCloseTo less brittle

## Testing

1. `gulp test:unit --specs=js/modules/util/email-popup-helpers-spec.js`


sebworks:
We could always subtract them as well and then check the diff.

anselmbradford:
True, I'll update to that if this starts failing. Could be clearer too.

PR #3917, opened 2018-03-21
Fix empty related posts bug 

richaagarwal:
It is unclear what the original intention of this code was, but currently it breaks if you hit this block more than once (which would happen if you have two or more empty related posts blocks) by causing the error `maximum recursion depth exceeded in cmp`

See discussion in GHE/platform/issues/2576

UPDATE: I haven't been able to figure out what `loop()` is supposed to be doing -- there's no real jinja documentation on it, and when I play around with it locally, it seems to expect a Wagtail block-like object (if I pass in an array or dictionary, it complains that it doesn't have the attribute 'block', 'meta', or 'value'). 

I also found this same logic in [`sidebar-breakout.html`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/sidebar-breakout.html#L97) so before I proceed with this PR & removing it there as well, I'd like to better understand what it's supposed to be doing.

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/sidebar-breakout.html
+++ b/cfgov/jinja2/v1/_includes/organisms/sidebar-breakout.html
@@ -52,7 +52,7 @@
                           block
                           block__flush-top
                           block__flush-bottom">
-                {% for block in page.sidebar_breakout recursive %}
+                {% for block in page.sidebar_breakout %}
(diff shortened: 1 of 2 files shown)


richaagarwal:
I removed this logic from `sidebar-breakout.html` as well.

PR #3922, opened 2018-03-22
Email popup coverage + email popup path fixes

anselmbradford:
## Additions

- Adds email popup unit test spec.
- Adds add-email-popup.js for shared popup initialization code. Also adds test for that module.

## Changes

- Changes the email popup so that it isn't reaching outside of its module for the language and instead the language is set on the popup organism itself.
- Adds `getDom` method to email popup for retrieving the base element.
- Adds dom check to email popup to throw error if base class isn't found.
- Moves BASE_CLASS to a static property on the email popup constructor.
- Fixes email popup paths where popup was not showing on OAH paths.
- Removes visible class from email popup on initialization, if present.

## Testing

1. `gulp clean && gulp build`
2. Popup should show on scroll on the following URLs:

```
/ask-cfpb/what-is-a-statute-of-limitations-on-a-debt-en-1389/
/ask-cfpb/what-is-the-best-way-to-negotiate-a-settlement-with-a-debt-collector-en-1447/
/ask-cfpb/what-should-i-do-when-a-debt-collector-contacts-me-en-1695/
/consumer-tools/debt-collection/
/owning-a-home/
/owning-a-home/mortgage-estimate/
```

Between a visit of each, go into Dev tools > Application > Local Storage and clear all the keys for the popups:

screen shot 2018-03-22 at 6 05 22 pm


cfgov/cfgov/settings/base.py
```diff
@@ -677,14 +677,18 @@
...
-        '/ask-cfpb/what-is-a-statute-of-limitations-on-a-debt-en-1389/',
-        '/ask-cfpb/what-is-the-best-way-to-negotiate-a-settlement-with-a-debt-collector-en-1447/',
-        '/ask-cfpb/what-should-i-do-when-a-debt-collector-contacts-me-en-1695/',
-        '/consumer-tools/debt-collection/',
-    ],
-    'oah': [
-        '/owning-a-home/',
-        '/owning-a-home/mortgage-estimate/',
-    ],
+    'debt': {
```
chosak: Did you consider adding the language setting to the popup-specific config in the template files [here](https://github.com/cfpb/cfgov-refresh/tree/master/cfgov/jinja2/v1/_includes/organisms/email-popup)? For example there could be a new `popup_language` setting defined in the template [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/email-popup/oah.html).

This works well but it feels a little bit complicated to have the popup URLs grouped under language like this, since language is just one configuration option that we want to pass in. If there's some additional option in future that we also need to pass, it might get difficult to do that here.

cfgov/v1/tests/templatetags/test_email_popup.py
```diff
@@ -27,32 +26,32 @@ def test_no_popups_configured_returns_blank(self):
...
         response = self.render('/page/not/configured/')
         self.assertEqual(response, '')
 
     @override_settings(
-        EMAIL_POPUP_URLS={'oah': ['/page/configured/']},
-        FLAGS={'EMAIL_POPUP_OAH': {'boolean': True}} 
+        EMAIL_POPUP_URLS={'oah': {'en': ['/page/configured/']}},
+        FLAGS={'EMAIL_POPUP_OAH': {'boolean': True}}
     )
     def test_popup_configured_and_flag_true_and_path_matches(self):
```
chosak: It'd be nice to add an additional unit test like this one that confirms that the language attribute `lang="something"` makes it into the template.

cfgov/cfgov/settings/base.py
anselmbradford: I like that much better. I didn't know it was possible. Fixed in https://github.com/cfpb/cfgov-refresh/pull/3922/commits/876592dab427f5ba0f56a76eee8a75be119f43ca

cfgov/v1/tests/templatetags/test_email_popup.py
anselmbradford: Sure, what would a test like this look like?

chosak: I'm no longer convinced an extra test is needed now that you've removed the language configuration from the settings dictionary. If you were using the dictionary (as before 876592d), it'd be nice to confirm that the dict key was being used and getting into the template. As it is now, the test would just be verifying that template content in one file (like [here](https://github.com/cfpb/cfgov-refresh/pull/3922/files#diff-afca900319cd83827ac48ee8eacf7498R3)) shows up when being rendered in another file ([here](https://github.com/cfpb/cfgov-refresh/pull/3922/files#diff-fafa50303a4ed1aaa70f935f5e4e29d1R15)). I don't think that adds much value.

PR #3924, opened 2018-03-23
Adjust footer-button spec

anselmbradford:
## Changes

- Adjust footer-button to immediately scroll to top if requestAnimationFrame is not supported, instead of silently doing nothing.
- Adjust footer-button spec to provide full branch coverage.

## Testing

1. `gulp test:unit --specs=js/modules/footer-button-spec.js` should pass with full coverage.

## Notes

- If this still fails on Travis periodically we could increase https://github.com/cfpb/cfgov-refresh/compare/fix_button_spec?expand=1#diff-c2f27422225b9f5a39bc1f4247dc96a7R54


cfgov/unprocessed/js/modules/footer-button.js
```diff
@@ -5,19 +5,14 @@
    scrolltop-animation-without-jquery
    ========================================================================== */
 
-
 // Required modules.
 const behavior = require( './util/behavior' );
```
sebworks: Switch to import statement.

anselmbradford: Fixed!

sebworks:
Merge when ready.

jimmynotjim:
At some point we should update the button to use the [broader scrollTo utility](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/js/modules/util/scroll.js) and remove these two files.

PR #3927, opened 2018-03-23
owning-a-home/rate-checker: spins out slider and adds tests

anselmbradford:
## Changes

- Moves range slider JS to its own class.
- Adds range slider tests.
- Moves range slider CSS to its own file.
- Uses range slider without jquery dependency.
- Removes range slider handlebars template.

## Testing

1. `npm install`.
1. `gulp clean && gulp build` and credit score range on /owning-a-home/explore-rates/ should work as before.
1. `gulp test:unit` should pass and show 98.25% coverage for range slider.

## Screenshot
screen shot 2018-04-09 at 4 09 58 pm


cfgov/unprocessed/apps/owning-a-home/js/explore-rates/Slider.js
```diff
@@ -0,0 +1,113 @@
+// Required modules.
+const atomicHelpers = require( '../../../../js/modules/util/atomic-helpers' );
+const standardType = require( '../../../../js/modules/util/standard-type' );
+
+const $ = require( 'jquery' );
```
sebworks: use import here

anselmbradford: What would this look like? Try changing it in `rate-checker.js` and running `gulp build`. Do you get an error in the dev console? If not what's the import statement you used look like?

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/Slider.js
```diff
@@ -0,0 +1,113 @@
...
+      }
+    } );
+  }
+
+  function _update() {
+    const handle = _dom.parentNode.querySelector( '.rangeslider__handle' );
+    const leftVal = +Number(
+      handle.style.left = handle.style.left.replace( 'px', '' )
+    );
+    _min = domValues.getSelection( 'credit-score' );
```
sebworks: If we are going to make this atomic then lets make it generic, this way we can reuse it. You should be able to pass in the min / max.

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/Slider.js
```diff
@@ -0,0 +1,113 @@
...
+   */
+  function getMax() {
+    return _max;
+  }
+
+  /**
+   * Initialize the range slider. http://andreruffert.github.io/rangeslider.js/
+   */
+  function render() {
+    $( '#' + element.id ).rangeslider( {
```
sebworks: I would check if there is a range slider which doesn't include jQuery

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/Slider.js
```diff
@@ -0,0 +1,113 @@
...
+   * @returns {number} Minimum value of the range.
+   */
+  function getMin() {
+    return _min;
+  }
+
+  /**
+   * @returns {number} Maximum value of the range.
+   */
+  function getMax() {
```
sebworks: I don't think using getters / setters makes sense unless you are performing some type of validation or property modification.  It's just more code and it really doesn't protect you. I say, let's convert this to a class and just use the `_` convention.

anselmbradford: How about this? https://github.com/cfpb/cfgov-refresh/pull/3927/commits/31351f78bb3e96d5c938f6cda06131c27dbe1470

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/Slider.js
anselmbradford: Hm, I'll be. Lookie lookie https://github.com/stbaer/rangeslider-js. I'll give it a try.

anselmbradford: Updated in https://github.com/cfpb/cfgov-refresh/pull/3927/commits/3fed220c49f4d482c67bbc0dccc61bf4e953fa3e

anselmbradford: Err, hang on. I needed to `gulp clean` and now find it's not working as expected.

sebworks:
@anselmbradford, the slider isn't working for me.

anselmbradford:
> @sebworks 

Yeah, sorry, [this comment](https://github.com/cfpb/cfgov-refresh/pull/3927#discussion_r177602217) got lost. Still working on it.

anselmbradford:
@sebworks This has been re-written and is ready for review again. Please re-pull the branch down. Thanks!!

cfgov/unprocessed/apps/owning-a-home/css/a-range.less
```diff
@@ -0,0 +1,115 @@
+/*
+    Range slider for credit score
+    --------------------------- */
+.a-range {
```
sebworks: not sure if this should be a molecule, instead of an atom

anselmbradford: I was going off of selects and inputs being atoms https://github.com/cfpb/capital-framework/tree/canary/src/cf-forms/src/atoms

anselmbradford: You may be right though with the slider label, since that brings it more in line with https://github.com/cfpb/capital-framework/blob/canary/src/cf-forms/src/molecules/form-fields.less

anselmbradford: Thinking about it further, I think it would be an atom. The label on the slider is part of the slider, as opposed to a form field which has an input and label that are separate entities. I'll merge this as an atom, but we can revisit no doubt if/when this moves to CF or into a more general component within cfgov-refresh.

PR #3933, opened 2018-03-28
owning-a-home/explore-rates/util - Adds test coverage for util functions

anselmbradford:
## Changes

- Merges `format-timestamp.js` into new `explore-rates/util.js` module.
- Adds ellipsis to template for before date stamp renders.
- Moves `renderTime` function to `explore-rates/util.js` module as `renderDatestamp` and removes jquery.
-  Moves `renderLoanAmount` function to `explore-rates/util.js` module as `calcLoanAmount` and `renderLoanAmount`.
- Moves `renderAccessibleData` function to `explore-rates/util.js` module.
- Adds unit test coverage for rate explorer util functions.

## Testing

1. `gulp clean && gulp build` and check the rate checker chart date stamp and loan calculator. Also, check `accessible-data` id table structure against production.
2. `gulp test:unit --specs=apps/owning-a-home/js/explore-rates/util-spec.js` should pass.

## Screenshots

Location of data stamp:
screen shot 2018-03-27 at 8 17 42 pm


cfgov/unprocessed/apps/owning-a-home/js/explore-rates/rate-checker.js
```diff
@@ -989,10 +977,12 @@ function init() {
     return false;
   }
 
+  // Record timestamp HTML element that's updated from date from API.
+  timeStampDom = document.querySelector( '#timestamp' );
+
   renderSlider();
   renderChart();
   renderLoanAmount();
-  renderTime();
```
anselmbradford: This doesn't appear to be needed. Here's the original function:

```
function renderTime( time ) {
  if ( time ) {
    time = formatTime( time );
  }

  $( '#timestamp' ).text( time );
}
```

So if that's called without the `time` argument. The timestamp element would render to `undefined`. Instead I added a default ellipsis to the template and updated the function to:

```
function renderDatestamp( elem, time ) {
  elem.textContent = formatTimestampMMddyyyy( time );
}
```

PR #3937, opened 2018-03-28
Remove unused ansible directory

richaagarwal:
I don't think we use this directory anymore, and it seems like it's not worth us maintaining these files when there are changes to the setup.  I would be curious if anyone is using this, but otherwise I think we should remove. 

A note for after this is (potentially) merged: When removing directories, sometimes you still end up with an empty directory locally.  To resolve that, you can run `git clean -fd` which will remove those for you. Alternatively, you can run `git clean -fdn` for a dry-run to see what would be removed.


chosak:
There are some additional files related to this directory, which is part of the Vagrant implementation for this repository:

- The repo [Vagrantfile](https://github.com/cfpb/cfgov-refresh/blob/master/Vagrantfile) makes use of the Ansible playbook.
- [docs/index.md](https://github.com/cfpb/cfgov-refresh/blob/master/docs/index.md) and [docs/usage.md](https://github.com/cfpb/cfgov-refresh/blob/master/docs/usage.md) both mention the Vagrant setup. These references should be removed if we're deprecating Vagrant entirely.
- The [`bin/cfgov`](https://github.com/cfpb/cfgov-refresh/blob/master/bin/cfgov) and [`bin/cfgovd`](https://github.com/cfpb/cfgov-refresh/blob/master/bin/cfgovd) scripts are included as part of the Vagrant setup described in the documentation.

If we're going to completely deprecate Vagrant/Ansible for this repository, then we should clean it all up at once.

I'd be okay with removing all of that from this repo as our internal Ansible repository lets you run this codebase using a Vagrant box, and the current code is likely out of date and won't be updated to include future improvements (like Postgres). It'd be nice if in future we could re-open source as much of the internal Ansible/Vagrant process as possible.

anselmbradford:
Covered in https://github.com/cfpb/cfgov-refresh/pull/4107

PR #3939, opened 2018-03-28
Remove tests that rely on how NULLs are sorted 

richaagarwal:
These tests promise functionality that is specific to MySQL.  With our impending move to PostgreSQL, the behavior in these two tests will no longer be the case (sorting NULLs first).
However, we don't actually rely on this ordering anywhere, so we can just remove the tests and not promise this behavior.

See platform issue 2569 for discussion on this. 

Note that the migration is because we updated the help text on the `order` field.

### DIFF
--- a/cfgov/v1/models/snippets.py
+++ b/cfgov/v1/models/snippets.py
@@ -132,8 +132,7 @@ class Resource(ClusterableModel):
         blank=True,
         help_text='Snippets will be listed alphabetically by title in a '
         'Snippet List module, unless any in the list have a number in this '
-        'field; those with an order value will appear at the bottom of the '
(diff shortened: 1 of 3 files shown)


cfgov/v1/models/snippets.py
```diff
@@ -132,8 +132,7 @@ class Resource(ClusterableModel):
         blank=True,
         help_text='Snippets will be listed alphabetically by title in a '
         'Snippet List module, unless any in the list have a number in this '
-        'field; those with an order value will appear at the bottom of the '
-        'list, in ascending order.'
+        'field; those with an order value will appear in ascending order. '
```
chosak: Extra space at the end of this string.

PR #3948, opened 2018-04-02
Remove unused analytics script

anselmbradford:
This script used to reference `span` tags, which don't support the on-change event, meaning it was non-functional. Further investigation shows this tag is not needed at all.

## Removals

- enterprise-dropdown-on-change.js


anselmbradford:
@kelleyholden

PR #3949, opened 2018-04-02
owning-a-home/rate-checker: Inlines small templates

anselmbradford:
Small alert templates were being added/removed from the dom via handlebars, but didn't have any variables that utilized what handlebars is best at. Probably best to always include them in the markup and set their visibility through the CSS class `u-hidden`. This makes the size of the HTML page larger slightly, but removes the handlebars dependency on those templates.

## Removals

- credit, result, fail, and dps alerts templates.

## Changes

- Removes jquery lookups for the alert markup.

## Testing

1. `gulp clean && gulp build`
2. When the credit score slider is brought to 600 there should be an alert on the chart and below the slider.
3. When the down payment is greater than the house price, there should be an alert that appears below the loan amount.


cfgov/unprocessed/apps/owning-a-home/js/explore-rates/rate-checker.js
```diff
@@ -16,18 +16,22 @@ import median from 'median';
...
 import { applyThemeTo } from './highcharts-theme';
 import { getSelection } from './dom-values';
-import { uniquePrimitives } from '../../../../js/modules/util/array-helpers';
 import {
   calcLoanAmount,
   renderAccessibleData,
   renderDatestamp,
   renderLoanAmount
 } from './util';
+import { uniquePrimitives } from '../../../../js/modules/util/array-helpers';
```
anselmbradford: This was a consequence of a rebase, but I suppose as a "u" it fits better here. They're all out of alphabetical anyway though.

PR #3957, opened 2018-04-03
Adds analytics util script

anselmbradford:
There is lots of redundant utility code in the analytics scripts. This moves some of that functionality to a utility module. It also makes it so that the console logs in the scripts only show up when the `debug-gtm=true` query string appears in the URL. 

The `track` utility script can likely be consolidated with `Analytics.js` in the future (which should perhaps move into the `analytics-gtm` app?

## Additions

- Add `analytics-util.js` and initial test spec.

## Changes

- Moves `delay` and `track` functions to the utility script.
- Adds helper function to check that an element exists before adding an event listener.
- Renames `bottom` variable `bottomPos` to fix New Relic error.
- Makes script that push to the dataLayer utilize the `track` function.

## Testing

To see that the logging function works, add this to `common.js` and run `gulp clean && gulp build`

```js
// common.js
function getQueryParameter( key ) {
  const url = window.location.href;
  const param = key.replace( /[\[\]]/g, '\\$&' );
  const regex = new RegExp( '[?&]' + param + '(=([^&#]*)|&|#|$)' );
  const results = regex.exec( url );
  if ( !results ) return null;
  if ( !results[2] ) return '';
  const decoded = decodeURIComponent( results[2].replace( /\+/g, ' ' ) );
  if ( decoded === 'true' ) return true;
  if ( decoded === 'false' ) return false;
  return decoded;
}

if ( getQueryParameter( 'debug-gtm' ) === true ) {
  console.log( `ANALYTICS DEBUG MODE:` );
}
```

Then visit http://localhost:8000/?debug-gtm=true and see the log in the dev console. Remove the query string and reload to see the log disappear.

### DIFF
--- a/cfgov/unprocessed/apps/analytics-gtm/js/util/analytics-util.js
+++ b/cfgov/unprocessed/apps/analytics-gtm/js/util/analytics-util.js
@@ -1,4 +1,3 @@
-
 /**
  * Check if an element exists on the page, and if it does, add listeners.
  * @param {[type]}   elem     [description]
@@ -35,7 +34,7 @@ function track( event, action, label ) {
     label: label
   } );
   analyticsLog( event, action, label );
-};
+}
 
 /**
  * Check if two hosts are the same.
--- a/cfgov/unprocessed/apps/analytics-gtm/js/util/analytics-util.js
+++ b/cfgov/unprocessed/apps/analytics-gtm/js/util/analytics-util.js
@@ -1,0 +1,102 @@
+
+/**
+ * Check if an element exists on the page, and if it does, add listeners.
+ * @param {[type]}   elem     [description]
+ * @param {[type]}   event    [description]
+ * @param {Function} callback [description]
+ */
+function addEventListenerToElem( elem, event, callback ) {
+  if ( elem ) {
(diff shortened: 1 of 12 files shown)


cfgov/unprocessed/apps/analytics-gtm/js/util/analytics-util.js
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Check if two hosts are the same.
+ * This only works with a www subdomain.
+ * @param {string} host1 - A URL.
+ * @param {string} host2 - Another URL.
+ * @returns {boolean} True if the hosts are equal, false otherwise.
+ */
+function hostsAreEqual( host1, host2 ) {
+  let h1 = document.createElement( 'a' );
```
sebworks: I would use an inner function so we don't have got repeat the code

anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/pull/3957/commits/dafdc3f8a9c79c16892698ca535e01651dcb24ce

PR #3961, opened 2018-04-04
Updates scroll tracking analytics

anselmbradford:
Scroll tracking analytics had some deprecated code https://stackoverflow.com/questions/19635188/why-is-body-scrolltop-deprecated

It also could be condensed to reuse code via helper functions.

Additionally, it tracked total page height instead of a percentage of hidden page height, which would be the area that scrolls.

## Changes

- Consolidated code to allow function to track any percentage of page scrolling area.
- Changes reporting to show number of seconds since page load, not since start of scrolling.

## Testing

1. `gulp clean && gulp build`
2. Visit http://localhost:8000/about-us/blog/?debug-gtm=true and check the dev console and scroll.

## Screenshots

screen shot 2018-04-04 at 5 19 58 pm

## Note:

Start of scroll is not currently reported. Should it be, @kelleyholden ?

### DIFF
--- a/cfgov/unprocessed/apps/analytics-gtm/js/event-listener-scroll-tracking.js
+++ b/cfgov/unprocessed/apps/analytics-gtm/js/event-listener-scroll-tracking.js
@@ -31,7 +31,7 @@ const startTime = new Date();
 const timeAtPageLoad = startTime.getTime();
 
 /**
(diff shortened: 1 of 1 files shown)


sebworks:
Looks good, I would give https://developers.google.com/web/updates/2016/06/passive-event-listeners a look though.

PR #3966, opened 2018-04-05
Make ask_cfpb tests compatible with PostgreSQL

chosak:
This change fixes two `ask_cfpb` Python unit tests that were failing when being run against a Postgres database. In both cases this seems to be due to differences in how predictable model IDs are when generated by Model Mommy - when using SQLite this seems to reliably start at 1 but with Postgres that isn't predictable.

At a high level these tests were failing because they were testing an assumed model ID instead of explicitly setting it in the test.

@richaagarwal and I paired on this; we broke up `AnswerModelTestCase.test_facet_map` into three smaller tests that more explicitly test a single piece of functionality and rely less on common setup being done in the test case `setUp`.

## Testing

To validate these changes against a local Postgres DB:

```sh
$ DATABASE_URL=postgres://user@localhost/db tox -e fast
```

## Checklist

### DIFF
--- a/cfgov/ask_cfpb/tests/models/test_pages.py
+++ b/cfgov/ask_cfpb/tests/models/test_pages.py
@@ -270,19 +270,70 @@ class AnswerModelTestCase(TestCase):
             page.get_template(HttpRequest()),
             'ask-cfpb/answer-page.html')
 
-    def test_facet_map(self):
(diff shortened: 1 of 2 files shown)


richaagarwal:
Thanks for the review @higs4281!

PR #3968, opened 2018-04-05
Update watch and clean front-end gulp tasks

contolini:
JS and Less files were being ignored for projects that live in `/unprocessed/apps/`.

## Changes

- Gulp's watch task now watches `paths.unprocessed + '/apps/**/[js|css]/**/*.[js|css]'`
- Gulp's clean task now cleans `paths.unprocessed + '/apps/**/[js|css]/**/*.[js|css]'`

## Testing

Clean task

1. `gulp` to build everything.
1. `/cf_gov/static_built/apps/regulations3k/css/` and `/cf_gov/static_built/apps/regulations3k/js/` should contain some files.
1. `gulp clean:css` and `gulp clean:js` should delete those files.

Watch task

1. `gulp watch`
1. Edit `/cfgov/unprocessed/apps/regulations3k/js/module1.js` and see if it recompiles the JS.
1. Edit `/cfgov/unprocessed/apps/regulations3k/css/layout.less` and see if it recompiles the CSS.


gulp/tasks/clean.js
```diff
@@ -1,12 +1,21 @@
 const configClean = require( '../config' ).clean;
```
anselmbradford: Let's get rid of the references to config altogether and remove the `clean` entry from `config.js`. All it is is a obfuscating shell over paths. `paths.processed + '/css/**/*'` is clearer than `configClean.css + '/**/*'` in my opinion. The only place for config settings is when they are used in multiple places.

contolini: Sounds great to me. Frequently tabbing over to the config file to look up paths is tedious.

gulp/tasks/clean.js
```diff
@@ -1,12 +1,20 @@
...
 const del = require( 'del' );
 const gulp = require( 'gulp' );
+const paths = require( '../../config/environment' ).paths;
 
 // Clean CSS out of /cfgov/static_built/css/
-gulp.task( 'clean:css', () => del( configClean.css + '/**/*' ) );
+gulp.task( 'clean:css', cb => {
+  del( paths.processed + '/css/**/*' );
+  del( paths.processed + '/apps/**/css/*' );
+  cb();
```
anselmbradford: You probably need `return cb()`. Does `gulp lint --quiet` complain about this?

contolini: No complaints from my machine. Although now as I look at this, `del` is async so there's a race condition here. I'll refactor.

gulp/tasks/watch.js
```diff
@@ -3,10 +3,18 @@
 
 const gulp = require( 'gulp' );
 const config = require( '../config' );
```
anselmbradford: Let's nix the config here too. I mean look at `config.legacy.cwd` all it resolves to is `paths.legacy`!

anselmbradford:
@contolini Looks like I really broke watch with gulp 4. Check it now, should be working. We do have a `gulp scripts:apps` command that we could use to only run when app JS (or the equivalent for CSS) is changed, if desired.

PR #3970, opened 2018-04-06
Upgrading Webpack to v4.5, enable caching via the uglifyOptions, and modifying output

sebworks:
Upgrading Webpack to v4.5, enable caching via the uglifyJS Plugin, and modifying Webpack output

## Changes

- Upgraded Webpack to version 4.5
- Modified various Webpack configs to enable caching UglifyJS files and modify console output

## Testing

1. Run gulp build and verify that the output no longer contains entrypoints.
    Example: `Entrypoint Expandable.js = Expandable.js`
2. Run `open node_modules/.cache/uglifyjs-webpack-plugin` and verify that the folder exists.

## Notes

-  We have a couple layers of caching / process avoidance going on, so we will need to pay attention to ensure the caches are being expired properly. 

## Checklist

### Browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


config/webpack-config.js
```diff
@@ -60,27 +60,36 @@ const COMMON_CHUNK_CONFIG = new webpack.optimize.SplitChunksPlugin( {
   name: COMMON_BUNDLE_NAME
 } );
 
+const STATS_CONFIG = {
+  stats: {
+    entrypoints: false
```
anselmbradford: This is such a short config, what do you think about putting the `false` value in a constant and then putting the rest in each webpack object? This would make the object spread unnecessary and simplify the webpack configs that aren't using multiple copies of `STATS_CONFIG`.

.eslintrc
```diff
@@ -9,7 +9,7 @@ parserOptions:
     globalReturn: false
     impliedStrict: true
     jsx: true
-    experimentalObjectRestSpread: false
+    experimentalObjectRestSpread: true
```
anselmbradford: Does this need an update in https://github.com/cfpb/development/blob/master/.eslintrc#L12 (and CF and generator-cf)? This is to avoid usage of `Object.assign`?

cfgov/unprocessed/apps/analytics-gtm/webpack-config.js
```diff
@@ -37,14 +38,19 @@ const COMMON_MODULE_CONFIG = {
         presets: [ [ 'babel-preset-env', {
           targets: {
             browsers: BROWSER_LIST.LAST_2_IE_9_UP
-          },
-          debug: true
```
anselmbradford: I'd be inclined to leave this as `debug: false` so it's easier to find where to turn it on, like is done with the uglify config. Or do you think that's too much clutter?

config/webpack-config.js
```diff
@@ -111,18 +121,21 @@ const appsConf = {
...
   module: COMMON_MODULE_CONFIG,
   mode: 'production',
   output: {
     filename: 'spanish.js'
   },
   plugins: [
     COMMON_UGLIFY_CONFIG
-  ]
+  ],
+  ...STATS_CONFIG
```
anselmbradford: Since this is the only webpack config where the `STATS_CONFIG` is used multiple times, you could modify the loop at the end of the file:

```js
let currConfig;
for ( const key in configExports ) {
  currConfig = configExports[key];
  currConfig.stats = STATS_CONFIG.stats; // Or use Object.assign(…)
  if ( NODE_ENV === 'development' ) {
    // eslint-disable-next-line guard-for-in
    Object.assign( currConfig, devConf );
  }
}
```

That would render the need for the turning on the experimental flag unnecessary.

Or alternatively, the object spread could be used for mixing in the dev config as well.

config/webpack-config.js
sebworks: I like it as an separate object because it can be extended when debugging and you don't have to update things in multiple places.  This is the entire list of stat options: https://webpack.js.org/configuration/stats/#stats

.eslintrc
sebworks: It's my understanding that object rest spread has reached stage 4 per: https://github.com/tc39/proposal-object-rest-spread.  I think its easier to read and less code.

config/webpack-config.js
sebworks: I don't think we have to use object spread or Object.assign uniformly in all cases.  I think the code reads wells as it is.

config/webpack-config.js
anselmbradford: Ah gotcha, I didn't realize there were so many settings in there. One thing I'm confused on is https://webpack.js.org/configuration/stats/#stats shows `entrypoints: false` like that's the default value. What's the default state of the stats object in the config? Is the default that it's absent?

.eslintrc
anselmbradford: Works for me! We should adjust the other eslints to match I'd say, so https://github.com/cfpb/development/blob/master/.eslintrc#L12 and https://github.com/cfpb/capital-framework/blob/canary/.eslintrc#L12

cfgov/unprocessed/apps/analytics-gtm/webpack-config.js
sebworks: fixed in https://github.com/cfpb/cfgov-refresh/pull/3970/commits/cc2efa06e6dbd3b47a7078f290cf89ead46dffa4.

config/webpack-config.js
sebworks: yeah, I'm not sure what the default values are and the string value doesn't work with Webpack stream. https://github.com/shama/webpack-stream/issues/163.

PR #3974, opened 2018-04-06
Remove reference to deprecated Youtube Flash Player

anselmbradford:
[Youtube Flash player has been deprecated since 2015](https://developers.google.com/youtube/player_parameters). This PR removes reference to the player.

## Changes

- Code comment cleanup.

## Removals

- Removes code in youtube enterprise analytics that converts old Youtube Flash Player to iframe.

## Testing

1. Open https://www.consumerfinance.gov/practitioner-resources/library-resources/ in IE8 Sauce Labs and note that clicking video opens it in a new window to youtube.com. So existing behavior is not to convert the embedded player.


anselmbradford:
cc @kelleyholden @lfatty

sebworks:
shackwave

PR #3994, opened 2018-04-11
Bugfix: Adds missing closing span in housing counselor

anselmbradford:
## Additions

- Adds missing closing span tag to website field in find-a-housing-counselor.

## Testing

1. There shouldn't be warnings in IE11 for "unmatched end tag" for /find-a-housing-counselor/?zipcode=05201 or similar.


anselmbradford:
This is the full tag:

screen shot 2018-04-11 at 10 03 07 am

PR #3999, opened 2018-04-11
Remove boxsizing.htc from codebase

anselmbradford:
boxsizing.htc is only for IE6 and IE7 and has long been removed from the rest of the codebase and shouldn't be needed in the files in this PR.

## Removals

- Last remnants of `boxsizing.htc`


sebworks:
Out with the old, in with the new

PR #4009, opened 2018-04-13
Remove browserSync and gulp watch

anselmbradford:
`browserSync` is in dependencies because it's a dependency in production as part of the gulp workflow, but it is not run. browserSync also has snyk reported vulnerabilities https://snyk.io/test/npm/browser-sync. ~~This PR sets up an architecture to keep it in devDependencies.~~

## Additions
- ~~Adds `browser-sync` utility for running browser-sync dependency in non-production environments.~~

## Removals

- ~~Removes `browsersync` gulp task and merges it with the `watch` task.~~
- browser-sync dependency.
- gulp watch task.

## Changes

- Sets the `NODE_ENV` value in `cli_flags.sh`

## Testing

1. Run `./frontend.sh` to get into development state.
2. ~~Run `gulp watch` and see that it runs.~~
3. ~~Edit a CSS file (like `main.less`) and save it and see that it re-compiles.~~
4. ~~Edit a JS file (like `common.js`) and save it and see that it re-compiles.~~
5. ~~ctrl+c to quit watch.~~
1. Run `./frontend.sh production` to get into production state.
2. ~~Run `gulp watch` and see that it says `NODE_ENV` is in `production`.~~
3. Run `gulp build` and see that it passes.

## Todo

- ~~If you run `./frontend.sh production` first it modifies package-lock.json, and when `./frontend.sh` is run subsequently, the devDependencies aren't installed—requiring `npm install` be run between the two. This PR should not be merged before this is resolved.~~


gulp/tasks/copy.js
```diff
@@ -16,9 +16,7 @@ function _genericCopy( src, dest ) {
     .pipe( gulpChanged( dest ) )
     .on( 'error', handleErrors )
     .pipe( gulp.dest( dest ) )
-    .pipe( browserSync.reload( {
-      stream: true
-    } ) );
+    .pipe( runBrowserSyncApp() );
```
Scotchester: I'm confused about why this is still needed here. Doesn't having browsersync running via gulp-watch nullify the need for this (and in the scripts and style tasks, too)?

jimmynotjim: Yeah, I think you have it right Ans. I'll give this a test later today and report back.

ascott1: @jimmynotjim did you have a chance to verify this?

Scotchester: Couldn't the `reload` call happen in `watch` as part of the actions it takes when it detects a change?

jimmynotjim: I'm actually on the side of "lets remove browser sync entirely" now. When I was testing this I noticed that the entire build process would be kicked off and as it worked it's way through everything the page would refresh multiple times, making it take longer than simply running the appropriate task and refreshing manually. I think we ought to simplify the build process as much as possible, then look at adding features back in as needed.

anselmbradford: Okay, I'll update this to remove browserSync and heavily spam the FEWD channel with news of its impending merge to solicit any latent protest.

anselmbradford:
browser-sync has been removed. Ready for review again. Thanks!

package.json
```diff
@@ -97,7 +96,8 @@
     "protractor-retry": "1.1.9",
     "selenium-webdriver": "4.0.0-alpha.1",
     "snyk": "1.70.2",
-    "wcag": "0.3.0"
+    "wcag": "0.3.0",
+    "yargs": "11.0.0"
```
anselmbradford: Note I had to add this because `yargs` was removed from package-lock when browser-sync is removed, but it's needed by protractor. Not sure why protractor doesn't install it, but adding it here fixes it.

PR #4013, opened 2018-04-16
Simpler, more flexible database configuration

chosak:
This set of changes attempts to make this project's Django database configuration less MySQL-specific, making it easier to run against other databases like SQLite and PostgreSQL.

Instead of specifying the Django database through a series of environment variables like `MYSQL_HOST` and `MYSQL_USER`, we want to standardize on the use of a single `DATABASE_URL` variable. This variable is supported and documented as part of the third-party [dj-database-url](https://github.com/kennethreitz/dj-database-url) package.

This lets you swap between databases by modifying a single variable from, say, `mysql://user:pw@host/name` to `postgres://user:pw@otherhost/othername`. This makes it significantly easier to run this project against different databases and reduces the number of variables needed in our deployment environments.

The various Django settings files have been condensed based on the introduction of this variable. The new logic works like this:

1. If `DATABASE_URL` is defined, Django will use it to define the default database.
2. If not, and if `MYSQL_HOST` is defined, Django will use a MySQL database as the default. This is to maintain compatibility with the current behavior. A warning is displayed if this logic is used, however, because, as part of the site transition to Postgres, support for these specific variables will be removed in future.
3. If neither of these are set, then there is no fallback database, and most Django actions will fail with an error. The exception to this is for unit tests, as described below.
4. When running the unit tests using `settings.test`, `settings.test_nomigrations`, and `settings.test_acceptance`, special logic is used. If a default database is already defined as described above, it'll be used for the tests -- this is different from the current logic but ensures that e.g. if you're running against Postgres that the tests will run against Postgres.
5. If no default database is defined via environment variables, for example upon initial checkout of this repository, an in-memory SQLite database is used for testing. @sebworks points out that using an in-memory database is faster than using one on disk.

The requirements files have been modified slightly to treat MySQL and PostgreSQL more equally. There's now a separate `requirements/postgres.txt` file with a single `psycopg2` requirement.

The definition of a second Postgres database and a Django router to support it (introduced in #3462 / 7ebe9a8) has been removed, as the Postgres conversion project will not be utilizing side-by-side databases.

Both the Docker (`.python_env`) and non-Docker (`.env`) environment variable templates have been modified to reflect the preference for using `DATABASE_URL`. The Docker path in `setup.sh` has been modified to ensure that `.python_env_SAMPLE` is used to create an initial `.python_env` if it doesn't exist. The default Docker Compose setup doesn't contain an explicit database selection, deferring instead to the environment (as the non-Docker setup does).

I've made a few other minor tweaks to the setup scripts and documentation to remove a few things that were outdated. I changed the default database selection in `.env_SAMPLE` from MySQL to SQLite to simplify development for new users and reduce the startup process. See the modified quickstart guide in `README.md`.

This PR also modifies `refresh-data.sh` so that it can work on gzipped archives without expanding them. This reduces the disk space needed when working with database dumps.

I apologize for the length of this PR; these are all related changes and I thought it best to introduce them as a set. I'm happy to break out into smaller PRs if preferred. Note that because the legacy MySQL environment variables are still supported this will not (yet) require any changes to the deployment pipeline. Future work will be needed to standardize on the use of `DATABASE_URL` there.

## Testing

1. If you run `tox` in a new environment, it'll run against an in-memory SQLite database.
2. If you set your `DATABASE_URL` to your local database, for example `postgres://user@localhost/cfgov`, the tests will run against Postgres:

    ```sh
    $ DATABASE_URL=postgres://user@localhost/cfgov tox -r -e fast
    ```

## Screenshots

Warning displayed if running without `DATABASE_URL`:

```
$ DATABASE_URL= MYSQL_HOST=localhost cfgov/manage.py dbshell
/Users/chosaka/Projects/cfgov-refresh/cfgov/cfgov/settings/base.py:197: UserWarning: 
The ability to define your MySQL database through the use of environment
variables like MYSQL_NAME will soon be deprecated in favor of the single
DATABASE_URL environment variable.

Please modify your environment to instead use something like this:

DATABASE_URL=mysql://username:password@host/dbname

See https://github.com/kennethreitz/dj-database-url for other examples of how
to define DATABASE_URL.

  warnings.warn(MYSQL_VARIABLES_DEPRECATED)
```

## Checklist


.env_SAMPLE
```diff
@@ -25,7 +25,7 @@ export DJANGO_HTTP_PORT=8000
 #export CFGOV_PROD_DB_LOCATION=<some_database_dump_url>
 
 # Password given to 'admin' user created by initial-data.sh script.
-#export WAGTAIL_ADMIN_PW=<wagtail_admin_user_password>
+export WAGTAIL_ADMIN_PW=admin
```
chosak: I changed this to have a default `admin` value as it makes it easier to get started from scratch. With a blank admin password it's not possible to login to the Wagtail admin, so we might as well default this to something.

.env_SAMPLE
```diff
@@ -45,23 +45,22 @@ export ALLOW_ADMIN_URL=True
...
-########################
-# MySQL Database Server.
-########################
+#########################################################################
+# Django database.
+#
+# See https://github.com/kennethreitz/dj-database-url for URL formatting.
+#########################################################################
+
+export DATABASE_URL=sqlite:///db.sqlite3
```
chosak: Default value when using `.env` is SQLite because that makes it easier for new users of this repo to get up and running without having to set up MySQL or Postgres.

.python_env_SAMPLE
```diff
@@ -8,3 +8,22 @@
...
+#
+# See https://github.com/kennethreitz/dj-database-url for URL formatting.
+#########################################################################
+# DATABASE_URL=sqlite:///db.sqlite3
+
+# DATABASE_URL=postgres://cfgov:cfgov@postgres/cfgov
+# Required for Postgres dbshell in Django < 1.9.
+# PGPASSWORD=cfgov
+
+DATABASE_URL=mysql://v1:v1@mysql/v1
```
chosak: Default database when using Docker is MySQL to maintain current behavior.

refresh-data.sh
```diff
@@ -12,39 +12,43 @@ refresh_dump_name=$1
...
 download_data() {
 	echo 'Downloading fresh production Django database dump...'
-	prod_archive="$refresh_dump_name".gz
+	prod_archive="$refresh_dump_name"
 	curl -o $prod_archive $CFGOV_PROD_DB_LOCATION
-	gunzip -f $prod_archive
 }
 
 refresh_data(){
-	echo 'Dropping db'
```
chosak: I should point out that this is also a change from the current behavior. During normal DB refresh it's difficult to actually delete and recreate the database in a generic way, without using the DB-specific commands (like Postgres `dropdb`). For this reason I've modified this script to just re-import the data from the dump and skip the drop/recreate. One side effect this should have is to potentially leave old tables around.

runserver.sh
```diff
@@ -10,25 +10,13 @@
...
-  mysql(){
-  if ! mysql.server status; then
-    mysql.server start
-  fi
-  }
-
-  # Run tasks to build the project for distribution.
-  server(){
   if [ "$1" == "ssl" ]; then
     echo '\033[0;32mStarting SSL Django server on port' $DJANGO_HTTP_PORT '...'
```
higs4281: not related to this PR, but if we want color output, I think we need to use `echo -e` or `printf`

chosak: I've long wondered what those characters were for!

PR #4020, opened 2018-04-17
Add Turbolinks JS library behind a feature flag

contolini:
Turbolinks speeds up page loads by fetching and caching content. Is it a good idea? Who knows. We've been experimenting with it on eRegs and it might be appropriate for the larger cf.gov ecosystem. This PR will enable turbolinks across cf.gov when the TURBOLINKS feature flag is enabled.

https://github.com/turbolinks/turbolinks

## Additions

- Turbolinks on-demand JS file.
- `TURBOLINKS` feature flag that loads the above file on cf.gov pages when enabled.

## Testing

1. `./frontend.sh`
1. Enable the `TURBOLINKS` feature flag.
1. Load http://localhost:8000/ and click around to a few random pages. You should see a new progress bar at the top of the screen. Subsequent visits to the same pages should be instantaneous.

## Notes

- I am not endorsing the use of turbolinks. It's kind of scary and would require a lot of testing to ensure we're not breaking any functionality. But it's pretty neat and worth checking out.

### DIFF
--- a/cfgov/jinja2/v1/_layouts/base.html
+++ b/cfgov/jinja2/v1/_layouts/base.html
@@ -119,6 +119,9 @@
       {% if js_source | length > 0 %}
         s.push( '{{ static('js/routes/' + request.path[1:] + 'index.js') }}' );
       {% endif %}
+      {% if flag_enabled('TURBOLINKS', request) %}
+        s.push( '{{ static('js/routes/on-demand/turbolinks.js') }}' );
+      {% endif %}
       {# Pass scripts to JavaScript loader. #}
       jsl(s);
     }();
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -614,7 +614,7 @@ FLAGS = {
     # To be enabled when owning-a-home/closing-disclosure/
     # and owning-a-home/loan-estimate/ are de-sheered.
     'OAH_FORM_EXPLAINERS': {},
-    
+
     # To be enabled when journey pages are released in Wagtail.
     'OAH_JOURNEY': {},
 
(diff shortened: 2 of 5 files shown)


anselmbradford:
I remember it from Rails and that's all I know.

sebworks:
Interesting, hmmm. I 'm not sure we want to handle this with JS as opposed to using [prerender](https://www.keycdn.com/blog/resource-hints/).  The downside to using JS is that it's more code that has to be sent / parsed by the browser. It also wont work when JS is disabled. 

Screw it, lets try it though.

contolini:
> Prerendering is very similar to prefetching in that it gathers resources that the user may navigate to next. The difference is that prerendering actually renders the entire page in the background, all the assets of a document.

Oh wow it actually renders the entire page in the background. That is magical and scary. I'm going to try it with eRegs.

Turbolinks modifies the browsing experience too much for my taste. I'll leave it here for folks to try but I'm not keen on it reaching production.

anselmbradford:
Sorry, what's happening with this @contolini ? I'm confused by your comment. This PR needed to go in for eregs? I'm wondering since it adds a dependency for package.json and some template logic. Would it make more sense for it to live in a branch outside of master, unless we're sure we're going to use it?

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -119,6 +119,9 @@
       {% if js_source | length > 0 %}
         s.push( '{{ static('js/routes/' + request.path[1:] + 'index.js') }}' );
       {% endif %}
+      {% if flag_enabled('TURBOLINKS', request) %}
+        s.push( '{{ static('js/routes/on-demand/turbolinks.js') }}' );
```
anselmbradford: Would we want this in base-common.html, which is both the english and spanish page?

contolini:
Sorry @anselmbradford, I should have provided more context. We wanted to add this to the new [regs3k](https://github.com/cfpb/cfgov-refresh/tree/master/cfgov/regulations3k) app but instead of hardcoding it I wanted to give team members the ability to toggle it on and off. I then figured, "Why not just add it every page of cf.gov? It might work across the site" so I stuck the conditional in `base.html`.

I think you're right that it should be in `base-common.html`. In a couple weeks we'll determine if we want to use it and at that point I'll get rid of this feature flag and add it to eReg's JS bundle (or just delete it).

anselmbradford:
👍

PR #4035, opened 2018-04-19
Remove some deprecated pieces of sheerlike

chosak:
This change removes the last remaining Sheerlike site ("owning-a-home"), along with a few bits of the `sheerlike` app that can be cleanly removed. There are still other pieces of `sheerlike` that are still in use.

## Checklist


cfgov/cfgov/settings/base.py
```diff
@@ -465,14 +465,6 @@
 LOGIN_REDIRECT_URL = '/login/welcome/'
 LOGIN_URL = "/login/"
 
-
```
sebworks: We still need this code.

chosak: Where does this code still get used?

sebworks: for the rate checker; that flag hasn't been flipped yet.

chosak: Right, that path still [uses](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/urls.py#L86) `SheerTemplateView`, but that doesn't seem to depend on the Sheer Site at all. All that's really doing is modifying how/where the template file gets loaded from.

chosak:
Marking on hold pending ratechecker migration.

cfgov/cfgov/settings/base.py
anselmbradford: The flag has now been flipped in production.

PR #3887, opened 2018-03-15
Convert from using the CFPB Minicons icon font to SVG icons

jimmynotjim:
Updates the project to use the latest SVG icon changes from Capital Framework.

## Changes

- Converted remaining Jinja templates to use the SVG icon template

## Testing

1. `npm install && gulp build`
2. `./runserver.sh`
3. Navigate through the site and ensure there are no errors and all icons are rendered as expected.

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows the standards laid out in the [development playbook](https://github.com/cfpb/development)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [x] Project documentation has been updated
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Browsers

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Internet Explorer 8, 9, 10, and 11
- [x] Edge
- [x] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [x] Keyboard friendly
- [x] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [x] Is useable without JS
- [x] Flexible from small to large screens
- [x] No linting errors or warnings
- [x] JavaScript tests are passing

PR #4037, opened 2018-04-19
Init notifications in the correct location

jimmynotjim:
Currently notifications are initialized within the formSubmit which lead
to errors where the notifications weren't being updated correctly (a new 
instance was created every submission). Moving them to their direct 
parent and passing them into the formSubmit resolves the issue.

## Changes

- Moves notification init to it's parent function
- Passes notification into the formSubmit function

## Testing

1. Run `gulp test:unit` and everything should continue to pass

## Checklist

### DIFF
--- a/test/unit_tests/js/organisms/FormSubmit-spec.js
+++ b/test/unit_tests/js/organisms/FormSubmit-spec.js
@@ -1,0 +1,80 @@
+const BASE_JS_PATH = '../../../../cfgov/unprocessed/js/';
+const FormSubmit = require( BASE_JS_PATH + 'organisms/FormSubmit' );
+const BASE_CLASS = 'o-email-signup';
+const HTML_SNIPPET = `
+  <div class="o-email-signup">
+    <header class="m-slug-header">
+      <h2 class="a-heading">
+        Stay informed
+      </h2>
+    </header>
+
+    <form id="o-email-signup_54"
+          class="o-form o-form__email-signup"
+          action="/subscriptions/new/"
+          method="POST"
+          enctype="application/x-www-form-urlencoded">
+
+      <div class="u-mb15">
+        <div class="m-notification
+                    m-notification__success">
+          <span class="m-notification_icon
+                       cf-icon"></span>
+          <div class="m-notification_content">
+            <div class="h4 m-notification_message"></div>
+          </div>
+        </div>
+      </div>
+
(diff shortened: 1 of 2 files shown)


anselmbradford:
Can you explain in more detail the circumstances where this was arising? Like how would I reproduce the issue on the master branch?

jimmynotjim:
@anselmbradford the issue came up during the SVG Icon transition and I regret not pulling it out of that PR sooner. It's been so long I've forgotten the exact details

Essentially the issue was that the notification exists in the static markup and the init searches it out relative to the passed HTMLNode. When the forms were submitted, it was re-initializing on every submission, rather than updating the existing instance. This was causing issues where the notification couldn't update the existing icon https://github.com/cfpb/cfgov-refresh/pull/3887/files#diff-32e184401b65625fc114c891e29b833cR130

jimmynotjim:
Turns out the issue was slightly different than we originally thought. The notification was never initialized within the formSubmit and when we moved the instance to the files that call the formSubmit we fixed it just by initializing the notification. Updated the formSubmit to initialize (as well as create the instance) within it's own init method and everything is working as expected again.

test/unit_tests/js/organisms/FormSubmit-spec.js
```diff
@@ -0,0 +1,69 @@
...
+            Email address
+          </label>
+          <input id="form_3" type="email" placeholder="example@mail.com" name="email" class="a-text-input a-text-input__full">
+        </div>
+        <p>The information you provide will permit the Consumer Financial Protection Bureau to process your request or inquiry.&nbsp;<a class="" href="/privacy/privacy-policy/">See more</a>.</p>
+        <input class="a-btn a-btn__full-on-xs" type="submit" value="Sign up">
+      </div>
+
+      <div class="form-group">
+        <input type="hidden" name="code" value="USCFPB_91">
```
anselmbradford: Do we need the value code here in the test?

anselmbradford:
LGTM +1

You could stack the attributes to wrap the long lines in the test HTML_SNIPPET if you wanted.

test/unit_tests/js/organisms/FormSubmit-spec.js
jimmynotjim: I don't think so, removed and amended the commit.

jimmynotjim:
Good call, fixed the attributes in the fixture.

PR #4041, opened 2018-04-20
Dom tools tests

anselmbradford:
## Additions

- Adds initial BAH dom-tools spec.

## Changes

- Minor refactoring of dom-tools:
  - Remove unnecessary nesting.
  - Removes returns of applyAll, which runs forEach, which doesn't appear to return anything but undefined.
  - Changes `return div.firstChild` to `div.children[0]` to avoid returning text nodes.
  - Returns boolean from `applyAll`
  - Don't run split on arrays on `applyAll`.
  - Make isEl method internal.
  - Documents methods.

## Testing

1. `gulp test:unit` should pass and show 20% coverage in dom-tools.
2. `gulp clean && gulp build` should pass and form explainers should be unchanged in functionality.

### DIFF
--- a/cfgov/unprocessed/apps/owning-a-home/js/dom-tools.js
+++ b/cfgov/unprocessed/apps/owning-a-home/js/dom-tools.js
@@ -1,15 +1,28 @@
 import fastDom from 'fastdom';
 
-const NO_OP = function( ) {
+const NO_OP = () => {
   // Placeholder function meant to be overridden.
 };
 
+/**
+ * Apply a function to one or more elements.
+ * @param {HTMLElement|NodeList} elements - An HTML element or a list of them.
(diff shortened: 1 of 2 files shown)


cfgov/unprocessed/apps/owning-a-home/js/dom-tools.js
```diff
@@ -1,138 +1,229 @@
...
+
+/**
+ * Check whether something is a NodeList, HTML element, or window.
+ * @param {*} selector Something, possibly a list, element or window instance.
+ */
+function hide( selector ) {
+  applyAll(
+    getEls( selector ),
+    element => fastDom.mutate(
+      () => ( element.style.display = 'none' )
```
anselmbradford: Can this use `u-hidden`?

PR #4043, opened 2018-04-20
Updates fake emails

anselmbradford:
We should always use example.com for filler domains, mail.com is a spammy actual domain.

## Changes

- Changes `mail.com` to `example.com` in email signup.

### DIFF
--- a/cfgov/jinja2/v1/_includes/macros/subscribe.html
+++ b/cfgov/jinja2/v1/_includes/macros/subscribe.html
@@ -26,7 +26,7 @@
         <input id="email-subscribe-form_email"
                type="email"
                name="email"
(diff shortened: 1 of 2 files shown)


cfgov/jinja2/v1/_includes/macros/subscribe.html
```diff
@@ -26,7 +26,7 @@
         <input id="email-subscribe-form_email"
                type="email"
                name="email"
-               placeholder="example@mail.com"
+               placeholder="mail@example.com"
```
Scotchester: I'm not sure this macro i actually still used anywhere, but doesn't hurt to make the change anyway.

PR #4046, opened 2018-04-22
Fix mega menu jank

contolini:
The mega menu is [hidden](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/mega-menu.html#L184) on page load and unhidden via JS when it [initializes](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/js/organisms/MegaMenu.js#L93). This causes pages' content to jump down after a split second.

eRegs is experimenting with using turbolinks and service workers to cache content. When a reg section's text instantly updates, it makes the mega menu jank really obvious and a little distracting. 

This PR removes the mega menu jank by replacing the `u-hidden` class with `u-hide-on-mobile` so that desktops show it by default.

## Changes

- Mega menu is no longer hidden on desktops at page load.

## Testing

1. Pull down branch and `./frontend.sh`.
1. Refresh random pages at various viewport sizes and keep an eye on the menu.

## Screenshots

An eRegs page:

| Before                                                                                                 | After                                                                                           |
|--------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------|
| eregs-jank | eregs-no-jank |

A non-eRegs page:

| Before                                                                                                      |  After                                                                                               |
|-----------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------|
| cf-jank | cf-no-jank |


sebworks:
🥇

anselmbradford:
Let me check some things before you merge plz

anselmbradford:
Okay, I was digging into why we did this originally at all since the menu is functional without JS.
I believe it was that due to the complexity of the menu we wanted the menu to be fully initialized error-free before showing it, so if something went wrong it would be really apparent. However, I think we added the CSS interactions after that time, as well as New Relic. I'm actually not seeing what the `hide-on-mobile` is adding. Turning off JS and throttling the load in the performance tab I don't see an experience different from prod. Should we strip out `_dom.classList.remove( 'u-hide-on-mobile' );` and  `u-hide-on-mobile` in the template completely?

contolini:
Huh good sleuthing @anselmbradford. I'll try removing it and do some testing.

ascott1:
I cannot express to you the joy that this PR brings me.

sebworks:
@anselmbradford, I think the `u-hide` was added to prevent the menu from flashing on the initial load. I say merge the PR and figure out if it can be removed later.

ascott1:
@contolini @anselmbradford what is the hold up on this? I would love to see this change in prod.

contolini:
@ascott1 @anselmbradford I'll merge this and we can experiment with removing the class at a later date.

anselmbradford:
OOoooooo I figured out why we had this! So comment out your `MAPBOX_ACCESS_TOKEN` in `.env` and source the `source .env` and restart the server. Then visit http://localhost:8000/find-a-housing-counselor/ and see that the menu is stuck open covering half the page content. So not sure if we care about that, but that's why we did it. If things went wrong the menu wouldn't be stuck covering things up.

ascott1:
Wow, that's some impressive detective work @anselmbradford!

PR #4059, opened 2018-04-26
Bump moto to more recent version and make related fixes

acrewdson:
Back in 80c7c7e88b4b82953728aedc7156307ea980dd7e we downgraded `moto` to 0.4.27 to fix some conflicting dependency versions that had surfaced through the addition of `venv-update` to the tox setup. At the time, I submitted a PR to `moto` to address the problem with their dependencies, and that eventually got merged. But we had not yet gotten round to bumping `moto` back up to a more recent version.

Now we have no choice: our old version of `moto` specifies an unpinned version of `flask`, and `flask` released version 1.0 today. In flask 1.0, the Jinja2 dep is bumped, meaning that we now have another Jinja2 version conflict when attempting to run `tox`.

All of which is to say, we are now bumping `moto` back up to a more recent version, and fixing a few related tests that needed to be updated as a side-effect. While `moto` 1.3.1 _is_ a lot more recent (one month old), it's not the newest version as of this writing (1.3.3). Upgrading to 1.3.3 would mean upgrading boto3, and we might prefer to kick the can down the road on that one for now.

Also note that we are bumping `requests` to the latest version to address a version conflict (newer `moto` requires newer `requests`). Upgrading `requests` is probably a good idea in general, and should not pose problems since `requests` claims minor versions do not introduce breaking changes.

### DIFF
--- a/cfgov/v1/tests/test_meta_image.py
+++ b/cfgov/v1/tests/test_meta_image.py
@@ -111,7 +111,7 @@ class TestMetaImage(TestCase):
     )
     def test_template_image_image_url_s3(self):
         """Meta image links should work if using S3 storage."""
-        mock_s3 = moto.mock_s3()
+        mock_s3 = moto.mock_s3_deprecated()
         mock_s3.start()
 
         s3 = boto.connect_s3()
--- a/cfgov/v1/tests/test_s3utils.py
+++ b/cfgov/v1/tests/test_s3utils.py
@@ -25,7 +25,7 @@ from v1.s3utils import (
 )
 class S3UtilsTestCase(TestCase):
     def setUp(self):
-        mock_s3 = moto.mock_s3()
+        mock_s3 = moto.mock_s3_deprecated()
         mock_s3.start()
         self.addCleanup(mock_s3.stop)
 
--- a/requirements/libraries.txt
+++ b/requirements/libraries.txt
@@ -31,7 +31,7 @@ pyelasticsearch==0.6.1
 python-dateutil==2.1
 PyYAML==3.11
 pytz
-requests==2.12.4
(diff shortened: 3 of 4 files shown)


chosak:
@acrewdson thank you for this PR. Unfortunately a couple of the cf.gov satellite repos also share these pinned requirements:

[regulations-site](https://github.com/cfpb/regulations-site/blob/1d365d1233c2265a7da281b1cdacc60536e573fe/setup.py#L17) has a pinned `requests==2.12.4` in its `setup.py`.

[owning-a-home-api](https://github.com/cfpb/owning-a-home-api/blob/bd383a3281bd5e4798a2f1edbc7454462224fc39/requirements/base.txt#L8) has `requests==2.12.4` pinned in its `requirements.txt` -- this won't cause a conflict at install time, but it'd be better if these were consistent.

Generally it'd be nice if all of the cf.gov satellite repos included their requirements in `setup.py` instead of requirements files -- this would let us validate things at install time and also let us remove a few things from the cfgov-refresh requirements that could be specified in the satellites.

@willbarton I find moto3 to be very useful and would prefer to use it over manual mocking.

willbarton:
> Generally it'd be nice if all of the cf.gov satellite repos included their requirements in setup.py instead of requirements files -- this would let us validate things at install time and also let us remove a few things from the cfgov-refresh requirements that could be specified in the satellites.

This would be a fantastic endeavor, particularly for owning-a-home-api. regulations-site shouldn't be long for this world, so it might make less sense.

> @willbarton I find moto3 to be very useful and would prefer to use it over manual mocking.

Alrighty. The boto API *does* make it quite difficult to mock AWS calls.

PR #4070, opened 2018-04-30
Add optional heading block to the table block

jimmynotjim:
When a user adds a table that is outside the full width text module,
there is no way to add a heading without relying on another block and
the spacing that goes with it (typically 60px). To alleviate this issue
I have added the heading block as a custom entry to the Table Block JS
as well as a necessary style fix to the Table Block CSS.

## Additions

- Adds heading block markup to table block template
- Adds JS handlers for new text input and select fields
- Adds heading inputs to table input template
- Adds CSS necessary to align heading block inputs

## Testing

1. Add or edit a table in Wagtail
2. Add heading text, select a level, and add an icon name
3. Save and view page, the appropriate heading level should be used with the text and icon entered in the admin.

## Screenshots

screen shot 2018-04-30 at 4 14 46 pm

## Notes

- Due to the nature of the table block, we could not use the existing heading block in the `organisms.py` file like elsewhere.

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/table.html
+++ b/cfgov/jinja2/v1/_includes/organisms/table.html
@@ -11,6 +11,14 @@
 
    value.data:            A multi-dimensional list representing table rows.
 
+   value.heading_text:    (Optional) The text of a heading.
+
+   value.heading_level:   The heading level (H2, H3, H4). Default: H2
+
+   value.heading_icon:    (Optional) A string containing cf-icon name
+                          that triggers the insertion of an icon before
+                          the heading.
+
    value.row_links:       Boolean indicating if the table contains row links.
 
    value.is_stacked:      Boolean indicating if the table columns
@@ -34,6 +42,8 @@
    value.column_widths:   A list of column width (classes) to be applied in
                           order to the table columns.
 
+   value.sortable:        Boolean indicating if the table is sortable.
+
    ========================================================================== #}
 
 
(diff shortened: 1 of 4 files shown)


sebworks:
I'm not sure if we should be adding an optional heading to the components just to control spacing.  We should probably stop wrapping every component in a [block](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/templates/render_block.html#L2) and just have a heading component.

Scotchester:
I see your point, Steven. Wrapping each Wagtail module in a block certainly leads to its fair share of excessive spacing. That said, if we stopped doing that, it's a little unclear to me how we would ensure that certain sections have _enough_ spacing between them. It definitely warrants further discussion, but it will take a while to figure out, and probably shouldn't hold up this PR.

jimmynotjim:
@sebworks This is a really good point and I maybe should have thought it over a little bit more. What if we add the Heading Block as a top level option to all the models and then fiddled with the bottom margin on just the heading blocks to remove the 60px? Then the heading level's natural margin would be utilized to create the appropriate spacing between the heading and the following block. 

One worry I have is would wrapping the heading and then the content that follows in their own divs break the content outline? If the heading doesn't live in the same wrapper that it's describing would we be creating a whole other issue with accessibility? 

Maybe the answer is somewhere in between, adding the heading block as a sub-module for all the organism blocks not just here and there.

jimmynotjim:
Tested this idea locally and my second theory is correct, the solo heading living in it's own block wrapper (whether that's a heading block or full-width text block) creates an awkward outline where it's unclear what content the heading is describing, which will likely lead to accessibility issues. I'm thinking we should eventually add the heading block as an option to all the organisms or replace the existing heading field that doesn't allow icons (like the Snippet lists).

Scotchester:
The idea of creating a new block solely for headings feels risky to me for reasons I can't quite articulate. Possibly because I have a vague memory of already going down this road when figuring out the InfoUnitGroups – remember when we had the notion of creating a "Module Introduction" (not sure if that's actually what we called it) block? We determined that was a bad idea at the time.

I also think there's some intangible value in keeping the heading content within the module it belongs to.

PR #4071, opened 2018-04-30
Prevent iframe_api loading on pages without iframes

anselmbradford:
Google analytics youtube script adds the youtube iframe_api to all pages. We should only add it if there's a youtube video on the page. This change checks the page for iframes and if none are found aborts the analytics script.

## Changes

- Checks for existence of iframes on the page before injecting youtube's iframe_api script.
- Minor reformatting of youtube script to avoid dangling comma starting a new line.

## Testing

1. `gulp clean && gulp build`
2. Visit http://localhost:8000/about-us/blog/video-a-message-from-director-cordray/
3. Open the dev console and note that the iframe_api script has loaded in the network tab.
4. Visit the homepage.
5. Note that the iframe_api script has not loaded.

## Note

- You can add `?debug-gtm=true` to the URL and see in the console that it'll say `ANALYTICS DEBUG MODE: There are no iframes on the page!` on pages where there aren't videos.
- If you see `Error parsing header X-XSS-Protection` errors in the dev console, that's a Chrome error fixed in 66 https://stackoverflow.com/questions/48714879/error-parsing-header-x-xss-protection-google-chrome

## Screenshot

screen shot 2018-04-30 at 6 20 10 pm

### DIFF
--- a/cfgov/unprocessed/apps/analytics-gtm/js/youtube-event-listeners-older-style.js
+++ b/cfgov/unprocessed/apps/analytics-gtm/js/youtube-event-listeners-older-style.js
@@ -18,11 +18,18 @@ import { analyticsLog } from './util/analytics-util';
       'There is already a function defined at window.onYouTubeIframeAPIReady;' +
       ' aborting LunaMetrics Google Analytics YouTube Tracking'
     );
+    return;
+  }
+
+  const iframes = document.getElementsByTagName( 'iframe' );
+
+  if ( iframes.length === 0 ) {
+    analyticsLog( 'There are no iframes on the page!' );
+    return;
   }
 
   // Invoked by the YouTube API once it has finished loading.
   window.onYouTubeIframeAPIReady = function() {
-    const iframes = document.getElementsByTagName( 'iframe' );
     const embeds = document.getElementsByTagName( 'embed' );
 
     digestPotentialVideos( iframes );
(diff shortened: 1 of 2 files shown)


cfgov/unprocessed/apps/analytics-gtm/js/youtube-event-listeners-older-style.js
```diff
@@ -18,11 +18,18 @@ import { analyticsLog } from './util/analytics-util';
       'There is already a function defined at window.onYouTubeIframeAPIReady;' +
       ' aborting LunaMetrics Google Analytics YouTube Tracking'
     );
+    return;
+  }
+
+  const iframes = document.getElementsByTagName( 'iframe' );
+
+  if ( iframes.length === 0 ) {
```
jimmynotjim: Could you do `!iframes.length` for brevity and consistency with how we usually check length elsewhere.

anselmbradford: It would work like that, but since it's talking about a quantity here I'd probably leave it as a number. `!` is casting it to a boolean. Seems kinda obfuscating. https://codepen.io/anon/pen/GdWxVW?editors=1111

jimmynotjim: Correct, but we often do the inverse as `if (array.length) {}` to check existence before doing something as well. Since you're not actually doing anything with the item count, there's nothing wrong with making it a boolean.

PR #4049, opened 2018-04-24
eRegs progressive web app

contolini:
This PR sets the foundation for service workers and general [PWA](https://developers.google.com/web/progressive-web-apps/checklist)-ness of the new eRegs. [sw-precache](https://github.com/GoogleChromeLabs/sw-precache) is used to generate a service worker that caches eRegs pages so that subsequent page loads are loaded from memory instead of the network. It should make the browsing of regulations super duper fast.

Service workers need to reside at the website's root (i.e. `https://consumerfinance.gov/service-worker.js`) so I named it `eregs-service-worker.js` to avoid any confusion. We may want to create new service workers for other parts of cf.gov in the future.

It's not picking up 100% of eReg's assets so the sw-precache config will need some further tweaking. You can verify the manifest and service worker are working by viewing the Application section of chrome dev tools:

screen shot 2018-04-24 at 1 59 23 am

screen shot 2018-04-24 at 1 59 38 am

You can also verify that the `prerender` resource hint is successfully downloading the next reg section page in the background by visiting `chrome://net-internals/#prerender`.

screen shot 2018-04-24 at 2 01 16 am

## Additions

- Custom webpack config file for eRegs
- Service worker generating script in the webpack file
- Unofficial-definitely-not-approved-but-still-cool eRegs icons designed by the amazing @niqjohnson 
- A [web app manifest](https://developers.google.com/web/fundamentals/web-app-manifest/) that tells browsers about the website

## Removals

- eRegs placeholder JS

## Testing

1. `./frontend.sh`
1. Load a new Regulations 3k reg section page.
1. Open the Application tab in chrome dev tools and see if the manifest and service worker are present.

## Todos

- Ensure all assets are being cached.
- Need to create a 200 offline page.
- The site's mobile experience needs to be improved in general.

## Screenshots

Android users will see a popup recommending they "install" eRegs by saving it to their home screen (please forgive the temporary awkward mobile menu).

eregs-pwa

PR #4072, opened 2018-05-01
eRegs progressive web app

contolini:
This is the same as #4049 but with the URLs updated to accommodate the new location at `cf.gov/regulations`. 

## Additions

- Custom webpack config file for eRegs
- Service worker generating script in the webpack file (the generated service worker is ignored by git)
- Unofficial-definitely-not-approved-but-still-cool eRegs icons designed by the amazing @niqjohnson 
- A [web app manifest](https://developers.google.com/web/fundamentals/web-app-manifest/) that tells browsers about the website

## Changes

- Use [routablepageurl](http://docs.wagtail.io/en/v2.0.1/reference/contrib/routablepage.html#the-routablepageurl-template-tag) instead of hardcoding prev/next URLs.

## Testing

1. `./frontend.sh`
1. Load a new Regulations 3k reg section page.
1. Open the Application tab in chrome dev tools and see if the manifest and service worker are present.

## Screenshots

screen shot 2018-04-24 at 1 59 23 am

screen shot 2018-04-24 at 1 59 38 am

As you click through a reg's sections, in the Network tab you should see the next page being prefetched and cached by the service worker (gear icon = service worker).

screen shot 2018-04-30 at 9 55 37 pm


cfgov/regulations3k/tests/test_models.py
```diff
@@ -125,7 +125,7 @@ def test_routable_page_get_context(self):
 
     def test_get_reg_nav_items(self):
         request = HttpRequest()
-        request.url = '/eregulations3k/1002/1002-4/'
+        request.url = '/regulations/1002/1002-4/'
```
higs4281: probably doesn't matter for the test, but we're using the form `/1002/4/` now.

cfgov/unprocessed/apps/regulations3k/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "cfpb-regulations-3000",
+  "name": "eRegulations",
   "description": "The consumerfinance.gov/eregulations website.",
   "homepage": "https://www.consumerfinance.gov/eregulations/",
```
higs4281: Do we want to go with `/regulations/` everywhere for now?

contolini: Yesh

cfgov/regulations3k/tests/test_models.py
contolini: Good call

cfgov/unprocessed/apps/regulations3k/package.json
higs4281: homepage too?

PR #4077, opened 2018-05-01
Don't show draft versions of reg pages on beta

chosak:
This change removes three regulation-related pages from the set of pages that are shown in their draft form in the beta deploy environment.

See platform#2721 for background.


chosak:
Thanks for the review @anselmbradford. No surplus testing necessary.

PR #4078, opened 2018-05-01
Fix an issue where richtext images are being distorted

jimmynotjim:
Currently richtext images are printed to the markup with static height
and width attributes. When the image is resized in one dirction, it
can be distorted in the opposite direction. Resetting the height and
width to auto avoids this issue allowing the image to take up the
dimensions of it's container without causing distortion.

## Additions

- New global `images.less` file because this change is limited to Wagtail and not necessary to be moved to CF. This file will also be the future home of whatever Wagtail high-dpi/responsive image solutions we explore.

## Testing

1. Gulp build
1. Create a table with fixed dimensions
1. Add an image to one of the columns (best if the column is narrow, like 20%)
1. Observe the image is not distorted in the way you can [see it is here](http://build.consumerfinance.gov/consumer-tools/mayg-by-topic/money-as-you-grow-bookshelf-at-home/)

## Screenshots
Before:

screen shot 2018-05-01 at 1 07 23 pm

After:

screen shot 2018-05-01 at 1 07 59 pm

## Checklist

### Browsers

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [ ] iOS Safari
- [ ] Chrome for Android

### DIFF
--- a/cfgov/unprocessed/css/images.less
+++ b/cfgov/unprocessed/css/images.less
@@ -1,0 +1,10 @@
+// Images inserted via rich text field are given strict width and height
+// attributes in the resulting markup. This resets the dimensions to allow
+// the image to resize to its container without distortion
+//
+// TODO: Check if this can be removed after upgrading to Wagtail 2.0
+
+.richtext-image {
+    height: auto;
+    width: auto;
+}
--- a/cfgov/unprocessed/css/main.less
+++ b/cfgov/unprocessed/css/main.less
@@ -59,6 +59,7 @@
    ========================================================================== */
 
 @import (less) "forms.less";
+@import (less) "images.less";
 @import (less) "media.less";
 @import (less) "meta.less";
 @import (less) "misc.less";


sebworks:
@jimmynotjim, I think we shouldn't depend on their class names and css for our frontend. It's subject to change at anytime. Wagtail allows the to modify the class names associated with the images in the [RTE](https://github.com/wagtail/wagtail/blob/76975525f7a3757b280572165f8717cc4c13bba6/docs/advanced_topics/customisation/page_editing_interface.rst#image-formats-in-the-rich-text-editor).

cfgov/unprocessed/css/images.less
```diff
@@ -0,0 +1,7 @@
+// Richtext images have set width and height attributes. Reset them so the
+// images fills it's container without distortion.
```
Scotchester: First, two pedantic things:
1. I'd open with "Images inserted in rich text fields have …".
2. Should be "image [singular] fills its [no apostrophe] container".

More importantly, it seems somewhat inaccurate. Images less wide than their container will not be scaled up to the width of their container, right?

Finally, it seems very possible that the default behavior could change with the implementation of Draftail in Wagtail 2.0. I suggest leaving a to-do to revisit that upon upgrading to 2.0.

Scotchester:
@sebworks That's a good point. We should probably be using the `image_formats.py` process to customize that. It will probably also come in handy for the work I would like to do during the next cycle to improve image responsiveness.

jimmynotjim:
I disagree. We already have three available formats, this is simply fixing a default that shouldn't exist and from the sounds of it may be removed, at which point the fix can also be removed. Registering new formats would mean either unregistering the existing formats and performing a migration for existing images, or asking users to understand how the image container is behaving and make the correct choice between images with static dimensions and those without. And then what do we do if Wagtail fixes the default behavior, perform another migration to move everything back again? This is essentially a patch and scoped specifically to the offending code without accidentally affecting anything else. If and when the markup is corrected this change can be removed without unintended consequences.

Scotchester:
Yeah, understanding now that there are three existing formats, I agree that adding a new one is probably best left for later.

sebworks:
Yeah, I just don't think we should rely on code from our CMS on our website in any capacity. JS, CSS, HTML, etc.. We don't include any Wagtail image styles but we are using the class names. It also changes the nature of our testing from content management functionality to visual changes in the website ( Wagtail class name changes affect if images our skewed on the website ). 

Having said all that, the scope of this is small and it shouldn't hold this PR up.

jimmynotjim:
@sebworks I hear you, and had originally targeted all images but then decided to narrow the selector as much as possible to avoid any unexpected changes elsewhere. Even if Wagtail doesn't revert the nature of the dimension attributes in 2.0, I think we should explore that ourselves as a long term solution.

PR #4081, opened 2018-05-02
Use PostgreSQL version 10 and cfpb username

chosak:
This change bumps the version of the PostgreSQL database created in Docker from 9.6 to 10.

It bumps Travis to version 10 as well, using the travis database and user that are created as part of v10 setup there.

It also standardizes the default Postgres username to `cfpb`, to be more consistent with the data dumps generated by https://github.com/cfpb/cfgov-mysql-to-postgres.

## Changes

- Use Postgres 10 in Docker and for testing in Travis.
- Standardize Postgres username to `cfpb` and database to `cfgov` everywhere.

## Testing

See Travis unit tests. You can also run the Docker setup with the Postgres `DATABASE_URL` in `.python_env_SAMPLE`.

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,11 @@ services:
   - postgresql
 
 addons:
-  postgresql: "9.6"
+  postgresql: "10"
+  apt:
+    packages:
+    - postgresql-10
+    - postgresql-client-10
   chrome: stable
 
 cache:
@@ -33,7 +37,7 @@ branches:
     - master
 
 before_script:
-  - psql -c 'create database travis_ci_test;' -U postgres
+  - psql -c "ALTER USER travis WITH PASSWORD 'travis';"
 
 script:
   - ./travis_run.sh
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -33,7 +37,7 @@ branches:
     - master
 
 before_script:
-  - psql -c 'create database travis_ci_test;' -U postgres
+  - psql -c "ALTER USER travis WITH PASSWORD 'travis';"
```
richaagarwal: Is there a reason we need to set the password?

chosak: Yes, if you look in [the Travis log](https://api.travis-ci.org/v3/job/374023986/log.txt) you can see that they install Postgres 10 with `initdb --auth-host md5`, which [requires password login for users](https://www.postgresql.org/docs/10/static/auth-methods.html#AUTH-PASSWORD). We could optionally create our own new user and set its password but I thought it less complex to just give the `travis` user a password to use. This method is also suggested in [this discussion](https://github.com/travis-ci/travis-ci/issues/8537#issuecomment-383758631).

PR #4084, opened 2018-05-03
Remove django-compressor

willbarton:
See platform#2725

`{% compress %}` is trying to write to a cache in the Django static assets location, which isn't possible due to permissions on our servers. This behavior changed in Wagtail 1.13 ("Use minified versions of CSS in the admin by adding minification to the front-end tooling") which is why we're only seeing this bug now.

This change modifies our custom admin templates to remove {% compress %} as in the 1.13 Wagtail admin templates. It also removes our dependence on django-compressor and updates the login.html and confirm.html templates to mirror recent changes in Wagtail.

## Checklist

### Browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


cfgov/templates/wagtailadmin/login.html
```diff
@@ -1,35 +1,40 @@
...
         {% endif %}
 
-        {% if 'build' in request.site.hostname %}
-            <h1>Log in to demo Wagtail</h1>
-        {% else %}
-            <h1>Log in to consumerfinance.gov</h1>
-        {% endif %}
-        <form action="{% url 'cfpb_login' %}" method="post" autocomplete="off">
+
+        {% block above_login %}
```
chosak: This logic feels outdated to me, especially given the various environments we have.

cfgov/templates/wagtailadmin/account/password_reset/confirm.html
```diff
@@ -1,19 +1,19 @@
 {% extends "wagtailadmin/admin_base.html" %}
-{% load compress static i18n %}
+{% load static i18n %}
```
chosak: Why is this `static` here but [`staticfiles` in Wagtail](https://github.com/wagtail/wagtail/blob/v1.13.1/wagtail/wagtailadmin/templates/wagtailadmin/account/password_reset/confirm.html#L2)? Looks like that was changed in Wagtail [here](https://github.com/wagtail/wagtail/commit/2444ee6e8d8ad2c3729e74cd73c93a6146261d1e#diff-c9bbbc9e5782ca30e7bcc08627d7629d).

cfgov/templates/wagtailadmin/login.html
```diff
@@ -1,35 +1,40 @@
...
-
+{% load static i18n %}
 
 {% block furniture %}
     <div class="content-wrapper">
-        {% if messages  or form.non_field_errors %}
+        {% if messages or form.errors %}
             <div class="messages">
                 <ul>
+                    {% if form.errors %}
```
richaagarwal: Is there a reason we don't want to enumerate the specific errors anymore?

cfgov/templates/wagtailadmin/account/password_reset/confirm.html
```diff
@@ -1,19 +1,19 @@
...
 {% block titletag %}{% trans "Set your new password" %}{% endblock %}
 {% block bodyclass %}login{% endblock %}
 {% load wagtailadmin_tags %}
```
chosak: Is this line still needed?

cfgov/templates/wagtailadmin/account/password_reset/confirm.html
```diff
@@ -1,19 +1,19 @@
...
-    {% compress css %}
-        <link rel="stylesheet" href="{% static 'wagtailadmin/css/layouts/login.css' %}" type="text/css" />
-    {% endcompress %}
+    <link rel="stylesheet" href="{% static 'wagtailadmin/css/layouts/login.css' %}" type="text/css" />
 {% endblock %}
 
 {% block furniture %}
     <div class="content-wrapper">
+
+    {% if validlink %}
```
chosak: Did you bring this in from [this commit](https://github.com/wagtail/wagtail/commit/84a2a45d49307b533f16eeb2901b368a65ad7453#diff-a9f218f2a6735e21fd62ef455224ca98)? I think perhaps this isn't available yet, it's still pending in the 2.1 release.

cfgov/templates/wagtailadmin/login.html
willbarton: Nope, I think I was too aggressive in bringing over what I thought were Wagtail changes.

cfgov/templates/wagtailadmin/account/password_reset/confirm.html
```diff
@@ -26,23 +26,34 @@
...
-
-        <form method="post">
+        <form method="post" novalidate>
             {% csrf_token %}
             <h1>{% trans "Set your new password" %}</h1>
             <ul class="fields">
                 {% for field in form %}
                     <li class="{% if field.field.required %}required{% endif %} {{ wrapper_classes }} {{ li_classes }} {% if field.errors %}error{% endif %}">
-                      <div class="field {{ field|fieldtype }} {{ field|widgettype }} {{ field_classes }}">
+                    <div class="field {{ field|fieldtype }} {{ field|widgettype }} {{ field_classes }}">
```
chosak: Do you know why we have this custom loop over fields instead of copying the explicit logic [in the Wagtail template](https://github.com/wagtail/wagtail/blob/v1.13.1/wagtail/wagtailadmin/templates/wagtailadmin/account/password_reset/confirm.html#L27)?

cfgov/templates/wagtailadmin/login.html
```diff
@@ -1,10 +1,9 @@
 {% overextends "wagtailadmin/login.html" %}
-{% load compress static i18n %}
-
+{% load static i18n %}
```
chosak: Same comment re `static` instead of [`staticfiles`](https://github.com/wagtail/wagtail/blob/v1.13.1/wagtail/wagtailadmin/templates/wagtailadmin/login.html#L2).

cfgov/templates/wagtailadmin/login.html
```diff
@@ -17,19 +16,24 @@
             </div>
         {% endif %}
 
-        {% if 'build' in request.site.hostname %}
-            <h1>Log in to demo Wagtail</h1>
-        {% else %}
-            <h1>Log in to consumerfinance.gov</h1>
-        {% endif %}
-        <form action="{% url 'cfpb_login' %}" method="post" autocomplete="off">
+        {% block above_login %}
```
chosak: This logic feels outdated given the number of Wagtail environments we have. "build" isn't really a good indicator for "demo Wagtail", and non-"build" doesn't necessarily mean "consumerfinance.gov".

cfgov/templates/wagtailadmin/login.html
```diff
@@ -64,10 +61,14 @@ <h1>Log in to consumerfinance.gov</h1>
...
                     </li>
                 {% endcomment %}
+                {% endblock %}
                 <li class="submit">
+                    {% block submit_buttons %}
                     <button type="submit" class="button button-longrunning" tabindex="3" data-clicked-text="{% trans 'Signing in...' %}"><span class="icon icon-spinner"></span><em>{% trans 'Sign in' %}</em></button>
+                    {% endblock %}
                 </li>
             </ul>
+        {% endblock %}
```
chosak: Can we bring in [this logic](https://github.com/wagtail/wagtail/blob/v1.13.1/wagtail/wagtailadmin/templates/wagtailadmin/login.html#L82) to add the username focus?

cfgov/templates/wagtailadmin/login.html
```diff
@@ -1,10 +1,9 @@
 {% overextends "wagtailadmin/login.html" %}
-{% load compress static i18n %}
-
+{% load static i18n %}
 
 {% block furniture %}
     <div class="content-wrapper">
-        {% if messages  or form.non_field_errors %}
+        {% if messages or form.non_field_errors %}
```
chosak: This logic has changed slightly [in 1.13](https://github.com/wagtail/wagtail/blob/v1.13.1/wagtail/wagtailadmin/templates/wagtailadmin/login.html#L14), now using `form.errors` here and just below.

willbarton:
@chosak Our custom password requirements, and outputting them, seems to be what's different about our password confirmation page.

cfgov/templates/wagtailadmin/login.html
willbarton: I think this has to be a larger discussion. I recall this being explicitly added this way. @schaferjh?

cfgov/templates/wagtailadmin/account/password_reset/confirm.html
willbarton: Because of our custom password requirements. It's possible for a password to not have a sufficient character count, not have certain types of characters. We have more possible errors than Wagtail does.

cfgov/templates/wagtailadmin/login.html
schaferjh: @willbarton @chosak it probably is outdated but I agree it was added purposefully back in the day. Just jumping in like this, I'm not sure of the consequences of removing it. Happy to have a quick chat if you're trying to move quickly. FYI @ascott1

cfgov/templates/wagtailadmin/login.html
willbarton: I *think* that gets brought over because the template [`overextends`](https://github.com/cfpb/cfgov-refresh/pull/4084/files/b99dd9c4ca1c8b33f62326f9e5d5717453783fd2#diff-c0e993a61a4307b99a36af5801d4e183L1) wagtailadmin/login.html? I will confirm.

cfgov/templates/wagtailadmin/login.html
chosak: @schaferjh thanks; it's definitely not urgent to remove. I'd be curious if the goal was more to warn people who might be editing cf.gov, or give them assurance that they're on a "demo" system if on build. For example, someone logging into beta or a demo dev server would see the "consumerfinance.gov" message which wouldn't be accurate.

schaferjh: @chosak that rings a bell. I'm sure the product owner (!!) asked to make it clear to users logging in to build that they were on a demo system and not editing the live site.

cfgov/templates/wagtailadmin/login.html
chosak: @willbarton I can confirm that this does get brought in.

cfgov/templates/wagtailadmin/login.html
chosak: I haven't implemented this change as we handle field errors differently from vanilla Wagtail.

chosak:
I've made some minor changes to the templates to address some of the open questions above. If this PR can get a review and merge we won't have to roll back 1.13 in #4089.

higs4281:
👍 Not on the review list, but this looks good to me and doesn't raise issues locally, although the true test will be on different servers.

Scotchester:
Trying to go through the password reset flow locally, when I go to the provided reset link, I get the following error:

> ### TemplateSyntaxError at /admin/password_reset/confirm/MTg/4vz-a9082daae9b165e06722/
> Invalid filter: 'fieldtype'

In template /src/cfgov-refresh/cfgov/templates/wagtailadmin/account/password_reset/confirm.html, error at line 30

PR #4085, opened 2018-05-03
Removes jQuery from owning-a-home analytics scripts

anselmbradford:
## Additions

- `addEventListenerToElems` utility method to analytics utils.
- `closest` utility method to analytics utils.

## Removals

- Removes jquery for owning-a-home analytics scripts.
- Removes delay requirement from owning-a-home disclosure pages analytics scripts.

## Changes

- Makes delays individual instances.

## Testing

1. `gulp clean && gulp build`
2. Visit `/owning-a-home/closing-disclosure/?debug-gtm=true`.
3. Open dev console.
4. Click pages, next/prev button, expandables, and image overlays and ensure analytics log correctly in console (e.g. should say "expanded" when expanding the expandable, and "collapsed" when it's collapsing).
3. Repeat for `/owning-a-home/loan-estimate/?debug-gtm=true`

## Notes

- Ideally the analytics could listen to an expand/collapse event on the expandables instead of inferring their state. What it the current thinking on the architecture of having components communicate with each other?

### DIFF
--- a/cfgov/unprocessed/apps/analytics-gtm/js/bah-closing-disclosure-listeners.js
+++ b/cfgov/unprocessed/apps/analytics-gtm/js/bah-closing-disclosure-listeners.js
@@ -1,68 +1,144 @@
-// TODO: Remove $.
-import $ from 'jquery';
-
 import {
-  delay,
+  addEventListenerToElems,
+  closest,
   track
 } from './util/analytics-util';
 
-const OAHCDAnalytics = ( function( $ ) {
+// Owning a Home /owning-a-home/closing-disclosure/
 
-  $( '.tab-link' ).click( function() {
-    const text = $( this ).text().trim();
+const OAHCDAnalytics = ( () => {
+
+  const expandableStates = {};
+  let lastExpandable;
+
+  /**
+   * Record state (expanded or collapsed) of expandables.
+   * @param {string} id - The unique HTML ID of an expandable.
+   */
+  function recordExpandableState( id ) {
+    if ( !expandableStates[id] ) {
+      expandableStates[id] = true;
+    } else if ( lastExpandable === id ) {
+      expandableStates[id] = !expandableStates[id];
+    }
+    lastExpandable = id;
+  }
+
+  /**
+   * Record state (expanded or collapsed) of expandables.
+   * @param {string} expandable - An expandable HTML element.
+   */
+  function isAnimatingExpandable( expandable ) {
+    let isAnimating = false;
+
+    if ( expandable.classList.contains( 'o-expandable__expanding' ) ||
+         expandable.classList.contains( 'o-expandable__collapsing' ) ) {
+      isAnimating = true;
+    }
+
+    return isAnimating;
+  }
+
+  /**
+   * @param {MouseEvent} event - Mouse event from the click.
+   */
+  function trackTabLinkClick( event ) {
+    const target = event.target;
+    const text = target.textContent.trim();
     track( 'OAH Closing Disclosure Interaction', 'Tab click', text );
-  } );
+  }
 
-  $( '.form-explainer_page-link' ).click( function() {
-    const pageNumber = 'Page ' + $( this ).attr( 'data-page' );
+  /**
+   * @param {MouseEvent} event - Mouse event from the click.
+   */
+  function trackFormExplainerPageLinkClick( event ) {
+    const target = event.target;
+    const pageNumber = 'Page ' + target.getAttribute( 'data-page' );
     track(
       'OAH Closing Disclosure Interaction', 'Page link click', pageNumber
     );
-  } );
+  }
 
-  $( '.form-explainer_page-buttons button' ).click( function() {
-    const currentPage = 'Page ' + $( '.form-explainer_page-link.current-page' ).attr( 'data-page' );
+  /**
+   * @param {MouseEvent} event - Mouse event from the click.
+   */
+  function trackFormExplainerPageButtonClick( event ) {
+    const target = event.currentTarget;
+    const currentPageDom =
+      document.querySelector( '.form-explainer_page-link.current-page' );
+    const currentPage = 'Page ' + currentPageDom.getAttribute( 'data-page' );
     let action = 'Next Page button clicked';
-    if ( $( this ).hasClass( 'prev' ) ) {
+    if ( target.classList.contains( 'prev' ) ) {
       action = 'Previous Page button clicked';
     }
     track( 'OAH Closing Disclosure Interaction', action, currentPage );
+  }
 
-  } );
+  /**
+   * @param {MouseEvent} event - Mouse event from the click.
+   */
+  function trackExpandableTargetsClick( event ) {
+    const elem = event.currentTarget;
+    const expandable = elem.parentNode;
+    const expandableID = expandable.id;
+    if ( isAnimatingExpandable( expandable ) ) { return; }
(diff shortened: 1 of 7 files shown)


cfgov/unprocessed/apps/analytics-gtm/js/util/analytics-util.js
```diff
@@ -12,13 +28,61 @@ function addEventListenerToElem( elem, event, callback ) {
...
+
+/* Search for support of the matches() method by looking at
+   browser prefixes.
+   @param {HTMLNode} elem
+   The element to check for support of matches() method.
+   @returns {Function} The appropriate matches() method of elem. */
+function _getMatchesMethod( elem ) {
+  return elem.matches ||
+       elem.webkitMatchesSelector ||
+       elem.mozMatchesSelector ||
```
sebworks: I think we just need the ms prefix

cfgov/unprocessed/apps/analytics-gtm/js/util/analytics-util.js
```diff
@@ -12,13 +28,61 @@ function addEventListenerToElem( elem, event, callback ) {
...
+}
+
+/**
+ * Get the nearest parent node of an element.
+ *
+ * @param {HTMLNode} elem - A DOM element.
+ * @param {string} selector - CSS selector.
+ * @returns {HTMLNode} Nearest parent node that matches the selector.
+ */
+function closest( elem, selector ) {
```
sebworks: I think we have three [separate](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/js/modules/util/dom-traverse.js#L66) versions of this now. We should probably consolidate them or just use an npm module.

cfgov/unprocessed/apps/analytics-gtm/js/bah-loan-estimate-listeners.js
```diff
@@ -1,68 +1,142 @@
...
+      action = 'Image Overlay click - expandable expanded';
+    }
+    track( 'OAH Loan Estimate Interaction', action, text );
+  }
+
+  const tabLinksDom = document.querySelectorAll( '.tab-link' );
+  addEventListenerToElems( tabLinksDom, 'click', trackTabLinkClick );
+
+  const formExplainerPageLinksDom =
+    document.querySelectorAll( '.form-explainer_page-link' );
```
sebworks: Why not just allow `addEventListenerToElems` to accept a selector or DOM element?

cfgov/unprocessed/apps/analytics-gtm/js/util/analytics-util.js
anselmbradford: `elem.msMatchesSelector` on the line below?

cfgov/unprocessed/apps/analytics-gtm/js/util/analytics-util.js
anselmbradford: Yeah that makes sense, though I think an npm module would be the way to go. I don't think analytics should be reaching out into the rest of the codebase. Actually we should probably look into refactoring it away here in the analytics. If we had the architecture for components to talk to each other, analytics could just listen to what's happening in the components and wouldn't need to snoop around the dom like this.

sebworks: >> If we had the architecture for components to talk to each other, analytics could just listen to what's happening in the components and wouldn't need to snoop around the dom like this.

Yep, once I get done with the rate checker, it's up next. 

>> I think an npm module would be the way to go

Yeah, just pick one and I'll remove the other code to use that one as well.

cfgov/unprocessed/apps/analytics-gtm/js/bah-closing-disclosure-listeners.js
```diff
@@ -1,68 +1,145 @@
...
+    if ( target.classList.contains( 'prev' ) ) {
       action = 'Previous Page button clicked';
     }
     track( 'OAH Closing Disclosure Interaction', action, currentPage );
+  }
+
+  /**
+   * @param {MouseEvent} event - Mouse event from the click.
+   */
+  function trackExpandableTargetsClick( event ) {
```
jimmynotjim: Instead of tracking the click and checking the state, would it better to watch for DOM changes with a MutationObserver?

cfgov/unprocessed/apps/analytics-gtm/js/bah-loan-estimate-listeners.js
anselmbradford: I updated this to `addEventListenerToSelector`

anselmbradford:
@sebworks @jimmynotjim I pushed up an update that makes the explainers share the same script logic since only the tracking label is different between them.

cfgov/unprocessed/apps/analytics-gtm/js/bah-closing-disclosure-listeners.js
anselmbradford: The analytics should support a much wider net than what we use elsewhere and not necessitate polyfills. MutationObserver does throw errors in New Relic once in awhile, though that may mean the expandables aren't working anyway. As mentioned in another comment I think what we want is the architecture to communicate between components so that analytics can just observe what events are being emitted instead of snooping in the dom.

screen shot 2018-05-04 at 1 09 40 pm

cfgov/unprocessed/apps/analytics-gtm/js/util/analytics-util.js
```diff
@@ -1,16 +1,13 @@
...
  * @param {Function} callback - The event handler.
  */
-function addEventListenerToElems( elems, event, callback ) {
-  let elem;
-  for ( const i in elems ) {
-    elem = elems[i];
-    if ( elem.addEventListener ) {
-      addEventListenerToElem( elems[i], event, callback );
-    }
+function addEventListenerToSelector( selector, event, callback ) {
```
jimmynotjim: I really like this change 👍

cfgov/unprocessed/apps/analytics-gtm/js/bah-explainers.js
```diff
@@ -0,0 +1,138 @@
...
+    const expandable = elem.parentNode;
+    const expandableID = expandable.id;
+    if ( isAnimatingExpandable( expandable ) ) { return; }
+    recordExpandableState( expandableID );
+
+    const tab = closest( elem, '.explain' ).querySelector( '.active-tab' );
+    const tabText = tab.querySelector( '.tab-label' ).textContent.trim();
+    let action = 'Expandable collapsed - ' + tabText;
+    const label = elem.querySelector( '.o-expandable_label' );
+    let text = '';
```
jimmynotjim: Why `let` and define on the next line? Doesn't look like this gets redefined at all.

cfgov/unprocessed/apps/analytics-gtm/js/util/analytics-util.js
anselmbradford: 👍 I'll get it in a PR after this one. I think I could actually refactor it away.

cfgov/unprocessed/apps/analytics-gtm/js/bah-explainers.js
anselmbradford: hm, yeah not sure what happened there. Fixed!

PR #3841, opened 2018-03-01
Upgrade Wagtail to 1.13.1

willbarton:
This PR bumps our Wagtail version to 1.1.13.1. 

Interesting things we've missed out on so far (that we get now):

- *New React-based admin!*
- Global unique identifiers on StreamField blocks
- URL caching throughout the request cycle with  `page.get_url(request=request)` instead of `page.url` (We'll have to chose to make use of this)

Possible concerns:

- StreamField defaults to `blank=False`. I haven't found a place where this affects us (in fact, we explicitly set `blank=True` where we seem to desire it already), but my testing has not been exhaustive
- Registering Hallo plugins with `registerHalloPlugin` is deprecated and replaced by rich text features. I have not replaced our use of `registerHalloPlugin` just yet — it's deprecated but not removed, and yes we should use configurable rich text features instead and it will be more future-friendly now that Draftail has replaced Hallo in Wagtail 2.0. But I think we should put our effort toward Python3/Django2/Wagtail2 and Draftail.

Full changelogs:

- http://docs.wagtail.io/en/v1.13.1/releases/1.11.html
- http://docs.wagtail.io/en/v1.13.1/releases/1.12.html
- http://docs.wagtail.io/en/v1.13.1/releases/1.13.html

This PR is holding for more thorough testing and input on opinionated choices I've made and described.

## Notes

- I modified the Data & Research conference registration form test `test_returns_form_block` because of the StreamBlock unique ids introduced in 1.12. These mean that a rendered StreamBlock isn't going to be equatable to a fixture. 

## Todos

- More thorough testing.

## Checklist

* [ ] PR has an informative and human-readable title
* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [development playbook](https://github.com/cfpb/development)
* [ ] Passes all existing automated tests
* [ ] Any *change* in functionality is tested
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated
* [ ] Reviewers requested with the [Reviewer tool](https://help.github.com/articles/about-pull-request-reviews/) :arrow_right:

PR #4089, opened 2018-05-04
Revert "Upgrade Wagtail to 1.13.1"

richaagarwal:
Reverts cfpb/cfgov-refresh#3841

This upgrade introduced some errors (see https://github.com/cfpb/cfgov-refresh/pull/4084) however that PR is not ready.  Since we are avoiding deploying to production as a result, I think we should revert the upgrade until the fixes are ready.

### DIFF
--- a/cfgov/data_research/fixtures/conference_registration_page.json
+++ b/cfgov/data_research/fixtures/conference_registration_page.json
@@ -3,7 +3,6 @@
     "fields": {
         "locked": false,
         "title": "Conference registration test",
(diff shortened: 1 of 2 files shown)


chosak:
@richaagarwal I'd like to close this PR as the fix in #4084 should have resolved the issues we were seeing with the upgrade. See also platform#2725. Any objection?

richaagarwal:
@chosak Thanks for taking care of that! Closed.

PR #4091, opened 2018-05-04
Remove gulp-mq-remove | Update gulp-clean-css

anselmbradford:
I noticed `gulp-mq-remove` does not appear to be used. 

## Removals

- Removes `gulp-mq-remove`.

## Changes

- Updates `gulp-clean-css` from `3.9.3` to `3.9.4`.

## Testing

1. `npm install && gulp clean && gulp build` should pass.


jimmynotjim:
@anselmbradford weren't we using this for IE8 styles? Don't know when we stopped but seems 👍 to me

anselmbradford:
It was replaced with `postcss-unmq` in https://github.com/cfpb/cfgov-refresh/pull/3599, but I guess I forgot to remove it from package.json.

PR #4098, opened 2018-05-07
Switch to gulp-uglify-es

anselmbradford:
The spanish language pages are using gulp-uglify, but they are throwing an error after uglifying. I switched to gulp-uglify-es (es6 support) and it appears to have solved the issue.

## Changes

- Swap gulp-uglify with gulp-uglify-es.

## Testing

1. `npm install`
2. `gulp clean && gulp build`
3. Visit consumerfinance.gov/es/obtener-respuestas/  and not there's an error in the dev console and search doesn't appear.
4. Visit http://localhost:8000/es/obtener-respuestas/ and note there is no error in the dev console and the search DOES appear.

## Screenshots

The error:
screen shot 2018-05-07 at 6 02 50 pm

## Note
The search doesn't appear to be working. I enter `¿Cómo puedo revisar mi informe de crédito?` and it says `No pudimos encontrar resultados para "None"`. I'm guessing this is an existing issue? Who's familiar with how this search works?

### DIFF
--- a/gulp/tasks/scripts.js
+++ b/gulp/tasks/scripts.js
@@ -14,7 +14,7 @@ const gulpModernizrBuild = require( 'gulp-modernizr-build' );
 const gulpNewer = require( 'gulp-newer' );
 const gulpRename = require( 'gulp-rename' );
 const gulpReplace = require( 'gulp-replace' );
-const gulpUglify = require( 'gulp-uglify' );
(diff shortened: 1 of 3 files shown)


sebworks:
I don't think this is the issue, It worked before our last [deploy](https://web.archive.org/web/20180504212623/https://www.consumerfinance.gov/es/obtener-respuestas/).  We should figure out what the issue was and acceptance test for it.

PR #4101, opened 2018-05-08
Add email signup to full width text module

virginiacc:
On the Buying a House project, we needed a way to make email signups more visible on the journey browse pages (currently the only option is to add them in the sidefoot). This adds the ability to create email signups as insets in the FullWidthText module. (See Hubcap issue 84 for more details)

## Additions

- Email signup in full width text module
- New H3 header & inline disclaimer options in email signup module

## Notes
- We decided to make the new header/disclaimer options available in the existing email signup module instead of creating a new module for a couple of reasons -- there might be future changes to email signup headers/disclaimers across the board, and it also might be useful in the short term to have the option for inline disclaimers in sidebar signups. 
- The email signup JS isn't currently added to the page when a signup is created in the full width text module. There's a PR open in wagtail-inventory that should fix this

## Testing

1. Check out branch
2. `python cfgov/manage.py migrate` and `gulp styles`
3. In admin, create a new browse page & add a full width text module with an email signup followed by a content block
4. Verify that the signup displays as expected on page and that disclaimer/header option toggles work as expected 
5. Verify that existing sidebar signup modules get the slug header/stacked disclaimer (the defaults for these options should mean that current signups are unchanged visually)

### Changes to email signup module in admin
screen shot 2018-05-08 at 12 16 15 pm

### Signup in full width text module with H3 & inline disclaimer

screen shot 2018-05-08 at 8 29 50 am

## Checklist

### Browsers

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [x] iOS Safari
- [x] Chrome for Android

### Accessibility

- [x] Keyboard friendly
- [x] Screen reader friendly

### Other

- [x] Is useable without CSS
- [x] Is useable without JS


cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -145,9 +145,34 @@
         border-bottom: 1px solid @horizontal-rule;
         margin-bottom: unit( ( @grid_gutter-width / 2 ) / @base-font-size-px, em );
     }
+
+    &__email {
```
sebworks: It doesn't have to happen here, but this probably should be a organism.

cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -145,9 +145,34 @@
...
+            border: 1px solid @block__border-top;
+            margin-bottom: unit( @grid_gutter-width / @base-font-size-px, em );
+            background-color: @block__bg;
+
+            .respond-to-range(@bp-sm-min, @bp-sm-max {
+                width: 100%;
+            });
+
+            .a-btn {
+                width: auto !important;
```
sebworks: why do you we need to use important here?

cfgov/unprocessed/css/molecules/form-field-with-button.less
```diff
@@ -30,12 +30,40 @@
...
 // TODO: Look into merging email-signup and form-field-with-button styles
 // Form field with button isn't used anywhere else
 .m-form-field-with-button {
     .m-form-field {
         margin-bottom: unit( (@grid_gutter-width / 2) / @base-font-size-px, em );
     }
+
+    // Note: inline info display breakpoints are set based on length of
+    // the default email signup disclaimer ('Privacy Act statement')
+     &_info {
```
sebworks: If we modify the copy, would we need to change the breakpoints?

virginiacc: @sebworks We would, although I think the copy is stable at this point. As I understand it, we generally use a `Privacy Act statement` link here, with the exception of a few cases where we substitute a sentence summarizing the statement text. 

We were thinking a future iteration (after more discussion about links vs. summary) might be to hard-code the text and possibly allow users to select the link location (it sounds like there are a couple of variations on the statement).

The breakpoints are specific to the email signup module, though, and the form-field-with-button is ostensibly generic (although currently only used in email signups). That raises questions about how to handle email signup specific code (like these breakpoints, or a change to the disclaimer field).

cfgov/unprocessed/css/molecules/inset.less
virginiacc: Looks like it isn't necessary -- removed in https://github.com/cfpb/cfgov-refresh/pull/4101/commits/af4e72470bbcff1d8fc444d93342d027b8824dd7

cfgov/unprocessed/css/molecules/inset.less
virginiacc: Should just the email inset be an organism?

cfgov/unprocessed/css/molecules/form-field-with-button.less
sebworks: yeah, I don't think the breakpoints should be contingent on the copy. I would try and use flexbox or something else.

sebworks:
👍, looks good to me other than the breakpoint issue.

cfgov/unprocessed/css/molecules/form-field-with-button.less
virginiacc: @sebworks The challenge is reversing the display order of the disclaimer and button, but I think I have it working with flexbox in https://github.com/cfpb/cfgov-refresh/pull/4101/commits/5c52c967862e2ffd57bcbdbd0ae72d744d3f3546.

PR #4116, opened 2018-05-10
Let smoketests pass if only one timeout occurs

higs4281:
Our smoketests sometimes fail because the initial test can time out
 while servers are restarting. This PR allows one timeout to occur
 without failing the build.

Also added a test of the test.

### DIFF
--- a/cfgov/scripts/http_smoke_test.py
+++ b/cfgov/scripts/http_smoke_test.py
@@ -223,11 +223,11 @@ def check_urls(base, full=False):
 
     if failures:
         logger.error("These URLs failed: {}".format(failures))
-    if timeouts:
+    if len(timeouts) > 1:
         logger.error("These URLs timed out after {} seconds: "
                      "{}".format(TIMEOUT, timeouts))
 
-    if failures or timeouts:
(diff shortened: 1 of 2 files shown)


cfgov/scripts/http_smoke_test.py
```diff
@@ -223,11 +224,11 @@ def check_urls(base, full=False):
 
     if failures:
         logger.error("These URLs failed: {}".format(failures))
-    if timeouts:
+    if len(timeouts) > ALLOWED_TIMEOUTS:
```
hillaryj: Let's output the URLs that timed out even if it's okay for one to time out.

Another option is to make a note that "%d timeouts are allowed due to httpd setup time, and the following URLs timed out but are allowed".

higs4281: √
Added logging of the timeout list.

Individual timeouts were already logged (from build 1335/console:):

```
09:58:58 + python http_smoke_test.py --base http://awswebl01 -v
09:58:58 /
09:59:28 http://awswebl01/ timed out
```

hillaryj: LGTM!

PR #4114, opened 2018-05-09
Prepare the project for SVG icons

jimmynotjim:
Updates all the files required for the SVG icon changes without actually
converting to the new SVG icons. Includes support for both the webfont
and SVG icons.

Co-authored-by: Scott Cranfill <scott.cranfill@cfpb.gov>
Co-authored-by: Andy Chosak <andy@chosak.org>

__I'm breaking this out of the big Icons pr to make it easier to review. Look forward to more "bitesized" PRs in place of that one.__

## Additions

- Adds mockfiles for both Python and JS tests

## Changes

- Updates Dockerfile to properly serve SVGs
- Updates Gulp task to support both Webfont and SVG icons
- Updates CF deps

## Testing

1. Run `npm install`
1. Run `tox -e fast`
1. Run `gulp test:unit`

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows the standards laid out in the [development playbook](https://github.com/cfpb/development)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [x] Project documentation has been updated
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Internet Explorer 8, 9, 10, and 11
- [ ] Edge
- [ ] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS
- [ ] Flexible from small to large screens
- [ ] No linting errors or warnings
- [ ] JavaScript tests are passing

PR #4120, opened 2018-05-10
Task per environment

anselmbradford:
If we only ran gulp tasks in the environments they are needed we would have a number of dependencies that wouldn't need to be installed. This PR adds the architecture to ./frontend.sh and gulp to allow segregating tasks to specific node environments.

## Removals

- `require-dir` dependency.

## Changes

- Moves dev dependencies that don't run in production environment to devDependencies in package.json.
- Updates gulpfile to introduce logic to only load tasks files that are called (except in top-level build, test, and default tasks).
- Updates gulpfile to support ignoring tasks from build based on environment.
- Updates `frontend.sh` to support rebuilding node_modules when switching environments.

## Testing

Testing development:
1. Run `./frontend.sh` (or `./frontend.sh development`, which is the default).
2. The npm deps are installed in node_modules, and a file `node_modules/NODE_ENV` is added, which will have the contents `development`.
3. Run `./frontend.sh` again. The npm modules should NOT install again and use the existing folder ("Dependencies are up to date." should be in the console log).
7. Run `echo $NODE_ENV`. It should output "development".
8. Run `gulp`. It should list all the gulp tasks in `gulp/tasks/` in the format `Requiring task ./gulp/tasks/[taskname].js`.
9. Run `gulp build`. It should show the same required tasks as above.
10. Run `gulp lint`. It should ONLY list `Requiring task ./gulp/tasks/lint.js` at the top of the output.
11. Run `gulp test:unit`. It should ONLY list `Requiring task ./gulp/tasks/test-unit.js` at the top of the output.

Testing gulp:
`gulp --tasks` outputs the tasks that can be run.
1. `gulp` should load and run all task files.
2. `gulp build` should load and run all task files (or a smaller subset when in production environment).
3. `gulp lint` should load only `lint.js` task.
4. `gulp test` should load all task files and run tests.
5. `gulp test:unit` should load only `test-unit.js` task.
6. `gulp test:unit:scripts` should load only `test-unit.js` task.

Testing production:
1. Run `./frontend.sh production`
2. The npm deps are installed in node_modules and a file `node_modules/NODE_ENV` is added, which will have the contents `production`.
3. Run `./frontend.sh production` again. The npm modules should NOT install again and use the existing folder.
4. Run `echo $NODE_ENV`. It should output "production".
5. Run `gulp`. It should list all the gulp tasks in `gulp/tasks/` in the format `Requiring task ./gulp/tasks/[taskname].js` EXCEPT audit, docs, lint, test-unit.
6. Run `gulp build`. It should show the same required tasks as above.
7. Run `gulp styles`. It should ONLY list `Requiring task ./gulp/tasks/styles.js` at the top of the output.
8. Run `gulp scripts:apps`. It should ONLY list `Requiring task ./gulp/tasks/scripts.js` at the top of the output.

## Notes

- For easier review, note that the test gulp tasks are unchanged, they are just split into two files to separate the unit and acceptance test code.

- I removed the `$NODE_DIR` variable from `frontend.sh`. It doesn't seem realistic that we wouldn't be using node_modules for this, so why not reference the path directly.

- Top-level tasks are in gulpfile.js. These tasks don't define their own tasks but instead call other tasks. We had `test.js` calling the lint task, which I don't think is right. Each task file should only be calling its own tasks, therefore this PR moves test to be top-level. The consequence of this is it acts like build and the default task in that it loads all gulp tasks. This is no worse than how it was before, it just means the test task doesn't utilize the optimizations that test:unit or test:acceptance use.


package.json
```diff
@@ -57,16 +56,10 @@
...
     "imagemin-svgo": "5.2.4",
-    "is-reachable": "2.4.0",
     "lightbox2": "2.10.0",
-    "localtunnel": "1.9.0",
     "lodash.throttle": "4.1.1",
     "postcss-unmq": "1.0.2",
-    "psi": "3.1.0",
-    "require-dir": "1.0.0",
     "sauce-connect-tunnel": "0.2.1",
-    "svg-inline-loader": "0.8.0",
```
anselmbradford: I removed this as I couldn't find a reference to it in the codebase. This will be needed in the SVG icons per https://github.com/cfpb/cfgov-refresh/pull/4114. Please add it back when it's used and put it in the `devDependencies` or `dependencies`, depending on which environment needs it (development or production). (cc @jimmynotjim)

frontend.sh
```diff
@@ -12,83 +12,101 @@ set -e
...
+write_checksum() {
+  echo -n "$DEP_CHECKSUM" > node_modules/CHECKSUM
+  echo "Wrote node_modules/CHECKSUM $DEP_CHECKSUM"
+}
+
+# Write file that says the node environment that we're in.
+# We can read this on next run to see if the checksum should be bashed.
+write_node_env() {
+    echo -n "${NODE_ENV}" > "node_modules/NODE_ENV"
+    echo "Wrote node_modules/NODE_ENV $NODE_ENV"
```
jimmynotjim: Minor, extra indentation here.

gulpfile.js
```diff
@@ -5,30 +5,121 @@
...
+ * @returns {string|undefined} File name that exists, otherwise undefined.
+ */
+function fileExists( filePattern ) {
+  let UNDEFINED;
+  if ( !filePattern ) {
+    return UNDEFINED;
+  }
+  const checkFile = `${ TASK_PATH }${ filePattern }.js`;
+  // eslint-disable-next-line no-sync
+  const doesFileExist = fs.existsSync( checkFile );
```
jimmynotjim: Since this isn't reused at all, is it worth moving the execution to the condition? It would clean up the repetitive naming a bit.

gulpfile.js
```diff
@@ -5,30 +5,121 @@
...
+  if ( doesFileExist ) {
+    return checkFile;
+  }
+  const newFile = filePattern.split( '-' );
+  newFile.pop();
+  return fileExists( newFile.join( '-' ) );
+}
+
+/*
+Check for command-line flag option (such as `lint` in `gulp lint`).
```
jimmynotjim: Minor, these lines should be prepended with a `*`

gulpfile.js
```diff
@@ -5,30 +5,121 @@
...
+const cliOption = process.argv.slice( 2 );
+let taskName;
+if ( cliOption.length > 0 ) {
+  taskName = cliOption[0].replace( /:/g, '-' );
+}
+const taskFile = fileExists( taskName );
+
+if ( taskFile ) {
+  requireTask( taskFile );
+} else {
```
jimmynotjim: It's difficult to follow where this ends, is it possible either to break out at L:79 or to move the contents here into it's own function and only execute that function within this half of the condition?

package.json
jimmynotjim: Whoops, good call. I'll make sure to add it back when it's used in the JS modules

gulpfile.js
anselmbradford: Fixed

gulpfile.js
anselmbradford: Fixed

frontend.sh
anselmbradford: Fixed

gulpfile.js
jimmynotjim: Did you push that change?

anselmbradford: blargh, check now

package.json
```diff
@@ -57,16 +56,10 @@
...
     "imagemin-optipng": "5.2.1",
     "imagemin-svgo": "5.2.4",
-    "is-reachable": "2.4.0",
     "lightbox2": "2.10.0",
-    "localtunnel": "1.9.0",
     "lodash.throttle": "4.1.1",
     "postcss-unmq": "1.0.2",
-    "psi": "3.1.0",
-    "require-dir": "1.0.0",
     "sauce-connect-tunnel": "0.2.1",
```
sebworks: This can probably be moved as well.

jimmynotjim:
`echo $NODE_ENV` didn't output anything for me. Everything else seems to be working.

jimmynotjim:
Oh wait, `gulp --tasks` is also outputting an error

```
|task_per_env {12} U:1 ✗| → gulp --tasks
module.js:529
    throw err;
    ^

Error: Cannot find module 'is-reachable'
    at Function.Module._resolveFilename (module.js:527:15)
    at Function.Module._load (module.js:476:23)
    at Module.require (module.js:568:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/Users/wilsong/repos/cfpb/cfgov-refresh/gulp/tasks/audit.js:5:21)
    at Module._compile (module.js:624:30)
    at Object.Module._extensions..js (module.js:635:10)
    at Module.load (module.js:545:32)
    at tryModuleLoad (module.js:508:12)
    at Function.Module._load (module.js:500:3)
```

package.json
anselmbradford: It's used by the acceptance tests, which are run in the production environment, so has to stay in dev-deps.

anselmbradford: Err I mean deps

anselmbradford:
> Oh wait, gulp --tasks is also outputting an error

Yeah, that's expected behavior, so you're running `gulp --tasks` in the production environment, which is going to be looking for dev-dependencies too. I don't think we have to worry about this because in production `gulp --tasks` is not going to be run, and locally we should be working in the dev environment and can run `gulp --tasks` from there and should only be running in the production environment to check that everything builds.

jimmynotjim:
Ah, maybe move that above the production testing steps so it doesn't come up for anyone else.

anselmbradford:
> Ah, maybe move that above the production testing steps so it doesn't come up for anyone else.

Whoops, good point! Done.

package.json
anselmbradford: Wait… what test commands are run on Jenkins?

package.json
jimmynotjim: @anselmbradford the PR that's using this was merged. Can you remove this change please.

anselmbradford: It's been moved to dev-deps, see https://github.com/cfpb/cfgov-refresh/pull/4120/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R91

package.json
anselmbradford: Okay, this is updated to not have the test task in the production environment. The discussion can continue in https://github.cfpb.gov/CFGOV/platform/issues/2709 to set Travis to the production environment + tests.

frontend.sh
```diff
@@ -12,83 +12,101 @@ set -e
...
     npm install --production --loglevel warn --no-optional
   fi
 }
 
+# Calculate checksum value.
+calc_checksum() {
+  if [ -f "package-lock.json" ]; then
+    DEP_CHECKSUM=$(cat package*.json cfgov/unprocessed/apps/**/package*.json | shasum -a 256)
+  else
+    DEP_CHECKSUM=$(cat package.json cfgov/unprocessed/apps/**/package.json | shasum -a 256)
```
sebworks: We can probably just use 

`DEP_CHECKSUM=$(cat package*.json cfgov/unprocessed/apps/**/package*.json | shasum -a 256)`

and get rid of the if / else statment.

PR #4126, opened 2018-05-11
Remove enterprise attribution cookies script

anselmbradford:
This analytics script was complicated and under utilized by the analytics team and after discussing with @kelleyholden we decided to remove it. It also caused errors in MS Edge (reported in New Relic), as well as was flagged for the setting of a client-side cookie by Checkmarx (reported by @lfatty).

## Removals

- `enterprise-attribution-cookies.js` analytics cookie tracking.


jimmynotjim:
@ans was this file not executed anywhere?

anselmbradford:
It's used by google tag manager, but we've removed the tag that uses it.

PR #4129, opened 2018-05-14
Mega Menu: Hack for WAVE validation empty button error

anselmbradford:
## Changes

- Adds a period to the mege menu's tab trigger as a hack to pass the WAVE validation.

## Testing

1. Pull this branch.
2. Check WAVE extension and see that there is one less error on the page vs master branch.

## Notes

- Not sure if this is the best way to satisfy the content requirement, but the idea was if for some reason this button was shown, it would only be a period and wouldn't otherwise visually disrupt things. 

## Todos

- Tab trigger should be its own module. https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/js/molecules/GlobalSearch.js#L30

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
+++ b/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
@@ -193,5 +193,13 @@
     {# Create a root menu at depth one.
        This is the 1st level of a 3-level menu. #}
     {{ _nav( 1, vars.nav_groups ) }}
-    <button class="{{ base_class ~ '_tab-trigger' }}" aria-hidden="true"></button>
+    <button class="m-global-search_tab-trigger u-visually-hidden"
+            aria-hidden="true">
+      .
+      {# The tab trigger is hidden, but is used to listen for keyup events
+      when the tab key is pressed on it while it has focus, so that
+      the menu can be collapsed.
+      The period is because HTML validation requires the button
+      contain some content. #}
+    </button>
 </nav>


marteki:
I don't think this would meet the content requirement or meet the accessibility requirement because the text isn't meaningful or descriptive of the button function when read out loud by screen readers.

From the documentation/more context tab of the WAVE extension, paying special attention to the "Why It Matters"" area:
> When navigating to a button, descriptive text must be presented to screen reader users to indicate the function of the button.

screen shot 2018-05-14 at 9 58 18 am

If there's a concern about visual disruption or what content should be in the text, let's talk about what is possible and what we want to do first before pushing code. (I think this might end up being another one of the mega menu fixes @schaferjh and @ascott1 called out that will need UX and design attention before devs could pick up work for it. Sigh. 😞 )

anselmbradford:
So the `aria-hidden="true"` should tell screen readers to ignore the button content. So these buttons should be visually hidden and hidden from screen readers, and the period is just to satisfy WAVE not being smart enough to pick that these use our `u-visually-hidden` class. Some other content is fine too, I just wasn't sure what to put here and went with a period so if for some reason the styles didn't apply they wouldn't inject a bunch of content into the page.

Scotchester:
Setting aside for now the fact that the loss of focus is still the root problem, is there any reason this element should be a `button` in the first place? Why not just make it a `div` or something and give it `tabindex="0"` to make it able to be tabbed to?

anselmbradford:
> Setting aside for now the fact that the loss of focus is still the root problem, is there any reason this element should be a button in the first place? Why not just make it a div or something and give it tabindex="0" to make it able to be tabbed to?

That's a good question. There may have been a technical reason it was a button but when it was created, but I see no reason not to explore refactoring it to not be one (and to also make it it's own component that can be shared with components that need this behavior). This PR would just take care of the validation error, but it should be looked at closer.

cfgov/jinja2/v1/_includes/molecules/global-search.html
```diff
@@ -87,6 +87,14 @@
                 </ul>
             </div>
         </form>
-        <button class="m-global-search_tab-trigger" aria-hidden="true"></button>
+        <button class="m-global-search_tab-trigger u-visually-hidden"
+                aria-hidden="true">
+          .
```
marteki: Change text to read "Collapse" instead of ".", and adjust the code comment for it.

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
```diff
@@ -193,5 +193,13 @@
     {# Create a root menu at depth one.
        This is the 1st level of a 3-level menu. #}
     {{ _nav( 1, vars.nav_groups ) }}
-    <button class="{{ base_class ~ '_tab-trigger' }}" aria-hidden="true"></button>
+    <button class="m-global-search_tab-trigger u-visually-hidden"
+            aria-hidden="true">
+      .
```
marteki: Change text to read "Collapse" instead of ".", and adjust the code comment describing it.

cfgov/jinja2/v1/_includes/molecules/global-search.html
anselmbradford: Fixed

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
anselmbradford: Fixed

PR #4134, opened 2018-05-14
Fix 'ssl' mode in standalone runserver.sh

acrewdson:
The args were not being passed, so you could not run your `runserver.sh` in ssl mode. I'm guessing this might have broken when Docker support was added.

(Could not help attempting to fix the ANSI escape code output for the 'Starting ... ' line while I was in there, though if that was not broken for anyone else, let me know and I can remove that commit; I was seeing the code itself, not the color, without `-e`.)

### DIFF
--- a/runserver.sh
+++ b/runserver.sh
@@ -10,6 +10,8 @@
 set -e
 
 standalone() {
+  local DJANGO_HTTP_PORT=${DJANGO_HTTP_PORT:-8000}
+
   if [ "$1" == "ssl" ]; then
     echo -e '\033[0;32mStarting SSL Django server on port' $DJANGO_HTTP_PORT '...'
     tput sgr0
--- a/runserver.sh
+++ b/runserver.sh
@@ -11,10 +11,12 @@ set -e
 
 standalone() {
   if [ "$1" == "ssl" ]; then
-    echo '\033[0;32mStarting SSL Django server on port' $DJANGO_HTTP_PORT '...'
+    echo -e '\033[0;32mStarting SSL Django server on port' $DJANGO_HTTP_PORT '...'
+    tput sgr0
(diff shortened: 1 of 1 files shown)


higs4281:
> without -e

Thank you, @acrewdson    
It appears we lost the color when the shebang was changed from `#!/bin/sh` to `#!/bin/bash`

runserver.sh
```diff
@@ -11,10 +11,12 @@ set -e
 
 standalone() {
   if [ "$1" == "ssl" ]; then
-    echo '\033[0;32mStarting SSL Django server on port' $DJANGO_HTTP_PORT '...'
+    echo -e '\033[0;32mStarting SSL Django server on port' $DJANGO_HTTP_PORT '...'
```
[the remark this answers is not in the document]
acrewdson: @chosak great idea, will do

acrewdson:
@higs4281 ah, that makes sense, thanks for confirming 👍

runserver.sh
```diff
@@ -10,11 +10,15 @@
 set -e
 
 standalone() {
+  local DJANGO_HTTP_PORT=${DJANGO_HTTP_PORT:-8000}
```
acrewdson: idk if this kind of intentional shadowing is kosher in bash, but I'm OK with it

chosak: Looks good to me. You can do all sorts of fun stuff with variables like this in bash. Another neat trick is to fail the script if an expected environment variable isn't defined:

```bash
#!/usr/bin/env bash
: "${DJANGO_HTTP_PORT?This message will print and the script will exit if not defined}"
```

acrewdson:
This branch is going to have conflicts with another branch I'm working on, so I will close this and move the changes over to that branch.

PR #4138, opened 2018-05-14
Misc local dev tweaks

acrewdson:
This PR got slightly out of hand. Please review commits individually.


.python_env_SAMPLE
```diff
@@ -29,3 +29,6 @@ MYSQL_ROOT_PW=root
 MYSQL_NAME=v1
 MYSQL_USER=v1
 MYSQL_PW=v1
+
+# Password given to 'admin' user created by initial-data.sh script.
+WAGTAIL_ADMIN_PW=admin
```
chosak: In the interest of simplification, how would you feel about deprecating the use of this environment variable? That would mean modifying this PR to remove this change, removing its definition in the `.env_SAMPLE` [here](https://github.com/cfpb/cfgov-refresh/blob/master/.env_SAMPLE#L28), and hardcoding in in the initial data script [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/scripts/initial_data.py#L29).

Benefits:

1. Currently, if you don't define the `WAGTAIL_ADMIN_PW` environment variable, the initial data script invokes `make_password(None)`, which returns a valid hash, but then the user is left unable to login. This means the script can potentially be run leaving behind a broken admin user.
2. We hardcode the admin username as `admin`; hardcoding the admin password as `admin` would only simplify and make it easier to use locally.
3. This would let us simplify the documentation. Currently the docs instruct users to [create a superuser](https://cfpb.github.io/cfgov-refresh/usage/#3-load-indexes-launch-site), which isn't necessary, and also [mention](https://cfpb.github.io/cfgov-refresh/installation/#load-initial-data-into-database) the `WAGTAIL_ADMIN_PW` environment variable. Instead we could simplify and say that if `initial-data.sh` is run, you'll have an `admin` / `admin` superuser.

Drawbacks:

1. Limited configurability around local admin user password?

acrewdson: I like this idea and will update the PR accordingly. Thanks @chosak

docs/installation.md
```diff
@@ -182,10 +177,12 @@ Get any errors? [See our troubleshooting tips.](#troubleshooting)
   (a way of running software in an isolated environment) and "images" (a
   snapshot of all of the files neccessary to run a container).
 - **Docker Compose**: Compose allows you to configure and run a collection of
-  connected containers (like a web application and it's database)
-- **Docker Machine**: Docker only runs natively on Linux and Windows. On OS X,
-  we'll use Docker Machine to start the Docker server in a virtual linux
-  environment (using Virtualbox)
+  connected containers (like a web application and its database)
```
chosak: Nitpicky, but missing `.` at the end of this sentence.

docs/installation.md
```diff
@@ -182,10 +177,12 @@ Get any errors? [See our troubleshooting tips.](#troubleshooting)
...
+- **Docker Machine**: In environments where Docker Engine is not supported, or
+  where it is not approved for use, Docker Machine can be used to create and
```
chosak: Perhaps "or where it is not available" is a more general way to describe this. :smile:

docs/installation.md
```diff
@@ -182,10 +177,12 @@ Get any errors? [See our troubleshooting tips.](#troubleshooting)
...
+  manage Docker hosts on virtual machines. For more information on Docker
+  Machine, see the 'How do I use Docker Machine' section in the
+  [usage guide](usage).
```
chosak: Can we use [the section link](https://cfpb.github.io/cfgov-refresh/usage/#usage-docker) here?

docs/installation.md
```diff
@@ -123,13 +123,8 @@ Where `[GHE]` is our GitHub Enterprise URL.
...
 
-If this is the first time you're setting up the project, run the following
-script to copy `.env_SAMPLE` to `.env`, export your environment variables,
-and activate your virtualenv for the first time.
-
-```bash
-source load-env.sh
-```
+If this is your first time setting up the project, copy `.env_SAMPLE` to
+`.env`, then `source .env` and activate your virtualenv.
```
chosak: For people who are less technical, it might be nice to include the explicit steps:

```sh
$ cp .env_SAMPLE .env
$ source .env
$ workon cfgov-refresh
```

docs/installation.md
```diff
@@ -198,14 +195,22 @@ If you are on a machine that is already set up to run Linux docker containers,
 please install [Docker Compose](https://docs.docker.com/compose/install/).
 If `docker-compose ps` runs without error, you can can go to step 2.
 
-#### Mac + Homebrew + Virtualbox quickstart
+#### Copy the `.python_env_SAMPLE` file over
 
-**Starting assumptions**: You already have homebrew and virtualbox installed.
+If this is your first time setting up the project, copy `.python_env_SAMPLE` to
+`.python_env`.
```
chosak: Similarly to above, it'd be nice to include the explicit command for less-technical users.

```sh
$ cp .python_env_SAMPLE .python_env
```

acrewdson:
@Scotchester thanks for the feedback, I will incorporate your suggestions

@schbetsy 👍

docs/installation.md
Scotchester: I don't know; that's why I asked :)

The man page says "Preserves structure and attributes of files but not directory structure," which I don't totally understand.

acrewdson: my guess is that it was was someone being overly-cautious, but I'll leave the `-a` in for now just to be safe

acrewdson:
@chosak @Scotchester PR has been updated based on your feedback

ascott1:
There's a lot of great stuff in here @acrewdson!

PR #4155, opened 2018-05-17
Adds accessible tagline atom

anselmbradford:
Fixes [GHE]/CFGOV/platform/issues/2610

Adds new accessible tagline atom that's shared between the header and footer.

## Additions

- Adds tagline atom.

## Changes

- Lowercases "government" in the tagline.

## Testing

1. Visit the homepage.
2. Resize to small screen size.
3. Check wrapping of tagline inside the hamburger menu at various sizes. It should generally lay out with US government all on one line.
4. Close hamburger menu.
3. Press command + f5 to enter VoiceOver.
4. Scroll to the footer.
5. click somewhere in the footer and press tab key.
5. Press ctrl + command + right arrow till you get to the tagline in the footer and verify it reads "An official website of the United States government"


Scotchester:
I can't figure out why the same non-breaking behavior exists in the footer version of this _without_ that span.

jimmynotjim:
Have you tried converting the `div` that wraps it all to a `p`? Just tried that locally and it seems to read it as one sentence.

Scotchester:
> How bad would it be if we let it break across two lines? I'll ask in the source issue.

For posterity, it was decided on GHE that we would stop worrying about having breaks within "United States government".

cfgov/unprocessed/css/molecules/global-eyebrow.less
```diff
@@ -40,7 +40,7 @@
           -----------------------
           .m-global-eyebrow.m-global-eyebrow__list
             .wrapper
-              .m-global-eyebrow_tagline
+              p.a-tagline
```
Scotchester: Extremely trivial, but the markup for this is using `div` still, not `p`. (I think `div` is the correct choice.)

anselmbradford: @jimmynotjim suggested `p`, though before I reworked this. Which would you'll like?

Scotchester: I think `div`, as it's not even a complete sentence.

jimmynotjim: Yeah, `div` seems fine and matched what we're doing with notifications. I only suggested `p` because it's read as one unit, even with inline elements breaking up the content within.

anselmbradford: Done!

PR #4159, opened 2018-05-17
Regs3K: Explicit subsections for nav items

higs4281:
The original secondary navigation code delivered a list of reg sections
to display, and reg subparts were extracted in-line as needed.

To enable expandable nav sections, subparts need to be available
separately.

To test the branch:
* load reg data with `./cfgov/manage.py loaddata regdata`
* check out <http://localhost:8000/regulations/1005/>

Subsections should show up as button headings in the side nav.


higs4281:
👍 This is much friendlier IMO.

PR #4162, opened 2018-05-17
Updates to support move to Postgres for local dev

acrewdson:
The main goal here is to shift the local dev setup/docs (both stand-alone and Docker-based) in the direction of Postgres and away from MySQL. Apologies for the slightly haphazard nature of the cleanup here – not all of it is strictly related to the task at hand.

### DIFF
--- a/.env_SAMPLE
+++ b/.env_SAMPLE
@@ -51,11 +51,15 @@ export ALLOW_ADMIN_URL=True
 export DATABASE_URL=sqlite:///db.sqlite3
 #export DATABASE_URL=postgres://cfpb@localhost/cfgov
 
-# The use of default_storage_engine here assumes use of MySQL 5.7. For
-# MySQL 5.5 or earlier, use storage_engine instead.
+#DEPRECATED:
 #export DATABASE_URL=mysql://root@localhost/v1?init_command='SET default_storage_engine=MYISAM'
 
-# Required for existing MySQL-specific database scripts.
+#########################################################################
+# Deprecated MySQL settings; we plan to leave these here during the MySQL to
+# Postgres transition period, in case there is a need for local MySQL-based
+# testing. Once we have fully moved to Postgres, we can remove this section.
+#########################################################################
+
 #export MYSQL_NAME=v1
 #export MYSQL_HOST=localhost
 #export MYSQL_USER=root
--- a/.python_env_SAMPLE
+++ b/.python_env_SAMPLE
@@ -14,18 +14,17 @@
 #
 # See https://github.com/kennethreitz/dj-database-url for URL formatting.
 #########################################################################
-# DATABASE_URL=sqlite:///db.sqlite3
+DATABASE_URL=postgres://cfpb:cfpb@postgres/cfgov
 
-# DATABASE_URL=postgres://cfpb:cfpb@postgres/cfgov
-# Required for Postgres dbshell in Django < 1.9.
-# PGPASSWORD=cfpb
-
-# The use of storage_engine here assumes use of MySQL 5.1. For MySQL 5.6
-# or later, use default_storage_engine instead.
-DATABASE_URL=mysql://v1:v1@mysql/v1?init_command='SET storage_engine=MYISAM'
-# Required for existing MySQL-specific database scripts.
-MYSQL_HOST=mysql
-MYSQL_ROOT_PW=root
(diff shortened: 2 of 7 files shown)


.python_env_SAMPLE
```diff
@@ -14,18 +14,16 @@
...
-# Required for existing MySQL-specific database scripts.
-MYSQL_HOST=mysql
-MYSQL_ROOT_PW=root
-MYSQL_NAME=v1
-MYSQL_USER=v1
-MYSQL_PW=v1
+#########################################################################
+# Postgres env vars to make tools like `psql` easier to work with (and
+# in the case of PGPASSWORD, it also makes `manage.py dbshell` work without
+# a password in Django versions older than 1.9.)
```
acrewdson: I think we need `PGPASSWORD` for `manage.py dbshell` to work inside the container because the db is on another "host" (container), causing it to prompt for a password. In stand-alone, it's localhost, and I'm guessing the default PG setup does not use password auth for localhost, so it's not an issue (but as you can see above in `.env_SAMPLE` I'm still adding those `PG*` env vars to make it easy to just run `psql` and friends with no args).

.env_SAMPLE
```diff
@@ -51,11 +51,26 @@ export ALLOW_ADMIN_URL=True
...
+#########################################################################
+# Postgres env vars to make tools like psql easier to work with
+#
+# See https://www.postgresql.org/docs/current/static/libpq-envars.html
+#########################################################################
+
+#export PGHOST=localhost
+#export PGDATABASE=cfgov
+#export PGUSER=cfpb
+#export PGPASSWORD=cfpb
```
chosak: How would you feel about removing this line and encouraging use of a password-less `cfpb` user locally? See above on line 52 where we don't provide a password.

I realize that the Docker setup does use a password.

docs/installation.md
```diff
@@ -141,6 +141,49 @@ If you setup Autoenv earlier, this will happen for you automatically when you
...
+```bash
+brew services start postgresql
+```
+
+Then create the database and associated user (enter `cfpb` when prompted for a
+password):
+
+```bash
+dropdb --if-exists cfgov && dropuser --if-exists cfpb
+createuser -P cfpb && createdb -O cfpb cfgov
```
chosak: See above; if we remove this `-P` parameter then we don't need to provide a password for this user.

docs/installation.md
```diff
@@ -141,6 +141,49 @@ If you setup Autoenv earlier, this will happen for you automatically when you
...
+specifies SQLite as the db:
+
+```bash
+# export DATABASE_URL=sqlite:///db.sqlite3
+```
+
+And then uncomment the line that tells Django to use Postgres:
+
+```bash
+export DATABASE_URL=postgres://cfpb@localhost/cfgov
```
chosak: If we want to use a password, then this line and the line in `.env_SAMPLE` needs to include `:cfpb` to provide the password.

.env_SAMPLE
```diff
@@ -51,11 +51,26 @@ export ALLOW_ADMIN_URL=True
...
 #export DATABASE_URL=mysql://root@localhost/v1?init_command='SET default_storage_engine=MYISAM'
 
-# Required for existing MySQL-specific database scripts.
+#########################################################################
+# Postgres env vars to make tools like psql easier to work with
+#
+# See https://www.postgresql.org/docs/current/static/libpq-envars.html
+#########################################################################
+
+#export PGHOST=localhost
```
chosak: Looking at these lines again, I wonder about the need/value of including them here. Would it be better to just encourage people to use `cfgov/manage.py dbshell`, and stick to the single `DATABASE_URL` variable? I'm not sure the increased complexity of more variables is worth being explicit.

acrewdson: I'm all for reducing the number of environment variables here, so I'm glad to remove these in that case 👍

.env_SAMPLE
acrewdson: @chosak Good call, I think – as you point out – I did that just to make it match what's in `docker-compose.yml`, but let's go for password-less outside of Docker as the default, will fix 👍

docs/installation.md
acrewdson: @chosak great catch

acrewdson:
.@chosak gives the go-ahead to merge

PR #4175, opened 2018-05-22
Change navigation landmark label

anselmbradford:
## Changes

- Change mega menu aria-label to `main menu` from `main navigation`.
- Remove `role="navigation", which is covered by the HTML5 `nav` tag.

## Testing

1. Press Cmd+F5 to activate VoiceOver
1. Press Ctrl+Opt+U to activate rotor mode
1. Press the right arrow until you reach the Landmarks menu
1. Press the down arrow to cycle through and hear "main menu navigation"

## Notes

- We could remove the `aria-label` completely ([it's optional](https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/landmarks/navigation.html)), however, if we had another `role="navigation"` on the page we'd want to have a label, so safer to include it.

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
+++ b/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
@@ -185,8 +185,7 @@
 <nav class="{{ base_class }}
             u-hide-on-mobile"
      data-js-hook="behavior_flyout-menu"
-     aria-label="main menu"
-     role="navigation">
+     aria-label="main menu">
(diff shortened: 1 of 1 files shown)


Scotchester:
I'm not sure if it affects what a screen reader will read, but the `role="navigation"` is redundant on `nav` elements. Try removing it and see what's read with your change.

anselmbradford:
> I'm not sure if it affects what a screen reader will read, but the role="navigation" is redundant on nav elements. Try removing it and see what's read with your change.

Done!

PR #4179, opened 2018-05-22
Run ondemand subtask with default gulp scripts run

Scotchester:
Betsy spent more time than she should have trying to debug why the mega menu was open on top of the page on Paying for College. The reason was because we don't run `gulp scripts:ondemand` with a default `gulp scripts` run. Seems like we should just do that, otherwise, devs have to know that they have to run that ondemand subtask manually.

It's also worth noting that we include the ondemand styles in the default `gulp styles` run, so this makes the behavior consistent.

## Additions

- `gulp scripts:ondemand` subtask to `gulp scripts` default task

## Testing

1. `rm -rf cfgov/static_built`
1. `gulp build`
1. Visit http://localhost:8000/paying-for-college/ and see the mega menu open on top of everything
1. Pull branch
1. `gulp build`
1. Refresh the page and see the mega menu functioning normally

## Checklist

### DIFF
--- a/gulp/tasks/scripts.js
+++ b/gulp/tasks/scripts.js
@@ -262,6 +262,7 @@ gulp.task( 'scripts',
     'scripts:apps',
     'scripts:external',
     'scripts:nemo',
-    'scripts:spanish'
+    'scripts:spanish',
+    'scripts:ondemand'
   )
 );


anselmbradford:
The original reason this was run on demand was because you would run `gulp scripts:ondemand`, then fetch the generated scripts and manually copy them over to the satellite repo (cfgov-sheer-templates, Salesforce, etc.). So we didn't want this running all the time unnecessarily unless someone was taking the manual step to make an update outside of this repo. How exactly is this necessary to run now with Paying for College?

Scotchester:
That may have been the original intent of the `ondemand` tasks, but as of now, they are also what provides the nonresponsive CSS and JS to the pages that need them.

PFC [extends `base_nonresponsive.html`](https://github.com/cfpb/django-college-costs-comparison/blob/master/comparisontool/templates/comparisontool/landing.html#L1) from cfgov-refresh. `base_nonresponsive.html` calls the nonresponsive [CSS](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/legacy/templates/front/base_nonresponsive.html#L92-L93) and [JS](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/legacy/templates/front/base_nonresponsive.html#L190) from their standard location in `static_built`. If they're not actually built there, pages like PFC that use `base_nonresponsive.html` are broken.

PR #4181, opened 2018-05-22
Adds TabTrigger module

anselmbradford:
Addresses TODO to provide shared tabtrigger module to global search and mega menu. This needs work and isn't working as it should, but this PR moves the code to its own module so it can be refined outside of the global search and menu. 

## Additions

- TabTrigger module for issuing tab press events in global search and mega menu.

## Testing

1. `gulp clean && gulp build`.
2. Click the site logo in the header.
3. Hit tab key till search has focus.
4. Hit space key to open search.
5. Hit tab key till leaving the search and watch that it collapses back to the default state.

## Notes

- As mentioned, this needs work and probably not the right approach overall, but the module can now be adjusted in isolation. Is buggy in VoiceOver (not possible to re-expand search after first visit).


cfgov/unprocessed/js/modules/TabTrigger.js
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @classdesc Initializes a new MegaMenu organism.
+ *
+ * @param {HTMLNode} element
+ *   The DOM element within which to search for the organism.
+ * @returns {TabTrigger} An instance.
+ */
+function TabTrigger( element ) {
+
+  const _that = this;
```
sebworks: make this `_this`

anselmbradford: Changed, though note `_that` is used in the expandables. Probably should be updated.

PR #4185, opened 2018-05-24
Add a parser to harvest a regulation from eCFR xml

higs4281:
This is a first pass at converting eCFR XML to regdown-based eregs pages.

The PR creates a migration because it adds two fields to the EffectiveVersion model:
- draft
- acquired

These fields are intended to allow the import of the latest version of a Part 
without overriding current content. It could also be used to work on future 
versions of regulations without having them go live.

The Part model's `effective_version` method is now set to exclude draft versions 
before looking for the latest effective version to display.

The `acquired` field should help in managing draft versions by showing when they 
were imported from eCFR.

The script will re-use Part objects if they exist, and if re-run the same day,
will re-use version and subpart objects as well, creating what it needs to produce 
a regulation tree:  part > version > subpart > section.

## Testing
You can give it a whirl via the Django shell by passing in a part number:

```
./cfgov/manage.py runscript ecfr_importer --script-args 1004
```

## Notes
- Appendices and Interpretations are out of scope for this PR.
- The results of a full run of this parser can be viewed on the build server at /regulations/


willbarton:
I still think we should have some kind of award for the order**s** of magnitude improvement on the eCFR parsing that regulations-parser did.

cfgov/regulations3k/scripts/ecfr_importer.py
```diff
@@ -0,0 +1,282 @@
...
+                title=element.find('HEAD').text.strip(),
+                label=part.part_number,
+                version=PAYLOAD.version
+            )
+            PAYLOAD.subparts['section_subparts'].append(_subpart)
+            subpart_sections = element.find_all('DIV8')
+            parse_sections(subpart_sections, part, _subpart)
+
+
+def ital_to_bold(soup):
```
willbarton: This is such a stupid little nitpick that I hesitate to comment, but I think I'd prefer "italic" to "ital".

cfgov/regulations3k/scripts/ecfr_importer.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+def run(*args):
+    if len(args) not in [1, 2]:
+        logger.info(
+            "Usage: ./cfgov/manage.py runscript "
+            "ecfr_importer --script-args "
+            "[PART NUMBER] [OPTIONAL XML FILE PATH]")
+        sys.exit(1)
+    elif len(args) == 1:
```
willbarton: I wonder, since we have to download all of Chapter X anyway, could we make PART NUMBER optional, and if it's not given, import all parts?

higs4281: I do fear admin pollution, as an unwanted full download would require deleting versions for every reg, and stakeholders have expressed a desire to limit the authority to delete.  
Maybe get some feedback from stakeholders/editors on who has authority to download and who has authority to delete?

willbarton: That's a good point... are we going to hook this up to a button in the admin?

higs4281: I would vote for a button and mini API, but I think we need some feedback.

willbarton: button

grapesmoker:
It's not surprising that this code generates an orders-of-magnitude speedup; the old code was based on running literally all the text through PyParsing, which is absurdly slow partly because of the overhead generated by all the objects it instantiates. By avoiding that and solving a more limited problem, you win a lot of speed at the (minor, in my view) cost of some generality. For what it's worth, I once rewrote the major components of the parser in Lisp (lol) and was able to parse the entirety of the initial Reg Z notice (just the regtext and the interpretations, not amendments or section-by-section) in about 2 minutes.

PR #4193, opened 2018-05-29
Remove IE8 from browserlist

anselmbradford:
See [GHE]/CFGOV/platform/issues/2788

## Removals

- IE8 config from analytics-gtm browserlist config file.


cfgov/unprocessed/apps/analytics-gtm/browserslist
```diff
@@ -8,4 +8,3 @@
 
 > 0.5% in my stats # Sets a threshold to cut off extreme edge case browsers.
 > 5% in US         # Sets a threshold for possible future users.
-Explorer >= 8      # Sets IE8 support to address known New Relic errors.
```
Scotchester: Will this drop appropriate support for IE 9 or 10?

anselmbradford: That's a good question, so possibly it will if those are less than 0.5% of our visitors (or less than 5% in US, as a set). I will re-add this as `Explorer >= 9`.

anselmbradford: Added as `Explorer >= 9`

PR #4195, opened 2018-05-29
Workaround to restore Wagtail S3 file overwriting

chosak:
(See cfgov/v1/storage.py for a very lengthy comment that explains the need for this commit in great detail.)

Wagtail 1.13 introduced some behavior that broke support for Django file storage systems that overwrite saved files that have the same name, for example the django-storages S3 backend we use with the default `AWS_S3_FILE_OVERWRITE=True`.

This commit adds an admittedly very hacky workaround to restore support for this behavior in Wagtail 1.13+. It's not pretty, but it works.

I've opened a PR to Wagtail to restore this behavior the right way at https://github.com/wagtail/wagtail/pull/4567; this code in cfgov-refresh would only be necessary until that is merged in, assuming it is at some point. The unit tests in this commit are copied from that PR.

This change should fix GHE#2750. @Scotchester and I colllaborated on this work.

This commit also includes some minor refactoring and simplification of how S3-related settings are set, to make unit testing easier.

## Testing

To test this from a development environment you need to set the appropriate environment variables to run locally while using S3: `S3_ENABLED`, `AWS_S3_ACCESS_KEY_ID`, `AWS_S3_SECRET_ACCESS_KEY`, `AWS_S3_URL`. Once those are set, you should be able to run a local Wagtail server and upload documents/image to S3. You should then be able to test overwriting files by, say, creating a document and then re-uploading a different file that has the same filename.

## Checklist

### DIFF
--- a/cfgov/v1/storage.py
+++ b/cfgov/v1/storage.py
@@ -1,0 +1,120 @@
+import inspect
+
+from storages.backends.s3boto import S3BotoStorage
+
+
+class WagtailOverwritingStorageMixin(object):
+    """Django storage mixin to support file overwriting in Wagtail 1.13+.
+
+    The Django storage backend convention is that if a file is uploaded to a
+    storage backend with the same name as another file already stored on that
+    backend, the new file's name will be changed:
+
+    https://docs.djangoproject.com/en/2.0/ref/files/storage/#django.core.files.storage.Storage.get_available_name
+
+    Specifically, the new file will have random characters appended to its
+    filename (for example, a second upload of "test.txt" would become something
+    like "test_abcdefg.txt").
+
+    The convention is that typically storage backends will not support
+    deliberate overwriting of files that they store. However, there are some
+    backends that can support this behavior; for example if users desire to
+    maintain a consistent relationship between source filename and filename
+    in storage.
+
+    Wagtail 1.13 included changes that broke support for backends with this
+    overwriting behavior for both documents and images:
+
+    https://github.com/wagtail/wagtail/pull/3821
+    https://github.com/wagtail/wagtail/pull/3822
+
+    The goal of these changes was to clean up old document/image files after
+    the user uploads a new one. Consider an existing document uploaded from
+    some filename "test.txt"; the source file was "test.txt" and the file on
+    the storage backend will be "test.txt". If a user uses the Wagtail UI to
+    replace the document's file with a new file "test2.txt", the change in
+    1.13 tries to clean up the old file by deleting "test.txt" from the storage
+    backend after "test2.txt" has been stored.
+
+    The Wagtail logic looks roughly like this:
+
+        - user uploads a new file for an existing document/image
+        - store the new file to the storage backend
+        - delete the document/image's old file from the storage backend
+
+    This logic is sound as long as it is safe to assume that new uploads with
+    identical filenames never overwrite each other and always get assigned a
+    unique filename when stored. However, as described above, this may not
+    always be the case.
+
+    If, instead, files get overwritten, then the final delete that happens
+    will in fact delete the new file that's just been stored.
+
+    This mixin attempts to prevent this from happening. It uses the Python
+    stack inspect capability to identify if a delete call is being invoked
+    from the Wagtail document/image edit views, and, if so, skips that delete.
+    """
+    def delete(self, name):
+        stack = inspect.stack()
+
+        if self._should_skip_wagtail_delete(stack):
+            return
+
+        super(WagtailOverwritingStorageMixin, self).delete(name)
+
+    def _should_skip_wagtail_delete(self, stack):
+        """Given a call stack, determine whether delete should be skipped."""
+
+        # See Python inspect documentation for structure of stack object:
+        # https://docs.python.org/2/library/inspect.html#the-interpreter-stack
+        caller_frame_record = stack[1]
+        caller_frame_object = caller_frame_record[0]
+        caller_frame_function_name = caller_frame_record[3]
+
+        caller_module = inspect.getmodule(caller_frame_object)
+        caller_module_name = caller_module.__name__
+
+        called_from = caller_module_name + '.' + caller_frame_function_name
+
+        # Only skip delete if it is being called by the wagtaildocs.edit or
+        # wagtailimages.edit views. We want deletes called from other places
+        # to work as expected.
+        skip_for_these_callers = (
+            'wagtail.wagtaildocs.views.documents.edit',
+            'wagtail.wagtailimages.views.images.edit',
+        )
+
+        if called_from not in skip_for_these_callers:
+            return
+
+        # We know this delete is being called from by specific Wagtail views of
+        # interest. Grab the variables from the stack that contain the filename
+        # of the document/image being replaced and the filename of the new
+        # document/image with which it is being replaced.
+        original_file = caller_frame_object.f_locals['original_file']
+        doc = caller_frame_object.f_locals['doc']
+
+        # If these two filenames are the same, we don't want to go ahead with
+        # the delete, because that will remove the new file that's just been
+        # provided to replace the old file.
+        #
+        # If these two filenames are different, we should go ahead with the
+        # deletion, because the intended Wagtail behavior is to clean up
+        # the old file that is being replaced.
+        return original_file.name == doc.file.name
+
+
+class OverwritingS3Storage(WagtailOverwritingStorageMixin, S3BotoStorage):
+    """Boto-based S3 Django storage that supports overwriting from Wagtail.
+
+    Based on the django-storages boto-based storage backend defined here:
+
+    https://github.com/jschneier/django-storages
+
+    This storage has a default setting AWS_S3_FILE_OVERWRITE=True that
+    overwrites existing files when files with the same name are uploaded.
+
+    This class adds v1.storage.WagtailOverwritingStorageMixin to support
+    that use case when using this storage in Wagtail 1.13+.
+    """
+    pass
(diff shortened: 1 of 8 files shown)


Scotchester:
> this code in cfgov-refresh would only be necessary until that is merged in

… until that is merged in and cfgov-refresh is updated to that version of Wagtail 2.1.x, so it could be a while :)

(just to further justify this hackery; we can't wait until both of those conditions are met to fix this)

cfgov/v1/storage.py
```diff
@@ -0,0 +1,120 @@
...
+    This logic is sound as long as it is safe to assume that new uploads with
+    identical filenames never overwrite each other and always get assigned a
+    unique filename when stored. However, as described above, this may not
+    always be the case.
+
+    If, instead, files get overwritten, then the final delete that happens
+    will in fact delete the new file that's just been stored.
+
+    This mixin attempts to prevent this from happening. It uses the Python
+    stack inspect capability to identify if a delete call is being invoked
```
willbarton: 😱

willbarton:
> … until that is merged in and cfgov-refresh is updated to that version of Wagtail 2.1.x, so it could be a while :)

@Scotchester Wagtail 1.13 is a LTS release. If this is a problematic regression, then we could try to make an argument for fixing it.

Scotchester:
@willbarton I understand your reticence.

> I'm left wondering why we shouldn't just set `AWS_S3_FILE_OVERWRITE=False`?

Because of the comment I made on https://github.com/wagtail/wagtail/pull/4567#issuecomment-392846530 – we have a longstanding practice of linking directly to docs on S3 with a files.consumerfinance.gov URL. If we didn't allow overwriting, we'd be unable to replace updated files in place and maintain those links.

I did have the thought that we could conceivably set `AWS_S3_FILE_OVERWRITE=False` and set it up so that it would automatically create an S3 redirect from the original file to the new file, but as we've recently found, those redirects are somewhat brittle and not as much in our control as we'd like.

cfgov/v1/storage.py
chosak: Oh, I know. I know.

cfgov/v1/storage.py
```diff
@@ -0,0 +1,120 @@
...
+
+        called_from = caller_module_name + '.' + caller_frame_function_name
+
+        # Only skip delete if it is being called by the wagtaildocs.edit or
+        # wagtailimages.edit views. We want deletes called from other places
+        # to work as expected.
+        skip_for_these_callers = (
+            'wagtail.wagtaildocs.views.documents.edit',
+            'wagtail.wagtailimages.views.images.edit',
+        )
```
willbarton: Would it be possible to call out where those other places are, explicitly? If we're going to do stack manipulation to force behavior we want, I think I'd prefer to explicitly enable it for callers, rather than vice-versa.

cfgov/v1/storage.py
```diff
@@ -0,0 +1,120 @@
...
+
+        if called_from not in skip_for_these_callers:
+            return
+
+        # We know this delete is being called from by specific Wagtail views of
+        # interest. Grab the variables from the stack that contain the filename
+        # of the document/image being replaced and the filename of the new
+        # document/image with which it is being replaced.
+        original_file = caller_frame_object.f_locals['original_file']
+        doc = caller_frame_object.f_locals['doc']
```
willbarton: It says above this works with images and docs, but it's looking for a `doc` local, which [`wagtail.wagtaildocs.views.documents.delete` has](https://github.com/wagtail/wagtail/blob/stable/1.13.x/wagtail/wagtaildocs/views/documents.py#L182) but [`wagtail.wagtailimages.views.images.delete` does not](https://github.com/wagtail/wagtail/blob/stable/1.13.x/wagtail/wagtailimages/views/images.py#L225) (and images aren't addressed in any of the tests). Is that correct? Does this only work with the document views, or should it work with both? The Wagtail behavior is the same for documents and images.

chosak:
@willbarton the 500 errors in GHE#2750 happen because the Document whose edit view is being viewed no longer has its backing file on S3, as it was deleted when it was replaced. Wagtail should properly show an error message in those cases -- and it does for local storage -- but it doesn't properly handle missing files on remote storage backends. [wagtail#4468](https://github.com/wagtail/wagtail/pull/4468) fixed this for images, but there needs to be an additional PR for documents. So the 500 error is caused by the file being missing **and** Wagtail not handling that missing file properly. This PR addresses the first part - preventing the files from being deleted on overwrite.

@gasman has left [a long and thoughtful comment](https://github.com/wagtail/wagtail/pull/4567#issuecomment-392934164) on [wagtail#4567](https://github.com/wagtail/wagtail/pull/4567). I'd like to understand more about long-term options such as [wagtail#1420](https://github.com/wagtail/wagtail/issues/1420), and agree that our use of `AWS_S3_FILE_OVERWRITE=True` violates some Django/Wagtail conventions -- but in the short term, I'm not sure there's a better way to do what we're currently doing without much larger changes to code or processes.

As a reminder, the current workflow appears to be:

1. Wagtail users can upload files as Documents, and get both a useable Wagtail document URL `/documents/1234/foo.pdf` **and** a dependable URL on the backing S3 storage (`bucket/f/foo.pdf`).
2. Wagtail users can use the Document edit view to upload a new version of an existing document, keeping the same filename, **without** breaking either of those URLs.

I'd love to hear any better suggestions for how to do this in Wagtail 1.13 without something like this PR. Believe me, `inspect` was a last resort! 😄

cfgov/v1/storage.py
chosak: I'm not sure I follow -- maybe this variable is just named poorly, but this is the list of places where we want to skip delete -- otherwise the default behavior works. We need to maintain the default functionality of `Storage.delete` in any other cases where it is used.

cfgov/v1/storage.py
chosak: Yes, nice catch! It should work with both. If we come to some consensus about moving forward with this approach I can fix this and add additional tests.

willbarton:
> Wagtail users can use the Document edit view to upload a new version of an existing document, keeping the same filename, without breaking either of those URLs.

@Scotchester @chosak thanks! I think what I was missing (also coffee).

willbarton:
> our use of `AWS_S3_FILE_OVERWRITE=True` violates some Django/Wagtail conventions

Well, now that I know about it, `S3BotoStorage`'s default of `AWS_S3_FILE_OVERWRITE=True` annoys me quite a bit. We didn't consciously chose to set that value. I imagine we've just assumed that was normal behavior. That's on django-storages.

cfgov/v1/storage.py
willbarton: Then does this work?

If we want to skip delete for `wagtail.wagtaildocs.views.documents.edit`, then shouldn't [line 88](https://github.com/cfpb/cfgov-refresh/pull/4195/files#diff-255059ee1b4d67208bda88e26493190bR88) return `True` so that [line 60](https://github.com/cfpb/cfgov-refresh/pull/4195/files#diff-255059ee1b4d67208bda88e26493190bR60) triggers the return before the super call?

Scotchester:
> Well, now that I know about it, S3BotoStorage's default of `AWS_S3_FILE_OVERWRITE=True` annoys me quite a bit. We didn't consciously chose to set that value. I imagine we've just assumed that was normal behavior. That's on django-storages.

Still, it is the behavior we _want_ in order to reupload docs without breaking the S3 URL.

cfgov/v1/storage.py
chosak: More evidence that I named these things poorly.

```py
if called_from not in skip_for_these_callers
```

This means

> if `delete` was called from a caller that isn't in the list of callers to skip `delete` for

meaning, we know we shouldn't skip delete, so `_should_skip_wagtail_delete` should return falsey.

chosak:
> reupload docs without breaking the S3 URL

This is the crux of the problem -- this behavior was sort of a loophole in Wagtail that was closed in 1.13. So we can try to reopen the loophole via hack here and/or PR to Wagtail, or reconsider our processes to be more consistent with standard behavior.

> That's on django-storages

[A warning about this](https://github.com/wagtail/wagtail/commit/fc46f12939f0076804433ed2aafdedcaacca5f38#diff-a47540faa28f483eecfcd2209fd987ba) was recently added to the Wagtail documentation.

willbarton:
Alright, aside from the "skip" terminology being slightly confusing when insufficiently caffeinated, I think this is *alright*.

But brainstorming alternative approaches (which I imagine you've already iterated through otherwise you wouldn't be stack inspecting), I'm wondering if it'd be possible to simply override the `edit` views for documents and images. I'm imagining a wrapper that just handles `POSTs`, something like:

```python
from wagtail.wagtailimages.views import wagtailimages_edit

def edit(request, image_id):
    if request.method == 'POST':
        original_file = doc.file
        form = DocumentForm(request.POST, request.FILES, instance=doc, user=request.user)
        if form.is_valid():
            doc = form.save()

            # Here is where the delete logic lived

            # Reindex the document to make sure all tags are indexed
            search_index.insert_or_update_object(doc)

            messages.success(request, _("Document '{0}' updated").format(doc.title), buttons=[
                messages.button(reverse('wagtaildocs:edit', args=(doc.id,)), _('Edit'))
            ])
            return redirect('wagtaildocs:index')
        else:
            messages.error(request, _("The document could not be saved due to errors."))

    else:
        return wagtailimages_edit(request, image_id)
```

I realize that this would be vastly more duplicative of Wagtail, and we'd have to keep up with changes that might happen in that code, but it'd be more readable and more obvious what was going on.

chosak:
Closing this PR based in part on feedback above and also due to some offline discussions with @Scotchester about another approach (custom Document model) that has more promise.

PR #4196, opened 2018-05-29
Remove cli flag

anselmbradford:
After chatting with @chosak, it was decided the `cli-flag.sh` script adds too much overhead for little gain. This PR merges this script into the frontend and backend scripts, respectively, and strips out the irrelevant parts so that bewds and fewds can adjust their scripts without the danger of impacting the other's workflow.

This also adds a `NODE_ENV` to the `.env` sample file, which can be set so that `NODE_ENV` is set to either `development` or `production` without running `./frontend.sh`.

## Removals

- Removes `cli-flag.sh` script.

## Changes

- Fixes issue where `setup.sh` was not passing production cli flag.
- Adds documentation about `NODE_ENV` usage.

## Testing

new testing instructions:
1. `./setup.sh` should run frontend and backend scripts in development mode.
1. `./setup.sh production` should run frontend in production mode and backend scripts in development mode.
1. `./frontend.sh` should run frontend script in development mode.
1. `./frontend.sh production` should run frontend script in production mode.
1. `./backend.sh` should run backend script in development mode.
1. `./backend.sh production` should run backend script in development mode.

old testing instructions:
1. `./setup.sh` should run frontend and backend scripts in development mode.
1. `./setup.sh production` should run frontend and backend scripts in production mode.
1. `./frontend.sh` should run frontend script in development mode.
1. `./frontend.sh production` should run frontend script in production mode.
1. `./backend.sh` should run backend script in development mode.
1. `./backend.sh production` should run backend script in production mode.

## Notes

- There exists the not-great workflow of having `NODE_ENV=development` in your `.env` file and running `./frontend.sh production` and then running `gulp test` and having it fail because you've gone from the development to production to development environments and have ended up with the production set of dependencies but the gulp set of tasks. Only way around that currently is to manually set your `NODE_ENV` the same as the frontend.sh script that was run (`./frontend.sh` or `./frontend.sh production`).

- You'll see this warning, which is existing behavior:
```
WARNING in configuration
The 'mode' option has not been set, webpack will fallback to 'production' for this value. Set 'mode' option to 'development' or 'production' to enable defaults for each environment.
```


willbarton:
@chosak yes. AFAIK, `backend.sh` is only ever used locally.

docs/usage.md
```diff
@@ -31,13 +31,21 @@ This setup script will remove and reinstall the project dependencies
...
+  You may also run `./backend.sh` or `./frontend.sh`
+  if you only want to re-build the backend or front-end, respectively.
+
+##### Setting environments
+
+The `NODE_ENV` environment variable can be set in your `.env` file to either
+`development` or `production`, which will affect how the build is made and what
+gulp tasks are available. To install dependencies of one environment
+or the other run `./frontend.sh` (dependencies and devDependencies)
+or `./frontend.sh production` (non-devDependencies).
```
Scotchester: The relationship between `NODE_ENV` and the parameter passed to `frontend.sh` could be made a little more clear, here.

Scotchester: Also, I think it might be more clear to say "dependencies only" instead of "non-devDependencies".

anselmbradford: Is that clear enough that it's talking about excluding https://github.com/cfpb/cfgov-refresh/blob/master/package.json#L67

Scotchester: I guess to be totally clear, we should say "dependencies but not devDependencies".

anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/pull/4196/commits/910da66a1b4234b1aca91fb2859915766ffa3672

anselmbradford:
@willbarton @chosak @Scotchester okay, pushed up a change that stripped out the production environment setting.

anselmbradford:
I may be misunderstanding this, but it's not strange that `./setup.sh production` is running the frontend in production mode, but the backend is in development (`/requirements/local.txt`)?

PR #4197, opened 2018-05-30
 Hide empty regdown paragraphs

contolini:
Inline interps have an erroneous margin at the top caused by the presence of an empty paragraph element.

caff6f7a-5dc2-11e8-93d4-bf9ec59bbf7b

The empty paragraph comes from opening interp labels in the regdown.

7aeceb3c-5dc4-11e8-92d4-92b894c3793c

I was originally thinking we should remove empty paragraphs but in retrospect we want to keep the IDs so that users can jump to those sections on the page. By replacing empty paragraphs with divs they naturally disappear from view but remain linkable.

Unrelated: I tweaked the inline interp pattern to support longer label prefixes (`12-b-1-Interp-2` instead of just `12-b-Interp-2`).

## Changes

- Empty regdown paragraphs are now divs instead of paragraphs.
- Inline interps can have longer prefixes.

## Checklist

### DIFF
--- a/cfgov/regulations3k/regdown.py
+++ b/cfgov/regulations3k/regdown.py
@@ -178,8 +178,12 @@ class LabeledParagraphProcessor(ParagraphProcessor):
             # the paragraph tag, label id, and initial text, then process the
             # rest of the blocks normally.
             label, text = match.group('label'), match.group('text')
-            p = util.etree.SubElement(parent, 'p')
-            p.set('id', label)
+            # Labeled paragraphs without text should use a div element
+            if text == '':
+                el = util.etree.SubElement(parent, 'div')
+            else:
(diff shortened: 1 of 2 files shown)


cfgov/regulations3k/regdown.py
```diff
@@ -178,26 +178,30 @@ def run(self, parent, blocks):
             # the paragraph tag, label id, and initial text, then process the
             # rest of the blocks normally.
             label, text = match.group('label'), match.group('text')
-            p = util.etree.SubElement(parent, 'p')
-            p.set('id', label)
+            # Labeled paragraphs without text should use a div element
```
willbarton: Out of curiosity, why should they use a `div`?

PR #4201, opened 2018-05-30
Mega menu: Add aria attributes to menu footer links

anselmbradford:
## Changes

- Adds aria- attributes to menu footer links.

## Testing

1. Pull down branch.
2. Visit the site locally.
3. `Cmd`+`F5` to start VoiceOver
4. Tab through to Consumer Tools (scrolling down slightly and clicking "Consumer Tools" and then tabbing from there may be the easiest)
5. `Enter` to open Consumer Tools
6. Tab to `Ask CFPB` and listen that it reads a complete sentence.
7. Tab to `Submit a complaint` and listen that it also reads a complete sentence.

## Note

- With the move to Mega Menu in Wagtail, the `_vars-mega-menu.html` template is deprecated, so we'll need a solution that works in Wagtail, but the changes here can be used as a reference for what's needed in Wagtail.

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/_vars-mega-menu.html
+++ b/cfgov/jinja2/v1/_includes/organisms/_vars-mega-menu.html
@@ -43,7 +43,7 @@
                 'replaces four disclosure forms with two new ones. ' +
                 'We have resources to help you comply.',
     }
-    
+
 } %}
 
 {% set consumer_tools_content = {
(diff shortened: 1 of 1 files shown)


cfgov/jinja2/v1/_includes/organisms/_vars-mega-menu.html
```diff
@@ -200,9 +200,25 @@
...
         'value': {
-            'content': 
-                '<p>Browse answers to hundreds of financial questions. <a href="/ask-cfpb/">Ask CFPB</a></p>
-                <p>Have an issue with a financial product? <a href="/complaint/">Submit a complaint</a></p>',
+            'content':
+                '<p>
+                    <span aria-hidden="true">
+                        Browse answers to hundreds of financial questions.
+                    </span>
+                    <a aria-label="Ask CFPB: browse answers to hundreds of financial questions."
```
marteki: Honest wondering: is there a reason why the phrases are switched so "Ask CFPB" comes first before the contextual sentence?

marteki: If the answer is not really, can we switch them around to match the visible text? That way, VoiceOver will lead with the phrase that's recognizable to people rather than the jargony product name.

anselmbradford: We can, but it sounds weird when read the other way around because it's not clear the jargony product name is not part of a longer missing sentence. The other link leads with a question `Have an issue with a financial product?`, while this one just reads `Browse answers to hundreds of financial questions.`. If I was speaking this I would say something like `Ask CFPB to browse answers to hundreds of financial questions.`

anselmbradford: Or `Browse answers to hundreds of financial questions with Ask CFPB.`

marteki: I like that last option!

anselmbradford: Fixed to be `Browse answers to hundreds of financial questions with Ask CFPB.`

PR #4204, opened 2018-05-31
regs3k: Make subpart section_range available in admin

higs4281:
Subparts in the admin tree currently display only subpart title, which 
isn't helpful if you want to edit a particular section of a large reg.

This PR elevates the Subpart's `section_range` value from a property to a
model field that gets populated when a subpart is saved. As a model field, 
we can display it in the admin:

section_ranges

## Testing
This PR includes a table migration, so you'll need to update your local db with `migrate`

```
./cfgov/manage.py migrate --noinput
```

Then populate the new field by saving all subparts:

```
./cfgov/manage.py shell
from regulations3k.models import Subpart
for each in Subpart.objects.all():
    each.save()
```

The subpart ranges should show up in your regulations admin.

### DIFF
--- a/cfgov/regulations3k/migrations/0005_subpart_section_range.py
+++ b/cfgov/regulations3k/migrations/0005_subpart_section_range.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('regulations3k', '0004_add_acquired_and_draft_to_effectiveversion'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='subpart',
+            name='section_range',
+            field=models.TextField(null=True, blank=True),
+        ),
+    ]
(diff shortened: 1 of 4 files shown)


willbarton:
> Then populate the new field by saving all subparts:
> 
> ```
> ./cfgov/manage.py shell
> from regulations3k.models import Subpart
> for each in Subpart.objects.all():
>    each.save()
> ```
> 
> The subpart ranges should show up in your regulations admin.

Do you want to create a data migration? 😇

willbarton:
I'm not sure if this is going to work as intended either... what if a section is added to the subpart? In order to update the text in the admin, the subpart will have to also be re-saved. We could do that in `Section.save()`, but that seems a bit smelly.

higs4281:
> what if a section is added to the subpart?

A good point. I was thinking section-adding would routinely be handled by the importer, which takes care of saving the subpart. But there is a case for a locally added section, in which case something, man or beast, would would have to save the subpart. Is there a better way to make range available in the admin?

willbarton:
Unfortunately I don't think so. It operates entirely on database fields. 

*Maaaaaaybe* we could do something creative with the queryset to construct the value from related fields in the SubpartModelAdmin?

willbarton:
I'm thinking something along the lines of adding annotations to the queryset:

```
from django.db.models import Max, Min
from regulations3k.models import Subpart
Subpart.objects.annotate(min_section=Min('sections__label'), max_section=Max('sections__label'))
```

(Which, without a tremendous amount of testing seems to mostly work?)

willbarton:
> Unfortunately I don't think so. It operates entirely on database fields.

I am entirely wrong about this. We're already using properties in a ModelAdmin here: https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jobmanager/wagtail_hooks.py#L58

PR #4208, opened 2018-05-31
Mega menu FMC column fix

virginiacc:
Fix `three-col` mega menu classes so they work correctly in header.css

## Changes

- Updates to `three-col` classes in mega menu template and less file

## Testing

1. Check out branch
2. Run `gulp styles`
3. Verify that FMC column displays correctly on   
    - http://localhost:8000 
    - http://localhost:8000/data-research/consumer-complaints/search/ 
(mega menu text will still be bold on search page)

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
+++ b/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
@@ -24,7 +24,7 @@
 {%- macro _classes( nav_depth, suffix='', additional_suffix='' ) -%}
 {%- set general_class = base_class ~ '_content' -%}
 {%- set depth_class = general_class ~ '-' ~ nav_depth -%}
-{%- set additional_class = depth_class ~ '__' ~  additional_suffix if additional_suffix else '' -%}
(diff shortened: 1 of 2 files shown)


cfgov/jinja2/v1/_includes/organisms/mega-menu.html
```diff
@@ -24,7 +24,7 @@
 {%- macro _classes( nav_depth, suffix='', additional_suffix='' ) -%}
 {%- set general_class = base_class ~ '_content' -%}
 {%- set depth_class = general_class ~ '-' ~ nav_depth -%}
-{%- set additional_class = depth_class ~ '__' ~  additional_suffix if additional_suffix else '' -%}
+{%- set additional_class = depth_class ~ suffix ~ '__' ~  additional_suffix if additional_suffix else '' -%}
```
jimmynotjim: If there are no additional suffixes, will the class end with `__`?

virginiacc: I think additional_class will just be set to '' if there's no additional_suffix value.

jimmynotjim: Ahhh, thanks, I was thinking the condition was only for the last item in the combined string.

PR #4217, opened 2018-06-01
Mega menu: Set the aria attributes on menu trigger and menu items

anselmbradford:
Fixes [GHE]/CFGOV/platform/issues/2791

## Changes

- Replaces aria-pressed roles with aria-haspopup="menu" and adds aria-role="menuitem" to menu list items.

## Testing

1. Open page in mobile size and run the WAVE chrome extension to see the aria-attributing.

## Screenshot

screen shot 2018-06-01 at 4 58 53 pm

## Notes:

- [This example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/menu-button/menu-button-links.html) notes that 

> The menuitem role is on the HTML a element contained by each li element so link behaviors are available when focus is set on the menu item. Another reason the menuitem role is not on the li element is that the semantics of descendants of ARIA menuitem elements are not exposed in the accessibility tree. That is, an item in a menu can only be a menuitem because accessibility APIs do not enable assistive technologies to render elements contained inside of an item in a menu.

However, [this site](https://www.levelaccess.com/challenges-mega-menus-standard-menus-make-accessible/) notes that

> Nested submenus can be added using the same syntax by placing the same structure within the appropriate list item. However, when adding a nested submenu, the triggering element that includes aria-haspopup must include role=”menuitem”, and not a button role. Focus management is critical and must be strictly observed.

Our nested menus are inside the `li`, not inside the `a`, so I put the `menuitem` on the `li`. However, this likely should be revisited when addressing navigating the menu with VoiceOver.

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
+++ b/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
@@ -51,7 +51,7 @@
     {% for nav_group in nav_groups %}
     {% set group = nav_group.value %}
     <div class="{{ _classes( nav_depth, '-list' ) }}">
-        
+
         {% if group.group_title %}
         <div class="h5
                   {{ base_class ~ '_group-heading' }}
@@ -74,7 +74,7 @@
             {%- endfor %}
         </ul>
     </div>
-    
+
     {% endfor %}
 </div>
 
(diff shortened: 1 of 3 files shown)


cfgov/jinja2/v1/_includes/organisms/mega-menu.html
```diff
@@ -179,15 +180,16 @@
     {% set has_children = nav_item.nav_groups | count > 0 or nav_item.nav_items | count > 0 %}
 
     <li class="m-list_item {{ _classes( nav_depth, '-item' ) }}"
-        {{ 'data-js-hook=behavior_flyout-menu' if has_children else '' }}>
+        {{ 'data-js-hook=behavior_flyout-menu' if has_children else '' }}
+        {{ 'aria-haspopup=menu' if has_children else '' }}
```
jimmynotjim: Any reason not to combine L:183 & L:184 since they're reliant on the same condition?

anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/pull/4217/commits/1e8272a836025fd80dba8bafafd735a2b573b6e9

PR #4225, opened 2018-06-05
Move aria-haspopup from list items to links

anselmbradford:
## Changes

- Moves `aria-popup` from `li` to `a` links in the submenus.
- Adds `role="menuitem"` to global CTA complaint link at mobile size.
- Adjust whitespace on attributes by adding `-`.

## Testing

1. Using desktop voiceover on the menu at mobile size you should hear "1 of 6" etc for the menu items.


jimmynotjim:
Tested locally on both Desktop at multiple sizes and iPhone and works as advertised. Still need to address the tab/swipe issue you discussed in chat, but I'm starting to think they're the same issue.

PR #4229, opened 2018-06-06
Invalidate pages if their social sharing image is too large

richaagarwal:
### Overview
We want to prevent social sharing images that are too large from getting saved to a page (currently they get saved and then don't display at all on Twitter).
See GHE/CFGOV/platform/issues/2377 for more context. 

### Notes / potential issues
 - Saving a non-Ask page displays the specific error, like so:
screen shot 2018-06-11 at 10 07 06 am

- While saving an Ask answer (a Django model) shows something more generic:
screen shot 2018-06-05 at 11 14 30 pm

I think this is related to this issue https://github.com/wagtail/wagtail/issues/3441  which was only recently fixed, so it would not be fixed in our current version of Wagtail.  As part of this work, I also updated the help text (hence the migrations), so I feel like it's not too big of a deal, but just wanted to flag.

### DIFF
--- a/cfgov/ask_cfpb/models/django.py
+++ b/cfgov/ask_cfpb/models/django.py
@@ -23,6 +23,7 @@ from wagtail.wagtailcore.models import Page
 from wagtail.wagtailimages.edit_handlers import ImageChooserPanel
 
 from v1.util.migrations import get_or_create_page
+from v1.util.util import validate_social_sharing_image
 
 
 html_parser = HTMLParser.HTMLParser()
@@ -376,7 +377,9 @@ class Answer(models.Model):
         help_text=(
             'Optionally select a custom image to appear when users share this '
             'page on social media websites. If no image is selected, this '
-            'page\'s category image will be used. Minimum size: 1200w x 630h.'
+            'page\'s category image will be used. '
+            'Recommended size: 1200w x 630h. '
+            'Maximum size: 4096w x 4096h.'
         )
     )
 
@@ -458,6 +461,10 @@ class Answer(models.Model):
             html_parser.unescape(self.snippet_es),
             html_parser.unescape(self.answer_es)))
         return html.strip_tags(unescaped).strip()
+
+    def clean(self):
+        super(Answer, self).clean()
+        validate_social_sharing_image(self.social_sharing_image)
 
     def cleaned_questions(self):
         cleaned_terms = html_parser.unescape(self.question)
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -31,6 +31,7 @@ from v1 import blocks as v1_blocks, get_protected_url
 from v1.atomic_elements import molecules, organisms
 from v1.models.snippets import ReusableText
 from v1.util import ref
+from v1.util.util import validate_social_sharing_image
 
 
 class CFGOVAuthoredPages(TaggedItemBase):
@@ -78,7 +79,8 @@ class CFGOVPage(Page):
         related_name='+',
         help_text=(
             'Optionally select a custom image to appear when users share this '
-            'page on social media websites. Minimum size: 1200w x 630h.'
+            'page on social media websites. Recommended size: 1200w x 630h. '
+            'Maximum size: 4096w x 4096h.'
         )
     )
 
@@ -125,6 +127,10 @@ class CFGOVPage(Page):
         ObjectList(sidefoot_panels, heading='Sidebar/Footer'),
         ObjectList(settings_panels, heading='Configuration'),
     ])
+
+    def clean(self):
+        super(CFGOVPage, self).clean()
+        validate_social_sharing_image(self.social_sharing_image)
 
     def get_authors(self):
         """ Returns a sorted list of authors. Default is alphabetical """
--- a/cfgov/v1/tests/test_social_sharing_image_validation.py
+++ b/cfgov/v1/tests/test_social_sharing_image_validation.py
@@ -1,0 +1,41 @@
+from django.core.exceptions import ValidationError
+from django.test import TestCase
+
+from model_mommy import mommy
+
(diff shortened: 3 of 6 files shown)


cfgov/v1/models/base.py
```diff
@@ -126,6 +128,9 @@ class CFGOVPage(Page):
         ObjectList(settings_panels, heading='Configuration'),
     ])
 
+    def clean(self):
```
chosak: Does this method need to call the `super` clean, to validate against any base Wagtail validators (like [this code](https://github.com/wagtail/wagtail/blob/v1.13.1/wagtail/wagtailcore/models.py#L456), that validates that the slug is unique)? This doesn't appear to cause problems via the UI (because it has additional checks via the Django form) but skipping that might allow you to programmatically create invalid pages.

cfgov/ask_cfpb/models/django.py
```diff
@@ -459,6 +461,9 @@ def answer_text_es(self):
             html_parser.unescape(self.answer_es)))
         return html.strip_tags(unescaped).strip()
 
+    def clean(self):
```
chosak: Does this method need to call the `super` clean to validate against the base model validators (like the rest of the fields)?

Like my page comment above, this doesn't appear to cause problems in practice via the UI - because none of the fields are required - but if they were, I think this could allow for saving of invalid models.

willbarton:
Forgot to use the review comment box (and you can't edit reviews without comments?) Looks straight-forward enough, but I do agree with @chosak that some unit tests would be appreciated!

richaagarwal:
Yes yes unit tests are coming @willbarton @chosak . Thank you for the reviews!!

richaagarwal:
I added some unit tests, as well as added validation for the minimum size as well.  If that seems reasonable (cc @ielerol @Scotchester ) I will go ahead and merge.

cfgov/v1/util/util.py
```diff
@@ -201,3 +202,16 @@ def extended_strftime(dt, format):
...
+
+
+def validate_social_sharing_image(image):
+    """Raises a validation error if the image is too large or too small."""
+    if image:
+        if image.width > 4096 or image.height > 4096:
+            raise ValidationError(
+                'Social sharing image must be less than 4096w x 4096h'
+            )
+        if image.width <= 1200 or image.height <= 630:
```
Scotchester: These should be `<`, rather than `<=`. Exactly 1200x630 is the ideal size.

cfgov/v1/util/util.py
```diff
@@ -201,3 +202,16 @@ def extended_strftime(dt, format):
...
+            raise ValidationError(
+                'Social sharing image must be greater than 1200w x 630h'
```
Scotchester: This should say "at least 1200w x 630h".

cfgov/ask_cfpb/models/django.py
```diff
@@ -376,7 +377,8 @@ class Answer(models.Model):
         help_text=(
             'Optionally select a custom image to appear when users share this '
             'page on social media websites. If no image is selected, this '
-            'page\'s category image will be used. Minimum size: 1200w x 630h.'
+            'page\'s category image will be used. Minimum size: 1200w x 630h. '
+            'Maximum size: 4096w x 4096h.'
```
Scotchester: I'd like the help text to state that 1200x630 is the ideal size, too. Perhaps:

"Minimum (and ideal) size: 1200w x 630h."

cfgov/v1/tests/test_social_sharing_image_validation.py
```diff
@@ -0,0 +1,57 @@
...
+            ValidationError,
+            'Social sharing image must be greater than 1200w x 630h'
+        ):
+            self.page.save()
+
+    def test_validation_error_thrown_if_height_too_small(self):
+        self.page.social_sharing_image = mommy.prepare(CFGOVImage, width=1500, height=629)
+        with self.assertRaisesRegexp(
+            ValidationError,
+            'Social sharing image must be greater than 1200w x 630h'
```
Scotchester: This (and the above) message will also need to change here in the test according to the change made to the actual message.

cfgov/v1/models/base.py
```diff
@@ -78,7 +79,8 @@ class CFGOVPage(Page):
         related_name='+',
         help_text=(
             'Optionally select a custom image to appear when users share this '
-            'page on social media websites. Minimum size: 1200w x 630h.'
+            'page on social media websites. Minimum size: 1200w x 630h. '
+            'Maximum size: 4096w x 4096h.'
```
Scotchester: Same help text comment applies here.

cfgov/ask_cfpb/models/django.py
ielerol: I endorse this suggestion, though I have a minor quibble with calling it "ideal," something about it feels off to me.  Maybe "Minimum (and recommended) size: 1200w x 620h"

I can't quite articulate what bugs me about "ideal" though, so, could be a personal problem.

richaagarwal: This reminds me that there was some conflation around minimum size and target size in the related issue.  Do we want to consider this a minimum at all then?  I was just trying to apply validation that's consistent with the help text, but if it's confusing I can remove the validation on minimum size, remove the help text on minimum size, and just use language around suggesting ideal/recommended size.

sarahsimpson09:
@richaagarwal My suggestion for the min/max help text: 

Recommended: 1200w X 630h; Maximum: 4096 x 4096

richaagarwal:
Thanks @sarahsimpson09! After some discussion, I've settled on removing the minimum size validation as well and will use the text Sarah suggested. (Recommended size: 1200w x 630h; Maximum size: 4096w x 4096h)

PR #4238, opened 2018-06-08
regs3k: Update ecfr_importer to handle appendices

higs4281:
Appendices and interpretations share the same DIVs in our source XML.
This change teases them out and processes them separately.

In the pursuit of small PRs, this is an incremental change, focusing on parsing just the main text of appendices with subheadings, and recognizing when they're using a different indentation scheme.

This code doesn't work for all regs yet, nor does it properly handle our oddities: pseudo-forms, tables and images.

Some other improvements included:
- A fix to the Section model's "numeric_label" property so that it doesn't choke on appendices, which have alpha section labels.
- A fix to the ordering of side navigation subparts and sections.
- The addition of "label" field to list display for sections in the admin, to help identify appendices.

## Testing
Appendix imports aren't working for all regs, but you can test B with:

```
./cfgov/manage.py runscript ecfr_importer --script-args 1002
```

There's a slight drop in test coverage that will be addressed in the next iteration.

### DIFF
--- a/cfgov/regulations3k/models/django.py
+++ b/cfgov/regulations3k/models/django.py
@@ -174,6 +174,8 @@ class Section(models.Model):
         part, number = self.label.split('-')[:2]
         if number.isdigit():
             return '\xa7\xa0{}.{}'.format(part, number)
+        else:
+            return ''
 
     @property
     def section_number(self):
@@ -182,4 +184,7 @@ class Section(models.Model):
 
     @property
     def title_content(self):
-        return self.title.replace(self.numeric_label, '')
+        if self.numeric_label:
+            return self.title.replace(self.numeric_label, '')
+        else:
+            return self.title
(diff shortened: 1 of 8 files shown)


cfgov/regulations3k/models/django.py
```diff
@@ -174,6 +174,8 @@ def numeric_label(self):
         part, number = self.label.split('-')[:2]
         if number.isdigit():
             return '\xa7\xa0{}.{}'.format(part, number)
+        else:
+            return ''
```
willbarton: How would you feel about returning a [`sortable_label()`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/regulations3k/models/django.py#L13) instead? It converts alphabetical labels into their positional numeric equivalent.

higs4281:
> refactor

absolutor.

cfgov/regulations3k/models/django.py
higs4281: √

PR #4240, opened 2018-06-11
Update favicon location and static reference

acrewdson:
As part of the Wordpress removal effort, we need to switch to using a favicon that's part of the Django app. Up till now the favicon has come from an internal Wordpress theme repo which gets added to the build artifact via no-drama's `--aux` option.

We also noticed that the existing `<link/>` reference to a favicon in the site's base template to a static file ('assets/favicon.ico') pointed to a favicon that, as far as we can tell, did not get placed on the filesystem as part of the existing Ansible deploy process. That is, the file did not seem to get placed there via `collectstatic` or some other build/deploy operation. 

To address the confusion we have moved an existing copy of the file (which is binary equal to the current favicon.ico being served on the site) so that it will be collected to the top of the `STATIC_ROOT`.

Also:
* I did not remove [this other favicon.ico](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/legacy/static/nemo/_/img/favicon.ico)
* I'm assuming we do not care about [this reference to a favicon.png](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/legacy/templates/front/base_print.html#L24)
* This PR has a corresponding PR in the GHE Ansible repo


chosak:
:+1: This looks good to me. I don't really know anything about that other favicon.ico, or about the PNG. I can confirm when running locally that http://localhost:8000/static/favicon.ico works, as we'd hoped, and that pages point to it in their base template.

Can @mistergone @Scotchester give this a FEWD's eye?

mistergone:
I can't speak to the mystery file, but these changes look good to me!

👍

Scotchester:
OK, after some helpful input from @chosak, here's my thinking:

The only other thing in the path proposed in this PR (`cfgov/v1/static/`) is the question_tips CSS file, which is for the Wagtail Admin, and I think that should get relocated to where the other Wagtail Admin CSS lives (`cfgov/templates/wagtailadmin/css/`). That would leave the favicon as the only thing in `cfgov/v1/static/`, which seems odd.

Additionally, the favicon belongs to the whole of consumerfinance.gov, and thus feels more at home outside of the `v1` app. Andy suggested `cfgov/core/static/`, which I think makes good sense. It will still get picked up automatically by Django's `collectstatic` there, and it's a sort of kindred spirit with the `robots.txt` file that already lives there.

Scotchester:
P.S.: Don't feel the need to deal with that admin CSS file in this PR. I can open a PR for that.

acrewdson:
@Scotchester Please confirm this looks right now – thanks!

PR #4241, opened 2018-06-11
Remove MySQL-specific local dev pieces

acrewdson:
Removes MySQL-related local dev env vars, scripts, and Docker config. No rush to merge this, and we may want to wait a bit longer.


acrewdson:
@chosak Right, this PR is only intended to affect local dev. 👍

higs4281:
Just a thought, but might it be worth leaving the configs as comments for the sake of devs who want to switch to mysql to inspect an old mysql dump?

acrewdson:
@higs4281 That's a good question, and maybe one @chosak might also have brought up at one point, now that I think about it. I guess I'm wondering: how far into the future do we think someone might still want to be looking at a MySQL dump? If it would not be much beyond the Postgres production cut-over (a few weeks from now), we could just hold off on merging this PR for a couple weeks. Or if it's indeterminate, and we don't really know, then yes, I'd be glad to add those commented-out lines back in this PR.

chosak:
I don't have strong feelings, but err on the side of simplicity. Part of the motivation behind adding [this pointer](https://github.com/cfpb/cfgov-refresh/blob/88bffb0d5796ddd5f0e7795eb07d6c80ccec47fb/.env_SAMPLE#L48) to the [dj-database-url](https://github.com/kennethreitz/dj-database-url) repo was to give a hint to devs who want to try to connect to other databases, including MySQL.

We could add a simple example MySQL config string under the other two, like this:

```
export DATABASE_URL=sqlite:///db.sqlite3
#export DATABASE_URL=postgres://cfpb@localhost/cfgov
#export DATABASE_URL=mysql://cfpb@localhost/cfgov
```

although this doesn't really add that much information. The specific details of the existing connection string (the `root` user and database named `v1`) are going to become legacy references to our current conventions and other MySQL scripts that are being removed in this PR. Those names shouldn't technically be necessary to use old dumps. The `?init_command='SET default_storage_engine=MYISAM'` part of the connection string is similarly a legacy thing and would only be needed when creating new tables, which people shouldn't be doing with old dumps.

Perhaps an alternate approach could be to remove the MySQL config as proposed in this PR and then add a page to the internal documentation describing how to load an old dump?

higs4281:
I'm sold on the alternate approach, removing MYSQL config as proposed and adding docs guidance for old dumps. I agree this will quickly become a rare need.

acrewdson:
@higs4281 @chosak 👍 Should I add the docs about loading old MySQL dumps as part of this PR?

chosak:
@acrewdson my suggestion would be to instead put those in the internal platform wiki, as they're specific to our internal use of dumps and not relevant to public users of this repository.

chosak:
@higs4281 :+1: we have a backlog task for that; that can't happen until after production moves off of MySQL which will be happening shortly.

willbarton:
Could we maybe just [remove `mysql.txt` line from Tox](https://github.com/cfpb/cfgov-refresh/blob/master/tox.ini#L139)? Do we need to run any tests against MySQL specifically anymore? That'd satisfying the ability to run Python 3 tests without manually removing/commenting-out that line.

acrewdson:
@chosak good idea, will do

chosak:
@willbarton 

> Could we maybe just remove mysql.txt line from Tox?

This occurred to me too - although, if we did this, we'd be deploying against MySQL without doing any testing at all against it. At least now we're testing that the MySQL package can be installed properly? Might as well wait a few weeks until this is all removed at once.

acrewdson:
@higs4281 @chosak I added a `Loading old MySQL dumps locally` section to the relevant platform wiki page.

higs4281:
Docs look great, and I think that's the right spot for them. Thanks for updating them.

PR #4246, opened 2018-06-12
Remove jest-cli

anselmbradford:
## Removals

- Removes `jest-cli` because it's installed by `jest` already.

## Testing

1. `./frontend.sh && gulp test:unit should pass`


anselmbradford:
Closing this for now till someone has time to look into why jest-cli is not installed at the top-level node_modules folder.

PR #4251, opened 2018-06-13
regs3k: Add Elasticsearch indexing

higs4281:
The regulations3k project will have a search component,
and this is an attempt to hook up an Elasticsearch index for regdown data.

## Testing
To try it out, you'll need an Elasticsearch index running locally, or use Docker setup. And it helps to have a few regs imported so we have content to play with.

```
./cfgov/manage.py migrate
./cfgov/manage.py update_index regulations3k
```

You'll need to create a `Regulations search page`, as a child of our landing page. The new page type should be one of your choices.

It doesn't need any content, but give it a slug (I used `reg-search` for this example).

Results are served by a routable child page that lives at `/regulations/reg-search/results/`

queries get attached to that URL, with options for free-text search ('q') or one or more reg numbers ('regs'). A query request would look like this:

`/regulations/reg-search/results/?q=disclosures&regs=1002,1004,1005`

Providing only one reg limits the search to that reg. Providing more than one widens the search. Providing no regs searches all active regs.

I have doubts that this is the path we want, but it can at least be a testing rig for search.


contolini:
This is amazing and will make working on the front-end a whole lot easier. 🎉

PR #4256, opened 2018-06-15
Add gulp npm script

### LINKED ISSUE #4254: Gulp woes
Gulp is something of a pain point, especially requiring a global install separate from this project. Here's some possible actions paired with the problems they solve:

1. Remove Gulp
  - Gulp is somewhat duplicative with webpack & npm scripts. We'd never encounter problems with gulp if we wholly remove it from the project.
2. Use the local gulp binary via an npm script
  - Rather than requiring a global install of `gulp-cli`, we can just create a `gulp` npm script that runs `gulp --`, which is then invoked with `npm run gulp build`, `npm run gulp clean`, etc.
3. In a pre-install step, add `./node_modules/.bin/` to the users path if it doesn't exist there already. This is kinda intrusive and could be a pain point because it isn't super clear what's going on, but would allow local gulp to be run automatically without having to wrap the calls as npm scripts.

Thoughts? @contolini @jimmynotjim @anselmbradford @cfarm @virginiacc @ascott1 @mistergone @Scotchester

wpears:
Closes #4254
## Additions

- A `gulp` npm script that allows the local gulp executable to be run

## Testing

 - `npm remove gulp-cli -g && which gulp`
    - Should be `gulp not found`
 - `npm run gulp`
    - Ensure gulp runs the build task
 - `npm run gulp clean`
  - Ensure gulp runs with passed arguments

### DIFF
--- a/frontend.sh
+++ b/frontend.sh
@@ -109,7 +109,7 @@ clean_and_install() {
 # Run tasks to build the project for distribution.
 build() {
   echo "Building project…"
-  gulp build
+  npm run gulp build
 }
 
 # Execute requested (or all) functions.
--- a/package.json
+++ b/package.json
@@ -93,7 +93,8 @@
   "scripts": {
     "postinstall": "node scripts/npm/apps-install",
     "preinstall": "npx enforce-node-version || (echo 'Please install npm v5.2 or greater' && exit 1)",
-    "test": "snyk test"
+    "test": "snyk test",
+    "gulp": "gulp --"
   },
   "snyk": true
 }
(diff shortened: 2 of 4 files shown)


wpears:
We should also update the [installation / usage instructions](https://cfpb.github.io/cfgov-refresh/installation/) as needed.

anselmbradford:
Do we want to update everything in this PR? Like:
- gulp install in Travis https://github.com/cfpb/cfgov-refresh/blob/master/travis_install.sh#L5
- gulp build command https://github.com/cfpb/cfgov-refresh/blob/master/frontend.sh#L112
- gulp test command in Travis https://github.com/cfpb/cfgov-refresh/blob/master/travis_run.sh#L11

anselmbradford:
Does this pass the `--travis` and `--headless` flags through? Wondering if that could be related to the Travis error.

PR #4262, opened 2018-06-18
Remove milliseconds from timestamp

anselmbradford:
Milliseconds were causing an HTML validation error because of the number of decimal places they used. Easiest to just remove them because it doesn't look like they are utilized anyway.

## Changes

- Removes milliseconds from timestamp.

## Testing

1. Pull down branch.
2. Visit homepage.
3. View source.
4. Search for `datetime` and see that `datetime=` attribute only contains seconds in the time, such as `2018-05-23T19:55:00-0400`.

### DIFF
--- a/cfgov/jinja2/v1/_includes/macros/util/format/datetime.html
+++ b/cfgov/jinja2/v1/_includes/macros/util/format/datetime.html
@@ -64,7 +64,7 @@
 
    Format a date for datetime metadata in the format of:
 
-       YYYY-MM-DDTHH:MM:SS
+       YYYY-MM-DDTHH:MM:SS-Z
 
    When given:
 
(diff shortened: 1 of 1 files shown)


cfgov/jinja2/v1/_includes/macros/util/format/datetime.html
```diff
@@ -64,7 +64,7 @@
 
    Format a date for datetime metadata in the format of:
 
-       YYYY-MM-DDTHH:MM:SS.MMMM
+       YYYY-MM-DDTHH:MM:SS
```
Scotchester: Could update this to show the time zone offset that appears at the end, too.

marteki: +1 to including the time zone.

anselmbradford: Done! As `-Z`

PR #4264, opened 2018-06-18
Remove obsolete pubdate boolean attribute

anselmbradford:
`pubdate` boolean attribute has been removed from the html spec. We could use microdata schema https://www.sitepoint.com/5-obsolete-features-html5/#2-the-pubdate-attribute, but since we don't use that elsewhere it's probably best to remove it altogether for now.

## Removals

- Remove obsolete `pubdate` boolean attribute

### DIFF
--- a/cfgov/jinja2/v1/ask-cfpb/answer-page.html
+++ b/cfgov/jinja2/v1/ask-cfpb/answer-page.html
@@ -41,7 +41,7 @@
                 block__flush-top
                 block__sub">
         {% if last_edited %}
-        <time datetime='{{ last_edited }}' pubdate class="answer-edited-date">updated {{ dt.format_date(last_edited) }}</time>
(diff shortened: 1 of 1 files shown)


anselmbradford:
> Could we use the datePublished value described in the SitePoint article? I'm seeing things online about the possible SEO impact of removing signifiers about the last published date; nothing conclusively, but could we err on the side of caution and keep the correct indicator instead of just removing it completely?

Since we don't use `pubdate` elsewhere, like [on the blog](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/post-preview.html#L102-L105), I was thinking it would be better if all our posts have the same format. We could add the schema markup, but it's probably better to do it all at once across all posts, which is likely out of scope here since this is just fixing the HTML validation error. Would you be all right with an issue for investigating whether the schema is necessary?

marteki:
@ans Yeah, that sounds like a better call.

anselmbradford:
[GHE]/CFGOV/platform/issues/2844 added.

PR #4267, opened 2018-06-19
Mega menu: Add aria attributes to footer text

virginiacc:
Mega menu footer links are often brief and rely on the surrounding text to be understood, so this automatically adds aria attributes to provide that context to users accessing the links via screen readers.

## Additions

- `NavFooter` block 
- MenuItem method that adds an `aria-label` attribute containing the full text of the paragraph to footer links and `aria-hidden` attributes to the surrounding text

## Changes

- Remove all editor options except paragraphs and links from mega menu nav footer content field

## Notes
- Automatically adding aria attributes depends on footer items being written so that the full text is usable as an aria-label on the link, which would necessitate a slight change to the text of the `Ask CFPB` link under `Consumer Tools`. 

## Testing

1. Check out branch
2. Run `python cfgov/manage.py migrate` 
3. Run`python cfgov/manage.py runscript migrate_mega_menu` if necessary to generate the mega menu
4. Turn on `WAGTAIL_MENU` flag to activate the Wagtail version of the mega menu 
5. Save a menu item (with or without edits to footer)
6. Open the mega menu and verify that aria attributes have been added to the footer items

## Screenshots
screen shot 2018-06-19 at 9 23 44 am

## Checklist

### Browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


cfgov/scripts/migrate_mega_menu.py
```diff
@@ -31,7 +31,7 @@
             'type': 'nav_footer',
             'value': {
                 'content': '<p>Browse answers to hundreds of financial ' +
-                'questions. <a href="/ask-cfpb/>Ask CFPB</a></p>' +
+                'questions with <a href="/ask-cfpb/">Ask CFPB</a></p>' +
```
jimmynotjim: Good catch on the closing quote.

virginiacc:
@anselmbradford That's a good point. Our content guidelines for footer items state that they should consist of or include a link, but it would probably make sense to verify that before adding the attributes just in case there's ever a need to add non-link content to the footer. I added a check for links in https://github.com/cfpb/cfgov-refresh/pull/4267/commits/460a10e3dfcc6e34b88730c20a335dd44f39c0af.

anselmbradford:
Sounds great! 👍 Works.

cfgov/v1/blocks.py
```diff
@@ -276,6 +277,44 @@ class NavGroup(blocks.StructBlock):
...
+        cleaned = super(NavFooter, self).clean(value)
+        if cleaned['content']:
+            html = expand_db_html(cleaned['content'].source)
+            cleaned['content'] = RichText(
+                self.accessible_links(html)
+            )
+        return cleaned
+
+    @staticmethod
+    def accessible_links(html):
```
richaagarwal: Do you think you could add a docstring here to describe specifically what this function alters?

cfgov/v1/blocks.py
```diff
@@ -276,6 +277,44 @@ class NavGroup(blocks.StructBlock):
...
+        help_text='If checked, this block will only show '
+        'on our sharing site (Content).',
+        label='Mark block as draft'
+    )
+    content = blocks.RichTextBlock(
+        required=False,
+        features=['link']
+    )
+
+    def clean(self, value):
```
chosak: If I follow this logic, this is taking what the user entered into the CMS editor, parsing it for links, modifying those links by adding attributes, and then saving that modified content in the database?

Is it possible instead to alter the links at serve time, like we already do extensively in something like `v1.parse_links`, or through a Wagtail editor hook of some kind? The downside of altering at save time is that we're persisting this particular implementation of link markup into the database - if in future it ever changes, we'll have to reprocess every link.

This code also makes some assumptions about the internal Wagtail logic around how links are stored in the DB (`expand_db_html` and `RichText` are undocumented). This makes it less future-proof than modifying at serve time.

virginiacc: @chosak That's a good point. I originally handled the parsing at serve time, but thought it might be cleaner to perform it once on save instead. If that's not the case because of possible future incompatibilities, I can switch back to parsing the links when the menu items are assembled.

richaagarwal: @chosak @virginiacc I'd actually be a fan of moving away from parsing for links at serve time (something I've been looking into as part of my work on how we can best manage external links) since it's redundant and also requires that people remember to put in the `parse_links` jinja filter when editing or creating a new template...  I plan to do a more complete write-up on why the current implementation doesn't work/ is buggy) but I hear the concerns about using `expand_db_html`.

chosak: @richaagarwal sure, I'd be interested to hear your thoughts.

Generally my main argument against storing these kinds of modifications in the DB is that it violates separation of content and presentation. The editor should be concerned with content only: what is being linked to and what should the link text say, and not need to care/think about HTML attributes. Things like link icons (`<a class>`) or labels (`<a aria-label>`) are part of the presentation, not the content. We don't want to have to think about modifying the page content every time we change the page presentation.

richaagarwal: @chosak Right, so I think there's a distinction between doing heavy processing to determine whether something is an external link, and then the actual mark-up that gets applied (which I agree would ideally happen at serve time).  But right now we're doing both at serve time, and we only should ever need to do the first part once / at page save time.  I know this particular PR isn't concerned with determining whether something is an external link or not, but if we do move away from having the `parse_links` filter called everywhere (inconsistently), then that wouldn't be as viable of an alternative.

cfgov/v1/blocks.py
virginiacc: [Docstring](https://github.com/cfpb/cfgov-refresh/blob/458bc21149d7015f1cc94264e099c0d24452044a/cfgov/v1/models/menu_item.py#L110-L115) added in https://github.com/cfpb/cfgov-refresh/pull/4267/commits/3c4d5a22a78f54ecdc0dc864930f6d3f68c6869f

cfgov/v1/blocks.py
virginiacc: @chosak @richaagarwal I switched to handling the links at serve time in https://github.com/cfpb/cfgov-refresh/pull/4267/commits/3c4d5a22a78f54ecdc0dc864930f6d3f68c6869f

PR #4295, opened 2018-06-29
Adds initial accessibility tracking custom gtm tag

anselmbradford:
This adds code for tracking the skip link button that shows when navigating the site using the tab key. This will have a corresponding tag in GTM.

## Additions

- Add accessibility interactions GTM tracking code listener.

## Changes

- Add dashes in jsdoc comment.

## Testing

1. Add `require( '../../apps/analytics-gtm/js/accessibility-interactions-listeners' );` to `common.js`
2. Run `npm run gulp clean && npm run gulp build`
3. Visit http://localhost:8000/?debug-gtm=true
4. Open the dev console
5. Hit the tab key and see that a tracking event logs to the console.


cfgov/unprocessed/apps/analytics-gtm/js/accessibility-interactions-listeners.js
```diff
@@ -0,0 +1,13 @@
+import {
+  addEventListenerToElem,
+  track
+} from './util/analytics-util';
+
+const AccessibleInteractionsListeners = ( function() {
```
jimmynotjim: This should be updated to an arrow function to match below and our updated standards.

anselmbradford: Fixed

anselmbradford:
How do we want to handle testing for these? I'm kinda inclined to ignore the analytics-gtm app in jest outside of the utility methods. For instance, in this one, say we set up a test with a mock of the skip link button HTML, the code is such that only line 9 (the query for `'#skip-nav'`) is unique to this file (the other methods are in the analytics utility, which do need coverage). So say the ID gets changed on the site, this tracking code will break here and tracking won't be reported to GTM, but the test will still pass. This may be a case where creating a test will add a maintenance burden that doesn't actually test for the scenario we want to avoid. I learned today that GTM can [trigger on a JavaScript error](https://www.tatvic.com/blog/javascript-error-tracking-google-analytics-via-google-tag-manager-unsung-hero-2/). Perhaps instead we would want to create an analytics utility (with test coverage) the standardizing how errors are thrown in the analytics, and then create a tag in GTM for reporting errors to the analytics team, who would then inform us of breakages. Just thinking out loud here.

anselmbradford:
@jimmynotjim I ignored the analytics-gtm scripts in the coverage, but not the analytics utility.

> How much work do you think it'll be to create the standard analytics utilities? Sounds like this is well out of scope of our project but maybe we can do it during the reset?

We have some pieces in place already, like we can log whether a query is not found https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/apps/analytics-gtm/js/util/analytics-util.js#L24. I also opened https://github.cfpb.gov/CFGOV/platform/issues/2869, since we have two approaches to integrating analytics. We may want to use https://github.com/cfpb/cfpb-analytics to include all the analytics utilities in cfgov-refresh and the satellite app repos.

jest.config.js
```diff
@@ -14,6 +14,7 @@ module.exports = {
     '<rootDir>/cfgov/unprocessed/apps/.+/node_modules/',
     '<rootDir>/cfgov/unprocessed/apps/.+/webpack-config.js$',
     '<rootDir>/cfgov/unprocessed/apps/.+/index.js$',
+    '<rootDir>/cfgov/unprocessed/apps/analytics-gtm/js/[a-zA-Z-]+.js$',
```
jimmynotjim: Ahhh, this is cool, much cleaner than comments in all the files as well.

PR #4298, opened 2018-06-29
Truncate overly-long referrers in feedback forms

chosak:
The migration from MySQL to Postgres introduced a bug where submissions to the feedback form would cause a 500 error if the referrer exceeded 255 characters. See platform#2857 for additional detail.

This change adds manual truncation of overly-long referrers so that these never get sent down into the database. This is done in a very minimal way by modifying the existing logic; a better way to do this would be to refactor the feedback forms to leverage Django form validation. I implemented the fix in this minimal way for expediency's sake.

New unit tests have been added that verify the new behavior. To properly run these tests against Postgres locally, use a command line like:

```
TEST_DATABASE_URL=postgres://user@localhost/db tox
```

Travis always runs unit tests using Postgres, so that should also provide some additional validation of these new tests.

## Checklist

### DIFF
--- a/cfgov/v1/handlers/blocks/feedback.py
+++ b/cfgov/v1/handlers/blocks/feedback.py
@@ -80,7 +80,20 @@ class FeedbackHandler(Handler):
                 )
             except (ValueError, TypeError):
                 pass
-            feedback.referrer = self.request.POST.get('referrer', '')
+
+            # The referrer is set manually here instead of being set on the
+            # ModelForm. Whereas MySQL silently truncates varchar fields that
+            # exceed the column length, Postgres does not and will raise a
+            # "value too long for type character varying" error if an
+            # excessively long referrer is saved.
+            #
+            # This code grabs the referrer from the request and truncates it
+            # to at most the length of the feedback model referrer field.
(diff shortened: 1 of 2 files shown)


cfgov/v1/handlers/blocks/feedback.py
```diff
@@ -80,7 +80,20 @@ def get_response(self, form):
...
+            # The referrer is set manually here instead of being set on the
+            # ModelForm. Whereas MySQL silently truncates varchar fields that
+            # exceed the column length, Postgres does not and will raise a
+            # "value too long for type character varying" error if an
+            # excessively long referrer is saved.
+            #
+            # This code grabs the referrer from the request and truncates it
+            # to at most the length of the feedback model referrer field.
+            referrer = self.request.POST.get('referrer', '')
+            referrer_field = feedback._meta.get_field('referrer')
```
higs4281: I wouldn't have thought to grab the max length directly from _meta. Nice.

chosak: Thanks @higs4281. I didn't want to have it hardcoded in two places.

PR #4303, opened 2018-07-02
Regs3k: Add processing for tables, forms, and images

higs4281:
Also: organize parsing operations in a parser module.

Regs imported with this code can be reviewed on build. Two good examples:
- /1026/H/ (184 images)
- /1010/A/ (14 tables)

## Notes:
* Since we limit external image sources via a whitelist, `www.ecfr.gov`
needs to be added for now so we can test image rendering options.
* Most of the changes in this PR are reorganization of code.
The main additions are:
    * The addition of a `parser` module for parsing machinery.
    * The addition of a RegTable class to parse and store tables.
    * Changes to ecfr_importer's `parse_appendix_elements` function
    to pick up tables, forms and images.
* TODO: Move the importer into the parser module and create a
management command to run it.


higs4281:
Agreed, @willbarton 
`regulations3k.parser` it is.

PR #4316, opened 2018-07-05
Add TreeDomModel

anselmbradford:
Fixes [GHE]/CFGOV/platform/issues/2867. Adds a TreeDomModel module for migrating tree-initialization code out of the MegaMenu module.

## Additions

- Adds TreeDomModel module for populating a tree with dom nodes that match a certain selector.

## Testing

1. `gulp test:unit` should pass and increase overall coverage past 50%.
2. `gulp clean && gulp build` and test that menu initializes and works as before.


cfgov/unprocessed/js/modules/TreeDomModel.js
```diff
@@ -0,0 +1,86 @@
...
+      return;
+    }
+    let child;
+    for ( let i = 0, len = children.length; i < len; i++ ) {
+      child = children[i];
+      let childNode = parentNode;
+      // If the element matches the selector, add a tree node and run callback.
+      const matchesSelector = _getMatchesMethod( child );
+      if ( matchesSelector.bind( child )( selector ) ) {
+        childNode = parentNode.tree.add( null, parentNode );
```
anselmbradford: @jimmynotjim this is the one area I'm not so into with this update. As the tree nodes are created and added to the tree their data payload is empty (`null`), since we don't know at this point that they will be holding menus (FlyoutMenu instances), the data payload is then set with [`treeNode.data = menu`](https://github.com/cfpb/cfgov-refresh/pull/4316/files#diff-7c890f8c0c7dc76e2ab1fba0db2d8f64R113) in the `_addMenu` function in the MegaMenu module. Consequently, the data payload is touched twice, once here to set it to null initially and once to add the menu when a new FlyoutMenu instance is created. Before creation of the menu and creation of the new tree node happened in the same function (`_addMenu`) so the data payload could be set at the same point, so this two-step process to setting it is a consequence of separating tree node creation from FlyoutMenu creation.

willbarton:
TREEDOM

anselmbradford:
Decided to close this as the extra abstraction is not worth it.

PR #4330, opened 2018-07-09
Fix HTML validation errors in Spanish pages

anselmbradford:
## Changes

- Converts tabs to spaces.
- Removes unnecessary roles.
- Removes unclosed link.
- Removes units from iframe element dimensions. 
- Converts div to span for error where div was not allowed in h1 at that point.

## Testing

1. HTML Validation errors are reduced on /es/ pages vs the live site. There is the error for "Bad value “none” for attribute “media” on element “link”: Media query ended prematurely." and for the FB like button on the Spanish homepage still.

## Notes

I suspect the FB like button was copy-pasted into the template with initialization that should be handled via JS. Anyone want to like the page and check that that works? ([GHE]/CFGOV/platform/issues/2878)


cfgov/jinja2/v1/es/comprar-casa/index.html
```diff
@@ -1,161 +1,197 @@
...
-		</div>
-	</div>
+    <div class="hero-container">
+        <img class="hero-image"
+             src="{{ static('nemo/_/img/mortgage-hero.png') }}"
+             alt=""
+         style="padding-left: 434px;">
+        <div class="hero-copy">
+            <h1>
+                <span class="heading-inner-small" style="display:block;">
```
anselmbradford: `style="display:block;` added here because I needed to convert the `div` to a `span`, because the former is not allowed in an `h1` at this point according to the HTML validator.

jimmynotjim:
> "Bad value “none” for attribute “media” on element “link”

Is expected, it's the same hack we came across on the rest of the pages.

PR #4332, opened 2018-07-09
Align inline interpretations to paragraph text

niqjohnson:
Align inline interpretations to the text of the paragraph they interpret

## Additions

- `inline-interpretation` class to inline interpretation expandables
- LESS to style inline interpretation expandables

## Testing

1. Fire up Regulations 3000 as normal
2. Go to any section that has inline interpretations
3. Verify that inline interpretation styles match those seen in the screenshot below; namely:
    - Inline interpretations are only as wide as the main column of text (i.e. they don't bleed past the right edge of the paragraph text)
    - Each Inline interpretation is indented to the same level as the interpreted paragraph right above it

## Screenshots

inline-interp-widths

## Notes

- These styles rely on inline interpretations always directly following the paragraph they interpret, which I believe will always be the case. If we don't think the next sibling combinator method used in these styles will be bulletproof enough, though, we can refactor to add a `level-N` class to the inline interpretations themselves instead.

## Checklist

### Browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [x] Keyboard friendly
- [x] Screen reader friendly

### Other

- [x] Is useable without CSS
- [x] Is useable without JS

### DIFF
--- a/cfgov/regulations3k/jinja2/regulations3k/inline_interps.html
+++ b/cfgov/regulations3k/jinja2/regulations3k/inline_interps.html
@@ -1,7 +1,8 @@
 <div data-qa-hook="expandable" class="o-expandable
             o-expandable__expanded
             o-expandable__borders
-            o-expandable__midtone">
+            o-expandable__midtone
+            inline-interpretation">
 
     <button class="o-expandable_target">
         <div class="o-expandable_header">
--- a/cfgov/unprocessed/apps/regulations3k/css/main.less
+++ b/cfgov/unprocessed/apps/regulations3k/css/main.less
@@ -9,6 +9,7 @@
 @import (less) './reg-pagination.less';
 @import (less) './reg-search.less';
 @import (less) './reg-text.less';
+@import (less) './reg-inline-interpretations.less';
 
 .section-nav {
   .u-clearfix;
--- a/cfgov/unprocessed/apps/regulations3k/css/reg-inline-interpretations.less
+++ b/cfgov/unprocessed/apps/regulations3k/css/reg-inline-interpretations.less
@@ -1,0 +1,19 @@
+/* ==========================================================================
+   Regulations 3000
+   Regulation inline interpretation styling and cf-expandable overrides
+   ========================================================================== */
+
+.inline-interpretation {
+    max-width: 41.875rem;
+}
+
+.generateInterpLevel(6);
+
+// Generate six indentation levels for inline interpretations
+.generateInterpLevel(@n, @i: 1) when (@i =< @n) {
+    .level-@{i} + .inline-interpretation {
+        margin-left: (@i * 2rem);
+        max-width: (41.875rem - (@i*2));
+    }
+    .generateInterpLevel(@n, (@i + 1));
+}


anselmbradford:
Should this actually be a modifier for expandable content? Is this a one-off, or will it be used elsewhere?

contolini:
@anselmbradford Good question -- I think it's a one-off thing. This PR relies on classes we assign to eRegs' paragraphs (`.level-@{i}`) and it's unlikely we'll be using those classes outside eRegs.

jimmynotjim:
Will the expandable always be preceded by the `.level-@{i}` paragraph? Could we use a sibling selector `.level-@{i} + .o-expandable` here instead of also adding a class to the expandable?

anselmbradford:
> Good question -- I think it's a one-off thing. This PR relies on classes we assign to eRegs' paragraphs (.level-@{i}) and it's unlikely we'll be using those classes outside eRegs.

Okay! Alternatively, you could have something like `.o-expandable_content__inline-interp`, which would likely be what it would need to become if integrated into CF (unless it's its own atomic component).

niqjohnson:
@anselmbradford: Yep, these indented, limited-width expandables are—at least right now—only used in the content of regulations. I couldn't think of good a use case for a style like this anywhere else on the site, so I thought a one-off implementation made sense. I don't have strong opinions on the class name, though; I can change it to something like `.o-expandable_content__inline-interpretation` if that makes more sense to everyone.

@jimmynotjim: Yep, the expandables are always going to be preceded by a `.level-@{i}` paragraph. I used the next sibling combinator of `.level-@{i} + .inline-interpretation` to select the expandables, but `.level-@{i} + .o-expandable` would work, too. It felt vaguely wrong to not give these special-case expandables their own class, though, so I used `.inline-interpretation` in the selector instead of `.o-expandable`.

jimmynotjim:
If you were going to follow BEM style, I'd say give it a class name like Ans suggests, but since it's it's own class unrelated to the expandable, there's no real benefit here. 

Alternatively, you could wrap the sub-level in a wrapper and not even have to give the p and expandable adjustment styles, just set the indentation on the wrapper. Not sure how much work that is so sibling selectors is entire fine here.

contolini:
Somewhat related: Is `.level-X` a less-than-ideal class name? Should we make it more BEMy like `.a-paragraph__indented--X` or something?

jimmynotjim:
@contolini that's a good question. Ideally a component should not set it's own layout styles because it should be a self-contained unit, so no margin, or floats on itself, but ok to do so on it's children (because they are also self contained). Instead the modifications for layout should come from the parent container. An example is the max line-length on paragraphs, the max-width comes from the `m-full-width-text` container, not on the paragraphs, but if we wanted to target the paragraph for a different layout modifications we'd do so with `m-full-width-text p`.

In this case wrapping the paragraph and expandable (guessing there's probably lists too?) in a container like `o-regulation-level` (just a suggestion, choose the name that's right for you) and then setting the left padding (because margin should be set on it's container) on `o-regulation-level` would be the best solution. If there was a necessary layout change on a child of `o-regulation-level` you'd do that with `.o-regulation-level p, .o-regulation-level .o-expandable, ...`. It's a bit hard to explain by text and it took me a while to grok, but if you want to chat more about our Atomic Design + BEM principles let me know.

niqjohnson:
Feels like reorganizing and renaming these elements might be a bigger thing than this PR—and my meager pseudo-FEWD skills—should handle. So how's this for a plan for now?

1. If everyone is OK with it, I'll merge this PR, since it seems to be working as intended and fits with our current non-BEM Regulations 3000 naming scheme.
2. After this is merged, we can think more about how best to structure things and, if we end up making any changes, opening up a separate PR for those?

That said, I'm also not opposed to closing this PR if we don't want to go any further down our non-BEM path.

jimmynotjim:
@niqjohnson I wouldn't hold up this work just to make it more Atomic and BEM like. I think it's better to release it as is and iterate on it. Might be a good idea to add it to the project backlog so we're sure it doesn't get lost though.

higs4281:
fyi, this is the most deeply indented interp in the most deeply indented source graf (/regulations/1026/35/#c-2-vii-A-1), and it hangs in there.

deepest_interp

PR #4339, opened 2018-07-11
ignore common.js in app tests and add test coverage for hasClass

anselmbradford:
## Changes

- Ignore `common.js` in app tests as it's equivalent to `index.js`.
- Add `hasClass` test for `dom-tools-spec.js` 

## Testing


test/unit_tests/apps/owning-a-home/js/dom-tools-spec.js
```diff
@@ -35,4 +35,20 @@ describe( 'dom-tools', () => {
       );
     } );
   } );
+
+  describe( 'hasClass()', () => {
+    it( 'should return true when element has class', () => {
```
jimmynotjim: Minor: updating the descriptions to include "the provided class" would make it clearer we're not testing if it has a class at all, I got mixed up on that when I saw the query selector for the false test was a class name.

anselmbradford: I omitted the `the` so it'll fit on one line :)

PR #4342, opened 2018-07-12
Regs3k: Add search-term highlighting

higs4281:
This is a relatively simple option that makes search terms bold, leaves a reasonable window of text around the search term for context and thus isn't too jarring when jumping to the result's reference link.

The highlighting can be test-driven on dev5 at`/regulations/reg-search/results/?q=appraisal+fee`

simple_highlighting

### Notes
Other Haystack options remove content up to the first instance of the search term, which has  drawbacks:
- Highlighting of the first instance of the term becomes redundant, since the terms are always the first words of a result.
- It often fails to provide reasonable context and can be disorienting when clicking on the result link.

### DIFF
--- a/cfgov/regulations3k/models/pages.py
+++ b/cfgov/regulations3k/models/pages.py
@@ -10,9 +10,7 @@ from django.template.loader import get_template
 from django.template.response import TemplateResponse
 from django.utils.functional import cached_property
 from django.utils.text import Truncator
-# from haystack.utils import Highlighter
 from haystack.query import SearchQuerySet
-from jinja2 import Markup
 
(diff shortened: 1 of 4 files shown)


cfgov/regulations3k/models/pages.py
```diff
@@ -59,7 +61,8 @@ def regulation_results_page(self, request):
         search_query = request.GET.get('q', '')  # haystack cleans this string
         sqs = SearchQuerySet()
         if search_query:
-            sqs = sqs.filter(content=search_query)
+            sqs = sqs.filter(content=search_query).highlight(
+                pre_tags=['<strong>'], post_tags=['</strong>'])
```
contolini: It's cool to see Haystack offers highlighting support out of the box.

PR #4345, opened 2018-07-12
Remove the roles from menu content blocks and links

jimmynotjim:
The links are spoken twice when using Voice Over's auto-read feature. Turns out that it reads the content as "[content] - menu item, [content]" instead of just "[content] - link". Removing the roles fixes this issue but brings back the issue of the content divs having no roles. Converted those to section elems and the VO experience is much improved, but there are now three instances where the sections have no heading (one at the top level, two at the third level). One possible solution is including headings within those that are visually hidden.

__NOTE: This is an experiment and intended to spark a discussion around our best practices, it works, but I don't suggest merging as is till we've had time to play and test it as a team.__ - Jimmy


jimmynotjim:
One other consideration that Cat just made me aware of is the `role="menu"` is intended for applications, not websites - https://github.com/cfpb/development/pull/181/files#diff-63cb6ce53ea17d2e235290c1057d6e00R117

The evidence for not relying on roles for menus seems to be mounting.

cfarm:
Removing the `menu` and `menuitem` roles from the navigation is 100% in line with W3C recommendations and the ARIA standards. I've written some guidance in the [development repo](https://github.com/cfpb/development/blob/100a8bd2e723b661e69742b52d5ed7b7819327fb/guides/screen-reader-differences.md#4-application-mode) about this type of misuse of the ARIA application widgets and the negative usability impact they have on screen reader users on Windows applications (such as JAWS and NVDA, the [two most popular screen reader apps](https://webaim.org/projects/screenreadersurvey7/#used)).

So in this case, it will fix a bug on VoiceOver and it will also improve the experience for users of JAWS and NVDA.

jimmynotjim:
Video of this branch's menu using Voice Over https://drive.google.com/open?id=1K8byWlvvo30_7mW6jtZsVEnvjXYgmq7A

cfarm:
@jimmynotjim Is this branch on a dev server so we can try it out on our iPhones to test iOS VoiceOver? If not, lmk if you need me to deploy it to one.

PR #4356, opened 2018-07-18
Mega menu: use relative urls for page links

virginiacc:
Ensures that mega menu links to Wagtail pages use relative paths so navigation works as expected on sharing site.

## Additions

- `get_relative_path` method on base page model

## Changes

- use `get_relative_path` instead of `url` method for Wagtail pages in the mega menu template

## Testing

1. Check out branch
2. Generate Wagtail version of mega menu if necessary: `python cfgov/manage.py runscript migrate_mega_menu`
3. Turn on `WAGTAIL_MENU` flag
4. Set up local sharing site (with an entry under `Sites` for sharing site)
5. Verify that mega menu links go to sharing site pages on sharing site and to default site pages on default site

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
+++ b/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
@@ -185,7 +185,7 @@
 {% macro _nav_item( nav_depth, nav_item ) %}
 {% set link = nav_item.link if nav_item.link else nav_item %}
 {% if link %}
-    {% set url = link.page_link.url if link.page_link else link.external_link | default('#') %}
+    {% set url = link.page_link.specific.get_relative_path() if link.page_link else link.external_link | default('#') %}
     {% set text = link.link_text %}
     {% set has_children = nav_item.nav_groups | count > 0 or nav_item.nav_items | count > 0 %}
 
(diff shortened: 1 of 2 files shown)


chosak:
@virginiacc not knowing much about the context/schedule for this change, would this instead be solved if we (at long last) moved to a single Wagtail Site? See GHE#2375 for some further discussion of this; now that we're off Wordpress, this is more unblocked than it used to be. Now might be a good time to make that change, and it'd pay off with simplification in lots of other places.

Technically, your step 4 above shouldn't require "with an entry under Sites for sharing site" -- we should need a Sharing Site, yes, for the default site, but not a separate Site as well.

This logic in this PR makes sense -- although there's some overlap with similarly unfortunate logic [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/routing.py#L33) -- but feels like a workaround that'd be better avoided if possible. As an additional minor concern, this logic adds a bunch of database queries per page load: the `Page.specific` and then getting the site for each page.

Scotchester:
Is there really no built-in way to get a root-relative URL for a page? :(

virginiacc:
@Scotchester The Wagtail page url methods normally return relative urls within a site, but for historical reasons we still have a separate Sites entry for our sharing site. I was a bit surprised there isn't a method that always returns a relative path, but I guess that wouldn't be necessary under usual circumstances.

Scotchester:
Ah, I guess that's what Andy was getting at. Thanks for the extra clarification.

virginiacc:
This is no longer necessary since we have moved to a single Site. Thanks, @chosak!

PR #4359, opened 2018-07-19
Bump TDP release to 1.0.4

nbucknor:
Added TDP Activity models

## Additions
- Added mptt and haystack to installed apps
- Updated TDP version to 1.0.4

@chosak @willbarton  @higs4281


Scotchester:
We need to hold releasing this, if possible, until we get the go ahead to release the logo updates (in 1.0.3).

PR #4364, opened 2018-07-20
regs3k: Lint paragraphs for italics and dashes.

higs4281:
QA testers identified some recurring format errors:
* Improper boldface on references such as **see** and **et seq.**
* Hyphens where em dashes should be.
* Interp graph references that didn't match the content (GHE issue 209)
* Some interp sections were not broken out in Reg X because of non-standard section naming.

This PR adjusts the importer and adds a paragraph linting routine to fix such cases.

## New dependency
Neither unicode nor html entities work as-is for passing an em dash character
from regdown to our templates, and either would sprinkle code through our regdown.  
To give editors a friendly way to handle em dashes, this PR adds a requirement for 
a Markdown extension,  `markdown-emdash`. It translates a triple hyphen 
--- a convention used in other flavors of Markdown and LaTex --- into an em dash.

This appears to play nice with Markdown tables (which also use multiple hyphens, to denote colums) and with our pseudoform underline extension.

It's still viable to enter a utf-8 '—' em dash character, but to use it an editor needs to know how to generate it or where to find a source to copy. The triple hyphen is more obvious and avoids using a non-ascii character.

## Testing

You can check the linting by importing a fresh copy of Reg E:

If you have a copy of ECFR xml in develop-apps, this would import it:

```
./cfgov/manage.py runscript ecfr_importer --script-args 1005 develop-apps/ECFR-title12.xml
```

Then uncheck 'draft' status on the newly imported effective-version. You should be able to
check the following spots for linted formatting:

### bold/italics

<http://localhost:8000/policy-compliance/rulemaking/regulations/1005/2/#b-2>

This spot shows both a bold 'see' and a bold 'et seq.'

Before linting:

before_linting_1005_2

After linting:

after_linting_1005-2

### em dashes

<http://localhost:8000/policy-compliance/rulemaking/regulations/1005/12/#c>

This graph should show a hyphen that is fixed to be an em dash.

Before linting:

before_dashing

After linting:

after_dashing

### Interp graph labels
The interp reference labels in <http://localhost:8000/policy-compliance/rulemaking/regulations/1026/1/> should now match the graphs they refer to.

### Reg X interp sections

If you import Reg X 1024 anew, you should find some more Interp sections (such as Interp-30).  
They were previously delivered inside interp-17 and thus ignored.

### DIFF
--- a/cfgov/regulations3k/parser/paragraphs.py
+++ b/cfgov/regulations3k/parser/paragraphs.py
@@ -6,6 +6,8 @@ by the patterns.IdLevelState class.
 """
 from __future__ import unicode_literals
 
+import re
+
 
 def bold_first_italics(graph_text):
     """For a newly broken-up graph, convert the first italics text to bold."""
@@ -34,6 +36,32 @@ def graph_top(graph_text):
         'paragraph')[0].partition(
         '12 CFR')[0].partition(
         '\xa7')[0][:200]
+
+
+def lint_paragraph(graph_text):
+    """Clean formatting anomalies.
+
+    - Missing em dashes
+    - restoring italics
+    """
+    fix1 = restore_emdash(graph_text)
+    fix2 = restore_italics(fix1)
+    return fix2
+
+
+def restore_italics(graph_text):
+    fix1 = re.sub(
+        r'\*\*(see)\*\*', '*\g<1>*', graph_text, flags=re.IGNORECASE)
+    fix2 = re.sub(
+        r'\*\*(et[.]? seq[. ]?)\*\*?', '*\g<1>*', fix1, flags=re.IGNORECASE)
+    return fix2
+
+
+def restore_emdash(graph_text):
+    stripped = graph_text.rstrip()
+    if stripped.endswith('-'):
+        return stripped + '--\n'
+    return graph_text
 
 
 def pre_process_tags(paragraph_element):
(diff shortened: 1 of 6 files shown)


cfgov/regulations3k/parser/paragraphs.py
```diff
@@ -36,6 +38,32 @@ def graph_top(graph_text):
...
+    fix1 = restore_emdash(graph_text)
+    fix2 = restore_italics(fix1)
+    return fix2
+
+
+def restore_italics(graph_text):
+    fix1 = re.sub(
+        r'\*\*(see)\*\*', '*\g<1>*', graph_text, flags=re.IGNORECASE)
+    fix2 = re.sub(
+        r'\*\*(et[.]? seq[. ]?)\*\*?', '*\g<1>*', fix1, flags=re.IGNORECASE)
```
contolini: What're all the possible combinations for `et. seq.`? Can there be both a `.` and a space after `seq`? If so, maybe `\*\*(et\.? seq\.? ?)\*\*?` would work better? https://regexr.com/3sr2f

higs4281: Good catch! Thank you.

contolini:
I can't figure out how to post an `approved` review after leaving a `comment` review so I'll just leave this emoji instead: ✅

PR #4373, opened 2018-07-25
Regs3k: Add support for copying Effective Versions

willbarton:
This PR adds a button to the Effective Version admin to allow copying an effective version and all its subparts and their sections to another version, giving regulations content editors the ability to duplicate an effective version instead of creating a new one from scratch.

### DIFF
--- a/cfgov/regulations3k/copyable_modeladmin.py
+++ b/cfgov/regulations3k/copyable_modeladmin.py
@@ -16,7 +16,7 @@ from treemodeladmin.views import TreeViewParentMixin
 class CopyButtonHelper(TreeButtonHelper):
 
     def copy_button(self, pk):
-        cn = self.edit_button_classnames
+        cn = 'button button-small button-secondary'
         return {
             'url': self.url_helper.get_action_url('copy', quote(pk)),
             'label': 'Copy',
--- a/cfgov/regulations3k/copyable_modeladmin.py
+++ b/cfgov/regulations3k/copyable_modeladmin.py
@@ -1,0 +1,91 @@
+from __future__ import unicode_literals
+
+from django.conf.urls import url
+from django.contrib.admin.utils import quote
+from django.contrib.auth.decorators import login_required
+from django.shortcuts import redirect
+from django.utils.decorators import method_decorator
+
+from wagtail.contrib.modeladmin.views import InstanceSpecificView
+
+from treemodeladmin.helpers import TreeButtonHelper
+from treemodeladmin.options import TreeModelAdmin
+from treemodeladmin.views import TreeViewParentMixin
+
+
+class CopyButtonHelper(TreeButtonHelper):
+
+    def copy_button(self, pk):
+        cn = self.edit_button_classnames
+        return {
+            'url': self.url_helper.get_action_url('copy', quote(pk)),
+            'label': 'Copy',
+            'classname': cn,
+            'title': 'Copy this {}'.format(self.verbose_name),
+        }
+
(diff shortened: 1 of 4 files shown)


cfgov/regulations3k/copyable_modeladmin.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class CopyButtonHelper(TreeButtonHelper):
+
+    def copy_button(self, pk):
+        cn = self.edit_button_classnames
+        return {
+            'url': self.url_helper.get_action_url('copy', quote(pk)),
+            'label': 'Copy',
+            'classname': cn,
```
higs4281: cn is an empty list. If you set 'classname' to the following, the button will match its siblings:

```
'classname': "button button-small button-secondary",
```

sibling_buttons

cfgov/regulations3k/wagtail_hooks.py
```diff
@@ -45,6 +48,34 @@ class EffectiveVersionModelAdmin(TreeModelAdmin):
     child_model_admin = SubpartModelAdmin
     parent_field = 'part'
 
+    def copy(self, instance):
+        # Create a new instance
+        new_version = EffectiveVersion(
+            authority=instance.authority,
+            source=instance.source,
+            part=instance.part,
+            effective_date=date.today(),
```
higs4281: I would vote for `instance.effective_date` as a starting point. 
'today()' will almost never be right, although it might be closest in time to a prospective new effective_date.

I worry that it might be confusing to editors if `acquired_date` and `effective_date` pop up as the same.

willbarton: Avoiding confusion was my motivation not to use the same `effective_date`, because that's currently the unique identifier for a human looking at the list of versions.

Created might be better, because then that becomes the distinguishing field for telling the difference between the version you copied and the new copy.

higs4281: It is hard to imagine a common use case for creating a new tree with the same effective date, so copying the old date would almost always be wrong, too. I don't object to today() as a starting point.

willbarton: Your changes in #4387 pretty much resolve all my concerns about clarity of what you've just done by clicking "Copy". We'll keep the effective date the same and update the created date.

cfgov/regulations3k/copyable_modeladmin.py
willbarton: I seem to have broken that getting the superfluous modeladmin heavy-customization code out of the way.

cfgov/regulations3k/wagtail_hooks.py
```diff
@@ -45,6 +48,34 @@ class EffectiveVersionModelAdmin(TreeModelAdmin):
     child_model_admin = SubpartModelAdmin
     parent_field = 'part'
 
+    def copy(self, instance):
+        subparts = instance.subparts.all()
+
+        # Copy the instance
+        new_version = instance
+        new_version.pk = None
+        new_version.acquired = date.today()
```
willbarton: Depending on whether this or #4387 gets merged first, this will have to be changed to the "created" field.

cfgov/regulations3k/wagtail_hooks.py
```diff
@@ -45,6 +48,34 @@ class EffectiveVersionModelAdmin(TreeModelAdmin):
...
+        new_version.draft = True
```
willbarton: Likewise for this line.

PR #4337, opened 2018-07-11
Revert #4334 Deprecate sheerlike middleware

chosak:
This reverts commit 793f0aa30a95e485c73197f4774edac6ac9ac449 (PR #4334).

The change in that PR (removing a manually inserted `request` object in the Jinja context) breaks some of the satellite apps - at least the [retirement](https://github.com/cfpb/retirement) one that uses the deprecated Django `render_to_response` method for rendering templates. Until these satellite apps are updated to use Django `render` (that provides the request), we can't make this change.

Reverting now to unblock other needed PRs/deployment that needs to happen today.

## Checklist

- [X] PR has an informative and human-readable title
- [X] Changes are limited to a single goal (no scope creep)
- [X] Code can be automatically merged (no conflicts)
- [X] Code follows the standards laid out in the [development playbook](https://github.com/cfpb/development)
- [X] Passes all existing automated tests
- [X] Any _change_ in functionality is tested
- [X] New functions are documented (with a description, list of inputs, and expected output)
- [X] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #4380, opened 2018-07-26
Remove Sheerlike middleware and app

chosak:
This change removes the `sheerlike` app and deprecates use of the `sheerlike.middleware.GlobalRequestMiddleware`. This middleware did two things:

1. Added a global `request` variable to the Jinja2 context, passed via the local Python thread.

This is no longer needed as the request should be properly passed into all Django and Jinja-rendered templates.

2. Adds the ability to access the request meta variables using dictionary syntax, e.g. `request.headers['foo']` gets converted to `request.META['HTTP_FOO']`.

We don't access headers this way anymore (`git grep headers -- *.py`).

All unit tests continue to pass.

This change was previously applied in #4334 but had to later be rolled back in #4337 because of an incompatibility with the optional retirement app and its dependence on a global `request` variable. This dependency has since been removed (in #4357) and is no longer a blocker.

~~In addition, this PR refactors the `get_protected_url` Jinja2 tag and `Page.generate_view_more_url` method, which previously depended on this global request variable as well. The definition of this tag has been moved to a more reasonable place, and unit tests [and documentation](https://github.com/cfpb/cfgov-refresh/compare/master...chosak:remove-sheerlike-middleware?expand=1#diff-09223057401711731471e7b7b6142fc4R17) have been added.~~ This was implemented in #4423 and has been rebased out of this PR.

## Checklist


cfgov/v1/jinja2tags.py
```diff
@@ -11,6 +13,27 @@ def date_formatter(dt, text_format=False):
     return extended_strftime(dt, format)
 
 
+def get_protected_url(context, page):
+    """Output a page's URL if it is live and '#' if not.
+
+    Used when inserting a link to a Wagtail page. If the page is live, we want
```
richaagarwal: I don't think this is used/ called at the time an editor is inserting a link -- it seems like it's scattered throughout templates and called wherever a developer remembered to wrap potential page links with this?  (Somewhat similar to the process we currently have for `parse_links`).  I know refactoring how it's called is out of scope for this PR, but don't quite follow this docstring.

chosak: @richaagarwal you're right that this is very similar to the current `parse_links` setup. Including this template tag is dependent on the developer, and I'm not sure if it's applied across the board. Unfortunately there's no easy way to simplify this using a middleware like you're doing.

I would actually prefer to remove this logic completely but to reduce the scope of this PR I just tried to clean it up and add tests to clarify what's already there.

Regarding the docstring, would a better way to start the wording be

> This method can be used in templates to insert a link to this page that will be redacted if the page is not yet live. If the page is live...

richaagarwal: That sounds good -- thanks!

chosak:
I've fixed a few unit tests that were broken in the first version of this PR, and pushed a few more commits that contain those fixes. I'd be grateful if these commits could be given a once-over:

- 53159dc - @Scotchester as we discussed, this removes use of `get_protected_url` from the `ItemIntroduction` block and adds logic to handle a missing filterable parent (see commit message for more details).
- 0e9fb60 - @virginiacc this removes use of `get_protected_url` from the `JobListingList` block and also corrects the unit tests to properly pass it in when rendering the block (see commit message for more details).

chosak:
Due to the changes in #4417, this PR now removes all of the remaining functionality of the `sheerlike` app. I've added an additional commit in a705141 that removes all remaining traces of that app.

richaagarwal:
Would it make sense to wait for the final outcome in GHE/CFGOV/platform/issues/2907 regarding `get_protected_url` and remove that altogether here?  Also happy to see that incrementally done, but just asking since this PR already introduces changes to it.

richaagarwal:
I noticed many of the tests on this branch are failing and also that the diff isn't showing the complete set of changes (e.g. the new definition of `get_protected_url` is nowhere to be found).  Not sure if something went wrong with the rebase?

chosak:
Ugh, thanks @richaagarwal, you're right, I messed up the rebase.

Regarding `get_protected_url`, I'd been thinking that if/when the decision to remove that is made, it'd make more sense to open another PR that just removed it from this repo. If that got reviewed/merged first, I could rebase this one. This PR has been open for a while and I've been hoping to get it merged (although screwing up the rebases doesn't help, thank you for looking at it) so didn't want to block it on that other discussion.

Looking at GHE#2907, though, I see that decision has been made 😄 so I will go ahead and open that other PR.

chosak:
Now that #4423 has been merged, the `sheerlike` app only contains a single piece of middleware. I've rebased this PR to a single commit (d9c1990); see that commit or the (edited) description above for a refresher on what that middleware does/did. This PR now removes the `sheerlike` app entirely from cfgov-refresh.

willbarton:
its-a-new-dawn

PR #4384, opened 2018-07-27
Support multiple caches for fragment cache extension

richaagarwal:
This generalizes the fragment cache extension code slightly so that what was previously the `fragment_cache` (defined in the environment) is not hardcoded to use the `post_preview` cache.  Instead, we explicitly pass in the name of the cache in the template, and use that to look up the cache.  

The reason for this change is to support additional caches, so that we're not using `post_preview` for different types of fragment caching.  Specifically, @virginiacc will be using fragment caching on the mega menu template, so this PR will allow us to set up a new cache - whether it's a more generic cache or one specific to the mega menu, it needs to be separate from the current `post_preview` cache.

For more context, this is the PR in which fragment caching was introduced: https://github.com/cfpb/cfgov-refresh/pull/3596

### DIFF
--- a/cfgov/v1/fragment_cache_extension.py
+++ b/cfgov/v1/fragment_cache_extension.py
@@ -1,4 +1,6 @@
-# Taken from http://jinja.pocoo.org/docs/2.10/extensions/
+# Based off of http://jinja.pocoo.org/docs/2.10/extensions/
+from django.core.cache import caches
+
 from jinja2 import nodes
 from jinja2.ext import Extension
 
@@ -23,11 +25,15 @@ class FragmentCacheExtension(Extension):
         # that line number to the nodes we create by hand.
         lineno = next(parser.stream).lineno
 
-        # now we parse a single expression that is used as cache key.
+        # The first argument is the cache key
         args = [parser.parse_expression()]
 
-        # if there is a comma, the user provided a timeout.  If not use
-        # None as second parameter.
+        # The second argument is the cache's name
+        if parser.stream.skip_if('comma'):
+            args.append(parser.parse_expression())
+
+        # If there is a third argument, the user provided a timeout.
+        # If not use None
         if parser.stream.skip_if('comma'):
             args.append(parser.parse_expression())
         else:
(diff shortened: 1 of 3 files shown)


Scotchester:
😮 

🤔 

Could this be related to the issue where blog post previews periodically just show the generic "Blog" category and icon instead of their usual "Info for consumers" or whatever? See: GHE/CFGOV/platform/issues/2825

richaagarwal:
@Scotchester Hm, not that I'm aware of.  This PR isn't addressing a bug, it's just modifying the way the fragment cache is configured to allow multiple caches.

Scotchester:
Sorry, wasn't referring to the changes in this PR, but to the general concept of the `post_preview` cache.

cfgov/v1/fragment_cache_extension.py
```diff
@@ -42,16 +48,15 @@ def parse(self, parser):
         return nodes.CallBlock(self.call_method('_cache_support', args),
                                [], [], body).set_lineno(lineno)
 
-    def _cache_support(self, name, timeout, caller):
+    def _cache_support(self, key, cache_name, timeout, caller):
         """Helper callback."""
-        key = self.environment.fragment_cache_prefix + name
```
chosak: Because these changes remove the use of both `self.environment.fragment_cache_prefix` and `self.environment.fragment_cache`, is there a need any longer to set the defaults as is done in the `__init__` method [here](https://github.com/cfpb/cfgov-refresh/pull/4384/files#diff-c73c333b7f1696264ffedeb4d9d8a100R12)? I think maybe that custom `__init__` can be removed if those are no longer needed.

richaagarwal:
@chosak Thanks so much for adding those tests!  Merging that into this branch.

richaagarwal:
@chosak Actually, I'm getting errors on all 4 of those tests so I'm going to undo that commit for now.  I tried working through it a little (e.g. both %s in line 62 needed to be wrapped in " ")  but kept getting other errors.  I'd like to unblock Virginia on some work but agree more test coverage would be great, so I'm going to proceed with merging this given existing tests pass, but maybe we can pair on getting your tests working soon?

cfgov/v1/fragment_cache_extension.py
richaagarwal: @chosak Good call, removed in 441832558dcbc336174f0ad52b931f24dbd15b4e

PR #4387, opened 2018-07-30
regs3k: Adjust admin sorting and labeling

higs4281:
User testing raised some issues with sorting and labeling in Wagtail.
* Subparts were not properly ordered when a reg had more than one section subpart.
* The "acquired" column heading for EffectiveVersion was confusing.
* The "draft" column for EffectiveVersion was confusing (red meant active)
* If more than one version were active, the live version wasn't obvious.

In addition to fixing the subpart sorting, this patch creates a new "status" column
for effective verisons that will distinguish between live, draft, previous, and future versions. 
So a reg with versions in each state might look like this:

version_status

## Testing

This patch includes a migration to rename the "acquired" field for EffectiveVersion, so first run the  migration:

```
./cfgov/manage.py migrate --noinput
```
Then:
- Dive into a reg with multiple section subparts, and subparts should be correctly ordered.
- Check out a reg with multiple versions and test the statuses.

**Notes**  
- An effective version can be a "Future version" only if its effective date is in the future AND it is  not a "draft."
- This addresses GHE issue 143 (subpart ordering) and an undocumented issue about the color of the  "draft" field (red was counter-intuitive).

### DIFF
--- a/cfgov/regulations3k/models/django.py
+++ b/cfgov/regulations3k/models/django.py
@@ -87,7 +87,7 @@ class EffectiveVersion(models.Model):
     authority = models.CharField(max_length=255, blank=True)
     source = models.CharField(max_length=255, blank=True)
     effective_date = models.DateField(blank=True, null=True)
-    acquired = models.DateField(blank=True, null=True)
+    created = models.DateField(blank=True, null=True)
     draft = models.BooleanField(default=False)
     part = models.ForeignKey(Part, related_name="versions")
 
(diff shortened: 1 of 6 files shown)


cfgov/regulations3k/models/django.py
```diff
@@ -87,7 +87,7 @@ class EffectiveVersion(models.Model):
     authority = models.CharField(max_length=255, blank=True)
     source = models.CharField(max_length=255, blank=True)
     effective_date = models.DateField(blank=True, null=True)
-    acquired = models.DateField(blank=True, null=True)
+    created = models.DateField(blank=True, null=True)
```
willbarton: I realize this would complicate the migration substantially, but I wonder if this should have a default of `date.today()`? That way creating a new, blank effective version always has the creation date.

higs4281: I'll add this if we need it. The importer and copy buttons were already resulting in created today(), but a default makes sense.

PR #4389, opened 2018-07-30
Fix post preview cache

richaagarwal:
## Overview

See GHE/CFGOV/platform/issues/2825 for bug report. 

This fixes a bug introduced by https://github.com/cfpb/cfgov-refresh/pull/3596, in which certain information on a post preview (such as the icon displayed) was getting cached across blog and newsroom pages, which displayed the icon differently, and with no way to differentiate.  This PR resolves it by moving that logic out of the cache.  

I also identified another issue where the author URLs were incorrectly getting cached, and this PR resolves it by ensuring a relative link is used on the blog and newsroom pages since it's unnecessary to pass in an absolute URL. 

## Testing

First, to view the bug locally, check out the `master` branch locally and run it with the cache enabled, e.g. `ENABLE_POST_PREVIEW_CACHE=1 ./runserver.sh`. You do not need to refresh your production dump, but you should be running on a production dump.  Depending on the exact data snapshot you have, you'll see that the icons are off either on http://localhost:8000/about-us/blog/ or on http://localhost:8000/about-us/newsroom/.  On whichever page the icons are off, the author URLs will be off too.   Alternatively, you can clear the cache by running the shell like so:
`ENABLE_POST_PREVIEW_CACHE=1 ./manage.py shell`
`from django.core.cache import caches`
`caches['post_preview'].clear()`
and then loading the newsroom page, and then the blog page -- you'll notice the blog page has the incorrect information because the newsroom page was cached first.

Now, switch to my branch `fix-post-preview`, and clear the cache.  Then run the server (with cache enabled still), and confirm that the links and icons are as expected.

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/post-preview.html
+++ b/cfgov/jinja2/v1/_includes/organisms/post-preview.html
@@ -60,12 +60,9 @@
 {% import 'macros/time.html' as time %}
 
 {% macro render(post, controls, url='', post_date_description='') %}
-    {% cache post.post_preview_cache_key, 'post_preview' %}
     {% set date_desc = controls.post_date_description or post_date_description or 'Published' %}
     {% set cat_controls = controls.categories %}
     {% set show_categories = cat_controls.show_preview_categories if cat_controls is defined else true %}
-    {% set page_url = url or get_protected_url(page) %}
-    {% set post_url = get_protected_url(post) %}
     <article class="o-post-preview">
         <div class="m-meta-header">
             <div class="m-meta-header_right">
@@ -83,19 +80,21 @@
                 {% if show_categories %}
                     {% import 'macros/category-slug.html' as category_slug %}
                     {% if cat_controls and 'newsroom' in cat_controls.page_type and is_blog(post) %}
-                        {{ category_slug.render('blog', page_url, '', false) }}
+                        {{ category_slug.render(category='blog', href=url) }}
                     {% else %}
                         {% for cat in post.categories.all() %}
                             {% if loop.index > 1 %}
                                 |
                             {% endif %}
-                            {{ category_slug.render(cat.name, page_url, '', false) }}
+                            {{ category_slug.render(category=cat.name, href=url) }}
                         {% endfor %}
                     {% endif %}
                 {% endif %}
             </div>
         </div>
-         {% if 'EventPage' in post.specific_class.__name__ %}
+        {% cache post.post_preview_cache_key, 'post_preview' %}
+        {% set post_url = get_protected_url(post) %}
+        {% if 'EventPage' in post.specific_class.__name__ %}
             {% set event = post.specific %}
             {% if event.start_dt %}
                 <div class="o-post-preview_image-container">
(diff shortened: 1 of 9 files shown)


cfgov/jinja2/v1/_includes/organisms/post-preview.html
```diff
@@ -164,15 +163,15 @@ <h3 class="o-post-preview_title">
             <div class="o-post-preview_byline-group">
             {% for author in post.get_authors() %}
                 {% if loop.index == 1  %}
-                    By <a href="{{ page_url }}?authors={{ author.slug }}">
+                    By <a href="{{ url }}?authors={{ author.slug }}">
```
richaagarwal: It's not obvious from the diff, but the changes here mean that `url` is either an empty string, or is passed in by the calling template.  The only time it is passed in is from a page with a post preview snapshot on it, of which there are two (http://localhost:8000/policy-compliance/amicus/, http://localhost:8000/policy-compliance/notice-opportunities-comment/), and otherwise we want a relative URL here, rather than hardcording the page, which presents issues for caching.

cfgov/jinja2/v1/_includes/organisms/post-preview.html
```diff
@@ -83,19 +80,21 @@
...
                             {% if loop.index > 1 %}
                                 |
                             {% endif %}
-                            {{ category_slug.render(cat.name, page_url, '', false) }}
+                            {{ category_slug.render(category=cat.name, href=url) }}
                         {% endfor %}
                     {% endif %}
                 {% endif %}
             </div>
         </div>
```
richaagarwal: This section above line 95 (the beginning of the cache) will no longer be cached, thus fixing the issue with the wrong icons displaying (determined by the `category_slug` macro).

cfgov/jinja2/v1/_includes/organisms/post-preview.html
```diff
@@ -83,19 +80,21 @@
...
-         {% if 'EventPage' in post.specific_class.__name__ %}
+        {% cache post.post_preview_cache_key %}
```
richaagarwal: This section above line 95 (the beginning of the cache) will no longer be cached, thus fixing the issue with the wrong icons displaying (determined by the category_slug macro).

chosak: Will this require some kind of manual step to clear the entire cache when this goes out? Let's say before the fix some page with primary key 123 has the existing section cached (the whole `render` macro) under cache key `post_preview_123`. After this change, the next time the page is rendered, the template will insert that cached content between the newly moved `{% cache %}{% endcache %}` tags. Will that make a double header appear -- first the new, non-cached one, and then a second one that was in the cache?

richaagarwal: Yes, exactly -- was planning to clear the cache as soon as this is deployed.

richaagarwal:
I want to expand on [this comment[(https://github.com/cfpb/cfgov-refresh/pull/4389#discussion_r206267295) and point out one remaining issue with this code, which is that the `url` passed in [from here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/sublanding-page/index.html#L41) still gets cached, it's just narrower in scope.  The default is now an empty string (instead of the absolute URL), and so the only time that `url` is not an empty string is from the code I link to above, which is triggered on a page with a post preview snapshot, which currently is only two pages (https://www.consumerfinance.gov/policy-compliance/amicus/, https://www.consumerfinance.gov/policy-compliance/notice-opportunities-comment/).  

Which is all to say that, if we expected there to be overlapping content between those two pages and other filterable pages, then we might have a problem with the `url` being incorrect.  As it stands, I don't expect this to be the case, but without knowing how a post preview snapshot may be used in the future, it's best to address it one way or the other.  I haven't yet come up with a straightforward way to address this, and I don't want to prevent the other fixes in this PR from going through, so I'm merging -- but wanted to mention it here for now. I will also create an issue in GHE to document this. Thanks to @chosak for pointing this out.

PR #4391, opened 2018-07-30
Unified env

wpears:
This PR adjusts how environment variables are handled in the docker setup so that we only need to specify them in the `.env` file and don't have to worry about keeping `.env`/`.python_env` in sync. Also contains a minor clean up of the Dockerfile which removes mariadb in favor of postgres and so also changes postgres to the default in `.env` 😈 

Docs updated to match nicer behavior.


docs/installation.md
```diff
@@ -160,17 +156,17 @@ dropdb --if-exists cfgov && dropuser --if-exists cfpb
...
-specifies SQLite as the db:
+If you absolutely need to use SQLite, you'll need to update your `.env` file
+to comment out the line that specifies Postgres as the db:
 
 ```bash
-# export DATABASE_URL=sqlite:///db.sqlite3
+# export DATABASE_URL=postgres://cfpb@localhost/cfgov
 ```
 
 And then uncomment the line that tells Django to use Postgres:
```
anselmbradford: `use Postgres` > `use SQLite`

chosak:
Does the [README](https://github.com/cfpb/cfgov-refresh/blob/master/README.md) also need to be updated as it assumes the default behavior will use SQLite?

anselmbradford:
I don't have the docker setup to check, but the `.env` bits LGTM 👍

.env_SAMPLE
```diff
@@ -175,4 +177,6 @@ export WATCHMAN_TOKENS=''
 # Project configuration - set up working state.
 ###############################################
 
-source activate-virtualenv.sh
+if [ ! -z $(command -v workon)]; then
```
Scotchester: This needs a space before the `]` to work, otherwise you get this error:

`-bash: [: missing `]'`

Scotchester:
Hmm... maybe I take that back. It seems to be causing an issue in the Docker setup:

```
python_1         | Performing system checks...
python_1         |
python_1         | System check identified no issues (0 silenced).
python_1         | Unhandled exception in thread started by <function wrapper at 0x7fadd84c21b8>
python_1         | Traceback (most recent call last):
python_1         |   File "/usr/lib64/python2.7/site-packages/django/utils/autoreload.py", line 229, in wrapper
python_1         |     fn(*args, **kwargs)
python_1         |   File "/usr/lib64/python2.7/site-packages/django/core/management/commands/runserver.py", line 116, in inner_run
python_1         |     self.check_migrations()
python_1         |   File "/usr/lib64/python2.7/site-packages/django/core/management/commands/runserver.py", line 168, in check_migrations
python_1         |     executor = MigrationExecutor(connections[DEFAULT_DB_ALIAS])
python_1         |   File "/usr/lib64/python2.7/site-packages/django/db/migrations/executor.py", line 19, in __init__
python_1         |     self.loader = MigrationLoader(self.connection)
python_1         |   File "/usr/lib64/python2.7/site-packages/django/db/migrations/loader.py", line 47, in __init__
python_1         |     self.build_graph()
python_1         |   File "/usr/lib64/python2.7/site-packages/django/db/migrations/loader.py", line 191, in build_graph
python_1         |     self.applied_migrations = recorder.applied_migrations()
python_1         |   File "/usr/lib64/python2.7/site-packages/django/db/migrations/recorder.py", line 59, in applied_migrations
python_1         |     self.ensure_schema()
python_1         |   File "/usr/lib64/python2.7/site-packages/django/db/migrations/recorder.py", line 49, in ensure_schema
python_1         |     if self.Migration._meta.db_table in self.connection.introspection.table_names(self.connection.cursor()):
python_1         |   File "/usr/lib64/python2.7/site-packages/django/db/backends/base/base.py", line 162, in cursor
python_1         |     cursor = self.make_debug_cursor(self._cursor())
python_1         |   File "/usr/lib64/python2.7/site-packages/django/db/backends/base/base.py", line 135, in _cursor
python_1         |     self.ensure_connection()
python_1         |   File "/usr/lib64/python2.7/site-packages/django/db/backends/base/base.py", line 130, in ensure_connection
python_1         |     self.connect()
python_1         |   File "/usr/lib64/python2.7/site-packages/django/db/utils.py", line 98, in __exit__
python_1         |     six.reraise(dj_exc_type, dj_exc_value, traceback)
python_1         |   File "/usr/lib64/python2.7/site-packages/django/db/backends/base/base.py", line 130, in ensure_connection
python_1         |     self.connect()
python_1         |   File "/usr/lib64/python2.7/site-packages/django/db/backends/base/base.py", line 119, in connect
python_1         |     self.connection = self.get_new_connection(conn_params)
python_1         |   File "/usr/lib64/python2.7/site-packages/django/db/backends/postgresql_psycopg2/base.py", line 176, in get_new_connection
python_1         |     connection = Database.connect(**conn_params)
python_1         |   File "/usr/lib64/python2.7/site-packages/psycopg2/__init__.py", line 130, in connect
python_1         |     conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
python_1         | django.db.utils.OperationalError: could not connect to server: Connection refused
python_1         | 	Is the server running on host "localhost" (127.0.0.1) and accepting
python_1         | 	TCP/IP connections on port 5432?
python_1         | could not connect to server: Cannot assign requested address
python_1         | 	Is the server running on host "localhost" (::1) and accepting
python_1         | 	TCP/IP connections on port 5432?
```

Looks like the Python container can't talk to the Postgres container anymore. Looking at the change set, I can't see why that would be the case, but I'm getting that output on this branch, while it works fine on `master`.

.env_SAMPLE
```diff
@@ -48,8 +48,10 @@ export ALLOW_ADMIN_URL=True
 # See https://github.com/kennethreitz/dj-database-url for URL formatting.
 #########################################################################
 
-export DATABASE_URL=sqlite:///db.sqlite3
-#export DATABASE_URL=postgres://cfpb@localhost/cfgov
+# export DATABASE_URL=sqlite:///db.sqlite3
+export DATABASE_URL=postgres://cfpb@localhost/cfgov
```
chosak: Unifying this environment variable is problematic because if not using Docker, we want the host value here to be `localhost` if running Postgres using something like Homebrew, but in Docker it needs to refer to the Postgres container, which is `postgres`.

.env_SAMPLE
```diff
@@ -177,7 +178,8 @@ export WATCHMAN_TOKENS=''
 # Project configuration - set up working state.
 ###############################################
 
-if [ ! -z $(command -v workon) ]; then
+# Local envvar overrides and virtualenv hook
+if [ $(uname -s) = "Darwin" ]; then
```
Scotchester: Could you explain this change just a little more for me? This will only be true on macOS, right? What if someone wants to develop on a Linux system? Why was the previous condition less good?

wpears:
😢 Confirmed with regeneration of `.env` from `.env_SAMPLE`.

Easy fix, just missed an export of `PGPASSWORD`

Scotchester:
Weird. I would've thought that it being in the URL would suffice.

Also, the comment "Required for Postgres dbshell in Django < 1.9" misled me to believe that the `PGPASSWORD` value was unrelated. As one final request, perhaps we can update that to be more clear about where that is used?

wpears:
@Scotchester kk, I removed the `cfpb:cfpb` bit and the comment about dbshell, which I think wasn't really helpful. Should be clear now that both envvars are required and what they each accomplish.

Scotchester:
Thanks, @wpears!

wpears:
@rossk @chosak how's this look now? Probably want to separate this from the dockerfile update in case that one eventually gets rejected

PR #4395, opened 2018-07-31
Regulations 3000: Landing page fields

willbarton:
This PR adds additional content fields to the Regulations landing page, including:

- `header`, with a hero molecule
- `content` with a custom full-width text block (`RegulationsFullWidthText`) that can contain:
    - A content block
    - A reusable text block
    - A custom block (`RegulationsList`) for listing available regulations in a table.

## Testing

1. Pull down this branch and run migrations, `./cfgov/manage.py migrate regulations3k`
2. Create or visit your local "Regulation landing page"
3. Add a hero to the header, and add a full width text block with content and a regulations list.
4. Preview or publish and visit the page to view!

## Screenshots

With these changes it is possible to construct a page that looks like this:

image


cfgov/regulations3k/jinja2/regulations3k/regulations-listing.html
```diff
@@ -0,0 +1,51 @@
...
+
+   heading:        The heading of the regulations list.
+
+   more_regs_url:  A string for the URL of the more regulations link.
+
+   more_regs_text: A string for the text of the more regulations link.
+
+   ========================================================================== #}
+{% if regulations.exists() %}
+<table class="o-table" data-bound="true">
```
niqjohnson: Doesn't have to hold up this PR, but I wonder if a using an unordered list instead of a table would make more sense here. On one hand, using an `o-table` is nice since it's a standard component and we don't have to make any front-end customizations to get it looking right. On the other, since we're now down to just a single column that has links to each reg (and eventually links to future amendments), I think an unordered list would be more semantic (though it would obviously require some custom styles to match the design).

willbarton: We certainly could do — I leave that up to you and @contolini.

willbarton: But then it might need to be titled "Unordered list of contents" 😛

contolini: @niqjohnson I think you're right that `<ul>` makes more sense. I bet we can get a list to mimic a table's styles pretty easily. To avoid holding up this PR let's open a separate issue to do it.

niqjohnson: @contolini, converting the table to an unordered list in a separate PR works for me. Thanks!

PR #4399, opened 2018-08-02
Add additional unit tests for fragment caching

richaagarwal:
This commit adds additional unit tests for the Jinja {% cache %} template tag, and its support for multiple caches in #4384.

### DIFF
--- a/cfgov/v1/tests/test_fragment_cache_extension.py
+++ b/cfgov/v1/tests/test_fragment_cache_extension.py
@@ -1,5 +1,6 @@
-from django.core.cache import caches
-from django.test import Client, TestCase
+from django.core.cache import cache, caches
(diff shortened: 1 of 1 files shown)


richaagarwal:
@chosak Sorry, not sure you saw in this comment https://github.com/cfpb/cfgov-refresh/pull/4384#issuecomment-409798774 that I needed to remove these since I couldn't get them to pass.  So you might want to restore your branch.  Would you like to pair on getting these working when you have time?

chosak:
@richaagarwal oh, I see. I misinterpreted this as merging to master and didn't look closely enough to see that it was merging into your branch. I've fixed the tests (I don't see how these ever worked before, actually!) and opened a new PR against master in #4411.

PR #4401, opened 2018-08-02
Regulations 3000: Enforce using the default block renderer in regdown/resolver tests

willbarton:
There doesn't seem to be a clean way to clear markdown extension config options in between tests, so this change is to always use the default block renderer (instead of another custom one) in all tests when testing things that are not the block renderer support.

This should stop some inconsistent test failures we were seeing in the resolver tests (the ultimate cause of those failures is still something I'm troubleshooting, but new tests should be written to test for that).


willbarton:
> This fixes python 2 tests but now python 3 fails with an odd ImportError: bad magic number in 'sheerlike.management': b'\x03\xf3\r\n'

@higs4281 I think you must have some Python 2-compiled .pyc files lying around from the sheerlike removal.

PR #4408, opened 2018-08-03
Regs3k fix svg namespace parsing

willbarton:
This PR fixes Regulations 3000 problems with parsing the svgs used in the inline interpretations expandable. These problems included styling errors, rendering errors, and 500 errors being generated by the Python-Markdown library failing to strip namespaces that were being introduced by ElementTree when including our inline interpretation template in the rendered markdown.


contolini:
@niqjohnson @ajbush Should the interp content (i.e. the `1. Application for credit` paragraph above) be indented? I think the original designs called for it to be flush with the expandable header?

niqjohnson:
Looks like inline interpretation paragraphs were indented before all the expandables updates, so this indentation isn't a regression or anything. Ideally, yes, the paragraphs would be flush left, but I'm fine if we want to make that a separate PR and keep this one to just the SVG changes.

contolini:
Alrighty I modified the indentation code, the interps now look like this: 

screen shot 2018-08-06 at 12 52 35 pm

PR #4409, opened 2018-08-03
Update webpack-config.js

anselmbradford:
atomic-component has been moved into Capital Framework under cf-atomic-component, so we no longer have to filter it out.

## Changes

- Changes exclusion regex to search for modules with `cf-` and removes search for `atomic-component`.

## Testing


anselmbradford:
Closing in favor of a non-fork PR

PR #4410, opened 2018-08-03
Adjusts webpack config regex to remove atomic-component

anselmbradford:
atomic-component has been moved into Capital Framework under cf-atomic-component, so we no longer have to filter it out.

## Changes

- Changes exclusion regex to search for modules with `cf-` or `cfpb-` and removes search for `atomic-component`.

## Testing

2. `gulp test:acceptance --headless` should pass.


cfgov/unprocessed/apps/analytics-gtm/webpack-config.js
```diff
@@ -29,9 +29,10 @@ const COMMON_MODULE_CONFIG = {
   rules: [ {
     test: /\.js$/,
 
-    /* The below regex will capture all node modules that start with `cf`
-      or atomic-component. Regex test: https://regex101.com/r/zizz3V/1/. */
-    exclude: /node_modules\/(?:cf.+|atomic-component)/,
+    /* The below regex will capture all node modules
+       that start with `cf-` or `cfpb-`.
+       Regex test: https://regex101.com/r/zizz3V/5. */
```
Scotchester: Mind removing the dot from the end of the URL? If you include it, it resolves to the first version of the regex.

anselmbradford: Fixed

PR #4413, opened 2018-08-06
Fix `parse_links` middleware so that Python 3 tests pass

willbarton:
The `parse_links` updates introduced some test failures when running the tests on Python 3 because of `bytes`/`str` differences in Python 2 and 3 (Django's `response.content` is `bytes` on Python 3). Because `parse_links` is done in middleware now this caused unrelated tests that use the Django test client to fail on Python 3 (like `regulations3k`, where @higs4281 and I have been trying to maintain Python 3-compatibility). This PR fixes that by decoding/encoding the `response.content` using the `response.charset`.

While I was in `cfgov/core`, I also updated the rest of the tests that were failing on Python 3 so that they will pass too, to get one more corner of cfgov-refresh ready for the inevitable.

## Checklist

### DIFF
--- a/cfgov/core/middleware.py
+++ b/cfgov/core/middleware.py
@@ -26,8 +26,12 @@ def should_parse_links(request_path, content_type):
     return True
 
 
-def parse_links(html):
+def parse_links(html, encoding=None):
     """Process all links in given html and replace them if markup is added."""
+    if encoding is None:
+        encoding = settings.DEFAULT_CHARSET
+    html = html.decode(encoding)
+
     soup = BeautifulSoup(html, 'html.parser')
     link_tags = get_link_tags(soup)
     for tag in link_tags:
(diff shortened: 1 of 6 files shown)


chosak:
@willbarton can you share the tox commands needed to run the tests you've changed against Python 3?

willbarton:
`tox -e unittest-py36-dj18-wag113-fast core`

richaagarwal:
@willbarton Thanks for this! Independent of this, would it make sense to add eregs to the blacklist so that `parse_links` doesn't touch it?

willbarton:
@richaagarwal it probably would, actually. Hopefully this doesn't break anything and it gets Python 3-support out of the way though.

willbarton:
@chosak We could discuss it. There're no changes to that blacklist here though.

PR #4417, opened 2018-08-06
Consolidate Jinja2 environment

chosak:
This change consolidates the definition of the Jinja2 template environment into a single module at `v1.jinja2_environment`. Previously, the environment was defined in two places: `sheerlike.SheerlikeEnvironment` and `v1.environment`. This split of logic made it hard to debug what was being placed in the environment, and the definitions in the app roots complicated imports at times.

See `cfgov.settings.base.TEMPLATES` for some new comments about how our template engines are defined.

In addition, this change reorganizes and refactors a bit all of the global functions and filters that we place into the Jinja2 environment. It defines a new `v1.jinja2tags.datetimes` to collect date/time-related functions that are used, and also moves the caching logic under `v1.jinja2tags.fragment_cache`. This is done in attempt to be more consistent about organizing all of our Jinja2 tags into `jinja2tags` modules within each app.

Various unit tests have been reorganized and added to ensure that our Jinja2 environment is well-tested. The existing behavior of the environment to support relative template paths has also been documented and tested; see `v1.jinja2_environment.RelativeTemplatePathEnvironment`.

Note that this change doesn't add any new functionality and should not change any functionality, but only reorganizes and removes unused code. Specifically these Jinja2 globals have been removed as they are no longer used anywhere in the codebase:

- `cfgovpage_objects`
- `get_latest_activities`
- `global_dict`
- `stripaccents`

## Notes

- This reorganization will help with the upgrade to Django 1.11, as some of the imports currently in `v1/__init__.py` are no longer permitted in that version. This refactor simplifies `v1/__init__.py` to conform with Django expectations about not importing models.

## Checklist


cfgov/v1/jinja2tags/__init__.py
```diff
@@ -0,0 +1,112 @@
+from six.moves import html_parser as HTMLParser
+
+from django.utils.module_loading import import_string
+
+from jinja2 import Markup, contextfunction
+from jinja2.ext import Extension
+
+from .datetimes import DatetimesExtension
+from .fragment_cache import FragmentCacheExtension
```
richaagarwal: I'm curious about the relative path imports here.  I think we're fairly inconsistent about this as-is, but had been under the impression we were moving towards explicit imports everywhere.  I have a slight preference for `from v1.jinja2tags.fragment_cache` even if it feels slightly redundant since we're inside that directory.

willbarton: I would definitely prefer absolute imports

PR #4427, opened 2018-08-09
Fix v1.jinja2tags.datetimes for running tests on Python 3

willbarton:
This PR adds an import for [`six.string_types` as `basestring`](https://pythonhosted.org/six/#six.string_types) to our `v1.jinja2tags.datetimes` module to ensure that its use of `basestring` doesn't cause errors when running tests that use the Django test client on Python 3.

### DIFF
--- a/cfgov/v1/jinja2tags/datetimes.py
+++ b/cfgov/v1/jinja2tags/datetimes.py
@@ -1,4 +1,5 @@
 import datetime
+from six import string_types as basestring
 
 from django.utils.timezone import template_localtime
 


higs4281:
Not sure I I'm out of sync somehow, but I also encounter string type clash when I run Python3 tests against regulations3k. If I change this line to look for a bytestring instead of unicode, tests pass.

https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/regulations3k/tests/test_models.py#L391

willbarton:
@higs4281 Yes, I added a quiet fix for that in #4425.

higs4281:
@willbarton I thought that rang a  bell.

PR #4444, opened 2018-08-16
Support running unit tests against Django 1.11

chosak:
Currently when trying to run Python unit tests against Django 1.11, for example with this command:

`tox -r -e unittest-py27-dj111-wag113-slow`

the test runner command fails immediately, before any tests are run. This change fixes a few things that are incompatible with Django 1.11, which allows the test command to run, and individual test failures to be reported. This does not actually make the tests work, but at least lets them run.

All tests continue to pass with Django 1.8.

Specific changes are:

- As of 1.9, the use of `settings.MIGRATION_MODULES` to disable migrations has changed, and so our logic in cfgov.settings.test_nomigrations must as well. See Django docs at https://docs.djangoproject.com/en/2.1/releases/1.9/#migrations.
- As of 1.10, support for string view paths to `url()` has been removed, and the module itself must be passed. See Django docs at https://docs.djangoproject.com/en/2.1/releases/1.10/#features-removed-in-1-10.
- The version of django-extensions we were using, 1.5.5, is not compatible with Django 1.11 due to the existence of a migration that tries to import from South. Support for Django 1.11 is added to django-extensions in version 1.7.8. This PR bumps the version to the latest one, 2.1.0, which will include Django 2.0 support in future.

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -739,3 +739,8 @@ PARSE_LINKS_BLACKLIST = [
     '/admin/',
     '/django-admin/'
 ]
+
+# Required by django-extensions to determine the execution directory used by
+# scripts executed with the "runscript" management command.
+# See https://django-extensions.readthedocs.io/en/latest/runscript.html.
+BASE_DIR = 'scripts'
--- a/cfgov/cfgov/settings/test_nomigrations.py
+++ b/cfgov/cfgov/settings/test_nomigrations.py
@@ -1,6 +1,13 @@
+from collections import defaultdict
+
+import django
+
 from core.utils import NoMigrations
 
 from .test import *
 
 
-MIGRATION_MODULES = NoMigrations()
+if django.VERSION[:2] < (1, 9):
+    MIGRATION_MODULES = NoMigrations()
+else:
+    MIGRATION_MODULES = defaultdict(None)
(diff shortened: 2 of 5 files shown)


chosak:
Additional dependency changes for Django 1.11 compatibility:

- django-csp 3.1 -> 3.4
- django-haystack 2.6.0 -> 2.7.0
- django-overextends 0.4.1 -> 0.4.3

Regarding django-haystack, @higs4281 we wanted to make sure that this upgrade wouldn't break any Elasticsearch usage in cfgov-refresh. To test, we used a production dump with a local ES server, and visited a few Ask pages. Specifically, we confirmed that Ask category pages properly query the ES server to get back a list of questions, and that doing a text search uses ES to do autocomplete. We did this for both the English and Spanish pages. Are there any other ES-related tests you can think of that we should do here?

higs4281:
@chosak `regulations3k` and `comparisontool` have ES indexes.

Running `update_index` for each with a prod dump should build the indexes.

Then test
- <http://localhost:8000/policy-compliance/rulemaking/regulations/search-regulations/results/?q=california&regs=1024>

If you have cost-comparison installed, you can test the search form when clicking "Get started" at: 
- <http://localhost:8000/paying-for-college/compare-financial-aid-and-college-cost/>

Or hit the comparison search api at
- https://www.consumerfinance.gov/paying-for-college/compare-financial-aid-and-college-cost/api/search-schools.json?q=boise

chosak:
@higs4281 thanks for the suggestions! I can confirm that all of those searches work.

PR #4445, opened 2018-08-16
Remove missing migrations script and unit test (post Django 1.11)

chosak:
~~This unit test checks whether there are any missing Django migrations. It duplicates logic that can be run independently using the [cfgov/core/scripts/missing_migrations.py](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/core/scripts/missing_migrations.py) script, via `./manage.py
runscript missing_migrations`. This is already run automatically in Travis via `tox -e missing-migrations`.~~

~~This unit test is the slowest of all Python unit tests (it takes longer than a second at times for me, due to invoking an external script), and given that its logic exists elsewhere and is automatically run on every PR, I propose that we remove this unit test.~~

See discussion and commits below; this PR now removes entirely our custom script that checks for missing migrations, and the unit test that invokes it. This should not be merged until we are on Django 1.11.

## Checklist


Scotchester:
Why don't we go the other way and remove the explicit invocation in Travis? If we take this out, then people who run the full suite of unit tests locally won't be notified if they're missing a migration.

chosak:
@schbetsy and I spent some time revisiting this as part of the ongoing upgrade to Django 1.11. Unfortunately the method for checking for missing migrations differs between 1.8 and 1.11 which complicates all of this code. I'm marking this PR as hold until such time as we are solely on 1.11 and can simplify again.

@Scotchester we considered that as an alternate path. There are definitely pros and cons to each approach.

If we include it as a unit test, it is nice that it gets run automatically as part of regular developer iterations. On the other hand, it then slows down the tests each time they are run, and makes it harder to work iteratively, say if you're changes to a model but don't want to make the migration until you're finished. Additionally, one could argue that checking for missing migrations doesn't belong as a _test_: it's not testing our code, but rather running a Django command to perform a logically independent function, more akin to linting.

If we include it on Travis, then we still ensure that it gets run before merges happen. Granted, developers might forget to run this locally, and have to correct their PRs, but one could say the same thing about the linter, or the acceptance tests.

Once we move to Django 1.11, we should be able to greatly simplify all of this code, as running `manage.py makemigrations --check` will return a non-zero exit code if there are migrations to be made, which is what we want and was difficult in Django 1.8. In this world, we can [modify our `tox.ini`](https://github.com/cfpb/cfgov-refresh/blob/master/tox.ini#L183) to run the command directly, this removing the need for a custom Python script (or test for that script).

Given that, I lean slightly towards holding this PR until after Django 1.11, removing the script and test, and just relying on tox/Travis to run the Django `makemigrations --check` on PRs (while still giving devs the ability to run locally with `tox -e missing-migrations` as they can today).

That being said, if you or others feel strongly the other way, we can definitely consider alternatives that are closer to the setup we have today.

Scotchester:
@chosak Thanks for the thorough answer! That all makes sense to me.

chosak:
@schbetsy could I please request another review (and, if it looks good, feel free to merge)? I've made some minor additional changes in the two most recent commits post-rebase.

PR #4450, opened 2018-08-20
Use boto3 instead of boto for Django 1.11 compatibility

chosak:
These set of changes change the AWS S3 package used by this project to boto3 from boto. See this post discussing differences:

https://boto3.readthedocs.io/en/latest/guide/migration.html

Generally, boto3 is a more modern library which is better maintained.

Direct uses of boto have been replaced with boto3, and relatedly this change modifies the django-storages backend to use its boto3 backend. The django-storages dependency has been bumped in version from 1.1.8 to 1.5.0; unfortunately we cannot yet use the most recent version, see
`requirements/libraries.txt` for an explanation.

Some of the Django settings have been simplified or ordered alphabetically related to this change. Specifically, we previously used an `AWS_S3_ROOT` environment variable to set an `AWS_S3_ROOT` setting that was used to instantiate a custom storages backend. This is no longer necessary as the more recent versions of django-storages directly support use of `AWS_LOCATION` to set the root path. As this is always "f" on all systems, I've hardcoded it here.

These changes are necessary to add Django 1.11 compatibility (mainly because django-storages 1.1.8 didn't support Django 1.11) but also get us more up-to-date in our use of S3.

## Testing

Unit tests should cover use of S3 helper methods.

You can also run a local server and enable S3 through environment variables (`S3_ENABLED=True`, set `AWS_S3_ACCESS_KEY_ID`, `AWS_S3_SECRET_ACCESS_KEY`, `AWS_STORAGE_BUCKET_NAME`, `AWS_S3_URL`). You should be able to browse in the Wagtail admin under the Documents section and see that file URLs are properly pointing to S3, and successfully download those files.

## Notes

The boto3 backend adds some logging statements under `boto` and `boto3` loggers at level `INFO` when doing things like opening a connection to S3. Due to the way we have logging set up in this project, you'll see those in your console when running locally, and they'll get written to server logs. We could suppress these if we wanted to.

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -382,18 +382,18 @@ ELASTICSEARCH_INDEX_SETTINGS = {
 ELASTICSEARCH_DEFAULT_ANALYZER = 'snowball'
 
 # S3 Configuration
+# https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#settings
+AWS_LOCATION = 'f'  # A path prefix that will be prepended to all uploads
 AWS_QUERYSTRING_AUTH = False  # do not add auth-related query params to URL
-AWS_S3_CALLING_FORMAT = 'boto.s3.connection.OrdinaryCallingFormat'
-AWS_S3_ROOT = os.environ.get('AWS_S3_ROOT', 'f')
+AWS_S3_FILE_OVERWRITE = False
 AWS_S3_SECURE_URLS = True  # True = use https; False = use http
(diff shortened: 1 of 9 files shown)


chosak:
@higs4281 thanks for catching, and I'm sorry for missing those! It's too bad those other boto references weren't caught by the unit tests. I've pushed another commit in 339ac15 that tries to fix those references; would you mind taking a look when you get a chance? I removed the JSON code as it doesn't seem to be used anywhere, but I can restore if you think it's important.

The requirement in college-costs will be removed by [college-costs#327](https://github.com/cfpb/college-costs/pull/327); thanks also for flagging that.

PR #4461, opened 2018-08-28
Adds django-mptt to libraries.txt

wpears:
Allows https://github.com/cfpb/teachers-digital-platform to be included in cfgov-refresh without a separate `pip install`.


requirements/libraries.txt
```diff
@@ -10,6 +10,7 @@ django-extensions==2.1.0
 django-flags==3.0.0
 django-haystack==2.7.0
 django-localflavor==1.1
+django-mptt==0.9.0
```
chosak: Would you mind adding a comment here indicating that this is required by the teachers-digital-platform package?

PR #4469, opened 2018-08-29
Alphabetize chart colors in Wagtail UI

anselmbradford:
## Changes

- Alphabetize color selection values in ChartBlock.

## Testing

1. Restart the local server.
2. Log into http://localhost:8000/admin/
3. Create a new Browse Page.
4. Add a Chart block to the page.
5. Set the required fields (if you try to save the page, it will notify you which are required).
6. Set the data source as `consumer-credit-trends/auto-loans/num_data_AUT.csv`.
7. See that color selection drop-down is alphabetize.
8. Preview the page and see that the chart renders.

## Screenshots

screen shot 2018-08-29 at 10 04 34 am


marteki:
Everything works to load and preview the page, but the dropdown isn't alphabetized. 😢 

screen shot 2018-08-29 at 11 59 19 am

Can't knowledgeably guess what might be going on, because I'm not well-versed in Python.

PR #4477, opened 2018-08-30
Properly disable migrations in Django 1.11

chosak:
PR #4444 (014df5a7) attempted to add the ability to run Python tests against Django 1.11, but did not properly disable migrations when running with `settings.test_nomigrations`.

This commit fixes testing without migrations (e.g. with `tox -e unittest-py27-dj111-wag113-fast`) in Django 1.11.

To test this change, you can run that command and notice how much faster the tests run, because the migrations aren't being run.

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/test_nomigrations.py
+++ b/cfgov/cfgov/settings/test_nomigrations.py
@@ -1,13 +1,6 @@
-from collections import defaultdict
-
-import django
-
 from core.utils import NoMigrations
 
 from .test import *
 
 
-if django.VERSION[:2] < (1, 9):
-    MIGRATION_MODULES = NoMigrations()
-else:
-    MIGRATION_MODULES = defaultdict(None)
+MIGRATION_MODULES = NoMigrations()
--- a/cfgov/core/tests/test_utils.py
+++ b/cfgov/core/tests/test_utils.py
@@ -1,5 +1,6 @@
 import unittest
 
+import django
 from django.test import TestCase
 
 from core.utils import (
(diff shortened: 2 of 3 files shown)


schbetsy:
When I run `tox -e unittest-py27-dj111-wag113-fas`, I get this as part of the output:
```
Creating test database for alias 'default'...
applying migration wagtail.wagtailcore.migrations.0002_initial_data
applying migration wagtail.wagtailcore.migrations.0025_collection_initial_data
```

Those lines aren't in the output on `master`. What does it mean?

--- 
Also, looks like this change shaves 30 seconds off the time it takes to run the tests. Nice work.

| | `master` | `nomigrations` |
|---|---|----|
| real  | 	2m2.190s | 1m29.533s |
| user |	1m56.133s | 1m25.326s |
| sys |	0m3.818s | 0m3.193s |

chosak:
@schbetsy those two migrations are Wagtail data migrations that don't change the database schema but do introduce data into the database. [The first one](https://github.com/wagtail/wagtail/blob/v1.13.4/wagtail/wagtailcore/migrations/0002_initial_data.py) creates a root Wagtail Page and a default Site along with various permissions objects. [The second one](https://github.com/wagtail/wagtail/blob/v1.13.4/wagtail/wagtailcore/migrations/0025_collection_initial_data.py) creates a root Collection object.

We have some unit tests that depend on these things existing; for example, any test that tries to create a Wagtail page will need there to be a Site and root Page in the database.

If we want to run the tests but skip migrations for speed, we need to make sure that data migrations like these are still run in order to ensure that the necessary data exists. The code that does that is [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/test.py#L86) in our custom test runner class.

The reason these messages appear in this branch but not in master is that in master all migrations are (erroneously) applied even if running with `settings.test_nomigrations` (`tox -e fast`). So when [this line](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/test.py#L112) is hit in the test runner, there's nothing to do because the migrations have already been run. It's only if they haven't already been run - because they were correctly skipped due to the change in this PR - that the test runner manually runs them and, as part of doing so, [prints that message](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/test.py#L113) to the console.

Thanks for taking those timing measurements!

PR #4478, opened 2018-08-30
Bump cfpb-chart-builder to 3.5.0

anselmbradford:
Brings in latest chart builder, which changes default behavior of y-axis values.

## Changes

- Bump `cfpb-chart-builder` to `3.5.0`.

## Testing

1. Run `./frontend.sh`
2. Log into local wagtail.
3. Create a new Browse page.
4. Add a ChartBlock.
5. Add required fields (set `consumer-credit-trends/auto-loans/inq_data_AUT.csv` as data source).
6. Save and preview and see that the y-axis values are integers. (If y-axis values are still millions or billions, inspect the page and reload to not use the cache.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3021,9 +3021,9 @@
       }
     },
     "cfpb-chart-builder": {
-      "version": "3.4.0",
-      "resolved": "https://registry.npmjs.org/cfpb-chart-builder/-/cfpb-chart-builder-3.4.0.tgz",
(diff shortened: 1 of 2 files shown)


anselmbradford:
> (I think the incorrect Y-axis label shown above would be fixed when cfpb/cfpb-chart-builder#129 is merged, a new cfpb-chart-builder release is created, and that release version is again updated here.)

You can set the y-axis label in Wagtail for the demos:
screen shot 2018-08-30 at 12 51 48 pm

marteki:
@anselmbradford I tried that on my local, and it didn't work! I think because of what was in this branch vs. what's on `master` from other PRs. Not concerned about it, until it crops up again (but not expecting it to).

PR #4482, opened 2018-08-31
First stab at full build in docker

### LINKED ISSUE #4462: Add git into python image
This allows us to include projects that are specified [like ccdb](https://github.com/cfpb/cfgov-refresh/blob/master/requirements/optional-public.txt#L7)

### LINKED ISSUE #4468: Incorporate front-end build tools into the docker image
In order to really decouple running cfgov and satellite apps from per-machine dependencies we need to provide all build dependencies in the image. This will allow users to run cfgov in total with the only hard dependency being docker.

wpears:
Closes #4468 
Closes #4462 

This is my first stab at running the full build in the backend. From a fresh repo, with the `cfgov-refresh_python` image removed, you should be able to simply `docker-compose up` to end up with working dockerized cfgov. (no front-end stuff, the .env_SAMPLE is copied over by default). The only thing you'd need to do afterwards to check it out at `localhost:8000`, is run `./shell.sh` and `./refresh-data.sh` with a dump once inside. This process could be done automatically (say, based on the existence of a `data.tar.gz` file or something) but I've elided that for now.

Satellite apps are currently unsupported as there are some decisions to be made around what is optional/what should be baked into the image that I didn't want to cloud this PR. That is, should `optional-public` be installed at build time? The argument **for** is to move all the painful installation stuff into one place and to then have everything available by default, the argument **against** is to limit the pain of that first image build. I'm more in the **for** camp, but reasonable people may disagree (and please do with comments on this PR).

If installation of the wheels **is** done at image build time, then, on container startup, we'd walk develop-apps, `pip uninstall` as needed, and `pip install -e` each app there. This preserves the ability to locally edit satellites without any mucking around inside the container, while also not affecting anyone who isn't editing satellites.

I'll be out for two weeks starting tomorrow 😅 , but I'll likely peep in on this issue to address questions/comments in the evenings over the next fews days.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,7 +3,6 @@ develop-apps
 collectstatic
 *.sql
 .github
-gulp
 node_modules
 docs
 .vendor
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,11 @@
 FROM centos:7
 RUN yum -y install https://download.postgresql.org/pub/repos/yum/10/redhat/rhel-7-x86_64/pgdg-centos10-10-2.noarch.rpm && \
-    yum -y install which gcc gcc-c++ kernel-devel mailcap make postgresql10 postgresql10-devel python-devel && \
+    yum -y install which gcc gcc-c++ git kernel-devel mailcap make postgresql10 postgresql10-devel python-devel && \
+    curl --silent --location https://rpm.nodesource.com/setup_10.x | bash - && \
+    yum -y install nodejs && \
     yum clean all
+WORKDIR /src
+ADD . /src
 ADD  https://bootstrap.pypa.io/get-pip.py /src/get-pip.py
-COPY requirements /src/requirements
-COPY extend-environment.sh /etc/profile.d/extend-environment.sh
-RUN python /src/get-pip.py && pip install -r /src/requirements/local.txt
+RUN python /src/get-pip.py && \
+  /src/setup.sh docker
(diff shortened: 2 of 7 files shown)


rosskarchner:
(that was after a complete removal of all containers and images)

wpears:
@rosskarchner, should be fixed.

rosskarchner:
Now it comes up with: 

> settings.DATABASES is improperly configured. Please supply the ENGINE value. Check settings documentation for more details.

Do I need to set the database in my .env file?

chosak:
@rosskarchner yes, see https://github.com/cfpb/cfgov-refresh/blob/master/.env_SAMPLE#L53.

chosak:
> we'd walk develop-apps, pip uninstall as needed, and pip install -e each app there

This feels a little bit too magical to me. I'm also in the **for** camp above and would vote to make `optional-public.txt` non-optional by including it in `requirements/local.txt` (and possibly also [in tox.ini](https://github.com/cfpb/cfgov-refresh/blob/master/tox.ini#L122)). If and when developers want to develop the apps locally, they could manually `pip install -e` as needed, the same as they would any other Python package.

rosskarchner:
@wpears good branch 11/10

willbarton:
@wpears is this PR deprecated by #5051?

wpears:
@willbarton yeah, I think in most respects it is. Closing!

PR #4489, opened 2018-09-06
Add LineChartIndex chart type

anselmbradford:
## Additions

- Adds LineChartIndex as option in Wagtail ChartBlock "Type" dropdown.

## Changes

- Adds additional text in Wagtail ChartBlock "Type" dropdown that indicates y-axis values for each chart type.
- Bumps `cfpb-chart-builder` to `5.6.0`.

## Testing

1. Run `./frontend.sh`
1. Run `python cfgov/manage.py migrate`
2. Restart local server.
3. Log into http://localhost:8000/admin/
4. Create a browse page and add a ChartBlock and required fields.
5. Set the type as "Line" and set data source as `consumer-credit-trends/auto-loans/num_data_AUT.csv` and preview to see that y-axis is millions/billions.
6. Change the type to "Line-Index" and set data source as `consumer-credit-trends/auto-loans/inq_data_AUT.csv` and preview to see that y-axis is integers and the label is "Index (January 2009 = 100)".

## Screenshots

screen shot 2018-09-06 at 5 15 45 pm

## Notes

- With the Line-Index type set, the y-axis label is hardcoded and can not be overridden from the custom y-axis field.


cfgov/v1/atomic_elements/organisms.py
```diff
@@ -900,7 +904,10 @@ class ChartBlock(blocks.StructBlock):
                   '"Data from the last six months are not final."')
     y_axis_label = blocks.CharBlock(
         required=False,
-        help_text='Custom y-axis label')
+        help_text='Custom y-axis label. '
+                  'NOTE: Line-Index chart y-axis '
```
marteki: Huh. This new note is not accurate.

screen shot 2018-09-07 at 1 49 19 pm

screen shot 2018-09-07 at 1 49 37 pm

anselmbradford: Uhhh that's not possible. My mistake, I see the issue. You need to run `./frontend.sh` to bump chart builder. Prior sure what you're seeing is the prior behavior before the bump. Added to testing instructions.

marteki: ... It also would have helped if I'd changed his graph from "Line" type to "Line-Index" type.

🤦‍♀️

PR #4495, opened 2018-09-07
Fix FLC height adjustment when interacting with multiselect filters

Scotchester:
As reported in https://GHE/CFGOV/platform/issues/2981, the height of a Filterable List Controls expandable is not being correctly adjusted when interacting with the multiselect filters inside. This PR takes care of that.

## Additions

- Added a new event dispatch when selections on a multiselect are updated, which I didn't end up using, but it might still be useful in the future.
- Added `resolve: { symlinks: false }` in `webpack-config.js` files, which I needed to test with an `npm link`ed cf-expandables. This PR no longer needs any changes in cf-expandables, but this will be useful to anyone in the future who wants to test local updates to CF components within cfgov-refresh.

## Changes

- Bind the instance of the FLC expandable's `transition` to the `expand` calls within `window.setTimeout`, because if not, `this` within the `expand` method ends up being `window`

## Testing

1. Pull branch
1. `gulp scripts`
1. Visit a page with FLC like http://localhost:8000/about-us/blog/
1. Open the FLC, click into the Topics multiselect, see that the FLC expands
1. Select one of the topics, see that the FLC expands when the pill is added above the multiselect
1. Select another topic
1. Click outside the multiselect to have it collapse, see that the FLC contracts
1. Click to remove each topic and see that the FLC contracts when they are removed

## Checklist

### Browsers

- [x] Chrome
- [x] Firefox
- [x] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Chrome for Android

### DIFF
--- a/cfgov/unprocessed/js/organisms/FilterableListControls.js
+++ b/cfgov/unprocessed/js/organisms/FilterableListControls.js
@@ -68,14 +68,14 @@ function FilterableListControls( element ) {
         multiSelect.addEventListener( 'expandBegin', function refresh() {
           window.setTimeout(
             _expandable.transition.expand.bind( _expandable.transition ),
-            300
+            250
           );
         } );
 
         multiSelect.addEventListener( 'expandEnd', function refresh() {
           window.setTimeout(
             _expandable.transition.expand.bind( _expandable.transition ),
-            300
+            250
           );
         } );
       } );
--- a/cfgov/unprocessed/js/organisms/FilterableListControls.js
+++ b/cfgov/unprocessed/js/organisms/FilterableListControls.js
@@ -78,13 +78,6 @@ function FilterableListControls( element ) {
             300
           );
         } );
-
-        multiSelect.addEventListener( 'selectionsUpdated', function refresh() {
-          window.setTimeout(
-            _expandable.transition.expand.bind( _expandable.transition ),
-            300
-          );
-        } );
       } );
     }
 
--- a/cfgov/unprocessed/js/organisms/FilterableListControls.js
+++ b/cfgov/unprocessed/js/organisms/FilterableListControls.js
@@ -66,11 +66,24 @@ function FilterableListControls( element ) {
     if ( _dom.classList.contains( 'o-filterable-list-controls' ) ) {
       multiSelects.forEach( function( multiSelect ) {
         multiSelect.addEventListener( 'expandBegin', function refresh() {
-          window.setTimeout( _expandable.transition.expand, 250 );
+          window.setTimeout(
+            _expandable.transition.expand.bind( _expandable.transition ),
+            300
+          );
         } );
 
         multiSelect.addEventListener( 'expandEnd', function refresh() {
-          window.setTimeout( _expandable.transition.expand, 250 );
+          window.setTimeout(
+            _expandable.transition.expand.bind( _expandable.transition ),
+            300
+          );
+        } );
+
+        multiSelect.addEventListener( 'selectionsUpdated', function refresh() {
+          window.setTimeout(
+            _expandable.transition.expand.bind( _expandable.transition ),
+            300
+          );
         } );
       } );
     }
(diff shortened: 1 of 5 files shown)


cfgov/unprocessed/js/organisms/FilterableListControls.js
```diff
@@ -66,11 +66,17 @@ function FilterableListControls( element ) {
     if ( _dom.classList.contains( 'o-filterable-list-controls' ) ) {
       multiSelects.forEach( function( multiSelect ) {
         multiSelect.addEventListener( 'expandBegin', function refresh() {
-          window.setTimeout( _expandable.transition.expand, 250 );
+          window.setTimeout(
+            _expandable.transition.expand.bind( _expandable.transition ),
+            300
```
anselmbradford: Why the change from 250 to 300?

anselmbradford:
> Added resolve: { symlinks: false } in webpack-config.js files, which I needed to test with an npm linked cf-expandables. This PR no longer needs any changes in cf-expandables, but this will be useful to anyone in the future who wants to test local updates to CF components within cfgov-refresh.

This seems like a lot of cruft to add across multiple configs for possible future testing purposes. Would it make more sense to add this in the CF docs with `npm link` to explain how to test outside of the repo? Unless I'm mistaken, the majority of changes we test in CF happen within the CF repo, not across repos.

cfgov/unprocessed/js/organisms/FilterableListControls.js
anselmbradford: Also, I'm forgetting why we need the delay at all? During testing, the delay in resize is kind of distracting.

Scotchester: Because we have to wait for the transition on the multiselect itself to complete to know what the new desired height is. I agree it's distracting, but I'm not sure what else we can do.

I _think_ I lengthened the delay for a similar reason? Had to be sure the multiselect was done animating. But it's possible it was conflated with other work I discarded. I'll try going back to 250.

Scotchester:
> This seems like a lot of cruft to add across multiple configs for possible future testing purposes. Would it make more sense to add this in the CF docs with `npm link` to explain how to test outside of the repo? Unless I'm mistaken, the majority of changes we test in CF happen within the CF repo, not across repos.

I don't know that it'd be possible to document how to do this succinctly. I also think it's impractical, when situations like this come up where a change _has_ to be tested in cfgov-refresh to know it's working as intended (or not), for someone to have to make all these changes temporarily. Our Webpack config is too complex for that to be a reasonable thing to do. These situations may come up infrequently, but I don't see any harm in just making the change now so we don't have to worry about it again.

cfgov/unprocessed/js/organisms/FilterableListControls.js
anselmbradford: Aww right. We'll need to have the multiselect dispatch its own event that the transition is complete. You might add a `TODO` for that since we both agree this could be better. But better than the live bug!

anselmbradford:
> I don't know that it'd be possible to document how to do this succinctly. I also think it's impractical, when situations like this come up where a change has to be tested in cfgov-refresh to know it's working as intended (or not), for someone to have to make all these changes temporarily. Our Webpack config is too complex for that to be a reasonable thing to do. These situations may come up infrequently, but I don't see any harm in just making the change now so we don't have to worry about it again.

You don't _have_ to `npm link`, we could promote copying the changes into `node_modules` directly. `npm link` is nice for the CF repo because changes are frequently happening there between the source and the docs (though I've had issues with it there). If you copy changes over, they'll be safely wiped out on next `npm install` after testing. At any rate, I'll approve this as this is a bug on prod. If you think this is necessary, that's fine, but since the webpack configs are complex if there's another way it would be nicer if we didn't have to add further configuration.

cfgov/unprocessed/js/organisms/FilterableListControls.js
Scotchester: Reverted this to 250 and rebased on master.

Scotchester:
As I'm traveling for the next three days, going to go ahead with the merge so we can get this deployed today.

PR #4516, opened 2018-09-13
Remove eRegulations

willbarton:
This PR removes the eRegulations dependencies and related configuration. This should enable us to proceed with the Django 1.11 upgrade.

~~I'm putting a hold label on this for right now until the launch dust clears for Regs3k.~~

## Removals

- eRegulations

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/production.py
+++ b/cfgov/cfgov/settings/production.py
@@ -89,22 +89,6 @@ CACHES.update({
         'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
         'LOCATION': '/tmp/eregs_cache',
     },
-    'eregs_longterm_cache': {
-        'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
-        'LOCATION': '/tmp/eregs_longterm_cache',
-        'TIMEOUT': 60*60*24*15,     # 15 days
-        'OPTIONS': {
-            'MAX_ENTRIES': 10000,
-        },
-    },
-    'api_cache': {
-        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
-        'LOCATION': 'api_cache_memory',
-        'TIMEOUT': 3600,
-        'OPTIONS': {
-            'MAX_ENTRIES': 1000,
-        },
-    },
     'post_preview': {
         'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
         'LOCATION': 'post_preview_cache',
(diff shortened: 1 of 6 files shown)


cfgov/cfgov/settings/production.py
```diff
@@ -87,23 +87,7 @@
...
-        },
-    },
-    'api_cache': {
-        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
-        'LOCATION': 'api_cache_memory',
-        'TIMEOUT': 3600,
-        'OPTIONS': {
-            'MAX_ENTRIES': 1000,
-        },
+        'LOCATION': '/tmp/django_cache',
```
chosak: Although this is a great change to make, I fear that this might cause problems during deployment. We could discuss/address this in another forum, but essentially, while the current servers happen to have `/tmp/eregs_cache` and `/tmp/eregs_longterm_cache` set up, I don't believe these are covered by the deployment scripts (e.g. Ansible).

See also internal platform#578. We could manually create a `/tmp/django_cache` or perhaps add that step to the Ansible deploy, but I might actually proposing removing the default file-based cache in favor of a database-based cache, since we're already using that for the post preview and mega menu (tables `post_preview_cache` and `default_fragment_cache`). We could also consolidate all of these uses into a single database cache table (which would be the new `default` cache), distinguishing their content by cache key prefix.

willbarton: I'll leave it alone for now.

saddest_face_ever_by_loveless_nights-d66ra4b

> See also internal platform#578. We could manually create a /tmp/django_cache or perhaps add that step to the Ansible deploy, but I might actually proposing removing the default file-based cache in favor of a database-based cache, since we're already using that for the post preview and mega menu (tables post_preview_cache and default_fragment_cache). We could also consolidate all of these uses into a single database cache table (which would be the new default cache), distinguishing their content by cache key prefix.

This sounds like an excellent idea.

cfgov/cfgov/urls.py
```diff
@@ -431,38 +431,16 @@ def flagged_wagtail_only_view(flag_name, regex_path, url_name=None):
...
-    flagged_url(
-        'REGULATIONS3K',
-        r'^eregulations/',
-        include_if_app_enabled('regulations', 'regulations.urls'),
-        fallback=redirect_eregs,
-        name='eregs-redirect',
-        state=False
-    ),
+    # Explicitly redirect eRegulations URLs to Regulations3000
+    url(r'^eregulations/', redirect_eregs, name='eregs-redirect'),
```
higs4281: I think this should be a greedier `eregulations/.*`

willbarton: Yup, I'll update with your changes.

willbarton:
lets-do-this

PR #4517, opened 2018-09-13
regsek: Fix redirect errors

higs4281:
Our redirect URL wasn't greedy enough, and some interpretation requests weren't matching.

Interp matching still needs work.

### DIFF
--- a/cfgov/regulations3k/views.py
+++ b/cfgov/regulations3k/views.py
@@ -214,4 +214,4 @@ def redirect_eregs(request, **kwargs):
                             new_base, part, section), permanent=True)
             else:
                 return redirect("{}{}/".format(new_base, part), permanent=True)
-    return redirect("{}{}/".format(new_base, part), permanent=True)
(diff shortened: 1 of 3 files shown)


cfgov/regulations3k/views.py
```diff
@@ -155,56 +156,62 @@ def redirect_eregs(request, **kwargs):
...
-            if pattern == INTERP_SECTION_RE:
-                section = INTERP_SECTION_DEFAULTS.get(part, '1')
-                if version_date:
-                    return redirect("{}{}/{}/Interp-{}/".format(
-                        new_base, part, version_date,
-                        section), permanent=True)
-                else:
-                    return redirect("{}{}/Interp-{}/".format(
-                        new_base, part, section), permanent=True)
+    if part_match:
```
willbarton: One quick thing, so we don't have to have the meat of this view indented so much, this could be `if not part_match: return redirect(new_base)` and then everything else could be outdented one tab.

PR #4519, opened 2018-09-13
Add the ability to respond to a Question in GovDelivery question to indicate conference registration

Scotchester:
A desire was expressed to capture conference registrations in GovDelivery on an existing list, rather than creating a new list. By having registrations use the GovDelivery API to "answer a question", the list can be segmented and conference-specific emails can be sent only to the portion of the list where the question has been answered. This PR implements the necessary changes to the Conf Reg Form in Wagtail to support that.

## Additions

- `govdelivery_question_id` field added to `ConferenceRegistrationForm` block
- `govdelivery_question_response` method added to `ConferenceRegistrationForm` form, called on `save` if the form's block has a value in its `govdelivery_question_id` field

## Removals

- Unused `subscribe` macro

## Testing

1. Pull branch
1. Create/edit a conf reg form (on a Browse page)
1. Test various states of the new GovDelivery Question ID field and its validation
1. `tox -e fast` to run the unit tests (🎩 @chosak for the help with those)

## Checklist


cfgov/data_research/tests/test_forms.py
```diff
@@ -10,22 +10,26 @@
 class ConferenceRegistrationFormTests(TestCase):
     capacity = 100
     govdelivery_code = 'TEST-CODE'
+    govdelivery_question_id = 12345
```
chosak: Minor thing that doesn't affect the execution of these tests: this (and other references below, see diff [here](https://gist.github.com/chosak/ef5cee23b6f56316357ed472af5bbe56)) should probably be a string `'12345'` to be consistent with how we expect the data to be passed in. This doesn't break the tests because [this line](https://github.com/cfpb/cfgov-refresh/pull/4519/files#diff-17c90d3c3fe8d73df852c60805e01d64R174) is mocked out, but in practice this will always need to be a string to avoid an error [here](https://github.com/cfpb/govdelivery/blob/master/govdelivery/api.py#L125).

cfgov/data_research/forms.py
```diff
@@ -140,17 +142,38 @@ def save(self, commit=True):
             # Subscribe this registrant to GovDelivery.
             self.govdelivery_subscribe(code=self.govdelivery_code, email=email)
 
+            # Update their question response, if appropriate.
+            if self.govdelivery_question_id:
+                self.govdelivery_question_response(
```
chosak: Unfortunately, because this requires two separate API calls, there's a possibility that an error on the second call would result in an incomplete registration.

Scotchester: I did notice that, and I think it's probably desirable. If we're counting on the question response to indicate who's registered in GovDelivery, then a failure on the second call _is_ an incomplete registration.

cfgov/data_research/tests/test_forms.py
Scotchester: Thanks, updating that.

cfgov/data_research/forms.py
chosak: Right, I see. Such an error would leave someone in a state where they were signed up to the GD list, but not with the correct question ID, and also not recorded in the DB. We'd lose their registration information (company, etc). Hopefully they'd see the error message and re-submit, but if not, they'd end up being left on the mailing list. These kind of errors would (reasonably) require some manual review to check/fix.

PR #4521, opened 2018-09-14
Add 10.0.2.2 to INTERNAL_IPS to support Django Debug Toolbar in Docker

willbarton:
This PR adds `10.0.2.2` to `INTERNAL_IPS` when `ENABLE_DEBUG_TOOLBAR` is set. This will allow usage of the Django Debug Toolbar in our Docker setup.

~~This PR makes the `INTERNAL_IPS` setting configurable to allow us to use Django Debug Toolbar in Docker when `ENABLE_DEBUG_TOOLBAR` is enabled.~~

~~This is a WIP because: @wpears @rosskarchner I'm not sure where the `10.0.2.2` IP address comes from, or whether it's specific to my machine. I discovered it by printing `request.META['REMOTE_ADDR']` in a request cycle to discover what IP requests from my Mac were coming from inside the python container. Is there a way we can make that information more easily discoverable? Is `10.0.2.2` consistent for everyone using our Docker setup?~~

## Testing

1. Add the following variable to your `.env` file:

   ```
   export ENABLE_DEBUG_TOOLBAR=''
   ```

2. Restart the `python` container
3. Visit http://localhost:8000 and observe that the Django Debug Toolbar is enabled.

## Todos

~~The source of the IP that needs to be added to enable the toolbar on docker needs to be documented.~~

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/local.py
+++ b/cfgov/cfgov/settings/local.py
@@ -40,7 +40,8 @@ LOGGING = {
 if os.environ.get('ENABLE_DEBUG_TOOLBAR'):
     INSTALLED_APPS += ('debug_toolbar',)
 
-    INTERNAL_IPS = ('127.0.0.1',)
+    INTERNAL_IPS = ('127.0.0.1', '10.0.2.2')
+
     MIDDLEWARE_CLASSES += ('debug_toolbar.middleware.DebugToolbarMiddleware',)
 
     DEBUG_TOOLBAR_CONFIG = {
--- a/.env_SAMPLE
+++ b/.env_SAMPLE
@@ -189,3 +189,10 @@ if [ $(uname -s) = "Darwin" ]; then
   export ES_HOST=localhost
   source activate-virtualenv.sh
 fi
+
+###############################################
+# Django Debug Toolbar
+###############################################
+
+# Enable the debug toolbar.
+# export ENABLE_DEBUG_TOOLBAR=True


cfgov/cfgov/settings/local.py
```diff
@@ -40,7 +40,9 @@
 if os.environ.get('ENABLE_DEBUG_TOOLBAR'):
     INSTALLED_APPS += ('debug_toolbar',)
 
-    INTERNAL_IPS = ('127.0.0.1',)
+    INTERNAL_IPS = [
+        i for i in os.environ.get('INTERNAL_IPS', '127.0.0.1').split(' ')
```
acrewdson: could this be written without the list comp and without the `split` arg, just: `os.environ.get('INTERNAL_IPS', '127.0.0.1').split()`?

chosak: Any reason why we shouldn't just add `10.0.0.2` to this list (maybe with a comment informing what it is there for) and leave it at that? It wouldn't be needed in non-Docker environments but would be much simpler and would avoid needing to define this in `.env`.

rosskarchner:
Instead of messing with IP's, how about setting SHOW_TOOLBAR_CALLBACK to a function that returns True if DEBUG is true (ignoring INTERNAL_IPS)  https://django-debug-toolbar.readthedocs.io/en/latest/configuration.html

willbarton:
> Instead of messing with IP's, how about setting SHOW_TOOLBAR_CALLBACK to a function that returns True if DEBUG is true (ignoring INTERNAL_IPS) https://django-debug-toolbar.readthedocs.io/en/latest/configuration.html

@rosskarchner I like the simplicity of this solution, but I'm not sure I want the toolbar on all the time. It *does* slow down local requests quite a lot.

willbarton:
@rosskarchner Oh! Right! Yes, it would! 

I'll do exactly what you suggest.

willbarton:
I've decided to go as simple as possible and add `10.0.2.2` to `INTERNAL_IPS`. If this becomes problematic, we can explore other alternatives. I've updated the PR accordingly.

I also added `ENABLE_DEBUG_TOOLBAR` to `.env_SAMPLE` and left it commented out.

PR #4529, opened 2018-09-19
Make Resources searchable and filterable, convert Snippet Lists to Resource Lists

Scotchester:
This started as simply wanting to make Resource snippets searchable and filterable (/cc @sonnakim) and evolved into simplifying Snippet Lists and renaming them to Resource Lists, as well. Review the first commit to see the work done for setting up a ModelAdmin view for Resources, and the subsequent commits to see the disentangling of Resources from Snippets.

Refer to GHE issues https://GHE/CFGOV/platform#2984 and https://GHE/CFPB/Hubcap#62 for more background.

Many thanks to @chosak for pairing with me on various pieces of this.

## Additions

- ModelAdmin view of Resources that supports searching and filtering them

## Changes

- Resources are no longer registered as snippets
- Snippet Lists have been renamed to Resource Lists and no longer pretend to support any other types of snippets (which they never really did)
- Fixes Snippet List header cell background color

## Testing

(assumes recent DB loaded)

1. Pull branch
1. `gulp styles`
1. Load an existing page that has a Resource List (née Snippet List) and confirm it still loads as expected, e.g., [MSEM](http://localhost:8000/consumer-tools/managing-someone-elses-money/) or [library outreach materials](http://localhost:8000/practitioner-resources/library-resources/outreach-materials-to-use-outside-the-library/)
1. Edit one of the Resource Lists on one of those pages and confirm that things save correctly and still load
1. Go to the new **Resources** item in the main Wagtail sidebar
1. Play around with searching & filtering
1. Use search or filters to edit one of the resources on one of the lists you saw before
1. Reload the page and see that the edits correctly took effect
1. Run the tests with `tox -e fast`

## Screenshots

screen shot 2018-09-19 at 12 53 34

## Notes

- To avoid the necessity for a data migration, I left the name of the StreamField block in BrowsePage and SublandingPage as `snippet_list`. If people feel strongly about it, I could rename and do the migration dance, but I know people are anxious to get these features available.

## Todos

- [ ] Notify Wagtail editors about this change via docs/announcement (/cc @schaferjh)

## Checklist


Scotchester:
Tagged @richaagarwal and @virginiacc on this as well, not necessarily for a full review, but in case this has any bearing on/implications on/synergy with what you all are working on.

richaagarwal:
@Scotchester Thanks for tagging me! This potentially impacts my work but should be manageable either way.  

I plan to look at the diff more carefully but at a high-level, I have a few questions/possible concerns:
- What is the reason for converting the Resource to a non-snippet Django model? Is it to use the built in model admin features?  Because you can search them by doing things like `Resource.objects.filter(link__contains=search_term))`  
- Would this break the existing functionality to include a Resource on any page? Or could you outline the functionality changes?  I don't know much about how it's currently used - are they only ever used as part of a Resource List?

schaferjh:
@Scotchester re: informing folks about the change, I'm happy to but where is the change in workflow documented? (and the very least it seems like it should go in `hubcap/wiki/Announcements-of-Wagtail-changes` but it looks like several things are missing from there...) 

Or perhaps the question is, what is the change in workflow beyond the additional option of searching to find the right snippet?

Scotchester:
@schaferjh 

> [W]hat is the change in workflow beyond the additional option of searching to find the right snippet?

1. When going to create or update resources, instead of first going to the **Snippets** page, one will now just click **Resources** in the main sidebar.
   - If you're editing a resource, you can now use the on-page search and filtering to find what you're looking for, rather than paging endlessly through resources that are ordered in a way that is unhelpful when they're all there together in the admin.
1. Not really a workflow change, but just an awareness thing: the module that used to be called "Snippet List" is now called "Resource List".

Other than that, it's exactly the same.

I will certainly update that announcements page; just wanted to make sure that these changes didn't seem onerous to you before I did so.

schaferjh:
@Scotchester ok, cool. These changes don't seem onerous. I am getting a little concerned about the amount of stuff in that main menu sidebar, but no reason to hold this up! 👍

willbarton:
For posterity I want to propose a possible way we could reduce the amount of stuff in the sidebar: I think we could probably combine Snippets and Resources into a single `ModelGroup` and remove Snippets from the sidebar altogether. Snippets are Django models, so we should be able to create a Wagtail ModelAdmin for them and add it to the same `ModelGroup` as `Resources`.

Scotchester:
I took @willbarton's advice and updated this PR to give all snippets a `ModelAdmin` and group them all under a new Snippets `ModelAdminGroup`, including Resources. Including Resources in that menu will make the change less disruptive to users already familiar with editing Resources.

Contacts and Reusable Texts remain registered as snippets so that they can be accessed as usual, with a `SnippetChooser`, but I'm leaving Resources deregistered; they aren't accessed via chooser, and their complexity cf. the other snippets warrants keeping them broken out into their own file. (Plus I'd have to undo all the annoying changes I already did to separate them out 😜)

Scotchester:
Here's how the new Snippets menu looks:

screen shot 2018-09-25 at 10 53 30

cfgov/v1/wagtail_hooks.py
```diff
@@ -163,3 +168,84 @@ class MegaMenuModelAdmin(ModelAdmin):
...
+    for resource in Resource.objects.all():
+        for tag in resource.tags.all():
+            tuple = (tag.slug, tag.name)
+            if tuple not in tag_list:
+                tag_list.append(tuple)
+
+    return sorted(tag_list, key=lambda tup: tup[0])
+
+
+class ResourceTagsFilter(admin.SimpleListFilter):
```
chosak: Nice-to-have: unit tests for this class. There are some complex Django unit tests that might be useful [here](https://github.com/django/django/blob/1.11.15/tests/admin_filters/tests.py#L251).

cfgov/v1/wagtail_hooks.py
```diff
@@ -163,3 +168,84 @@ class MegaMenuModelAdmin(ModelAdmin):
 def clear_mega_menu_cache(sender, instance, **kwargs):
     from django.core.cache import caches
     caches['default_fragment_cache'].delete('mega_menu')
+
+
+def get_resource_tags():
+    tag_list = []
+
+    for resource in Resource.objects.all():
```
chosak: I suspect this lookup could be done in a more optimal way. This probably doesn't matter much given the size of the data being queried, though.

higs4281:
This is a great improvement!

I worried that the entity escaping seen on the reusable-text list view (`/admin/v1/reusabletext/`) might be disconcerting to editors, but functionally it's not a problem, and it actually gives you a clue that some Latin-1 pasting has been going on.

escaping

cfgov/v1/wagtail_hooks.py
Scotchester: Working on this. I've added some tests for the `get_resource_tags` function above here.

Scotchester: Tests for ResourceTagsFIlter have now been added!

Scotchester: The `ResourceTagsFilter` tests were based on Wagtail's tests here: https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/modeladmin/tests/test_simple_modeladmin.py#L53-L63

cfgov/v1/tests/test_wagtail_hooks.py
```diff
@@ -192,3 +193,48 @@ def test_get_resource_tags_returns_alphabetized_list(self):
...
+    def get(self, **params):
+        return self.client.get('/admin/v1/resource/', params)
+
+    def test_no_params_returns_all_resources_in_order(self):
+        response = self.get()
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.context['result_count'], 3)
+        self.assertEqual(
+            list(response.context['object_list']),
+            list(Resource.objects.all().order_by('title'))
```
higs4281: Nit:
- `.order_by('title')` seems unnecessary here, because the model specifies the sort. The test succeeds without it.

Scotchester: Huh! That's surprising, because the [ordering on the model is actually `'order', 'title'`](https://github.com/cfpb/cfgov-refresh/pull/4529/files#diff-1a53f67ccb2f8350dd16136cfe0378d2R110). But you're right, it passes without that.

Scotchester: I just realized that it's because I didn't set an `order` on any of the test snippets 🤦‍♂️ Update coming.

Scotchester: Update pushed (amended previous commit and rebased everything on top of master).

PR #4560, opened 2018-09-27
Add tests to custom makemessages command

chosak:
This change adds a Python test for the v1 app's custom `makemessages` Django management command, which overrides the default one documented at

https://docs.djangoproject.com/en/1.11/ref/django-admin/#django-admin-makemessages

Our custom makemessages command adds support for block tags like this:

```
{% trans %}
Multi-line
statement
here
{% endtrans %}
```

which is not supported in Django templates.

This change also removes support in this command for Django 1.8 now that this project is solely targeted at Django 1.11.

## Testing

To test this locally, you can verify that the unit test passes, and then remove our custom makemessages command (including the .pyc file if on Python 2). The built-in command will fail as it does not detect the multi-line Jinja2 tag.

Additionally, you can run through the makemessages workflow documented [here](https://cfpb.github.io/cfgov-refresh/translation/) to confirm that the command still works.

## Checklist

### DIFF
--- a/cfgov/v1/management/commands/makemessages.py
+++ b/cfgov/v1/management/commands/makemessages.py
@@ -5,18 +5,12 @@
 # The code is based on this StackOverflow question and answer:
 # https://stackoverflow.com/questions/2090717/getting-translation-strings-for-jinja2-templates-integrated-with-django-1-x
 # And the Django docs:
-# https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#customizing-the-makemessages-command
+# https://docs.djangoproject.com/en/1.11/topics/i18n/translation/#customizing-the-makemessages-command
 
 import re
 
-from django import VERSION
 from django.core.management.commands import makemessages
-
-
-if VERSION[:2] < (1, 11):
-    from django.utils.translation import trans_real
-else:
-    from django.utils.translation import template as trans_real
+from django.utils.translation import template as trans_real
 
 
 class Command(makemessages.Command):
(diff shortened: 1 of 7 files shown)


schbetsy:
When I run `tox -r -e fast` I get an error on the `makemessages` test:

```
CommandError: Can't find msguniq. Make sure you have GNU gettext tools 0.15 or newer installed.
```

But I seem to have an up-to-date gettext installed (unless these are referring to different tools?):

```
> brew upgrade gettext
Error: gettext 0.19.8.1 already installed
```

chosak:
@schbetsy quite honestly I can't remember, but I do have that `brew link gettext --force` way back in my bash history. This is my local version:

```sh
$ gettext --version
gettext (GNU gettext-runtime) 0.19.8.1
```

schbetsy:
Okay, after running `brew link gettext --force`, this now works for me. How would you feel about adding that to the documentation somewhere before we merge? 

I get the feeling we'll get a lot of questions about it after we merge, since people won't be able to run the unit tests locally without doing that command. Should we do anything else to communicate the change?

schbetsy:
Yep, I like that plan! And I think the documentation you proposed looks good.

chosak:
@schbetsy I've updated the documentation and added logic to skip the translation test without gettext.

---

The documentation looks like this when rendered:

image

image

---

The tests will be skipped if you don't have gettext:

```sh
$ brew unlink gettext
$ tox -e fast v1.tests.management.commands.test_makemessages
s
Ran 1 test in 0.000s

OK (skipped=1)
```

and pass if you do:

```sh
$ brew link --force gettext
$ tox -e fast v1.tests.management.commands.test_makemessages
.
Ran 1 test in 0.092s

OK
```

PR #4561, opened 2018-09-28
Update StdoutCapturingTestRunner to Django 1.11

chosak:
This change updates the logic in our custom [StdoutCapturingTestRunner](https://github.com/cfpb/cfgov-refresh/blob/337f9a69aba8f56eb0791d4554463088c482d878/cfgov/cfgov/test.py#L195) class to simplify it a bit and remove a reference to Django 1.8.

Unfortunately this change actually has the side effect of making the unit tests for this custom test runner a bit more complicated.

Between 1.8 and 1.11, a change was made to address this Django ticket:

https://code.djangoproject.com/ticket/26981

The ability was added for Django TestRunner classes to provide custom kwargs to the underlying unittest test runner, which is the thing that actually runs the tests. This was implemented in this PR:

https://github.com/django/django/pull/7009

That PR added the ability for Django TestRunner classes to define a custom get_test_runner_kwargs method to provide custom kwargs.

By default the unittest test runner (TextTestRunner) prints its status to the console using stderr (this is the output that contains `...E..F` and shows how many tests are run). When we actually test our test runner, we want to suppress this output.

See comment in the modified test file for a bit more explanation.

## Checklist

### DIFF
--- a/cfgov/cfgov/test.py
+++ b/cfgov/cfgov/test.py
@@ -31,8 +31,6 @@ except ImportError:
         sys.stdout = new_target
         try:
             yield
-        except Exception:
-            raise
         finally:
             sys.stdout = sys.__stdout__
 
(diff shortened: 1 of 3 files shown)


schbetsy:
It seems something is amiss with the setup here. I added a `print` statement to a randomly-chosen test, and it didn't trigger the runtime error we want:
screen shot 2018-10-01 at 10 50 24 am

chosak:
@schbetsy good catch - I just tried this on master and it doesn't fail there either. I'll see if I can figure out what's happening here.

schbetsy:
Ah, (as we discussed offline) `TEST_RUNNER=cfgov.test.StdoutCapturingTestRunner tox -e fast` works as expected. I'd recommend putting a sentence about this in the back-end testing docs. otherwise, LGTM.

chosak:
Thanks @schbetsy. I've added a note to [the backend testing doc page](https://cfpb.github.io/cfgov-refresh/testing-be/) about use of that environment variable:

image

PR #4564, opened 2018-10-02
Upgrade Python Libraries

schbetsy:
Upgrade python libraries where possible.
Note: I did not touch any libraries that are elasticsearch related. I also only touched libraries that are in the `test` and `libraries` requirements collections, since those are covered by unit tests.

## Checklist

### Browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/requirements/libraries.txt
+++ b/requirements/libraries.txt
@@ -1,5 +1,5 @@
 address==0.1.1
-beautifulsoup4==4.5.1
+beautifulsoup4==4.6.3
 boto3==1.7.80
 dj-database-url==0.5.0
 djangorestframework==3.6.4
--- a/requirements/libraries.txt
+++ b/requirements/libraries.txt
@@ -19,7 +19,7 @@ django-overextends==0.4.3
 django-storages==1.5.0
 django-treebeard==3.0
 django-watchman==0.15.0
-edgegrid-python==1.0.10
+edgegrid-python==1.1.1
 elasticsearch==2.4.1
 govdelivery==1.2
 Jinja2==2.7.3
--- a/requirements/libraries.txt
+++ b/requirements/libraries.txt
@@ -24,7 +24,7 @@ elasticsearch==2.4.1
 govdelivery==1.2
 Jinja2==2.7.3
 jsonfield==2.0.2
-lxml==4.1.0
+lxml==4.2.5
 Markdown==2.6.11
 markdown-emdash==0.1.0
 ntplib==0.3.3
--- a/requirements/libraries.txt
+++ b/requirements/libraries.txt
@@ -1,7 +1,7 @@
 address==0.1.1
 beautifulsoup4==4.5.1
 boto3==1.7.80
-dj-database-url==0.4.2
+dj-database-url==0.5.0
 djangorestframework==3.6.4
 djangorestframework-csv==2.0.0
 django-csp==3.4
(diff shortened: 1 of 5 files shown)


schbetsy:
@chosak  So, I'm finding a bunch of libraries to update (and still finding more)... but I realized that just because tests pass doesn't mean the new versions are guaranteed to work when deployed. 

I realized it might be safer to just update a few libraries at a time and see what happens. What's your opinion on how much to do at once?

requirements/libraries.txt
```diff
@@ -5,51 +5,35 @@ dj-database-url==0.4.2
...
 jsonfield==2.0.2
-jsonschema==2.0.0
 lxml==4.1.0
 Markdown==2.6.11
 markdown-emdash==0.1.0
 ntplib==0.3.3
 openpyxl==2.5.0
 pyelasticsearch==0.6.1
-python-dateutil==2.1
-PyYAML==3.11
```
chosak: I agree that this should be removed, as it's not needed anywhere in cfgov-refresh. It is required by college-costs and included [in its setup.py](https://github.com/cfpb/college-costs/blob/master/setup.py#L16) with `>=3.11,<3.14`.

Ideally, we could remove this and college-costs would bring it in as part of its `pip install`. But, like django-mptt, which is also included here because it is required by TDP (#4461), if we don't include PyYAML, then college-costs might not work right when run in Docker.

This is because the docs [recommend](https://github.com/cfpb/cfgov-refresh/blob/master/docs/usage.md#develop-satellite-apps) cloning satellite repos into a local develop-apps directory which is automatically added to the `PYTHONPATH` [in the Python container](https://github.com/cfpb/cfgov-refresh/blob/8c780180d5352fcd4d7f7ee63aee343fc8f91e9f/extend-environment.sh).

Adding satellite app clones to the Python path like this means that when cfgov-refresh code tries to import the [`paying_for_college`](https://github.com/cfpb/college-costs/tree/master/paying_for_college) app, say, it'll work. But if that app then tries to import `yaml`, it'll fail, because PyYAML was never installed in the Python environment.

A better way forward (that has been discussed a bit in other venues) might be to recommend that folks in Docker use `pip install -e` to install satellite apps ([as we do](https://cfpb.github.io/cfgov-refresh/development-tips/#tip-loading-satellite-apps) for the non-Docker setup). This would install the satellites' dependencies into the environment properly, and eliminate the requirement to duplicate them here. I would support that change but it's probably out of scope for this particular PR. For now perhaps we could just add a comment for dependencies like PyYAML (like we already have for django-mptt) indicating why it remains.

schbetsy: That makes sense. Thanks for the explanation. I added PyYAML back in, with a comment, as you suggested.

schbetsy:
I believe this is ready for review again, @chosak.

cfgov/cfgov/settings/base.py
```diff
@@ -386,6 +386,7 @@
 AWS_S3_FILE_OVERWRITE = False
 AWS_S3_SECURE_URLS = True  # True = use https; False = use http
 AWS_STORAGE_BUCKET_NAME = os.environ.get('AWS_STORAGE_BUCKET_NAME')
+AWS_DEFAULT_ACL = None  # Default to using the ACL of the bucket
```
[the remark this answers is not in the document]
schbetsy: I see. No, not intentional. I took the documentation's recommendation without enough thought as to what we have those values set to.

Since this is complex, I created a separate ticket for it. I'll revert this change for this PR.

PR #4584, opened 2018-10-10
Fix doubled border in mobile footer

### LINKED ISSUE #3575: Double border in footer links at mobile
## Current behavior

- When at mobile size the links in the footer stack on top of each other. Since it's three columns at desktop sizes, the three stacked columns end up with double borders between each stacked column. This is visible after the "Open Government" and "Tribal" link.

## Expected behavior

- Borders are not doubled up at mobile.

## Steps to replicate behavior

1. Resize consumerfinance.gov to mobile and scroll to the footer.

## Screenshots

__Current__

screen shot 2017-11-09 at 9 44 53 am

wpears:
Closes #3575 

Does so by zeroing out the border-top-width of all list items except in the first column, which corresponds to the top of the footer in the collapsed mobile view.

To see the fix:

 - Pull branch
 - `npm run gulp styles`
 - Open any page, shrink browser to a mobile view, and bask in the non-doubled borders

### DIFF
--- a/cfgov/unprocessed/css/organisms/footer.less
+++ b/cfgov/unprocessed/css/organisms/footer.less
@@ -158,6 +158,16 @@
             margin: 0;
         }
 
+        .o-footer_col:nth-child(n+2) {
+            .o-footer_list {
+                .respond-to-max( @bp-xs-max, {
+                    .m-list_item .m-list_link {
+                        border-top-width: 0;
+                    }
+                } );
+            }
+        }
+
         .respond-to-min( @bp-sm-min, {
             .grid_column(8);
             border-right: 1px solid @gray-40;
@@ -174,6 +184,14 @@
     }
 
     &-middle-right {
+        .o-footer_list {
+            .respond-to-max( @bp-xs-max, {
+                .m-list_item .m-list_link {
+                    border-top-width: 0;
+                }
+            } );
+        }
+
         .respond-to-min( @bp-sm-min, {
             .grid_column(4);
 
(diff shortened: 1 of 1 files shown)


cfgov/unprocessed/css/organisms/footer.less
```diff
@@ -158,6 +158,16 @@
             margin: 0;
         }
 
+        .o-footer_col:nth-child(n+2) {
+            .o-footer_list {
+                .respond-to-max( @bp-xs-max, {
```
anselmbradford: Should the media query go on the outside of this entire block?

cfgov/unprocessed/css/organisms/footer.less
```diff
@@ -174,6 +184,14 @@
     }
 
     &-middle-right {
+        .o-footer_list {
+            .respond-to-max( @bp-xs-max, {
```
anselmbradford: Should the media query go out the outside of this entire block, instead of some class selectors on the outside and some on the inside?

wpears: Yeah I think that's probably cleaner less, fixed in 534d1f2.

cfgov/unprocessed/css/organisms/footer.less
wpears: Yeah I think that's probably cleaner less, fixed in 534d1f2.

wpears:
:+1: comments and spacing updated

PR #4589, opened 2018-10-12
Regs3k: Add a section-linking script

higs4281:
This script is mostly a run-once utility to add relative internal links
to a regulation's active regdown content. It was run once against our
currently live regulations.

The script will not alter:
* Any section it has added links to before
* Any section with no internal section references to link to
* Interpretation sections, because adding links there would corrupt
the interpretation resolver process.

The script could be run against any future regulations that are imported,
as an option in the Jenkins `regs3k-import-regulation` job.

The linking routine is not included in the ecfr_importer itself because
the links are specific to the Bureau's URL structure, and they would pollute 
the content for anyone else using the importer.


cfgov/regulations3k/scripts/insert_section_links.py
```diff
@@ -0,0 +1,82 @@
...
+def insert_section_links(regdown):
+    """Turn internal section references into links."""
+    section_refs = SECTION_RE.findall(regdown)
+    if not section_refs:
+        return
+    index_head = 0
+    for i, ref in enumerate(section_refs):
+        url = get_url(ref)
+        if url:
+            link = '<a href="{}" data-linktag="{}">{}</a> '.format(url, i, ref)
```
willbarton: Should it be an `a` tag or a Markdown link?

higs4281: For better or worse, I wanted script-added links to be unique and identifiable, so that the script wouldn't operate twice on the same reference if run again, and to provide a hook in case we wanted to strip out script-added links without touching editor-added markdown links.

The tags aren't as editor-friendly, but tag pollution could be undone.

PR #4595, opened 2018-10-18
Change method of setting origin param on YouTube URLs

Scotchester:
The fact that we don't have a reliable way to produce this `origin` string across different environments is probably causing us some issues.

See https://GHE/CFGOV/platform/issues/2895#issuecomment-186171 for further explanation.

### DIFF
--- a/cfgov/jinja2/v1/_includes/macros/video-player.html
+++ b/cfgov/jinja2/v1/_includes/macros/video-player.html
@@ -42,7 +42,7 @@
     {% if video_url.find( 'origin=' ) == -1 -%}
         {% set video_url = _buildParam(
              video_url,
-             'origin=' ~ request.scheme ~ '://' ~ request.META['HTTP_HOST']
+             'origin=' ~ request.scheme ~ '://' ~ request.get_host()
         ) %}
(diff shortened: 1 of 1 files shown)


cfgov/jinja2/v1/_includes/macros/video-player.html
```diff
@@ -39,12 +39,6 @@
     {% if video_url.find( 'enablejsapi=' ) == -1 -%}
         {% set video_url = _buildParam( video_url, 'enablejsapi=1' )  %}
     {% endif -%}
-    {% if video_url.find( 'origin=' ) == -1 -%}
-        {% set video_url = _buildParam(
-             video_url,
-             'origin=' ~ request.scheme ~ '://' ~ request.META['HTTP_HOST']
```
wpears: I think that instead of deleting this entirely, changing `request.META['HTTP_HOST']` to `request.get_host()` will fix this in at least beta and prod

Scotchester: I think you're right that this will work OK with some other configuration changes in some of our environments. I have amended the commit and rebased.

PR #4607, opened 2018-10-23
Make jobmanager tests work with Python3

willbarton:
This PR makes the jobmanager tests pass with Python 3. 

Two tests continue to fail because there seems to be a bug in the `AtomicTableBlock` rendering that differs between Python 2.7 and 3.6. The `AtomicTableBlock` tests also fail the same way, so I am making the choice to fix the problem when we get to those tests.

## Checklist

### DIFF
--- a/cfgov/jobmanager/models/django.py
+++ b/cfgov/jobmanager/models/django.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, unicode_literals
 
 from django.db import models
+from django.utils.six import python_2_unicode_compatible
 
 from wagtail.wagtailadmin.edit_handlers import FieldPanel, InlinePanel
 from wagtail.wagtailcore.fields import RichTextField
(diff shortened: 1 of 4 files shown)


cfgov/cfgov/test.py
```diff
@@ -183,9 +184,12 @@ def assertPageIncludesHtml(self, page, s):
         request = RequestFactory().get('/')
         request.user = Mock()
 
-        rendered_html = page.serve(request).render()
+        rendered_html = page.serve(request).render().content
+        if six.PY3:
```
chosak: Why is this extra step needed? Is it because the regexes [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/test.py#L173) aren't bytestrings?

willbarton: That's part of it, but there are places where we try to pass strings to those regexes as well.

If we want `assertHtmlRegexpMatches` to just expect to take `bytes` instead of `str` that's fine, it's just a different change.

willbarton: Other thought, I'm not sure why we're using a custom `assertPageIncludesHtml` instead of the Django `assertInHTML`

PR #4608, opened 2018-10-24
Backend Linting: Ignore W504 and fix W605

willbarton:
A recent update to flake8 started enforcing some new warnings, [W504 and W605](https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes).

W504 is a sibling to W503, [which we already ignore](https://github.com/cfpb/cfgov-refresh/blob/master/.flake8#L7). 

W605 is an explicit check for escape sequences that don't belong in certain string types without a leading `r`. (See the Python [String and Byte literals](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals) documentation.)

I've fixed those errors in this PR. They were all caused by regular expression strings that were not explicitly prefixed with `r`.

This PR will fix the recent Travis backend linting failures that are currently blocking PRs.


cfgov/v1/models/learn_page.py
```diff
@@ -200,7 +200,7 @@ class EventPage(AbstractFilterPage):
                   "It can be obtained by clicking on Share > "
                   "Embed on Youtube.",
         validators=[
-            RegexValidator(regex='^https?:\/\/www\.youtube\.com\/embed\/.*$')
+            RegexValidator(regex=r'^https?:\/\/www\.youtube\.com\/embed\/.*$')
```
Scotchester: Is this gonna require a migration?

willbarton: No, a change to the validator shouldn't.

willbarton: Confirmed with `tox -e missing-migrations`.

PR #4609, opened 2018-10-24
Make agreements tests work with Python 3

willbarton:
The agreements tests and code makes some assumptions about text encoding in zip files that needs to be more explicit in Python 3.

This PR also cleans up some bytes-to-string comparisons, generators-to-lists, and import moves.

### DIFF
--- a/cfgov/agreements/management/commands/_util.py
+++ b/cfgov/agreements/management/commands/_util.py
@@ -1,5 +1,9 @@
-import os
+from __future__ import unicode_literals
 
+import os
+import six
+
+from django.utils.encoding import force_text
 from django.utils.text import slugify
 
 import boto3
@@ -7,7 +11,10 @@ from agreements.models import Issuer
 
 
 def s3_safe_key(path, prefix=''):
-    key = prefix + path.encode('utf-8')
+    if six.PY2:
+        key = prefix + force_text(path)
+    else:
+        key = prefix + path
     key = key.replace(' ', '_')
     key = key.replace('%', '')
     key = key.replace(';', '')
@@ -42,9 +49,13 @@ def save_agreement(agreements_zip, pdf_path, outfile,
     uri_hostname = 'https://files.consumerfinance.gov'
     s3_prefix = 'a/assets/credit-card-agreements/pdf/'
 
-    zipinfo = agreements_zip.getinfo(pdf_path)
+    if six.PY2:
+        zipinfo = agreements_zip.getinfo(pdf_path)
+    else:
+        zipinfo = agreements_zip.getinfo(force_text(pdf_path, 'cp437'))
+
     if windows:
-        path = pdf_path.decode('windows-1252')
+        path = force_text(pdf_path, 'windows-1252')
     else:
         path = pdf_path
     try:
--- a/cfgov/agreements/management/commands/_util.py
+++ b/cfgov/agreements/management/commands/_util.py
@@ -45,19 +45,17 @@ def get_issuer(name):
 
 
 def save_agreement(agreements_zip, pdf_path, outfile,
-                   windows=False, upload=False):
+                   upload=False):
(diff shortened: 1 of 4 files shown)


cfgov/agreements/management/commands/_util.py
```diff
@@ -42,9 +49,13 @@ def save_agreement(agreements_zip, pdf_path, outfile,
     uri_hostname = 'https://files.consumerfinance.gov'
     s3_prefix = 'a/assets/credit-card-agreements/pdf/'
 
-    zipinfo = agreements_zip.getinfo(pdf_path)
+    if six.PY2:
+        zipinfo = agreements_zip.getinfo(pdf_path)
+    else:
+        zipinfo = agreements_zip.getinfo(force_text(pdf_path, 'cp437'))
```
chosak: What is the context for the 'cp437' here?

cfgov/agreements/tests/test_legacy.py
```diff
@@ -132,13 +132,13 @@ def test_issuer_paging_frontend(self):
 
         path = reverse('issuer_search', kwargs={'issuer_slug': issuer.slug})
         resp = self.client.get(path)
-        self.assertTrue('page=2' in resp.content)
-        self.assertFalse('page=1' in resp.content)
+        self.assertTrue(b'page=2' in resp.content)
```
chosak: Proposal: instead of hassling with bytes, use Django's built-in test assertions:

```suggestion
        self.assertContains(resp, 'page=2')
        self.assertNotContains(resp, 'page=1')
```

cfgov/agreements/tests/test_legacy.py
```diff
@@ -132,13 +132,13 @@ def test_issuer_paging_frontend(self):
...
+        self.assertFalse(b'page=1' in resp.content)
 
         resp = self.client.get(path + '?page=2')
-        self.assertTrue('page=1' in resp.content)
-        self.assertFalse('page=2' in resp.content)
-        self.assertFalse('page=3' in resp.content)
+        self.assertTrue(b'page=1' in resp.content)
```
chosak: Same suggestion as above:

```suggestion
        self.assertContains(resp, 'page=1')
        self.assertNotContains(resp, 'page=2')
        self.assertNotContains(resp, 'page=3')
```

cfgov/agreements/management/commands/_util.py
willbarton: Will add a comment about this change. But in brief (or not?):

Unless a particular flag is set in the zip file, it will default to MS-DOS encoding, which is `cp437`, according to the [zip file spec](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT):

> D.1 The ZIP format has historically supported only the original IBM PC character 
encoding set, commonly referred to as IBM Code Page 437.  This limits storing 
file name characters to only those within the original MS-DOS range of values 
and does not properly support file names in other character encodings, or 
languages. To address this limitation, this specification will support the 
following change. 
> 
> D.2 If general purpose bit 11 is unset, the file name and comment should conform 
to the original ZIP character encoding.  If general purpose bit 11 is set, the 
filename and comment must support The Unicode Standard, Version 4.1.0 or 
greater using the character encoding form defined by the UTF-8 storage 
specification.  The Unicode Standard is published by the The Unicode
Consortium (www.unicode.org).  UTF-8 encoded data stored within ZIP files 
is expected to not include a byte order mark (BOM).

willbarton: I think I might finagle that change a little bit more, because I'm still not sure it makes sense with the cp1252 encoding right below it.

willbarton: I've refactored the file encoding handling. There are still a couple of places where I don't fully understand what Python 2 is doing, but once we move to Python 3 the encodings flow through the import in a sensible way.

PR #4610, opened 2018-10-24
Make legacy tests work on Python 3

willbarton:
This is a cleanup of map/filter generators and translating unittest `assertItemsEqual` to `assertCountEqual`.


cfgov/legacy/housing_counselor/cleaner.py
```diff
@@ -71,9 +71,9 @@ def title_case(s):
     )
 
     parts[0] = parts[0].title()
-    parts = map(
+    parts = list(map(
         lambda part: part.title() if part not in lower_case else part,
         parts
-    )
+    ))
```
chosak: Is this change needed as the resulting map `parts` is joined just below?

willbarton: Oh, no, I don't think so.

PR #4611, opened 2018-10-24
Add new page attribute `is_searchable` and limit external links query to those pages

richaagarwal:
## Overview

Fixes a bug where there are duplicate results in /admin/external-links (the new external links search) because pages that inherit from `CFGOVPage` (all pages) or e.g.  `AbstractFilterPage` are having their sidefoot and/or header searched at each level of inheritance. You can reproduce this bug by adding a link to the sidefoot or header of an activity log page and then searching for that link in http://localhost:8000/admin/external-links/. 

## Notes

- These changes also make more explicit which pages we’ve audited to include in the search.  Previously, we were searching pages like `RegulationLandingPage`, though we’ve done nothing to add its fields to the search index (`header` and `content` in this case).  It would’ve included results based on the `sidefoot` since that is inherited from CFGOVPage, but I believe partially searching like that would be confusing. 
- Though not all v1 page *types* have had `is_searchable` added to them, all v1 *pages* are searchable because they themselves are being searched, or a subclass or superclass is being searched.

### DIFF
--- a/cfgov/ask_cfpb/models/pages.py
+++ b/cfgov/ask_cfpb/models/pages.py
@@ -201,6 +201,8 @@ class AnswerLandingPage(LandingPage):
     ])
     objects = CFGOVPageManager()
 
+    is_searchable = False
+
     def get_context(self, request, *args, **kwargs):
         from ask_cfpb.models import Category, Audience
         context = super(AnswerLandingPage, self).get_context(request)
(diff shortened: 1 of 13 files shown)


richaagarwal:
This doesn't quite work as stated, and I'm closing this in favor of https://github.com/cfpb/cfgov-refresh/pull/4627

PR #4619, opened 2018-10-31
Nonresponsive base template: Remove unnecessary inclusion of jQuery/JS and CSS

anselmbradford:
## Removals

- Unused JS/CSS for jobs sign up.
- Unused images in the jobs pages.

## Changes

- Moves jquery out of `base_nonresponsive.html` and into the implementing templates that actually use jquery. Also, does the same with the legacy nemo `functions.js` file and for the choosen jquery plugin.

## Testing

1. Check for JS errors for missing `$` or `jQuery` in the following pages:
 - http://localhost:8000/jobs/technology-innovation-fellows/
 - http://localhost:8000/jobs/supervision/
 - http://localhost:8000/practitioner-resources/students/knowbeforeyouowe/
 - http://localhost:8000/practitioner-resources/students/
 - http://localhost:8000/practitioner-resources/students/helping-borrowers-find-ways-to-stay-afloat/
 - http://localhost:8000/find-a-housing-counselor/
 - http://localhost:8000/fair-lending/
 - http://localhost:8000/credit-cards/agreements/


cfgov/agreements/templates/agreements/index.html
```diff
@@ -20,8 +20,10 @@ <h1>Credit card agreement database</h1>
 
 <div id="ccagrhome">
     <p>
-        The CFPB maintains a database of credit card agreements from hundreds of card issuers.
-            Using the tool below, you can search for an agreement by the name of the issuer.
+        The CFPB maintains a database
```
anselmbradford: This is just fixing the line length.

cfgov/templates/jobmanager/supervision.html
```diff
@@ -13,7 +17,7 @@
 {% block breadcrumbs %}
 	<li><a href="/" class="internal-link">Home</a></li>
```
wpears: Might as well match the indent of the lines below

cfgov/legacy/templates/front/base_nonresponsive.html
```diff
@@ -61,8 +61,9 @@
     <link rel="apple-touch-icon" size="114x114" href="{% static 'nemo/_/img/apple-touch-icon-114x114-precomposed.png' %}">
     <link rel="apple-touch-icon" size="144x144" href="{% static 'nemo/_/img/apple-touch-icon-144x144-precomposed.png' %}">
 
+    {% block jquery %}
```
wpears: Can this empty block be removed?

chosak:
[It looks like](https://github.com/search?q=user%3Acfpb+base_nonresponsive.html&type=Code) the [django-college-costs-comparison](https://github.com/cfpb/django-college-costs-comparison) satellite app also depends on this base template. Have you checked to make sure that it isn't depending in some way on the JS/CSS that's being removed here?

cfgov/legacy/templates/front/base_nonresponsive.html
anselmbradford: I think it's needed to tell the templates that extend this template where to place jquery. I would have placed jquery in the `app_js` block, but it was originally in the head so I aimed to keep it in the same place.

cfgov/templates/jobmanager/supervision.html
anselmbradford: Looking at some of the other ones, I removed the indents in the block, because they're not used elsewhere. Has been updated!

anselmbradford:
> It looks like the django-college-costs-comparison satellite app also depends on this base template. Have you checked to make sure that it isn't depending in some way on the JS/CSS that's being removed here?

Yes! PR https://github.com/cfpb/django-college-costs-comparison/pull/18

cfgov/legacy/templates/front/base_nonresponsive.html
wpears: Ahh [indeed](https://docs.djangoproject.com/en/1.7/topics/templates/#id1)!

PR #4622, opened 2018-11-02
Convert RelatedPosts block to use include_block

chosak:
This change modifies the way that the RelatedPosts StreamField block is included and rendered in page sidebars. Instead of rendering it using a macro (as we do most of our blocks), this change renders it using the [standard Wagtail `{% include_block %}` method](http://docs.wagtail.io/en/v2.3/topics/streamfield.html#template-rendering).

This change moves the related posts logic from the page class to the block. This consolidates all related posts logic into a single place. Python tests have also been moved accordingly.

To test this, you can visit these two pages using a local server running against a production dump:

http://localhost:8000/data-research/credit-card-data/
http://localhost:8000/about-us/blog/askcfpb-your-questions-about-credit-reports-and-scores/

These two pages represent instances where this block is included in a sidebar-breakout and a streamfield-sidefoot.

This commit makes a minor change to some of the logic in the related posts template; specifically, even if there are no posts to be rendered, it still applies the `block__flush-top` class to the block div. In practice this should never happen, as all pages should have some related posts. Even if it did, using that class on the div seems reasonable, even if the block is only displaying the "View more" link.

See also GHE/platform#378 for more background on why this change is beneficial. If this change seems good to folks, we could consider modifying other StreamField blocks to be included in a similar way, which would reduce some of the complexity of our templates.

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/templates/streamfield-sidefoot.html
+++ b/cfgov/jinja2/v1/_includes/templates/streamfield-sidefoot.html
@@ -1,4 +1,3 @@
-{% import 'molecules/related-posts.html' as related_posts with context %}
 {% import 'molecules/related-metadata.html' as related_metadata with context %}
 {% import 'molecules/social-media.html' as social_media with context %}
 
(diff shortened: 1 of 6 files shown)


cfgov/jinja2/v1/_includes/templates/streamfield-sidefoot.html
```diff
@@ -1,23 +1,18 @@
...
     {% for block in streamfield %}
         {% if 'related_posts' in block.block_type %}
-            {% set posts = page.related_posts(block) %}
-            {% if posts %}
-                <div class="block {{ 'block__flush-top' if loop.index == 1 else '' -}}">
-                    {{ related_posts.render(posts, block, half_width) }}
-                </div>
-            {% endif %}
+            <div class="block{{ ' block__flush-top' if loop.first else '' }}">
+            {% include_block block %}
```
Scotchester, line 9: Wish I had noticed when I reviewed this a year and a half ago that the closing `</div>` went missing here. We've had improperly nested sidebar tags on pages with Related Posts ever since 😞

PR #4627, opened 2018-11-07
Remove duplicates from external links search results

richaagarwal:
This is a more simplified version of addressing what https://github.com/cfpb/cfgov-refresh/pull/4611 was attempting to address, by removing duplicates at the time of the search. 

See PR above and GHE/CFGOV/enhanced-cms/issues/39 for more context.

In addition to removing the duplicates:
  - We've updated the help text (thank you @marteki and @nataliafitzgerald)
  - Added tests (if you run those same tests off master currently, they will fail, as expected)

### DIFF
--- a/cfgov/search/views.py
+++ b/cfgov/search/views.py
@@ -29,6 +29,8 @@ class SearchView(View):
 
         for cls in get_page_models():
             pages += list(cls.objects.search(url))
+
+        pages = self.remove_duplicates(pages)
         pages = sorted(pages, key=lambda k: k.title)
 
         contacts = list(
@@ -52,6 +54,14 @@ class SearchView(View):
             'num_page_results': num_page_results,
             'num_snippet_results': num_snippet_results,
         })
+
+    @staticmethod
+    def remove_duplicates(pages):
+        seen = set()
+        for page in pages:
+            if page.pk not in seen:
+                seen.add(page.pk)
+                yield page
 
 
 def results_view(request):
--- a/cfgov/search/templates/search/external_links.html
+++ b/cfgov/search/templates/search/external_links.html
@@ -25,7 +25,10 @@
 
     {% block content_main %}
     <div class="nice-padding external-links">
-        <p class="external-links-intro">Enter an external URL to see where it occurs on our site. If a link appears within more than one section of a given page, such as the body and sidebar, the results will list that page multiple times--once for each occurrence.</p>
+        <p class="external-links-intro">Enter an external link to see a list of
+          pages and snippets that contain that URL. Each page result will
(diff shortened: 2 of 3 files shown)


cfgov/search/views.py
```diff
@@ -54,6 +56,14 @@ def post(self, request):
         })
 
 
+def remove_duplicates(pages):
+        seen = set()
```
chosak: Can this please be made indented by 4 spaces to match consistency with the rest of the file?

marteki:
@richaagarwal - @nataliafitzgerald and I updated the helper text. I just pushed it up to this branch for you.

richaagarwal:
Thank you @marteki!  I have incorporated your changes and @chosak's feedback in my latest commits.

richaagarwal:
@chosak Thank you for the feedback!  I added a couple tests and changed the function to a static method.   

I also added an additional filter on the search since it seems unnecessary to search all pages -- just creatable ones -- in this commit: https://github.com/cfpb/cfgov-refresh/pull/4627/commits/479a7099f61561ccec479a35ac18f892795d3194.  Regardless of that, the duplicates are removed, but it seemed to make sense to narrow the scope of the search for performance.  

If you could take a look at the tests I added and that change I'd appreciate one final review!

chosak:
@richaagarwal 👍 tests look good! I have one minor suggestion that doesn't impact the testing logic at all but allows for a shorter and simpler test: instead of constructing the page and then using the  Wagtail StreamValue class to set the StreamField, set it as a constructor argument. See [this diff](https://github.com/cfpb/cfgov-refresh/compare/remove-dupes...chosak:remove-dupes-streamfield-content#diff-d6d072ad7619c97c3ffa408c81da4163L62) I made on a branch to demonstrate. To my eye this is a little more readable and straightforward about what the content of the StreamField is, and it also avoids using the internal StreamValue creation approach. I know we do it how you've done it here elsewhere in the codebase, though.

Regarding filtering out pages that aren't createable: do you have a sense for the performance benefit? The concern I'd have (especially if this goes undocumented) is that we might encounter a situation in future where we purposefully set `is_creatable` to false on a page type for the purposes of preventing its creation in the UI. Say we have a singleton page type (like the recent TDP index page) and we only ever want one of these ever, anywhere. So we might use a script to prevent it, and then use `is_creatable` to hide it from the UI. If we did that, those pages would disappear from the search results (or never appear if we did that from the beginning) and it wouldn't be clear to the user why that was.

richaagarwal:
@chosak Thanks! I've incorporated your updates to the tests; I agree it's more readable that way.  

And good point on `is_creatable`, I've removed that logic for now and can always revisit later if there's performance reasons to do so.

PR #4635, opened 2018-11-16
Convert to ES6 import/export and upgrade babel and cfpb-chart-builder to latest.

anselmbradford:
## Changes

- Converts requires to ES6 import/exports (except in assets for some individual apps).
- Updates to cfpb-chart-builder 5.0.0, which upgrades highcharts dependency to 6.x.
- Updates babel to latest (including making babel-config not invisible).

## Testing

1. `./frontend.sh`
2. `gulp test` should pass.
3. Check pages and look for any `exports` error in dev console.
4. Pay particular attention to [Mortgage Performance Trends](https://www.consumerfinance.gov/data-research/mortgage-performance-trends/) and [CCT](https://www.consumerfinance.gov/data-research/consumer-credit-trends/), as this is a major update for those.

Should be checked in supported browsers.

## Note

- Charts likely won't render in IE in VirtualBox because of cross domain errors loading the data. Sauce labs connect should be used instead.

### DIFF
--- a/cfgov/unprocessed/js/modules/behavior/FlyoutMenu.js
+++ b/cfgov/unprocessed/js/modules/behavior/FlyoutMenu.js
@@ -1,9 +1,14 @@
 // Required modules.
-const BaseTransition = require( '../../modules/transition/BaseTransition' );
-const behavior = require( '../../modules/util/behavior' );
-const breakpointState = require( '../../modules/util/breakpoint-state' );
-const EventObserver = require( '../../modules/util/EventObserver' );
-const standardType = require( '../../modules/util/standard-type' );
+import BaseTransition from '../../modules/transition/BaseTransition';
+import { checkBehaviorDom } from '../../modules/util/behavior';
+import * as breakpointState from '../../modules/util/breakpoint-state';
+import EventObserver from '../../modules/util/EventObserver';
+import {
+  BEHAVIOR_PREFIX,
+  JS_HOOK,
+  noopFunct,
+  UNDEFINED
+} from '../../modules/util/standard-type';
 
 /**
  * FlyoutMenu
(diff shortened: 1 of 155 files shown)


package.json
```diff
@@ -66,8 +66,11 @@
     "xdr": "0.5.3"
   },
   "devDependencies": {
+    "babel-core": "7.0.0-bridge.0",
```
anselmbradford: This is needed for babel-jest in addition to the version in dependencies, see https://jestjs.io/docs/en/getting-started.html#using-babel

anselmbradford:
Codecov can be ignored I think. It appears to be triggering on the change from requires to imports.

anselmbradford:
nudgy
Still looking for a review

cfgov/unprocessed/js/modules/behavior/FlyoutMenu.js
```diff
@@ -1,9 +1,14 @@
 // Required modules.
-const BaseTransition = require( '../../modules/transition/BaseTransition' );
-const behavior = require( '../../modules/util/behavior' );
-const breakpointState = require( '../../modules/util/breakpoint-state' );
-const EventObserver = require( '../../modules/util/EventObserver' );
-const standardType = require( '../../modules/util/standard-type' );
+import BaseTransition from '../../modules/transition/BaseTransition';
+import { checkBehaviorDom } from '../../modules/util/behavior';
+import * as breakpointState from '../../modules/util/breakpoint-state';
```
wpears: Mostly a nit, but it would be more concise if throughout this PR to move from collecting exports in `{ ... }` then importing them with `import * as...` to exporting objects as `default`, eg:
```
export default {
  get,
  isInDesktop
}
```

Then importing with `import breakpointState from ...`

anselmbradford: Unfortunately default exports break IE10:

screen shot 2018-11-27 at 9 51 29 am

We may need a polyfill. Dirty little secret: we don't automatically polyfill with babel-polyfill currently as it can be overly aggressive.

wpears: Interesting! Thanks for the info, definitely something to look into further.

PR #4638, opened 2018-11-19
Change map layout to use featured content module

anselmbradford:
## Changes

Housing counselor: 
- Updates zipcode input form and map to use featured content module.
- Removes hero block in jinja template.
- Exports utility methods from hud.js and adds unit test coverage for those methods. Also adjusts their logic to return consistent types.

## Testing

1. `gulp clean && gulp build`
1. Visit http://localhost:8000/find-a-housing-counselor/ and resize window to see how zipcode form and map stack.

## Screenshots

screen shot 2018-11-19 at 10 59 35 am
screen shot 2018-11-19 at 10 59 45 am


anselmbradford:
Codecov can be ignored. This JS has zero coverage before.

PR #4639, opened 2018-11-20
Add the Docker for Mac internal gateway IP

willbarton:
~~This PR adds the Docker for Mac gateway IP to our list of `INTERNAL_IPS` in the local settings so that when Django Debug Toolbar is enabled it a browser on the local Mac is recognized as being local.~~

~~I've also added comments around that setting to describe why each IP is there.~~

This PR adds a `.env` variable for the `INTERNAL_IPS` setting that lets us configure that setting based on our local Docker configuration to enable the Django Debug Toolbar.


cfgov/cfgov/settings/local.py
```diff
@@ -40,7 +40,10 @@
 if os.environ.get('ENABLE_DEBUG_TOOLBAR'):
     INSTALLED_APPS += ('debug_toolbar',)
 
-    INTERNAL_IPS = ('127.0.0.1', '10.0.2.2')
+    # Local running instances: 127.0.0.1
+    # VirtualBox docker-machine instances: 10.0.2.2
+    # Docker for Mac instances: 172.18.0.1
+    INTERNAL_IPS = ('127.0.0.1', '10.0.2.2', '172.18.0.1')
```
chosak: Is 172.18.0.1 guaranteed to be the host IP? See the [docs](https://docs.docker.com/docker-for-mac/networking/#use-cases-and-workarounds):

> The host has a changing IP address (or none if you have no network access). From 18.03 onwards our recommendation is to connect to the special DNS name host.docker.internal, which resolves to the internal IP address used by the host.

willbarton: That's a good point, although neither `host.docker.internal` nor `gateway.docker.internal` point to the correct IP. 

It seems like the correct IP is listed by `docker inspect cfgov-refresh_python_1` under NetworkSettings/Networks/cfgov-refresh_default/Gateway.

willbarton:
@chosak @Scotchester I've changed the way this PR is implemented. It now looks for an `INTERNAL_IP` variable in the environment and will use that to populate the `INTERNAL_IPS` tuple. I've also added the variable to the `.env_SAMPLE` file with a note about how to discover what it should be using `docker inspect`.

If that variable is not set, `INTERNAL_IPS` defaults to `127.0.0.1`. 

I've also removed the old hard-coded VirtualBox-based docker IP. It should be discoverable the same way.

cfgov/cfgov/settings/local.py
willbarton: Resolving this conversation. See comment below.

cfgov/cfgov/settings/local.py
```diff
@@ -40,7 +40,7 @@
 if os.environ.get('ENABLE_DEBUG_TOOLBAR'):
     INSTALLED_APPS += ('debug_toolbar',)
 
-    INTERNAL_IPS = ('127.0.0.1', '10.0.2.2')
+    INTERNAL_IPS = (os.environ.get('INTERNAL_IP', '127.0.0.1'), )
```
chosak: Not being super familiar with the Docker setup, if you were to provide an `INTERNAL_IP` variable here, and thus `INTERNAL_IPS` would get set to **only** that variable's value (not including `'127.0.0.1'`), would it break any functionality when accessed via localhost from within the Python container?

chosak: Not being super familiar with the Docker setup, if you were to provide an `INTERNAL_IP` variable here, and thus `INTERNAL_IPS` would get set to **only** that variable's value (not including `'127.0.0.1'`), would it break any functionality when accessed via localhost from within the Python container?

.env_SAMPLE
```diff
@@ -197,3 +197,10 @@ fi
 
 # Enable the debug toolbar.
 # export ENABLE_DEBUG_TOOLBAR=True
+
+# Set an interal IP for the debug toolbar
```
Scotchester:
```suggestion
# Use an alternate internal IP for the Django Debug Toolbar.
```

.env_SAMPLE
```diff
@@ -197,3 +197,10 @@ fi
...
+# This is necessary for it to work with docker
```
Scotchester:
```suggestion
# This is necessary for it to work with our Docker dev setup.
```

.env_SAMPLE
```diff
@@ -197,3 +197,10 @@ fi
...
+# The internal IP will be listed in the output of this command:
+#   docker inspect cfgov-refresh_python_1
+# Under NetworkSettings/Networks/cfgov-refresh_default/Gateway
```
Scotchester:
```suggestion
# (under NetworkSettings/Networks/cfgov-refresh_default/Gateway)
```

.env_SAMPLE
```diff
@@ -197,3 +197,10 @@ fi
 
 # Enable the debug toolbar.
```
Scotchester:
```suggestion
# Enable the Django Debug Toolbar.
```

cfgov/cfgov/settings/local.py
willbarton: Is there any functionality to access from within the Python container? I don't think so.

.env_SAMPLE
Scotchester: No, I didn't mean that second thing. You're probably right that it's implied enough, at least until such potential time as we use Docker in production.

Scotchester: Revised the suggested edit.

PR #4640, opened 2018-11-20
Finding housing counselor: Fix results behavior

anselmbradford:
Results list had hardcoded results count and non-functional PDF and print buttons. It also didn't work with invalid zipcode values.

Error message should close out [GHE]/CFGOV/platform/issues/2580

## Changes

- Make results list summary text dynamic based on result count and zip code.
- Update print button JS to remove jQuery.
- Update hud.css to hud.less and add less rule for abbr tag wrapping PDF.
- Wrap long paragraphs.
- Remove unused CSS.
- Updates "5-digit" copy to "five-digit".
- Updates "zip code" to "ZIP code" per our style guides.

## Testing

1. `gulp clean && gulp build`
2. Visit http://localhost:8000/find-a-housing-counselor/
3. Enter a valid ZIP code.
4. Click "Print list" button and see that print dialog shows up.
5. Click "Save list as PDF" link and see that PDF saves.
6. Enter an invalid ZIP code.
7. See that error notification shows up and results list does not appear.

## Screenshots

screen shot 2018-11-20 at 2 57 01 pm

---

screen shot 2018-11-20 at 2 57 15 pm

### DIFF
--- a/cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
+++ b/cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
@@ -119,10 +119,7 @@
                         </section>
 
                     </div>
-                    {% if zipcode %}
-                    {% if not zipcode_valid %}
-
-                    {% else %}
+                    {% if zipcode and zipcode_valid %}
                     <div class="block">
                         <div class="results-header">
                             <ul class="m-list
@@ -223,7 +220,6 @@
                         </table>
                     </div>
 
-                    {% endif %}
                     {% endif %}
 
                     <div class="block">
--- a/cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
+++ b/cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
@@ -21,7 +21,30 @@
                     <div class="block block__flush-top block__flush-top">
                         <h1>Find a housing counselor</h1>
                         <div class="rich-text">
-                            <p>Housing counselors throughout the country can provide advice on buying a home, renting, defaults, foreclosures, and credit issues. Using the search box below, you can find one near you. The counseling agencies on this list are approved by the U.S. Department of Housing and Urban Development (HUD) and they can offer independent advice about whether a particular set of mortgage loan terms is a good fit based on your objectives and circumstances, often at little or no cost to you. This list will show you several approved agencies in your area. There is also a <a class="a-link a-link__icon" href="https://apps.hud.gov/offices/hsg/sfh/hcc/hcs.cfm"><span class="a-link_text">list of nationwide HUD-approved counseling agencies</span> {{ svg_icon( 'external-link' ) }}</a>.</p>
+                            <p>
+                                Housing counselors throughout the country can
+                                provide advice on buying a home, renting,
+                                defaults, foreclosures, and credit issues.
+                                Using the search box below,
+                                you can find one near you.
+                                The counseling agencies on this list are
+                                approved by the U.S. Department of Housing
+                                and Urban Development (HUD) and they can
+                                offer independent advice about whether a
+                                particular set of mortgage loan terms is a
+                                good fit based on your objectives and
+                                circumstances, often at little
+                                or no cost to you. This list will show you
+                                several approved agencies in your area.
+                                There is also a
+                                <a class="a-link a-link__icon"
+                                   href="https://apps.hud.gov/offices/hsg/sfh/hcc/hcs.cfm">
+                                   <span class="a-link_text">
+                                       list of nationwide
+                                       HUD-approved counseling agencies
+                                   </span>
+                                   {{ svg_icon( 'external-link' ) }}</a>.
+                            </p>
                         </div>
                     </div>
                     <div class="block block__border">
--- a/cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
+++ b/cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
@@ -226,7 +226,38 @@
                         <div class="o-full-width-text-group">
                             <div class="m-full-width-text">
                                 <h3>Paperwork Reduction Act statement</h3>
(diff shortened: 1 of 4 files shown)


cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
```diff
@@ -75,17 +97,42 @@ <h1>Find a housing counselor</h1>
...
-                                </button>
-                            </div>
+                            <ul class="m-list
+                                       m-list__horizontal
+                                       hud_hca_api_results_actions">
+                                <li class="m-list_item">
+                                    <button class="a-btn a-btn__link"
+                                       id="hud_print-page-link">
+                                        Print list
+                                        {{ svg_icon( 'print' ) }}
```
niqjohnson: This icon isn't in the current version of the tool and I'm not sure we have a super consistent way we apply link button icons like this across the site, so I'd vote for no icon for now.

cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
```diff
@@ -75,17 +97,42 @@ <h1>Find a housing counselor</h1>
...
+                                    <a class="a-btn a-btn__link"
+                                       id="generate-pdf-link"
+                                       href="{{ pdf_url }}"
+                                       target="_blank"
+                                       rel="noopener noreferrer">
+                                        Save list as
+                                        <abbr title="Portable Document Format">
+                                          PDF
+                                        </abbr>
+                                        {{ svg_icon( 'save' ) }}
```
niqjohnson: This icon isn't in the current version of the tool and I'm not sure we have a super consistent way we apply link button icons like this across the site, so I'd vote for no icon for now.

cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
```diff
@@ -31,17 +31,39 @@ <h1>Find a housing counselor</h1>
...
+                                            {% if zipcode and not zipcode_valid %}
+                                            <div class="u-mb15">
+                                                <div class="m-notification
+                                                            m-notification__default
+                                                            m-notification__error
+                                                            m-notification__visible">
+                                                    {{ svg_icon( 'warning-round' ) }}
+                                                    <div class="m-notification_content" role="alert">
+                                                        <div class="h4 m-notification_message">
+                                                            You have entered an invalid ZIP code.
```
niqjohnson: This bit of error text in the current version of the tool reads "Sorry, you have entered an invalid zip code." I think capitalizing ZIP in the new version makes sense (that seems like an obvious error in the current version), but should we leave "Sorry" in there?

cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
```diff
@@ -31,17 +31,39 @@ <h1>Find a housing counselor</h1>
...
+                                                        </div>
+                                                        <div class="m-notification_explanation">
+                                                            Enter a valid five-digit ZIP code.
```
niqjohnson: This bit of error text in the current version of the tool reads "Please enter a valid five-digit ZIP code above." Getting rid of the "above" in the new version makes sense since the form is no longer above the error, but should we leave the "Please" in there?

cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
anselmbradford: Yeah I checked the `zip` with Sarah and it should be capitalized. I removed the "Sorry," and the "Please" because it makes the error notification very wordy. See this versus this:

screen shot 2018-11-20 at 3 30 35 pm

screen shot 2018-11-20 at 3 30 46 pm

Also, what are we apologizing for? The user entered the zip wrong! If anything, the "please" probably makes more sense to retain, but I'd vote we aim for brevity in notifications.

cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
anselmbradford: See note in other comment. I vote for brevity, but you have the final call on this one.

cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
anselmbradford: [It actually is](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/housing_counselor/templates/housing_counselor/index.html#L148) but isn't rendered. I'm guessing at some point it broke and no one noticed. However, let me know if in this modern incarnation it SHOULD have an icon?

cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
anselmbradford: Same as print, it is, but is broken on the current site.

cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
niqjohnson: Interesting! This does seem like something that used to show up and then silently broke, so I say let's restore it. The only small change is that the icons should not be underlined (see the external link icon for an example).

cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
niqjohnson: As above: Interesting! This does seem like something that used to show up and then silently broke, so I say let's restore it. Couple of small changes here: the icons should not be underlined (see the external link icon for an example) and I think we use the download icon instead of the save icon nowadays.

cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
niqjohnson: I'm not sure of the history of this content and thus am loathe to change it too much. So, for now, let's go with correcting "zip" to "ZIP" but leaving everything else the same: "Sorry, you have entered an invalid ZIP code."

cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
anselmbradford: Fixed!

cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
anselmbradford: Fixed!

cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
anselmbradford: Fixed!

anselmbradford:
@niqjohnson Updated screenshots:

screen shot 2018-11-20 at 5 32 03 pm

screen shot 2018-11-20 at 5 32 14 pm

PR #4643, opened 2018-11-21
Allow more flexible date formats in Filterable List validation

schbetsy:
This is part of the UI Improvements project, which is improving the user experience for filterable lists.

## Changes

- Allow single-digit month and day in filterable list date input. e.g. 10/2/2018, 1/31/2009
- Allow two-digit year in date inputs. e.g. 10/10/09

## Additions

- Add a few front-end tests to demonstrate which date formats are allowed

## Testing

1. With this branch running, visit http://localhost:8000/about-us/newsroom/
2. Try inputting different dates into the Date Rage "From:" and "To:" fields.
3. See that the following date formats all allow you to submit, and get a valid response:
     - 10/25/16 (two-digit year)
     - 9/1/16 (single-digit month and day)
     - 10/25/2016 (four-digit year)
     - 9/1/2016
     - 10/2016 (just month and year)
     - 7/2017
     - 10/16, 4/18 (month and two-digit year)
     - 2016

## Checklist

## Testing checklist

Should not change any visual behavior (except showing an error box under fewer conditions)


cfgov/unprocessed/js/modules/util/validators.js
```diff
@@ -21,8 +21,9 @@ const typeCheckers = require( '../../modules/util/type-checkers' );
  */
 function date( field, currentStatus ) {
   const status = currentStatus || {};
+  // Date regex matches date patterns allowed in cfgov/v1/forms.py FilterableDateField
```
anselmbradford: We're certainly not consistent with this, but it probably makes sense to link to a regex test bed for this regex in the code comment, like is done at https://github.com/cfpb/cfgov-refresh/blob/master/config/webpack-config.js#L24
That'll make it easier to update in the future, if needed. Totally not necessary, but an idea if you feel like it.

cfgov/unprocessed/js/modules/util/validators.js
```diff
@@ -21,8 +21,9 @@ const typeCheckers = require( '../../modules/util/type-checkers' );
...
   const dateRegex =
-    /^\d{2}$|^\d{4}$|^\d{2}\/(?:\d{4}|\d{2})$|^\d{2}\/\d{2}\/\d{4}$/;
+    /^\d{2}$|^\d{4}$|^(?:\d{1}|\d{2})\/(?:\d{4}|\d{2})$|^(?:\d{1}|\d{2})\/(?:\d{1}|\d{2})\/\d{4}$/;
```
chosak: - The first option `^\d{2}$` matches a solo 2-digit year input string like `16` but not a 4-digit year like `2016`. Should it be `^\d{4}$` instead?
- The second option `^(?:\d{1}|\d{2})\/(?:\d{4}|\d{2})$` looks good and does match `1/16`, `10/16`, `1/2016`, `10/2016`.
- The third option `^(?:\d{1}|\d{2})\/(?:\d{1}|\d{2})\/\d{4}$` matches m/d/y with 4-digit years like `1/10/2016` but doesn't seem to match 2-digit years like `10/1/16` as documented [here](https://github.com/cfpb/cfgov-refresh/pull/4643/files#diff-a20999810035f4df9b089b5ce85e66b6R22) as a valid Python option. Should it be `^(?:\d{1}|\d{2})\/(?:\d{1}|\d{2})\/(?:\d{4}|\d{2})$` instead?

cfgov/unprocessed/js/modules/util/validators.js
schbetsy: Good idea. I've added that now.

cfgov/unprocessed/js/modules/util/validators.js
schbetsy: - The first option was extraneous. Thanks for calling that out.
- Good call, I forgot to add the two-digit option to the third option. I've added it now.

anselmbradford:
I get "invalid data" with `10/25/16`

screen shot 2018-11-21 at 3 20 33 pm

schbetsy:
@anselmbradford Did you pull the latest version of this branch? I believe I fixed that bug in 6e9f2c4.

schbetsy:
I have to run `./frontend.sh` whenever I want to bring in new changes. Does that make it work?

anselmbradford:
> I have to run ./frontend.sh whenever I want to bring in changes to javascript. Does that make it work for you?

It's still happening with `./frontend.sh`

PR #4645, opened 2018-11-21
Makes more than one skip-link per page possible

anselmbradford:
Find a housing counselor page has two skip links. The default one in the header and one to skip over the map to the list of results. This PR makes the skip link CSS tied to a class instead of an ID so more than one skip link can be added to a page.

## Changes

- Converts `skip-link` ID to class.
- Adds skip link to find a housing counselor page for skipping over the map.
- Adds 10px margin to skip link.

## Testing

1. Make sure `HUD_TOOL_IMPROVEMENTS` is set to boolean true.
1. Visit http://localhost:8000/find-a-housing-counselor/ and tab through the page to show the "Skip to results" skip link. Shift+tab to move back to the "Skip to main content" skip link.
1. http://localhost:8000/paying-for-college/manage-your-college-money/ and tab through the page to show the "Skip to main content" skip link.

## Screenshots

screen shot 2018-11-21 at 3 02 12 pm

screen shot 2018-11-21 at 3 02 19 pm


cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -34,7 +34,9 @@
 {# Overlay for the page. Used for the mobile mega menu. #}
 <div class="a-overlay u-hidden"></div>
 
-<a href="#main" id="skip-nav">Skip to main content</a>
+<div class="skip-nav">
+    <a href="#main" class="skip-nav_link">Skip to main content</a>
+</div>
```
Scotchester: Why convert this to `div > a` instead of just leaving it as an `a`?

cfgov/unprocessed/apps/analytics-gtm/js/accessibility-interactions-listeners.js
```diff
@@ -5,8 +5,8 @@ import {
 
 const AccessibleInteractionsListeners = ( () => {
 
-  // Listen for focus on the "skip link" button.
-  const skipLinkButton = document.querySelector( '#skip-nav' );
+  // Listen for focus on the "skip link" button. This will grab the top link.
```
Scotchester: Do we not want this to be able to handle additional skip links, too?

cfgov/unprocessed/css/skip-nav.less
```diff
@@ -9,26 +9,29 @@
...
-    top: 0;
-    outline: 0;
-    transition: transform .1s ease, background .2s linear;
+      top: 10px;
+      left: 10px;
+      outline: 0;
+      transition: transform .1s ease, background .2s linear;
+    }
+  }
 }
```
Scotchester: Please maintain the four-space indentation in this file.

cfgov/v1/jinja2/v1/header.html
```diff
@@ -13,7 +13,9 @@
 {# Skip Link to skip to main content #}
 {# TODO: When projects standardize on #main,
    remove this and empty after-header div #}
-<a href="#after-nav" id="skip-nav">Skip to main content</a>
+<div class="skip-nav">
+    <a href="#after-nav" class="skip-nav_link">Skip to main content</a>
+</div>
```
Scotchester: We should probably pull this out into its own include file so that we're not duplicating this markup here and in `base.html`, but it's fine for now.

cfgov/unprocessed/apps/analytics-gtm/js/accessibility-interactions-listeners.js
anselmbradford: The purpose of this script is to see whether someone tabbed into the page from the start, so it should be fine triggering on just the first skip link. If we want both we need to set up an tag for it in google tag manager. I'll check with analytics, but it wouldn't be needed for this PR and could be its own thing, if analytics wants it.

Scotchester: Cool. Might be worth tweaking the comment slightly to make that more clear.

cfgov/unprocessed/css/skip-nav.less
anselmbradford: Fixed

cfgov/unprocessed/apps/analytics-gtm/js/accessibility-interactions-listeners.js
anselmbradford: Done

cfgov/jinja2/v1/_layouts/base.html
anselmbradford: As discussed in chat, this is because the skip link is absolutely positioned, but needs a container.

Added a modifier class and increased the margin, so now it looks like:

screen shot 2018-11-26 at 1 06 41 pm

screen shot 2018-11-26 at 1 06 36 pm

cfgov/v1/jinja2/v1/header.html
anselmbradford: Agreed. For futureland

cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
```diff
@@ -104,6 +104,17 @@ <h1>Find a housing counselor</h1>
                                         </div>
                                     </div>
                                 </form>
+
+                                {% if zipcode and zipcode_valid %}
+                                <div class="skip-nav">
+                                    <a class="skip-nav_link
+                                              skip-nav_link__flush-left"
+                                       href="#hud_results-list_container">
+                                       Skip to results
```
Scotchester: I think maybe this should be reworded. Might make the user wonder what they're skipping. "Skip map of results" maybe? /cc @niqjohnson

PR #4650, opened 2018-11-27
CCDB version bump

AdamZarger:
Updating the version of `ccdb5-api`

When a user exported complaints from Complaint Search it used an Elasticsearch plugin called dataformat that occasionally caused the Elasticsearch cluster to crash under heavy load. The new version of `ccdb5-api` uses a Python script to generate CSV and JSON export files in order to take that load off of Elasticsearch. More information can be found [here](https://github.com/cfpb/ccdb5-api/pull/81)

## Changes

- Updated version of the `ccdb5-api` optional requirement

## Checklist

### Browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


AdamZarger:
@chosak When we were testing this change (prior to this PR) we ran into an issue where the import step was failing due to not being able to import `certs` from `requests`: [tavis output here](https://travis-ci.org/cfpb/ccdb5-api/builds/452516755?utm_source=github_status&utm_medium=notification). We updated to the requests version to 2.20.1 and that fixed the import issue.

We definitely want to conform with the modernization effort and change this requirement back to a range if possible. I can go back and attempt the range again to see if it still throws the same error on import. If it does throw that error, would you be available to help troubleshoot it?

chosak:
@AdamZarger absolutely. I see this error message in your Travis output:

> RequestsDependencyWarning: urllib3 (1.24.1) or chardet (3.0.4) doesn't match a supported version!

Travis is trying to install `urllib3==1.24.1` even though requests 2.19.1 [requires `urllib3>=1.21.1,<1.24`](https://github.com/requests/requests/blob/v2.19.1/setup.py#L54) (`chardet==3.0.4` is fine as 2.19.1 [requires `>=3.0.2,<3.1.0`](https://github.com/requests/requests/blob/v2.19.1/setup.py#L52)). I wonder where that `1.24.1` requirement is coming from?

AdamZarger:
@chosak Hm, that's a good question. `ccdb5-api` was using urllib to form the URL that we sent to the dataformat plugin in Elasticsearch. Since this PR moves to a new export process, I went ahead and tried to remove the references to urllib left in the requirements. Unfortunately it still looks like it's throwing the same error? I'm not sure where the urllib reference is coming from. Here's the [PR where I attempted the changes](https://github.com/cfpb/ccdb5-api/pull/83) and the [travis output](https://travis-ci.org/cfpb/ccdb5-api/builds/460338711?utm_source=github_status&utm_medium=notification)

Repeating my edit from above comment: I went ahead and tested the range version for `requests` ([See Travis results here](https://github.com/cfpb/ccdb5-api/pull/83)). It looks like the previous range of `>=2.18.4,<2.20` causes the `certs` error I mentioned, but if we update the range to include the `.1` (ie, `>=2.18.4,<2.20.1`) then the error goes away. Could we make this update to the version range of requests to satisfy the modernization effort?

chosak:
Sorry that I missed your edit the first time! The problem with just expanding the range in the ccdb5-api repo is that in practice, when cf.gov is deployed, it will use the pinned version in cfgov-refresh requirements anyway. So the server will actually have 2.19.1, regardless of what the range is in the ccdb5-api setup.py file. And if there is something in ccdb5-api that doesn't work with 2.19.1, it will break.

I will try to reproduce that Travis failure locally to see if I can figure out where that urllib3 requirement is coming from; that seems to be the root of the problem.

Somewhat relatedly, I also opened https://github.com/cfpb/ccdb5-api/pull/84 which addresses another recent incompatibility introduced in cfgov-refresh. It'd be nice if that could go into the next ccdb5-api release as well.

Scotchester:
Please note that we need to be updating requests everywhere to 2.20.0: https://GHE/CFGOV/platform/issues/3104

AdamZarger:
After working with @chosak on the dependencies ([see PR here](https://github.com/cfpb/ccdb5-api/pull/83)), we have ironed out a dependency order that satisfies the app's requirements as well as the modernization standards. We additionally added proper unit test coverage and updated the version of Django Flag ([see PR here](https://github.com/cfpb/ccdb5-api/pull/84))

I've created a new release with these fixes and updated the version in this PR appropriately. @Scotchester Are there any further changes needed for the requests package update you mentioned?

Scotchester:
@AdamZarger @chosak I'm a little confused by cfpb/ccdb-api#83. Can we not work out the Requests situation so that we're using >= 2.20.1 both in ccdb-api and cfgov-refresh? Is there some kind of incompatibility that would take a while to work out?

AdamZarger:
Closing this PR in favor of https://github.com/cfpb/cfgov-refresh/pull/4662

PR #4655, opened 2018-11-30
Include highcharts directly

wpears:
## Changes

We already depend on highcharts transitively via cfpb-chart-builder. This PR just makes this dependency explicit which allows yarn & npm installs to generate `node_modules` with highcharts as a direct dependency (ie, `node_modules/highcharts` exists).

This allows us to [import highcharts directly](https://github.com/cfpb/cfpb-chart-builder/pull/156) in both npm and yarn.


anselmbradford:
Oh bummer, we should probably move cfpb-chart-builder to https://github.com/cfpb/cfgov-refresh/tree/master/cfgov/unprocessed/apps at some point.

PR #4659, opened 2018-12-04
Filter: Fix some linter warnings and cleanup

anselmbradford:
## Changes

- Rename `_initEvents` to `_initAnalyticsEvents` to better describe what it does.
- Rename `_setNotification` to `_showNotification` to better describe what it does.
- Scopes the multiselect query to filterable-list-controls so this doesn't reach outside of the filterable list control for multiselects.
- Minor linter warning cleanup. 

## Testing

## Notes

- I removed some linter ignores to report this in the linter report. We may want to re-add these after we're done working in the filters.


anselmbradford:
Codecov can be ignored.

PR #4660, opened 2018-12-04
Bump DATA_UPLOAD_MAX_NUMBER_FIELDS

higs4281:
Wagtail sends field updates via a query string, and on some pages
the number of fields exceeds Django's default max of 1,000 key/value pairs.
For environments that allow admin operations, we'll boost the max
so the admin doesn't encounter 400 errors on complex pages. 

Related to internal issue 3175.

## Testing
Try to save a draft of page 10808.

Django docs suggest setting the limit to `None`, which means no limit, but in testing 2,000 covers our craziest pages.


chosak:
:+1: to this change, but the Travis backend test failures are unexpected.

higs4281:
Two changes were required to resolve Travis issues:
- A change to data_research tests to avoid setting primary keys on test objects.
- A Travis config change for boto tests, suggested in this Travis issue thread:
    - https://github.com/travis-ci/travis-ci/issues/7940#issuecomment-311326476

PR #4664, opened 2018-12-07
core-js and fancy-log are true deps; this reflects that

wpears:
Needed for running with the `--production` flag


anselmbradford:
Whoopsie, thanks 👍

PR #3024, opened 2017-06-13
Click-to-dial phone number (header)

mistergone:
This PR adds markup to the phone number in the "eyebrow" - this markup allows users to click to dial the number.

## Additions
- Markup added to `global-eyebrow.html` and `mega-menu-for-ia-testing.html`
 - I'm not sure what `mega-menu-for-ia-testing.html` is for, but I found the number there, so I marked it up!

## Testing
1. Install this branch and run a server
2. Open the homepage in a mobile browser
3. Click the phone number (in the hamburger menu) to dial!

## Screenshots
screen shot 2017-06-13 at 10 46 47 am

## Notes
- The "hamburger menu" does not appear for me in Blackberry, so I was not able to test this in Blackberry. This also occurs on production, so I suspect a bug for the Blackberry browser.
- The code in `cfgov-refresh` for the Spanish language site already has `tel:` anchors. Hooray!

## Todos
- This is only one instance of the phone number on the site. Other instances are being tracked down.

## Checklist
* [x] PR has an informative and human-readable title
* [x] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [x] Code follows the standards laid out in the [development playbook](https://github.com/cfpb/development)
* [ ] Passes all existing automated tests
* [x] New functions include new tests
* [x] New functions are documented (with a description, list of inputs, and expected output)
* [x] Placeholder code is flagged
* [x] Visually tested in supported browsers and devices
* [x] Project documentation has been updated
* [ ] Reviewers requested with the [Assignee tool](https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/) :arrow_right:

PR #4673, opened 2018-12-11
Housing counselor: Add links for phone and email addresses

anselmbradford:
Phone numbers should probably be clickable for mobile users.
Email addresses should probably be as well.

## Changes

- Makes search result item phone numbers `tel:` URL links.
- Makes search result email address `mailto:` URL links.

## Testing

1. Visit https://www.consumerfinance.gov/find-a-housing-counselor/?zipcode=12047 and see that phone and email are not links in the results.
2. Pull this branch and visit http://localhost:8000/find-a-housing-counselor/?zipcode=12047 and see that phone and emails are links in the results list.

## Screenshots

screen shot 2018-12-11 at 10 35 44 am

## Notes

- I added `tel:+1-` to add a "+1-" to the beginning of the number. I'm not sure if this is necessary, but I saw it elsewhere. However, this will be an error if any of the data has a country code in the number. Any way we can check that? (cc @schbetsy)


niqjohnson:
Although I haven't tested this out yet, this change makes sense to me, especially since this is how we mark up phone numbers and email addresses on the rest of cf.gov.

But, quick question for @schaferjh: do you know of any reason why we didn't mark up phone numbers and email addresses like this in the old version? Would this change surprise anyone or have any unintended consequences?

schaferjh:
@niqjohnson I do not know why numbers were not marked up in the previous version, but I support doing it now since the page is responsive and this more easily allows folks to put information to good use!

anselmbradford:
I imagine it was just because it was an old page. This was a relatively recent addition to other numbers, like https://github.com/cfpb/cfgov-refresh/pull/3024

cfgov/housing_counselor/jinja2/housing_counselor/improved_index.html
```diff
@@ -217,12 +217,18 @@ <h2 class="h4">
                                                 {% endif %}
 
                                                 {% if counselor.phone1 %}
-                                                <b>Phone:</b> {{ counselor.phone1 }}
+                                                <b>Phone:</b>
+                                                <a class="a-link" href="tel:+1-{{ counselor.phone1 }}">
+                                                    <span class="a-link_text">{{ counselor.phone1 }}</span>
```
Scotchester: I don't think we need the `a-link` class or the whole `span.a-link_text` element, since these links don't have icons.

anselmbradford: Good catch. I was following the formatting of the website URL, but that does have an icon. Updated.

PR #4675, opened 2018-12-11
Scope expandable initialization to relevant dom element

anselmbradford:
Fixes [GHE]/CFGOV/platform/issues/2981

## Changes

- Bump cf-expandables, cf-atomic-component, and cf-tables to pull in latest change that allows scoping of AtomicComponent initialization to a particular DOM node.

## Testing

1. `yarn install && gulp clean && gulp build`
2. Visit http://localhost:8000/about-us/newsroom/ and see that when you add Topics to the filter the filter changes in height and the "Apply filter" remains visible.


Scotchester:
I guess it's out of scope to standardize on either `_dom` or `dom`?

anselmbradford:
> I guess it's out of scope to standardize on either _dom or dom?

`_dom` is inside FilterableListControls, a class, and is meant to designate it is not a property that is part of the FilterableListControls public API. `dom` is not inside a class so has no such constraints.

PR #4678, opened 2018-12-11
Add "Creating and Editing Wagtail Components" to docs

Scotchester:
This new docs page (`editing-components.md`) is aimed at addressing the following two user stories:

As a developer, I want to… 
- …edit existing Wagtail atomic components using HTML, CSS, JavaScript, and Python so that they work for my project without breaking existing functionality on other parts of cf.gov.
- …create new Wagtail components using writing HTML, CSS and JavaScript and based on atomic design principles so that I can create new functionality for my project that doesn't already exist.

## Additions

- New docs page, "Creating and Editing Wagtail Components"

## Changes

- Some cleanup of "Notes on Atomic Design"
- Fixed a link to the Wagtail repo to go to wagtail/wagtail instead of the deprecated torchbox/wagtail
- A small amount of cleanup to the `tag.html` template that I came across when adapting it for a better atom example in "Notes on Atomic Design" 

## Notes

- There are probably parts where I'm missing the forest for the trees, or missing a certain species of tree because I'm so enamored with the maples, or something. Please poke holes in it!
- I think that "Notes on Atomic Design" could use even more of an overhaul, but that seems out of scope for this PR.

## To do

- [ ] Add link to new page to the sidebar

## Checklist


docs/editing-components.md
```diff
@@ -0,0 +1,376 @@
...
+
+Before you dive in further,
+[check out the Atomic Structure page](../atomic-structure/)
+and familiarize yourself with our basic concepts
+of atoms, molecules, and organisms.
+
+
+### The Python class
+
+A component's fields and other properties are defined in a Python class file
```
chosak: We have quite a few StreamField blocks that live outside of these three files. These are often visible to Wagtail editors alongside block types that are (for example, [`v1.blocks`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/blocks.py)). Is this documentation explicitly including these from "Wagtail components"?

docs/editing-components.md
```diff
@@ -0,0 +1,376 @@
...
+corresponding to the level of the component.
+
+These files—`atoms.py`, `molecules.py`, and `organisms.py`—are located in
+[`cfgov/v1/atomic_elements`](https://github.com/cfpb/cfgov-refresh/tree/master/cfgov/v1/atomic_elements).
+
+A simple component class looks like this:
+
+```python
+class RelatedContent(blocks.StructBlock):                      # 1
```
chosak: We should mention the possibility of using custom `get_context` methods to add things to the block context, as documented [here](http://docs.wagtail.io/en/v2.3/topics/streamfield.html#template-rendering).

docs/editing-components.md
```diff
@@ -0,0 +1,376 @@
...
+First: the markup.
+
+
+### The HTML template
+
+As mentioned above, each module's Python class has a `Meta` class
+that defines the location of its template file.
+In the `RelatedContent` example,
+it was `template = '_includes/molecules/related-content.html'`.
+That path is relative to `cfgov/jinja2/v1/`.
```
chosak: This is true in this particular case, but other block template HTML is loaded elsewhere. This statement should be made more general to refer to how Django template paths are used.

docs/editing-components.md
```diff
@@ -0,0 +1,376 @@
...
+
+When Wagtail renders a block,
+it includes the values of its sub-blocks in an object named `value`.
+Above, you can see where the sub-blocks are output with
+[Jinja2 expression tags](http://jinja.pocoo.org/docs/2.10/templates/#expressions).
+In particular, note how the `links` `ListBlock` is iterated over,
+and the values of _its_ `Hyperlink` child blocks are output.
+
+That's about as simple an example as it gets, but block templates can get much
+more complex when they have lots of child blocks and grandchild blocks.
```
chosak: This might be a place to mention custom `get_context` as mentioned above. Additionally, use of [`{% include_block %}`](http://docs.wagtail.io/en/v2.3/topics/streamfield.html#template-rendering) could be mentioned here to include child blocks.

docs/editing-components.md
```diff
@@ -0,0 +1,376 @@
...
+
+
+### Adding some style
+
+If a component needs any custom styling not already provided
+by Capital Framework or cfgov-refresh,
+you can it by creating a new Less file for the component.
+This file should live in `cfgov/unprocessed/css/<atoms|molecules|organisms>`.
```
chosak: Is this always going to be true if these blocks live outside of those three directories?

docs/editing-components.md
```diff
@@ -0,0 +1,376 @@
...
+
+
+## How-to guides
+
+
+### Adding, editing, or removing fields
+
+#### Adding a field
+
+1. Locate the Python class of the component you want to add a field to in
```
chosak: See above re: generalizing the location of blocks.

docs/editing-components.md
willbarton: I would ask that "Python class file" be changed to "Python module" or just "Python file". Python doesn't have a Java-esq concept of "class files".

docs/editing-components.md
```diff
@@ -0,0 +1,376 @@
...
+[a common `sidefoot` StreamField](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/models/base.py#L95-L107)
+(so named for the fact that it appears on the right side on some page types,
+but in the footer on others) on the sidebar tab.
+
+
+---
+
+!!! note "Warning: Migrations may be required!"
+    Changes to either of the Python bits will usually require
+    a new schema migration file, and _may_ require a data migration.
```
willbarton: "Python bits"? 😛 

I think this note creates more mystery and more uncertainty than necessary. It can say conclusively something like:

> Adding or changing fields on either Python class will require a new [Django schema migration](https://docs.djangoproject.com/en/1.11/topics/migrations/), and changing fields on a pre-existing block will require a [data migration](https://docs.djangoproject.com/en/1.11/topics/migrations/#data-migrations). 

At least until Wagtail can handle changes to StreamFields a bit better.

You could even explain a little bit of why (that changes to the block appear as changes to the page model to the Django migration system), but maybe that's too much info, or better below. I just don't think migrations and when they need to happen should be mysterious.

docs/editing-components.md
willbarton: Agreed. See above about blocks that will only be used inside a certain app. 

We also have a bad habit of putting absolutely all templates regardless of which Django app uses them into this one template path. That makes the packaging problem much harder.

docs/editing-components.md
```diff
@@ -0,0 +1,376 @@
...
+of atoms, molecules, and organisms.
+
+
+### The Python class
+
+A component's fields and other properties are defined in a Python class file
+corresponding to the level of the component.
+
+These files—`atoms.py`, `molecules.py`, and `organisms.py`—are located in
+[`cfgov/v1/atomic_elements`](https://github.com/cfpb/cfgov-refresh/tree/master/cfgov/v1/atomic_elements).
```
willbarton: We have [other apps that use `blocks.py`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/data_research/blocks.py), which is more in line with Wagtail convention. We should make a distinction between blocks that need to be available to every part of the CMS (maybe to our general pages, Learn Page, etc) and those that are specific to app contexts (like the Data & Research blocks linked to above, or say, if TDP needed to create a custom block for their Activity Pages).

docs/editing-components.md
willbarton: Seconding the question.

docs/editing-components.md
willbarton: Seconded.

docs/editing-components.md
```diff
@@ -0,0 +1,376 @@
...
+
+
+### Creating migrations for StreamField blocks
+
+**tl;dr:**
+
+1. `./cfgov/manage.py makemigrations -n <description_of_changes>`
+1. [Determine if a data migration is needed](#you-may-also-need-a-data-migration)
+
+#### Schema migrations
```
willbarton: It would be nice to link to the [Django documentation](https://docs.djangoproject.com/en/1.11/topics/migrations/) somewhere here.

docs/editing-components.md
```diff
@@ -0,0 +1,376 @@
...
+
+Any time you edit a Python file related to a component,
+it's a good idea to check and see whether or not you need to include
```
willbarton: It's a good idea to check, but again, I feel strongly that we should demystify when and why you need to create a migration. Added or changed blocks on a page model, and added or changed fields on a block will require a schema migration. 

I also really think it's worth explaining *why* — that all fields on all blocks appear on page models for the purposes of Django's migration system. So changing a field on a block needs to create a schema migration for every page that uses that block, since each one of those pages, to the Django migration system, have had their fields changed.

docs/editing-components.md
```diff
@@ -0,0 +1,376 @@
...
+
+If anything shows up, you need a schema migration.
+Run the following, editing it to give your migration a name
+that briefly describes the change(s) you're making:
+
+```bash
+./cfgov/manage.py makemigrations -n <description_of_changes>
+```
+
+Include the created file (located in `cfgov/v1/migrations/`)
```
willbarton: As with above, if your page model that has changed (i.e. it uses a block that you've modified) is not in `v1`, the migration file will not be in `v1`. It will be in whichever Django app the page model is in.

Additionally, it's possible that if you change a `v1` model, there may be migrations that are created in other apps as a result — changing the Hero molecule will, for example, also require a migration for  `regulations3k.models.pages.RegulationLandingPage`, and that migration will live in `regulations3k/migrations`.

I would generalize this path to be wherever the page models are, and I *think* that makes more sense with the explanation of why schema migrations happen when you change a block I talked above above.

docs/editing-components.md
```diff
@@ -0,0 +1,376 @@
...
+- … are deleting a block (or converting it to a new type of block)
+  and don't want to lose any existing data stored in that field …
+
+… then you will need a data migration!
+
+In other words, if an existing field is changing,
+any data stored in that field has to be carefully migrated to the new field,
+unless you're OK with jettisoning it.
+
+**Data migrations are not easy.**
```
willbarton: They don't have to be hard! We shouldn't be discouraging people 😢

willbarton:
Hit the button before I typed a review comment. I'm explicitly requesting changes because the more I read the more convinced I became of my "demystify migrations" argument. I really think being explicit about what changes require migrations and why is exceptionally important.

docs/editing-components.md
Scotchester: @chosak @willbarton Proposed rewrite for the start of this section to address the three comments above:

---

### The Python class

A component's fields and other properties are defined in a Python class.
These classes are located in a number of different files across the repository,
but there are two major categories they fall into:

1. Files corresponding to a general-purpose, site-wide atomic component.
   These files—`atoms.py`, `molecules.py`, and `organisms.py`—are located in
   [`cfgov/v1/atomic_elements`](https://github.com/cfpb/cfgov-refresh/tree/master/cfgov/v1/atomic_elements).
2. Files that are specific to a particular sub-app, such as regulations3k's
   [blocks.py](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/regulations3k/blocks.py).

There are other places where StreamField block classes are defined
(particularly blocks that are only ever used as fields within another block),
but these are the two most common locations
where top-level Wagtail modules are stored.

chosak: @Scotchester love it. What would you think about adding something like "are defined in a Python class, typically a subclass of Wagtail's [`StructBlock`](http://docs.wagtail.io/en/v2.3/topics/streamfield.html#structblock)."

Scotchester: @chosak Sounds good, will do!

docs/editing-components.md
Scotchester: Proposed addition, to be placed under the 8 point list explaining the example:

---

There are two common optional things that are also used in component classes:

1. [Overriding the default `get_context` method](http://docs.wagtail.io/en/v1.13.4/topics/streamfield.html#streamfield-get-context) to pass additional data to the template
2. [Adding component-specific JavaScript](#adding-some-javascript) via the `Media` class

Scotchester: Addressed in a47039c

docs/editing-components.md
Scotchester: @chosak Not sure I understand enough about general Django template path behavior to address this. Can you suggest something?

docs/editing-components.md
Scotchester: Addressed in a47039c

docs/editing-components.md
Scotchester: Addressed in a47039c

docs/editing-components.md
Scotchester: Addressed in a47039c

docs/editing-components.md
Scotchester: Addressed in a47039c

docs/editing-components.md
Scotchester: Addressed in a47039c

docs/editing-components.md
Scotchester: Addressed in a47039c

docs/editing-components.md
Scotchester: Addressed in a47039c

docs/editing-components.md
Scotchester: Sorta addressed in a47039c by moving most of it to the Wagtail Migrations page.

docs/editing-components.md
Scotchester: Addressed in a47039c

docs/editing-components.md
Scotchester: Addressed in a47039c

docs/editing-components.md
chosak: Sure, how about this:

Frontend rendering of Wagtail StreamField blocks to HTML can be controlled by writing a Django template and associating it with the block type. A custom block definition can specify the template path in its `Meta` class:

```py
from wagtail.core import blocks

class PersonBlock(blocks.StructBlock):
    name = blocks.CharBlock()
    email = blocks.EmailBlock()

    class Meta:
        template = 'myapp/blocks/person_block.html'
```

StreamField block templates are loaded by the Django template loader in the same way that Django page templates are. The specified template path must be loadable by one of the Django template engines configured in [`settings.TEMPLATES`](https://docs.djangoproject.com/en/1.11/ref/settings/#std:setting-TEMPLATES) (for this project, the [Django templates backend](https://docs.djangoproject.com/en/2.1/topics/templates/#django.template.backends.django.DjangoTemplates) and the [Jinja2 backend](https://docs.djangoproject.com/en/2.1/topics/templates/#django.template.backends.jinja2.Jinja2)). See [the Django documentation](https://docs.djangoproject.com/en/1.11/topics/templates/#usage) for more details on the search algorithm used to locate a template.

docs/editing-components.md
```diff
@@ -0,0 +1,380 @@
...
+![A screenshot of a Related Content module with heading, paragraph, and links fields](img/related-content-module.png)
+
+Note again that what we think of as **fields** are _also_ blocks,
+and what we think of as **components** or **modules**
+are a special kind of block, `StructBlock`,
+that comprise the sub-blocks that are our fields.
+
+There are two common optional things that are also used in component classes:
+
+1. [Overriding the default `get_context` method](http://docs.wagtail.io/en/v1.13.4/topics/streamfield.html#streamfield-get-context)
```
chosak: Thanks for adding this. Unfortunately this doesn't always work because of our mixed support for `{% include_block %}`. Still good to include it as a forcing factor for us to fix that.

docs/editing-components.md
```diff
@@ -0,0 +1,380 @@
...
+```
+
+This sets up a StreamField named `content` that allows for
+the insertion of any of those seven listed blocks into it.
+
+To make the `RelatedContent` module (shown above) available to this StreamField,
+we'd add a new entry to this list following the same format:
+`('related_content', molecules.RelatedContent()),`.
+
+Most page types have a couple StreamFields (`header` and `content`) in the
```
chosak: "a couple of"? Maybe "two" to be more formal?

docs/wagtail-migrations.md
```diff
@@ -1,12 +1,97 @@
...
+- … are renaming a block within a page's StreamField definition …
+- … are deleting a block (or converting it to a new type of block)
+  and don't want to lose any existing data stored in that field …
+
+… then you will need a data migration!
+
+In other words, if an existing field is changing,
+any data stored in that field has to be migrated to the new field,
+unless you're OK with jettisoning it.
+
```
willbarton: Might be nice to have something here about Django data migrations themselves, including the boiler plate we talked about. 

~~~suggestion

To generate an empty migration file for your data migration, run:

```python
cfgov/manage.py makemigrations --empty yourappname
```

You can also copy the code below to get started with `forward()` and `backward()` functions to migrate your model's data:

```python
from django.db import migrations


    MyModel = apps.get_model('yourappname', 'MyModel')
    for obj in MyModel.objects.all():
        # Make forward changes to the object
        pass


def backwards(apps, schema_editor):
    MyModel = apps.get_model('yourappname', 'MyModel')
    for obj in MyModel.objects.all():
        # Make backward changes to the object
        pass


class Migration(migrations.Migration):
    dependencies = []
    operations = [
        migrations.RunPython(forwards, backwards),
    ]
```

The `forwards()` and `backwards()` functions are where any changes that need to happen to a model's data are made.

~~~

I'm not sure I'm satisfied with this, but some boiler plate for Django and Wagtail page models is good to have, I think, given the specificity of the example below to StreamField migrations.

docs/editing-components.md
```diff
@@ -0,0 +1,380 @@
...
+However, if you…
+
+- … are renaming an existing field …
+- … are deleting a field (or converting it to a new type of field)
+  and don't want to lose any existing data stored in that field …
+- … are renaming a block within a page's StreamField definition …
+- … are deleting a block (or converting it to a new type of block)
+  and don't want to lose any existing data stored in that field …
+
+… then you will need a data migration!
```
willbarton: Same as with this list in the migration doc (below — reviewed this files in reverse order).

docs/wagtail-migrations.md
```diff
@@ -1,12 +1,97 @@
...
-## Migrating StreamFields
 
-StreamFields do not follow a fixed structure, rather they're a freeform sequences of blocks. Making a change to a StreamField involves both creating a [Django schema migration](https://docs.djangoproject.com/en/1.11/topics/migrations/#workflow) and a custom [Django data migration](https://docs.djangoproject.com/en/1.11/topics/migrations/#data-migrations). The data migration needs to modify both the existing Wagtail pages that correspond to the changed model and all revisions of that page. It also needs to be able to manipulate the StreamField contents.
+## Table of contents
 
-To this end, there are some utility functions in cfgov-refresh that make this easier. Using these utilities, a Django data migration that modifies a StreamField would use the following format:
+1. [Reference material](#reference-material)
+1. [Schema migrations](#schema-migrations)
+1. [Data migrations](#data-migrations)
+   1. [Utility functions](#utility-functions)
```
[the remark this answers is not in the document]
willbarton: You totally don't have to act on this. It's my own neuroses talking.

docs/wagtail-migrations.md
```diff
@@ -1,12 +1,97 @@
...
+  and don't want to lose any existing data stored in that field …
+- … are renaming a block within a page's StreamField definition …
+- … are deleting a block (or converting it to a new type of block)
+  and don't want to lose any existing data stored in that field …
+
+… then you will need a data migration!
+
+In other words, if an existing field is changing,
+any data stored in that field has to be migrated to the new field,
+unless you're OK with jettisoning it.
```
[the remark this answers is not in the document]
chosak: The use of "jettisoning" implies that the data will be lost, which isn't necessarily true. It will still exist in the database (at least until the next time the page is saved).

docs/wagtail-migrations.md
```diff
@@ -1,12 +1,97 @@
-# Wagtail and Django data migrations
+# Wagtail Migrations
 
-Django data migrations with Wagtail can be challenging because programmatic editing of Wagtail pages [is difficult](https://github.com/torchbox/wagtail/issues/1101), and pages have both revisions and StreamFields. This document is intended to describe ways we try to address these challenges in cfgov-refresh.
+Adding or changing fields on Wagtail page models or StreamField block classes
```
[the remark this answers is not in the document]
Scotchester: As written, your suggestion seems to imply that Wagtail page models are not Django models. How about "Adding or changing fields on Django models, Wagtail page models (which are a particular kind of Django model), or StreamField block classes …"?

docs/wagtail-migrations.md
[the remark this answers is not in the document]
Scotchester: But so much more annoying to reorder them or add in new sections ;)

docs/wagtail-migrations.md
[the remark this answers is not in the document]
willbarton: That works for me.

docs/wagtail-migrations.md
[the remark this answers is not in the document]
willbarton: It depends on what the migrations are doing. If the "field" in question is a regular model field, then the schema migration is deleting that column from the database, so that data is lost.

It's only the quirk of StreamFields being stored as JSON that means *their* fields are the most special of snowflakes.

docs/wagtail-migrations.md
```diff
@@ -1,12 +1,97 @@
...
+run the following, editing it to give your migration a name
+that briefly describes the change(s) you're making:
+
+```bash
+./cfgov/manage.py makemigrations -n <description_of_changes>
+```
+
+It is a good idea to wait to do this until just before merging your new code
+into the `master` branch, to avoid potential conflicts with other developers'
+migrations that might make it into `master` before yours.
```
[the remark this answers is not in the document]
Scotchester: I suspected this might be controversial... but in my experience, rarely is it true that you need the migration to be able to test the changes. (I don't understand _why_ that is.) But I suppose it would be safer to not say that and just tell people to look out for migration conflicts.

docs/wagtail-migrations.md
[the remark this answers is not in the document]
Scotchester: Still in the DB or no, it's probably safest to encourage people to think of the data as lost, no?

docs/wagtail-migrations.md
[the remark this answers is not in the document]
willbarton: You don't need a migration to test StreamField changes, but you do need one to test model field changes. Again, StreamFields are annoyingly *special*.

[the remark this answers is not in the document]
chosak: > I don't understand _why_ that is.

This is likely because you are testing using `tox -e fast`, which uses [test Django settings](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/test_nomigrations.py#L6) that set [`settings.MIGRATION_MODULES`](https://docs.djangoproject.com/en/1.11/ref/settings/#migration-modules) to `None`, which does this:

> When you supply None as a value for an app, Django will consider the app as an app without migrations regardless of an existing migrations submodule. This can be used, for example, in a test settings file to skip migrations while testing (tables will still be created for the apps’ models).

In this case, the lack of migration doesn't affect the tests, and the database tables are created based on`models.py`. If you run plain `tox`, which does run migrations, and creates database tables based on those migrations, then the missing migration might cause problems with tests that access those tables.

docs/wagtail-migrations.md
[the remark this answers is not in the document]
willbarton: I agree 💯%. Data that only exists until the next time an object is saved is data that might as well be treated as lost.

docs/wagtail-migrations.md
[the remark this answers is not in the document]
willbarton: Oh, true! Didn't think of that either. I was thinking of when you `runserver`.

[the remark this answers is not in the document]
Scotchester: Yeah, I was also only referring to `runserver` testing.

docs/wagtail-migrations.md
```diff
@@ -1,12 +1,97 @@
...
+If you…
+
+- … are renaming an existing field …
+- … are deleting a field (or converting it to a new type of field)
+  and don't want to lose any existing data stored in that field …
+- … are renaming a block within a page's StreamField definition …
+- … are deleting a block (or converting it to a new type of block)
+  and don't want to lose any existing data stored in that field …
+
+… then you will need a data migration!
```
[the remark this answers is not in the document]
willbarton: It was more formatting suggestion than text suggestion.

docs/editing-components.md
Scotchester: Added in 968e783c4997ad98d51c1f8310159a1a68772b07

docs/editing-components.md
```diff
@@ -0,0 +1,380 @@
+# Creating and Editing Wagtail Components
+
+cfgov-refresh implements a number of components
```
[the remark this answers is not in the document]
Scotchester: Yes, good idea! Added in 968e783c4997ad98d51c1f8310159a1a68772b07

docs/editing-components.md
Scotchester: Yeah, I'm not gonna try to explain that in this doc :)

docs/editing-components.md
Scotchester: Addressed in 968e783c4997ad98d51c1f8310159a1a68772b07

docs/wagtail-migrations.md
```diff
@@ -1,12 +1,97 @@
-# Wagtail and Django data migrations
+# Wagtail Migrations
```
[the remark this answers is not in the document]
Scotchester: Addressed in 968e783c4997ad98d51c1f8310159a1a68772b07

docs/wagtail-migrations.md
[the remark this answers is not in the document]
Scotchester: Addressed in 968e783c4997ad98d51c1f8310159a1a68772b07

docs/wagtail-migrations.md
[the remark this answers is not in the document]
Scotchester: (As I just did in 968e783c4997ad98d51c1f8310159a1a68772b07)

docs/wagtail-migrations.md
```diff
@@ -1,12 +1,97 @@
...
+into the concepts presented throughout this page:
+
+- [Django migrations documentation](https://docs.djangoproject.com/en/1.11/topics/migrations/)
+- [Django data migrations documentation](https://docs.djangoproject.com/en/1.11/topics/migrations/#data-migrations)
+- [Wagtail Streamfield migrations documentation](https://docs.wagtail.io/en/v1.13.4/topics/streamfield.html#migrations)
+
+
+## Schema migrations
+
+Any time you are working with StreamField block or Wagtail page definitions,
```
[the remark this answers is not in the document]
Scotchester: Addressed in 968e783c4997ad98d51c1f8310159a1a68772b07

docs/wagtail-migrations.md
```diff
@@ -1,12 +1,97 @@
...
+a Django schema migration will be required.
```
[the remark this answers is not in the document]
Scotchester: Addressed in 968e783c4997ad98d51c1f8310159a1a68772b07

docs/wagtail-migrations.md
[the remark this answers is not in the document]
Scotchester: Addressed in 968e783c4997ad98d51c1f8310159a1a68772b07

docs/wagtail-migrations.md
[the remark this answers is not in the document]
Scotchester: Addressed in 968e783c4997ad98d51c1f8310159a1a68772b07

docs/wagtail-migrations.md
Scotchester: Addressed in 968e783c4997ad98d51c1f8310159a1a68772b07

docs/editing-components.md
Scotchester: Addressed in 968e783c4997ad98d51c1f8310159a1a68772b07

docs/wagtail-migrations.md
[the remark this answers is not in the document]
willbarton: Addressed **beautifully**. ❤️

chosak:
Do you need a modification to mkdocs.yml to add this new page to the docs?

Scotchester:
Rebased from master and added the new page to the nav.

docs/editing-components.md
```diff
@@ -0,0 +1,408 @@
...
+
+There are a few things happening here:
+
+1. The `RelatedContent` class is a subclass of
+   [Wagtail's `StructBlock`](https://docs.wagtail.io/en/v1.13.4/topics/streamfield.html#structblock),
+   which allows for the combination of a fixed number of other sub-blocks
+   (see previous comment about blocks being nested within other blocks)
+   into a single unit (what we'd think of as a "module" in the Wagtail editor).
+   This one has three sub-blocks (lines 2, 3, and 4).
+2. The `heading` field uses the basic Wagtail `CharBlock`, which results in a
```
cfarm: these mentions of the Wagtail block types would be a great place to link to the Wagtail docs again:

https://docs.wagtail.io/en/v1.13.4/topics/streamfield.html#charblock
https://docs.wagtail.io/en/v1.13.4/topics/streamfield.html#richtextblock
https://docs.wagtail.io/en/v1.13.4/topics/streamfield.html#listblock

docs/editing-components.md
```diff
@@ -0,0 +1,408 @@
...
+   to pass additional data to the template
+2. [Adding component-specific JavaScript](#adding-some-javascript)
+    via the `Media` class
+
+#### Adding it to a StreamField
+
+Components are made available in the page editing interface
+by adding them to one of a page types's StreamFields.
+These are usually the first things in a page's class definition.
+For example, see this snippet from `blog_page.py`:
```
cfarm: another good place for a link:

`blog_page.py` https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/models/blog_page.py

docs/editing-components.md
```diff
@@ -0,0 +1,408 @@
...
+into the template (as described at the end of
+[the Python class section](#the-python-class) above),
+those context variables can also be output with simple Jinja2 expression tags:
+`{{ context_var }}`.
+
+#### Adding some style
+
+If a component needs any custom styling not already provided
+by Capital Framework or cfgov-refresh,
+you can it by creating a new Less file for the component.
```
cfarm: typo, should be "you can add it"

docs/editing-components.md
```diff
@@ -0,0 +1,408 @@
...
+
+That's about as simple an example as it gets, but block templates can get much
+more complex when they have lots of child blocks and grandchild blocks.
+Also, if a block definition has overridden `get_context` to pass other data
+into the template (as described at the end of
+[the Python class section](#the-python-class) above),
+those context variables can also be output with simple Jinja2 expression tags:
+`{{ context_var }}`.
+
+#### Adding some style
```
cfarm: I would prefer this heading be more straightforward like "Adding CSS" or just "CSS" so that a quick scan of the TOC or the page is more likely to grab a reader's eye

docs/editing-components.md
cfarm: This paragraph elides some decision making that is required at this stage that we don't really have documented. Do you think we have room here to add more info on how to know if A) the styling you want already exists and B) your project requires custom styles and what the process is for that? 

I am thinking we can link to existing components docs in CF and link to the list of atomic components in this repo on Github for A, and for B, mention that new designs may also require a more involved process in which case they should refer to our internal standards on Hubcap.

docs/editing-components.md
```diff
@@ -0,0 +1,408 @@
...
+   `field_name = blocks.BlockName()`.
+   - Replace `field_name` with a succinct name for what data the field contains
+   - Replace `BlockName` with one of the
+     [basic Wagtail block types](https://docs.wagtail.io/en/v1.13.4/topics/streamfield.html#basic-block-types).
+     Sometimes we create our own custom blocks that can be used, as well.
+     See, for example, the
+     [`HeadingBlock`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/blocks.py#L147-L165),
+     [used in `InfoUnitGroup`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/atomic_elements/organisms.py#L54),
+     among other places.
+1. Add any desired parameters:
```
cfarm: Are you planning to fix the nested lists issue that shows up here (and elsewhere on the page) where the Markdown is just adding `<li>`'s without adding the `<ul>` for the nested list? 
screen shot 2018-12-14 at 11 59 32 am

docs/migrations.md
```diff
@@ -0,0 +1,224 @@
...
+will be required.
+
+To automatically generate a schema migration,
+run the following, editing it to give your migration a name
+that briefly describes the change(s) you're making:
+
+```bash
+./cfgov/manage.py makemigrations -n <description_of_changes>
+```
+
```
cfarm: I would recommend adding a link to existing migrations here to better illustrate the naming pattern and what a good name looks like. Something like:

> For examples of good migration names, see this [list of migrations in cfgov-refresh](https://github.com/cfpb/cfgov-refresh/tree/master/cfgov/v1/migrations).

docs/migrations.md
```diff
@@ -0,0 +1,224 @@
...
+Be aware that each migration must have
+a unique four-digit number at the beginning of its filename,
+which is a regular source of conflicts between pull requests
+that are in flight at the same time.
+If a PR with a migration gets merged between the time you create your migration
+and the time that your PR is ready for merging,
+you will have to update your branch as normal to be current with master
```
cfarm: This paragraph is important and a good reference, I think it could be its own section, and some of the content could be better written as an ordered list with steps. It could come after the note below, with its own sub-heading here, e.g. "Git conflicts in migrations"

cfarm: This could also be done by a later PR

docs/editing-components.md
Scotchester: Good idea! Addressed in 89f23454e89e7befda7ec3cbe69e2b4ca61cd59e

docs/editing-components.md
Scotchester: Added in 89f23454e89e7befda7ec3cbe69e2b4ca61cd59e

docs/editing-components.md
Scotchester: Addressed in 89f23454e89e7befda7ec3cbe69e2b4ca61cd59e

docs/editing-components.md
Scotchester: I have endeavored to largely keep this doc to [a how-doc and not a why-doc](https://codeascraft.com/2018/10/10/etsys-experiment-with-immutable-documentation/), but I think I was able to address your request succinctly enough in 89f23454e89e7befda7ec3cbe69e2b4ca61cd59e

docs/editing-components.md
```diff
@@ -0,0 +1,408 @@
...
+for almost the entire project, it is not necessary
+to tell the Python model anything about a component-specific stylesheet
+(for general-purpose, site-wide components).
+That is _not_ the case with JavaScript, as we will see in the next section.
+
+!!! note
+    If you're working on a component that belongs to a particular sub-app,
+    its Less file should live in `cfgov/unprocessed/<app-name>/css/`,
+    but how those Less files get built and included on their pages is
+    not something this document is prepared to discuss at the moment.
```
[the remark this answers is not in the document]
Scotchester: Because I don't personally actually know 😆 and I think it's different for different apps.

docs/editing-components.md
```diff
@@ -0,0 +1,408 @@
...
+```python
+class RelatedContent(blocks.StructBlock):
+    …  # see first example on this page
+
+    class Media:
+        js = ['related-content.js']
+```
+
+(The `related-content.js` file would need to be placed in
+`cfgov/unprocessed/js/molecules/`.)
```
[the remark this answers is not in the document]
Scotchester: Added a reminder to the start of the front-end section and links in the CSS and JS sections in 89f23454e89e7befda7ec3cbe69e2b4ca61cd59e

docs/editing-components.md
Scotchester: Huh. Honestly, I never ran it locally to take a look. TIL that the MD renderer that MkDocs uses requires 4-space indentation for nested lists. Fixed in 89f23454e89e7befda7ec3cbe69e2b4ca61cd59e

docs/migrations.md
Scotchester: Added in 89f23454e89e7befda7ec3cbe69e2b4ca61cd59e

docs/migrations.md
Scotchester: Rearranged a bit as you suggested. I don't know how well it can be made into an ordered list, though. The steps would be pretty dependent on the particulars of a person's local setup.

Scotchester:
Addressed @cfarm's comments, rebased on master. Any final concerns before merging?

PR #4690, opened 2018-12-13
Housing counselor: Improve map loading experience

anselmbradford:
We sporadically get this error in New Relic:

screen shot 2018-12-12 at 7 04 45 pm

It donned on me that this is happening because while we wait for the page to load before running our scripts, we don't wait for the mapbox.js to load, and we're expecting `L` (mapbox's global namespace) to be available. Most of the time it is, but occasionally it isn't.

Instead, we should default to the no-js view for the search box and then dynamically load the mapbox CSS and JS and when the JS has successfully loaded, change the search view to the JS layout and initialize the map. Mondo improved experience.

## Changes

- Moves mapbox CSS and JS loading from the template to the mapbox initialization script.
- Only calls map initialization after mapbox script has loaded.
- Tweak `no-js` CSS state to apply when `no-js` is on featured content module instead of on a parent (since this is added to the template, it should be fine to not also include CSS for when it's on the parent).
- Fix code comments for formatting and correctness.
- Scope global variables passed in from the template inside a `cfpbHudSettings` object attached to the `window`.

## Testing

1. Pull this branch down.
2. `gulp clean && gulp build`
3. Visit http://localhost:8000/find-a-housing-counselor and see that map loads.
4. Put a typo in `MAPBOX_JS_URL` URL value and `gulp clean && gulp build` again.
5. Reload housing counselor page and see that search area defaults to no-js view.


mistergone:
Looks great! It always makes me happy when our features work without JS. 😄

PR #4689, opened 2018-12-12
Post-Housing Counselor cleanup

schbetsy:
Remove code associated with the legacy Housing Counselor page, including the feature flag that toggled the switchover.

## Removals

- `HUD_TOOL_IMPROVEMENTS` feature flag, as well as the code that checked its value to determine which template to render
- Mapbox-related images that were duplicated
- django template for legacy version of the page

## Changes

- simplify directory structure in `/housing_counselor/tests`

## Testing

1. Running this branch, visit http://localhost:8000/find-a-housing-counselor/
2. Ensure that it is completely unchanged.

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [ ] Code can be automatically merged (no conflicts)
- [x] Code follows the standards laid out in the [development playbook](https://github.com/cfpb/development)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

## Testing checklist
Does not change any cf.gov functionality.

PR #4691, opened 2018-12-13
Set icon size and alignment to be geographically correct

wpears:
Fixes misalignment in master

## Testing
 - Checkout and yarn run gulp build
 - Enter any zipcode on the find a housing counselor page
 - Zoom in and out, note that all markers should remained pointing to the same spot at every zoom level (and that spot should correspond to the address)


wpears:
Not sure about the necessity of those files if the code is being included as an app. Seems entirely duplicative with the code in `unprocessed/apps`. Perhaps @ans or @mistergone can chime in?

anselmbradford:
Yeah `housing_counselor/static/hud/` is just there to support the old version of the page if the feature flag is turned off. Should be removed in https://github.com/cfpb/cfgov-refresh/pull/4689

mistergone:
👍  Thanks so much, @wpears !!

PR #4699, opened 2018-12-14
Move FCM class to organisms, convert rendering of it to include_block

Scotchester:
This PR addresses a few things, most urgently the inability to compare revisions on any page that has an FCM (due to the outdated `FeaturedContent` `template` property). 

It moves the Python code to organisms to be consistent with that change in the front-end code, and removes the `macro`-based rendering setup to the more standard Wagtail `include_block` method. 

It also corrects a lingering template issue with empty link list markup when no link was set up.

## Changes

- Moves the `RelatedPosts` class from `molecules.py` to `organisms.py`
- Updates page templates to render `RelatedPosts` blocks via `include_block` instead of `macro.render()`
- Fixes `related-posts.html` template logic to not render empty link list markup if no links are present

## Removals

- Unused `macro_link` stuff in `related-posts.html`

## Testing

1. Pull branch
2. Visit pages of each type with an FCM and confirm that they are still working
   - Browse: http://localhost:8000/es/prepararse-para-comprar-casa/hipotecas/
   - Sublanding: http://localhost:8000/consumer-tools/auto-loans/
   - Sublanding Filterable: http://localhost:8000/consumer-tools/auto-loans/
   - Browse Filterable: One doesn't exist that uses an FCM. You can make your own.
   - Newsroom Landing: Doesn't currently uses an FCM. You can make your own.

## Checklist

### DIFF
--- a/cfgov/regulations3k/jinja2/regulations3k/browse-regulation.html
+++ b/cfgov/regulations3k/jinja2/regulations3k/browse-regulation.html
@@ -1,7 +1,6 @@
 {% extends 'regulations3k/regulations3k-side-nav.html' %}
 
 {% import 'form_block.html' as form_block with context %}
-{% import 'organisms/featured-content.html' as featured_content with context %}
 {% import 'regulations3k-search-bar.html' as search_bar with context %}
 {% import 'templates/render_block.html' as render_block with context %}
 {% import 'templates/streamfield-sidefoot.html' as streamfield_sidefoot with context %}
--- a/cfgov/jinja2/v1/_includes/organisms/featured-content.html
+++ b/cfgov/jinja2/v1/_includes/organisms/featured-content.html
@@ -60,40 +60,43 @@
 <section class="o-featured-content-module
                 o-featured-content-module__center">
     <div class="o-featured-content-module_text">
-        <div class="a-heading a-heading__icon">
-            {{ category_icon.render( fcm_label(value.category) ) | safe }}
-            {{ fcm_label( value.category ) }}
-        </div>
+        {% if value.category %}
+            <div class="a-heading a-heading__icon">
+                {{ category_icon.render( fcm_label(value.category) ) | safe }}
+                {{ fcm_label( value.category ) }}
+            </div>
+        {% endif %}
         <h2>{{ value.heading }}</h2>
         {{ value.body | safe }}
-        {% if (value.show_post_link and value.post) or (value.links[0] and value.links[0].text) %}
+        {% if (value.show_post_link and value.post) or (value.links[0].text and value.links[0].url) %}
             <ul class="m-list m-list__links">
-                {% if value.show_post_link and value.post %}
+                {% if value.show_post_link and value.post_link_text and value.post %}
                     <li class="m-list_item">
                         <a class="m-list_link" href="{{ pageurl(value.post) }}">
                             {{ value.post_link_text }}
                         </a>
                     </li>
                 {% endif %}
-                {% if value.links[0] and value.links[0].text %}
+                {% if value.links[0].text and value.links[0].url %}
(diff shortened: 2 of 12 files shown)


cfgov/regulations3k/jinja2/regulations3k/browse-regulation.html
```diff
@@ -1,7 +1,6 @@
 {% extends 'regulations3k/regulations3k-side-nav.html' %}
 
 {% import 'form_block.html' as form_block with context %}
-{% import 'organisms/featured-content.html' as featured_content with context %}
```
chosak: Am I reading this correctly that this organism isn't actually used on this Regs3k page type? It doesn't seem to be one of the [supported block types](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/regulations3k/models/pages.py#L201)?

Scotchester: That does seem correct to my eyes, as well. @willbarton or @contolini, can you confirm?

contolini: Hmmm I bet we copied a template when we started regs3k and accidentally left some extraneous stuff.

chosak:
I don't want to blow up the scope of your PR, but while you're in here:

- Is it a bug that [`category`](https://github.com/cfpb/cfgov-refresh/pull/4699/files#diff-dcf2a8703e945442ba8591d07085a645R939) is not required, but a null category [is not checked for in the template](https://github.com/cfpb/cfgov-refresh/pull/4699/files#diff-1960b1d27b4bfc1a4eb279c986d73522R64)? If you leave the category blank, the page validation passes, but you see a None on the page:

    image
- Similarly, [`image`](https://github.com/cfpb/cfgov-refresh/pull/4699/files#diff-dcf2a8703e945442ba8591d07085a645R946) is also optional. I don't think that [this](https://github.com/cfpb/cfgov-refresh/pull/4699/files#diff-dcf2a8703e945442ba8591d07085a645R946) `required=False` is doing what we want, because [this](https://github.com/cfpb/cfgov-refresh/pull/4699/files#diff-1960b1d27b4bfc1a4eb279c986d73522R96) `elif value.image` is still evaluating as true even if you don't specify an image. This is probably because the `ImageBasic` block does always exist; what we really want to check is `value.image.upload`. See also https://github.com/wagtail/wagtail/issues/2665 which explains that `required` doesn't have any affect on a sub-StructBlock. If you don't specify an image, you get this kind of thing in the page source:

    image

    Per this logic, [this](https://github.com/cfpb/cfgov-refresh/pull/4699/files#diff-1960b1d27b4bfc1a4eb279c986d73522R91) `if value.video` is also probably always true, although here we're correctly checking against `.url` and `.id`.
- Conversely, our `Hyperlink` [does seem to have](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/atomic_elements/atoms.py#L77) some kind of custom `required` logic. We don't enforce that links have URLs, which allows you to enter links that don't point anywhere:

    image

    Validating this would require a change to `Hyperlink`, but the template for the FCM could add a check for `value.links[0].url` here to at least skip empty links.
- `post_link_text` [is included in the template](https://github.com/cfpb/cfgov-refresh/pull/4699/files#diff-1960b1d27b4bfc1a4eb279c986d73522R74) even though [it is also not required](https://github.com/cfpb/cfgov-refresh/pull/4699/files#diff-dcf2a8703e945442ba8591d07085a645R944). Should it default to the page title? If you don't enter a title, a weird empty link is inserted in the page source:

    image

Scotchester:
@chosak All good catches, thanks! Fixed them up in the latest commit, then rebased on `master` and force-pushed.

PR #4703, opened 2018-12-14
Catch intermittent ConnectionError on HUD data fetch

schbetsy:
[The call](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/housing_counselor/views.py#L68) that fetches Housing Counselor results from s3 occasionally fails. Previously when it did, the user would see a 500 error page. Now, they will see a standard error message asking them to try the request again.

## Additions

- Catch `ConnectionError`s, which were resulting in 500 errors
- Display new error message in that case

## Testing
There's no way to reproduce the error on demand, unfortunately. 
I guess test that the page and its interactions are unchanged.

## Screenshots
screen shot 2018-12-14 at 5 13 33 pm

## Todos

- [ ] Get the language of the error message approved

## Checklist


chosak:
This approach looks reasonable to me. Did you consider implementing some kind of retry to the request, for example as discussed [here](https://www.peterbe.com/plog/best-practice-with-retries-with-requests)?

One risk of implementing this fix is that we lose the ability to know how often the requests fail. At least now with the 500 errors we have some measure of this. Could we add a logging statement (with no identifying information about the request) so there'd be something to look up if we wanted to measure how often this happens?

schbetsy:
Thanks for the suggestion, @chosak. I've tried implementing the Retry in 47043ea. I'm not 100% confident that I'm using the correct strategy, since I can't recreate the failure. Do you think this will work?

I don't know how to insert logging into the `requests_retry_session`, but I've added logging in the case where the `ConnectionError` persists through all of the retries. Does that look helpful, or should I add more information?

PR #4704, opened 2018-12-14
Switches email-popup close widget from an anchor to a button.

wpears:
Closes GHE/CFGOV/platform/issues/3129

## Changes
 - Switches close button to an actual button. This inserts it into the tab index and is more semantically correct.
 - Adjusts styling to match current anchor styling

## Testing

1. Pull branch and `yarn run gulp styles`
2. Go to http://localhost:8000/owning-a-home
 - You may need to localStorage.clear() and refresh if the "Buying A House?" popup doesn't activate on scroll.
3. Once the popup is activated, you should be able to tab to the close button and close the popup with enter.
4. Styling of the close button should match the one at https://consumerfinance.gov/owning-a-home

### DIFF
--- a/cfgov/unprocessed/css/organisms/email-popup.less
+++ b/cfgov/unprocessed/css/organisms/email-popup.less
@@ -40,24 +40,6 @@
           top: unit( (@grid_gutter-width / 3) / @base-font-size-px, em);
           right: unit( (@grid_gutter-width / 2) / @base-font-size-px, em);
           cursor: pointer;
-
-          button {
-              background: none;
-              border: none;
-              color: @link-text;
-
-              &:hover {
-                  color: @link-text-hover;
-              }
-
-              &:focus {
-                  outline: thin dotted;
-              }
-
-              &:active {
-                  color: @link-text-active;
-              }
-          }
       }
   }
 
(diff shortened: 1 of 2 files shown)


anselmbradford:
Could you utilize `a-btn a-btn__link` from CF to make a button that looks like a link and not need to add more styles here?

wpears:
Excellent, I was hoping something like that class existed ^^;

anselmbradford:
Works as advertised! The after is getting an underline that the before does not have, but I don't mind it. You may want to confirm with a UXer though.

Before:
screen shot 2018-12-17 at 1 14 38 pm

After:
screen shot 2018-12-17 at 1 14 43 pm

cfgov/unprocessed/css/organisms/email-popup.less
```diff
@@ -40,6 +40,10 @@
           top: unit( (@grid_gutter-width / 3) / @base-font-size-px, em);
           right: unit( (@grid_gutter-width / 2) / @base-font-size-px, em);
           cursor: pointer;
+
+          button {
+            border: none;
```
anselmbradford: Since this is overriding a value from an external library it wouldn't hurt to put a small comment here 

`// Override default border on Capital Framework button links.`

Jess may feel it's not necessary to override at all.

schaferjh:
@wpears @anselmbradford looking good. the underline is fine!

PR #4706, opened 2018-12-17
Add auto-generated StreamField documentation

chosak:
This commit adds a new page of reference documentation to the site that describes our use of Wagtail StreamFields. It lists all custom StreamField blocks that can be added to a page (either directly or as a child of another block that can) and all pages/models that can include those blocks.

Due to the large number of blocks and fields, I created a script to automatically generate the document. I've committed the script as a Django management command and included documentation on how to run it.

## Testing

See http://localhost:8000/streamfield-blocks/ when running the local documentation server to review the content.

## Todos

As a follow up to this PR I'd like to add this script to tox (and add a check for up-to-date documentation to Travis), but am committing this first to solicit review on the content/format.

## Checklist


cfgov/core/management/commands/generate_streamfield_docs.py
```diff
@@ -0,0 +1,127 @@
...
+    def handle(self, *args, **options):
+        blocks, models = get_blocks_and_models()
+
+        template = loader.get_template('core/docs/streamfield-blocks.md')
+        md = template.render({
+            'blocks': blocks,
+            'models': models,
+        })
+
+        self.stdout.write(md)
```
willbarton: I think it would be nice if this had an option to write out the output with `-o` or something like that instead of us having to rely entirely on shell output redirection. I could see that potentially getting tricky if this is scripted in, say, yaml.

docs/development-tips.md
```diff
@@ -98,11 +98,23 @@ This tool exposes various useful pieces of information about things like HTTP he
...
 
 Every time a PR is merged to master, Travis will build and deploy the documentation to https://cfpb.github.io/cfgov-refresh. See [How we use Travis CI](https://github.com/cfpb/cfgov-refresh/blob/master/docs/travis.md) for more info.
 
-To add new pages to the navigation, edit the [mkdocs.yml](https://github.com/cfpb/cfgov-refresh/blob/master/mkdocs.yml) file in the root directory.
\ No newline at end of file
+To add new pages to the navigation, edit the [mkdocs.yml](https://github.com/cfpb/cfgov-refresh/blob/master/mkdocs.yml) file in the root directory.
+
+The source Markdown file used to generate one page of reference documentation
+can be generated dynamically from the source code by running the following
+Django management command from within a Python environment:
```
willbarton: The wording here is a bit awkward. Could we be state explicitly what is generated from source code with something like:

> The markdown file for the StreamField blocks documentation is generated from source code. To generate that documentation, run the following Django management command etc:

cfgov/core/management/commands/generate_streamfield_docs.py
```diff
@@ -0,0 +1,127 @@
...
+from wagtail.wagtailcore.models import Page
+
+
+class Command(BaseCommand):
+    help = 'Generate Wagtail StreamField block documentation'
+
+    def handle(self, *args, **options):
+        blocks, models = get_blocks_and_models()
+
+        template = loader.get_template('core/docs/streamfield-blocks.md')
```
willbarton: Thinking ahead, it might be nice if this command took the template path as an input, rather than hard-coding it. Right now with the template file in `cfgov/core/docs`, we end up with some of our documentation (the wording around the templated content) in a non-obvious location well outside of `docs/`.

willbarton:
Looking at the resulting file locally, it's really, really long, and I'm having trouble figuring out how I'd use it. I wonder if there could be better ways to break it up so that it's not an enormous page without much info. 

We also need to add docstrings in so very many places for this to be useful.

willbarton:
Also, this head makes my head explode:

> Django models containing StreamFields

cfgov/core/management/commands/generate_streamfield_docs.py
chosak: Where would you propose this template live instead? I considered putting it somewhere like `docs/templates` but the template is really only useful in the context of the management command. Additionally, the template intro comment specifically cites the management command as the thing to run.

cfgov/core/management/commands/generate_streamfield_docs.py
chosak: Nice suggestion, implemented in 71d0ed9.

docs/development-tips.md
chosak: :+1: Implemented in b1c2b28.

chosak:
> if there could be better ways to break it up so that it's not an enormous page

I agree completely. This isn't particularly useful right now, but hopefully provides a framework to build on? We could split the various parts of this page (blocks, pages using those blocks) into different pages.

> makes my head explode

Agreed.

cfgov/core/management/commands/generate_streamfield_docs.py
chosak: If this page stood alone from the rest of the Markdown MkDocs pages (had a different style, etc) it'd be easier to package the template/command independently. Given that we want the built page to live in the source repository under `docs/` among the other pages, though, I'm not sure what the best separation would be.

Originally (in [this branch on my fork](https://github.com/cfpb/cfgov-refresh/compare/master...chosak:dynamic-streamfield-docs#diff-ea4000de77ab8d50011836372a0fe126)) I tried to separate a bit more by having the management command just generate the JSON and then have MkDocs load the JSON at render time. This would keep the document content/layout in `docs/` and limit the Django interaction with generating source data. My preference is still towards that approach, but I got some feedback that people instead preferred having the fully rendered doc in the repo, so went in the direction in this PR.

Scotchester: If you're referring to me, I don't think I meant to say that. I'm not sure I even realized that an MkDocs page _could_ load JSON at render time. I think that's fine, actually. I just wanted the _data_ to be snapshotted at the point in time when stuff is merged to master, so it's always accurate with that.

Scotchester: I don't know and don't feel like I have enough time to think it through, to be honest. I also like the idea of being able to package this separately, as @willbarton suggests, but we've gotta focus on MVPs as we are approaching the finish line of this project. Do what you think is best 😀

cfgov/core/jinja2/core/docs/streamfield-blocks.md
```diff
@@ -0,0 +1,94 @@
...
+#### {{ model.name }}
+
+- Module: `{{ model.fullname }}`
+- StreamFields:
+{% for field in model.streamfields %}
+    - `{{ field.name }}`{% if field.required %} (required){% endif %}
+{%- endfor %}
+
+{% if model.docstring -%}
+{{ model.docstring }}
```
Scotchester: It's kinda weird that the model.docstrings just return a list of all available stuff in that model if a real docstring isn't found. I'm not sure it adds much value; maybe if we could ignore (or list once at the top) all the stuff inherited from `Page`/`CFGOVPage`?

cfgov/core/management/commands/generate_streamfield_docs.py
willbarton: Honestly, my experience with the way the mini projects happen, if it's not a separate package *now*, it'll never be. See https://github.com/cfpb/wagtail-treemodeladmin vs https://github.com/cfpb/wagtail-treemodeladmin

cfgov/core/jinja2/core/docs/streamfield-blocks.md
willbarton: I also worry slightly that docstrings are (traditionally) written as Restructured Text-ish because of `pydoc` (although that might just me showing my age) and it's included directly as Markdown. It might not be a bad idea to document that the docstrings will be rendered as Markdown.

willbarton: I agree with @Scotchester though that this is weird in both formatting and along side the models that do have docstrings:

image

It took me a while to figure out where that was coming from to comment on it.

willbarton:
I think before this were to get merged, I'd really want us to:

- Fill in docstrings for all of our StreamField blocks (because right now it's just a great big list of all possibilities, and it doesn't strike me as much more than a static version of the Wagtail block inventory we have already)
- Ensuring that the template values are accurate, since I've only just learned that the `template` attribute is ignored entirely in the way we render blocks.

Are these things we can reasonably do before the end of this project?

Scotchester:
> Ensure that the template values are accurate, since I've only just learned that the `template` attribute is ignored entirely in the way we render blocks.

Not entirely, just sometimes :)

willbarton:
> Not entirely, just sometimes :)

I feel sooooo much better now 😛

chosak:
Per the excellent feedback above, this isn't the best approach for this documentation on this project right now. The increased complexity and lack of automatic updates makes it likely that this documentation would quickly get out of date. If we did want to do something automatic, @willbarton's suggestion [above](https://github.com/cfpb/cfgov-refresh/pull/4706#discussion_r242303869) about an independent Wagtail package that provides this kind of documentation probably makes more sense.

PR #4709, opened 2018-12-17
Add a docs Docker container that serves the docs on port 8888

willbarton:
This PR adds another Docker container called "docs" that will run `mkdocs serve` on port 8888. This means that we always have a running copy of the docs along side the running cfgov-refresh on port 8000.

The "docs" container runs Python 3.6 to run MkDocs.

Because of the version of Python 3.6 (3.6.6 from EPEL), I had to downgrade the "futures" requirement. That does not seem to cause any problems for MkDocs or other pacakges.

This is a proof-of-concept PR for your consideration.

## Testing

1.  Run `docker-compose up`. The "docs" container should get built along with the others.
2. View cfgov-refresh at http://localhost:8000/ as usual
3. View docs at http://localhost:8888/ 

## Checklist

### DIFF
--- a/docker/docs/Dockerfile
+++ b/docker/docs/Dockerfile
@@ -1,0 +1,7 @@
+FROM centos:7
+RUN yum install -y which gcc gcc-c++ kernel-devel make epel-release && \
+    yum install -y python36 && \
+    yum clean all
+ADD  https://bootstrap.pypa.io/get-pip.py /src/get-pip.py
+COPY requirements /src/requirements
+RUN python3.6 /src/get-pip.py && pip install -r /src/requirements/docs.txt
--- a/docker/docs/Dockerfile
+++ b/docker/docs/Dockerfile
@@ -2,6 +2,7 @@ FROM centos:7
 RUN yum install -y which gcc gcc-c++ kernel-devel make epel-release && \
     yum install -y python36 && \
     yum clean all
-ADD  https://bootstrap.pypa.io/get-pip.py /src/get-pip.py
 COPY requirements /src/requirements
-RUN python3.6 /src/get-pip.py && pip install -r /src/requirements/docs.txt
+RUN python3.6 -m ensurepip
+RUN pip3 install -U pip
+RUN pip3 install -r /src/requirements/docs.txt
--- a/docs/development-tips.md
+++ b/docs/development-tips.md
@@ -109,3 +109,11 @@ To add new pages to the navigation, edit the [mkdocs.yml](https://github.com/cfp
 
 When running cfgov-refresh using [Docker-compose](https://cfpb.github.io/cfgov-refresh/installation/#docker-compose-installation)
 this documentation is running by default at http://localhost:8888.
+
+When not using Docker, you can run these docs with:
+
+```bash
+mkdocs serve -a :8888
+```
+
+And access on http://localhost:8888.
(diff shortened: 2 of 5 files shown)


docs/development-tips.md
```diff
@@ -105,4 +105,7 @@ by [mkdocs](https://www.mkdocs.org/user-guide/deploying-your-docs/).
 
 Every time a PR is merged to master, Travis will build and deploy the documentation to https://cfpb.github.io/cfgov-refresh. See [How we use Travis CI](https://github.com/cfpb/cfgov-refresh/blob/master/docs/travis.md) for more info.
 
-To add new pages to the navigation, edit the [mkdocs.yml](https://github.com/cfpb/cfgov-refresh/blob/master/mkdocs.yml) file in the root directory.
\ No newline at end of file
+To add new pages to the navigation, edit the [mkdocs.yml](https://github.com/cfpb/cfgov-refresh/blob/master/mkdocs.yml) file in the root directory.
+
+When running cfgov-refresh using [Docker-compose](https://cfpb.github.io/cfgov-refresh/installation/#docker-compose-installation)
+this documentation is running by default at http://localhost:8888.
```
Scotchester: It's probably worth expanding this to include how to run it with a standard virtualenv setup, too.

Scotchester:
This works great!

docs/development-tips.md
willbarton: What, with `docker-compose up`? 😛 

Good idea. When we redo the IA of these docs it'd be nice to move the documentation meta-documentation to somewhere better.

docker/docs/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM centos:7
+RUN yum install -y which gcc gcc-c++ kernel-devel make epel-release && \
+    yum install -y python36 && \
+    yum clean all
+ADD  https://bootstrap.pypa.io/get-pip.py /src/get-pip.py
```
rosskarchner: I think this might be a preferable way of getting pip (I *just* learned about it):

`python36 -m ensurepip`

afterwards, you should be able to refer to pip as `pip3`

rosskarchner: It seems like you might want to follow up with a `pip3 install -U pip` though

willbarton: I know the `get-pip` script fetches the latest version, but it's not a bad idea for `ensurepip` I guess. I'll look into how it works.

willbarton: It looks like we could do [`python36 -m ensurepip --upgrade`](https://docs.python.org/3/library/ensurepip.html#command-line-interface)

rosskarchner: I tried that, and it didn't do what I expected (I still ended up with pip 10)

willbarton: Interesting, same for me:

```
You are using pip version 10.0.1, however version 18.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

Alas.

willbarton: Using both `ensurepip` and `pip install -U pip` in da69ace!

docs/development-tips.md
willbarton: @Scotchester I've added a brief bit on running the docs without Docker, but I think I'm going to break out this section of the docs into their own "Contributing to the docs" document.

PR #4712, opened 2018-12-18
Mortgage Performance Trends: Fix incorrect store

anselmbradford:
Fixes [GHE]/CFGOV/platform/issues/3211

When converting to ES6 modules the map was erroneously importing the chart's store and actions.

## Changes

- Update import statements to use actual name of class that is imported.
- Fix incorrect import of chart store on maps.

## Testing

1. `gulp clean && gulp build`
2.  Visit http://localhost:8000/data-research/mortgage-performance-trends/mortgages-90-or-more-days-delinquent/
3. Change state or date drop-downs and see that map updates.

## Note

- `actions` should really not be generic across chart and map and instead should be named after what is being imported, namely `chartActionCreators` and `mapActionCreators`.


anselmbradford:
Codecov patch can be ignored.

PR #4715, opened 2018-12-19
Rate checker: Clean up `event` variable

anselmbradford:
There was a mix of `event` and `evt` used for event names. Unfortunately, one spot used one but passed in the other, leading to this error:

screen shot 2018-12-19 at 9 28 50 am

## Changes

- Fixes misnamed `event` variable name.
- ~~Standardizes on `evt` name for event object within rate explorer script.~~
- Standardizes on `event` name for event object within rate explorer script.

## Testing

In Firefox 63 or earlier:

1. `gulp clean && gulp build`
2. Pull this branch and visit /owning-a-home/explore-rates/
3. Change the house price input field and see that there are no errors in dev console. Do the same in https://www.consumerfinance.gov/owning-a-home/explore-rates/ and see that there are.


cfgov/unprocessed/apps/owning-a-home/js/explore-rates/rate-checker.js
```diff
@@ -857,15 +857,15 @@ function registerEvents() {
...
-    if ( allowedKeys.indexOf( key ) === -1 || ( event.shiftKey && key !== 9 ) ) {
-      event.preventDefault();
+    if ( allowedKeys.indexOf( key ) === -1 || ( evt.shiftKey && key !== 9 ) ) {
+      evt.preventDefault();
     }
   } );
 
   /* Check if input value is a number.
      If not, replace the character with an empty string. */
   $( '.calc-loan-amt .recalc' ).on( 'keyup', function( evt ) {
```
anselmbradford: See how `evt` is passed in, but `event` is used below.

anselmbradford:
`event` is a special one as it traditionally has some special interactions with a global `event` (see "Cruddy Scoping 2" in https://stackoverflow.com/questions/1510094/is-event-a-reserved-word-in-javascript). This may be why the error this is fixing shows up in older FF but not newer FF. We're probably safe going forward with `event` but it makes me a little uneasy standardizing on it. 
In this rate-checker file in particular we had more `evt`s than `event`s so I went the other way. What do you think? Still go with `event`?

cfarm:
> Also, I am personally biased against abbreviations that save so few keystrokes and reduce clarity.

What he said 👍

anselmbradford:
Okay, I've updated it in the rate-checker script to `event`, but as noted above be aware of the historical warnings around this, necessitating rules like https://eslint.org/docs/rules/no-restricted-globals and https://eslint.org/docs/rules/no-catch-shadow (no longer applicable, because IE8). Generic names like `event` and `error` may lead to less clarity and searchability despite being spelled out.

cfgov/unprocessed/apps/owning-a-home/js/explore-rates/rate-checker.js
```diff
@@ -205,7 +205,7 @@ function updateView() {
         if ( ( new Date( results.timestamp ) ).getDate() !== parseInt( results.timestamp.split( 'T' )[0].split( '-' )[2], 10 ) ) {
           _timestamp = _timestamp.split( 'T' )[0] + 'T15:00:00Z';
         }
-      } catch ( evt ) {
+      } catch ( err ) {
```
anselmbradford: This never should have been `evt`

contolini:
I agree that abbreviations are dumb but in this scenario it's the use of `event` that caused the bug:

```js
button.addEventListener( 'click', evt => {
  console.log( event ); // doesn't throw but a developer would expect it to
} );
```

HOWEVER, I do hate having one file that uses `evt` whenever every other file in the codebase uses `event`. So I vote for `event` even though it's less-than-ideal from a technical perspective.

anselmbradford:
> I think we should go other way on this. A search of all JS files in cfgov-refresh shows 2638 results for event and 215 results for evt.

Yeah but check what those 2638 results are actually matching on ;)

richaagarwal:
I know it might be a much larger effort, but is there no way to give the local variables a more specific/descriptive name, e.g. `keyboard_event` or whatever the best practice is?

anselmbradford:
> I didn't consider the special issues with event that you originally noted. Given that, and the slimmer majority than originally thought, I would be OK moving toward evt.

Whoops, well it's updated to event now.


> I know it might be a much larger effort, but is there no way to give the local variables a more specific/descriptive name, e.g. keyboard_event or whatever the best practice is?

In theory we could for keyboard and click events, but I think we have straight up instances of "Event" too, though I suppose those could be `genericEvent` or something. It's slightly tricky because you have to log the event to get the type unless you're sure of what type it is.

PR #4716, opened 2018-12-19
Housing counselor: Update hud_print.css

anselmbradford:
Fixes [GHE]/CFGOV/platform/issues/3206

## Changes

- Updates print styles for find a housing counselor.

## Testing

2. Visit http://localhost:8000/find-a-housing-counselor/?zipcode=52246 and click "Print list"

## Screenshots

PDF from print below:
~~[Find a Housing Counselor | Consumer Financial Protection Bureau.pdf](https://github.com/cfpb/cfgov-refresh/files/2695743/Find.a.Housing.Counselor.Consumer.Financial.Protection.Bureau.pdf)~~

[Find a Housing Counselor | Consumer Financial Protection Bureau.pdf](https://github.com/cfpb/cfgov-refresh/files/2704095/Find.a.Housing.Counselor.Consumer.Financial.Protection.Bureau.pdf)

## Note

- Since the external link icon appears on its own line in the page's first paragraph, when the icon is hidden in the print view, the sentence ending period ends up on its own line. Not sure we can do much about this, though we should aim to get the external link icon on the line before in the desktop view too by adjusting the copy.


anselmbradford:
@niqjohnson Updated and ready for review. Latest print view is updated in the original post (link for convenience: https://github.com/cfpb/cfgov-refresh/files/2704095/Find.a.Housing.Counselor.Consumer.Financial.Protection.Bureau.pdf)

cfgov/unprocessed/apps/find-a-housing-counselor/css/hud_print.css
```diff
@@ -54,136 +65,51 @@
...
+      margin-bottom: 0;
+      padding-bottom: 0;
     }
 
-    tr.hud_hca_api_result {
-        border-bottom: 1px solid #babbbd !important;
-        padding: 1.5em 0 !important;
+    /* Item result row. */
+    tr {
+      display: grid !important;
```
niqjohnson: What does this look like in a browser that doesn't support CSS Grid Layout?

cfgov/unprocessed/apps/find-a-housing-counselor/css/hud_print.css
```diff
@@ -54,136 +65,51 @@
...
+      grid-row-start: 1;
     }
 
-    .hud_hca_api_counselor_address,
-    .hud_hca_api_site,
-    .hud_hca_api_tel,
-    .hud_hca_api_lang,
-    .hud_hca_api_serv,
-    .hud_hca_api_distance {
-        page-break-inside: avoid;
```
niqjohnson: Looks like the old version avoided page breaks inside results. Is that something we want to keep? Personally I kind of like avoiding page breaks there to keep all info about a result on the same page, but I don't feel super strongly either way.

cfgov/unprocessed/apps/find-a-housing-counselor/css/hud_print.css
```diff
@@ -1,32 +1,48 @@
...
      */
+    a:after {
+        content: '';
+    }
 
-    a[href^="javascript:"]:after,
-    a[href^="#"]:after {
-        content: "";
+    a,
+    a:link,
```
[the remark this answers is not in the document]
anselmbradford: I didn't think so. A printed page won't be clickable 😉

cfgov/unprocessed/apps/find-a-housing-counselor/css/hud_print.css
anselmbradford: With this included it made one result per page. I think we wanted to stack them as close as possible?

cfgov/unprocessed/apps/find-a-housing-counselor/css/hud_print.css
anselmbradford: 🤦‍♂️ Nevermind. I was on the wrong branch.

anselmbradford: This is IE9. Later IEs look equivalent.

screen shot 2018-12-21 at 5 32 29 pm

niqjohnson: That looks about as nice as one could expect for legacy IE, and it looks like all of our other supported browsers support `grid`, so everything looks good here. Thanks for checking!

cfgov/unprocessed/apps/find-a-housing-counselor/css/hud_print.css
niqjohnson: Yeah, I was worried something like one result per page might happen. I definitely prefer *not* one result per page over keeping the info for a listing together, so let's leave this one as is.

cfgov/housing_counselor/jinja2/housing_counselor/index.html
```diff
@@ -16,9 +16,40 @@
...
+                        <p>
+                            The counseling agencies on this list are approved
+                            by the U.S. Department of Housing and Urban
+                            Development (HUD), and they can offer independent
+                            advice about whether a particular set of mortgage
+                            loan terms is a good fit based on your objectives
+                            and circumstances, often at little or no cost to
+                            you. This list shows you several approved agencies
+                            in your area. You can find other approved counseling
+                            agencies at the Bureau of Consumer Financial
```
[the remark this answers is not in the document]
niqjohnson: Via email, @schaferjh said that this can be changed back to "Consumer Financial Protection Bureau" once again.

cfgov/unprocessed/apps/find-a-housing-counselor/css/hud_print.css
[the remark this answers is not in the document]
niqjohnson: Mostly I was noticing that the old print styles left links underlined, and it looks like our general print styles leave links underlined elsewhere on the site, so I was thinking we might want to keep the underlines. But, given that someone has likely chosen to specifically *print* this page rather than make a PDF (otherwise they probably would have used the PDF button), I'm fine keeping the underlines hidden.

cfgov/housing_counselor/jinja2/housing_counselor/index.html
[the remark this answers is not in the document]
anselmbradford: Fixed! https://github.com/cfpb/cfgov-refresh/pull/4716/commits/dca36830ff5927bf56d842ce4b0c17dee1953e63

anselmbradford:
> I tried it out in Safari (12.0.2), too, and the table headers were still showing up only once at the very top of the results list

I see this too. What should be do about this? It appears to be how Safari handles the headers. We could hide them on all browsers?

niqjohnson:
@anselmbradford: Is Safari's mishandling responsive table headers when printing a global thing on cf.gov? It looks like the headers show up correctly if you try to print https://cfpb.github.io/capital-framework/components/cf-tables/, but I know that page isn't 100% in sync with what we're using on cf.gov (and I couldn't find a responsive table like that to test out elsewhere on cf.gov).

At any rate, if this is a global thing, then I say let's keep things as they are and make any needed Safari fixes elsewhere.

anselmbradford:
@niqjohnson I'm not aware of elsewhere where we are using the full-width table either. Let's do what you suggest and merge this and address the Safari issue separately. If that sounds good with you can you update your review status?

niqjohnson:
@anselmbradford: Figuring out what's up with Safari separately sounds good to me. This looks good to go now.

PR #4728, opened 2018-12-28
Documentation for hud_generate_json management command

schbetsy:
An overview of the Find a Housing Counselor tool.

## Additions

- a documentation page under the "Back End Guidelines" category, called "Housing Counselor Data Processing" 

## Notes

- I'm not up to speed on our documentation standards. Feedback welcome on where this should live, its organization, formatting, and content, and anything else.

## Checklist

### DIFF
--- a/docs/housing-counselor-tool.md
+++ b/docs/housing-counselor-tool.md
@@ -1,0 +1,228 @@
+# Find a Housing Counselor
+
+The [Find a Housing Counselor tool](https://www.consumerfinance.gov/find-a-housing-counselor/) allows users to search for HUD-approved housing counselors.
+Users enter a U.S. ZIP code, and the tool returns a list of the ten housing counselors nearest to that ZIP code location.
+The page also displays a map of the results using Mapbox.
+
+When the user enters a ZIP code in the page's search box, the [Django view](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/housing_counselor/views.py#L84) fetches a JSON file of the results from our Amazon S3 bucket.
+The results are inserted into the [page template](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/housing_counselor/jinja2/housing_counselor/index.html) and the [Mapbox map](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/apps/find-a-housing-counselor/js/common.js).
+The page also contains a link to a PDF version of the results, which is also stored in S3.
+The files are publicly accessible, so the tool can run on localhost or in a container without any change in behavior.
+
+This page documents the process we use to generate the JSON and PDF files the Find a Housing Counselor tool relies on.
+
+
+## Housing Counselor Data Processing
+
+The tool gets its data from the U.S. Department of Housing and Urban Development (HUD).
+HUD provides an [API](https://data.hud.gov/housing_counseling.html) to their list of approved counseling agencies.
+A daily job, `cf.gov-housing-counselor-data` on our external Jenkins server,
+queries HUD data and produces the JSON and PDF files we use for the Find a Housing Counselor tool.
+It performs the following steps, each of which is optional
+and configured using parameters before starting the job.
+
+  1. [Geocode ZIP codes](#geocode-zip-codes) (`GEOCODE_ZIPCODES`)
+  2. [Generate JSON files](#generate-json-files) (`MAKE_JSON`)
+  3. [Generate HTML files](#generate-html-files) (`MAKE_HTML`)
+  4. [Generate PDFs](#generate-pdfs) (`MAKE_PDF`)
+  5. [Upload to S3](#upload-to-s3) (`UPLOAD_TO_S3`)
+
+
+### Geocode ZIP codes
+
+The `GEOCODE_ZIPCODES` step generates a file of all ZIP codes in the United States and their location latitude and longitude and saves it in the Jenkins workspace.
+By default, this step is not enabled.
+Since ZIP code geographical information rarely changes, we run this step rarely by manually enabling the option in the Jenkins job.
+
+If enabled, this step calls the [`hud_geocode_zipcodes`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/housing_counselor/management/commands/hud_geocode_zipcodes.py) management command.
+The management command in turn calls the `BulkZipCodeGeocoder` in [`geocoder.py`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/housing_counselor/geocoder.py).
+`BulkZipCodeGeocoder` uses the [Mapbox](https://www.mapbox.com/) geocoding API to determine which 5-digit number sequences are ZIP codes and fetch their latitude and longitude values.
+The management command inserts this data into a CSV and saves it to `./zipcodes.csv` on the Jenkins job workspace.
+
+The generated file looks like this:
+
+```csv
+12305,42.81,-73.94
+12306,42.77,-73.96
+12307,42.81,-73.93
+12308,42.82,-73.93
+12309,42.81,-73.91
+12325,42.88779,-73.99597
+12345,42.80856,-74.02737
+```
+
+
+### Generate JSON files
+
+When enabled, the `MAKE_JSON` step generates a JSON file of housing counselor data for each ZIP code in the U.S.
+Each file contains the ten results geographically nearest to the ZIP code's latitude and longitude.
+This step is enabled by default.
+
+This step calls the [`hud_generate_json`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/housing_counselor/management/commands/hud_generate_json.py) management command,
+which performs the following steps:
+
+  1. fetch agency listings from HUD
+  2. clean the results
+  3. fill in any missing latitude and longitude values
+  4. create files of the 10 nearest results for each U.S. ZIP code
+
+
+#### Fetch agency listings
+
+(_in [`fetcher.py`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/housing_counselor/fetcher.py)_)
+
+Request every housing counselor in HUD's database
+with a request to `https://data.hud.gov/Housing_Counselor/searchByLocation?Lat=38.8951&Long=-77.0367&Distance=5000`.
+
+It returns thousands of results like this:
+
+```json
+{
+    "services": "DFC,FBC,PPC,RHC",
+    "languages": "ENG,SPA",
+    "agc_STATUS": "A",
+    "agc_SRC_CD": "HUD",
+    "counslg_METHOD": "Face to Face Counseling,Group Counseling,Phone Counseling",
+    "agcid": "80790",
+    "adr1": "1234 N. Example St.",
+    "adr2": " ",
+    "city": "SPRINGFIELD",
+    "email": "counselor@example.org",
+    "fax": "999-888-7777",
+    "nme": "EXAMPLE COMMUNITY HOUSING SERVICES",
+    "phone1": "111-222-3333",
+    "statecd": "MI",
+    "weburl": "www.example.org",
+    "zipcd": "48219-8888",
+    "agc_ADDR_LATITUDE": "42.442658",
+    "agc_ADDR_LONGITUDE": "-83.28329",
+    "parentid": "81228",
+    "county_NME": "",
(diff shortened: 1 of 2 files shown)


docs/housing-counselor-data.md
```diff
@@ -0,0 +1,163 @@
+# Housing Counselor Data Processing
+
+The [Find a Housing Counselor tool](https://www.consumerfinance.gov/find-a-housing-counselor/) gets its data from the US Department of Housing and Urban Development. HUD provides an API to their list of approved counseling agencies.
+
+We generate the JSON data for the Find a Housing Counselor tool using the `hud_generate_json` management command. This page describes how that command works to transform HUD data to the JSON file format we use for the FAHC tool, including the following steps:
+  1. fetch agency listings from HUD
+  2. clean the results
+  3. fill in any missing latitude and longitude values
+  4. create files of the 10 nearest results for each US ZIP code
```
Scotchester:
```suggestion
  4. create files of the 10 nearest results for each U.S. ZIP code
```

docs/housing-counselor-data.md
```diff
@@ -0,0 +1,163 @@
...
+Request every housing counselor in HUD's database with this request:
+
+```
+https://data.hud.gov/Housing_Counselor/searchByLocation?Lat=38.8951&Long=-77.0367&Distance=5000
+```
+
+It returns thousands of results like this:
+
+```json
+    {
```
Scotchester: Without the context of the start of the file, the indentation in this example is unnecessary and could be removed.

docs/housing-counselor-data.md
```diff
@@ -0,0 +1,163 @@
...
+        "colonias_IND": "Y",
+        "migrantwkrs_IND": "N"
+    },
+```
+
+## Clean the results
+
+(_in `fetcher.py`_)
+
+We replace the 'languages' value with the full language names. We replace the 'services' value with fully spelled out service descriptions. Both of these mappings come from the HUD API.
```
Scotchester:
```suggestion
We replace the `languages` values with the full language names. We replace the `services` values with fully spelled out service descriptions. Both of these mappings come from the HUD API.
```

docs/housing-counselor-data.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Fetch agency listings
+
+(_in `fetcher.py`_)
+
+Request every housing counselor in HUD's database with this request:
+
+```
+https://data.hud.gov/Housing_Counselor/searchByLocation?Lat=38.8951&Long=-77.0367&Distance=5000
+```
```
Scotchester: Maybe pull this out of a separate code fence and up with the sentence leading into it, like:

> Fetch a list of every housing counselor in HUD's database with a request to `https://data.hud.gov/Housing_Counselor/searchByLocation?Lat=38.8951&Long=-77.0367&Distance=5000`.

docs/housing-counselor-data.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Clean the results
+
+(_in `fetcher.py`_)
+
+We replace the 'languages' value with the full language names. We replace the 'services' value with fully spelled out service descriptions. Both of these mappings come from the HUD API.
+
+(_in `cleaner.py`_)
+
+Clean the data: Convert latitude and longitude values to Floats. Convert the the city and organization name to title case. Ensure email appears to be an email (otherwise leave it blank). Add `http://` to the URL if it is present.
```
Scotchester:
```suggestion
Clean the data: Convert latitude and longitude values to floats. Convert the the city and organization name to title case. Ensure email appears to be an email (otherwise leave it blank). Add `http://` to the URL if it is not already present.
```

docs/housing-counselor-data.md
```diff
@@ -0,0 +1,163 @@
...
+
+(_in `geocoder.py`_)
+
+If any counselor records are missing their latitude or longitude data, we fill in those values with the lat/long location of the agency's ZIP code. This uses a saved data file with mappings from each ZIP code to its latitude and longitude values. We create the file of mappings ahead of time, using the `hud_geocode_zipcodes` management command.
+
+## Create collections of results by ZIP code
+
+(_in `generator.py`_)
+
+Create a local sqlite database and define a `distance_in_miles` function in it. Fill the database with a three-column table: For each counselor in the list, include its latitude and longitude (both in radians) and all of its information from the HUD API as a JSON text string.
```
Scotchester:
```suggestion
Create a local SQLite database and define a `distance_in_miles` function in it. Fill the database with a three-column table: For each counselor in the list, include its latitude and longitude (both in radians) and all of its information from the HUD API as a JSON text string.
```

docs/housing-counselor-data.md
```diff
@@ -0,0 +1,163 @@
...
+
+For each ZIP code in the US, query the database to find the 10 closest housing counselors to the lat/long of that ZIP. Put the information in a JSON structure like this (but with ten results instead of two):
```
Scotchester:
```suggestion
 For each ZIP code in the U.S., query the database to find the 10 closest housing counselors to the lat/long of that ZIP. Put the information in a JSON structure like this (but with ten results instead of two):
```

docs/housing-counselor-data.md
```diff
@@ -0,0 +1,163 @@
...
+        "state_NME": "New York",
+        "agc_ADDR_LONGITUDE": -73.75812,
+        "colonias_IND": "N"
+    }]
+}
+```
+
+Save the resulting JSON into a file on the workspace.
+
+After the management command is finished, we then upload the generated JSON files to Amazon s3.
```
Scotchester:
```suggestion
After the management command is finished, we then upload the generated JSON files to Amazon S3.
```

Scotchester:
P.P.S: I realize now that my comments about the automation kinda contradict what I said in the GHE issue. Upon further reflection, I do think it makes sense to talk about it as long as we don't disclose the Jenkins URL. Maybe also include a tl;dr section at the top for someone who isn't interested in the particulars of what the management commands are doing, they just want to get the data into the tool for local development.

mkdocs.yml
```diff
@@ -21,6 +21,7 @@ nav:
     - Testing: testing-be.md
     - Translation: translation.md
     - Caching: caching.md
+    - Housing counselor data processing: housing-counselor-data.md
```
chosak: We've discussed this a bit as part of a larger documentation IA refactor, but this might be a good opportunity to move this down to the "API" section and rename it to something like "Site content" or "What's on cf.gov" (there's probably a much better name for this). What do you think @Scotchester?

docs/housing-counselor-data.md
```diff
@@ -0,0 +1,163 @@
+# Housing Counselor Data Processing
+
+The [Find a Housing Counselor tool](https://www.consumerfinance.gov/find-a-housing-counselor/) gets its data from the US Department of Housing and Urban Development. HUD provides an API to their list of approved counseling agencies.
```
chosak: Could we add a link to [the HUD API](https://data.hud.gov/housing_counseling.html) here?

docs/housing-counselor-data.md
```diff
@@ -0,0 +1,163 @@
...
+
+```json
+    {
+        "services": "DFC,FBC,PPC,RHC",
+        "languages": "ENG,SPA",
+        "agc_STATUS": "A",
+        "agc_SRC_CD": "HUD",
+        "counslg_METHOD": "Face to Face Counseling,Group Counseling,Internet Counseling,Phone Counseling",
+        "agcid": "80790",
+        "adr1": "24331 W. Eight Mile Road",
```
chosak: It occurs to me that it might be a good idea to modify this data (both here and below) into a made-up counselor instead of including specific details of a real one. That way in future if this real one changes its data or is closed, we won't have out-of-date data in our docs. We also wouldn't want to be seen as specifically endorsing or highlighting one counselor.

docs/housing-counselor-data.md
```diff
@@ -0,0 +1,163 @@
+# Housing Counselor Data Processing
+
+The [Find a Housing Counselor tool](https://www.consumerfinance.gov/find-a-housing-counselor/) gets its data from the US Department of Housing and Urban Development. HUD provides an API to their list of approved counseling agencies.
+
+We generate the JSON data for the Find a Housing Counselor tool using the `hud_generate_json` management command. This page describes how that command works to transform HUD data to the JSON file format we use for the FAHC tool, including the following steps:
```
chosak: How would you feel about adding links throughout this document to the Python source files being referenced, e.g. [`hud_generate_json`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/housing_counselor/management/commands/hud_generate_json.py)? This would make it easier for developers to follow along in the code when reading this document.

docs/housing-counselor-data.md
chosak: Instead of local, perhaps we should say ["in-memory"](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/housing_counselor/generator.py#L38) here so as not to imply that a file gets left behind in the workspace when this is run.

mkdocs.yml
Scotchester: I'm not sure. The content of the existing pages in the API section is significantly different than this page; they are about APIs we _provide_. I think I like it better here in the back-end section, for now.

schbetsy:
Thanks for your comments, @Scotchester and @chosak. I'm incorporating your feedback and adding the sections you requested. This adds significantly more content to the page, so I'll be taking a day or so to write and edit it.

Scotchester:
@schbetsy No rush! Thank you for your work on this.

schbetsy:
@Scotchester and @chosak : I think I've incorporated all of your comments, so this page is ready for another round of review. Note: I'm planning to squash the PR to one commit before I merge.

docs/housing-counselor-tool.md
```diff
@@ -0,0 +1,228 @@
...
+
+
+### Upload to S3
+
+When enabled, the `UPLOAD_TO_S3` step uploads the contents of the `jsons` and `pdfs` directories to an Amazon S3 bucket where it can be accessed by consumerfinance.gov.
+This step is enabled by default.
+
+The files are publicly accessible, e.g.:
+
+ - http://files.consumerfinance.gov/a/assets/hud/jsons/12345.json
```
chosak: We have a redirect set up on http:// requests to files.cf.gov to https://, and files.cf.gov now supports HTTPS, so these two items can be combined into a single one: https://files.consumerfinance.gov/a/assets/hud/jsons/12345.json

schbetsy: Interesting. Does that mean this code is out of date? https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/s3utils.py#L4

chosak: Nice find! I'd forgotten about that code.

If I recall, I think the idea behind that code was to support generating URLs for arbitrary files on S3, where the S3 bucket name is defined by the `AWS_STORAGE_BUCKET_NAME` Django setting. In theory one could run cfgov-refresh with that setting set to something other than "files.consumerfinance.gov" (for example, if/when build gets set up with its own S3 bucket).

Currently, however, the only use of that code is in the housing counselor logic [here](https://github.com/cfpb/cfgov-refresh/blob/50b9a4458f18836492b6147fb3cfbc9f83645537/cfgov/housing_counselor/views.py#L48) -- but this seems wrong, because here we don't want the URLs to be dependent on that setting, because the files only ever exist in one place (on files.cf.gov). Even if other instances of cf.gov (like build) that had their own S3 buckets for, say, Wagtail uploads, we'd still want their housing counselor page to be pulling files from files.cf.gov.

Given that, maybe it makes sense to change [`settings.HOUSING_COUNSELOR_S3_PATH_TEMPLATE`](https://github.com/cfpb/cfgov-refresh/blob/fb16e906bc4935669f880c270ccf4e32b930b068/cfgov/cfgov/settings/base.py#L324) to be a full path like `https://files.consumerfinance.gov/a/assets/hud/{format}s/{zipcode}.{format}`, thus ensuring that every cf.gov instance everywhere would always use the right bucket. This would let us remove this s3utils file entirely, as it's not used anywhere else.

I think we'll need to do this anyway if/when build gets its own bucket, because otherwise the housing counselor download links would all be wrong... @schbetsy what do you think? I can open a PR with this change if you think it makes sense.

schbetsy: PR for this change: https://github.com/cfpb/cfgov-refresh/pull/4759

PR #4730, opened 2018-12-28
Filters: Convert fieldCache object hash to a Map

anselmbradford:
`fieldCache` object was using HTMLNodes as keys, which get converted to strings, which aren't unique across instances of HTMLNodes, leading to some values overwriting others. Maps are restricted to strings as keys, which solves the issue (all credit to @wpears)

## Changes

- Convert `fieldCache` object to Map.

## Testing

2. `gulp clean && gulp build` and test adding and applying a filter, like on /about-us/newsroom/

### DIFF
--- a/cfgov/unprocessed/js/modules/util/FormModel.js
+++ b/cfgov/unprocessed/js/modules/util/FormModel.js
@@ -25,7 +25,7 @@ function FormModel( form ) {
     ]
   };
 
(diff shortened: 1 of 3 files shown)


anselmbradford:
On hold pending merging of https://github.com/cfpb/cfgov-refresh/pull/4733

PR #4732, opened 2019-01-02
Add option to Export Ask data from Wagtail admin

richaagarwal:
This introduces a self-service option for downloading Ask CFPB data.  See GHE/CFGOV/enhanced-cms/issues/86 for more context.

Currently, this is done by a developer via a management script, and the spreadsheet is then shared with the team requesting it.  This PR adds a self-service option via the Wagtail admin. The spreadsheet produced will remain the same. 

## Screenshots
- Download interface
screen shot 2019-01-02 at 3 16 56 pm

## TODO
- Add test coverage

### DIFF
--- a/cfgov/ask_cfpb/templates/admin/export.html
+++ b/cfgov/ask_cfpb/templates/admin/export.html
@@ -1,0 +1,22 @@
+{% extends "wagtailadmin/base.html" %}
+{% load i18n %}
+{% block titletag %}{% trans "Export Ask data" %}{% endblock %}
+
+{% block content %}
+
+    {% include "wagtailadmin/shared/header.html" with title="Ask CFPB Export" icon="download" %}
+
+    <div class="nice-padding">
+        <h3>
+            Download a spreadsheet of all Ask CFPB answers. <br>
+            This may take up to a minute or so to complete.
+        </h3>
+        <form method="post">
+            {% csrf_token %}
+            <button type="submit" class="button">
+                Download CSV
+            </button>
+        </form>
+    </div>
+
+{% endblock %}
--- a/cfgov/ask_cfpb/scripts/export_ask_data.py
+++ b/cfgov/ask_cfpb/scripts/export_ask_data.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 import datetime
 from six.moves import html_parser as HTMLParser
 
+from django.http import HttpResponse
 from django.utils import html
 
 import unicodecsv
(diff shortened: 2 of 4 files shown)


cfgov/ask_cfpb/templates/admin/export.html
```diff
@@ -0,0 +1,22 @@
...
+    {% include "wagtailadmin/shared/header.html" with title="Ask CFPB Export" icon="download" %}
+
+    <div class="nice-padding">
+        <h3>
+            Download a spreadsheet of all Ask CFPB answers. <br/>
+            This may take up to a minute or so to complete.
+        </h3>
+        <form method="post">
+            {% csrf_token %}
+            <button type="submit" name="Export CSV" class="button"/>
```
anselmbradford: `<button` has erroneous `/>` at end. Should be `>`.

anselmbradford: Also, I'd avoid the `name` attribute having a space in it. Is this format needed on the backend? Could it be `name="export-csv"` or similar?

hillaryj:
I'm going to review this further, but as a first note I highly recommend adding a start/stop date selector for users to specify what time period they want data from.

Currently, the Django feedback export web interface will throw an error if we try to export too many records, which at this point is not quite the 3 months of a quarterly feedback export much less all the records.

richaagarwal:
@anselmbradford @higs4281 Thanks for the feedback!  Some of that template code was copypasta so I appreciate the attention to detail :) 

Regarding the name in the sidebar, I don't have a strong opinion, but am wondering if we should wait until there are additional utilities to generalize it?  I'd also be curious to hear from my teammates, @marteki and @nataliafitzgerald, if either of you have thoughts on the naming or the layout of the sidebar.

richaagarwal:
@hillaryj Thanks for the note!  I didn't know that about the feedback export.  This export is an export of all Ask CFPB answers - so filtering it by date would be somewhat arbitrary -- you could do by date created or something, but in general they're not creating new questions (currently 1120 questions, which I don't think will grow much/ at all).  The way the data might change over time is edits to existing fields or metadata - but I don't think we expect the data to grow.  Let me know if you still have concerns though!

richaagarwal:
I've addressed the template feedback, and added a couple tests - https://github.com/cfpb/cfgov-refresh/pull/4732/files#diff-b641b2fbe2a94dbfc11084d89faeb30aR108 - would welcome any feedback on the tests!

hillaryj:
Whoops, I misread that as exporting Ask CFPB **Feedback** not **Answers**. I retract the issue with the export - that's an issue with the sheer volume of feedbacks.

PR #4733, opened 2019-01-03
Remove IE10 JavaScript support

anselmbradford:
Gives IE10 the `no-js` treatment, like IE9 before it.

## Removals

- Removes `dataset` polyfill for the video player.

## Changes

- Requires `Map` support for the JS state of the site to work.
- Minor code comment tweaks.

## Testing

1. Pull branch and run `./frontend.sh`
2. Open local site in IE10. http://localhost:8000/prepaid-rule should launch the video in a new tab and play the video in-browser in a modern browser.
3. http://localhost:8000/owning-a-home/ and charts on http://localhost:8000/data-research/mortgage-performance-trends/mortgages-30-89-days-delinquent/ should not work. 

## Note
We use `pointer-event` in CF `a-select` and depend on modernizr to add the `no-csspointerevents` class globally (https://github.com/cfpb/capital-framework/blob/canary/src/cf-forms/src/atoms/select.less#L79 and https://github.com/cfpb/cfgov-refresh/blob/master/gulp/tasks/scripts.js#L58). I've left this structure in place since it's for the CSS.


cfgov/unprocessed/js/modules/VideoPlayer.js
```diff
@@ -6,9 +6,6 @@ import { assign } from './util/assign';
 import { noopFunct } from './util/standard-type';
 import * as jsLoader from './util/js-loader';
 
-// TODO: Remove data-set ponyfill when IE10 support is dropped.
```
Scotchester: RIP pony-fill 🎠

cfarm:
Before this can get approved and merged, we need the info about this change documented for discussion like you did previously for IE9: https://GHE/CFPB/hubcap/issues/109

anselmbradford:
[Updated babel config to](https://github.com/cfpb/cfgov-refresh/pull/4733/commits/4c4b7605e7adac2e436f0f200035614cf81a3193):

screen shot 2019-01-11 at 11 11 07 am

PR #4741, opened 2019-01-07
String util: Name function with its intention

anselmbradford:
`stringValid` does not describe what is being done to determine validity. This changes the function name to `stringHasRestrictedChars` to better describe what it does.

## Changes

- Change string utility `stringValid` function name to `stringHasRestrictedChars`.

## Testing


anselmbradford:
> This works for me, but I think stringHasSpecialChars might be an even more accurate/agnostic function name.

I had the same idea, but since it excludes hyphen and underscore I thought it was maybe more nuanced, though would hyphen and underscore not be considered special characters? I guess at least hyphen wouldn't.

Scotchester:
Hmm... I guess we might need more information about what the negative effects of using those characters in an option value are? From what I can see in the spec, [attribute values can contain any characters](https://html.spec.whatwg.org/multipage/dom.html#attributes).

Scotchester:
Then perhaps we want a `stringHasCharsThatNeedEncodingInURLs` function? 😆 Not sure how to practically shorten that… but maybe it's fine? Maybe `stringNeedsEncodingInURLs`?

anselmbradford:
> Then perhaps we want a stringHasCharsThatNeedEncodingInURLs function? 😆 Not sure how to practically shorten that… but maybe it's fine? Maybe stringNeedsEncodingInURLs?

Well… the function doesn't know the string is intended for URLs, which has the same issues as `stringValid` (it's "valid" for a particular context). It just excludes characters that are on a restricted list that is in a venn diagram with particular special character collections. [Character entities](https://brajeshwar.github.io/entities/) include hyphen, but not underscore. [URL-encoding](https://en.wikipedia.org/wiki/Percent-encoding#Percent-encoding_reserved_characters) excludes quotes, underscore, and hyphen). I think I like my original naming 😁. We should add a code comment in the function though that it is intended for URL-encoding.

Scotchester:
> Well… the function doesn't know the string is intended for URLs, which has the same issues as `stringValid` (it's "valid" for a particular context).

I'm suggesting that the function _should_ know that. There are too many possible kinds of characters that could be special in different ways for `stringHasSpecialChars` to be a useful name.

But given the purpose of it, perhaps it would be even better to ditch the utility function and have the consuming function (`_sanitizeOptions`) just check whether `encodeURI` would have any effect on the string?

anselmbradford:
> I'm suggesting that the function should know that. There are too many possible kinds of characters that could be restricted for different reasons for stringHasRestrictedChars to be a useful name.

I get that, but the function doesn't know. It's possible it could be used outside of a URL-encoding context, so that intention is best left to a code comment above the regex. The function checks the string against restricted characters, I don't think we need to say what those characters are in the name.

> But given the purpose of it, perhaps it would be even better to ditch the utility function and have the consuming function (_sanitizeOptions) just check whether encodeURI would have any effect on the string?

Like:

```js
if ( stringHasRestrictedChars( item.value ) ) {
```

…becomes…

```js
const value = item.value;
if ( encodeURI( value ) !== value ) {
```

Is that what you mean? I think I would be okay with that. I'm not sure of the performance of encoding the whole thing versus searching it, but the difference has to be minuscule and we could remove the utility.

anselmbradford:
@Scotchester 
Okay I removed the utility. I realized it's not exactly a one-to-one change because the restricted character regex included `?` and `%`, which will not change when encoded, so if a string with those characters makes it through it will not fail. However, this should probably be handled on the backend anyway, and as we discussed this seems like an arbitrary limit on the values, when it would support many of the characters excluded.

anselmbradford:
Superceded by https://github.com/cfpb/cfgov-refresh/pull/4775

PR #4746, opened 2019-01-08
Switch ask search div to label and show outline on focus

wpears:
Closes GHE/CFGOV/platform/issues/3131

## Additions

-

## Removals

- CSS rule disabling the default outline for `.input-with-btn input`

## Changes

- Moves ask search-bar_label to a true label

## Testing

1. Pull & yarn run gulp build
2. Navigate to /ask-cfpb/ and focus on the search bar. The default dotted outline should appear.
3. Note that stylistically the search-bar_label is unchanged (although now it is more semantically correct).

Slight caveat to point 3, you'll note that clicking on the label text will now focus on the search input. I think this is a feature, so I didn't disable this behavior, although it is different from live. Also note that the cursor remains `default` over the label rather than `text`. I think this makes sense as well (and seems to match other labels on the site) but is another slight difference from live.

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/search-bar.html
+++ b/cfgov/jinja2/v1/_includes/organisms/search-bar.html
@@ -42,7 +42,7 @@
               {{ value.additional_classes }}">
     <form method="get" action="{{ value.action }}" class="content-l">
         {% if value.label %}
-        <label for="q" class="o-search-bar_label o-search-bar_col">
+        <label for="o-search-bar_query" class="o-search-bar_label o-search-bar_col">
             <h4 class="u-mb0">
                 {{ value.label | safe }}
             </h4>
@@ -51,7 +51,7 @@
         <div class="o-search-bar_input o-search-bar_col">
             <div class="input-with-btn input-with-btn__bar">
                 <div class="input-with-btn_input {% if value.autocomplete %} m-autocomplete{% endif %}">
-                    <input class="a-text-input" type="text" name="q" id="q" value="{{ value.q }}">
+                    <input class="a-text-input" type="text" name="o-search-bar_query" id="o-search-bar_query" value="{{ value.q }}">
                 </div>
                 <div class="input-with-btn_btn">
                     <button class="a-btn" name="btnSearch" id="btnSearch" type="submit">
--- a/cfgov/jinja2/v1/_includes/organisms/search-bar.html
+++ b/cfgov/jinja2/v1/_includes/organisms/search-bar.html
@@ -42,11 +42,11 @@
               {{ value.additional_classes }}">
     <form method="get" action="{{ value.action }}" class="content-l">
         {% if value.label %}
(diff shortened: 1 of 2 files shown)


cfgov/unprocessed/css/enhancements/forms.less
```diff
@@ -165,7 +165,6 @@ input {
 
         &:focus {
             box-shadow: 0 0 0 1px @pacific inset;
-            outline: none;
```
anselmbradford: I wonder what this was used for originally. Looks like this change will affect the global search too, but it looks fine as far as I can tell.

wpears:
Here are the differences between live and with the label change, not the cursor difference and the input hover state (and disregard the wrong input height, that's just a local font issue).
live_with_cursor
label_as_label

anselmbradford:
> note the cursor difference and the input hover state (and disregard the wrong input height, that's just a local font issue).

Ah, well I guess that's how our other labels work (no pointer hand).

keelerr:
@wpears, quick question just to make sure I'm understanding the new functionality. 

If hover is triggered when you mouse over the search box (and your screenshots showing cursor placement are just inaccurate), @designlanguage and I both like how the text input field is highlighted, but find it problematic that the cursor does not indicate text input. 

But based on the location of your cursor in the screenshots, It looks like the hover state is triggered when you hover over the green box. Is this the case?  If it is, we think it's a little unusual that the smaller text field is highlighted, but the cursor is not aligned to that selection nor does it indicate it's ready for input. 

Ideal: When you hover over input field a blue outline around field appears and the cursor looks like it's ready for text input.

wpears:
Also, it's possible to remove the hover and click behavior of the label, but this removes the larger-target-to-click benefit while also going against default browser behavior. Moreover, if a user hovers over and clicks the text (say, if they thought it was a link to a search page), I think we'd rather focus them on the input element than just have nothing happen.

keelerr:
@wpears, thanks for the explanation! It sounds great. And, I think leaving the cursor as a cursor makes the most sense, since it is clickable and will get you to the focused input.

cfgov/jinja2/v1/_includes/organisms/search-bar.html
```diff
@@ -42,16 +42,16 @@
...
             <h4 class="u-mb0">
                 {{ value.label | safe }}
             </h4>
-        </div>
+        </label>
         {% endif %}
         <div class="o-search-bar_input o-search-bar_col">
             <div class="input-with-btn input-with-btn__bar">
                 <div class="input-with-btn_input {% if value.autocomplete %} m-autocomplete{% endif %}">
-                    <input class="a-text-input" type="text" name="q" id="q" value="{{ value.q }}">
```
richaagarwal: I think changing the value of `name` and `id` here from `q` to something else is breaking Ask search on production. According to @virginiacc, the search is expecting/ looking for `q`.  I think there's a related javascript file that depends on it.

PR #4750, opened 2019-01-09
Filterable Lists: Prevent 500 error when filter year is < 1900

schbetsy:
Bugfix: Previously, when a user entered a date earlier than 1/1/1900 in the Date Range fields on a Filterable List page, they would receive a 500 error. This happened because the `set_interpreted_date_values` method in this file uses `strftime` to format date input for display. `strftime` can only handle dates whose year is 1900 or later.

After this fix, when the user enters a date earlier than 1900, they see a validation error that says "Please enter a date later than 1/1/1900."

## Additions

- Added a validator on the `FilterableDateField` class that requires dates to be later than 1/1/1900

## Testing

1. Running this branch, visit any filterable list, e.g. http://localhost:8000/about-us/newsroom/
2. Open the filter controls, enter a date in the "from" and/or "to" field earlier than 1/1/1900.
3. Press "apply filters"
4. See that the page doesn't show the django error page. Instead, it shows a validation error specifying that dates must be later than 1/1/1900.

## Screenshots

screen shot 2019-01-09 at 4 39 03 pm

## Notes

- I chose 1/1/1900 as the cutoff date because that's the earliest date that our code (specifically the `strftime` method) can handle. I could choose a later date to be the cutoff if that would make more sense from a UX perspective.
- I made up the error message ("Please enter a date later than 1/1/1900."). What would be a better way to word it?

## Checklist


anselmbradford:
- Tiny inconsistency: it's possible to enter `1/1/1900` for the date and apply the filter. According to the error message dates later than 1/1/1900 should be entered, implying 1/1/1900 itself should not work. Perhaps update the message to `Please enter a date later than 12/31/1899.` or adjust the logic to exclude 1/1/1900.

- When the error message is shown, a warning message that there are no filter results is also shown. Do we want both messages or just the error message? I think this is a @schaferjh question.

schbetsy:
Good point, @anselmbradford. I've updated the message to say "Please enter a date 1/1/1900 or later". I've also updated the screenshot above to show the "no results" message.

schaferjh:
@anselmbradford @schbetsy interesting little edge case here! I don't think the "no results" message is necessary -- after all, the error message is there because filtering did not actually happen successfully.

Scotchester:
What if we autoadjusted the date to be 1/1/1900?

schbetsy:
@Scotchester My first draft actually had that behavior, but I decided to change it. My reasoning: Looking at the logs, when this error has occured, it's been because the user meant to type a year like 2018, but they fumbled and typed 1018 instead. If I were the user in that situation, I'd want my attention called to the error so I could correct it, instead of the page subbing a meaningless date in instead.

Scotchester:
Hah, of course. That makes way more sense. Sorry for my stupidity :)

chosak:
Could this possibly be addressed by using [`dateutil.parser`](https://dateutil.readthedocs.io/en/stable/parser.html) instead of `strftime` [here](https://github.com/cfpb/cfgov-refresh/blob/5ea6bfd2a2fde6b864bfb3f20b526d93da1e5a20/cfgov/v1/forms.py#L188), as it seems to support earlier dates in a reasonable way?

```py
>>> from dateutil.parser import parse
>>> parse('7/4/19').date()
datetime.date(2019, 7, 4)
>>> parse('7/4/76').date()
datetime.date(1976, 7, 4)
>>> parse('7/4/1776').date()
datetime.date(1776, 7, 4)
```

schbetsy:
@chosak We need a method that does the opposite: taking a date object and turning it into a formatted string. If there's a better method for that than `strftime`, I'd love to use it, I just wasn't sure how to find one.

schbetsy:
@anselmbradford and @schaferjh : Thanks for the feedback. I've now updated the template so the "no results" message only shows when there aren't errors on the form.

schbetsy:
Could I get a final review on this one? @chosak @Scotchester

cfgov/v1/tests/test_filterable_list_form.py
```diff
@@ -89,6 +89,12 @@ def test_filter_by_title(self):
         self.assertEquals(len(page_set), 1)
         self.assertEquals(page_set[0].specific, page1)
 
+    def test_validate_date_after_1900(self):
```
chosak: For someone coming to this code in the future, it might be nice to have another test which is identical to this one but works because the date is valid. That way it'd be easier to see that the difference in expected behavior is due only to the date.

schbetsy: Makes sense. Added that test in 54de85a8111b458651806f8b840c026c1a816bd4.

PR #4756, opened 2019-01-11
Enforce only one filterable list per page

chosak:
This change modifies the behavior of BrowseFilterablePage and SublandingFilterablePage to ensure that at most one FilterControls ~~or Feedback block~~ is entered into the page's content StreamField. Thanks to @anselmbradford for suggesting this change.

We only ever want at most one of these blocks on a page. After this change, trying to save a page with multiple blocks of these types will produce a validation error like this:

image

Unfortunately there's currently no client-side validation in Wagtail, i.e. it still lets you add the block and only prevents saving. Client-side validation of block limits will be added as part of the Wagtail StreamField rewrite being done in https://github.com/wagtail/wagtail/issues/4473.

This logic could probably be expanded to other pages and blocks. I did an audit of all pages and we don't have duplicate filterable lists anywhere -- as far as I can tell, the exisiting logic won't work with multiple filterable lists anyway. It's a little harder to audit the Feedback blocks. @Scotchester, do you think it's a safe assumption to limit those to 1 per page? Could this also be expanded to include TextIntroductions, or do we sometimes have multiple of these on a page? Any other places you can think of to add a limit?

Aside: this implementation isn't particularly DRY, which makes me think there might be a way of consolidating how we've defined these pages and StreamFields.

## Checklist

### DIFF
--- a/cfgov/v1/models/browse_filterable_page.py
+++ b/cfgov/v1/models/browse_filterable_page.py
@@ -3,6 +3,7 @@ from django.db import models
 from wagtail.wagtailadmin.edit_handlers import (
     FieldPanel, ObjectList, StreamFieldPanel, TabbedInterface
 )
+from wagtail.wagtailcore.blocks import StreamBlock
 from wagtail.wagtailcore.fields import StreamField
 from wagtail.wagtailcore.models import PageManager
 from wagtail.wagtailsearch import index
@@ -14,6 +15,21 @@ from v1.models.base import CFGOVPage
 from v1.util.filterable_list import FilterableListMixin
 
 
+class BrowseFilterableContent(StreamBlock):
+    """Defines the StreamField blocks for BrowseFilterablePage content.
+
+    Pages can have at most one filterable list.
+    """
+    full_width_text = organisms.FullWidthText()
+    filter_controls = organisms.FilterControls()
+    feedback = v1_blocks.Feedback()
+
+    class Meta:
+        block_counts = {
+            'filter_controls': {'max_num': 1},
+        }
+
+
 class BrowseFilterablePage(FilterableFeedPageMixin,
                            FilterableListMixin,
                            CFGOVPage):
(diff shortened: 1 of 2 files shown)


chosak:
@willbarton I considered that; it has the slight side effect of reordering the child blocks in the pages (right now, the BFP has full width, filter controls, feedback, and this would make it filter controls, feedback, full width). It's also not obvious where that parent StreamBlock definition should live.

Another option would be to make some kind of `CountLimitingStreamBlock` base that only defines the max counts:

```py
class CountLimitingStreamBlock(StreamBlock):
    class Meta:
        block_counts = {
            'filter_controls': {'max_num': 1},
            'feedback': {'max_num': 1},
        }
```

In theory this could include the max counts for all blocks across all pages (assuming these are consistent). Then any page-specific StreamBlock could inherit from this, ensuring that it complies with the limits. I didn't really love this approach either.

Scotchester:
`CountLimitingStreamBlock` seems like an OK concept, but it would rely on consistently naming those blocks `filter_controls` and `feedback`, wouldn't it?

I would definitely extend it to all places where feedback forms appears (and probably also email signup forms, as our current code does not support more than one on a page, IIRC).

chosak:
> it would rely on consistently naming those blocks

That's a good point @Scotchester, and another reason it might not be a good idea to apply this broadly. This came up as part of the FilterableList work @anselmbradford is doing, so maybe it's worth making a small change for that single purpose (stripping Feedback from this PR) and considering broader application in a future PR?

chosak:
I've simplified this PR to only enforce the singleton block count for FilterControls blocks, punting a larger design discussion about how to handle this more generally down the road.

@Scotchester @willbarton could I please get another look at this?

cfgov/v1/models/browse_filterable_page.py
```diff
@@ -14,18 +15,29 @@
...
                            FilterableListMixin,
                            CFGOVPage):
     header = StreamField([
         ('text_introduction', molecules.TextIntroduction()),
         ('featured_content', organisms.FeaturedContent()),
     ])
-    content = StreamField([
-        ('full_width_text', organisms.FullWidthText()),
-        ('filter_controls', organisms.FilterControls()),
-        ('feedback', v1_blocks.Feedback()),
```
Scotchester: Would it be possible to define the max inline here, rather than adding a whole new class?

chosak: Unfortunately not the way the StreamField constructor works. There's no way to provided this meta information here -- that would be a nice improvement, though.

PR #4759, opened 2019-01-14
Settings point to correct bucket for Housing Counselor files

schbetsy:
The JSON and PDF files that power the Find a Housing Counselor tool are stored on files.consumerfinance.gov for all instances of cf.gov, regardless of where the other S3 files (e.g. Wagtail uploads, etc) are stored for that instance.

So, the `HOUSING_COUNSELOR_S3_PATCH_TEMPLATE` should not depend on the `AWS_STORAGE_BUCKET_NAME` setting, as it did before. This PR updates that setting to always point to files.consumerfinance.gov, and removes the `s3utils.py` utility, since it was not called by any other code.

## Removals

- Remove the [`s3utils.py`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/s3utils.py) utility and its associated tests, since these are no longer used.

## Changes

- Update the [`HOUSING_COUNSELOR_S3_PATCH_TEMPLATE`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/base.py#L324) setting to point to the files.consumerfinance.gov bucket. This will not change for different environments.

## Testing

1. Running this branch, visit http://localhost:8000/find-a-housing-counselor.
2. All functionality--including searching for a ZIP, using the map, and downloading a PDF of results--should be unchanged.

## Checklist


cfgov/housing_counselor/views.py
```diff
@@ -45,7 +44,7 @@ def s3_url(format, zipcode):
             format=format,
             zipcode=zipcode
         )
-        return https_s3_url_prefix() + path
+        return path
```
chosak: Could simplify a bit by just `return`ing the first line in this method.

PR #4760, opened 2019-01-14
Update Notification API

anselmbradford:
- Notification had two methods, `setTypeAndContent` and `setContent`. To simplify this I've changed it to one method, `update`, which requires the type. We always use the type where this is used so this should be fine, and in the future requiring the type will be more self-documenting.

- Actual constants (not just `const`) don't need to appear inside the constructor. I'd originally done this to contain everything within the scope of the constructor as an organizational choice, but we really don't need to do this since the imports appear outside also. This will make it so each instance isn't redefining the constants unnecessarily.

## Changes

- Change Notification's public APIs `setTypeAndContent` and `setContent` with `update`.
- Require notification type when updating the notification content.
- Update strings to template strings.
- Refactor logic for checking supported types to have a guard clause (removes if/else block).
- Update typos in Notification's tests.

## Testing

## Notes

- The `default` state that Notifications begin with is not currently a state that can be set with the public API. We probably should add this for completeness.


anselmbradford:
codecov/patch can be ignored. I don't understand what triggers that but says base coverage is missing.

PR #3117, opened 2017-07-11
Feedback form on success behavior

virginiacc:
Makes form replacement the default success behavior for feedback forms. Currently some feedback forms are replaced by the success notification (on `owning-a-home/help-us-improve` & Spanish Ask answer pages), but the default behavior (on '/owning-a-home', Ask English answer pages) is to leave the form on the page and add the notification to it. This makes the behavior consistent across the feedback form modules.

## Additions

- Add method to `Notification.js` to get the notification element.

## Changes
- In `feedback.html`
  - Add the block background to the feedback form itself instead of the containing div
  - Move the notification into the feedback form so warning messages appear inside the background
- In `feedback-form.js`
  - Always replace feedback form with success notification
- In `FormSubmit.js`
  - If `replaceForm` option selected, swap form out for its notification module (ensures that notification doesn't appear inside `block__bg`)

## Testing

- Check out branch & run gulp
- Navigate to [Ask answer page](http://localhost:8000/ask-cfpb/how-quickly-can-i-get-money-after-i-deposit-a-check-into-my-checking-account-what-is-a-deposit-hold-en-1023/) and verify that feedback form is replaced on success

## Screenshots
Feedback form error:
screen shot 2017-07-10 at 3 22 29 pm

Feedback form success:
screen shot 2017-07-10 at 3 23 57 pm

## Checklist

* [ ] PR has an informative and human-readable title
* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [development playbook](https://github.com/cfpb/development)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated
* [ ] Reviewers requested with the [Assignee tool](https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/) :arrow_right:

PR #4761, opened 2019-01-14
Feedback form on success w/ PII language

wpears:
Pulls in changes from https://github.com/cfpb/cfgov-refresh/pull/3117 (rebased onto master) with updated PII language from GHE/CFGOV/platform/issues/3114.

The underlying template for `cfgov/jinja2/v1/_includes/blocks/feedback.html` has changed quite a bit since this PR was first proposed, so I'd recommend going back over it carefully to make sure it looks/operates as expected.

### DIFF
--- a/cfgov/jinja2/v1/_includes/blocks/feedback.html
+++ b/cfgov/jinja2/v1/_includes/blocks/feedback.html
@@ -6,15 +6,21 @@
 
 {# JS-disabled result handling #}
 
-
-<div class="o-feedback
-            block
-            {% if value.was_it_helpful_text %}
-                block__bg
-                block__border
-                block__padded-top
-            {% endif %}">
-    <div class="u-mb15">
+<div class="o-feedback block" 
+	        {% if value.radio_intro %} 
+	        	data-replace="true" 
+	        {% endif %}>
+	<form method="post"
+	      action="."
+	      class="o-form
+	      		 {% if value.was_it_helpful_text %}
+	      		 block
+		         block__bg
+		       	 block__border
+		       	 block__padded-top
+	       		 {% endif %}"
+	      novalidate>
+	<div class="u-mb15">
         {% if get_messages(request) %}
         {% for message in get_messages(request) %}
             {{ notification.render(message.tags.split(' ')[0], true, message.message) }}
@@ -23,10 +29,6 @@
             {{ notification.render('default', false, '') }}
         {% endif %}
     </div>
-    <form class="o-form"
-          method="post"
-          action="."
-          novalidate>
         <input type="hidden" name="csrfmiddlewaretoken" value="{{ csrf_token }}">
         <input type="hidden" name="form_id" value="{{ form_id }}">
         <input type="hidden" name="referrer" value="{{ settings.referrer }}">
--- a/cfgov/jinja2/v1/_includes/blocks/feedback.html
+++ b/cfgov/jinja2/v1/_includes/blocks/feedback.html
@@ -6,10 +6,7 @@
 
 {# JS-disabled result handling #}
 
-<div class="o-feedback block" 
-	        {% if value.radio_intro %} 
-	        	data-replace="true" 
-	        {% endif %}>
+<div class="o-feedback block">
 	<form method="post"
 	      action="."
 	      class="o-form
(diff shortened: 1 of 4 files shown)


cfgov/unprocessed/js/organisms/FormSubmit.js
```diff
@@ -161,12 +158,9 @@ function FormSubmit( element, baseClass, opts ) {
     }
 
     function fadeInMessage() {
-      if ( opts.minReplacementHeight ) {
-        _baseElement.style.marginBottom = Math.min( _formElement.offsetHeight, 100 ) + 'px';
-      }
-      _formElement.style.display = 'none';
       _notification.setTypeAndContent( _notification.SUCCESS, message );
       _notification.show();
+      _baseElement.replaceChild( _notification.getElement(), _formElement );
```
anselmbradford: What do think about using `_notificationElement` in place of `_notification.getElement()` and leaving the notification API alone?

cfgov/jinja2/v1/_includes/blocks/feedback.html
```diff
@@ -93,7 +92,7 @@
                     {% if page.language == 'es' %}Comentario adicional{% else %}Additional comment{% endif %} <small class="a-label_helper">({% if page.language == 'es' %}opcional{% else %}optional{% endif %})</small>
                 {% endif %}
             </label>
-            <p class="u-mb15"><small><em>{% if page.language == 'es' %}Nota: No incluya información confidencial, como su nombre, información de contacto, número de cuenta o número de seguro social en este campo.{% else %}Note: Do not include sensitive information like your name, contact information, account number, or social security number in this field.{% endif %}
+            <p class="u-mb15"><small><em>{% if page.language == 'es' %}Nota: No incluya información confidencial, como su nombre, información de contacto, número de cuenta o número de seguro social en este campo.{% else %}Please do not share any personally identifiable information (PII), including, but not limited to: your name, address, phone number, email address, Social Security number, account information, or any other information of a sensitive nature.{% endif %}
```
anselmbradford: Can this be broken across multiple lines?

cfgov/jinja2/v1/_includes/blocks/feedback.html
```diff
@@ -6,15 +6,18 @@
...
+	      action="."
+	      class="o-form
+	      		 {% if value.was_it_helpful_text %}
+	      		 block
+		         block__bg
+		       	 block__border
+		       	 block__padded-top
+	       		 {% endif %}"
+	      novalidate>
+	<div class="u-mb15">
```
anselmbradford: Could this use some combination of `block`? I think `block block__sub block__flush-top` may do it. That's off the top of my head though.

cfgov/unprocessed/js/organisms/FormSubmit.js
virginiacc: Hmm, yes -- it looks like `_notificationElement` is already defined in this module, so that would make sense.

wpears:
@virginiacc I noticed this doesn't seem to replace the form when selecting "No" on the was-helpful radio. Is that the intended behavior or should the notification always replace the form?

wpears:
Turns out this was caused by stale code in the service worker zzz, it wasn't that selecting "No" was the cause, it was the second attempt after a hard reload would pull cached, old code from the service worker.

cfgov/jinja2/v1/_includes/blocks/feedback.html
wpears: I have no real opinion on this, but does something in this change require those new classes? Was that just an outstanding issue with this code that it should have had eg, `block__sub` and `block__flush-top` and it would be convenient to sneak it in here?

anselmbradford: Yeah, may be out of scope for here so feel free to ignore it. I thought we can set 15px margin with blocks, though it may only be 15px padding that can be set and 30px margin. `block` combinations are very verbose, but we probably want to be consistent with using them and then could swap in abbreviated patterns for all of them if we ever get to that. Food for thought.

cfgov/unprocessed/js/molecules/Notification.js
```diff
@@ -158,6 +158,10 @@ function Notification( element ) {
     return this;
   }
 
+  function getElement() {
```
anselmbradford: You'll want to remove the changes here if they are unused now, or add the jsdoc description for this method.

PR #4762, opened 2019-01-14
Handle regs3k highlighting TypeError

higs4281:
Occasionally the regulations search produces a null "highlighted" value,
which throws a TypeError when rendering results.

Related issues:
* Internal alerts issue 2361
* Internal platform issue 3184

## Testing
I can't replicate the TypeError locally. 
These params caused the last error in prod:
`regulation: 1026, query: "no"`

To run the new unit test for TypeError:

```
tox -e fast regulations3k
coverage report -m | egrep "Stmts|regulations3k/models"
```

[Short description explaining the high-level reason for the pull request]

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

### Browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Edge
- [ ] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


chosak:
I'm able to reproduce this locally using [this localhost URL](http://localhost:8000/policy-compliance/rulemaking/regulations/search-regulations/results/?q=%20&results=25&order=relevance).

If I apply your fix, this search (for a space character) returns thousands of hits, which eventually fails with a "Result window too large" error. This generates a page that says that there are lots of results, but doesn't show any:

image

What about an alternate fix that instead adds a `.strip()` [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/regulations3k/models/pages.py#L65)? This would treat a space as if it were an empty string, and avoid this problem in a different way.

higs4281:
@chosak Thanks for finding this gap. We should catch both errors.

cfgov/regulations3k/models/pages.py
```diff
@@ -98,8 +103,13 @@ def regulation_results_page(self, request):
             pre_tags=['<strong>'],
             post_tags=['</strong>']).models(SectionParagraph)
         for hit in sqs:
+            try:
+                snippet = Markup(" ".join(hit.highlighted))
+            except TypeError as e:
+                logger.info("Query string {} produced a TypeError: {}".format(
```
chosak: How would you feel about using `logger.warning` here instead?

cfgov/regulations3k/models/pages.py
```diff
@@ -98,8 +103,13 @@ def regulation_results_page(self, request):
...
+                    search_query, e))
+                continue
```
chosak: @higs4281 with this change, if you do a search for ", you get the warning about no results, but the sidebar still shows the result count. If you click on one of the regulations, you get the "Showing X matches" but there are no matches shown.

What about using `hit.text` here instead?

```py
if hit.highlighted:
    snippet = " ".join(hit.highlighted)
else:
    snippet = hit.text

...
'snippet': Markup(snippet),
...
```

This works even if you remove [this](https://github.com/cfpb/cfgov-refresh/pull/4762/files#diff-5ac6519566563e4cf77d313d88876facR78) single character check. Are there cases where searching for single characters might be useful, e.g. "Z"? ([working link with my proposed change](http://localhost:8000/policy-compliance/rulemaking/regulations/search-regulations/results/?q=Z&results=25&order=relevance))

cfgov/regulations3k/models/pages.py
higs4281: Good idea.

PR #4764, opened 2019-01-15
Add recreating migrations to migration doc

willbarton:
This moves the formerly internal "Recreating migrations" document into our existing public "Django and Wagtail migrations" document. It re-works the introductory content of the "Recreating migrations" section to reference the rest of the document and ensure the information that it needs to reference is there. Otherwise this is a straight forward move of the existing content.

## Checklist

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -14,6 +14,7 @@ additionally, changing field names or types on an existing block will require a
 1. [Data migrations](#data-migrations)
    1. [Wagtail-specific consideration](#wagtail-specific considerations)
    1. [Utility functions](#utility-functions)
+1. [Recreating migrations](#recreating-migrations)
 
 
 ## Reference material
@@ -31,7 +32,9 @@ into the concepts presented throughout this page:
 Any time you add or change a field on a Django model, Wagtail page model
 (which are a particular kind of Django model), or StreamField block classes, a
 [Django schema migration](https://docs.djangoproject.com/en/1.11/topics/migrations)
-will be required.
+will be required. 
+When that field is a StreamField, this includes changes to the properties of a child block,
+or even something minor like a help text string.
 
 To automatically generate a schema migration,
 run the following, editing it to give your migration a name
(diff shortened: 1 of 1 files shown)


docs/migrations.md
```diff
@@ -31,7 +32,9 @@ into the concepts presented throughout this page:
 Any time you add or change a field on a Django model, Wagtail page model
 (which are a particular kind of Django model), or StreamField block classes, a
 [Django schema migration](https://docs.djangoproject.com/en/1.11/topics/migrations)
-will be required.
+will be required. 
+When that field is a StreamField, this includes changes to the properties of a child block,
+or even something minor like a help text string.
```
chosak: The "or" here confused me on first read. Changes to something minor like a help text string require migrations for any Django model field, not just StreamFields. Maybe change to "child block, even for changes as small as a modified help text string."

docs/migrations.md
```diff
@@ -14,6 +14,7 @@ additionally, changing field names or types on an existing block will require a
 1. [Data migrations](#data-migrations)
    1. [Wagtail-specific consideration](#wagtail-specific considerations)
```
chosak: There's a missing space here in the anchor link.

docs/migrations.md
```diff
@@ -228,3 +231,95 @@ If commit is `True` (default),
...
+    To recover such a dump, you'll need to check out the code at the point before the recreation was done, 
+    migrate from N->N+1, and then check out latest and migrate forwards.
+
+2. Any open pull requests at the time of the recreation that reference 
+    or depend on some of the existing migrations 
+    will need to be modified to instead refer to the new migration files.
+
+Migrations can be recreated with this process:
+
+1. Remove all existing migration files:
```
chosak: One tricky problem: what about other repositories ([teachers-digital-platform](https://github.com/cfpb/teachers-digital-platform/blob/master/teachers_digital_platform/migrations/0001_initial.py#L15)) that depend on cfgov-refresh migrations?

docs/migrations.md
Scotchester:
```suggestion
    1. [Wagtail-specific considerations](#wagtail-specific-considerations)
```

docs/migrations.md
```diff
@@ -31,7 +32,9 @@ into the concepts presented throughout this page:
 Any time you add or change a field on a Django model, Wagtail page model
 (which are a particular kind of Django model), or StreamField block classes, a
```
Scotchester:
```suggestion
(which are a particular kind of Django model), or StreamField block class, a
```

docs/migrations.md
Scotchester: I support explicitly mentioning things like `help_text` that people might not expect to cause a need for a migration, but isn't the first part of this statement covered by the "or Streamfield block class" phrase above?

willbarton: I think removing the "or" works:

```
When that field is a StreamField, this includes changes to the properties of a child block,
even something minor like a help text string.
```

willbarton: @Scotchester it adds context for "child block". You could have a thrice-nested block and changing the `help_text` still requires a migration on a page that uses a block that uses a block that uses the block you changed. That's not necessarily obvious.

docs/migrations.md
willbarton: I think we resolved this in chat — but as long as the squashed migration has the `includes` list in step 4, it should be fine.

docs/migrations.md
willbarton: Nevermind, I see what you're saying. Will make it a bit clearer.

docs/migrations.md
```diff
@@ -29,9 +30,9 @@ into the concepts presented throughout this page:
 ## Schema migrations
 
 Any time you add or change a field on a Django model, Wagtail page model
-(which are a particular kind of Django model), or StreamField block classes, a
+(which are a particular kind of Django model), or StreamField block class, a
 [Django schema migration](https://docs.djangoproject.com/en/1.11/topics/migrations)
-will be required.
+will be required. This includes changes as small as modifying the `help_text` string.
```
willbarton: I think this is the simplest way to make the point, based on both @chosak and @Scotchester's suggestions.

PR #4769, opened 2019-01-16
Optimize Ask Export queries

richaagarwal:
In https://github.com/cfpb/cfgov-refresh/pull/4732, I introduced an option to export Ask data from the Wagtail admin.  Unfortunately, it was timing out on production as it was taking longer than a minute to run (locally it would take about 30 seconds).  

Thanks to feedback from other developers and much help from @chosak, I was able to significantly reduce the runtime by modifying the queries.  Instead of making several subqueries on each answer, we do a couple bulk queries upfront to get all the attributes we need.  There's a bit more code to specify the `prefetch_fields`, pair each answer with its answer pages, and construct the lists of `ManyToMany` objects (e.g. subcategories) -- but it means we only have to make 2 queries (one set of answers and one set of answer pages) versus over 10,000 queries (number of answers * objects we're querying off each of them). 

Running locally, the export now takes less than 3 seconds to complete.

### DIFF
--- a/cfgov/ask_cfpb/scripts/export_ask_data.py
+++ b/cfgov/ask_cfpb/scripts/export_ask_data.py
@@ -8,7 +8,8 @@ from django.utils import html
 
 import unicodecsv
 
-from ask_cfpb.models import Answer
+from ask_cfpb.models.django import Answer
+from ask_cfpb.models.pages import AnswerPage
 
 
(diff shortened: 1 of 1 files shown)


higs4281:
I deployed the branch to dev4, and the download time was about 7 seconds!

richaagarwal:
@higs4281 Thank you for the review!!

PR #4772, opened 2019-01-17
Remove tags from FOIA reports page

mistergone:
The new designs for filtered list results require the removal of tags from the FOIA reports page (all results are tagged FOIA and only tagged FOIA).

## Changes
- Adds a conditional to post-preview.html that negates the generation of tag markup when the `page_type` is `foia-freq-req-record`

## Testing

1. Pull down this branch.
2. Check out http://localhost:8000/foia-requests/foia-electronic-reading-room/
3. There shouldn't be topic tags!

## Notes
- After a lengthy discussion amongst the UI Improvements Team, we recommend that we use `controls.categories.page_type` to control the generation of markup in the `post-preview.html` template, which will allow some flexibility in the design between filtered list results pages.

## Checklist

### Browsers

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [x] iOS Safari
- [x] Chrome for Android

### Accessibility
- [x] Keyboard friendly
- [x] Screen reader friendly

### Other
- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/post-preview.html
+++ b/cfgov/jinja2/v1/_includes/organisms/post-preview.html
@@ -178,7 +178,7 @@
             {% endfor %}
             </div>
 
-            {% if post.tags.exists() %}
+            {% if post.tags.exists() and controls.categories.page_type != 'foia-freq-req-record' %}
                 {%- import 'tags.html' as tags %}
                 {{ tags.render(post.related_metadata_tags(), hide_heading=true, is_wagtail=True) }}
             {% endif %}
--- a/cfgov/jinja2/v1/_includes/organisms/post-preview.html
+++ b/cfgov/jinja2/v1/_includes/organisms/post-preview.html
@@ -63,6 +63,7 @@
     {% set date_desc = controls.post_date_description or post_date_description or 'Published' %}
     {% set cat_controls = controls.categories %}
     {% set show_categories = cat_controls.show_preview_categories if cat_controls is defined else true %}
+    {% set display_settings = get_fl_display_settings(controls.categories.page_type) if cat_controls is defined else get_fl_display_settings('None') %}
     <article class="o-post-preview">
         <div class="m-meta-header">
             <div class="m-meta-header_right">
@@ -176,9 +177,8 @@
                       </a>
                 {% endif %}
             {% endfor %}
-            </div>
 
-            {% if post.tags.exists() and controls.categories.page_type != 'foia-freq-req-record' %}
+            {% if post.tags.exists() and display_settings.show_tags %}
                 {%- import 'tags.html' as tags %}
                 {{ tags.render(post.related_metadata_tags(), hide_heading=true, is_wagtail=True) }}
             {% endif %}
--- a/cfgov/jinja2/v1/_includes/organisms/post-preview.html
+++ b/cfgov/jinja2/v1/_includes/organisms/post-preview.html
@@ -63,7 +63,7 @@
     {% set date_desc = controls.post_date_description or post_date_description or 'Published' %}
     {% set cat_controls = controls.categories %}
     {% set show_categories = cat_controls.show_preview_categories if cat_controls is defined else true %}
-    {% set display_settings = get_fl_display_settings(controls.categories.page_type) if cat_controls is defined else get_fl_display_settings('None') %}
+    {% set display_settings = get_fl_display_settings(controls.categories.page_type) if controls.categories is defined else get_fl_display_settings('None') %}
     <article class="o-post-preview">
         <div class="m-meta-header">
             <div class="m-meta-header_right">
@@ -177,7 +177,7 @@
                       </a>
                 {% endif %}
             {% endfor %}
-
+            </div>
             {% if post.tags.exists() and display_settings.show_tags %}
                 {%- import 'tags.html' as tags %}
                 {{ tags.render(post.related_metadata_tags(), hide_heading=true, is_wagtail=True) }}
--- a/cfgov/v1/util/ref.py
+++ b/cfgov/v1/util/ref.py
@@ -267,6 +267,7 @@ def get_category_children(category_names):
         for category in category_names
     )))
 
+
 def get_fl_display_settings(filterable_page_type):
     settings = {
         'page_type': filterable_page_type,
--- a/cfgov/v1/util/ref.py
+++ b/cfgov/v1/util/ref.py
@@ -266,3 +266,12 @@ def get_category_children(category_names):
         dict(categories_dict[category]).keys()
         for category in category_names
     )))
+
+def get_fl_display_settings(filterable_page_type):
+    settings = {
+        'page_type': filterable_page_type,
+        'show_tags': True,
+        }
+    if "foia-freq-req-record" in filterable_page_type:
+        settings['show_tags'] = False
+    return settings
--- a/cfgov/v1/jinja2tags/__init__.py
+++ b/cfgov/v1/jinja2tags/__init__.py
@@ -103,6 +103,7 @@ class V1Extension(Extension):
             'is_report': ref.is_report,
             'is_filter_selected': contextfunction(is_filter_selected),
             'render_stream_child': contextfunction(render_stream_child),
+            'get_fl_display_settings': ref.get_fl_display_settings,
         })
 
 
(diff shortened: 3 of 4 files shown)


Scotchester:
Did you consider adding a new option to the `FilterControls` component to turn off tag rendering?

anselmbradford:
@Scotchester we did, but consider that filterable lists have three possible result types: post previews (most things), activity-list (which is checked via the [page_type](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html#L251)), and info-units (which is set via the render as [50-50 option in Wagtail](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/atomic_elements/organisms.py#L907)). We currently have 15 page types. Multiply that by the options for what appear on the filters and you get a very large number of possible combinations. We're thinking to move toward setting the configuration via the Page Type and limit the customization possible in Wagtail (in that case there would be a Page Type for everyone has a story and the "render preview items as 50-50" checkbox would go away).

mistergone:
I also consulted with designers who worked on the filtered list results designs, and they indicated we probably aren't going to see the need for a lot more filtered list pages with these specific constraints in the forseeable future. Thus, there's little need for content editors to be able to select the options, while more options in the controls can lead to more conflicts and potential human error.

Trust me - I tend towards, "Let's add a checkbox that does this" over something hard-coded, but... this is not the only "one-off" design change I need to implement this way. So by the time I'm done, I would be adding two or more checkboxes to the "control panel", each of which is for a single page on the site. I believe this is already the case for the "50/50" option.

So there's a legitimate debate here - do we add more and more options to the Wagtail control panels, or do we sequester these things in "developer land" inside templates and the like? I end up settling on the latter because we don't expect content editors to need to create a number of additional pages where, for instance, topics are hidden, the date is hidden, etc. I also think additional checkboxes add additional confusion.

I do think, if we go down the path of `page_type` dictating some elements of markup, that we should have a small "hint text" in the filterable list "control panel" that indicates that - something like, "_Page Type determines categories that will display in the filter controls, as well as determining what elements of the results might appear._" Then that drop-down isn't as mysterious as it is currently. :)

cfgov/jinja2/v1/_includes/organisms/post-preview.html
```diff
@@ -178,7 +178,7 @@ <h3 class="o-post-preview_title">
             {% endfor %}
             </div>
 
-            {% if post.tags.exists() %}
+            {% if post.tags.exists() and controls.categories.page_type != 'foia-freq-req-record' %}
```
schbetsy: This breaks when I visit http://localhost:8000/policy-compliance/amicus/. It gives this error message: `'None' has no attribute 'categories'`. Do some lists not get a 'controls' variable?

anselmbradford: 😭

anselmbradford: OOooo it looks to me that amicus uses the post preview, but is not a filterable list. Of course! We need to be careful with post preview edits because it's just something that's consumed by filterable lists, but they are independent things.

anselmbradford:
I just noticed a convenience variable `cat_controls` [is defined at the top](https://github.com/cfpb/cfgov-refresh/blob/9203688a253071c2d7f4ebd2ac477a36fb3f6340/cfgov/jinja2/v1/_includes/organisms/post-preview.html#L64). @mistergone I'd reconcile your edit with that by either only referencing `cat_controls` or getting rid of that variable and referencing `controls.categories` directly throughout. Probably should use the variable.

Scotchester:
@mistergone and @anselmbradford, you make a solid case. I am willing to get on board with configuring by page type behind the scenes, however I'd like to see the logic moved to the page type classes and send a context variable like `show_tags` to the template. As this PR is not urgent, I'd prefer to modify it to do that than to merge this and do it as a future enhancement. @andy and I discussed some ideas for how we could achieve this, and we'd be glad to pitch in to make the shift.

Scotchester:
(Changing the approach would solve the issue Betsy raised, too.)

mistergone:
> @mistergone and @anselmbradford, you make a solid case. I am willing to get on board with configuring by page type behind the scenes, however I'd like to see the logic moved to the page type classes and send a context variable like `show_tags` to the template. 

I love that idea! I'll annoy you directly about it. 😄

mistergone:
@anselmbradford , @Scotchester , @chosak - I've changed around how this was done (which took a bit of learning, but I know more about Python and Django now!) and I think this is probably the best approach. Let me know what you think!

I need some assistance adding tests for the new function I added to `ref.py` before I'm finished, so any volunteers there would be appreciated!

mistergone:
It's also worth noting this is designed to be extendable - there are other features by page type which can be added to the function _ad infinitum_ (or perhaps _ad nauseum_ 😉 ).

cfgov/v1/util/ref.py
```diff
@@ -266,3 +266,12 @@ def get_category_children(category_names):
         dict(categories_dict[category]).keys()
         for category in category_names
     )))
+
+def get_fl_display_settings(filterable_page_type):
+    settings = {
+        'page_type': filterable_page_type,
+        'show_tags': True,
+        }
+    if "foia-freq-req-record" in filterable_page_type:
```
anselmbradford: Should this be `if "foia-freq-req-record" == filterable_page_type`?

cfgov/jinja2/v1/_includes/organisms/post-preview.html
```diff
@@ -63,6 +63,7 @@
     {% set date_desc = controls.post_date_description or post_date_description or 'Published' %}
     {% set cat_controls = controls.categories %}
     {% set show_categories = cat_controls.show_preview_categories if cat_controls is defined else true %}
+    {% set display_settings = get_fl_display_settings(controls.categories.page_type) if cat_controls is defined else get_fl_display_settings('None') %}
```
anselmbradford: Since you reference `cat_controls` already in here, should this be:

`{% set display_settings = get_fl_display_settings(cat_controls.page_type) if cat_controls is defined else get_fl_display_settings('None') %}`

anselmbradford: Actually, this could be shortened to this, right?

```
{% set display_settings = get_fl_display_settings(cat_controls.page_type if cat_controls is defined else 'None') %}
```

cfgov/jinja2/v1/_includes/organisms/post-preview.html
```diff
@@ -176,9 +177,8 @@ <h3 class="o-post-preview_title">
                       </a>
                 {% endif %}
             {% endfor %}
-            </div>
```
anselmbradford: Why was this closing div removed? Looking at the source file it doesn't look properly close anymore?

cfgov/jinja2/v1/_includes/organisms/post-preview.html
mistergone: Oh, whoops. I meant to divorce this from `cat_controls` entire. I had originally piggy-backed on it, but then I realized we might be able to iterate into eliminating that `cat_controls` line and integrating that `show_preview_categories` into the `display_settings` thing I'm doing now.

I'll remove the reference to `cat_controls` in my edits.

cfgov/jinja2/v1/_includes/organisms/post-preview.html
mistergone: Hrm. Looks like an accident. I don't recall seeing that in my `diff`s so I don't know how that happened.

cfgov/v1/util/ref.py
mistergone: I used `in` because it is used elsewhere in similar cases. There are also some extraneous characters in the output, so I think `in` was the reliable choice here. I'm all ears to best practice on this!

anselmbradford: I'm not sure either, but I thought `in` probably refers to a collection, so array or object. Whereas `==` would be string comparison. You may want to do string comparison with whitespace trimming, e.g.`filterable_page_type.strip()`.

chosak:
I've opened #4791 as an alternate approach to this and #4790.

mistergone:
Thanks for your clarification, Andy! I am going to make sure I understand the code you wrote, and then cancel my PRs and approve yours. 😄

mistergone:
This PR is being closed in favor of the solution in #4791

PR #4771, opened 2019-01-17
Break apart FilterableListControls

anselmbradford:
FilterableListControls contains the controls, notifications, results and pagination. It really should instead have a parent module, FilterableList, that includes those elements. 

## Additions

- Adds FilterableList module to encompass FilterableListControls, Notification, the results (post previews, activity log, or info units), and pagination.

## Removals

- filterable-list-controls.less and moves margin logic to `block` and the JS.
- Removes `filter_type` option (see changes from #4774)

## Changes

- Removes filterable list template `render` macro and calls `value` values directly, like other templates that render directly from the block.
- Update cucumber feature file to remove 0 from dates. 
- Hide result list when there are errors on the client-side.
- Add label to server-side field error message, for consistency with client-side error message (see note section below though for differences that remain).

## Testing

1. Pull branch and run `gulp clean && gulp build`
2. Check the following states.

### No posts state

Example URL: http://localhost:8000/about-us/events/
No content shown.
<table>
  <tr>
    <td></td>
    <td></td>
    <td colspan="4">notification</td>
    <td colspan="3">results</td>
    <td></td>
  </tr>
  <tr>
    <td></td>
    <td>filter control</td>
    <td>single field error</td>
    <td>multi-field error</td>
    <td>⚠</td>
    <td>✓</td>
    <td>post-preview</td>
    <td>activity-log</td>
    <td>info-unit</td>
    <td>pager</td>
  </tr>
  <tr>
    <td>NO-JS</td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
  </tr>
  <tr>
    <td>JS</td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
    <td></td>
  </tr>
</table>

---

Example URL: http://localhost:8000/about-us/events/
Content for "No posts message" in Wagtail added.
<table>
  <tr>
    <td></td>
    <td></td>
    <td colspan="4">notification</td>
    <td colspan="3">results</td>
    <td></td>
  </tr>
  <tr>
    <td></td>
    <td>filter control</td>
    <td>single field error</td>
    <td>multi-field error</td>
    <td>⚠</td>
    <td>✓</td>
    <td>post-preview</td>
    <td>activity-log</td>
    <td>info-unit</td>
    <td>pager</td>
  </tr>
  <tr>
    <td>NO-JS</td>
[...]

PR #2628, opened 2016-12-29
Remove deprecated Django calendar app

chosak:
Now that the leadership calendar has been moved to Wagtail, we can remove the deprecated Django calendar application. This PR removes the application completely.

## Removals

- Django `cal` app, and reference to it from settings.

## Testing

- Using a production data dump, verify that leadership calendar can be viewed at [http://localhost:8000/about-us/the-bureau/leadership-calendar/](http://localhost:8000/about-us/the-bureau/leadership-calendar/), and that the Django/Wagtail admins can be browsed with no legacy calendar references.

## Notes

- This PR will leave behind `cal` database tables and references to `cal` types in the Django `contenttypes` table. These can be manually removed.

## Checklist

* [X] Changes are limited to a single goal (no scope creep)
* [X] Code can be automatically merged (no conflicts)
* [X] Passes all existing automated tests
* [X] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)
* [X] Reviewers requested with the [Assignee tool](https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/) :arrow_right:

PR #4774, opened 2019-01-18
Remove form_type from FilterableList organism

chosak:
Note: this PR is to the **filterable_list_results_container** branch, onto @anselmbradford's PR #4771.

## Removals

This change removes the `form_type` ChoiceBlock from the FilterableList organism (formerly named FilterControls, before #4771). In the past, FilterableLists could either show a list of things or be used to generate PDFs (for example, for the leadership calendar). Since (I believe) #2628, we don't have any examples of this on the site anymore (@Scotchester can you confirm?) so this logic can be simplified.

This change requires a new migration due to the changing schema of the organism. In theory a data migration could also be created to remove any `form_type` data from the database, but this data won't cause a problem hanging around, so I lean against creating one unless others feel differently.

## Checklist


cfgov/jinja2/v1/_includes/organisms/filterable-list.html
```diff
@@ -83,29 +81,27 @@
...
-                    {# The if statement here allows access to the
-                        notification_opts variable outside of the for loop.
-                    #}
-                    {% if notification_opts.update({
-                        'type': 'success',
-                        'is_visible': true,
-                        'message': count ~ ' filtered result' ~ count | pluralize('s')
-                    }) %}{% endif %}
-                {% endif %}
+        {% if posts and form.is_valid() %}
```
anselmbradford: Looks like this should be `{% if posts is defined and form.is_valid() %}` otherwise http://localhost:8000/about-us/newsroom/?categories=blog&topics=financial-education&authors=holly-petraeus&from_date=2017 doesn't show the warning message.

chosak: Changed to more simply `{% if form.is_valid() %}` in 33319d5; `posts` should always be defined because `filter_data['page_set']` is always defined [here](https://github.com/cfpb/cfgov-refresh/blob/ff8f89f3ca29582ee3504ffb9d9086e6bef6d86c/cfgov/v1/util/filterable_list.py#L77).

PR #4758, opened 2019-01-11
Filterable Lists: Show a customizable message when there are no posts

schbetsy:
Previously, when a Filterable List page had no pages to filter (e.g. the Events page right now), the page would show the filter controls, which didn't make sense, and which had a broken Topic multi-select.

Now, a Wagtail user can specify a two-part message (header and body) in the "no posts message" and "no posts explanation" fields. That message will show up instead of the filter controls when there are no pages to show.

When there are pages to filter, nothing changes.

## Additions

- Added two fields to the Wagtail editing page for Filterable List pages. The "No posts message" and "No posts explanation" fields determine what message will be displayed when there are no pages to filter.

## Changes

- In the Filterable List Controls template, only show the controls if there are any posts to filter. If not, show the message specified in the "No posts message" field in Wagtail.

## Testing

1. Running this branch, visit http://localhost:8000/about-us/events/.
2. See that the filter controls aren't shown.
3. Log in to the Wagtail editor, navigate to the Events page. http://localhost:8000/admin/pages/1090/edit/
4. Enter text in the "No posts message" and "No posts explanation" fields. Save draft and publish.
    1. You might run into an issue where other fields on the page cause validation errors. I got around these by deleting the "Full Width Text" block at the bottom of the page and entering anything in the "Post date description" field.
5. Visit the events page again: http://localhost:8000/about-us/events/. See that your message appears on the page.

## Screenshots

Wagtail:
screen shot 2019-01-14 at 4 17 19 pm

Event page:
screen shot 2019-01-14 at 2 12 43 pm

## Todos

- I'll need to work with someone who has access to Wagtail to make fill in the content in Prod.

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
[...]

PR #4775, opened 2019-01-18
Multiselect: Remove client-side data check

anselmbradford:
Now that https://github.com/cfpb/cfgov-refresh/pull/4758 is in, which I think this logic was added for, we can strip out this client-side data check of the select menu options. If this throws an error it seems like it should be fixed on the backend data generating the markup.

## Removals

- `stringValid` utility class.
- Logic in multiselect to check value option values against a restricted character regex.

## Testing

1. `gulp clean && gulp build` should pass and the multiselect in the filter should work as before on http://localhost:8000/about-us/blog/


Scotchester:
I'm not sure I follow the connection between this and #4758.

Regardless, I agree that it doesn't seem like this code should be needed. The filter form that takes the value of the option and puts it into the query string should just encode the value. If we want people to avoid using characters that would need encoding for aesthetic reasons, that should be a check on the tags when they are created.

anselmbradford:
> I'm not sure I follow the connection between this and #4758.

The old behavior was to show the filter, but there was nothing to filter. Since, I think there were no posts to pull topics from (I think?), the multiselect would bail out and render as regular select markup (see [this wayback machine snapshot](https://web.archive.org/web/20170816162626/https://www.consumerfinance.gov/about-us/events/)). In theory that's good, but the page should really just not show the filter, which Betsy fixed.

chosak:
@wpears by creation time do you mean when a Wagtail editor adds a tag to a page? I believe that Wagtail uses [django-taggit](https://github.com/alex/django-taggit) on the backend and [jQuery Tag-it!](https://github.com/wagtail/wagtail/blob/master/wagtail/admin/static_src/wagtailadmin/js/vendor/tag-it.js) on the frontend. The interface in the admin to add or edit tags on a page looks like this:

image

When rendered on the page (in master) this looks like this:

image


The UI doesn't let you enter a `<` character.

wpears:
@chosak Cool. html escaping was the main worry here.

PR #4780, opened 2019-01-22
Updates Owning a Home copy to Buying a House

anselmbradford:
There is mixed use of "Owning a Home" and "Buying a House" (e.g. see breadcrumb link [here](https://www.consumerfinance.gov/owning-a-home/explore-rates/) and [here](https://www.consumerfinance.gov/owning-a-home/closing-disclosure/)). This PR updates copy to "Buying a House".

## Changes

- Changes `Owning a Home` copy to `Buying a House`.
- Changes `OAH` abbreviation to `BAH`.

## Testing

1. Check the /owning-a-home/ pages and see that they say Buying a House.

## Screenshots

Before:
screen shot 2019-01-22 at 6 51 54 pm

After:
screen shot 2019-01-22 at 6 51 36 pm


cfgov/v1/handlers/blocks/feedback.py
```diff
@@ -97,7 +97,7 @@ def success(self):
                      'message': "Be sure to also sign up for our email list "
                      "to get our blog posts and other tips about homebuying "
                      "and mortgages in your inbox. We'll also let you know "
-                     "when we make updates to Owning a Home."}
+                     "when we make updates to Buying a House."}
```
anselmbradford: @higs4281 I presume this is a safe change, but how is this tested?

higs4281: 👍 Looks safe to me. 
The feedback block is pretty [well-tested](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/tests/handlers/blocks/test_feedback.py), but we don't test the specific success language for suggestion responses.

PR #4781, opened 2019-01-23
Remove non-wagtail newsroom logic from utils.py

anselmbradford:
Since the structure of newsroom/press resources is identical to events/request a speaker, it looks like we can follow the structure there and fix the issue where "Press resources" doesn't get highlighted in the side navigation menu.

## Removals

- Legacy logic for the secondary navigation menu when the newsroom page was not in wagtail.

## Testing

1. Visit your local wagtail admin and set the following:
  - On the "newsroom" page in Wagtail, check this setting in the sidefoot:
screen shot 2019-01-22 at 7 04 47 pm
  - On the "press resources" page in Wagtail, uncheck this setting in the config:
screen shot 2019-01-22 at 7 11 21 pm

Publishing both pages should show a menu like below…

## Screenshots

screen shot 2019-01-22 at 7 18 16 pm

screen shot 2019-01-22 at 7 18 20 pm

## Todos

- When this goes to production the wagtail steps in the "testing" section above need to be performed. I'm not certain what the best way to do this is?

### DIFF
--- a/cfgov/v1/util/util.py
+++ b/cfgov/v1/util/util.py
@@ -55,27 +55,6 @@ def get_secondary_nav_items(request, current_page):
     # children
     has_children = False
 
-    # Handle the Newsroom page specially.
-    # TODO: Remove this ASAP once Press Resources gets its own Wagtail page
-    if page.slug == 'newsroom':
-        return [
-            {
-                'title': page.title,
-                'slug': page.slug,
-                'url': page.relative_url(request.site),
-                'children': [
-                    {
-                        'title': 'Press Resources',
-                        'slug': 'press-resources',
-                        'url': '/newsroom/press-resources/',
-                    }
-                ],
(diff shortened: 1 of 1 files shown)


anselmbradford:
> I would recommend running those steps manually on production (and on build). We could consider making a data migration to do it automatically but that feels like overkill.

Sounds good. Do I need to track that this gets done or is someone responsible for those kinds of Wagtail updates on prod?

chosak:
@anselmbradford at a minimum you should track it, at best you can feel free to do it! 😄

chosak:
@anselmbradford this has been deployed and I've made the changes on production to the [newsroom](https://www.consumerfinance.gov/about-us/newsroom/) and [press resources](https://www.consumerfinance.gov/about-us/newsroom/press-resources/) pages.

PR #4790, opened 2019-01-29
[HOLD] CFPB-Researchers date fix

mistergone:
__HOLD:__ This PR will wait for the work in #4772 

This PR adds a display setting (in the `get_fl_display_settings` function in `ref.py`) that allows for hiding dates in filtered lists. This is a specific need for the CFPB Researchers page, where the date is irrelevant to the viewer (although it is used in the back-end as a sorting mechanism).

To use this option, the CFPB Researchers page must be edited so that the filtered list uses the "CFPB Researchers" page type in the filterable list options.

## Additions
- Add `cfpb-researchers` to `page_types`
- Add an option to `get_fl_display_settings` for `show_date`
- Add logic to hide the date when `show_date` is `False`

## Testing
1. Pull down this branch and get it running!
2. Go into Wagtail, find the 'CFPB Researchers' page, scroll down, and under "FILTERABLE LIST" options, use the Page Type dropdown to select "CFPB Researchers."
3. Check out http://localhost:8000/data-research/cfpb-researchers/ and make sure there are no dates.
4. Check out other filterable list pages and make sure they have dates as expected.

## Screenshots
screen shot 2019-01-29 at 11 48 13 am

## Notes
-

## Todos
-

## Checklist
- [ ] PR has an informative and human-readable title
- [ ] Changes are limited to a single goal (no scope creep)
- [ ] Code can be automatically merged (no conflicts)
- [ ] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [ ] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated
- [ ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Internet Explorer 8, 9, 10, and 11
- [ ] Edge
- [ ] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS
- [ ] Flexible from small to large screens
- [ ] No linting errors or warnings
- [ ] JavaScript tests are passing

PR #1522, opened 2016-03-01
Fixes the Sublanding pages Text Introduction rendering bug

kurtw:
The text intro block was never being accessed in the template. Also, the sidebar-breakout organism template needs the context supplied to it. This fixes the issue by changing up the template quite a bit and moving the text introduction to only be located in the content streamfield. It does a data migration so nothing is lost.

## Additions

- supplies context to call to sidebar-breakout

## Removals

- conditional wrapping the blocks of a header
- text introduction from header streamfield on sublanding page

### Before pulling down
- Make or go to a sublanding page and create a text introduction in the header streamfield

### Pull down
- run `./cfgov/manage.py migrate`
- Return to the sublanding page and see if the text intro has moved to the content streamfield and that text intro has been removed from the header streamfield
- Go to the page and verify that the text intro renders properly
- Go back to the sublanding page and create a sidebar breakout.
 - verify the the text intro renders with the breakout
- Remove the Breakout and add a Hero to the page in the header streamfield and verify that renders properly as well

## Review

- @kave 
- @richaagarwal 

## Screenshots
Text intro with Sidebar Breakout
screen shot 2016-02-25 at 11 04 42 am
Text intro Without Sidebar Breakout
screen shot 2016-02-25 at 11 05 08 am

PR #4791, opened 2019-01-29
Allows page type-specific post preview rendering

chosak:
This change proposes an alternative to #4772 and #4790.

This change makes it possible to disable the rendering of post preview tags and dates when rendered as part of a filterable list. This is now possible via [code in the `FilterableList` StreamField block](https://github.com/chosak/cfgov-refresh/blob/98b59e44bdb6783005dc0f87f30e7c904dbad6f3/cfgov/v1/atomic_elements/organisms.py#L918), and is dependent on the "page_type" set on the FL, which is an optional setting that a page editor can define.

This change adds a new page type of "cfpb-researchers" and disables rendering of post preview dates for it (#4790). It also disables rendering of post preview tags for the "foia-freq-req-record" page type (#4772).

A side effect of this change is to add a proper Wagtail StreamField block template for the `PostPreviewSnapshot` block type.

## Additions

- Creates new template for `PostPreviewSnapshot` StreamField blocks, extracting the existing code in the `SublandingPage` template, which is the only place these blocks can be placed. This allows this block to be rendered using Wagtail's built-in `{% include_block %}` syntax, and, in future, to be placed on other pages if desired.
- Adds a new FilterableList page type for "CFPB Researchers", that does not display post dates.

## Changes

- Modifies the rendering of FilterableLists set to the "FOIA Frequently Requested Record" so that post previews do not show tags.

## Testing

1. To verify default functioning of FilterableLists shows dates and tags, visit [the Newsroom locally](http://localhost:8000/about-us/newsroom/) and compare with [the version in production](https://www.consumerfinance.gov/about-us/newsroom/).
2. To see the changed behavior of the FOIA page, compare [its local version](http://localhost:8000/foia-requests/foia-electronic-reading-room/) with [the version in production](https://www.consumerfinance.gov/foia-requests/foia-electronic-reading-room/) and confirm that tags no longer appear.
3. To see the behavior for the new CFPB Researchers tag, [edit the page locally](http://localhost:8000/admin/pages/4833/edit/) and set its FilterableList page type to "CFPB Researchers". Preview the page (you may need to set a "Post date description", which will be ignored) and compare its local version to [the version in production](https://www.consumerfinance.gov/data-research/cfpb-researchers/) and confirm that dates no longer appear.
4. To confirm proper functioning of the new `PostPreviewSnapshot` block template, test the two Sublanding pages that currently use that block to verify that the local page version matches the version in production. Note especially that the HTML rendering and tag links are the same.

    1. Amicus program: [local](http://localhost:8000/policy-compliance/amicus/), [production](https://www.consumerfinance.gov/policy-compliance/amicus/)
    2. Notice and opportunities to comment: [local](http://localhost:8000/policy-compliance/notice-opportunities-comment/), [production](https://www.consumerfinance.gov/policy-compliance/notice-opportunities-comment)

## Notes

This change attempts to make `PostPreviewSnapshot` into a proper Wagtail StreamField block. We don't currently have a `PostPreview` block -- we have a template, but it's not a StreamField block. For this reason, it seemed more straightforward to add `show_date` and `show_tags` as parameters to that template's top-level macro than to try to add it to an existing structure or create a whole new structure for display settings.

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/filterable-list.html
+++ b/cfgov/jinja2/v1/_includes/organisms/filterable-list.html
@@ -31,6 +31,10 @@
 
    filter_data.page_set: Result set of posts from the filtered content.
 
+   show_post_dates: Whether to render a date in post previews.
+
+   show_post_tags: Whether to render the tags in post previews.
+
    ========================================================================== #}
 
 {% import 'organisms/filterable-list-controls.html' as filterable_list_controls with context %}
@@ -181,9 +185,14 @@
                         </div>
                     </div>
                 {% else %}
-                    {% import 'organisms/post-preview.html' as post_preview with context %}
+                    {% import 'organisms/post-preview.html' as post_preview without context %}
                     {% for post in posts %}
-                        {{ post_preview.render(post, value) }}
+                        {{ post_preview.render(
+                            post,
+                            value,
+                            show_date=show_post_dates|default(true),
+                            show_tags=show_post_tags|default(true)
+                        ) }}
                     {% endfor %}
                 {% endif %}
 
--- a/cfgov/jinja2/v1/_includes/organisms/filterable-list.html
+++ b/cfgov/jinja2/v1/_includes/organisms/filterable-list.html
@@ -190,8 +190,8 @@
                         {{ post_preview.render(
                             post,
                             value,
-                            show_date=show_post_dates|default(true),
-                            show_tags=show_post_tags|default(true)
+                            show_date=show_post_dates | default(true),
+                            show_tags=show_post_tags | default(true)
                         ) }}
                     {% endfor %}
                 {% endif %}
--- a/cfgov/jinja2/v1/sublanding-page/index.html
+++ b/cfgov/jinja2/v1/sublanding-page/index.html
@@ -32,14 +32,7 @@
         {% elif block.block_type == 'feedback' %}
             {{- form_block.render(block, 'content', loop.index0) -}}
         {% elif 'post_preview_snapshot' in block.block_type %}
-            <div class="block
-                        {{ block.block.meta.classname if block.block.meta.classname else '' }}">
-                {% set limit = block.value.limit | int %}
-                {% set posts = page.get_browsefilterable_posts(limit) %}
-                {% for post in posts %}
-                    {{ post_preview.render(post, controls=none, url=pageurl(post.parent()), post_date_description=block.value.post_date_description) }}
-                {% endfor %}
-             </div>
+            {% include_block block %}
         {% elif 'filter_controls' in block.block_type %}
             <div class="block block__flush-top">
                 {% include_block block %}
--- a/cfgov/v1/util/ref.py
+++ b/cfgov/v1/util/ref.py
@@ -46,6 +46,7 @@ page_types = [
     ('rule-under-dev', 'Rule Under Development'),
     ('story', 'Story'),
     ('ask', 'Ask CFPB'),
+    ('cfpb-researchers', 'CFPB Researchers'),
 ]
 
 fcm_types = [
--- a/cfgov/jinja2/v1/_includes/organisms/post-preview-snapshot.html
+++ b/cfgov/jinja2/v1/_includes/organisms/post-preview-snapshot.html
@@ -1,0 +1,33 @@
+{# ==========================================================================
+
+   Post preview snapshot
+
+   ==========================================================================
+
+   Description:
+
+   Render a list of post previews for a page when given:
+
+   value: Data object from an instance of the PostPreviewControls StreamField
+          block.
+
+   value.limit: Number of posts to show.
+
+   value.post_date_description: Text to display next to date, e.g. "Published".
+
+   ========================================================================== #}
+
+{% import 'organisms/post-preview.html' as post_preview with context %}
+
+<div class="block">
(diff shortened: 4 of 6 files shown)


cfgov/jinja2/v1/sublanding-page/index.html
```diff
@@ -32,14 +32,7 @@
         {% elif block.block_type == 'feedback' %}
             {{- form_block.render(block, 'content', loop.index0) -}}
         {% elif 'post_preview_snapshot' in block.block_type %}
-            <div class="block
-                        {{ block.block.meta.classname if block.block.meta.classname else '' }}">
```
chosak: This second class doesn't ever seem to get used in practice; both Sublanding pages that use `PostPreviewSnapshot` have `<div class="block ">`. @anselmbradford @Scotchester should something else go here [in its new location](https://github.com/cfpb/cfgov-refresh/blob/98b59e44bdb6783005dc0f87f30e7c904dbad6f3/cfgov/jinja2/v1/_includes/organisms/post-preview-snapshot.html#L22)?

cfgov/jinja2/v1/_includes/organisms/filterable-list.html
```diff
@@ -181,9 +185,14 @@
...
                 {% else %}
-                    {% import 'organisms/post-preview.html' as post_preview with context %}
+                    {% import 'organisms/post-preview.html' as post_preview without context %}
                     {% for post in posts %}
-                        {{ post_preview.render(post, value) }}
+                        {{ post_preview.render(
+                            post,
+                            value,
+                            show_date=show_post_dates|default(true),
+                            show_tags=show_post_tags|default(true)
```
anselmbradford: Trivial and we don't adhere to it, but spacing out filters can make them stand out a bit more `show_tags=show_post_tags | default(true)`

cfgov/jinja2/v1/sublanding-page/index.html
anselmbradford: Hmm it appears [this](https://github.com/cfpb/cfgov-refresh/pull/1522) is where it was added. Maybe something to do with that layout where the breakout sidebar was contained in the same box as the text intro. I think it's fine to remove.

I noticed something unrelated looking at this. The text introduction has two `block__flush-top` classes added. Anyone see why that's the case?
screen shot 2019-01-30 at 10 17 30 am

cfgov/v1/util/ref.py
```diff
@@ -46,6 +46,7 @@
     ('rule-under-dev', 'Rule Under Development'),
     ('story', 'Story'),
     ('ask', 'Ask CFPB'),
+    ('cfpb-researchers', 'CFPB Researchers'),
```
anselmbradford: Can the page types list be alphabetized?

chosak: I wondered that too but was hesitant to reorder this list without understanding how else it's used or what content editor expectations are around this list. Who would be the best person to opine on this?

cfgov/jinja2/v1/sublanding-page/index.html
chosak: @anselmbradford it looks like that gets rendered in the Sublanding page template [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/sublanding-page/index.html#L10), which calls [this macro](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/templates/render_block.html). That renders `block__flush-top` twice, once [on this line](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/templates/render_block.html#L3) because it's the first element, and then again [on this line](https://github.com/cfpb/cfgov-refresh/blob/026ff4960ebb533e99b1b42335b4afd3b970c8b5/cfgov/jinja2/v1/_includes/templates/render_block.html#L12) because of [this `Meta.classname` definition](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/atomic_elements/molecules.py#L111).

cfgov/v1/util/ref.py
mistergone: I'd defer to @Scotchester but it seems to me alphabetical (by displayed text, the second parameter) would be the way to go, since that would make it easy to find what you're looking for in the Wagtail editor dropdown.

cfgov/jinja2/v1/sublanding-page/index.html
anselmbradford: How best to fix that?

chosak: As with all of these, I'd like to deprecate our use of `render_block` and move all of the StreamField blocks to using the Wagtail standard `{% include_block %}`, if we can. We could do that for `TextIntroduction` - does this actually cause a problem, or is it just unnecessary duplication?

Scotchester: It's just unnecessary duplication, but yeah, I'd say it gives us added motivation to migrate `TextIntroduction` to `{% include_block %}`.

cfgov/v1/util/ref.py
Scotchester: I agree that that would be the best way to alphabetize it, and I can't see why it would present a problem to do so. Surely no code relies on the order of the list.

PR #4797, opened 2019-02-04
Simplify RateCheckerChartMenu and fix BAH linter warnings

anselmbradford:
## Additions

- Adds missing jsdocs to BAH methods.
- Adds missing MIME type for PNG in chart export.

## Removals

- Removes "RateCheckerChartMenu.STATES" export as it was only exported for the tests.
- Removes test of internal private method "_setState".
- Unused references in Slider.js.

## Changes

- Converts open/closed state of chart download menu to simple strings from an object to remove need for Object.assign.
- Converts logic stored in data- attributes to directly use textContent of list nodes in download menu.
- Removes complexity warning from exportChart method by converting switch into if/else and an object lookup hash.
- Update RateCheckerChartMenu test markup structure to follow that of production.

## Testing

1. Pull the branch and `gulp clean && gulp build`
2. The "Download chart" options in the upper-right of the chart on http://localhost:8000/owning-a-home/explore-rates/ should work (first three will download an image, last will open the print dialog box). Clicking the "Download chart" text will open/close the menu.
3. `gulp test:unit` should pass.

## Screenshots

screen shot 2019-02-04 at 3 47 22 pm

### DIFF
--- a/cfgov/jinja2/v1/owning-a-home/explore-rates/index.html
+++ b/cfgov/jinja2/v1/owning-a-home/explore-rates/index.html
@@ -341,10 +341,10 @@ home price, down payment, and more can affect mortgage interest rates.
                                 Download chart
                             </button>
                             <ul class="chart-menu_options">
-                                <li data-export-type="0">PNG</li>
-                                <li data-export-type="1">SVG</li>
-                                <li data-export-type="2">JPEG</li>
-                                <li data-export-type="3">Print chart</li>
+                                <li>PNG</li>
+                                <li>SVG</li>
+                                <li>JPEG</li>
+                                <li>Print chart</li>
                             </ul>
                         </div>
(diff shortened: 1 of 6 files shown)


anselmbradford:
codecov can be ignored.

anselmbradford:
> Probably out of scope for this PR, but it'd be nice if the chart exports included a title and the x-axis label. Also, why does the print version have blue instead of green bars?

Hmmmmmmm got me, but good observation! You're saying you want to print what you see on screen when you click print?? 😜

anselmbradford:
Issues for noted issues:
https://[GHE]/CFGOV/platform/issues/3281
https://[GHE]/CFGOV/platform/issues/3280

Scotchester:
Thanks for opening those!

PR #4800, opened 2019-02-05
Add requirements.txt file for Snyk to consume

Scotchester:
This file's purpose is to be checked for vulnerabilities by Snyk, which requires a file named `requirements.txt`.

## Additions

- `requirements.txt` file for Snyk

## Testing

We won't be able to test this until we merge and use the Snyk UI to try and add this file to watch it.

## Notes

- I'm not certain Snyk will use the `-r` directive to pull in the dependencies from the other files. I guess we'll have to try it and see.
- Another potential solution would be to put each requirements file in a subdirectory of its name, and then have the files all be named `requirements.txt`. We could then add one or more of those files to what Snyk is monitoring. Both that and the solution proposed here are inelegant, but this seems preferable to me.

## Checklist

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -1,0 +1,4 @@
+# This file's purpose is to be checked for vulnerabilities by Snyk,
+# which requires a file named requirements.txt
+
+-r deployment.txt


marcesher:
@Scotchester this makes sense to me, provided that snyk is smart enough to follow included dependencies files.

ascott1:
I think that we need to merge it to test. An alternative would be to create a dummy repo for testing this approach.

willbarton:
image

Scotchester:
No good :(

screen shot 2019-02-06 at 11 44 38

marcesher:
huh, bummer.

I wonder:

1. perhaps we can submit a support ticket, user voice, or some other thing to snyk for consideration
2. would github actions open up an easy way to just automatically copy the contents of deploy.txt into requirements.txt when changes are made to it? I'm thinking: what are options we have for getting the benefit of snyk here without creating a ton of work on maintaining duplicate files

anselmbradford:
Hmm my latest PR is getting this now:
screen shot 2019-02-06 at 1 50 44 pm

ascott1:
👍 Thanks for taking a wack at this @Scotchester

willbarton:
I feel really uncomfortable with option 2, since that would require us to have our requirements specified in two places *committed in the repo* that could easily get out-of-sync if someone's not careful.

chosak:
Snyk's command line tool offers additional flexibility that the web UI does not, for example [specifying alternate filenames for requirements.txt](https://snyk.io/docs/cli-test/#manifest-files-with-custom-names). Could we use the command line tool to monitor for vulnerabilities instead of relying on the web UI?

willbarton:
I would add, GitHub seems to have no issue reading separate requirements files:

https://github.com/cfpb/cfgov-refresh/network/alerts?q=is%3Aclosed

anselmbradford:
> Snyk's command line tool offers additional flexibility that the web UI does not, for example [specifying alternate filenames for requirements.txt](https://snyk.io/docs/cli-test/#manifest-files-with-custom-names). Could we use the command line tool to monitor for vulnerabilities instead of relying on the web UI?

For the node modules we already have the snyk cli set up at `npm test`, so this may be fairly straightforward to adjust.

Scotchester:
The nice thing about getting it monitored in the Snyk.io app is that it monitors proactively and emails us, it doesn't just happen on PR checks.

But, perhaps we can live with only doing it on PR checks for cfgov-refresh, given the amount of traffic it gets.

willbarton:
What's the advantage of Snyk over GitHub's built-in vulnerability monitoring?

Scotchester:
> What's the advantage of Snyk over GitHub's built-in vulnerability monitoring?

I'm not sure. Possibly none? @ascott1?

ascott1:
Snyk does a much more thorough job. It uses both CVE and NVD's vulnerability databases as well as maintaining their own independent list that is compiled from several other sources and security researches at Snyk. IIRC Github only uses CVE.

ascott1:
And I agree that we should work with Snyk to encourage them to implement a way to scan multiple and uniquely named requirement files. In the meantime, I don't think it's _that_ awful of an anti-pattern to explore pulling the production dependencies into a `requirements.txt` file. 

That said, I would also be comfortable with running the `snyk-cli` on Python dependencies as part of Travis build, if it's that big of a sticking point for y'all.

willbarton:
Let's run `snyk-cli` in Travis then. I'm strongly opposed to committing a generated `requirements.txt` file.

PR #4802, opened 2019-02-05
Untangle BAH apps and routes paths

anselmbradford:
It appears that the BAH landing page and the mortgage-estimate page should be showing [a popup](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/base.py#L724-L725), however, mortgage-estimate is not showing a popup.

Also, the page-specific form-explainer initialization code in the /routes/ path can be consolidated in the BAH apps code, since those pages have their own templates and /routes/ is only for pages that don't have specific templates.

## Additions

- Adds add-email-popup initialization code to BAH landing page and mortgage-estimate page.

## Changes

- Move form-explainer initialization code to BAH apps code and imports into the closing-disclosures and loan-estimate templates.
- Minor linter fixes.

## Testing

2. http://localhost:8000/owning-a-home/ and http://localhost:8000/owning-a-home/mortgage-estimate/ should both show the email popup when scrolling down the page (you'll need to "Clear site data" in the Application tab in dev console between visits).
3. Visit http://localhost:8000/owning-a-home/loan-estimate/, http://localhost:8000/owning-a-home/closing-disclosure/, and http://localhost:8000/owning-a-home/explore-rates/ and see that those tools work. Check that the email signup works.

## Screenshots

screen shot 2019-02-05 at 4 10 57 pm

### DIFF
--- a/cfgov/jinja2/v1/owning-a-home/closing-disclosure/index.html
+++ b/cfgov/jinja2/v1/owning-a-home/closing-disclosure/index.html
@@ -23,7 +23,8 @@
         !function(){
           {# Include site-wide JavaScript. #}
           var s = [
-            '{{ static('apps/owning-a-home/js/common.js') }}'
+            '{{ static('apps/owning-a-home/js/common.js') }}',
+            '{{ static('apps/owning-a-home/js/form-explainer/index.js') }}'
           ];
           jsl(s);
         }()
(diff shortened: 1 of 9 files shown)


anselmbradford:
codecov can be ignored. It's because of the addition of an `init` method to the form-explainer, but that whole class isn't tested so that's a separate issue.

PR #4804, opened 2019-02-05
Data migrations to convert Image & Text and Link Blob Groups to Info Unit Groups

Scotchester:
[Info Unit Groups came on the scene quite a while ago]() and we are (I am) long overdue for migrating the data out of the blocks they replaced. This PR begins that work.

## Additions

- Data migration for Image & Text Groups to Info Unit Groups
- Data Migration for Link Blob Groups to Info Unit Groups

## Changes

-  Modify `get_stream_data` migration utility to handle corrupt revisions
  - Specifically, handle the case where a revision is missing the StreamField named `field_name`.

## Testing

1. Pull branch
1. Refresh your database (which runs the migrations)
1. Look at a wide range of pages to confirm that data has transferred as expected (use the Block Inventory on Content to look for pages to check)
1. Update your local Block Inventory with `./cfgov/manage.py block_inventory` and see that your local Block Inventory menu no longer offers to search for the following:
   - `v1.atomic_elements.molecules.HalfWidthLinkBlob` – It's still there! See note below.
   - `v1.atomic_elements.molecules.ImageText2575`
   - `v1.atomic_elements.molecules.ImageText5050`
   - `v1.atomic_elements.molecules.ThirdWidthLinkBlob`
   - `v1.atomic_elements.organisms.HalfWidthLinkBlobGroup`
   - `v1.atomic_elements.organisms.ImageText2575Group`
   - `v1.atomic_elements.organisms.ImageText5050Group`
   - `v1.atomic_elements.organisms.ThirdWidthLinkBlobGroup`

## Notes

- One instance of `HalfWidthLinkBlob`s remains – the weird special case on the `HomePage`'s `header` StreamField. We'll take care of that in a future PR.

## Todos

In this PR:
- [x] Write tests for the migration functions
- [x] Remove the organism classes

In future PRs:
- [ ] Remove the molecule classes
- [ ] Convert the singleton Half Width Link Blob molecules on the `HomePage` to Info Units
- [ ] Remove the old Info Unit and Info Unit Group templates

## Checklist


cfgov/v1/migrations/0136_migrate_imagetextgroups_to_infounitgroup.py
```diff
@@ -0,0 +1,153 @@
+from django.db import migrations
+
+from v1.util.migrations import get_stream_data, set_stream_data
+
+
```
chosak: It would be extremely helpful both to review this PR and for future developers looking at this migration to include a comment here documenting the different formats of these two things. For example, something like:

```py
"""
ImageText2575Group organisms has this format:

{
    # string, not required
    'heading': 'Heading',

    # boolean, not required
    'link_image_and_heading': True,

    # list of ImageText2575 molecules
    'image_texts': [
        {
            # string, not required
            'heading': 'Heading',

            # rich text, not required
            'body': 'Rich text, maybe with <a linktype="page" id="123">links</a>',

            # etc....
        },
        ...
    ]
}
"""
```

I know this information exists in the source files but having it in the migration would make it easier to understand what this is doing. Would adding something like this to both of these files be too much to ask?

cfgov/v1/migrations/0136_migrate_imagetextgroups_to_infounitgroup.py
```diff
@@ -0,0 +1,153 @@
...
+    new_value['format'] = new_format
+
+    return new_value
+
+
+def migrate_streamfield_forward(page_or_revision, streamfield_name, new_type):
+    """ Migrate a StreamField belonging to the page or revision """
+    old_stream_data = get_stream_data(page_or_revision, streamfield_name)
+
+    if old_stream_data != []:
```
chosak: Can this be `if old_stream_data:`? Is `old_stream_data` ever `None`?

cfgov/v1/migrations/0136_migrate_imagetextgroups_to_infounitgroup.py
Scotchester: I'll see if I can do this without too much pain :)

Scotchester:
@chosak Thanks for your feedback! I've addressed it in 217222a.

WIP label removed; this is ready for a final review.

cfgov/v1/migrations/0136_migrate_imagetextgroups_to_infounitgroup.py
```diff
@@ -0,0 +1,278 @@
...
+        ('v1', 'BlogPage', 'content'),
+        ('v1', 'BrowsePage', 'content'),
+        ('v1', 'LandingPage', 'content'),
+        ('v1', 'LearnPage', 'content'),
+        ('v1', 'SublandingPage', 'content'),
+    ]
+
+    revision_model = apps.get_model('wagtailcore.PageRevision')
+
+    for app, page_type, streamfield_name in models_and_fields:
```
willbarton: There's quite a bit of code here that's duplicated between these two migrations and https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/util/migrations.py. I realize why — the need to pass `new_type` based on the page mode... but I can't help feeling that `migrate_page_types_and_fields()` could be used with `partial` passing `new_type`. Something akin to:

```python
from functools import partial

def migrate_streamfield_forward(page_or_revision, data, new_type=None):
    …

    other_page_mapper = partial(migrate_streamfield_forward, new_type=info_unit_group)
    learn_page_mapper = partial(migrate_streamfield_forward, new_type=info_unit_group_25_75_only)
    migrate_page_types_and_fields(
        apps,
        [
            ('v1', 'DocumentDetailPage', 'sidefoot', 'related_metadata'),
            ('v1', 'BlogPage', 'content'),
            ('v1', 'BrowsePage', 'content'),
            ('v1', 'LandingPage', 'content'),
            ('v1', 'SublandingPage', 'content'),
        ],
        other_page_mapper
    )
    migrate_page_types_and_fields(
        apps,
        [
            ('v1', 'LearnPage', 'content'),
        ],
        learn_page_mapper
    )
```

But I don't feel strongly enough to recommend you re-work all of this. It's good as it is.

Scotchester:
> I do wonder about the lack of backwards migrations — I can't think of when we've ever needed them, but I feel a slight discomfort without them (perhaps I'll bring this to the BEWD huddle today). Do you consider the image text group to info unit group migration to be too lossy? (Looks like `is_widescreen` and `is_button` are lost, but is that because they're already unused?)

`is_button` is a design pattern we no longer want to support, and `is_widescreen` is not needed in the IUG implementation.

The crux of why I didn't want to get into a backward migration is that we're taking four existing blocks and moving them all to one. It would take a lot of logic to look at an InfoUnitGroup and figure out which of the previous two types of blocks to go backward to. Also, IUGs are already in widespread use with data that doesn't map back neatly to the previous blocks' fields (heading icons at the group level, group intro RTFs, etc.)

Thanks for the review. I will update those docstrings, clear up the `browse_page.py` conflict, and bump the migrations up a number.

PR #4810, opened 2019-02-07
Update browser testing checklist in PR template

Scotchester:
IE8 is no longer testable with Sauce Labs.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -45,13 +45,13 @@
 
 ### Browsers
 
-- [ ] Chrome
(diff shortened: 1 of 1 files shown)


anselmbradford:
FWIW IE8 can be used via VirtualBox with https://github.com/xdissent/ievms

Scotchester:
> FWIW IE8 can be used via VirtualBox with https://github.com/xdissent/ievms

Is that a "Request Changes" review? 😁 

Those things are so hard to maintain, and our IE8 traffic is so low at this point—0.13% (that's thirteen hundredths of one percent)—I think it's justifiable to stop testing in IE8.

PR #4816, opened 2019-02-08
Remove UNDEFINED from standard-type

anselmbradford:
Our linter flags `undefined` values because of the https://eslint.org/docs/rules/no-undefined rule. 
To get around this we utilize an uninitialized variable to hold the value of `undefined`. However, we inconsistently handle this, sometimes defining a `let UNDEFINED` variable in a module and sometimes importing the standard-type module and referencing the one in there. While it may make sense to always use the one from the standard-type module, the additional dependency complexity of doing that versus the cost of creating a new variable each time does not make sense, at very least from a pure number-of-lines-of-code perspective.

## Removals

- Removes `UNDEFINED` value from the `standard-type` module.

## Testing

1. Tests should pass and site should load error-free.

### DIFF
--- a/cfgov/unprocessed/js/molecules/Autocomplete.js
+++ b/cfgov/unprocessed/js/molecules/Autocomplete.js
@@ -3,7 +3,6 @@ import { assign } from '../modules/util/assign';
 import * as atomicHelpers from '../modules/util/atomic-helpers';
 import { ajaxRequest } from '../modules/util/ajax-request';
 import { bindEvent } from '../modules/util/dom-events';
-import { UNDEFINED } from '../modules/util/standard-type';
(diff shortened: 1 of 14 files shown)


cfgov/unprocessed/js/molecules/Autocomplete.js
```diff
@@ -86,6 +85,7 @@ function Autocomplete( element, opts ) {
    */
   function init() {
     if ( !atomicHelpers.setInitFlag( _dom ) ) {
+      let UNDEFINED;
       return UNDEFINED;
```
anselmbradford: For this use, we could just `return`, but I think it's slightly clearer to set what's returned.

PR #4817, opened 2019-02-11
Multiselect: Add model

anselmbradford:
Fixes https://[GHE]/CFGOV/hmda-migration/issues/16

## Additions

- Adds MultiselectModel for managing state of the multiselect items, including filtering, selection, and highlighting.

## Changes

- Changes the search behavior to remove the minimum characters needed to perform a search (was 3, now it's 1).
- Prefixes multiselect filter utility classes with `u-`.

## Testing

1. `gulp test:unit --specs=js/molecules/MultiselectModel-spec.js` should have 100% passing coverage.
2. No errors in `gulp test:unit`.
3. `gulp clean && gulp build`
4. Compare filter on http://localhost:8000/about-us/newsroom/ to production. The one difference should be that when you start typing into the input, the list filters, whereas on production it will only filter after minimum of three characters are entered. Try hitting "enter" and tabbing and using the up/down arrows in the list, while being aware of existing known issues (https://github.cfpb.gov/CFGOV/platform/issues/3282, https://github.cfpb.gov/CFGOV/platform/issues/3283).

## Screenshots

filter-search

### DIFF
--- a/cfgov/unprocessed/css/molecules/multiselect.less
+++ b/cfgov/unprocessed/css/molecules/multiselect.less
@@ -58,11 +58,11 @@ select.o-multiselect {
             margin-bottom: 0;
         }
 
-        &.filtered li:not( .filter-match ) {
+        &.u-filtered li:not( .u-filter-match ) {
             display: none;
         }
 
-        &.no-results,
+        &.u-no-results,
         &.max-selections {
             li {
                 display: none;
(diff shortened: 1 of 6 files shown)


anselmbradford:
Codecov can be ignored since it has no base coverage for MultiselectModel.
Also, accesslint is stalled and can be ignored too.

PR #4818, opened 2019-02-12
Change page model to handle editing of Ask entries

higs4281:
This is the first step in migrating Ask content to a simpler editing
structure. It exposes content on the AnswerPage and allows version
comparisons.

Metadata will be handled in subsequent steps.

Notes
* The 'tags' and 'tags_es' properties on the Answer model, which are
used in indexing, are being changed to 'clean_tags' and 'clean_tags_es'
to avoid a future collision with the 'tags' field that exists on all Wagtail pages.
* I turned off exports of the content fields (question, statement, snippet, answer) from the Answer model. so that page content won't get clobbered by any changes to an Answer object.
* I added helper text to the AnswerPage's search_tags field to make clear that it's not being used yet.


cfgov/ask_cfpb/models/django.py
```diff
@@ -547,6 +546,10 @@ def has_live_page(self):
                 return True
         return False
 
+    # def sync_related_questions(self, page):
```
richaagarwal: Can this commented out code be deleted for now?

cfgov/ask_cfpb/models/pages.py
```diff
@@ -523,13 +528,23 @@ class AnswerPage(CFGOVPage):
...
-    snippet = RichTextField(
-        blank=True, help_text='Optional answer intro', editable=False)
-    created_at = models.DateTimeField(auto_now_add=True)
-    updated_at = models.DateTimeField(auto_now=True)
-    publish_date = models.DateTimeField(default=timezone.now)
+    last_edited = models.DateField(
+        blank=True,
+        null=True,
+        help_text="Change the date to today if you make a significant change.")
+    question = models.TextField(blank=True)
```
richaagarwal: Is there a reason the `question` field becomes a `TextField` in place of a `RichTextField`?

higs4281: The question has always been a TextField in the Answer model. We didn't see a need to allow  markup in questions.

cfgov/ask_cfpb/models/django.py
higs4281: Gone!

cfgov/ask_cfpb/models/django.py
```diff
@@ -556,14 +555,20 @@ def create_or_update_page(self, language=None):
             _parent = english_parent
             _slug = self.slug
             _question = self.question
-            _snippet = self.snippet
-            _answer = self.answer
+            # _snippet = self.snippet
```
richaagarwal: Do you want to remove these or leave them commented out?  In the next iteration we'll be removing this code entirely so it doesn't matter too much either way but might be easier to remove.

cfgov/ask_cfpb/models/pages.py
```diff
@@ -4,19 +4,20 @@
 import re
 from six.moves.urllib.parse import urlparse
 
+# from django import forms
```
richaagarwal: Can this line be deleted?

higs4281: good catch. gone!

PR #4819, opened 2019-02-12
Multiselect: Fix incorrect event dispatching error

anselmbradford:
Fixes https://[GHE]/CFGOV/platform/issues/3282
Introduced in https://github.com/cfpb/cfgov-refresh/pull/1684

## Changes

- Fixes issue where checkbox input was incorrectly dispatching a string as an event, not an event object.
- Removes `dom-traverse` dependency from Multiselect.
- Enabled keyboard interaction unit test.

## Testing

1. Visit /blog
2. Expand filter and click into Topic multiselect.
3. Open the dev console.
3. Hit "enter" key once.
4. Hit the "enter" key again.
5. There should be no error in the console and the checkbox should check and add a selection.

## Screenshots

Before:
input-before

After:
input-after


anselmbradford:
Codecov can be ignored, it's flagging on code that is behind interactions that aren't covered by unit tests.

PR #4820, opened 2019-02-13
Ensure our data migrations migrate blocks that are children of other blocks

willbarton:
This PR adds some recursion to our Wagtail StreamField-specific data migration utilities to ensure that we can migrate blocks that are children of other blocks.

By way of illustration, [`FullWidthText` block](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/atomic_elements/organisms.py#L712) itself contains a variety of other blocks that could also exist within a StreamField directly. So if we want to migrate all `EmailSignUp`s, we would previously have had to look inside the `FullWidthText` separately for its `EmailSignUp`. 

For example, `LearnPage` contains an `email_signup` `EmailSignUp` block in its `content` StreamField. `BlogPage` has both a `FullWidthText` and a `EmailSignUp` block in its `content` StreamField. If we wanted to perform the same migration on the `EmailSignUp` in both places, we'd have to extract the `EmailSignUp` block from the `FullWidthText` block first. Perhaps something like:

```python
def full_width_text_mapper(page_or_revision, data):
    email_signup = data['email_signup']
    data['email_signup'] = email_signup_mapper(page_or_revision, email_signup)
    return data

def email_signup_mapper(page_or_revision, data):
    # Mung mung
    return data

    migrate_page_types_and_fields(
        apps,
        [(v1', 'BlogPage', 'header', 'email_signup'),],
        full_width_text_mapper
    )
    migrate_page_types_and_fields(
        apps,
        [(v1', 'BlogPage', 'content', 'full_width_text'),],
        full_width_text_mapper
    )
```

With this change, we just have to write one mapper function and have one call to `migrate_page_types_and_fields`, and include the path from parent to child block names in a list:

```python
def email_signup_mapper(page_or_revision, data):
    # Mung mung
    return data

    migrate_page_types_and_fields(
        apps,
        [
            ('v1', 'BlogPage', 'content', ['email_signup', ]),
            ('v1', 'BlogPage', 'content', ['full_width_text', 'email_signup', ]),
        ]
        full_width_text_mapper
    )
```

## Checklist

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -156,20 +156,18 @@ from v1.util.migrations import migrate_page_types_and_fields
 
 
 def forward_mapper(page_or_revision, data):
-    data = dict(data)
     # Manipulate the stream block data forwards
     return data
 
 
 def backward_mapper(page_or_revision, data):
-    data = dict(data)
     # Manipulate the stream block data backwards
     return data
 
 
 def forwards(apps, schema_editor):
     page_types_and_fields = [
-        ('myapp', 'MyPage', 'field_name', 'block_type'),
+        ('myapp', 'MyPage', 'field_name', 'block_name'),
     ]
     migrate_page_types_and_fields(
         apps,
@@ -180,7 +178,7 @@ def forwards(apps, schema_editor):
 
 def backwards(apps, schema_editor):
     page_types_and_fields = [
-        ('myapp', 'MyPage', 'field_name', 'block_type'),
+        ('myapp', 'MyPage', 'field_name', 'block_name'),
     ]
     migrate_page_types_and_fields(
         apps,
@@ -197,12 +195,29 @@ class Migration(migrations.Migration):
 ```
 
 `field_name` is the name of the StreamField on the Page model that contains the blocks to migrate. 
-`block_type` is the type of block within a StreamField that contains the data to be migrated.
+`block_name` is the name of the block within a StreamField that contains the data to be migrated.
 
 StreamBlocks can themselves also contain child blocks. 
-`migrate_page_types_and_fields` will run the given mapper function on 
-all StreamBlocks within the StreamField and 
-all child blocks that are of `block_type`.
+The block name can be given as a list of block names 
+that form the "path" to the block that needs to be migrated. 
+For example :
+
+```python
+def forwards(apps, schema_editor):
+    page_types_and_fields = [
+        ('myapp', 'MyPage', 'field_name', ['parent_block', 'child_block']),
+    ]
+    migrate_page_types_and_fields(
+        apps,
+        page_types_and_fields,
+        forward_mapper
+    )
+```
+
+In this example, 
+a block with the name `child_block` 
+that is inside a block named `parent_block` 
+will be passed to the `forward_mapper` function.
 
 The `data` that gets passed to the `forward_mapper` or `backward_mapper`
 is a JSON-compatible Python `dict` that corresponds to the block's schema.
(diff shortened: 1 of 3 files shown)


docs/migrations.md
```diff
@@ -171,20 +169,24 @@ def backward_mapper(page_or_revision, data):
 
 def forwards(apps, schema_editor):
     page_types_and_fields = [
-        ('myapp', 'MyPage', 'streamfield_name', 'streamblock_type'),
+        ('myapp', 'MyPage', 'field_name', 'block_type'),
```
Scotchester: The added explanation below helps, but since this is in the context of our special Wagtail migration methods, I would leave this as `streamfield_name` to make it very clear what that param is expecting.

docs/migrations.md
```diff
@@ -194,25 +196,57 @@ class Migration(migrations.Migration):
...
+`field_name` is the name of the StreamField on the Page model that contains the blocks to migrate. 
+`block_type` is the type of block within a StreamField that contains the data to be migrated.
+
+StreamBlocks can themselves also contain child blocks. 
+`migrate_page_types_and_fields` will run the given mapper function on 
+all StreamBlocks within the StreamField and 
+all child blocks that are of `block_type`.
+
+The `data` that gets passed to the `forward_mapper` or `backward_mapper`
+is a JSON-compatible Python `dict` that corresponds to the block's schema.
```
Scotchester: Is it already a `dict`? Do we need the `data = dict(data)` lines in the example mappers? I feel like I've not needed those in the past. Will try to confirm with my migration in progress.

docs/migrations.md
```diff
@@ -194,25 +196,57 @@ class Migration(migrations.Migration):
...
+
 ### Utility functions
 
 These functions, defined in `v1.util.migrations`,
-are used in the above data migration example.
+are used in the above data migration example. 
+They reduce the amount of boiler-plate required in data migrations
```
Scotchester:
```suggestion
They reduce the amount of boilerplate required
```

docs/migrations.md
```diff
@@ -194,25 +196,57 @@ class Migration(migrations.Migration):
...
+The `data` that gets passed to the `forward_mapper` or `backward_mapper`
+is a JSON-compatible Python `dict` that corresponds to the block's schema.
+
 ### Utility functions
 
 These functions, defined in `v1.util.migrations`,
-are used in the above data migration example.
+are used in the above data migration example. 
+They reduce the amount of boiler-plate required in data migrations 
+to work with Wagtail StreamField data.
```
Scotchester:
```suggestion
to work with Wagtail StreamField data in data migrations.
```

docs/migrations.md
willbarton: What I'm trying to make clear is that it's the name of the field on the Django model — coming back to this to write a data migration after six months or so it took me a while to realize this.

docs/migrations.md
willbarton: I was wondering that myself when writing this. I don't think we need that line? But I want to make sure.

docs/migrations.md
Scotchester: I suspected that's what you would say :) I'm fine with that.

willbarton: Heh. I'm totally open to making it clearer in any way possible — like I said, what I originally wrote was totally not obvious to me now :)

chosak:
I have concerns about using the terms `type` and `block_type` here, when these things are really more appropriately the **name** of the block, which can vary based on the context.

For most of our atomic elements we tend to usually make the block name similar to the block type when defined at the StreamField top level (e.g. `'text_introduction'` for TextIntroduction), but not always (e.g. `'contact'` for `MainContactInfo`). Once we get into sub-blocks, the names are more often different than the type.

We could potentially encourage/enforce a standard way of naming blocks and sub-blocks, but my instinct is that it'd be overly restrictive (and would need to be documented).

If we want to go with the direction proposed in this PR, could we be more clear that we are referring to the name or key of the block, not the type? And recognize the caveat that there might be places where the names conflict or differ?

Another approach that @Scotchester and I have discussed offline would be to define the signature of inputs to migrate more explicitly, e.g. `page_types_and_fields` would become a tuple of (app, model, field, path to block), where the last element can have between 1 and N levels:

```py
page_types_and_fields = [
    ('v1', 'BlogPage', 'content', ('email_signup',)),
    ('v1', 'BlogPage', 'content', ('full_width_text', 'email_signup',)),
]
```

In this case the migrate function would use those path elements to determine which blocks to migrate.

---

Are these changes backwards compatible with any existing data migrations that exist and might be using the `migrate_page_types_and_fields` function?

willbarton:
> Another approach that @Scotchester and I have discussed offline would be to define the signature of inputs to migrate more explicitly, e.g. `page_types_and_fields` would become a tuple of (app, model, field, path to block), where the last element can have between 1 and N levels:

Ooooh. I do like how explicit that is a *lot* better. It would still require two lines for the Blog Page email signup block though — which is the change that inspired this. @Scotchester any opinions?

> Are these changes backwards compatible with any existing data migrations that exist and might be using the `migrate_page_types_and_fields` function?

Yes, these changes are backwards-compatible to the extent I've been able to determine (the tests for those migrations pass) — and we're not interested in going backwards anymore, so theoretically any data migrations that have run on production before this gets deployed will never have to run again.

chosak:
> heoretically any data migrations that have run on production before this gets deployed will never have to run again

This is true for existing cf.gov environments, and for local environments that bootstrap using a production dump, but those old migrations still get run on a new database (`./initial-data.sh`) and as part of the full tox tests. So the code in those data migrations still needs to be able to at least run without exception.

Scotchester:
> > Another approach that @Scotchester and I have discussed offline would be to define the signature of inputs to migrate more explicitly, e.g. `page_types_and_fields` would become a tuple of (app, model, field, path to block), where the last element can have between 1 and N levels:
> 
> Ooooh. I do like how explicit that is a _lot_ better. It would still require two lines for the Blog Page email signup block though — which is the change that inspired this. @Scotchester any opinions?

I don't mind the added line if it makes the method signature more explicitly clear.

willbarton:
Ok, I will update this PR to use the explicit path instead. I like that there's less magic in that.

docs/migrations.md
willbarton: Nope, looks like the casting as a dict is unnecessary.

willbarton:
@Scotchester @chosak alrighty, I think I've got the path-based block names working. I've also confirmed this doesn't break the existing migrations on a blank/initial-data database.

willbarton:
> I have concerns about using the terms `type` and `block_type` here, when these things are really more appropriately the **name** of the block, which can vary based on the context.

I've also tried to replace all references to "block type" with "block name".

willbarton:
@Scotchester and I found a bug where subsequent attempts at recursion would make earlier `migrated = True` values get overridden, resulting in migrated fields not being ultimately written back to the page. I've fixed that in the latest commit.

Issue #4822, opened 2019-02-13
Update docker installation instructions


wpears:
The current docker installation instructions presume that a Mac user is working with docker-machine rather than Docker for Mac. This is largely not the case for current users of cfgov-refresh and the docs should be updated accordingly.

Probably the best thing to do in the interim would be to have a installation w/ docker-machine and installation w/ Docker for Mac to meet both use cases.

PR #4825, opened 2019-02-14
Unset docker-machine variables if using Docker for Mac

contolini:
If your system is not using docker-machine, ensure those environment variables aren't set or mysterious issues will arise. Related to #4822.

See: 

https://forums.docker.com/t/initial-install-of-docker-for-mac-could-not-read-ca-certificate/9170
https://docs.docker.com/machine/reference/env/
https://[mm]/cfpb/pl/uy1r1fc35389jx3kufk1rqrrfh

## Additions

- Unset lingering `docker-machine` env variables.

## Testing

1. `source mac-virtualbox-init.sh` should run a-okay


contolini:
> I think the ultimately sensible thing might be to just say we only support native docker and remove all the docker machine and boot2docker stuff.

I agree. We should eventually remove all the old docker-machine references (and delete this file entirely) but that feels like a much larger task than the simple stop gap in this PR.

willbarton:
> I think the ultimately sensible thing might be to just say we only support native docker and remove all the docker machine and boot2docker stuff.

Sublime agreement.

PR #4831, opened 2019-02-19
Move search-tags field on AnswerPage

higs4281:
The search-tags field is moving from the Answer model to the
AnswerPage model, and it should be displayed with other tag and
metadata fields for the page, instead of on the content tab.


higs4281:
This should wait for other metadata fields.

PR #4839, opened 2019-02-21
Update Ask CFPB export fields 

richaagarwal:
## Overview

See GHE issue GHE/CFGOV/Ask-Evo/issues/40

Now that https://github.com/cfpb/cfgov-refresh/pull/4818 has been merged (!!), we need to update the Ask export to pull certain fields from the `AnswerPage` instead of the `Answer` (specifically, `answer`, `question`, and `snippet`).  We will further update this with upcoming work to migrate all fields to the `AnswerPage`, but for now we need to do this so that the export continues to be accurate and up-to-date.

## Changes

I've updated the `test_export_script_assemble_output` test so that it lives with the rest of the `ExportAskDataTests` tests, both for improved readability and to minimize conflicts with the setUp data in the `AnswerModelTestCase`. 

The only functional change this introduces is that the export will now output live versions of what's on a page, whereas previously since it was querying directly from the Answer, it exported the latest revision. We have a pending to-do to clarify with CE whether this will be an issue or if we need to make this return draft content (which will likely slow the script down), but in the meantime, I wanted to get this in so that we have something working at least with the live content.

Lastly, this PR introduces a one-time script* we need to run in order to fix certain unsupported characters that are causing issues in the export (they may have made their way in from Word copy + paste.)  We plan to address these unicode issues more widely with a future migration, but for now I'm tackling the characters that specifically cause the export to choke.  As noted in the GHE issue, this impacts only 14 pages which are safe to publish.

*I will open a PR to remove this script once it runs on production. I am open to suggestions on how else to run this but would like to avoid copying and pasting code into terminal :) 

## Local testing

If you'd like to test this out locally, you can pull down this branch and import a relatively recent database dump.

From http://localhost:8000/admin/export-ask, you can attempt to download the csv, but it will fail - you should see this error in the browser:
screen shot 2019-02-25 at 5 48 52 pm

You'll need to run the script that takes care of the problematic characters, by running:
```
python cfgov/manage.py runscript replace_unsupported_characters
```
This should impact 14 pages.

Now, if you try to download the csv again, it should work. Hooray!

For extra testing, you can confirm that (published) changes you've made to an AnswerPage get reflected in the csv.

### DIFF
--- a/cfgov/ask_cfpb/scripts/replace_unsupported_characters.py
+++ b/cfgov/ask_cfpb/scripts/replace_unsupported_characters.py
@@ -1,0 +1,33 @@
+from __future__ import unicode_literals
+
+from ask_cfpb.models.pages import AnswerPage
+
+
+unsupported = ['\x92', '\x93', '\x94', '\x96']
+
+
+def fix(text):
+    text = text.replace('\x92', "'")
+    text = text.replace('\x93', '"')
+    text = text.replace('\x94', '"')
+    text = text.replace('\x96', '-')
+    return text
+
+
+def run():
+    for page in AnswerPage.objects.all():
+        fixed = False
+
+        for string in unsupported:
+            if string in page.question:
+                page.question = fix(page.question)
+                fixed = True
+            if string in page.answer:
+                page.answer = fix(page.answer)
+                fixed = True
+            if string in page.snippet:
+                page.snippet = fix(page.snippet)
+                fixed = True
+
+        if fixed:
+            page.save_revision().publish()
--- a/cfgov/ask_cfpb/scripts/export_ask_data.py
+++ b/cfgov/ask_cfpb/scripts/export_ask_data.py
@@ -48,11 +48,13 @@ def assemble_output():
         'related_questions',
         'next_step__title')
     answers = list(Answer.objects.prefetch_related(*prefetch_fields).values(
-        'id', 'answer', 'answer_es', 'question', 'question_es',
-        'snippet', *prefetch_fields))
+        'id', *prefetch_fields))
+
     answer_pages = list(AnswerPage.objects.values(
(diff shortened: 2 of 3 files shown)


cfgov/ask_cfpb/scripts/replace_unsupported_characters.py
```diff
@@ -0,0 +1,31 @@
...
+                fixed = True
+            if string in page.answer:
+                page.answer = fix(page.answer)
+                fixed = True
+            if string in page.snippet:
+                page.snippet = fix(page.snippet)
+                fixed = True
+
+        if fixed:
+            page.save_revision().publish()
```
higs4281: Can we be sure that any page we find with bad characters can be published? It might be the case currently, but perhaps not in the future.

richaagarwal: Right -- so unless they're planning on making any changes in the next week or so, I think this is safe as a one-time thing.  That's why I'm not sure it should be merged into the codebase or not; I was hoping to run it just once so that the updates to the Ask export can get merged.

cfgov/ask_cfpb/scripts/replace_unsupported_characters.py
```diff
@@ -0,0 +1,31 @@
+from ask_cfpb.models.pages import AnswerPage
+
+
+unsupported = [u'\x92', u'\x93', u'\x94', u'\x96']
```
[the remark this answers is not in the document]
richaagarwal: Ooh good idea!  I think that will be useful for the fuller conversion we'll want to do when we're ready to publish all live + draft pages, but for now am trying to contain it to the characters that cause the export to fail.  After we run this in production I'm happy to open a PR to delete this script entirely or update it to reflect a complete conversion.

PR #4840, opened 2019-02-21
Deprecate non-Wagtail mega menu

chosak:
This change removes the `WAGTAIL_MENU` feature flag and removes the template for the non-Wagtail mega menu.

The Wagtail mega menu has been turned on in production since August 2018 (over six months), but the cfgov-refresh repository still defaults to using the non-Wagtail version unless the `WAGTAIL_MENU` flag is set. It happens to be set in production database dumps, but not when starting from an empty database.

This change removes the `WAGTAIL_MENU` flag and the non-Wagtail menu template.

## Notes

If you run the site without any menu items defined (say, from an empty database after `./initial-data.sh`) there's a little extra whitespace where the menu would go:

image

I looked for a minute at removing this but it's not as simple as just removing the HTML elements that make up the menu because of (a) the way we initialize the header JS and (b) the way that the complaint link is part of the mega menu on smaller screen sizes.

## Checklist


anselmbradford:
Any modifications needed to https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/mega-menu.html#L227? How's that called now?

chosak:
@virginiacc thank you for the review. That's an interesting thought, although the [current documentation](https://cfpb.github.io/cfgov-refresh/mega-menu/) is really thorough (thanks to you, I think!) and is already written as if the Wagtail menu is the only way that the menu can be edited. In practice I don't think developers tend to ever encounter the case where there is no menu if they work with database dumps.

@anselmbradford I don't think so? That's called from [here](https://github.com/cfpb/cfgov-refresh/pull/4840/files#diff-07fa1ae187ed14f88d55bc0feb237ff3R73) given the list of Wagtail-specified menu items.

PR #4842, opened 2019-02-22
Add a note that backward migrations are not required

willbarton:
Per a decision made by the backend team in our last huddle, backward data migrations are not required in cfgov-refresh. This change adds a note to that effect in the migration documentation.

## Screenshots

image

## Checklist

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -124,6 +124,10 @@ class Migration(migrations.Migration):
 The `forwards()` and `backwards()` functions are where any changes
 that need to happen to a model's data are made.
 
+!!! note
+    While backwards migrations are necessary in external libraries that we create, 
+    we do not require them in cfgov-refresh. 
+
 ### Wagtail-specific considerations
 
 Django data migrations with Wagtail can be challenging because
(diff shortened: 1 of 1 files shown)


docs/migrations.md
```diff
@@ -124,6 +124,10 @@ class Migration(migrations.Migration):
 The `forwards()` and `backwards()` functions are where any changes
 that need to happen to a model's data are made.
 
+!!! note
+    While backwards migrations are necessary in external libraries that we create,
```
chosak: Is "encouraged" a better word to use here than "necessary"?

docs/migrations.md
```diff
@@ -124,6 +124,10 @@ class Migration(migrations.Migration):
 The `forwards()` and `backwards()` functions are where any changes
 that need to happen to a model's data are made.
 
+!!! note
+    While backwards migrations are necessary in external libraries that we create, 
+    we do not require them in cfgov-refresh.
```
chosak: Would it be worth adding more in the way of explanation, something like

...we do not require them in cfgov-refresh because we prefer not to rollback migrations that have already been applied.

docs/migrations.md
willbarton: Perhaps? But I feel like any exceptions should be adequately explained (I would hate to encounter a library that I couldn't migrate backwards, personally). I like the stronger language, personally.

PR #4845, opened 2019-02-22
Add ability to log DB queries to console

chosak:
This commit adds the ability to (optionally) log Django database queries to the console by defining a new `ENABLE_SQL_LOGGING` environment variable.

This is useful when trying to debug database interactions, and works against the Django server, interactive shell, and management commands.

The documentation on developer tips has been updated to include this ability.

## Screenshots

New documentation section:

image

## Notes

This is similar to some functionality that [django-extensions](https://github.com/django-extensions/django-extensions) already provides through its `shell_plus` and `runserver_plus` commands, but doing it via our local settings lets us more easily turn this on globally for things like management commands.

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/local.py
+++ b/cfgov/cfgov/settings/local.py
@@ -23,7 +23,7 @@ LOGGING = {
     'disable_existing_loggers': False,
     'handlers': {
         'console': {
-            'level': 'INFO',
+            'level': 'DEBUG',
             'class': 'logging.StreamHandler',
         }
     },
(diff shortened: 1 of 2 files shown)


cfgov/cfgov/settings/local.py
```diff
@@ -23,7 +23,7 @@
     'disable_existing_loggers': False,
     'handlers': {
         'console': {
-            'level': 'INFO',
+            'level': 'DEBUG',
```
chosak: This change shouldn't affect how things are logged anywhere else because [we define the fallback logger](https://github.com/cfpb/cfgov-refresh/blob/541656522efca8570d3ab7355a5da246dc521a38/cfgov/cfgov/settings/local.py#L33) with a level of `INFO`. That should filter out any `logger.debug()` messages unless the console handler is explicitly invoked.

PR #4851, opened 2019-02-26
Move rest of Answer fields onto AnswerPage

richaagarwal:
Move remaining Answer fields onto Answer page!  

This retires the `Answer` model from the admin, in addition to the rest of the Ask CFPB specific models other than `Category` and `Subcategory` which will be dealt with in upcoming work.

No data is being deleted yet - just copied over. 

There's a migration script (`answer_to_answerpage_migration`) that needs to run immediately after deployment - until it runs, there will be a brief period of downtime for Answer pages (a few minutes or so) - I think we can manage this by deploying after-hours.

After the migration script, there's a sanitization script (`replace_unsupported_characters`) that we should run.


cfgov/ask_cfpb/models/pages.py
```diff
@@ -586,22 +554,81 @@ class AnswerPage(CFGOVPage):
...
+        on_delete=models.SET_NULL,
+        related_name='redirect_to_pages',
+        help_text="Choose another AnswerPage to redirect this page to")
+    featured = models.BooleanField(
+        default=False,
+        help_text=(
+            "Check to make this one of two featured answers "
+            "on the landing page."))
+    featured_rank = models.IntegerField(blank=True, null=True)
+    category = models.ManyToManyField(
```
richaagarwal: I'll remove this -- this is just from before when we thought we'd migrate everything over first.  Once we get the final mapping from CE we can open up another PR to handle categories and subcategories.

cfgov/ask_cfpb/models/pages.py
```diff
@@ -586,22 +548,72 @@ class AnswerPage(CFGOVPage):
...
-            FieldRowPanel([FieldPanel('last_edited')])
-        ],
-            heading="Visible time stamp"),
-        FieldPanel('question'),
-        FieldPanel('statement'),
-        FieldPanel('snippet'),
-        FieldPanel('answer'),
-        FieldPanel('search_tags'),
-        FieldPanel('redirect_to'),
+            FieldPanel('question', classname="title"),
```
higs4281: I think applying `classname="title"` to questions and statements make them difficult to edit because the text is too big to fit in the form window.

cfgov/ask_cfpb/scripts/answer_to_answerpage_migration.py
```diff
@@ -0,0 +1,50 @@
...
+            if page.redirect_to and page.redirect_to.english_page:
+                page.redirect_to_page = page.redirect_to.english_page
+            page.search_tags = page.answer_base.search_tags
+            page.last_edited = page.answer_base.last_edited
+            for related_question in page.answer_base.related_questions.all():
+                page.related_questions.add(related_question.english_page)
+
+        if page.status_string != 'draft':
+            # We need to publish for the answer ID, next step, and search tags
+            page.save_revision().publish()
```
higs4281: I think we should pass a user to `page.save_revision()` to avoid anonymous revisions.

cfgov/ask_cfpb/models/pages.py
```diff
@@ -586,22 +548,72 @@ class AnswerPage(CFGOVPage):
...
+            "Choose only subcategories that belong "
+            "to one of the categories checked above."))
+    search_tags = models.CharField(
+        max_length=1000,
+        blank=True,
+        help_text="Search words or phrases, separated by commas")
+    related_resource = models.ForeignKey(
+        RelatedResource,
+        blank=True,
+        null=True
```
higs4281: I think we'll need to add `on_delete=models.SET_NULL` for related_resource.

higs4281:
I strongly favor creating the two missing Spanish pages before merging this PR.  
There is no harm in creating two draft pages, and if they're truly unwanted, it's trivial to move them to trash. But it won't be trivial to dig into history and create them by hand if they are indeed wanted.

cfgov/ask_cfpb/models/pages.py
richaagarwal: @higs4281 @virginiacc Is this something we should take care of / remove now, or will your enhancements take care of this kind of thing?

richaagarwal:
I think remaining work on this is

- Rebase 
- Squash the migrations we're introducing?
- Potentially update `view_answer` in `ask_cfpb/views.py` to grab the answer_base ID instead so that pages aren't broken while we deploy (though my preference would be to leave it as-is and deploy off-hours).

cfgov/ask_cfpb/models/pages.py
higs4281: I left those classname specifiers off in the initial migration, because it can make long questions hard to read. I don't think we should re-introduce them.

higs4281:
> squash migrations
I vote for this, since we have to redo the v1 migration anyway since there have been two migrations merged before ours.

> Revert `view_answer` to use answer_base.id
I vote for this as well. We shouldn't have two sources of truth for a value that isn't visible in the admin.

cfgov/jinja2/v1/ask-cfpb/answer-page.html
```diff
@@ -112,19 +112,19 @@ <h3 class="answer-snippet">
...
-            {{ page.answer_base.next_step.text | safe }}
+            {{ page.related_resource.text | safe }}
         </div>
     {% endif %}
     {% if page.sidebar_panels %}
             {{ streamfield_sidefoot.render(page.sidebar) }}
     {% endif %}
     {% if category and subcategories %}
-        <div class="block {% if not page.answer_base.next_step %}block__flush-top{% endif %}">
+        <div class="block {% if not page.next_step %}block__flush-top{% endif %}">
```
virginiacc: I think this `next_step` reference also needs to be updated to `page.related_resource`.

higs4281: great catch.

richaagarwal: Should we test this type of thing?  Like - if there exists a `related_resource` we expect to see it on the rendered page?  Not sure if that is overkill but we do something similar with our molecules tests.

PR #4860, opened 2019-03-01
Wagtail: Add notification molecule to sublanding pages

anselmbradford:
## Changes

- Update `notification.html` to support `value` passed in from wagtail.

## Testing

1. Create a new sublanding page in wagtail and add a notification to the content area. ([Shortcut](http://localhost:8000/admin/pages/add/v1/sublandingpage/319/))
2. Set the "message" only and preview the page to see that the notification type is the default and the message shows up.
3. Change the type drop-down and preview the page to see the color and icon change in the notification.
4. Add the explanation text and preview the page to see that it is added below the notification message.
5. Add links and preview the page to see that links appear after the explanation text.
5. See that hardcoded notifications on e.g. http://localhost:8000/about-us/blog/ still work (e.g. enter date before 1900).

## Screenshots

NEW SCREENSHOTS (4/5)

screen shot 2019-03-05 at 11 46 21 am

screen shot 2019-03-05 at 10 33 07 am

screen shot 2019-03-05 at 10 33 17 am


cfgov/jinja2/v1/_includes/molecules/notification.html
```diff
@@ -47,3 +47,27 @@
...
 
 {% endmacro %}
+
+{% if value %}
+
+    {#
+      TODO: This if/else is for removing the explanation if it's empty HTML.
+            Investigate a better way to do this that doesn't repeat the call.
+    #}
+    {% if ( value.explanation | trim ) == '<div class="rich-text"><p><br/></p></div>' %}
```
anselmbradford: @chosak Quite certain there's a better way than this but I don't know what it is. Do you? I'm checking that the explanation is an empty rich text block here and excluding it from the render call if needed. I tried `{% set value.explanation = none %}` and got an error (`cannot assign attribute on non-namespace object`).


cannot assign attribute on non-namespace object
--


.

chosak: @anselmbradford this is a known wontfix issue ([1](https://github.com/wagtail/wagtail/issues/3111), [2](https://github.com/wagtail/wagtail/issues/4046)) for Wagtail, addressed in the new Wagtail 2.x rich text editor, which unfortunately we can't move to right now.

We could maybe do something like [this](https://github.com/wagtail/wagtail/issues/2336#issuecomment-396536853). Is there anywhere else on the site where we've encountered this, or need to do boolean tests on rich text fields?

anselmbradford: That note at the bottom…

> The html module is available only in Python 3.

Does that mean we can't use this approach yet?

anselmbradford: If the resolution here is to leave a `TODO` I'm fine with that, if you are.

chosak: @anselmbradford we don't have to do it exactly as proposed in that link. The problem with the existing version is that `<div class="rich-text"><p><br/></p></div>` isn't guaranteed -- it might have zero `<br/>`s, one, multiple.

I'll take another look a little later today to see if I can find a simple workaround.

One question, though: does this have to be rich text, or could it just be text? What is the use case for rich text -- having headings? Can you be more specific about exactly which rich text elements are to be allowed in this notification?

anselmbradford: > One question, though: does this have to be rich text, or could it just be text? What is the use case for rich text -- having headings? Can you be more specific about exactly which rich text elements are to be allowed in this notification?

Good question. We have instances that have links. That's the only requirement that I thought necessitated rich text.

chosak:
@anselmbradford can you take [this commit from my branch](https://github.com/cfpb/cfgov-refresh/compare/ans_wagtail_notification...chosak:ans_wagtail_notification) for a spin? This simplifies the logic by testing the rich text content using the Jinja2 [`striptags`](http://jinja.pocoo.org/docs/2.10/templates/#striptags) filter: essentially, if stripping HTML tags from the rich text content leaves an empty string, treat the field as empty.

Per [your comment above](https://github.com/cfpb/cfgov-refresh/pull/4860#discussion_r262181096), I also [limited the rich text field](https://github.com/cfpb/cfgov-refresh/compare/ans_wagtail_notification...chosak:ans_wagtail_notification#diff-6ca91ea473f91ef1fd574632e0f19595R194) to support links only.

anselmbradford:
@chosak Thanks so much for looking into this and for your solution. It looks like we're going to simplify the explanation text and not make it rich text after all and place the links below it.

PR #4861, opened 2019-03-04
Update tests and remove unneeded code.

higs4281:
In addition to restoring tests, this drops some more code blocks, including:

- answerproxy models for haystack
- search indexing for categories
- all facet_map code and tests
- answer deletion tests


cfgov/ask_cfpb/models/django.py
```diff
@@ -127,8 +123,7 @@ def __str__(self):
         return self.name
 
     def featured_answers(self, language):
```
richaagarwal: Should this just be moved to the `AnswerPage` at this point?

cfgov/ask_cfpb/models/django.py
```diff
@@ -127,8 +123,7 @@ def __str__(self):
...
-        from .pages import AnswerPage
```
richaagarwal: I know this functionality will go away eventually with removal of the landing page, but I'm wondering if in the interim we should move this over to the `AnswerPage`?

higs4281: I don't see an advantage here. We can redo tests and migrations, etc., but we'll end up taking it back out in the end. I don't see the value for a short-term condition.

PR #4862, opened 2019-03-04
Fix missing YouTube video preview images

Scotchester:
This PR addresses https://GHE/CFGOV/platform/issues/3287, which reports that YouTube videos are no longer showing their correct preview images, but instead are all showing the fallback CFPB logo.

I traced the cause back to [this change](https://github.com/cfpb/cfgov-refresh/commit/50170057f0fc808d2b6aebc9984b9650d378af31#diff-75e5fea40be00a78d02194824cce179aR44). As noted by [our `assign` util's JSDoc description](https://github.com/cfpb/cfgov-refresh/blob/fix-youtube-thumbnails/cfgov/unprocessed/js/modules/util/assign.js#L17-L19), "fields of every next source will override same named fields of previous sources", so by removing the placeholder `dataSet` variable that was in place prior to that change, the `id` of `this.iFrameProperties` was replacing the `id` in `this.baseElement.dataset`, and thus in the `data-id` attribute of the `baseElement` markup, which is in turn used to generate the URL for the preview image.

## Changes

- Fix `VideoPlayer` JS so that it no longer accidentally overwrites data attributes on `baseElement`

## Testing

1. Load up any page with a video (like http://localhost:8000/prepaid-rule/) and see that it shows the CFPB logo, not a proper image preview
1. Pull branch
1. `gulp scripts`
1. Reload the page and see the thumbnail now appears

## Checklist

### Browsers

- [x] Chrome on desktop
- [x] Firefox
- [x] Safari on macOS
- [x] Edge
- [x] Internet Explorer ~~9, 10, and~~ 11
- [x] Safari on iOS
- [x] Chrome on Android

### DIFF
--- a/cfgov/unprocessed/js/modules/VideoPlayer.js
+++ b/cfgov/unprocessed/js/modules/VideoPlayer.js
@@ -40,8 +40,9 @@ function VideoPlayer( element, options ) {
   _this = this;
   options = options || {};
   this.baseElement = _ensureElement( element, options.createIFrame );
+  const baseDataset = assign( {}, this.baseElement.dataset );
   this.iFrameProperties = assign(
-    this.baseElement.dataset,
+    baseDataset,
     this.iFrameProperties
   );
 
--- a/cfgov/unprocessed/js/modules/VideoPlayer.js
+++ b/cfgov/unprocessed/js/modules/VideoPlayer.js
@@ -40,9 +40,9 @@ function VideoPlayer( element, options ) {
   _this = this;
   options = options || {};
   this.baseElement = _ensureElement( element, options.createIFrame );
-  const baseDataset = assign( {}, this.baseElement.dataset );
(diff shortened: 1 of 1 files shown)


Scotchester:
Seems like CodeCov is ignoreable, as VideoPlayer.js has no unit tests as of yet.

anselmbradford:
I still don't get it. `const baseDataset = assign( {}, this.baseElement.dataset );` is assigning to an empty object, how does using `this.baseElement.dataset` directly make a difference?

Scotchester:
I did a little pairing with Ans and we figured out a slightly more efficient way to do it, which I have just pushed.

Longer-term, it might be nice to get rid of the `assign` util and just assign things manually to reduce complexity and the chance of bugs like this. Or just switch to the stdlib `Object.assign` when we drop JS support for IE 11.

PR #4863, opened 2019-03-05
AnswerPage tweaks

richaagarwal:
- Add feedback form to AnswerPage with after_create_page hook
- Remove unneeded Audience and NextStep code
- Modify related questions to a ParentalManyToManyField
- Remove redundant social_sharing_image from metadata section on page


cfgov/ask_cfpb/wagtail_hooks.py
```diff
@@ -111,3 +100,37 @@ def register_export_menu_item():
 @hooks.register('register_admin_urls')
 def register_export_url():
     return [url('export-ask', export_data, name='export-ask')]
+
+
+def get_feedback_stream_value(page):
```
higs4281: This is a better place for the code, but we need a test.

cfgov/ask_cfpb/wagtail_hooks.py
```diff
@@ -111,3 +100,37 @@ def register_export_menu_item():
...
+
+
+@hooks.register('after_create_page')
+def after_create_page(request, page):
+    if isinstance(page, AnswerPage):
+        page.content = StreamValue(
+            page.content.stream_block,
+            get_feedback_stream_value(page),
+            is_lazy=True)
+        page.save_revision(user=request.user)
```
virginiacc: This is very minor since it's unlikely to happen often, but I was wondering if we should publish the revision if the user is publishing the page? Otherwise the page's status would be live + draft rather than live, which might be confusing for the user.

richaagarwal:
Closing this since @higs4281 came up with something cleaner :)

PR #3122, opened 2017-07-12
Fix the layout and styles for heroes

jimmynotjim:
Capital Framework made a dramatic change in the way we build heros to
provide future options for responsive images. These changes broke the
existing hero in this repo.

## Removals

- Removed the old hero less file

## Changes

- Updated the hero template to match markup in CFv4 examples

## Testing

1. run `./frontend.sh`

## Screenshots

screen shot 2017-07-12 at 10 31 51 am

screen shot 2017-07-12 at 10 32 37 am

screen shot 2017-07-12 at 10 32 08 am

screen shot 2017-07-12 at 10 32 16 am

screen shot 2017-07-12 at 10 45 15 am

screen shot 2017-07-12 at 10 45 08 am

## Notes

- In order to test each hero type, you'll have to grab the static urls and replace them in the template or your browsers inspection window. Maybe we should find a way to pull images from the server instead of locally in the future.
- See https://github.com/cfpb/capital-framework/pull/469 for history of hero change.

## Todos

- Update the image and padding settings to be programmatically set and delivered via Django

## Checklist

* [x] PR has an informative and human-readable title
* [x] Changes are limited to a single goal (no scope creep)
* [x] Code can be automatically merged (no conflicts)
* [x] Code follows the standards laid out in the [development playbook](https://github.com/cfpb/development)
* [x] Passes all existing automated tests
* [x] New functions include new tests
* [x] New functions are documented (with a description, list of inputs, and expected output)
* [x] Placeholder code is flagged
* [x] Visually tested in supported browsers and devices
* [x] Project documentation has been updated
* [x] Reviewers requested with the [Assignee tool](https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/) :arrow_right:

PR #4864, opened 2019-03-05
Remove links from hero in Wagtail

anselmbradford:
Wagtail included an option to include links, but it appeared not to actually work as the template includes no references to links.

## Removals

- Remove links option from heros in Wagtail.

## Testing

1. 

## Screenshots

Before:

screen shot 2019-03-05 at 12 29 47 pm

After:

screen shot 2019-03-05 at 12 30 08 pm


Scotchester:
The decision to stop allowing links in heroes was decided in https://github.com/cfpb/design-manual/issues/454 and implemented template-wise in https://github.com/cfpb/cfgov-refresh/pull/3122. Either due to forgetting or wanting to keep that PR's scope constrained, the field was not removed from the molecule at that time. Thanks for submitting this PR, Ans.

anselmbradford:
@Scotchester It is done! I also re-ordered the fields so the image-related ones were contiguous

screen shot 2019-03-05 at 1 46 11 pm

cfgov/regulations3k/migrations/0021_rm_hero_links.py
```diff
@@ -0,0 +1,23 @@
...
+from django.db import migrations
+import wagtail.wagtailcore.blocks
+import wagtail.wagtailcore.fields
+import wagtail.wagtailimages.blocks
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('regulations3k', '0021_rm_hero_links'),
```
Scotchester: Something happened here with your update. This needs to be pointing to regulations3k migration 0020.

anselmbradford: Fixed

PR #4871, opened 2019-03-07
Use standard feedback module

higs4281:
With a tweak to the v1 feedback template, we can use the
standard feedback module without any customization.

Current pages will pick up the content changes in the
v1 feedback forms.

For new pages, editors will need to add a feedback module, but they
won't have to change any values.

Code coverage is back to 100%!


higs4281:
It agree it would be nice to have feedback automatically initialized, even though it's a single-click task.
But for this step I wanted to stick to standard practice and keep changes to current pages to a minimum.

PR #4878, opened 2019-03-08
Update migration script so that it's working off of latest revision

richaagarwal:
Currently, the script is skipping any draft revisions that exist, since by default querying for a page gives back the live version.  This fixes that. 

I've included an updated sanitization script as well.

### DIFF
--- a/cfgov/ask_cfpb/scripts/answer_to_answerpage_migration.py
+++ b/cfgov/ask_cfpb/scripts/answer_to_answerpage_migration.py
@@ -72,10 +72,9 @@ def run():
             for related_question in page.answer_base.related_questions.all():
                 page.related_questions.add(related_question.english_page)
 
-        # Save & unpublish the draft page
+        # Save & publish our changes
+        page.save_revision(user=user).publish()
+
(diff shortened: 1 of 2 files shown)


cfgov/ask_cfpb/scripts/answer_to_answerpage_migration.py
```diff
@@ -60,8 +72,10 @@ def run():
             for related_question in page.answer_base.related_questions.all():
                 page.related_questions.add(related_question.english_page)
 
-        if page.status_string != 'draft':
-            # We need to publish for the answer ID, next step, and search tags
-            page.save_revision(user=user).publish()
-        else:
+        # Save & unpublish the draft page
```
higs4281: Might be simpler to publish all pages to make sure revisions are in order, then unpublish drafts.

```python
        page.save_revision(user=user).publish()
        if draft:
            page.unpublish()
```

richaagarwal: Good catch!

PR #4879, opened 2019-03-08
Notification: convert explanation to TextBlock and remove safe filter

anselmbradford:
## Changes

- Convert notification explanation to a TextBlock to make input easier.
- Remove `| safe` filter from explanation since it is no longer a rich text input.

## Testing

1. Create a new sublanding page and add a notification. If you add more than one line to explanation the text area should grow to accommodate the text.
2. Add a new browse page and add a conference registration form.
3. Fill in the required fields. Add a link to the success message Rich Text Field. 
4. Publish the form, fill in the details  and hit the "register" button and see that the success notification has a notification with HTML:
 
Screen Shot 2019-03-12 at 5 31 17 PM


anselmbradford:
> This works as advertised, but we may want to consider leaving the safe filter on there as an undocumented way to use HTML to add some inline formatting, in a pinch. I get that part of the idea was not using RTF, but I like allowing that as a middle ground. Not a dealbreaker for me by any means, though.

Oh interesting, I didn't realize you could put HTML into a TextBlock… now it looks like we didn't need the RTF in the beginning 🤦‍♂️ 
@chosak do you have a preference? I know you asked about this.

> Maybe also add this to the other page types we want to have it while we're here and already creating a migration file?

I was going to do that in another issue, but sure I can add it here to cut down on migration files!

chosak:
> @chosak do you have a preference?

I would prefer not to encourage putting HTML in TextBlocks. Using `| safe` is somewhat dangerous because it trusts whatever the editor puts in the field, including, say `<script>` tags.

That being said, we do unfortunately have some existing cases where we manually render the notification with embedded HTML tags. I believe the conference registration form template does this. Its success/error/at-capacity messages are all rich text. [Here's](https://github.com/cfpb/cfgov-refresh/blob/461b0c95ea2e586d5c6e3d8fdc2bc36889fe879a/cfgov/data_research/jinja2/data_research/conference-registration-form.html#L14) where the success message is rendered, and here's what was used last time (redacted):

image

anselmbradford:
> That being said, we do unfortunately have some existing cases where we manually render the notification with embedded HTML tags. I believe the conference registration form template does this. Its success/error/at-capacity messages are all rich text. Here's where the success message is rendered, and here's what was used last time (redacted):

Yeah thinking about it I lean toward no `| safe` as well and re-adding it if for some reason there was a use case for it. As much as we can lock down these patterns the better we can manage what to check changes against. The hardcoded ones can do whatever they want (though from a design standpoint should likely follow the Wagtail one). It looks like the conference one has a link on the last line and could be recreated with the Wagtail one easily without markup in `explanation`.

Scotchester:
As noted in my original comment, I'm not strictly hung up on allowing the `safe` filter, but Andy raises a better point: the notification macro needs to better handle the situation if the passed in `explanation` is from a rich text field – i.e., it shouldn't wrap it in a paragraph.

Scotchester:
> As noted in my original comment, I'm not strictly hung up on allowing the `safe` filter, but Andy raises a better point: the notification macro needs to better handle the situation if the passed in `explanation` is from a rich text field – i.e., it shouldn't wrap it in a paragraph.

Realized that the conference registration form is passing into `message`, not `explanation`. This is good to go 👍

PR #4880, opened 2019-03-08
RSSFeed molecule improvements

chosak:
This PR refactors the existing RSSFeed molecule so that Wagtail editors no longer choose from a hardcoded choice of feeds but instead let the module autoselect the feed link (and whether one should be rendered) based on the page on which it appears.

These changes also include some frontend improvements (simplifying multiple templates into one) by @mistergone.

Fixes platform#2584.

## Testing

With a production dump, visit these three pages and verify that the RSS feeds link as intended:

1. [Blog](http://localhost:8000/about-us/blog/) -> [Blog feed](http://localhost:8000/about-us/blog/feed/)
2. [Newsroom](http://localhost:8000/about-us/newsroom/) -> [Newsroom feed](http://localhost:8000/about-us/newsroom/feed/)
3. [Press resources](http://localhost:8000/about-us/newsroom/press-resources/) -> [Newsroom feed](http://localhost:8000/about-us/newsroom/feed/)

In addition, individual Blog pages (like [this one](http://localhost:8000/about-us/blog/tax-season-protect-yourself-irs-imposter-scams/)) should link to the [Blog feed](http://localhost:8000/about-us/blog/feed/), and individual Newsroom pages (like [this one](http://localhost:8000/about-us/newsroom/ffiec-members-adopt-policy-statement-report-examination/)) should link to the [Newsroom feed](http://localhost:8000/about-us/newsroom/feed/).

## Screenshots

The RSSFeed StreamField block will now appear in the Wagtail editor like this:

image

image

## Checklist


chosak:
@mistergone thanks. I didn't think that would be possible, but I checked again, and wrapping [this definition of `admin_text`](https://github.com/cfpb/cfgov-refresh/blob/27a9db34a8d0f7a8f0e05a6ab02a81fdb1a1ba7f/cfgov/v1/atomic_elements/molecules.py#L335) in `django.utils.safestring.mark_safe` allows us to pass HTML elements through to the admin.

I tried wrapping the "RSS Feed:" in a `<strong></strong>` and it looks like this:

image

Wrapping it in an `<h3></h3>` looks like this:

image

Can you suggest some HTML to use here to make it look nice and more visible?

Scotchester:
The `h3` feels appropriate to me (drop the colon, though).

cfgov/jinja2/v1/_includes/molecules/rss-feed.html
```diff
@@ -9,8 +9,20 @@
    Creates a block containing RSS feed markup.
 
    ========================================================================== #}
-
+{% macro render( url ) %}
 <div class="block block__sub" data-qa-hook="rss-subscribe-section">
```
anselmbradford: `block block__sub` should very likely not appear in the molecule, but instead should appear where this molecule is included.

mistergone: Can you elaborate, Ans? When I collapsed `rss.html` and `rss-feed.html` into one file, I included this `div`, but is that problematic from an atomic standpoint or...? Andy and I aren't sure. 😄

anselmbradford: Yeah I'd see it as problematic from an atomic standpoint. 
See this diagram:

Screen Shot 2019-03-08 at 5 00 27 PM

Component A works across all kinds of contexts because the margin is set by its parent container. Component B runs into problems because each new context requires customizations of the component.

chosak: @anselmbradford do you know what `data-qa-hook="rss-subscribe-section"` does? I see [some tests](https://github.com/cfpb/cfgov-refresh/search?q=rss-subscribe-section&unscoped_q=rss-subscribe-section) that refer to it, but does that do anything? Right now that's duplicate both outside and inside of the element. It would simplify to be able to remove it.

anselmbradford: The idea was to have an attribute dedicated to the element that the tests were looking for so that the CSS class could change and the tests would still pass. So depending on how the tests are written I'm pretty sure that should go on the component itself, not its container.

chosak:
@anselmbradford do the changes in 6a767ee satisfy proper atomic element usage?

cfgov/v1/feeds.py
```diff
@@ -63,3 +63,29 @@ def serve(self, request, format='html'):
             return FilterableFeed(self, context)(request)
         else:
             return super(FilterableFeedPageMixin, self).serve(request)
+
+
+def get_page_ancestor_with_feed(page):
```
Scotchester: I think the naming of this function and its local variables is somewhat confusing, given that the passed `page` itself could be found and returned. Consider renaming or refactoring to eliminate the weird inclusive `get_ancestors` behavior.

cfgov/v1/feeds.py
```diff
@@ -63,3 +63,29 @@ def serve(self, request, format='html'):
...
+    """Return the closest page ancestor with a feed.
+
+    Includes the specified page in the search. Returns None if no ancestors
+    inherit from FilterableFeedPageMixin.
+    """
```
Scotchester: Trivial stylistic point: Don't we prefer a blank line after a docstring? (Also below in `get_rss_feed_for_page`.

chosak: Not that I know of. The PEP [says](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) only that the closing quotes need to be on their own line.

cfgov/v1/feeds.py
chosak: Thanks for the suggestion @Scotchester. How does the change in c5bfc77 look to you?

PR #4883, opened 2019-03-08
Analytics: Create tab-navigation-tracking.js

anselmbradford:
## Additions

- Adds `tab-navigation-tracking.js`

## Testing

1. `gulp clean && gulp build` should generate `/apps/analytics-gtm/js/tab-navigation-tracking.js`


anselmbradford:
Note that the Navigation type used in this script is deprecated https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigation/type

Scotchester:
What are the consequences if `if ( window.PerformanceNavigation )` returns `false`?

PR #4886, opened 2019-03-11
Notification: Remove subtle style

anselmbradford:
There is a subtle notification style that delivers styles to all pages for the exclusive use of the notification on https://www.consumerfinance.gov/consumer-tools/financial-well-being/
This converts the subtle style to the standard notification.

## Removals

- Removes subtle notification style.

## Testing

1. Pull branch and build it `gulp clean && gulp build`.
2. Compare http://localhost:8000/consumer-tools/financial-well-being/ to https://www.consumerfinance.gov/consumer-tools/financial-well-being/

## Screenshots

Before:
Screen Shot 2019-03-11 at 3 30 45 PM

After:
Screen Shot 2019-03-11 at 3 34 20 PM

## Notes

- The spacing between the notification explanation and the list of links is very tight. We probably need to adjust this in the cf-notification styles pending designer recommendations.


anselmbradford:
Spacing issue in notification has been addressed in https://github.com/cfpb/capital-framework/pull/898, which should show up in the next CF release and update in cfgov-refresh.

anselmbradford:
Spacing issue has landed here and updated screenshot looks like:

Screen Shot 2019-04-12 at 8 26 10 AM

keelerr:
@anselmbradford this is fine with me. If additional design refinements are needed, at least they will all be consistent.

PR #4891, opened 2019-03-12
Notification: Remove bespoke regs3k notification

anselmbradford:
Regs3k has its own notification template and less file. This converts that notification into a container for the v1 notification and uses utility classes to eliminate the less file in interest of phasing out the one-off notification for regs3k.

## Changes

- Use v1 notification in place of regs3k notification.

## Testing

1. Run migrations with `python cfgov/manage migrate` and run a local server.
2. Visit http://localhost:8000/policy-compliance/rulemaking/regulations/ and compare to https://www.consumerfinance.gov/policy-compliance/rulemaking/regulations/

## Screenshots

Before:

Screen Shot 2019-03-12 at 1 31 09 PM

After:

Screen Shot 2019-03-12 at 1 31 03 PM

## Notes

- This changes the right-hand margin of the notification so it aligns with the content. I *think* this was the original intent?


Scotchester:
Shouldn't we just remove all this code and add the new notification module to `RegulationLandingPage`?

anselmbradford:
> Shouldn't we just remove all this code and add the new notification module to RegulationLandingPage?

It requires a data migration, which was deemed not a priority at this time given the effort. This is a step in that direction though.

Scotchester:
Adding a field that did not exist prior does not require a data migration.

Scotchester:
> Adding a field that did not exist prior does not require a data migration.

I made a bad assumption here. It would indeed require a data migration to automatedly migrate the existing data, but there is currently only one existing instance of the regs3k `NotificationBlock`, so we can just blow it away and re-add it immediately after deployment with minimal "downtime".

I will take over this PR for Ans, as he's heading out on vacation, and see those changes through.

willbarton:
@Scotchester @anselmbradford discussed that with me, and I think it might be worth asking the content owners if we could remove that notification from the regs pages altogether (it's been 6 months since launch). I'm not sure it's worth migrating the content of the notification from this block to another for that reason.

Scotchester:
> @Scotchester @anselmbradford discussed that with me, and I think it might be worth asking the content owners if we could remove that notification from the regs pages altogether (it's been 6 months since launch). I'm not sure it's worth migrating the content of the notification from this block to another for that reason.

That would be fine, too. Please do ask them, if you don't mind, @willbarton.

Regardless of what they say, should we add the new v1 notification molecule to `RegulationLandingPage`, anyway, for any possible future notices we want to put up on that page?

willbarton:
@Scotchester Yes, it should be added to the RegulationLandingPage too — we've had to put notices up on eRegs before (for updates that should appear but haven't yet and things like that) so it would probably be worth preserving that ability.

anselmbradford:
Force pushed an update that removes the regs notification and adds the standard notification at the top:

Screen Shot 2019-07-12 at 2 32 38 PM

To test, create a new regs landing page and note that you can now add a notification or a full-width-text:

Screen Shot 2019-07-12 at 2 38 46 PM

Ready for a review again.

cfgov/regulations3k/models/pages.py
```diff
@@ -214,7 +213,6 @@ class RegulationPage(RoutablePageMixin, SecondaryNavigationJSMixin, CFGOVPage):
 
     content = StreamField([
         ('info_unit_group', organisms.InfoUnitGroup()),
-        ('full_width_text', RegulationsFullWidthText()),
```
anselmbradford: I was unsure if this was correct, since this is in the RegulationPage class, but `RegulationsFullWidthText` only contained a notification in the `blocks.py`, which was removed, so seemed it should be removed here. I wasn't sure how to test this page type since it doesn't show up in the list when creating a page.

chosak: RegulationsFullWidthText [inherits from](https://github.com/cfpb/cfgov-refresh/pull/4891/files#diff-358ce821d7c89205bf820b087d01742cL100) plain FullWidthText, so it has other blocks defined and exists on other pages. If you're removing the notification block only you might be able to replace this line with

`('full_width_text', organisms.FullWidthText()),`

If it's helpful you can [use wagtail-inventory](http://localhost:8000/admin/inventory/?form-TOTAL_FORMS=3&form-INITIAL_FORMS=0&form-MIN_NUM_FORMS=0&form-MAX_NUM_FORMS=3&form-0-has=includes&form-0-block=regulations3k.blocks.RegulationsFullWidthText&form-1-has=includes&form-1-block=&form-2-has=includes&form-2-block=) to see all of the pages that currently have a RegulationsFullWidthText.

anselmbradford: Cool, pushed up that update in https://github.com/cfpb/cfgov-refresh/pull/4891/commits/24a652a6978c487cb8d554822c72ac9f2c67107f
Looks like that doesn't need a data migration, unless I'm looking at the wrong thing?

willbarton: It shouldn't need a data migration, since we're just throwing the data that's there away!

Scotchester: Locally, the Notification block was still visible in the page editor and it was still looking for the `regulations3k/notification.html` template when viewing the page. I had to manually remove the Notification block from the editor to get the page to preview without error. Is that expected? If so, not a big deal, we'll just need to do this on the live site, as well. Seems like we could go ahead and do it, even, given the stakeholder's response about it.

Scotchester: Belay that comment; seems to have been a caching issue. After stopping and restarting my server, everything's fine.

PR #4895, opened 2019-03-15
Remove deprecated URL from smoke tests

chosak:
The page https://www.consumerfinance.gov/about-us/advisory-groups/
no longer exists, which causes the smoke tests to fail when run with `--full` option.

To test, run this command from within a cf.gov Python environment:

```sh
$ python cfgov/scripts/http_smoke_test.py --base https://www.consumerfinance.gov --full -v
```

It will fail on master because of this page, but pass with the change proposed here.

## Checklist

### DIFF
--- a/cfgov/scripts/http_smoke_test.py
+++ b/cfgov/scripts/http_smoke_test.py
@@ -111,7 +111,6 @@ FULL_RUN = [
     '/about-us/careers/',
     '/about-us/careers/current-openings/',
     '/about-us/doing-business-with-us/',
-    '/about-us/advisory-groups/',
     '/about-us/innovation/',
     '/about-us/contact-us/',
     '/eregulations/',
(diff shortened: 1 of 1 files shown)


higs4281:
It looks as if the page got renamed (new slug):
https://www.consumerfinance.gov/about-us/advisory-committees/

higs4281:
Downside of hard-coded URL list. I wonder who did *that*?

chosak:
Thanks @higs4281, sorry I was a bit too quick on the draw. I'm going to say that it's not worth adding back in to the (arbitrary) smoke test URL list, unless others disagree.

higs4281:
I don't think it needs to be in the list.

Scotchester:
Yeah, it's not exactly a critical page.

higs4281:
The theory was to hit every URL in the top level of the megamenu (which does include `about-us/advisory-committees/`, but that list should be derived programmatically, and that should be on *my* plate.

PR #4901, opened 2019-03-19
Migrate portal-page tags

higs4281:
The Ask project needed a new tagging scheme to relate answer pages to portal pages. 
Our method is to use PortalTopic and PortalCategory models to handle the relations.

This method will allow editors to add or change portal topics and categories without need of a code deployment, and it avoids the hard-coded method of defining tag categories used in our [util.ref](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/util/ref.py) file.

## Fixtures
There are 14 json fixtures included in this merge that will be used to populate the new models and to apply tag references to answer pages. The fixtures and migration script can be dropped after this  migration is deployed.

## Other cleanups
Since the migration requires cycling through all answer pages, we'll take the opportunity to clean up other parts of the answer page admin, such as the default text for feedback modules and the naming of some admin fields and sections.

## Testing
This branch has been deployed a number of times to the DEV3 server for demo and review by product owners, and it is available for continued review there.

These steps can be used to test the migration locally:
- Load a fresh copy of the cfgov databse.
- Migrate the models and load portal topics and categories with these commands:
```bash
./cfgov/manage.py migrate
./cfgov/manage.py loaddata portal_topics
./cfgov/manage.py loaddata portal_categories
```
Then run the migration script to apply tags to answer pages. This takes 3 to 4 minutes:

```bash
./cfgov/manage.py runscript migrate_tagging
```

After that, you should be able to see portal topic and category tags applied to answer pages, and every answer page's feedback module should have this text in the "Was it helpful text" field:  

"Was this answer helpful to you?"


richaagarwal:
This looks great, @higs4281!  

My only concerns are regarding the multiselect display of the portal topics & categories (feels like they take up too much room right now, & in a way encourages them to select multiples, as opposed to just entering in a single tag, but I think we can continue to think about the best way to approach this), and what we plan to do with the non-portal topics that were included in the spreadsheet. 

I don't think those concerns should prevent us from merging this, though.

cfgov/ask_cfpb/scripts/migrate_tagging.py
```diff
@@ -0,0 +1,150 @@
...
+                row['CategoryTag2'],
+                row['CategoryTag3'],
+                row['CategoryTag4']]
+            if tag.strip()]
+        category_pks = [CATEGORY_IDS.get(tag) for tag in categories
+                        if CATEGORY_IDS.get(tag)]
+        entry = {
+            'ask_id': row.get('ask_id'),
+            'portal_ids': [portal_pk],
+            'see_all_ids': category_pks,
```
richaagarwal: This is very minor but I find the use of `see_all_ids` a little confusing here.  We've been referring to these two fields are portal topics and categories. They do have page corollaries but given that you use `CATEGORY_IDS` above it's a little inconsistent.  I know it's a one-time script, though, so it doesn't matter too much.

higs4281: You hit the nail. The see-all naming was ill-conceived, but the json taggin files used it, and since this script is disposable, I didn't think it was worth redoing the 12 json files.

higs4281:
> multiselect display

I share your concern. I'm not happy with any of the display choices here, and settled on multiselect for its simplicity. I'm hoping response from editors will help make a choice.

PR #4902, opened 2019-03-20
Add HMDA Explore page temporarily, with feature flag

schbetsy:
Collaboration with @contolini

Add a static page that duplicates what is currently served at [cf.gov/hmda/explore](https://www.consumerfinance.gov/data-research/hmda/explore). In the near future, we will be removing the static site that is currently served by Apache at [cf.gov/hmda](https://www.consumerfinance.gov/data-research/hmda/).  This is the only page from that site we will still need.

With the `HMDA_LEGACY_REVIEW` flag, it will be served at `/data-research/hmda-new/explore`, so that stakeholders can review it on content.cf.gov. When we're ready to publish the page, (and when we've updated the Apache settings not to redirect traffic for that url), we'll switch to the `HMDA_LEGACY_PUBLISH` flag, and it will serve the page at `/data-research/hmda/explore`.

## Additions

- `hmda` app (we'll be adding more to this app later)
- URLs, view, and template for the HMDA Explore page

## Testing

1. Run this branch
2. In the Wagtail admin, update Wagtail Flags to enable the `HMDA_LEGACY_REVIEW` flag
3. Visit the new page at localhost:8000/data-research/hmda-new/explore
4. In the Wagtail admin, update Wagtail Flags to disable the `HMDA_LEGACY_REVIEW` flag and enable the `HMDA_LEGACY_PUBLISH` flag
5. Visit the page at localhost:8000/data-research/hmda/explore
    - The page will be the same at both URLs

The page should be identical to the page at https://www.consumerfinance.gov/data-research/hmda/explore EXCEPT for a few details:
- The new page removes the nav bar at the top
- The new page updates the notification box above the explorer tool to include a notification that the HMDA API will be deprecated

We have mockups if you want to see them.

## Notes

- This is a temporary addition. This will only exist until the Public Data Platform is disabled (later this year)

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


cfgov/hmda/templates/orange-explorer.html
```diff
@@ -0,0 +1,2287 @@
+<!DOCTYPE html>
```
contolini: This HTML file was created by scraping https://www.consumerfinance.gov/data-research/hmda/explore and inlining all the CSS and JS to make it a single standalone file with no external dependencies.

cfgov/cfgov/settings/base.py
```diff
@@ -688,6 +691,13 @@
 
     # Test financial well-being hub pages on Beta
     'FINANCIAL_WELLBEING_HUB': [('environment is', 'beta')],
+
+    # Test migrated HMDA content pages. Delete after HMDA content launch
+    'HMDA_LEGACY_REVIEW': [{'condition': 'site',
+        'value': 'content.consumerfinance.gov:80'}],
```
contolini: At first I was questioning if this URL should appear on GH but it's elsewhere in the repo so 🤷‍♂️ .

willbarton: Theoretically this condition could be created in the admin on content rather than hard-coded.

I do wonder — I don't think we still have a separate Wagtail `Site` for content, do we?

Scotchester: I think you're right, Will. There's a Sharing Site set up at that subdomain, but I assume the site condition doesn't look at that. Even if it did, the port is 443.

chosak: An alternate way to do this would be to use `'environment is': 'content'`, which would match any hostnames that you happened to hit the content server with, and would also make it easier to test locally with, e.g.

```sh
$ DEPLOY_ENVIRONMENT=content ./runserver.sh
```

schbetsy: Thanks for the feedback. I was patterning my flag condition after other ones that I saw in Wagtail, but I'll use one of the corrected versions you mentioned instead. I'll set up the condition using the Wagtail admin instead of in code.

schbetsy:
@Scotchester @willbarton @chosak Any of you wanting to do more review? If possible, we'd like to get this merged soon, so stakeholders can see it on Content.

cfgov/cfgov/urls.py
```diff
@@ -409,7 +410,18 @@ def this_view_always_raises_http404(request, *args, **kwargs):
         'FINANCIAL_WELLBEING_HUB',
         r'^practitioner-resources/financial-well-being-resources/',
         'financial-well-being-resources'
-    )
+    ),
+
+    # Temporary: HMDA Legacy pages
+    # Will be deleted when HMDA API is retired (hopefully Summer 2019)
+    flagged_url('HMDA_LEGACY_REVIEW',
```
chosak: FWIW, there exists an (undocumented) `flagged_wagtail_template_view` helper that would allow you to eliminate the hmda/views.py file and just refer to the template directly in urls.py. That would look like this:

```py
urlpatterns = [
    ...
    url(r'data-research/hmda/explore$',
        flagged_wagtail_template_view(
            'HMDA_LEGACY_REVIEW',
            'orange-explorer.html'
        ),
        name='legacy_explorer_draft'
    ),
    ...
]
```

I'm not sure this is an improvement over what is already here, which maybe indicates that the helper could be made better somehow.

schbetsy:
Thanks for the feedback! I incorporated @chosak 's comments, and will merge when Travis is green.

PR #4905, opened 2019-03-21
Add CloudFront files.cf.gov distribution to CDN tools

willbarton:
[Short description explaining the high-level reason for the pull request]

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -478,18 +478,26 @@ GOOGLE_ANALYTICS_SITE = ''
 REGSGOV_BASE_URL = os.environ.get('REGSGOV_BASE_URL')
 REGSGOV_API_KEY = os.environ.get('REGSGOV_API_KEY')
 
-# Akamai
+# CDNs
+WAGTAILFRONTENDCACHE = {}
+
 ENABLE_AKAMAI_CACHE_PURGE = os.environ.get('ENABLE_AKAMAI_CACHE_PURGE', False)
 if ENABLE_AKAMAI_CACHE_PURGE:
-    WAGTAILFRONTENDCACHE = {
-        'akamai': {
-            'BACKEND': 'v1.models.akamai_backend.AkamaiBackend',
-            'CLIENT_TOKEN': os.environ.get('AKAMAI_CLIENT_TOKEN'),
-            'CLIENT_SECRET': os.environ.get('AKAMAI_CLIENT_SECRET'),
-            'ACCESS_TOKEN': os.environ.get('AKAMAI_ACCESS_TOKEN')
-        },
+    WAGTAILFRONTENDCACHE['akamai'] = {
+        'BACKEND': 'v1.models.akamai_backend.AkamaiBackend',
+        'CLIENT_TOKEN': os.environ.get('AKAMAI_CLIENT_TOKEN'),
+        'CLIENT_SECRET': os.environ.get('AKAMAI_CLIENT_SECRET'),
+        'ACCESS_TOKEN': os.environ.get('AKAMAI_ACCESS_TOKEN')
     }
 
+ENABLE_CLOUDFRONT_CACHE_PURGE = os.environ.get('ENABLE_CLOUDFRONT_CACHE_PURGE', False)
+if ENABLE_CLOUDFRONT_CACHE_PURGE:
+    WAGTAILFRONTENDCACHE['cloudfront'] = {
+        'BACKEND': 'wagtail.contrib.wagtailfrontendcache.backends.CloudfrontBackend',
+        'DISTRIBUTION_ID': {
+            'files.consumerfinance.gov': os.environ.get('CLOUDFRONT_DISTRIBUTION_ID_FILES')
+        }
+    }
 
 # CSP Whitelists
 
--- a/cfgov/v1/admin_views.py
+++ b/cfgov/v1/admin_views.py
@@ -1,14 +1,19 @@
+import logging
+
 from django.conf import settings
 from django.contrib import messages
 from django.http.response import HttpResponseForbidden
 from django.shortcuts import render
 
-from wagtail.contrib.wagtailfrontendcache.utils import purge_url_from_cache
+from wagtail.contrib.wagtailfrontendcache.utils import PurgeBatch
 
 from requests.exceptions import HTTPError
 
-from v1.admin_forms import AkamaiFlushForm
+from v1.admin_forms import CacheInvalidationForm
 from v1.models.akamai_backend import AkamaiBackend, AkamaiHistory
+
+
+logger = logging.getLogger(__name__)
 
 
 def cdn_is_configured():
@@ -17,14 +22,35 @@ def cdn_is_configured():
 
 
 def purge(url=None):
+    akamai_config = settings.WAGTAILFRONTENDCACHE.get('akamai', {})
+    cloudfront_config = settings.WAGTAILFRONTENDCACHE.get(
+        'cloudfront', {})
+
     if url:
-        purge_url_from_cache(url)
-        return "flushed %s" % url
+        # Use the Wagtail frontendcache PurgeBatch to perform the purge
+        batch = PurgeBatch()
+        batch.add_url(url)
+
+        # If the URL matches any of our cloudfront distributions, invalidate
+        # with that backend
+        if any(k for k in cloudfront_config.get('DISTRIBUTION_ID', {})
+               if k in url):
+            logger.info('Purging {} from cloudfront'.format(url))
+            batch.purge(backends='cloudfront')
+
+        # Otherwise invalidate with our default backend
+        else:
+            logger.info('Purging {} from akamai'.format(url))
+            batch.purge(backends='akamai')
+
+        return "invalidated %s" % url
+
     else:
-        config = settings.WAGTAILFRONTENDCACHE['akamai']
-        backend = AkamaiBackend(config)
+        # purge_all only exists on our AkamaiBackend
+        backend = AkamaiBackend(akamai_config)
+        logger.info('Purging entire site from akamai')
         backend.purge_all()
-        return "flushed entire site"
+        return "invalidated entire site"
 
 
 def manage_cdn(request):
@@ -34,35 +60,39 @@ def manage_cdn(request):
     user_can_purge = request.user.has_perm('v1.add_akamaihistory')
 
     if request.method == 'GET':
-        form = AkamaiFlushForm()
+        form = CacheInvalidationForm()
+
     elif request.method == 'POST':
         if not user_can_purge:
             return HttpResponseForbidden()
 
-        form = AkamaiFlushForm(request.POST)
+        form = CacheInvalidationForm(request.POST)
         if form.is_valid():
             url = form.cleaned_data['url']
             history_item = AkamaiHistory(subject=url or "entire site",
                                          user=request.user)
+
             try:
                 message = purge(url)
                 history_item.message = message
                 history_item.save()
                 messages.success(request, message)
-
             except Exception as e:
                 if isinstance(e, HTTPError):
                     error_info = e.response.json()
                     error_message = "{title}: {detail}".format(**error_info)
                 else:
                     error_message = repr(e)
+
                 history_item.message = error_message
                 history_item.save()
                 messages.error(request, error_message)
+
         else:
             for field, error_list in form.errors.items():
                 for error in error_list:
                     messages.error(request, "Error in %s: %s" % (field, error))
+
     history = AkamaiHistory.objects.all().order_by('-created')[:20]
     return render(request, 'cdnadmin/index.html',
                   context={'form': form,
--- a/cfgov/v1/templates/cdnadmin/index.html
+++ b/cfgov/v1/templates/cdnadmin/index.html
@@ -10,13 +10,13 @@
     {% include "wagtailadmin/shared/header.html" with title="CDN Manager" icon="tag" %}
 
     <div class="nice-padding">
-        <h2>Flush Akamai</h2>
+        <h2>Cache Invalidator</h2>
 
     {% if user_can_purge %}
         <div class="help-block help-info">
             <p>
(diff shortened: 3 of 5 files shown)


cfgov/v1/admin_views.py
```diff
@@ -17,14 +22,35 @@ def cdn_is_configured():
...
-        return "flushed %s" % url
+        # Use the Wagtail frontendcache PurgeBatch to perform the purge
+        batch = PurgeBatch()
+        batch.add_url(url)
+
+        # If the URL matches any of our cloudfront distributions, invalidate
+        # with that backend
+        if any(k for k in cloudfront_config.get('DISTRIBUTION_ID', {})
+               if k in url):
+            logger.info('Purging {} from cloudfront'.format(url))
```
Scotchester:
```suggestion
            logger.info('Purging {} from CloudFront'.format(url))
```

cfgov/v1/admin_views.py
```diff
@@ -17,14 +22,35 @@ def cdn_is_configured():
...
+        'cloudfront', {})
+
     if url:
-        purge_url_from_cache(url)
-        return "flushed %s" % url
+        # Use the Wagtail frontendcache PurgeBatch to perform the purge
+        batch = PurgeBatch()
+        batch.add_url(url)
+
+        # If the URL matches any of our cloudfront distributions, invalidate
```
Scotchester:
```suggestion
        # If the URL matches any of our CloudFront distributions, invalidate
```

cfgov/v1/admin_views.py
```diff
@@ -17,14 +22,35 @@ def cdn_is_configured():
...
+        # with that backend
+        if any(k for k in cloudfront_config.get('DISTRIBUTION_ID', {})
+               if k in url):
+            logger.info('Purging {} from cloudfront'.format(url))
+            batch.purge(backends='cloudfront')
+
+        # Otherwise invalidate with our default backend
+        else:
+            logger.info('Purging {} from akamai'.format(url))
```
Scotchester:
```suggestion
            logger.info('Purging {} from Akamai'.format(url))
```

cfgov/v1/admin_views.py
```diff
@@ -17,14 +22,35 @@ def cdn_is_configured():
...
+            batch.purge(backends='akamai')
+
+        return "Submitted invalidation for %s" % url
+
     else:
-        config = settings.WAGTAILFRONTENDCACHE['akamai']
-        backend = AkamaiBackend(config)
+        # purge_all only exists on our AkamaiBackend
+        backend = AkamaiBackend(akamai_config)
+        logger.info('Purging entire site from akamai')
```
Scotchester:
```suggestion
        logger.info('Purging entire site from Akamai')
```

cfgov/v1/templates/cdnadmin/index.html
```diff
@@ -10,13 +10,13 @@
     {% include "wagtailadmin/shared/header.html" with title="CDN Manager" icon="tag" %}
 
     <div class="nice-padding">
-        <h2>Flush Akamai</h2>
+        <h2>Cache Invalidator</h2>
 
     {% if user_can_purge %}
         <div class="help-block help-info">
             <p>
                 Enter a full URL (e.g., <code>https://www.consumerfinance.gov/about-us/</code>)
```
Scotchester:
```suggestion
                 Enter a full URL (e.g., <code>https://www.consumerfinance.gov/about-us/</code>
                 or <code>https://files.consumerfinance.gov/f/documents/example.pdf</code>)
```

Scotchester: There may be a weird spacing thing going on here related to GH's suggested change feature. Please confirm that the indentation remains at 16 spaces once/if the suggested change is accepted.

cfgov/cfgov/settings/base.py
```diff
@@ -479,18 +479,26 @@
...
+        'CLIENT_SECRET': os.environ.get('AKAMAI_CLIENT_SECRET'),
+        'ACCESS_TOKEN': os.environ.get('AKAMAI_ACCESS_TOKEN')
     }
 
+ENABLE_CLOUDFRONT_CACHE_PURGE = os.environ.get('ENABLE_CLOUDFRONT_CACHE_PURGE', False)
+if ENABLE_CLOUDFRONT_CACHE_PURGE:
+    WAGTAILFRONTENDCACHE['cloudfront'] = {
+        'BACKEND': 'wagtail.contrib.wagtailfrontendcache.backends.CloudfrontBackend',
+        'DISTRIBUTION_ID': {
+            'files.consumerfinance.gov': os.environ.get('CLOUDFRONT_DISTRIBUTION_ID_FILES')
```
chosak: Can this line (and the Akamai ones above) please be changed to use `os.environ['KEY']` instead of `os.environ.get('KEY')`? If the variable isn't in the environment, this should raise an exception, instead of passing by without error and sticking `None` in the dict.

cfgov/v1/admin_views.py
```diff
@@ -17,14 +22,35 @@ def cdn_is_configured():
...
+        # with that backend
+        if any(k for k in cloudfront_config.get('DISTRIBUTION_ID', {})
+               if k in url):
+            logger.info('Purging {} from cloudfront'.format(url))
+            batch.purge(backends='cloudfront')
+
+        # Otherwise invalidate with our default backend
+        else:
+            logger.info('Purging {} from akamai'.format(url))
+            batch.purge(backends='akamai')
```
chosak: Shouldn't `backends` be a list, per [this comment](https://github.com/wagtail/wagtail/blob/v1.13.4/wagtail/contrib/wagtailfrontendcache/utils.py#L145)?

cfgov/v1/admin_views.py
```diff
@@ -17,14 +22,35 @@ def cdn_is_configured():
...
+        # Use the Wagtail frontendcache PurgeBatch to perform the purge
+        batch = PurgeBatch()
+        batch.add_url(url)
+
+        # If the URL matches any of our cloudfront distributions, invalidate
+        # with that backend
+        if any(k for k in cloudfront_config.get('DISTRIBUTION_ID', {})
+               if k in url):
+            logger.info('Purging {} from cloudfront'.format(url))
+            batch.purge(backends='cloudfront')
```
chosak: Shouldn't `backends` be a list, per [this comment](https://github.com/wagtail/wagtail/blob/v1.13.4/wagtail/contrib/wagtailfrontendcache/utils.py#L145)?

cfgov/v1/models/caching.py
```diff
@@ -83,3 +91,19 @@ def purge_all(self):
...
+    if hasattr(settings, 'AWS_S3_CUSTOM_DOMAIN'):
+        media_base = settings.AWS_S3_CUSTOM_DOMAIN
+
+    url = urljoin(media_base, instance.url)
+
+    logger.info("Invalidating cache for " + url)
+
+    batch = PurgeBatch()
+    batch.add_url(url)
+    batch.purge(backends='cloudfront')
```
chosak: Shouldn't `backends` be a list, per [this comment](https://github.com/wagtail/wagtail/blob/v1.13.4/wagtail/contrib/wagtailfrontendcache/utils.py#L145)?

cfgov/v1/models/caching.py
```diff
@@ -83,3 +91,19 @@ def purge_all(self):
...
         )
         resp.raise_for_status()
+
+
+@receiver(post_save, sender=Document)
+@receiver(post_save, sender=CFGOVRendition)
+def cloudfront_cache_invalidation(sender, instance, **kwargs):
+    media_base = settings.MEDIA_URL
+    if hasattr(settings, 'AWS_S3_CUSTOM_DOMAIN'):
+        media_base = settings.AWS_S3_CUSTOM_DOMAIN
```
chosak: What does this do? Is `AWS_S3_CUSTOM_DOMAIN` only used for buildfiles?

rosskarchner: I think it's only used on production/content--  it's how django storages knows to generate URL's with 'https://files.consumerfinance.gov' https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#cloudfront

cfgov/cfgov/settings/base.py
richaagarwal: @chosak We already raise a value error if `AkamaiBackend` isn't initialized properly in `akamai_backend.py`.  It seems a bit overkill to prevent the entire site from running if it's misconfigured.

cfgov/cfgov/settings/base.py
```diff
@@ -479,18 +479,26 @@
...
-            'ACCESS_TOKEN': os.environ.get('AKAMAI_ACCESS_TOKEN')
-        },
+    WAGTAILFRONTENDCACHE['akamai'] = {
+        'BACKEND': 'v1.models.caching.AkamaiBackend',
+        'CLIENT_TOKEN': os.environ.get('AKAMAI_CLIENT_TOKEN'),
+        'CLIENT_SECRET': os.environ.get('AKAMAI_CLIENT_SECRET'),
+        'ACCESS_TOKEN': os.environ.get('AKAMAI_ACCESS_TOKEN')
     }
 
+ENABLE_CLOUDFRONT_CACHE_PURGE = os.environ.get('ENABLE_CLOUDFRONT_CACHE_PURGE', False)
```
richaagarwal: If both Akamai and Cloudfront are enabled, how does Wagtail know which backend to use when publishing a page?

cfgov/cfgov/settings/base.py
chosak: Thanks @richaagarwal, I didn't realize we had that check there.

cfgov/cfgov/settings/base.py
richaagarwal: Nevermind just saw the logic in `admin_views.py`!

Scotchester:
> One question: so the AkamaiHistory model is still displayed on the CDN Manager page, but is the list updated with files.cf.gov requests? If not, should it be (and be renamed to something more generic)? Users might be confused if they enter a URL on that page but then don't see it show up in the list. If we don't want to add it there, can we add some clarifying language around how the list only contains Akamai requests?

It does show up. Here's what it looks like:

Screen Shot 2019-04-04 at 18 02 43

Scotchester:
Tested this out locally with a fake distribution ID and it's working as expected, but I wonder if we should output a failure message of some kind on the page if the result of the request is definitely not a success, e.g., when testing here with a fake distribution ID and you get the message "Couldn't purge path '/f/documents/cfpb_fraud-protection-tool-for-justice-involved-individuals_handout.pdf' from CloudFront (DistributionId=fake_cloudfront_dist_id). ClientError: NoSuchDistribution The specified distribution does not exist."

I'm thinking something to the effect of "Something went wrong trying to purge the cache for the requested UR. Please contact a cf.gov developer for assistance." I suppose you could even output the results of the command so that they could relay that to the dev.

chosak:
> so the AkamaiHistory model is still displayed on the CDN Manager page, but is the list updated with files.cf.gov requests? 

> It does show up.

Thanks @Scotchester. In that case @willbarton can we rename the `AkamaiHistory` model to be something more generic, if it's no longer Akamai-specific?

cfgov/v1/models/caching.py
willbarton: Correct, it's only used for production — and it appears to be the only way to get the files.consumerfinance.gov FQDN.

willbarton:
> Thanks @Scotchester. In that case @willbarton can we rename the AkamaiHistory model to be something more generic, if it's no longer Akamai-specific?

Sure... I was trying to keep the database-related migratory changes to a minimum, especially since renaming models is non-trivial.

willbarton:
@Scotchester I tried to get to the result of the invalidation request, but I couldn't, which is why I changed the "success" language to be a successful submission of the invalidation request. I will try again — I don't recall right off why I couldn't.

willbarton:
@Scotchester Yeah, the wagtailfrontendcache CloudFront backend [does not return the result of the invalidation](https://github.com/wagtail/wagtail/blob/stable/1.13.x/wagtail/contrib/wagtailfrontendcache/backends.py#L152), all it does [is log the result](https://github.com/wagtail/wagtail/blob/stable/1.13.x/wagtail/contrib/wagtailfrontendcache/backends.py#L173-L179).

cfgov/v1/admin_views.py
willbarton: Listified in 570be79

cfgov/v1/admin_views.py
willbarton: Listified in 570be79

cfgov/v1/models/caching.py
willbarton: Listified in 570be79

cfgov/v1/admin_views.py
willbarton: These logging statements were actually intended to reference the `settings.WAGTAILFRONTENDCACHE` names. I've clarified the language on that somewhat.

cfgov/v1/admin_views.py
willbarton: Same as above, clarified in b270d39.

cfgov/v1/admin_views.py
willbarton: Same as above, clarified in b270d39.

cfgov/v1/templates/cdnadmin/index.html
willbarton: So weird that the spacing worked out like that. Fixed in ff8ebe3.

willbarton:
> I'll take it on faith that you didn't break anything in the meantime :)

That's what I always do!

PR #4906, opened 2019-03-25
Add HMDA outage banner behind feature flag

contolini:
We will be periodically taking down both HMDA's API and the HMDA Explorer page over the next few months. During outages we will redirect users to a not-yet-completed Data & Research page that will have info about our our plans to retire the HMDA API. When we automatically redirect users to this page, we'd like to add a banner at the top that mentions the outage. We're doing this with a feature flag triggered by a GET parameter called `hmda-outage`.

## Additions

- `HMDA_OUTAGE` feature flag
- Banner in `header.html` that sits alongside our other optional banners.

## Testing

1. Visit http://localhost:8000/data-research/?hmda-outage=True and you should see a banner at the top.
1. Visit http://localhost:8000/data-research/ and there should not be a banner at the top.

## Screenshots

Screen Shot 2019-03-24 at 11 23 17 PM

## Notes

- Related to https://github.com/cfpb/django-flags/issues/30

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -700,6 +700,14 @@ FLAGS = {
     # Publish new HMDA content pages
     # Delete after HMDA API is deprecated (hopefully Summer 2019)
     'HMDA_LEGACY_PUBLISH': [],
+
+    # The HMDA API and HMDA explorer pages will temporarily be taken down at
+    # TBD intervals. We use a GET parameter during downtime to trigger an
+    # explanatory banner about the outages.
+    'HMDA_OUTAGE': [
+        {'condition': 'parameter', 'value': 'hmda-outage', 'required': True},
+        {'condition': 'path matches', 'value': r'^/data-research', 'required': True}
+    ]
 }
 
 
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -704,6 +704,7 @@ FLAGS = {
     # The HMDA API and HMDA explorer pages will temporarily be taken down at
     # TBD intervals. We use a GET parameter during downtime to trigger an
     # explanatory banner about the outages.
+    # Delete after HMDA API is deprecated (hopefully Summer 2019)
     'HMDA_OUTAGE': [
         {'condition': 'parameter', 'value': 'hmda-outage', 'required': True},
         {'condition': 'path matches', 'value': r'^/data-research', 'required': True}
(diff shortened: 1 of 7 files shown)


cfgov/hmda/tests/test_outage_banner.py
```diff
@@ -0,0 +1,20 @@
...
+    def render(self, path):
+        request = RequestFactory().get(path)
+        request.flag_conditions = get_flags()
+        template = Template(
+            '{% load hmda_banners %}'
+            '{% hmda_outage_banner request %}'
+        )
+        return template.render(Context({'request': request}))
+
+    def test_banner_renders(self):
```
schbetsy: Would you mind adding an additional test to show that the banner doesn't show up when the flag conditions aren't met?

cfgov/cfgov/settings/base.py
```diff
@@ -700,6 +700,14 @@
     # Publish new HMDA content pages
     # Delete after HMDA API is deprecated (hopefully Summer 2019)
     'HMDA_LEGACY_PUBLISH': [],
+
+    # The HMDA API and HMDA explorer pages will temporarily be taken down at
+    # TBD intervals. We use a GET parameter during downtime to trigger an
+    # explanatory banner about the outages.
+    'HMDA_OUTAGE': [
+        {'condition': 'parameter', 'value': 'hmda-outage', 'required': True},
+        {'condition': 'path matches', 'value': r'^/data-research', 'required': True}
```
schbetsy: I was initially worried that the path condition is scoped too widely, since it could trigger for any page in /data-research. But since the GET param will only be added in a very specific context, I think it's fine.

cfgov/hmda/tests/test_outage_banner.py
contolini: I have a confession: this test is a farce that doesn't actually use the URL to test the flag's logic. Please don't tell @chosak. I will rewrite it to make it more useful before @chosak notices. I really hope @chosak doesn't see this.

schbetsy: I pushed an update to add tests showing that the banner conforms to the flag conditions.

schbetsy:
Any more review for this one?

willbarton:
@schbetsy the `assertIn`s that are failing in Python 3 can be replaced by `self.assertContains(response, ...)`.

schbetsy:
Thank you @willbarton!

Any other review for this one?

PR #4907, opened 2019-03-25
Remove Audience Pages

richaagarwal:
Audience pages have been deleted in production and redirects have been set up for each of them, so we can now remove the corresponding code.

To test, you can pull down the latest production dump and confirm that the following are redirected:

http://localhost:8000/ask-cfpb/audience-older-americans
http://localhost:8000/ask-cfpb/audience-parents
http://localhost:8000/ask-cfpb/audience-servicemembers
http://localhost:8000/ask-cfpb/audience-students

See GHE/CFGOV/Ask-Evo/issues/16 for more context


richaagarwal:
> I think we should be removing these audience links from the landing page.
> 
> https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/ask-cfpb/landing-page.html#L82-L84
> 
> audience_links

Ah good call, @higs4281.  I wonder if these are hardcoded anywhere else?

higs4281:
@richaagarwal Can't find any other audience links. We had already removed them from other Ask pages.

richaagarwal:
I've updated the landing page to remove references to audience pages, and rebased.  @higs4281 @virginiacc any other concerns before I merge?

PR #4910, opened 2019-03-26
Update portal-tag-migration to handle multiple topics

higs4281:
CE identified Ask CFPB answer pages to tag with multiple portal topics,
and this brings them into the fold and marks a primary topic when needed.

## Testing
- Load the latest prod db and migrate portal tags:
```bash
./cfgov/manage.py runscript migrate_tagging
```
Then edit pages to see tags applied.

Pages with multiple portal topics (and thus a primary topic designated):
- http://localhost:8000/admin/pages/6224/edit/
- http://localhost:8000/admin/pages/7798/edit/

Pages with single topics (and thus no primary): 
- http://localhost:8000/admin/pages/5928/edit/
- http://localhost:8000/admin/pages/6588/edit/

The script also has an option to run against a single Ask ID.  
To test this option, pass `--script-args` and an Ask ID:

```bash
./cfgov/manage.py runscript migrate_tagging --script-args 321
```

This should migrate and publish two pages, the English and Spanish versions of 321.

### DIFF
--- a/cfgov/ask_cfpb/scripts/migrate_tagging.py
+++ b/cfgov/ask_cfpb/scripts/migrate_tagging.py
@@ -140,7 +140,9 @@ def update_page(page, entry):
 
 
 def apply_tags(data, singleton=None):
-    """Apply tagging values to answer pages, or to a single Ask ID."""
+    """
+    Apply tags to all answer pages, or only to pages linked to a single Ask ID.
+    """
     if singleton:
         if singleton not in [d['ask_id'] for d in data]:
             return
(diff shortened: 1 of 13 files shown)


cfgov/ask_cfpb/scripts/migrate_tagging.py
```diff
@@ -96,55 +122,74 @@ def update_feedback_default(stream_value):
...
+        page.portal_topic.add(TOPIC_OBJECTS.get(topic_id))
+    for category_id in category_ids:
+        page.portal_category.add(CATEGORY_OBJECTS.get(category_id))
+    page.save_revision(user=USER).publish()
+    if initial_status == 'draft':
+        page.unpublish()
+
+
+def apply_tags(data, singleton=None):
+    """Apply tagging values to answer pages, or to a single Ask ID."""
```
richaagarwal: Can we update this docstring to something like "Apply tagging values to all answer pages, or to a single set of answer pages determined by their Ask ID"?

higs4281: √

PR #4915, opened 2019-03-28
Ask page creation

higs4281:
Add a routine for creating Ask IDs and related answer pages with self-healing URLs.

When an answer page is created, it needs to have an Ask ID to share
with its sister language and to use in creating a self-healing URL
in the form of `slug-language-id`.

The legacy Answer model previously handled Ask ID creation, and can
continue to do so even though all other content and operations have
moved to the publishable page.

## Testing
**For English**
- Navigate to the Ask CFPB landing page (http://localhost:8000/admin/pages/5702/)
- Click `ADD A CHILD PAGE`, choose `Answer page` as the page type, and enter a new
question in the page's title field, and click `SAVE DRAFT`.
- You should get back a page with '-en-[an Ask ID]' appended to the title,
and on the CONFIGURATION tab, the url should be [slugified-question]-en-[ASK ID]
- If you check the page in the Django admin, the new answer object should
show up when you call the page's `answer_base` value.

```python
from ask_cfpb.models import AnswerPage
page = AnswerPage.objects.get(pk=[NEW_PAGE_ID])
page.answer_base
```

**For Spanish**
- Navigate to the Obtener respuestas landing page (http://localhost:8000/admin/pages/5704/)
- Click `ADD A CHILD PAGE`, choose `Answer page` as the page type, and enter a new
question in the page's title field, and click `SAVE DRAFT`.
- You should get back a page with '-es-[an Ask ID]' appended to the title,
and on the CONFIGURATION tab, the url should be [slugified-question]-es-[ASK ID]


cfgov/ask_cfpb/wagtail_hooks.py
```diff
@@ -97,3 +98,44 @@ def register_export_menu_item():
...
+def create_answer_id(request, page):
+    """
+    Create an incremented Answer ID for a new AnswerPage and attach it.
+
+    Also create a sister-language page to keep languages in sync.
+    """
+    if isinstance(page, AnswerPage) and page.answer_base is None:
+        new_answer_base = Answer(
+            last_user=request.user,
+            question=page.title)
```
richaagarwal: Do we still need the Answer to store the question?

higs4281: No, but I thought it would be useful to devs who may have reason to traverse Answer. We should make Answer's __str__ method return question, so there's some context to an Answer object beyond it's primary key.

cfgov/ask_cfpb/wagtail_hooks.py
```diff
@@ -97,3 +98,44 @@ def register_export_menu_item():
...
+        new_answer_base.save()
+        new_id = new_answer_base.pk
+        if page.get_parent().slug == 'obtener-respuestas':
```
richaagarwal: This is minor but it might be nice to make this a function that takes in the parent slug ('obtener-respuestas' or 'ask-cfpb') as the argument and then returns the sister page. Then page.language and the slug could be passed in as variables instead of hardcoded.

higs4281:
@richaagarwal I refactored the create_answer_id hook, which exposed a couple weaknesses in the tests!
Thanks for the careful review.

PR #4917, opened 2019-03-28
Only download database dump it neccessary

rosskarchner:
- adds the -z flag to to the curl command in refresh-data.sh, which will make it compare the timestamps of the remote and local file, and only download if the remote file is newer.
- also adds the -R flag to preserve the remote timestamp

```
-z, --time-cond <time>

(HTTP FTP) Request a file that has been modified later than the given time and date,
or one that has been modified before that time. The <date expression> can be all sorts
of date strings or if it doesn't match any internal ones, it is taken as a filename
and tries to get the modification date (mtime) from <file> instead. 
See the curl_getdate(3) man pages for date expression details.

Start the date expression with a dash (-) to make it request for a document that is
older than the given date/time, default is a document that is newer than the
specified date/time.

If this option is used several times, the last one will be used.

-R, --remote-time

When used, this will make curl attempt to figure out the timestamp of the remote file,
and if that is available make the local file get that same timestamp.
```

## Additions

- -Z and -R flags as described above

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/refresh-data.sh
+++ b/refresh-data.sh
@@ -25,13 +25,26 @@ EOF
 
 download_data() {
     echo 'Downloading fresh production Django database dump...'
-    prod_archive="$refresh_dump_name"
-    if test -e "$prod_archive"
-       then 
-           curl -R -o "$prod_archive" -z "$prod_archive" "$CFGOV_PROD_DB_LOCATION"
-       else
-           curl -R -o "$prod_archive" "$CFGOV_PROD_DB_LOCATION"
-       fi
+    skip_download=0
+
+    # If the file already exists, check its timestamp, and skip the download
+    # if it matches the timestamp of the remote file.
+    if test -e "$refresh_dump_name"; then
+        timestamp_check=$(curl -s -I -R -z "$refresh_dump_name" "$CFGOV_PROD_DB_LOCATION")
+        if [[ "$timestamp_check" == *"304 Not Modified"* ]]; then
+            echo 'Skipping download as local timestamp matches remote timestamp'
+            skip_download=1
+        fi
+    fi
+
+    if [[ "$skip_download" == 0 ]]; then
+        curl -R -o "$refresh_dump_name" "$CFGOV_PROD_DB_LOCATION"
+    fi
+}
+
+check_data() {
+    echo 'Validating local dump file'
+    gunzip -t "$refresh_dump_name"
 }
 
 refresh_data() {
@@ -58,4 +71,5 @@ else
     fi
 fi
 
+check_data
 refresh_data
--- a/refresh-data.sh
+++ b/refresh-data.sh
@@ -28,9 +28,9 @@ download_data() {
     prod_archive="$refresh_dump_name"
     if test -e "$prod_archive"
 	    then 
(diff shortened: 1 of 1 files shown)


refresh-data.sh
```diff
@@ -26,7 +26,12 @@ EOF
 download_data() {
     echo 'Downloading fresh production Django database dump...'
     prod_archive="$refresh_dump_name"
-    curl -o "$prod_archive" "$CFGOV_PROD_DB_LOCATION"
+    if test -e "$prod_archive"
```
chosak: Would you mind terribly replacing these tabs with 4 spaces? Per [the development repo](https://github.com/cfpb/development) we seem to have standardized on that.

rosskarchner: done. I think my editor added the tabs

chosak:
One thought: if you happen to Ctrl+C while the download is happening, and then you're left with an invalid dump locally, and re-run the script, it won't re-download, because the timestamps line up. Is there any kind of equivalent check for file size? Or, if not, how hard would it be to add some kind of `--force` option to the script to force it to re-download?

rosskarchner:
> One thought: if you happen to Ctrl+C while the download is happening, and then you're left with an invalid dump locally, and re-run the script, it won't re-download, because the timestamps line up. Is there any kind of equivalent check for file size? Or, if not, how hard would it be to add some kind of `--force` option to the script to force it to re-download?

I can't find a curl option like that-- I wonder if it's sufficient to check the validity of the gzip file (which it seems like you can do with `gunzip -t`)

chosak:
Thanks @rosskarchner, that's a good idea. It'd also be nice if we let the user know when the download is being skipped, so people aren't confused about what is happening.

I played around a bit and came up with [this](https://github.com/cfpb/cfgov-refresh/compare/refresh-smarter...chosak:refresh-smarterer); it uses an initial silent HEAD request to check the timestamp, and parses the output to determine if the file should be re-downloaded or not. What do you think?

rosskarchner:
In that version, if check_data fails, the user is left to remove the corrupt file?

chosak:
Yes, I didn't think to delete it.

Is that something we'd definitely want to do? For example, I know some people like to run the script by pointing it at a local dump, like `./refresh_data.sh path/to/dump.sql.gz`. Let's say someone happened to unzip their local dump to work on it, and then erroneously ran `./refresh_data.sh path/to/dump.sql`. Would it be expected behavior that their local file would be deleted?

rosskarchner:
I'm thinking your *probably* right-- deleting the file might be a shade too clever

rosskarchner:
I merged @chosak's improvements branch and rebased, so this is pretty much his code now ;)

PR #4937, opened 2019-04-10
Run full tox tests on Travis

chosak:
This change modifies how the Python tests are run on Travis so that the "slow" version is run instead of the "fast" version. The "slow" version runs the full set of Django migrations.

It also modifies our Travis configuration to use [tox-travis](https://tox-travis.readthedocs.io/en/stable/), which allows us to simplify the .travis.yml a bit. This PR also removes the separate `missing-migrations` environment in favor of moving the migrations check into the standard `unittest` environment, but only having it run on Travis using Python 2.7. This is accomplished via a bit of environment variable trickery.

## Testing

- When running `tox -e fast` locally, backend tests will run against Python 2 without running migrations and without checking for missing migrations.
- When running `tox` locally, backend tests will run against both Python 2 and 3 with migrations but without checking for missing migrations.
- When Travis runs, it'll simultaneously run these three steps:
   - Backend tests against Python 2 with migrations and with checking for missing migrations
   - Backend tests against Python 3 with migrations but without checking for missing migrations (we can't easily do this yet)
   - Frontend build and tests

## Screenshots

Travis run for this PR:

image

It doesn't seem to take significantly more time than our current builds do.

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: python
 dist: trusty
 
+python:
+- 2.7
+- 3.6
+
 git:
   depth: 10
 
@@ -41,25 +45,21 @@ env:
     - COVERALLS_PARALLEL=true
     - PGPORT=5433
     - TEST_DATABASE_URL=postgres://travis:travis@localhost:5433/travis
+    - TRAVIS_CHECK_MIGRATIONS="./manage.py makemigrations --dry-run --check"
+  matrix:
+    - RUNTEST=backend
+    - RUNTEST=frontend
+
+matrix:
+  exclude:
+    # We only want to run the frontend build once.
+    - python: 2.7
+      env: RUNTEST=frontend
 
 jobs:
-  fast_finish: true
   include:
-    - stage: Run Tests
-      env: RUNTEST=frontend
-      if: type = pull_request
-      python: 2.7
-    - env:
-        - RUNTEST=backend
-        - TEST_RUNNER=cfgov.test.StdoutCapturingTestRunner
-      if: type = pull_request
-      python: 2.7
-    - env:
-        - RUNTEST=backend3
-        - TEST_RUNNER=cfgov.test.StdoutCapturingTestRunner
-      if: type = pull_request
-      python: 3.6
-    - stage: Docs Deploy
+    # This stage will run after the default testing stage succeeds.
+    - stage: Build docs and push them to GitHub Pages
       if: type != pull_request
       env: RUNTEST=docs
       deploy:
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -10,6 +10,7 @@ additionally, changing field names or types on an existing block will require a
 ## Table of contents
 
 1. [Reference material](#reference-material)
+1. [Do I need to create a migration?](#do-i-need-to-create-a-migration)
 1. [Schema migrations](#schema-migrations)
 1. [Data migrations](#data-migrations)
    1. [Wagtail-specific consideration](#wagtail-specific-considerations)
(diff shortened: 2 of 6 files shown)


willbarton:
> It doesn't seem to take significantly more time than our current builds do.

yay

willbarton:
I wonder if we want to just always run the migrations check if you're running the full `tox`. I *think* that would be my inclination, since the context for running `tox` is to run with all migrations anyway.

chosak:
@higs4281 @willbarton thanks for the reviews.

> folks might be used to running `tox -e check-migrations`

That's a great point. Is this something that a new documentation tip could address, maybe [here](https://cfpb.github.io/cfgov-refresh/development-tips/)?

> always run the migrations check if you're running the full `tox`

I considered doing this -- it seems reasonable, and I can see pros and cons. If I recall, the reason for not including the check in tox originally was that folks might like to iteratively unit test without needing to create a migration, preferring to leave it for the end. Making the change you suggest would break that use case for anyone who runs the full `tox` as part of their iterative process. I don't know if there's anyone who does this.

higs4281:
Running full `tox` will fail regardless if a model or column has been added, although it won't fail if there's a migration required for a change to help_text.

I have doubts that devs routinely run full `tox`, but I think it's common to get snagged on a missing migration.

Would it be possible to add the migration check to `tox -e lint`?

willbarton:
Anecdotally, I run `tox -e fast path.to.specific.tests` when I'm iterating on code.

chosak:
@higs4281 the downside of adding the check to the `lint` environment is that we'd have to add all of the Django/Wagtail requirements there too. Currently that environment has [its own](https://github.com/cfpb/cfgov-refresh/blob/444bd5caa6ec62acc09d6c8b92ead4181824e74a/tox.ini#L97) smaller set of dependencies.

That might be okay, but then we'd also have to add some additional logic to skip the missing migrations check when running linting under Python 3 (unless we do a bunch of work to fix that).

Looking at this again, I fear that my attempt to be clever and use tox-travis has actually made it harder to support the `tox -e missing-migrations` use case. It might be easier to support that without doing that piece. I'll take another crack at this.

higs4281:
I do think it's pretty painless to run `./cfgov/manage.py makemigrations --dry-run`, since makemigrations is a well-used command.

chosak:
@higs4281 good point. It'd be nice to know who uses `tox -e missing-migrations`, and how much they'd miss it if it were gone. For better or worse it's not documented anywhere now.

@cfpb/cfgov-backends @cfpb/cfgov-frontends do you use `tox -e missing-migrations` in your work? How would you feel if it went away, and was replaced by a documented recommendation to run `cfgov/manage.py makemigrations --dry-run` instead?

Scotchester:
> [D]o you use `tox -e missing-migrations` in your work? How would you feel if it went away, and was replaced by a documented recommendation to run `cfgov/manage.py makemigrations --dry-run` instead?

I prefer to just run `./cfgov/manage.py makemigrations --dry-run` if I suspect I might need one. Probably because I assume, perhaps incorrectly, that doing that will be faster than spinning up `tox`.

willbarton:
@chosak I do favor removing as much magic as possible. I think we should encourage folks to run `cfgov/manage.py makemigrations --dry-run` manually. Since if you *are* missing a migration you'll need to run `makemigrations` anyway.

I think at this point I don't have a good suggestion of how to add it to CI though.

chosak:
> I don't have a good suggestion of how to add it to CI though.

As written, this PR does do the migrations check via [this bit of trickery](https://github.com/cfpb/cfgov-refresh/blob/d64b6f57aa567f95b3264624f1261f33dc22b9f1/.travis.yml#L48), including it in the slow tox run but only on Travis. If the team is okay with saying farewell to `tox -e missing-migrations`, then I could add documentation about the recommended `cfgov/manage.py makemigrations` command here.

chosak:
@higs4281 @Scotchester @willbarton I've added another commit with some documentation on `makemigrations --dry-run`, as a new section on [this page](https://cfpb.github.io/cfgov-refresh/migrations/):

image

Scotchester:
@chosak Looks good!

PR #4940, opened 2019-04-10
Expose language value on blog pages.

higs4281:
Blog pages' parent class, the AbstractFilterPage, hides a page's language
value. This may have been an oversight, but it's becoming an unfortunate choice
as we translate more site pages, including blogs, into other languages.

This PR just exposes the existing `language` field in the page admin,
allowing editors to set the proper language for a translated page.
This in turn will enable automatic translation of template labels and
other microcopy to match the chosen language.

Exposing `language` on the CONFIGURATION panel should have no immediate effects itself.

## Testing
Open a blog page, such as <http://localhost:8000/admin/pages/10404/edit/>, and scroll to the bottom of the page's CONFIGURATION tab. You should be able to see and change the language value.

### DIFF
--- a/cfgov/v1/models/learn_page.py
+++ b/cfgov/v1/models/learn_page.py
@@ -67,6 +67,7 @@ class AbstractFilterPage(CFGOVPage):
             FieldPanel('comments_close_by'),
         ], 'Relevant Dates', classname='collapsible'),
         MultiFieldPanel(Page.settings_panels, 'Scheduled Publishing'),
+        FieldPanel('language', 'Language'),
     ]
 
     # This page class cannot be created.


Scotchester:
Is there a reason we don't inherit the CFGOVPage `settings_panels`, as we do with other page types?

higs4281:
@Scotchester 
Not sure why AbstractFilterPage was given a generate_edit_handler() method, but it would have been nice if it inherited the way most other pages do.

Scotchester:
@higs4281 Should we change that in this PR? It seems like we just need to inherit the CFGOVPage `settings_panels`, which already include `language`, and then add a few filterable-page-specific things. But it may be easier said than done.

higs4281:
@Scotchester Reworking the `generate_edit_handler` function for 2K+ pages (blog and legacy blog) seems risky and outside the scope of this PR, whose goal is to expose one field.

Scotchester:
I don't even understand what the purpose of that function is, but you're right, this is a fine solution for now.

richaagarwal:
Let's make a note (mental or otherwise) of the maybe unnecessary `generate_edit_handler` function - I think maybe at one point it was meant to be used to pass in different options for the content panel?  There's a lot else here we could clean up so I agree it's best to do this as a separate thing, if we do anything.

PR #4946, opened 2019-04-12
Add translations for portal topic and portal headings

higs4281:
Topic and portal headings are used in common by English and Spanish
answer pages, and each has an English and Spanish heading that can be set
by an editor when creating a new tag.

Here we add the headings to our translation `django.po` files so that templates
can call a topic tag and get the right translation.

This also enables translations on Ask's TagResultsPage to allow those
pages to use a single template when Spanish Ask is migrated.


cfgov/ask_cfpb/models/pages.py
```diff
@@ -288,6 +288,18 @@ class TagResultsPage(RoutablePageMixin, AnswerResultsPage):
 
     objects = CFGOVPageManager()
 
+    def set_language(self):
```
richaagarwal: Would it make sense to add this to `CFGOVPage` directly? Or is it only needed on a couple page types?

higs4281: CFGOVPage already [has activization](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/models/base.py#L224-L225), but it's in the `serve` method, which routable pages do not call. That's why they need their own.

The current routables are probably going away soon, so I'm not too concerned about repeated code at this point. I'm looking for a good way to add translations to all routables. Unfortunately the mixin isn't the place.

PR #4963, opened 2019-04-23
Add routable pages for Ask topic see-all searching

higs4281:
Ask category pages are going away, to be replaced by consumer-tools portal pages. 

The currently planned 11 portal pages divide up the Ask CFPB universe among them, leaving some of the portals (e.g. mortages) with hundreds of associated answers. To help users narrow down that universe, we're adding portal topic search pages to let users drill down by 5 new categories and drill further by entering search terms.

This work continues the unification of English and Spanish handling, so that both languages are served from the same templates. Answer content for each language is maintained on separate pages, but the languages will share the same templates and design. Some template microcopy, such as element labels, are translated via the Django translation framework and the included `django.po` file.

The code and associated Wagtail pages have been deployed to DEV3 for review. A few examples:

#### Spanish
- [DEV3]/es/obtener-respuestas/prestamos-para-vehiculos/respuestas/
- [DEV3]/es/obtener-respuestas/cuentas-bancarias/respuestas/

#### English
- [DEV3]/consumer-tools/auto-loans/answers/
- [DEV3]/consumer-tools/bank-accounts/answers/

## Testing

To test locally, you'll need to check out the branch, get a recent production db load, migrate and run an index update, and add the related pages. After checking out the branch:
```bash
./cfgov/manage.py migrate
./cfgov/manage.py update_index ask_cfpb --remove
./cfgov/manage.py runscript create_see_alls --script-args 'es'
./cfgov/manage.py runscript create_see_alls --script-args 'en'

The script will create the English and Spanish search pages and, where needed, their parent pages.

### CODE REVIEW

File: cfgov/ask_cfpb/models/pages.py [hunk 1]
```diff
@@ -241,6 +244,210 @@ def page_js(self):
...
 
 
+class PortalSearchPage(
+        RoutablePageMixin, SecondaryNavigationJSMixin, CFGOVPage):
+    """
+    A routable page type for Ask CFPB portal search ("see-all") pages.
+    """
+
+    objects = CFGOVPageManager()
+    portal_topic = models.ForeignKey(
```
chosak: Could you specify `related_name='portal_search_pages'` or something equivalent, to make the reverse relation a little nicer? Without specifying something, the default on PortalTopic objects would be something like `PortalTopic.portalsearchpage_set`, which isn't as night.

It also occurs to me that, if there's always a 1:1 relationship between PortalSearchPages and PortalTopics, in theory this could be a [OneToOneField](https://docs.djangoproject.com/en/dev/ref/models/fields/#django.db.models.OneToOneField), but that might cause problems with, say, experimenting with other pages or having old pages to the trash.

cfgov/ask_cfpb/models/pages.py
```diff
@@ -241,6 +244,210 @@ def page_js(self):
...
+            deactivate_all()
+        return super(PortalSearchPage, self).get_context(
+            request, *args, **kwargs)
+
+    def get_nav_items(self, request, page):
+        """Return nav items sorted by an arbitrary order."""
+        pk_sort_order = [1, 4, 5, 2, 3]
+        hand_sorted_categories = []
+        for pk in pk_sort_order:
+            category = PortalCategory.objects.get(pk=pk)
```
chosak: Doing the loop this way means that you'll have 5 separate database queries, one for each PortalCategory. You could instead do something like `PortalCategory.objects.values_list('pk', 'heading')` to pull back all PKs and headings in a single query, and then work off of that list.

cfgov/ask_cfpb/models/pages.py
```diff
@@ -241,6 +244,210 @@ def page_js(self):
...
+        if self.language != 'en':
+            activate(self.language)
+        else:
+            deactivate_all()
+        return super(PortalSearchPage, self).get_context(
+            request, *args, **kwargs)
+
+    def get_nav_items(self, request, page):
+        """Return nav items sorted by an arbitrary order."""
+        pk_sort_order = [1, 4, 5, 2, 3]
```
chosak: If these primary keys are going to be hardcoded in code like this, dependent on the fixture file, what is the advantage of using a PortalCategory model versus just defining a hardcoded Python module that defines the categories?

cfgov/ask_cfpb/models/pages.py
higs4281: - 1:1
    - The portals are in flux, to put it mildly, and some pages are likely to be trashed as it plays out.

- related_name
    - √

cfgov/ask_cfpb/models/pages.py
higs4281: @chosak I'm wondering what the best setup would be If, hypothetically, the arbitrary order were changed or we were suddenly allowed to sort alphabetically.

chosak: Is your concern that those changes would require a code change and redeploy? Wouldn't that still be required in this case?

higs4281: True that code change is inevitable. I was aiming to have an ideal structure in place if we were freed from arbitrary order. And to me the ideal is a category lineup that can be managed by an editor without deployments. That was one goal of PortalCategory and PortalTopic.

Maybe adding an editable sort-order field to PortalCategory could accomplish both?

cfgov/ask_cfpb/models/pages.py
```diff
@@ -241,6 +244,211 @@ def page_js(self):
...
+            'paginator': paginator,
+            'current_page': page_number,
+            'get_secondary_nav_items': self.get_nav_items,
+        })
+        return TemplateResponse(
+            request,
+            self.get_template(request),
+            context)
+
+    @route(r'^(?P<category>[^/]+)/$')
```
richaagarwal: There's a lot of repeated code here and in `portal_topic_page` above.   Would it be possible to use something reusable like what I did originally in my `see-all` branch: https://github.com/cfpb/cfgov-refresh/compare/see-all#diff-e83ae18f3e2ce1056a644da835042016R289

cfgov/ask_cfpb/models/pages.py
```diff
@@ -241,6 +244,211 @@ def page_js(self):
...
+    def portal_category_page(self, request, **kwargs):
```
richaagarwal: There's a lot of repeated code here and in portal_topic_page above. Would it be possible to use something reusable like what I did originally in my see-all branch: https://github.com/cfpb/cfgov-refresh/compare/see-all#diff-e83ae18f3e2ce1056a644da835042016R289

cfgov/ask_cfpb/models/pages.py
```diff
@@ -241,6 +244,211 @@ def page_js(self):
...
+            count = sqs.count()
+            if count == 0:
+                # No results, so let's try to suggest a better query
+                suggestion = SearchQuerySet().models(
+                    AnswerPage).spelling_suggestion(search_term)
+                if suggestion == search_term:
+                    suggestion = None
+                elif (request.GET.get('correct', '1') == '1' and
+                        flag_enabled('ASK_SEARCH_TYPOS', request=request)):
+                    sqs = self.query_base.filter(content=suggestion)
```
richaagarwal: This is no longer filtering with the provided category.  I think refactoring the search code as I did in https://github.com/cfpb/cfgov-refresh/tree/refactor-search-suggestion (not functional yet, I know, but as a proof of concept) might make this code more workable and easier to test.

higs4281: I fixed the filter, and we can refactor.

cfgov/ask_cfpb/models/pages.py
higs4281: Thank you for spotting this. I refactored to piggyback on the single query used in the `category_map` property, which is now an ordered dict that implements the custom sort.

I still have hope that we can lose this custom sort.

richaagarwal:
@higs4281 As discussed, I opened up a pull request against this branch that takes care of some of the issues I brought up here - https://github.com/cfpb/cfgov-refresh/pull/4966 - I'd ideally like to get that reviewed/merged in before we merge here.

cfgov/ask_cfpb/models/pages.py
```diff
@@ -241,6 +247,226 @@ def page_js(self):
...
+            request,
+            self.get_template(request),
+            context)
+
+    @route(r'^(?P<category>[^/]+)/$')
+    def portal_category_page(self, request, **kwargs):
+        context = self.get_context(request)
+        category_slug = kwargs.get('category')
+        self.portal_category = self.category_map.get(category_slug)
+        heading = self.get_heading(self.portal_category)
```
virginiacc: I noticed that the portal category pages show the same title as the portal topic page. Could we either set the page's title using the heading value (or some variation on it) here so the [title tag will be correctly populated](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_layouts/base-common.html#L53) or override the template's `title` block to do the same thing?

cfgov/ask_cfpb/models/pages.py
richaagarwal: I like the idea of adding a sort field to PortalCategory, if it's a must that they're sorted in a particular way.

cfgov/ask_cfpb/fixtures/portal_categories.json
```diff
@@ -4,79 +4,39 @@
     "pk": 1,
     "fields": {
         "heading": "Basics",
-        "heading_es": ""
+        "heading_es": "Lo b\u00e1sico"
```
richaagarwal: Do we need these updates, or this fixture more generally, now that they've been initialized in the Wagtail admin?

higs4281: They're useful for testing.

PR #4955, opened 2019-04-16
Replace Google Maps with Mapbox API on Events page template

cfarm:
[Short description explaining the high-level reason for the pull request]

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [ ] PR has an informative and human-readable title
- [ ] Changes are limited to a single goal (no scope creep)
- [ ] Code can be automatically merged (no conflicts)
- [ ] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [ ] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated
- [ ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Internet Explorer 9, 10, and 11
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS
- [ ] Flexible from small to large screens
- [ ] No linting errors or warnings
- [ ] JavaScript tests are passing

PR #4967, opened 2019-04-24
Replace Google Maps with Mapbox on events pages

contolini:
Replaces Google Map's API with Mapbox's to generate static map images on event pages. Mapbox can only generate static map images from GPS coordinates so we first have to use their geocoding API to get coordinates from the provided city/state.

This PR is a continuation of #4955. I took @cfarm's geocoding logic and moved it into the EventPage model. This preserves the functionality of the `location_image_url()` method so we don't have to touch any templates or worry about no-JS users.

@chosak has big plans to fork a widget like [wagtailgmaps](https://github.com/springload/wagtailgmaps) and port it to Mapbox to improve our handling of addresses but that's a task for another day.

## Additions

- New `venue_coords` field that isn't shown to the user. It's calculated and stored when an event page is saved.

## Testing

1. `tox -e fast v1.tests.models.test_event_page`
1. Define [`MAPBOX_ACCESS_TOKEN`](https://GHE/gist/contolini/b0697be36fe0f62bd835ca93b3e83574) in `.env`
1. Visit the [archive event page](http://localhost:8000/about-us/events/archive-past-events/) and the map thumbnails should have little Mapbox logos in the corner instead of Google Maps.

## Screenshots

Screen Shot 2019-04-24 at 11 06 59 AM


cfgov/v1/models/learn_page.py
```diff
@@ -301,20 +304,46 @@ class EventPage(AbstractFilterPage):
...
+        geocode_location_url = geocode_api_url + location + '.json'
+
+        params = {'access_token': settings.MAPBOX_ACCESS_TOKEN}
+        response = requests.get(geocode_location_url, params=params)
+
+        if response.status_code != 200:
+            return venue_coords
+
+        try:
+            geo_data = json.loads(response.text)
```
chosak: You can simplify this by using the built-in [`response.json()`](https://2.python-requests.org//en/master/user/quickstart/#json-response-content) method.

cfgov/v1/models/learn_page.py
```diff
@@ -301,20 +304,46 @@ class EventPage(AbstractFilterPage):
...
 
+    def get_venue_coords(self):
+        # Default to Washington DC coordinates
+        venue_coords = '-77.039628,38.898238'
+
+        if not self.venue_city or not self.venue_state:
+            return venue_coords
+
+        location = '{} {}'.format(self.venue_city, self.venue_state)
+        geocode_api_url = 'https://api.mapbox.com/geocoding/v5/mapbox.places/'
```
chosak: Per the [Mapbox API documentation](https://docs.mapbox.com/api/search/#endpoints), we need to use the `mapbox.places-permanent` endpoint because we are going to store these coordinates in the database. We already use that endpoint elsewhere in the housing counselor app [here](https://github.com/cfpb/cfgov-refresh/blob/61ae49a621dd231cc0902613284df21f64cb58d5/cfgov/housing_counselor/geocoder.py#L175).

cfgov/v1/models/learn_page.py
```diff
@@ -301,20 +304,46 @@ class EventPage(AbstractFilterPage):
...
-            'scale': scale,
-            'size': size,
-            'zoom': zoom
-        }
-        url = 'https://maps.googleapis.com/maps/api/staticmap?'
-        return '{url}{options}'.format(
-            url=url,
-            options=urlencode(options)
+        if not self.venue_coords:
+            self.save()
```
chosak: This call worries me a bit. I don't think it's a good idea to `save()` the page model _at render time_ if the coordinates aren't available. For example, how does this work when you are previewing a new draft page, or a draft revision of an existing page? Will it save a version of the page just because you previewed?

contolini: Makes sense! I'll remove it and we can write a script to update old events.

cfgov/v1/models/learn_page.py
```diff
@@ -301,20 +303,44 @@ class EventPage(AbstractFilterPage):
...
+        venue_coords = '-77.039628,38.898238'
+
+        if not self.venue_city or not self.venue_state:
+            return venue_coords
+
+        location = '{} {}'.format(self.venue_city, self.venue_state)
+        api = 'https://api.mapbox.com/geocoding/v5/mapbox.places-permanent/'
+        location_api_url = api + location + '.json'
+
+        params = {'access_token': settings.MAPBOX_ACCESS_TOKEN}
```
willbarton: Should we test here to see if `MAPBOX_ACCESS_TOKEN` is defined before we send the request?

contolini: Yesh, I'll check for it on L310.

contolini:
Codecov front end reporting can be ignored.

contolini:
Thanks @willbarton! The data migration works great.

PR #4970, opened 2019-04-26
Move request.path handling from base template to template tags

willbarton:
This PR moves `request.path` splitting from inside the base template to template tags. It also forces ascii-encoding on the result of the splitting to avoid issues where paths have non-ascii characters and subsequent template lookups based on those paths cause encoding errors resulting in a 500 for end-users. See discussion in the cfgov-alerts channel.

## Checklist

### DIFF
--- a/cfgov/v1/templatetags/app_urls.py
+++ b/cfgov/v1/templatetags/app_urls.py
@@ -1,0 +1,14 @@
+from django import template
+
+
+register = template.Library()
+
+
+@register.simple_tag
+def app_url(request):
+    return request.path.split('/')[1].encode('ascii', 'ignore')
+
+
+@register.simple_tag
+def app_page_url(request):
+    return request.path[1:].encode('ascii', 'ignore')
--- a/cfgov/v1/templatetags/app_urls.py
+++ b/cfgov/v1/templatetags/app_urls.py
@@ -1,3 +1,5 @@
+import six
+
 from django import template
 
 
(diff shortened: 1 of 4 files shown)


cfgov/v1/templatetags/app_urls.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+@register.simple_tag
+def app_url(request):
+    return request.path.split('/')[1].encode('ascii', 'ignore')
+
+
+@register.simple_tag
+def page_url(request):
+    return request.path[1:].encode('ascii', 'ignore')
```
willbarton: `ignore` is my solution here to avoid the unicode errors we were seeing — it's brute-force, but effective. As you can see in the tests below `päge` will simply become `pge`. If/when we need to support unicode characters in these paths, we should probably be a little bit more introspective about our approach (and figure out why unicode paths don't work within our Apache/WSGI environment right now).

higs4281: I think this will break the search-by-tags search, because tags with non-ascii characters won't be found.

Could we use `urllib.quote` to make sure the URL is ascii but preserve the characters?

willbarton: This shouldn't break URL handling for those pages because it's only used for Jinja2 JavaScript file includes [here](https://github.com/cfpb/cfgov-refresh/pull/4970/files#diff-89b43ee86bbd4952dc44cf751cfd775cR99) and [here](https://github.com/cfpb/cfgov-refresh/pull/4970/files#diff-89b43ee86bbd4952dc44cf751cfd775cR119). It won't affect anything else.

higs4281: √ Comment rescinded!

PR #4971, opened 2019-04-26
Restore Spanish templates for Spanish search-by-tag results.

higs4281:
## HOLD

I think we should wait to let #4970 do its job. Then this roll-back shouldn't be necessary.

This sends traffic for the now-unlinked-to Spanish search-by-tags page back to the Spanish /es/ template path, which doesn't involve file-system lookups based on URL.

## Test
Try a search-by-tags URL with diacriticals locally:

<http://localhost:8000/es/obtener-respuestas/buscar-por-etiqueta/discriminaci%C3%B3n_de_cr%C3%A9dito/>


higs4281:
Fix not needed.

PR #4973, opened 2019-04-27
Add a portal_topic field to SublandingPage

higs4281:
The portal_topic relation will allow an editor to assign a portal page
to a topic, which will define portal pages in Wagtail and avoid having
to have a hard-coded list of them. Portal pages use the SublandingPage model, 
and they can have SublandingPage siblings that are not portals.

The relation will also enable creation of links, labels and nav items
where needed, in English and Spanish.

## Testing
With a fresh db load, run the `create_see_alls` script.
```bash
./cfgov/manage.py runscript create_see_alls --script-args 'es'
./cfgov/manage.py runscript create_see_alls --script-args 'en'
```
Now, for our 11 portal pages, the template (/cfgov/jinja2/v1/sublanding-page/index.html) will have access to `page.portal_topic` and its relations. 

## Notes
- This also updates the create_see_alls.py script to add portal_topic relations to our 11 portals, whether live or draft.
- The `portal_category_page` routable is adjusted so that those pages get markup titles that  differentiate from their parent `portal_topic_page`.


cfgov/ask_cfpb/models/pages.py
```diff
@@ -340,7 +340,12 @@ def portal_category_page(self, request, **kwargs):
             portal_topics=self.portal_topic.heading,
             language=self.language,
             portal_categories=self.portal_category.heading)
-        return self.get_results(request)
+        # call results here to activate translations, to get the right title
```
richaagarwal: Why do translations need to be activated?

higs4281: Ah great point. I was initially including "answers" in the category title, which needed to be translated, but we just decided to ditch that, so we don't need to activate early.

higs4281: Early `get_results` invocation removed in 96445fc

PR #4974, opened 2019-04-29
Make seach terms safe by disallowing unsafe characters.

higs4281:
Adding a list of unsafe characters and a function to remove them from queries.

This satisfies internal issue #50.

## Note
I removed the "Clean" import and usage from search.py, because Clean is [included by default](https://django-haystack.readthedocs.io/en/master/inputtypes.html#clean) when passing a string to SearchQuerySet.


richaagarwal:
I love how simple this change is :)

PR #4977, opened 2019-04-30
Remove unused notification classes

anselmbradford:
`m-notification__default` is not a class that exists. `m-notification` on its own without a modifier IS the default.

## Removals

- Removes `m-notification__default` from the docs.
- Removes `data-js-hook="state_atomic_init"` from code examples, which should be added via JS.


anselmbradford:
CF PR https://github.com/cfpb/capital-framework/pull/916

PR #4981, opened 2019-05-01
Adjust Ask breadcrumbs to lead to portal pages.

higs4281:
Currently, Ask CFPB breadcrumbs on answer and category pages lead back
to categories and the ask-cfpb root.

When we decommission category and subcategory pages, breadcrumbs should
lead back to portal pages instead, falling back to see-all pages if necessary.

## Testing
This commit has been deployed to the internal DEV3 server for review.

You can also check breadcrumbs locally:

First, turn the new breadcrumbs on by activating the ASK_CATEGORIES_OFF flag: 
1. In the Wagtail admin, go to **Settings --> Flags**
2. Scroll to the bottom and click on ADD A CONDITION
3. Select Flag ASK_CATEGORIES_OFF
4. Set Condition name to **boolean** and Expected value to **True**
5. Hit SAVE NEW CONDITION

Then check some answer page breadcrumbs:

Answer page with live portal has breadcrumb back to portal:
- http://localhost:8000/ask-cfpb/i-owe-more-on-my-current-loan-than-my-current-vehicle-is-worth-what-do-i-need-to-know-if-i-buy-a-new-vehicle-en-779/

Page with no live portal points to its see-all topic page:
- http://localhost:8000/ask-cfpb/what-is-a-payday-loan-en-1567/

Spanish pages (none have live portals) point to see-all page:
- http://localhost:8000/es/obtener-respuestas/debo-mas-en-mi-prestamo-de-lo-que-vale-mi-vehiculo-que-debo-saber-si-compro-un-vehiculo-nuevo-es-779/
- http://localhost:8000/es/obtener-respuestas/me-gustaria-poder-tener-la-ayuda-de-mis-amigos-o-familiares-con-el-pago-de-mis-facturas-y-actividades-bancarias-cuales-son-mis-opciones-es-1145/

## Notes
- This puts the switch behind a feature flag, `ASK_CATEGORIES_OFF`
- I moved one-off scripts that have been used to `/deprecated/`, for later removal.


cfgov/ask_cfpb/models/pages.py
```diff
@@ -687,14 +739,19 @@ class AnswerPage(CFGOVPage):
     objects = CFGOVPageManager()
 
     def get_context(self, request, *args, **kwargs):
+        portal_topic = self.primary_portal_topic or self.portal_topic.first()
```
richaagarwal: Can we set the `primary_portal_topic` to the single portal_topic in the database so that we don't have to determine it each time?  It'd be nice to be able to rely on one field.

higs4281: Relying on one field is the ideal we were not allowed to implement.

Setting `primary_portal_topic` on every answer would unite the fields for now but would fail the next time an editor applies just one topic, which is what we strongly recommend. Our help_text discourages editors from using the primary_portal_topic field at all, but editors would see topics routinely selected there and in the portal_topic field.

I'd prefer the logical `or` to risking admin confusion, but I'm open to another way.

richaagarwal: @higs4281 That makes sense given what they're currently instructed to do.  

I was wondering if we might be able to reverse that guidance, such that they're always filling out `primary_portal_topic` and discouraged from using the multiselect unless they absolutely have to indicate multiple topics.  In those cases, it'd be the same process as now - they have to fill out both fields.  This allows us to rely on one field for breadcrumbs purposes, and I think probably makes it clearer in the admin that we would prefer they select a single topic in general.   I understand this is probably out of scope for this particular PR but something we may want to consider.

higs4281: Reversing instructions would work if editors honor the instructions. If they don't, and just enter a single portal-topic, our routine breaks. Maybe we could make primary_portal_topic required, but we'd need a migration and perhaps some re-training.

higs4281: Looking at the options, I still prefer the current config:
- The current config succeeds even if editors don't honor the help text.
- Reversing instructions introduces a failure state If an editor chooses multiple topics but no primary.
- Reversing instructions doesn't reduce complexity. It just pushes it elsewhere, such as to the [search_index](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/ask_cfpb/search_indexes.py#L34-L35), which would now need to check two fields to make sure it gets all of an answer's selected topics, and there may be duplicates. The same would be true for any operation that needs to collect all answers for a given topic: you'd need to check two AnswerPage fields.

richaagarwal: Ah that's true - that makes sense!

PR #4992, opened 2019-05-06
Updates for Google schema pilot

virginiacc:
Google has asked for some updates to the schema pilot running on answer pages. This adds a feature flag that can be used with path conditions to wrap a page's full content in markup.

## Additions

- Adds a feature flag and conditionals in answer template to apply `HowTo` schema markup around entire content of an answer based on path
- Adds `h3` to whitelist

## Removals
- H1/H2 testing feature flag on answer page (this will be removed anyway in the open answer enhancements PR)

## Testing

1. Check out branch
2. Add a new condition for `HOW_TO_SCHEMA` flag: `path matches` `^/ask-cfpb/what-is-the-best-way-to-negotiate-a-settlement-with-a-debt-collector-en-1447/`
3. Navigate to [answer page 1447](http://localhost:8000/ask-cfpb/what-is-the-best-way-to-negotiate-a-settlement-with-a-debt-collector-en-1447/) and check its source. Verify that the answer content is wrapped in a div with attributes `itemscope="" itemtype="http://schema.org/HowTo"` and H1 has attribute `itemprop="name"`
4. Remove path condition and verify that the schema markup is no longer applied to containing div and H1 on answer page

## Screenshots
Screen Shot 2019-05-06 at 9 48 05 PM

Screen Shot 2019-05-06 at 9 22 00 AM


cfgov/cfgov/settings/base.py
```diff
@@ -700,7 +700,10 @@
     'HMDA_OUTAGE': [
         {'condition': 'parameter', 'value': 'hmda-outage', 'required': True},
         {'condition': 'path matches', 'value': r'^/data-research', 'required': True}
-    ]
+    ],
+    
+    #Add HowTo schema markup to answer page
```
richaagarwal: Minor: Could there be a space added between # and the text?

richaagarwal:
Is there a reason this uses a flag?  I'm trying to understand what the condition is but if it's always for an answer page, then isn't the flag redundant since the template is specifically for answer pages?

cfgov/cfgov/settings/base.py
virginiacc: Updated in 6748d40

virginiacc:
@richaagarwal I updated the testing instructions and added tests in 7428f3224c35582df9e8ce814d7e537520db20b2.

cfgov/cfgov/settings/base.py
```diff
@@ -700,7 +700,10 @@
     'HMDA_OUTAGE': [
         {'condition': 'parameter', 'value': 'hmda-outage', 'required': True},
         {'condition': 'path matches', 'value': r'^/data-research', 'required': True}
-    ]
+    ],
+    
+    # Add HowTo schema markup to answer page
+    'HOW_TO_SCHEMA': []
```
willbarton: Could we add a note like @schbetsy did for the [`HMDA_OUTAGE` flag](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/base.py#L696-L699) about when the flag can be deleted?

willbarton: It might also be nice to have in that comment the expectation that it's intended to use a `path matches` condition and example.

virginiacc: @willbarton Added in 79ca85862359f1f24d33455a95f47bff158b8870

willbarton: Much appreciated!

higs4281: We could provide an example:

```python
# Intended for use with path conditions in the admin, such as:
#    condition: path matches, value: ^/ask-cfpb/
```

virginiacc: Example added in 385dc07c962f61193473c77eba9304c21e330993

PR #4997, opened 2019-05-09
Run Python 2.7 and Python 3.6 containers with Docker

willbarton:
This PR adds a Python 3.6 container running cfgov-refresh. As part of this change, the `attach.sh` and `shell.sh` scripts are removed and replaced with updated Docker usage documentation that describes how to attach to or run a shell in the desired container (since there are now two of them).

This PR includes some other minor cleanup of the usage documentation, including a table of contents so that one can navigate directly to the section of the page one is interested in.

Implementationally, I have moved the Python container's `Dockerfile` and `cfgov/entrypoint.sh` to `docker/python2` and added a corresponding `docker/python3`. This seems like a better way to organize our Docker-related files and scripts. The Python containers join the Elasticsearch and docs container in `docker/`.

## Testing

1. Run `docker-compose up`
2. Verify that you can get to the Python 2 container at http://localhost:8000/ and the Python 3 container at http://localhost:8333/
3. Read the updated usage documentation at http://localhost:8888/usage/#docker

## Screenshots

```
python2_1        | Performing system checks...
python2_1        |
python3_1        | Performing system checks...
python3_1        |
python2_1        | System check identified no issues (0 silenced).
python3_1        | System check identified no issues (0 silenced).
python2_1        | May 09, 2019 - 09:53:08
python2_1        | Django version 1.11.20, using settings 'cfgov.settings.local'
python2_1        | Starting development server at http://0.0.0.0:8000/
python2_1        | Quit the server with CONTROL-C.
python3_1        | May 09, 2019 - 09:53:08
python3_1        | Django version 1.11.20, using settings 'cfgov.settings.local'
python3_1        | Starting development server at http://0.0.0.0:8000/
python3_1        | Quit the server with CONTROL-C.
```

## Notes

The name of the Python 2.7 container is changed to `python2`/`cfgov-refresh_python2_1`.

The `attach.sh` script has been apparently broken for a while, as it is trying to attach to a container name that no longer exists (at some point the Python container name changed from `cfgovrefresh_python_1` to `cfgov-refresh_python_1` (and, with this PR, to `cfgov-refresh_python2_1`/cfgov-refresh_python3_1`).

The Docker usage documentation has also been out-of-date for quite a while. I updated it and reorganized it based on common tasks I know *I* need to perform, but I'd love to have others' opinions.

## Checklist


willbarton:
Specifically tagging our known-Docker experts to review what I've done to the Dockerfiles here ;)

rosskarchner:
I don't feel too strongly about this-- but it could be a nice place to use [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) via a single Dockerfile. One way it would be a win, is right now you're basically doing this:

image

With a multi-stage build, you could do that common stuff just once, saving time and disk space:

image

The compose file wouldn't need to change much, since [you can reference individual stages](https://docs.docker.com/compose/compose-file/#target).

willbarton:
@rosskarchner interesting! I definitely think that's a better way to go — I don't really like the duplication here right now.

rosskarchner:
Or, maybe even simpler: How about a "fat" container with both versions of python, and you switch by overriding the default entrypoint or startup command?

willbarton:
@rosskarchner I want Python 2.7 and 3.6 running side-by-side though. By "switch" do you mean you'd switch which version you use on `docker-compose up`?

rosskarchner:
No, you can have two seperate containers, based on the same image. One runs the python 2 entrypoint, the other runs the python 3 entrypoint

Scotchester:
Does this open the door to somehow testing out Wagtail 2 from the same repo in a Python 3 container? Or does that get too complex with juggling requirements and such?

rosskarchner:
I think in practice, it might look something like this:

```
    python2:
        build: 
            context: ./
            dockerfile: ./docker/python2/Dockerfile
        image: python
        ports:
            - "8000:8000"
        environment:
            ES_HOST: elasticsearch
        volumes:
            - ./:/src/cfgov-refresh
            - ./develop-apps:/src/develop-apps
        entrypoint:
            - sh
            - /src/cfgov-refresh/docker/python2/entrypoint.sh
        depends_on:
            - elasticsearch
            - postgres
        working_dir: /src/cfgov-refresh
        stdin_open: true
        tty: true
    python3:
        image: python
        ports:
            - "8333:8000"
        environment:
            ES_HOST: elasticsearch
        volumes:
            - ./:/src/cfgov-refresh
            - ./develop-apps:/src/develop-apps
        entrypoint:
            - sh
            - /src/cfgov-refresh/docker/python3/entrypoint.sh
        depends_on:
            - elasticsearch
            - postgres
        working_dir: /src/cfgov-refresh
        stdin_open: true
        tty: true
```

willbarton:
@rosskarchner so the approach there would be to have a single Python `Dockerfile` that installs Python 2 and Python 3 and the requirements in each, and then separate entrypoint scripts (or a switch in the entrypoint script)?

willbarton:
> Does this open the door to somehow testing out Wagtail 2 from the same repo in a Python 3 container? Or does that get too complex with juggling requirements and such?

@Scotchester I don't think so — there's very little reason to want to run Wagtail 2 along side Wagtail 1.13, and in practice it would introduce a *lot* of complexity in the codebase to support running them side-by-side. I think we want a clean upgrade from 1.13 to 2.x.

rosskarchner:
> @rosskarchner so the approach there would be to have a single Python `Dockerfile` that installs Python 2 and Python 3 and the requirements in each, and then separate entrypoint scripts (or a switch in the entrypoint script)?

Yeah, though I like it a bit less now that you've pointed out that requirements need to be installed twice (though I guess that happens either way)

docs/usage.md
```diff
@@ -128,133 +146,119 @@ yarn run gulp test:acceptance # Run only acceptance (in-browser) tests on produc
...
-`eval $(docker-machine env)`
+### Environment variables
 
-It may be helpful to run `docker-machine env` by itself, so you understand
-what's happening. Those variables are what allows `docker-compose` and the
-`docker` command line tool, running natively on your Mac, to connect to the
-Docker server running inside VirtualBox.
+Environment variables from your `.env` file are sourced 
+when the python containers starts 
+and when you [access the running python containers](#access-the-containers-shell).
```
Scotchester: Has that always happened?! Or is that a new feature being added by the `RUN echo 'source /src/cfgov-refresh/.env' >> ~/.bashrc` line you put in the `Dockerfile`?

```suggestion
and when you [access the running Python containers](#access-the-containers-shell). 
```

docs/usage.md
```diff
@@ -128,133 +146,119 @@ yarn run gulp test:acceptance # Run only acceptance (in-browser) tests on produc
...
 
-`eval $(docker-machine env)`
+### Environment variables
 
-It may be helpful to run `docker-machine env` by itself, so you understand
-what's happening. Those variables are what allows `docker-compose` and the
-`docker` command line tool, running natively on your Mac, to connect to the
-Docker server running inside VirtualBox.
+Environment variables from your `.env` file are sourced 
+when the python containers starts
```
Scotchester:
```suggestion
when the Python containers start
```

docs/usage.md
```diff
@@ -128,133 +146,119 @@ yarn run gulp test:acceptance # Run only acceptance (in-browser) tests on produc
...
-what's happening. Those variables are what allows `docker-compose` and the
-`docker` command line tool, running natively on your Mac, to connect to the
-Docker server running inside VirtualBox.
+Environment variables from your `.env` file are sourced 
+when the python containers starts 
+and when you [access the running python containers](#access-the-containers-shell). 
+Your local shell environment variables, however, 
+are not visible to applications running in Docker.
+To add new environment variables, simply add them to the `.env` file, 
+stop compose with ctrl-c,
```
Scotchester:
```suggestion
stop docker-compose with Ctrl+C, 
```

docs/usage.md
```diff
@@ -128,133 +146,119 @@ yarn run gulp test:acceptance # Run only acceptance (in-browser) tests on produc
...
-`docker` command line tool, running natively on your Mac, to connect to the
-Docker server running inside VirtualBox.
+Environment variables from your `.env` file are sourced 
+when the python containers starts 
+and when you [access the running python containers](#access-the-containers-shell). 
+Your local shell environment variables, however, 
+are not visible to applications running in Docker.
+To add new environment variables, simply add them to the `.env` file, 
+stop compose with ctrl-c, 
+and start it again with `docker-compose up`.
```
Scotchester: Could you leave the containers running and just access the shell? Or would the env vars not propagate to the running server processes?

docs/usage.md
```diff
@@ -128,133 +146,119 @@ yarn run gulp test:acceptance # Run only acceptance (in-browser) tests on produc
...
 
-If you use autoenv (described in the stand-alone intructions) or something
-similar, you might consider adding `eval $(docker-machine env)` to your .env
-file. You could also achieve the same results (and start the VM if it's not
-running yet) with `source mac-virtualbox-init.sh`
+### Access the containers' shell
```
Scotchester:
```suggestion
### Access the containers' shells
```

docs/usage.md
```diff
@@ -128,133 +146,119 @@ yarn run gulp test:acceptance # Run only acceptance (in-browser) tests on produc
...
-Any further Docker documentation will assume you are either in a shell where
-you have already run `eval $(docker-machine env)`, or you are in an environment
-where that's not neccessary.
+- Python 2.7: `docker exec -it cfgov-refresh_python2_1 bash`
+- Python 3.6: `docker exec -it cfgov-refresh_python3_1 bash`
+- Elasticsearch: `docker exec -it cfgov-refresh_elasticsearch_1 bash`
+- PostgreSQL: `docker exec -it cfgov-refresh_postgres_1 bash`
 
-#### Run manage.py commands like migrate, shell, and dbshell, and shell scripts like refresh-data.sh
+### Run management commands
```
Scotchester:
```suggestion
### Run Django management commands
```

docs/usage.md
```diff
@@ -128,133 +146,119 @@ yarn run gulp test:acceptance # Run only acceptance (in-browser) tests on produc
...
-If Compose is running, stop it with CTRL-C. Run:
+The next time you run `docker-compose up` the new requirements will be in place.
 
-`docker-compose build python`
+### Work on satellite apps
 
-This will update your Python image. The next time you run `docker-compose up`,
-the new requirements will be in place.
+If you need to work on any of our 
+[satellite apps](https://github.com/cfpb/cfgov-refresh/blob/master/requirements/optional-public.txt),
```
Scotchester:
```suggestion
[satellite apps](/satellite-repos/), 
```

docs/usage.md
```diff
@@ -128,133 +146,119 @@ yarn run gulp test:acceptance # Run only acceptance (in-browser) tests on produc
...
+cd develop-apps
+git clone https://github.com/cfpb/college-costs
 
-- the first time you run `docker-compose up` (or the first time after you
-re-create the Docker Machine VM)
-- any time you run `docker-compose build`
+# Build the front-end:
+cd college-costs
+./setup.sh
+```
```
Scotchester: What about installing their requirements?

docs/usage.md
```diff
@@ -128,133 +146,119 @@ yarn run gulp test:acceptance # Run only acceptance (in-browser) tests on produc
...
-#### Develop satellite apps
+- Python 2.7: `docker attach cfgov-refresh_python2_1`
+- Python 3.6: `docker attach cfgov-refresh_python3_1`
 
-Check out any apps you are developing into the develop-apps directory. These
-will automatically be added to the
-[PYTHONPATH](https://docs.python.org/2/using/cmdline.html#envvar-PYTHONPATH),
-and apps contained within will be importable from Python running in the
-container.
+When you're done, you can detach with `CTRL-p CTRL-q`.
```
Scotchester:
```suggestion
When you're done, you can detach with `Ctrl+P Ctrl+Q`.
```

docs/usage.md
```diff
@@ -128,133 +146,119 @@ yarn run gulp test:acceptance # Run only acceptance (in-browser) tests on produc
...
 
-For example, if your app is called 'foobar', in a repo called foobar-project,
-you could clone foobar-project in to develop apps:
 
-`git clone https://github.com/myorg/foobar-project`
+#### Useful Docker commands
```
Scotchester:
```suggestion
### Useful Docker commands
```

docs/usage.md
willbarton: It has always happened, but instead of that echo in the `Dockerfile`, it was happening in the `shell.sh` every time it was invoked.

docs/usage.md
willbarton: Yeah... it just depends on where you want the environment variables. You have to restart the container to update them for the running Django. But if you just need them in the shell, you don't have to. 

I don't think that distinction is worth calling out in the docs though – the `.env` vars exist to be consumed by Django (except for `CFGOV_PROD_DB_LOCATION`), so I think it's a reasonable assumption that anyone changing them would want them changed for the running Django process.

docs/usage.md
willbarton: Good point — that has to happen in the container shell. I'll add that.

docs/usage.md
willbarton: The capitalization was taken directly from the Docker documentation 😛 .

Scotchester: I saw that, but what are they smoking? 😝 

I did wonder if it implied a Shift that could cause problems, but I feel like most people would ignore the case of the final letter and only add Shift if it's explicitly written. cf. the shortcut keys shown in macOS menus, for example.

willbarton:
@rosskarchner I think I've got both containers using the same image, per your suggestion above.

docs/usage.md
willbarton: I don't think this is correct — It's about accessing a specific container's shell rather than accessing multiple shells at the same time.

Would "Access a container's shell" be a better change?

docs/usage.md
willbarton: Added in 0ea058b.

docs/usage.md
```diff
@@ -128,133 +146,119 @@ yarn run gulp test:acceptance # Run only acceptance (in-browser) tests on produc
...
-Python container, which you can do with `./shell.sh`
-- There is not *yet* a good way to use SSL/HTTPS, but that is in the works
-- You won't ever need to use `backend.sh` or `runserver.sh`
+All of these containers are configured in our 
+[`docker-compose.yml` file](https://github.com/cfpb/cfgov-refresh/blob/master/docker-compose.yml). 
+See the [Docker documentation](https://docs.docker.com/compose/compose-file/) 
+for more about the format and use of this file.
 
-### How do I...
+The following URLs are mapped to your host from the containers:
```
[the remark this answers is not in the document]
willbarton: More explicit in 0ea058b.

docs/usage.md
willbarton: Used the title "Access a container's shell" in 0ea058b.

willbarton:
I'll squash the doc updates once this PR is approved before it merges.

docs/usage.md
Scotchester: Sure, that works, too.

willbarton:
One issue I notice that I'm not sure how to get around is that `cfgov/manage.py` uses `#!/usr/bin/env python` to run. This means that opening a shell in the `python3` container and running management commands will still use Python 2 (because it's the default in the CentOS 7 image). I'm not entirely sure how to get around that (without using something like pyenv instead of the CentOS python packages).

rosskarchner:
this seems to work:
try adding `ln -s /bin/python3 /usr/local/bin/python ` to the python3 entrypoint

willbarton:
That works!

PR #4998, opened 2019-05-09
Add "Events" category label in Recent Updates section of homepage

schbetsy:
In the "Recent Updates" section of the homepage, posts that are Events weren't getting the correct header and icon above them.

## Additions

- A method, used in the index template, to check whether a "recent update" post is an event

## Testing

1. Running this branch, enter the Wagtail admin and create a new `EventPage` under Events
2. Visit the http://localhost:8000.
3. The event you created should show up in the "Recent Updates" section, and it should be visually consistent with the other posts in that section, including a header that says "Events", with a calendar icon.

## Screenshots
Screen Shot 2019-05-09 at 11 07 27 AM

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/index.html
+++ b/cfgov/jinja2/v1/index.html
@@ -197,6 +197,8 @@
                                 <h3 class="h4">
                                     {% if is_blog(post) %}
                                         {{ category_slug.render(category='blog', href=page_url) }}
+                                    {% elif is_event(post) %}
+                                        {{ category_slug.render(category='Events', href=page_url) }}
                                     {% elif is_report(post) %}
                                         {{ category_slug.render(category='report', href=page_url) }}
                                     {% else %}
--- a/cfgov/v1/util/ref.py
+++ b/cfgov/v1/util/ref.py
@@ -262,6 +262,11 @@ def is_blog(page):
         return True
 
 
+def is_event(page):
+    if 'Event' in page.specific_class.__name__:
+        return True
+
+
 def is_report(page):
     for category in page.categories.all():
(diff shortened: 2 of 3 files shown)


cfgov/jinja2/v1/index.html
```diff
@@ -197,6 +197,8 @@ <h2>Recent updates</h2>
                                 <h3 class="h4">
                                     {% if is_blog(post) %}
                                         {{ category_slug.render(category='blog', href=page_url) }}
+                                    {% elif is_event(post) %}
```
schbetsy: This seemed like the least invasive way to make the change we wanted. I didn't want to update any category lists, since I wasn't sure how that would affect functionality elsewhere. But I'm open to other ways to accomplish this if anyone sees a nicer solution.

cfgov/v1/util/ref.py
```diff
@@ -262,6 +262,11 @@ def is_blog(page):
         return True
 
 
+def is_event(page):
```
schbetsy: This file is doing a lot, and the way categories, slugs, and page types interact is pretty confusing. I didn't think this task should include any refactoring, but at some point it would be nice to simplify it, if possible.

willbarton: Yeah — refactoring that file comes up every time we have to touch it, but we never do ☹️ .

cfgov/jinja2/v1/index.html
willbarton: I think this makes sense, given that `is_blog(post)` is already a pattern here.

PR #4999, opened 2019-05-10
Add schema blocks to answer page

virginiacc:
Adds blocks to answer page streamfield to support pilot test of Google FAQ and HowTo schema.

## Additions
- FAQ and HowTo schema blocks and templates

## Changes
- Wrap answer page content blocks in `.row` container divs
- Apply spacing between `.row` divs

## Testing

1. Check out branch
2. Run `gulp styles` and `python cfgov/manage.py migrate`
3. Test FAQ schema:
    - open [an answer page](http://localhost:8000/admin/pages/5764/edit/) in Wagtail admin

    - Add a FAQ schema block in the answer content streamfield:
      Screen Shot 2019-05-09 at 6 09 35 PM

    - Verify that the output is wrapped in schema markup
      Screen Shot 2019-05-09 at 6 12 07 PM

    - You can also verify the schema by pasting the FAQ HTML into the [Google structured data testing tool](https://search.google.com/structured-data/testing-tool)
      Screen Shot 2019-05-09 at 6 11 35 PM

4. Test HowTo schema
    - open [an answer page](http://localhost:8000/admin/pages/10386/edit/) in Wagtail admin

    - Add a How To schema block in the answer content streamfield:
      Screen Shot 2019-05-09 at 6 52 05 PM

    - Verify that the output is wrapped in schema markup
      Screen Shot 2019-05-09 at 6 52 46 PM
    
    - The HowTo schema also requires some page-level markup that is applied with a feature flag. To verify the complete markup structure in the [Google structured data testing tool](https://search.google.com/structured-data/testing-tool), add a new `HOW_TO_SCHEMA` flag condition for the path of the page you've edited (ie, `^/ask-cfpb/what-is-a-529-savings-plan-en-2074/`). 
      
      Screen Shot 2019-05-10 at 4 42 46 PM

      Load the answer page, copy its source, and paste it in the tool. You should see something like this:

      Screen Shot 2019-05-10 at 3 50 35 PM

      The top-level `name` value is the page's question.

## Notes
I initially thought about creating a combined `schema` block with a type selector, but the markup for FAQ and HowTo diverged enough in the latest recommendations from Google that a single template for both would be complicated. I think keeping them separate might be more future-proof (and a bit more comprehensible for users, since we can use schema-specific labels like `Steps` vs `Questions`), but I'm also open to looking at the single block approach again if anyone thinks that would be preferable.


virginiacc:
@higginsw Thanks for reviewing and catching the HowTo issue! I forgot to add that the HowTo schema needs the page-level markup applied by the `HOW_TO_SCHEMA` flag. I updated the testing instructions to include that step.

PR #5000, opened 2019-05-10
Add a new snippet model for glossary terms

richaagarwal:
## Overview

Adds a new model GlossaryTerm for storing "key terms" as snippets. The scripts create glossary terms from parsing existing key terms pages, such as https://www.consumerfinance.gov/consumer-tools/student-loans/key-terms/, or from existing Answer pages tagged as "Key terms".   The CSV is used to map term names to answer IDs in the case of populating from Answer pages, but also to track Spanish translations (in both cases).

I've split out the scripts into English and Spanish since populating those require slightly different sources of information. There are some manual updates that we'll make to the terms after this is deployed to take care of some mark-up that wasn't easy to migrate programmatically.  For instance, the `$$$$$$$$$$` strings in the CSV were added by @sonnakim and will be dealt with before any of this becomes consumer-facing.

More context is available in GHE/CFGOV/Ask-Evo/issues/102

## Next steps 
Once this PR is deployed and the scripts `migrate_glossary_terms` and `migrate_glossary_terms_spanish` are run, the terms will exist under /admin/v1/glossaryterm/. This does not contain any consumer-facing changes.

A subsequent PR will contain updates to the template and view for the PortalSearchPage. This will populate & replace the "key terms" section on pages like https://www.consumerfinance.gov/consumer-tools/student-loans/answers/ and https://www.consumerfinance.gov/es/obtener-respuestas/prestamos-para-vehiculos/respuestas/palabras-claves/.

## Screenshots
screencapture-localhost-8000-admin-v1-glossaryterm-edit-167-2019-05-13-17_59_52

## Testing
To run this locally, pull down this branch and a recent database dump.  Run migrations, and then run:

```
python cfgov/manage.py runscript migrate_glossary_terms
python cfgov/manage.py runscript migrate_glossary_terms_spanish
```

Then navigate to http://localhost:8000/admin/v1/glossaryterm/


cfgov/v1/models/snippets.py
```diff
@@ -79,3 +81,83 @@ class RelatedResource(index.Indexed, models.Model):
...
+        blank=True
+    )
+    answer_page_es = models.ForeignKey(
+        'ask_cfpb.AnswerPage',
+        related_name='glossary_terms_es',
+        null=True,
+        blank=True
+    )
+    portal_topic = ParentalKey(
+        'v1.PortalTopic',
```
virginiacc: Do we need to allow for the possibility that a glossary term could be associated with multiple portal topics?

richaagarwal: @virginiacc I don't think so -- there's a few glossary term names like "Amortization" that appear in multiple topics, and they have their own definitions etc.  If they really want to duplicate content across portals I think it might just be easiest to create a new snippet and copy and paste but I'd say we shouldn't design for it. 

@sonnakim any thoughts?

cfarm:
:tada::tada::tada::tada::tada::tada::tada::tada::tada:5️⃣ 0️⃣ 0️⃣ 0️⃣ :tada::tada::tada::tada::tada::tada::tada::tada:

cfgov/ask_cfpb/fixtures/glossary_terms.csv
```diff
@@ -0,0 +1 @@
...
Student loans,Public Service Loan Forgiveness,641,Perdón de préstamo por servicio público
Student loans,Servicemembers Civil Relief Act (SCRA),,
Student loans,Stafford loan,549,Préstamos Stafford
Student loans,Standard repayment,,
Student loans,Student loan servicer,583,Administrador de préstamos estudiantiles
Student loans,Subsidized loan,573,Préstamo subsidiado
Student loans,Total loan balance,,
Student loans,Tuition repayment plan,563,Plan de pago de matrícula
Student loans,Unsubsidized loan,,
Student loans,Work study,,
```
higs4281: I wonder if it would be friendlier to include the header in the CSV rather than adding it later.
Also breaking each line would make the CSV reviewable, especially if there are any changes.

cfgov/v1/models/snippets.py
virginiacc: Ok, that makes sense. We can always revisit if there's a need for multiple topics down the road.

cfgov/ask_cfpb/fixtures/glossary_terms.csv
richaagarwal: Thanks for the help on this, @higs4281! Fixed in the latest.

cfgov/v1/models/snippets.py
```diff
@@ -79,3 +81,83 @@ class RelatedResource(index.Indexed, models.Model):
 
     def __str__(self):
         return self.title
+
+
+@python_2_unicode_compatible
+@register_snippet
+class GlossaryTerm(index.Indexed, models.Model):
+    term_en = models.CharField(
+        max_length=255,
```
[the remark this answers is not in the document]
richaagarwal: After discussion with @virginiacc, we've decided to keep it the way it is as it's meant to be a single word and a text area is not ideal.

cfgov/ask_cfpb/scripts/migrate_glossary_terms.py
```diff
@@ -0,0 +1,92 @@
...
+    Parses the terms & definitions on a given key terms page,
+    and creates new glossary term snippets out of them.
+    """
+    for block in page.specific.content:
+        if block.block_type != 'full_width_text':
+            continue
+        for child in block.value:
+            if child.block_type == 'content':
+                contents = bs(child.value.source, 'html.parser')
+                headings = contents.findAll('h3')
```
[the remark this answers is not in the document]
richaagarwal: Fixed in the latest commit, thanks for catching this!

PR #5005, opened 2019-05-13
Add language fields to RelatedResource snippet.

higs4281:
To facilitate unification of English and Spanish content-handling, we need
RelatedResource snippets to be able to provide content in either language.

Here we add `title_es` and `text_es` fields to the model, plus methods for
translated fields, allowing unified templates to call
`{{ page.related_resource.trans_title(page.language) }}` or
`{{ page.related_resource.trans_text(page.language) }}`
to pick up related-resource content in the correct language, without need
of separate template code.

Existing English templates that call for `title` or `text` on a related resource will not be affected.

EVO issue 118 also requests the consolidation of complaint related-resources on answer pages.  
The `use_generic_complaint_resource.py` script addresses that request.

## Testing
The `use_generic` script can be tested locally by pulling in a recent db copy and running these commands:

```bash
./cfgov/manage.py migrate
./cfgov/manage.py runscript use_generic_complaint_resource
```
The script should take about 3 minutes to set the generic resource on 1,334 pages.

Call up a live English page, and it should show the generic complaint related-resource.
<http://localhost:8000/ask-cfpb/what-does-it-mean-to-renew-or-roll-over-a-payday-loan-en-1573/>

generic_complaint_resource

The generic resources are also applied to related Spanish pages, but there is no text translation yet for the generic snippet, so the related resource should not render. Per the issue's requirements, when a translation for the title and text are provided, the resource will show up on Spanish pages as well.

## Checking counts
The complaint resource snippets are only used on Ask CFPB answer pages.  
Before running the script, related-resource usage counts should look like this:

Related resource | answer-page count
:--------------- | ----------------:
Credit reporting complaint | 188
Payday loan complaint | 72
Money transfers complaint | 56
Debt collection complaint | 170
Auto loan complaint | 162
Submit a bank account complaint | 260
Vehicle or consumer loan complaint | 0
Student loan complaint | 146
Mortgage complaint | 186
Submit a credit card complaint | 94
Submit a complaint | 22

After running the script, usage counts should be like this:

Related resource | answer-page count
:--------------- | ----------------:
Credit reporting complaint | 0
Payday loan complaint | 0
Money transfers complaint | 0
Debt collection complaint | 0
Auto loan complaint | 0
Submit a bank account complaint | 0
Vehicle or consumer loan complaint | 0
Student loan complaint | 0
Mortgage complaint | 0
Submit a credit card complaint | 0
Submit a complaint | 1356

The unwanted snippets can be deleted after this is deployed.

Counts can be checked this way:

```python
from v1.models.snippets import RelatedResource

complaint_pks = [1, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24]

for resource in RelatedResource.objects.filter(pk__in=complaint_pks):
    resource, resource.answerpage_set.count()
```

In the admin, you can check snippet usage by plugging a snippet pk  
into this URL form: `/admin/snippets/v1/relatedresource/[PK]/usage/`


cfgov/v1/models/snippets.py
```diff
@@ -69,16 +69,28 @@ def __str__(self):
 @python_2_unicode_compatible
 @register_snippet
 class RelatedResource(index.Indexed, models.Model):
-    title = models.CharField(
-        max_length=255
-    )
-    text = RichTextField()
+    title = models.CharField(max_length=255)
```
richaagarwal: What do you think of renaming these fields to `title_en` etc, so that `trans_title()` below can just be `title()` and `text()` etc ?  I think Django handles it as a renaming and not a new/ deleted field, so it shouldn't cause any migration issues.

higs4281: I was trying to avoid breaking any existing template usages, which would not pass a language value. But on second thought I think a bare template call for .title or .text without passing a language should return the English version if the default value is 'en'. 
If this tests out, I'll swap.

cfgov/jinja2/v1/ask-cfpb/answer-page.html
```diff
@@ -72,14 +72,14 @@ <h1 {% if flag_enabled('HOW_TO_SCHEMA', request) %}itemprop="name"{% endif %}>
                     </span>
                 </button>
             </div>
-        {% if page.related_resource and page.language != 'es' %}
+        {% if page.related_resource and page.related_resource.trans_text(page.language) | striptags %}
```
virginiacc: @higginsw I noticed that if a value hasn't been entered for a resource's Spanish text field, the "Next steps" section is output with `None` values on a Spanish page with that resource. Adding an `is not none` condition for the `trans_text` value seems to fix it.

cfgov/v1/models/snippets.py
richaagarwal: @higs4281 I can't find any other references to related_resource on a template other than the answer-page template, but if there are or there's a concern for future templates that don't have access to the language, and `.title` doesn't work, `.title()` certainly should.

higs4281: A bare template call does not work, so using title() and text() would break any current template (or other ) use. It does appears that our answerpage is the only template use, but in troublehsooting today with @nataliafitzgerald we discovered snippet usage reporting can be unreliable. 

There is at least one other use besides in templates: the export_ask_data.py script, which exports related-resource values and would break.

I don't think it's worth chasing this stuff or risking breaking other parts of the site for the sake of a translated field name that will, as you note, probably only be used in the answerpage template. I'd vote for backward compatibility and a `trans_` value name.

cfgov/jinja2/v1/ask-cfpb/answer-page.html
higs4281: Thank you for catching this @virginiacc 
I solved it in the snippet trans method.

cfgov/v1/models/snippets.py
richaagarwal: @higs4281 That makes sense to me.  If I have time after this project, I'd like to go through all the places we have translated fields and try to standardize them, so maybe we can eventually agree on an approach, but this seems great for now.

cfgov/jinja2/v1/ask-cfpb/answer-page.html
virginiacc: Works perfectly!

cfgov/v1/models/snippets.py
higs4281: I totally agree one approach is ideal. 

I was down with the simplicity of using the main field names as the translated values, but I'm liking the explicit nature of `trans_` and how it mirrors Django template language.

PR #5006, opened 2019-05-13
Add volume to postgres container to persist refreshed data

wpears:
Currently, the data loaded by `./refresh-data.sh` doesn't persist across container shutdown (`docker-compose down`). This adds a volume to fix that. Now data will persist across both container shutdown and docker shutdown.

Data is updated as normal, running `./refresh-data.sh` again will freshen the data and the new data will be persisted.

If for some reason this data needs to be wiped directly, you can just `docker-compose down && docker volume remove postgres_data` and an empty volume will be newly created on your next `docker-compose up`, but I doubt this will be a common workflow.


willbarton:
I have no problems with this PR, but I don’t think I’ve encountered this as a problem — my changes to the database (with refresh-data.sh and changes made by hand) seem to persist?

PR #5010, opened 2019-05-15
Remove AnswerCategoryPage and clean up Category/Subcategory references

richaagarwal:
As a follow-up to the recent work we've done on replacing AnswerCategoryPages with Portal Search "See All" Pages, as well as redirecting those URLs, we can now officially remove AnswerCategoryPages from our codebase. 

I've also cleaned up category and subcategory references where possible, although it looks like a category is still used on an AnswerPage to populate its `meta_image`.   Further clean-up can be done if those images are ever migrated off a category.  For now it seems harmless to leave it there.

I've also removed all the scripts in `/ask_cfpb/scripts/deprecated` since many of those have references to AnswerCategoryPages and it felt confusing to keep them around if they wouldn't work. They can still be found/referenced in PRs and our git history.

The AnswerPage template referenced the category page URLs (as the "Learn more about ... " links), so I updated those to use the same logic as breadcrumbs (which involved refactoring that a bit) - link to the portal page if it exists, otherwise link to the portal search page.

This work is being tracked in GHE/CFGOV/Ask-Evo/issues/114


cfgov/ask_cfpb/models/pages.py
```diff
@@ -113,51 +103,23 @@ def get_ask_breadcrumbs(
...
-            crumbs[0].update({
-                'title': title,
-                'href': crumbs[0]['href'].format(slug)
-            })
-        else:
-            portal_search_page = portal_topic.portal_search_pages.get(
-                language=language)
-            title = portal_search_page.title
+        title = page.title
+        href = crumbs[0]['href'].format(portal_slug)
```
virginiacc: I think this is an existing issue, but it looks like generating the href with a slugified version of the portal topic title doesn't work in all cases -- the slug for the 'Fraud and scams' portal page is `fraud`, not `fraud-and-scams`, so the breadcrumb link on a [fraud and scams](http://localhost:8000/ask-cfpb/how-do-i-dispute-a-charge-on-my-credit-card-bill-en-61/) answer page 404s.

richaagarwal: Fixed in b39c9d6c34306a77181591e0afc1e6d400e57140, thanks for catching this!

PR #5015, opened 2019-05-16
Spanish home page

virginiacc:
Adds the ability to create a Spanish home page based on cf.gov's English home page.

## Additions

- Wagtail page type for `Spanish home page`
- `index_es.html` template, based on `index.html`, with translations

## Testing

1. Check out branch

2. Run `python cfgov/manage.py migrate`

3. Under CFGov, create a new `Spanish home page` and set its language to Spanish. It should look like this:

    home-page-es

4. Test adding featured content using the half width link blob option in the `Header` section of the admin. The content should display in the well below the hero:

    home-page-es-featured


cfgov/jinja2/v1/index_es.html
```diff
@@ -0,0 +1,183 @@
...
+          media="(max-width: 600px)">
+    <link rel="preload"
+          href="{{ static('img/home-hero_lg.jpg') }}"
+          as="image"
+          media="(min-width: 601px)">
+{%- endblock %}
+
+
+{% block desc -%}
+    Our vision is a consumer finance marketplace that works for American consumers,
```
richaagarwal: Does this need to be translated?

virginiacc: @richaagarwal Yes, that's a good point. I don't think we have that translated yet, so I removed this override. The main desc block defaults to this text but will show anything entered in the description field in the page's admin, so we can start by entering some custom text there. Once we have a translation, we can set it up for the base template and remove the custom text.

higs4281: Just curious: Isn't this only visible if you view source and find the meta description? 
In general, do we need to worry about translating meta values? I'm sure there are people who would appreciate it, but it seems low on the priority scale for translating our site.

virginiacc: @higs4281 I think the meta description can appear in search results, so it's not an immediate priority but we might want to document it as a possible future translation.

PR #5019, opened 2019-05-17
Add translations to organisms/email-signup

higs4281:
To use the email signup block on Spanish pages, we just need to
add translation wrappers around some of the labeling, and
add translation pairs to `django.po`.

The email module has been pulled off most Ask pages, but it's on the formerly hard-coded Comprar casa page and on all the Spanish blog pages and will get translated there.

There will still be intro phrases to translate in places and when new intros are written, but this is a start.


cfgov/locale/es/LC_MESSAGES/django.po
```diff
@@ -31,6 +31,18 @@ msgstr "Hipoteca"
...
 msgstr "Hipotecas"
 
+msgid "Buying a Home?"
+msgstr "¿Quiere comprar una vivienda?"
+
+msgid "Buying a House?"
+msgstr "¿Quiere comprar una vivienda?"
+
+msgid "Sign up for email tips and info to help you through the process."
+msgstr "Suscríbase a nuestra lista de email con consejos útiles sobre hipotecas."
```
virginiacc: Do we need to include email heading text here, or would it be entered in the signup module?

higs4281: Not strictly needed, because you'd expect Spanish content to be entered on Spanish pages. 
I included it as a fallback because I had seen some English intros used (improperly) on Spanish pages and wanted to test the translation.

PR #5029, opened 2019-05-21
Make the hard-coded sequence for portal category navigation explicit

higs4281:
The `PORTAL_CATEGORY_SORT_ORDER` var in ask_cfpb/models/pages.py was a list
of portal category primary keys. This makes the variable an ordered dict
to give context to the keys.


higs4281:
covered by https://github.com/cfpb/cfgov-refresh/pull/5056

PR #5048, opened 2019-05-30
Fix indexing for Ask CFPB answers

higs4281:
Indexing broke when we moved answer content into streamfields.
This adjusts the indexing routine to extract content out of the streamfield blocks.

## Testing

With a recent db load (and a running elasticsearch instance), pull in
the branch and update the elasticsearch index:
```bash
./cfgov/manage.py update_index ask_cfpb --remove
```

The index run should complete, and you should get results on local search pages: 
- <http://localhost:8000/ask-cfpb/search/?q=negotiate>
- <http://localhost:8000/es/obtener-respuestas/buscar/?q=credito+meses>

### DIFF
--- a/cfgov/ask_cfpb/search_indexes.py
+++ b/cfgov/ask_cfpb/search_indexes.py
@@ -1,4 +1,9 @@
+from __future__ import unicode_literals
+
+from django.utils.html import strip_tags
 from haystack import indexes
+
+import unidecode
 
 from ask_cfpb.models.pages import AnswerPage
 from search import fields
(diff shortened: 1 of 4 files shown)


cfgov/ask_cfpb/search_indexes.py
```diff
@@ -37,9 +36,42 @@ def prepare_portal_topics(self, obj):
...
+        stripped = strip_tags(raw_text)
+        try:
+            text = unidecode.unidecode(stripped)
+        except Exception:
+            pass
+        return text
+
     def prepare(self, obj):
         data = super(AnswerPageIndex, self).prepare(obj)
-        data['suggestions'] = data['text']
```
richaagarwal: What is `data['text']` currently?

higs4281: The "text" field of the search index, which is the bucket of indexed content.

higs4281: Both ways work, so I reverted to data['text'].

richaagarwal:
Thanks for taking care of this, @higs4281!  I'm wondering if (for a future PR), we should look into adding tests for this somehow, though not exactly sure how we'd set that up.

PR #4982, opened 2019-05-02
Ask answer streamfield

virginiacc:
Moves answer page content into streamfield blocks and makes other updates to the answer page layout.

## Additions

- Migration to move answer content into streamfield
- Blocks and templates for tips and answer content

## Removals
- Remove separate `search-bar.html` organism since it was only being used for Ask

## Changes
- Update answer page layout
- Move content into streamfield blocks

## Testing

1. Check out branch
2. Run migration & `gulp styles`
3. Navigate to an English and Spanish answer page. Ensure its content still displays correctly, the search bar works, and the style/layout matches screenshot.
4. Add a tip to the content of an answer page and verify that it displays as in design.

## Screenshots

answer-page

## Design
22fa5280-45ad-11e9-8010-e61f48f7d896

## Checklist

- [ ] PR has an informative and human-readable title
- [ ] Changes are limited to a single goal (no scope creep)
- [ ] Code can be automatically merged (no conflicts)
- [ ] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [ ] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated
- [ ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Internet Explorer 9, 10, and 11
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS
- [ ] Flexible from small to large screens
- [ ] No linting errors or warnings
- [ ] JavaScript tests are passing

PR #5050, opened 2019-05-31
Fix Ask CFPB export script to work with new answer_content streamfield

richaagarwal:
## Overview
In https://github.com/cfpb/cfgov-refresh/pull/4982, we introduced a new StreamField `answer_content`, to replace the RichTextField `answer` on an `AnswerPage`.  In https://github.com/cfpb/cfgov-refresh/pull/5024 we did some clean-up and removed the deprecated `answer` field.  However, the changes I made to the export script in that PR introduced a bug, since I wasn't parsing the streamfield properly.  This PR fixes that.

## Notes
There's not a great way to convert streamfields with many different possible blocks / fields into an easy-to-read cell in an Excel spreadsheet, so I decided to only grab the `text` block.  However, we have a few AnswerPages with FAQ and HowTo blocks that don't define `text`, so in those cases I needed to grab the `description` from those blocks.  

This might be a bit brittle as those blocks evolve or the streamfield is used in different ways, so I'd personally lean towards not including the answer content in the export at all - but for now this feels like a reasonable fix.

### DIFF
--- a/cfgov/ask_cfpb/scripts/export_ask_data.py
+++ b/cfgov/ask_cfpb/scripts/export_ask_data.py
@@ -62,8 +62,20 @@ def assemble_output():
         output['Language'] = page['language']
         output['RelatedResource'] = page['related_resource__title']
         output['Question'] = page['question'].replace('\x81', '')
-        output['Answer'] = clean_and_strip(
-            page['answer_content']).replace('\x81', '')
+        answer_streamfield = page['answer_content'].stream_data
+        answer_text = filter(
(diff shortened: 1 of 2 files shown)


richaagarwal:
It looks like the back-end tests are failing on Travis for python3.  If someone else wants to take a peek I probably won't have time to come back to this until Monday.

higs4281:
> It looks like the back-end tests are failing on Travis for python3. If someone else wants to take a peek I probably won't have time to come back to this until Monday.

There was a filter iterator that needed to be made a list. Fixed.

PR #5051, opened 2019-05-31
WIP: Multistage Dockerfile, towards a production-ready image

### LINKED ISSUE #4468: Incorporate front-end build tools into the docker image

rosskarchner:
A proof-of-concept for producing a complete, deployable docker image of consumerfinance.gov
Closes #4468

## Additions

- The Dockerfile now uses [multiple stages](https://docs.docker.com/develop/develop-images/multistage-build/) to (optionally) build a complete, theoretically deployable docker image.
- docker-compose.prodlike.yml: provides a way to run the deployable image, alongside our usual Postgres and Elasticsearch images.
- docker-compose.override.yml: necessary because of [some nuances](https://docs.docker.com/compose/extends/#adding-and-overriding-configuration) around how multiple compose files work-- it will automatically get included when you do the usual `docker-compose up`, but won't interfere when you use the 'prodlike' compose file above.

image

## Changes

- unifies the python2 and python3 entrypoint scripts
- uses mod_wsgi-express (via the 'runmodwsgi' manage.py command) everywhere, so even in development you're running under Apache.

## Testing

The developer experience of using `docker compose up` should be relatively unchanged. It's worth testing that assumption. 

**The following expects a downloaded zip of our internal cfgov-fonts repo, in the same directory as the Dockerfile. It must be called "cfgov-fonts-master.zip"**.  The Download Zip button in github gives you the right thing.

If you want to launch the "prodlike" stack, use:
`docker-compose -f docker-compose.yml -f docker-compose.prodlike.yml up`

This will take a while, but should result in an image that contains just the dependencies necessary to run the app, as well as built static files. Like the regular docker-compose setup, it will also spin up elasticsearch and postgres.  requirements/optional-public.txt  WILL be installed for python2. It's not yet working with python3.

Unlike the regular docker-compose setup, the 'cfgov-refresh' directory is NOT being shared between your mac and Docker, and there is no automatic sharing of environment variables or .env files. The image is self-contained. If you want to pass in environment variables, you must do so through the mechanisms provided by Docker.

Outside of compose, if you want to play with the image, you can do something like:

`docker build . -t cfgov-deployable` to build the final image.

If you want to poke around with one of the intermediate stages, you can pass in the `--target` argument to `docker build`, for example:

`docker build . --target cfgov-build -t cfgov-build` 

These commands create images, you will need [docker run](https://docs.docker.com/engine/reference/run/) to make them go. That might look like:

`docker run -it -e "DATABASE_URL=postgresql://cfpb:cfpb@someserver/dbname" cfgov-build bash` to get a shell and configure a database.

## Notes

-

## Todos

- [ ] We're running under apache now, but we don't have all of our standard Apache config (like redirects, aliases, etc). This might be a place to integrate @wpears work.
- [ ] Maybe we can strip out other unnecessary files before we copy them to the deployment image (for example, any CSS, javascript, and images in the 'cfgov' directory, after collectstatic has been run)

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


rosskarchner:
With the latest push, the deployment image now runs as the Apache user, not root.

Dockerfile
```diff
@@ -1,11 +1,33 @@
...
+EXPOSE 8000
+
+
+FROM cfgov-runtime AS cfgov-dev
+
+# Install build-required packages
+RUN curl -sL https://rpm.nodesource.com/setup_10.x | bash -
+RUN curl --silent --location https://dl.yarnpkg.com/rpm/yarn.repo | tee /etc/yum.repos.d/yarn.repo
+
+RUN yum install -y nodejs yarn make gcc gcc-c++ kernel-devel unzip && \
```
wpears: Do you need nodejs and yarn here with the commands on lines 23 and 24?

Dockerfile
```diff
@@ -1,11 +1,33 @@
...
+    yum install -y httpd-devel postgresql10-devel && \
+    yum install -y python-devel python-pip python36-devel && \
+    yum clean all && rm -rf /var/cache/yum
+RUN npm install -g gulp
```
wpears: Is global gulp required here? I believe we're intending to use the gulp local to the project

Dockerfile
wpears: As an expansion on this, we can make a choice here on whether we want node and yarn in the dev image at all. If users will be running yarn commands outside of the container with the results mapped in via the volume (which I think is the intended workflow?), then node and yarn aren't needed here. Of course, we'd then have to install them in the build image, but this would limit the size of the dev image.

Then again, we can just tell people to only use yarn commands *in* the containers, and tell people they no longer need to worry about node/yarn on there systems. This would come with some dev friction, performance cost, and headache around ensuring `static_built` on the local machine isn't mapping over container copies (in practice this should work fine, but there can be footguns).

Dockerfile
rosskarchner: I'm not sure-- maybe I'll take it out and see if anything breaks

rosskarchner: You were right, that line was unneccesary

Dockerfile
rosskarchner: those lines set up the repos, I think they still need to be yum installed.

On the other question, 🤷‍♂?  You're right, if nobody will use yarn and node in the dev image, it could be punted to the build.

Dockerfile
```diff
@@ -14,17 +39,58 @@ COPY requirements /src/requirements
...
+COPY package.json .
+COPY code.json .
+COPY yarn.lock .
+COPY babel.config.js .
+COPY gulpfile.js .
+COPY gulp gulp
+COPY config config
+COPY jest.config.js .
+COPY cfgov/ ./cfgov/
+COPY cfgov-fonts-master.zip /tmp/fonts.zip
```
hkeeler: Isn't this going to make it so this image is only buildable by someone (or some service) with access to our GHE.  Could we just copy the checkout(s) right into `static.in`, that way the build would still work if that directory was empty?

rosskarchner: right now, collectstatic fails if those files aren't present--  so I think 

> only buildable by someone (or some service) with access to our GHE

describes the status-quo?

rosskarchner:
July 1st updates:

The deployment image is now Python 3 only, and the "build" step now generates an artifact, which is unpacked on the deployment image.

the build-artifact.sh script can also be used to generate an artifact deployable on RHEL6 (either Docker or a real server). Try that with 

```
(build front-end assets first)
docker build . -f Dockerfile.rhel6  -t rhel6build:latest

docker run -it -v $(pwd):/src/cfgov-refresh -v $(pwd)/cache:/root/.cache/pip rhel6build:latest scl enable python27 httpd24 "./build-artifact.sh demo 80"
```

If you are interested in how the artifact is built, check out https://github.com/cfpb/cfgov-refresh/blob/docker-multistage/build-artifact.sh

In contrast to our current "Drama Free Django" artifact format, this requires no interaction with pip, virtualenv, or collectstatic at deployment time. 

Deployment *would* be matter of uploading cfgov-refresh/build/demo.tgz to a server,  unpacking it in /srv/cfgov/ and running /srv/cfggov/cfgov-activate-next.sh (and maybe /src/cfgov/cfgov-migrate-next.sh )

This will install a script at /etc/cfgov-apache/apachectl that could be used to start and stop apache. I haven't yet figured out if there is a way to hook this into the 'standard' service startup routine for Apache 2.4 on RHEL 6, or if we would have to switch over to using this apachectl script.

I haven't yet said anything about settings. This deployment mechanism assumes that the environment variables that the app depends on are set *in the operating system environment*, and thus has no equivalent of Drama Free Django's environment.json. Stay tuned for a related PR in our internal ansible repo.

wpears:
Closes #4468

rosskarchner:
I've backed out the changes described [above](https://github.com/cfpb/cfgov-refresh/pull/5051#issuecomment-507383446), and archived it [over here](https://github.com/cfpb/cfgov-refresh/blob/docker-multistage-artifact-experiment/build-artifact.sh)

rosskarchner:
closing this, just to remove any confusion

PR #5053, opened 2019-06-03
Raise exceptions on New Relic status check failure

chosak:
The send_new_relic_messages_to_sqs script used in our cf.gov-alerts-poll-newrelic job doesn't check to make sure that the API call to New Relic succeeds before it tries to access its JSON. We've seen quite a few job failures with unhelpful error messages ("ValueError: No JSON object could be decoded") and it would be more helpful to actually see what happens when this request fails.

This change uses the built-in [`raise_for_status()`](https://2.python-requests.org/en/master/user/quickstart/#json-response-content) method to first ensure that the response was successful.

## Checklist

### DIFF
--- a/cfgov/alerts/newrelic_alerts.py
+++ b/cfgov/alerts/newrelic_alerts.py
@@ -21,6 +21,7 @@ class NewRelicAlertViolations(object):
         violations_url = (self.newrelic_url +
                           'alerts_violations.json?only_open=true')
         r = requests.get(violations_url, headers=headers)
+        r.raise_for_status()
         response_json = r.json()
 
         # Filter on the policy name


chosak:
For future reference, the error we get back from New Relic is "429 Client Error: Too Many Requests for url: https://api.newrelic.com/v2/alerts_violations.json?only_open=true".

The documentation on NR API rate limiting is [here](https://docs.newrelic.com/docs/apis/rest-api-v2/requirements/api-overload-protection-handling-429-errors). According to that, the API responses contain some useful headers like `NewRelic-OverloadProtection-Reset` that indicate how long until the limit is reset. It also suggests that we

> Send your queries less often; in particular, query less frequently than once per minute (the refresh rate for agent data).

Currently we poll NR every minute. I'm going to open a PR to change this to every 3 minutes as a way to back off our usage without significantly delaying the alerts we might receive.

PR #5067, opened 2019-06-17
Remove cf-icon-path variable

anselmbradford:
This variable should not be needed anymore, since v10 CF removes the cf-icon webfonts. CF complimentary PR here https://github.com/cfpb/capital-framework/pull/933

## Removals

- Remove cf-icon-path variable


anselmbradford:
Turns out this variable should be used but wasn't https://github.com/cfpb/capital-framework/pull/933

PR #5071, opened 2019-06-18
Clean up copy gulp task

anselmbradford:
We create variables to point to paths that are only used once in all but one case, this PR removes the reference to the "config" file and points to the paths used directly, making it clearer by looking at the task what is happening. It also removes the deprecated webfont icon tasks.

## Removals

- Deprecated iconsOld task for webfont icons.

## Changes

- Moves copy config paths directly into the copy gulp task.

## Testing

1. `gulp build` should pass.


gulp/tasks/copy.js
```diff
@@ -1,7 +1,14 @@
...
-const configCopy = require( '../config' ).copy;
 const handleErrors = require( '../utils/handle-errors' );
+const paths = require( '../../config/environment' ).paths;
+
+/*
+  Path to the cf-icons SVG icons folder,
+  which gets copied into the static directory on production
+  so SVG references in CSS get resolved.
+*/
+const iconSrc = paths.modules + '/cf-icons/src/icons/*.svg';
```
Scotchester: Want to use a template string here?

gulp/tasks/copy.js
```diff
@@ -17,36 +24,43 @@ function _genericCopy( src, dest ) {
...
+    paths.processed
+  );
+  return stream;
 } );
 
 gulp.task( 'copy:lightbox2', () => {
-  const lightbox2 = configCopy.lightbox2;
-  return _genericCopy( lightbox2.src, lightbox2.dest );
+  const stream = _genericCopy(
+    [ `${ paths.modules }/lightbox2/dist/**/*` ],
```
Scotchester: Seems like this array isn't actually necessary, yeah?

gulp/tasks/copy.js
anselmbradford: Sure, updated it.

gulp/tasks/copy.js
anselmbradford: Hmm doesn't look like it. This was the original format of the source, but I don't see how it's needed as an array either.

PR #5082, opened 2019-06-21
Pin django-js-asset package to match TDP

chosak:
The [TDP satellite app](https://github.com/cfpb/teachers-digital-platform/) includes pinned package dependencies for `django-mptt==0.9.0` and `django-js-asset==1.1.0`. Currently cfgov-refresh only includes the pinned dependency on django-mptt.

The django-mptt dependency includes [an unpinned dependency](https://github.com/django-mptt/django-mptt/blob/615cd547625bef2ce86c46d7e071787f194ac924/setup.py#L20) on django-js-asset. There have been [several releases](https://pypi.org/project/django-js-asset/#history) of django-js-asset since the 1.1.0 version. This means that, depending on how the cfgov-refresh and TDP dependencies are installed, you might end up with a later version of django-js-asset, that doesn't match what is specified in TDP.

For this reason, this PR fixes the django-js-asset here in cfgov-refresh, to ensure that it always matches what is specified in TDP.

## Testing

Create a virtualenv using Python 3 and try doing

```sh
$ pip install -r requirements/local.txt -r requirements/optional-public.txt
```

Doing this produces a warning on master (`ERROR: teachers-digital-platform 1.0.33 has requirement django-js-asset==1.1.0, but you'll have django-js-asset 1.2.2 which is incompatible.`) but none on this branch.

## Checklist

### DIFF
--- a/requirements/libraries.txt
+++ b/requirements/libraries.txt
@@ -6,6 +6,8 @@ django-csp==3.4
 django-extensions==2.1.3
 django-flags==4.1.0
 django-haystack==2.7.0
+# django-js-asset is required by teachers-digital-platform
+django-js-asset==1.1.0
 django-localflavor==1.6.2
 # django-mptt is required by teachers-digital-platform
 django-mptt==0.9.0


chosak:
> Would we want to explore updating django-js-asset in TDP at some point?

That would probably be a question for the TDP team. I don't know what this package does.

PR #5087, opened 2019-06-25
Expand our videos' iframe container on all screen sizes

contolini:
For some reason we were only adjusting video iframe's height and video controls position on larger screens. They are now adjusted on all screen sizes.

I have tested these changes on a handful of FCM and non-FCM videos and it seems okay but I could use more eyes if anyone's able.

- http://0.0.0.0:8000/consumer-tools/managing-someone-elses-money/
- http://0.0.0.0:8000/about-us/blog/mortgage-closing-scams-how-protect-yourself-and-your-closing-funds/
- https://[mattermost]/cfpb/pl/841dszdrrtfx5dpamz7tq7j89h

**Before:**

Screen Shot 2019-06-25 at 3 36 14 PM

**After:**

Screen Shot 2019-06-25 at 3 35 08 PM

### DIFF
--- a/cfgov/unprocessed/css/video-player.less
+++ b/cfgov/unprocessed/css/video-player.less
@@ -121,11 +121,20 @@
     margin: auto;
 }
 
+.video-player_iframe-container {
+    height: 100%;
+    margin: 0;
+}
+
+.video-player_video-actions-container {
+    right: 0;
+    left: auto;
+}
+
 .respond-to-min( @bp-sm-min, {
     .video-player_video-actions-container {
         position: absolute;
         top: 0;
-        left: @video-width__px;
     }
 
     .video-player .m-social-media,
@@ -210,20 +219,6 @@
         .u-flexible-container_inner();
     }
 } );
-
-
-.respond-to-min( @bp-xl-min, {
-    .video-player_iframe-container {
-        height: 100%;
-        margin: 0;
-    }
-
-    .video-player_video-actions-container {
-        right: 0;
-        left: auto;
-    }
-} );
-
 
 // Special styles for playing within an FCM
 .o-featured-content-module {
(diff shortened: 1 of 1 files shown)


cfgov/unprocessed/css/video-player.less
```diff
@@ -121,6 +121,16 @@
...
+    margin: 0;
+}
+
+.video-player_video-actions-container {
+    right: 0;
+    left: auto;
+}
+
 .respond-to-min( @bp-sm-min, {
     .video-player_video-actions-container {
```
niqjohnson: This might just be some sort of cached style I'm not clearing right, but when I test out FCM videos, the close button looks wonky at all screen sizes:

wonky-close

Looks like maybe the `left: auto;` is getting overridden by the `left: @video-width__px;` in this media query? Alternatively, it's 100% possible I didn't build something right and this isn't actually a problem.

contolini: Nooooope you're right. I'll spend some more time with this.

contolini:
@niqjohnson it appears the `left` positioning was cruft from the previous FCM layout. I removed it and the button behaves nicer now. Give it a whirl.

Scotchester:
Thanks so much for jumping in on this while I was out, @contolini. This is working great for me at all screen sizes on Firefox except for XS, where the padding is unnecessarily black (and too large on bottom):

Screen Shot 2019-06-26 at 12 03 44

This seems to be unrelated to the changes currently in this PR, but I'd like to take a little time to fix it here all the same. I'll handle that and push to this branch when done.

Scotchester:
Made that fix, rebased on master, and force-pushed. @contolini mind giving it a look-see?

Scotchester:
Thanks, @niqjohnson!

PR #5092, opened 2019-06-27
Migrate rural or underserved tool into cfgov-refresh

anselmbradford:
Migrates https://github.com/cfpb/rural-or-underserved-test into this repo.

## Additions

- Rural-or-underserved-tool app assets and jinja template.
- Census API to external API call whitelist.

## Testing

1. Pull branch and run `./frontend.sh && gulp clean && gulp build` and run a local server.
2. Visit http://localhost:8000/rural-or-underserved-tool/ and compare to https://www.consumerfinance.gov/rural-or-underserved-tool/

## Screenshots

rural

rural2

## Notes
- There are many classes that can likely be cleaned out, but this provides the initial migration of this app.
- Layout differs in some respects to the live site by necessity of fitting into a standard layout. Big differences:
  - About section is always visible.
  - Maps in the result table retain a top border line.


chosak:
In Firefox the dropdown has visibly changed from this:

image

to this:

image

Neither of these look great, but the latter looks a bit too squeezed to me, and even more squeezed than the screenshot you shared.

Additionally, the FAQs don't show up as collapsed for me:

image

and in the developer console, I see this warning:

image

This is all running locally, after doing a `gulp clean && gulp build` as you suggest. I did see this warning during the build:

```sh
[12:25:01] Using gulpfile ~/Projects/cfgov-refresh/gulpfile.js
[12:25:01] Starting 'build'...
[12:25:01] Starting 'styles'...
[12:25:01] Starting 'scripts'...
[12:25:01] Starting 'images'...
[12:25:01] Starting 'styles:apps'...
[12:25:01] Starting 'styles:featureFlags'...
[12:25:01] Starting 'styles:ie'...
[12:25:01] Starting 'styles:modern'...
[12:25:01] Starting 'styles:nemo'...
[12:25:01] Starting 'styles:ondemand'...
[12:25:01] Starting 'scripts:polyfill'...
[12:25:01] Starting 'scripts:modern'...
[12:25:01] Starting 'scripts:apps'...
[12:25:01] Starting 'scripts:external'...
[12:25:01] Starting 'scripts:nemo'...
[12:25:01] Starting 'scripts:ondemand'...
[12:25:01] Starting 'images:apps'...
[12:25:01] Starting 'images:general'...
[12:25:01] Starting 'styles:nemoProd'...
[12:25:01] Starting 'styles:nemoIE8'...
[12:25:03] Started generating service worker file...
[12:25:03] Copying eRegs' manifest...
App dependencies not installed, please run from project root: npm --prefix ./cfgov/unprocessed/apps/rural-or-underserved-tool install ./cfgov/unprocessed/apps/rural-or-underserved-tool
[12:25:03] Starting 'scripts:ondemand:header'...
[12:25:03] Starting 'scripts:ondemand:footer'...
[12:25:03] Starting 'scripts:ondemand:nonresponsive'...
[12:25:03] Finished 'styles:featureFlags' after 1.91 s
[12:25:03] webpack-stream - No files given; aborting compilation
[12:25:03] Successfully copied eRegs' manifest to ./cfgov/static_built/apps/regulations3k/regulations3k-manifest.json
[12:25:03] Generated cfgov/regulations3k/jinja2/regulations3k/regulations3k-service-worker.js, which will precache 0 files, totaling 0 bytes.
```

Note the message about ROUT dependencies not being installed. Is this expected?

anselmbradford:
Updated drop-down to use CF `a-select` styling in https://github.com/cfpb/cfgov-refresh/pull/5092/commits/84a379ef303322bee60c89d280a5adf59c4cd060

Screen Shot 2019-06-27 at 1 11 25 PM

It's pretty tight at mobile. Not sure how we want to handle that.

cfgov/unprocessed/apps/rural-or-underserved-tool/js/dom-tools.js
```diff
@@ -0,0 +1,303 @@
...
+
+/* Code copied from jQuery with minimal modifications.
+   XHTML parsers do not magically insert elements in the
+   same way that tag soup parsers do. So we cannot shorten
+   this by omitting <tbody> or other required elements. */
+const firstTag = /<([a-z][^\/\0>\x20\t\r\n\f]+)/;
+const wrapMap = {
+  'col':     [ 2, '<table><colgroup>', '</colgroup></table>' ],
+  'default': [ 0, '', '' ],
+  'option':  [ 1, '<select multiple=\'multiple\'>', '</select>' ],
```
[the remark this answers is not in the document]
anselmbradford: Unused in this PR and this creates a `select` around options, so there may well have an associated label.

cfgov/jinja2/v1/rural-or-underserved/index.html
```diff
@@ -0,0 +1,961 @@
...
+                                    </button>
+                                    <input id="file-name"
+                                           name="file-name"
+                                           class="a-text-input"
+                                           value="No file chosen">
+                                </div>
+                            </p>
+
+                            <p>
+                                <input class="a-btn" type="submit" value="Check addresses" />
```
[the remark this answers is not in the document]
anselmbradford: Converted to a button.

cfgov/jinja2/v1/rural-or-underserved/index.html
```diff
@@ -0,0 +1,961 @@
...
+                                    <!-- Styled faux overlay component. -->
+                                    <button class="a-btn
+                                                   a-btn__secondary
+                                                   a-btn__flush-right">
+                                        Select file
+                                    </button>
+                                    <input id="file-name"
+                                           name="file-name"
+                                           class="a-text-input"
+                                           value="No file chosen">
```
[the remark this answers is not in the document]
anselmbradford: Added visually hidden label.

cfgov/jinja2/v1/rural-or-underserved/index.html
```diff
@@ -0,0 +1,961 @@
...
+                                        name="address1"
+                                        type="text" />
+                                </span>
+                            </p>
+                            <p>
+                                <a class="no-visit" href="#" id="add-another" title="Add another address">
+                                    Add another address
+                                </a>
+                            </p>
+                            <input class="a-btn" type="submit" value="Check addresses" />
```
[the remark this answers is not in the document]
anselmbradford: Converted to a button.

cfgov/jinja2/v1/rural-or-underserved/index.html
```diff
@@ -0,0 +1,961 @@
...
+                                    <label class="a-label u-visually-hidden" for="address1">
+                                        Street address, city, state/territory,
+                                        or ZIP code.
+                                    </label>
+                                    <input
+                                        class="a-text-input
+                                            a-text-input__full
+                                            input-address"
+                                        name="address1"
+                                        type="text" />
```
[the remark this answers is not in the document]
anselmbradford: Added id for label.

cfgov/jinja2/v1/rural-or-underserved/index.html
```diff
@@ -0,0 +1,961 @@
...
+
+                    <h2>
+                        Check status of properties for loans extended in
+                        <small class="u-inline-select">
+                            <div class="m-form-field m-form-field__select">
+                                <label class="a-label a-label__heading u-visually-hidden" for="test_select_default">
+                                    Year of loan
+                                </label>
+                                <div class="a-select">
+                                    <select id="year" name="year">
```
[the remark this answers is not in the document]
anselmbradford: Added id for label.

niqjohnson:
@anselmbradford: I pushed up a change to convert the FAQs from individual expandables to expandable groups to match a little more closely what's there now. Take a look at that and see what you think.

Also, there's a slight style bug I'm seeing where the expandable headings aren't vertically centered because they're still getting the `margin-bottom` from `.h4` applied to them:

Current | Expected
---- | ----
current | expected

Looking at other expandable groups (like the ones on the [TILA-RESPA Integrated Disclosure FAQs](https://www.consumerfinance.gov/policy-compliance/guidance/tila-respa-disclosure-rule/tila-respa-integrated-disclosure-faqs/) page), I think this might have something to do with the order the Capital Framework styles are getting imported? It looks like `.o-expandable_label` is meant to override the bottom margin on `.h4`, but instead `.h4` is overriding `.o-expandable_label`. It's a minor thing, but I wasn't sure the best way to fix it.

niqjohnson:
@anselmbradford, I pushed up a couple more small style updates (more info on those in the GHE issue). I'm pretty secure in the design side of the updates but much less secure about the CSS I used to make the design side happen, so feel free to give the code a look and change anything that needs changing.

niqjohnson:
@anselmbradford, one other thing I noticed while I was poking around: it looks like the `body` of the page is getting an `overflow-x: hidden;` style from somewhere at small screen sizes (maybe `mega-menu.less` if Chrome's dev tools are to be believed). That's preventing the page from horizontally scrolling on small screens to let you view the full width of the result tables.

anselmbradford:
> one other thing I noticed while I was poking around: it looks like the body of the page is getting an overflow-x: hidden; style from somewhere at small screen sizes (maybe mega-menu.less if Chrome's dev tools are to be believed). That's preventing the page from horizontally scrolling on small screens to let you view the full width of the result tables.

Hmm, yeah most definitely from the mega-menu, since the hamburger menu sits along side the page's contents to enable it to slide, so this sounds like an issue we need to resolve within the table itself.

niqjohnson:
@anselmbradford: I wonder if we should convert the tables to the [horizontal scroll version](https://cfpb.github.io/capital-framework/components/cf-tables/#responsive-table---horizontal-scroll-variation)? Seems like in theory what we want to happen, but I don't think I've every actually seen one of our scrolling tables in the wild yet, so I'm not sure if that's the right way to go.

anselmbradford:
> Also, there's a slight style bug I'm seeing where the expandable headings aren't vertically centered because they're still getting the margin-bottom from .h4 applied to them:

@niqjohnson I fixed this in https://github.com/cfpb/cfgov-refresh/pull/5092/commits/1363d7749dfffe408737da76c5ffe85d93e1d599 You were right, the order was wrong (actually cf-core was being delivered twice).

anselmbradford:
> I wonder if we should convert the tables to the horizontal scroll version? Seems like in theory what we want to happen, but I don't think I've every actually seen one of our scrolling tables in the wild yet, so I'm not sure if that's the right way to go.

Converted to scrolling tables in https://github.com/cfpb/cfgov-refresh/pull/5092/commits/e9a816701e19fc68c9251fbb01ee121cd436bc8b
Also, I converted the summary counters at the top (the large numbers) to inline-flex instead of inline-grid, since they overlapped with the text if they were two digits.
Lastly, I wasn't sure if the show more buttons were too close?

Screen Shot 2019-07-12 at 8 03 09 PM

cfgov/jinja2/v1/rural-or-underserved/index.html
```diff
@@ -294,20 +294,24 @@ <h2 class="js-table-title">
                             First-lien covered transactions extended in <span class="chosenYear">2015</span> that are secured by these properties can be counted as rural or underserved in your determination of your first-lien covered transactions in <span class="chosenYear">2015</span> that are secured by properties located in rural or underserved areas.
                         </p>
 
-                        <table class="table" style="border-width: 1px;" width="100%">
```
niqjohnson: Can we restore the 100% width to all the tables and still keep them scrollable? I think for Wagtail-generated tables we do it with a `class=u-w100pct` on the `table`, but I'm not sure if that will conflict with the scrollable variation. I think making all the tables 100% width again will help with quickly scanning the page since it puts the columns in the same position across all four sections, which means your eye won't have to jump around as much to see the same columns as you scroll.

cfgov/jinja2/v1/rural-or-underserved/index.html
```diff
@@ -294,20 +294,24 @@ <h2 class="js-table-title">
...
+                                        <th width="22%">Rural or Underserved?</th>
+                                    </tr>
+                                </thead>
+                                <tbody class="data"></tbody>
+                            </table>
+                        </div>
+
+                        <p class="block block__flush-top block__flush-bottom block__padded-top block__padded-bottom">
+                            <a data-table="rural" class="button-more u-hidden a-btn a-btn__secondary no-decoration" href="#" id="ruralMore">View 10 more</a>
+                            <a data-table="rural" class="margin-left-1 view-all u-hidden no-visit" href="#" id="ruralAll">View all</a>
```
niqjohnson: Let's give these buttons the same separation as similar button combos on cf.gov, like the "Apply filters" and "Clear filters" buttons in filter modules. I think that's a 5px separation.

cfgov/jinja2/v1/rural-or-underserved/index.html
```diff
@@ -294,20 +294,24 @@ <h2 class="js-table-title">
...
-                                <th width="15%">County</th>
-                                <th width="22%">Rural or Underserved?</th>
-                                </tr>
-                            </thead>
-                            <tbody class="data"></tbody>
-                        </table>
-
-                        <a data-table="rural" class="button-more u-hidden a-btn a-btn__secondary no-decoration" href="#" id="ruralMore">View 10 more</a>
-                        <a data-table="rural" class="margin-left-1 view-all u-hidden no-visit" href="#" id="ruralAll">View all</a>
+                        <div class="o-table o-table-wrapper__scrolling">
```
niqjohnson: It looks like our scrolling tables have zebra striping baked in. Ordinarily adding zebra striping seems fine to me, but in this case I think we might want to remove it since it gets weird in the top two result tables. Every other row in those tables is initially hidden since they contain the expandable map, so when the results load it looks like those tables aren't striped at all. Then when you expand the map for a row, the results row is not striped but it's corresponding map row is. So maybe overriding the default zebra striping is the way to go here? Alternatively, I'd also be fine if we got zebra striping working in the top two results tables in a way that (a) showed when the results load and (b) striped a result row and its corresponding map row the same color.

cfgov/jinja2/v1/rural-or-underserved/index.html
anselmbradford: Hmm, 100% is set on `o-table` [here](https://github.com/cfpb/capital-framework/blob/master/packages/cf-tables/src/cf-tables.less#L118), but the scrollable table wraps the table in a div, so `o-table` ends up on the container, not the table. I don't particularly like having two table patterns where `o-table` is applied, so this may need to get threshed out in cf-tables. But in the mean time, we can add it here.

cfgov/jinja2/v1/rural-or-underserved/index.html
anselmbradford: It'll be updated to this in the next push
Screen Shot 2019-07-16 at 10 18 24 AM

cfgov/jinja2/v1/rural-or-underserved/index.html
anselmbradford: I overrode the zebra striping and added a TODO for adjusting the styles in cf-tables.

cfgov/jinja2/v1/rural-or-underserved/index.html
anselmbradford: I overrode the table width and added a TODO for adjusting the styles in cf-tables.

cfgov/jinja2/v1/rural-or-underserved/index.html
anselmbradford: Also opened https://github.com/cfpb/capital-framework/issues/946

Scotchester: These should all be `button` elements, too.

anselmbradford: Ah, true

cfgov/jinja2/v1/rural-or-underserved/index.html
niqjohnson: Looks good 👍

cfgov/jinja2/v1/rural-or-underserved/index.html
niqjohnson: Looks good 👍

cfgov/jinja2/v1/rural-or-underserved/index.html
niqjohnson: The spacing looks good to me

PR #5094, opened 2019-06-27
Regulations: Deprecate letter codes in favor of short names

willbarton:
This change removes our letter codes and adds short names in their place. A regulation with the letter code `Z` will now have the short name `Regulation Z`. This will let us add regulations that don't have letter codes but might be referred to in other ways, like `Payday Rule`.

## Checklist


cfgov/regulations3k/migrations/0027_remove_part_letter_code.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('regulations3k', '0026_migrate_letter_code_to_short_name'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
schbetsy: I think I heard that this step (removing the unused field after a data migration) can't happen as part of the same release as the data migration. Is that the decision you landed on, @chosak and @higs4281 ?

chosak: Theoretically, to conservatively avoid errors, yes, it should be removed in a subsequent deployment. Consider the order that things happen:

0. Current version of source code is running, depending on `letter_code` field.
1. New version of source code is copied to servers, that uses `short_name` field instead. This version is not activated yet.
2. New migrations in new version are run, while old version is still running.
   1. Migration 25 adds the `short_name` field.
   2. Migration 26 copies data from `letter_code` to `short_name`.
   3. Migration 27 removes the `letter_code` field.
3. New version of source code is activated.

Critically, any requests coming into the servers between 2.iii. and 3. will trigger 500 errors, because the running source code (the old version) still depends on the `letter_code` column that has been removed from the database. If all goes well with the deploy, this might only be a few seconds (in practice, the time between lines 46 and 80 in [this file](https://GHE/cfpb-infra/ansible-consumer-finance/blob/master/roles/frontend/tasks/activate_cfgov_release.yml#L46) (requires GHE domain). If one of those steps fails, however, the server would be left in a broken state.

The tradeoff between the additional complexity of another PR and the chance of errors happening depends on the risk tolerance of these particular pages.

willbarton: We should probably just come down on one side or another as a best-practice, and put it in the documentation. I don't mind being more conservative with the migrations and our current configuration, personally.

For now, I've removed the last migration in 730740a.

willbarton: 15d5e9b adds the field back to the model itself.

Issue #5111, opened 2019-07-12
Outdated documentation about Node version


chosak:
The current documentation cites requirements of

> Node.js v8.5 or higher, and npm v5.2 or higher"

But the frontend build seems to require at least Node 10. The docs should be updated.

Markdown source is [here](https://github.com/cfpb/cfgov-refresh/blob/dd09c94e5571a2f6154a07747d78ca1586b8d35b/docs/installation.md#front-end-dependencies), live version is [here](https://cfpb.github.io/cfgov-refresh/installation/#front-end-dependencies).

## Screenshots

![image](https://user-images.githubusercontent.com/654645/61132536-8e83fc00-a489-11e9-92b8-6f40ee4bd311.png)

anselmbradford:
Fixed in https://github.com/cfpb/cfgov-refresh/pull/5114

PR #5114, opened 2019-07-15
Update installation.md to address outdated node version

anselmbradford:
Fixes https://github.com/cfpb/cfgov-refresh/issues/5111

## Changes

- Remove node/npm version from docs and defer to saying user should receive an error on outdated version.
- Remove CF note since CF link goes to the same place.
- Minor formatting updates.

## Testing
- `mkdocs serve -a :8888` and visit local http://127.0.0.1:8888/installation/#front-end-dependencies


willbarton:
Or http://localhost:8888/installation/#front-end-dependencies for those of us docker users 😉

PR #5110, opened 2019-07-11
Add ability to build and test drama-free-django builds

chosak:
This PR adds the ability to execute the cf.gov [drama-free-django](https://github.com/cfpb/drama-free-django) build using a CentOS 6 Docker container. This creates a DFD build that can be successfully run using Python 2 on a CentOS 6 system. This makes public some of our build process which is currently configured via an internal Jenkins repository.

This also adds the ability to test a DFD build artifact, again using a CentOS 6 container. Testing consists of both extracting the artifact and running Django [`collectstatic`](https://docs.djangoproject.com/en/1.11/ref/contrib/staticfiles/#collectstatic). This will give us much earlier detection of missing/misconfigured static assets as part of PR checks (ping @anselmbradford).

Besides adding the ability for developers to run these steps, this PR also adds these steps to the automatic Travis CI checks that run as part of this repository. You can see an example run on my fork [here](https://travis-ci.org/chosak/cfgov-refresh/jobs/557505868), which took around 5 minutes, comparable to the rest of the build steps.

## Testing

To generate a DFD build from your local source code, execute this command from the root of your clone of the cfgov-refresh repository:

```sh
./docker/drama-free-django/build.sh
```

This will generate a DFD artifact named cfgov_current_build.zip in that same root location. Besides generating that file, this will also touch your filesystem by running the frontend build in your source code directory.

To then run a basic test using that artifact, execute this command from the same location:

```sh
./docker/drama-free-django/test.sh
```

This will run a container, expand the artifact, and run Django collectstatic. This won't touch your local filesystem at all.

You should also be able to see both of these steps running as part of a new Travis "drama-free-django" step.

## Notes

- I purposefully avoided using Dockerfiles here, instead choosing to pass complete scripts to plain centos:6 containers. I did this because, without a Docker image registry or a way to periodically purge the Docker image cache, we'd have to deal with maintaining/upgrading version
upgrades for things like pip/yarn if we included them in the base image. Using simple scripts felt more straightforward and allows us to more easily reuse code across Jenkins, Travis, and local development.
- The testing script contains a bit of magic to get around restrictions caused by our use of non-public webfont files. I thought that the benefits of running collectstatic on Travis outweighed the slight hackiness of this approach.

## Todos

@rosskarchner had a good idea about mapping volumes to allow Travis to cache pip and yum caches across builds. This would speed up the build and test process.

## Checklist

- [X] PR has an informative and human-readable title
- [X] Changes are limited to a single goal (no scope creep)
[...]

PR #5117, opened 2019-07-17
Run Docker DFD build in Travis in detached mode

chosak:
The drama-free-django build using a Docker container that runs in Travis (introduced recently in #5110) seems to fail strangely sometimes, e.g.

https://travis-ci.org/cfpb/cfgov-refresh/jobs/560138901#L1314

This works locally. I suspect this is some kind of timeout, but am not sure. This change modifies the way that the Travis script works, so that it runs the container in detached mode and then attaches to the logs. I've verified that this works (at least once) on my fork.

This change is intended to fix the Travis failures currently happening on #5116.

## Checklist

### DIFF
--- a/docker/drama-free-django/travis.sh
+++ b/docker/drama-free-django/travis.sh
@@ -1,6 +1,11 @@
 #!/usr/bin/env bash
 
-docker run \
-    -v `pwd`:/cfgov \
-    centos:6 \
-    /bin/bash -c "/cfgov/docker/drama-free-django/_build.sh && /cfgov/docker/drama-free-django/_test.sh"
+container_id=$(
+    docker run \
+        -d \
+        -v `pwd`:/cfgov \
+        centos:6 \
+        /bin/bash -c "/cfgov/docker/drama-free-django/_build.sh && /cfgov/docker/drama-free-django/_test.sh"
+)
+
+docker logs -f $container_id


chosak:
@hkeeler good question! I got the idea to run in detached mode from [this Travis documentation page](https://docs.travis-ci.com/user/docker/) and figured it was worth a try.

Adding `set -x` is a good idea. In the case of the failing runs on @Scotchester's PR, the job just seemed to end unexpectedly in the middle of the collectstatic command [here](https://github.com/cfpb/cfgov-refresh/blob/16b834a0ac5dfe907c8ffc36153ba5726013243d/docker/drama-free-django/_test.sh#L58). That's a command that generates a whole bunch of output. I hypothesized that maybe the way that generated output combined with the container shutting down was somehow causing the error.

The error here is in the middle of the regular collectstatic output:

image

[Here's](https://github.com/travis-ci/travis-ci/issues/6018) a lengthy Travis issue where people suggest using sleep statements in similar cases to keep the container alive long enough for all of the output to get written. I can't imagine how collectstatic would actually be _failing_ in these jobs, unless there's some kind of memory condition or something else unexpected. I can try to see if I can reproduce this failure reliably on my fork and see if sleep somehow gives additional clues; I agree that running detached feels like a weird hack without clear justification.

PR #5122, opened 2019-07-23
Add apache

wpears:
This PR adds our formerly private apache conf files into our open source repo. The benefits are manifold:
 - better alignment with out open source policy
 - easier to set up a production-like system locally (specifically, adding an option to run a docker container running apache and invoking django through wsgi, rather than using runserver. This will come in a later PR in order to not clutter this one).
 - with the point above, it will also be easier and faster to change and test our apache setup (or even experiment with replacing apache ^_______^)


cfgov/apache/conf.d/alias.tmpl
```diff
@@ -0,0 +1,20 @@
...
+Alias /favicon.ico ${STATIC_PATH}/favicon.ico
+Alias /foia/quarterly ${STATIC_PATH}/foia/quarterly
+# Facebook site auth
+Alias /lln8595c61g9qnvuwvtlcwo1k6kem8.html ${STATIC_PATH}/lln8595c61g9qnvuwvtlcwo1k6kem8.html
+Alias /robots.txt ${STATIC_PATH}/robots.txt
+Alias /static/f/ ${STATIC_PATH}/nemo/_/f/
+# Preserves legacy URL compatibility, but this is not actually using Wordpress
+Alias /wp-content/themes/cfpb_nemo/_/  ${STATIC_PATH}/nemo/_/
+
+# Static assets served from files on disk.
```
chosak: I think these staticsites sections should be removed now that #5092 moved rural-or-underserved-tool from a static site into Django.

cfgov/apache/conf.d/newsroom.conf
```diff
@@ -0,0 +1,12 @@
+SetEnv NEWSROOM enable
```
chosak: This newsroom.conf file is obsolete and has been removed from the internal repository; it should be removed here as well.

cfgov/apache/conf.d/headers.conf
```diff
@@ -0,0 +1,30 @@
...
+</LocationMatch>
+
+# this was a test, that we may want to refer back to later. Leaving it (but commented) for posterity
+#<LocationMatch ^\/company-signup\/$>
+#        Header always set Edge-Control: no-store
+#</LocationMatch>
+
+<LocationMatch ^\/consumer-tools/prepaid-cards\/>
+        Header always set Edge-Control: no-store
+</LocationMatch>
```
chosak: This doesn't include the Access-Control-Allow-Origin header currently defined in the private headers.conf. Currently we set that to https://www.consumerfinance.gov on production and beta -- which, frankly, seems wrong, which makes me think something isn't configured right there:

```sh
$ curl -I https://beta.consumerfinance.gov
HTTP/1.1 200 OK
...
Access-Control-Allow-Origin: https://www.consumerfinance.gov
```

How has this not bitten us on beta? What role does this header play?

On the dev servers we instead set the headers like this:

```
Header set Access-Control-Allow-Origin: "*"
Header set Access-Control-Allow-Headers: "X-Requested-With, Content-Type, Access-Control-Allow-Headers"
```

Shouldn't we set these similarly here?

wpears: That header is being provided in a conditional left in ansible:
https://GHE/cfpb-infra/ansible-consumer-finance/blob/remove-apache/roles/frontend/templates/apache-conditionals.conf.j2#L38-L43

cfgov/apache/conf.d/alias.conf
```diff
@@ -11,3 +11,4 @@ Alias /robots.txt ${STATIC_PATH}/robots.txt
 Alias /static/f/ ${STATIC_PATH}/nemo/_/f/
 # Preserves legacy URL compatibility, but this is not actually using Wordpress
 Alias /wp-content/themes/cfpb_nemo/_/  ${STATIC_PATH}/nemo/_/
+Alias /static/ ${STATIC_PATH}/
```
chosak: The comment in line 1 (`# These paths are served by WSGI as configured in wsgi.conf.`) is no longer accurate, since these aren't served by WSGI, but are served by Apache. Can it be removed?

Also, if we're aliasing `/static/` to `${STATIC_PATH)/`, then can/should we remove most of the explicit list of files above (e.g. `Alias /code.json ${STATIC_PATH}/code.json`, since they're now redundant? The only ones that aren't redundant are `/static/f` to `${STATIC_PATH}/nemo/_/f/` and the `wp-content` one.

wpears: I'll certainly remove the comment, but for the second part I don't think they're actually redundant, because the url is, eg, `consumerfinance.gov/code.json` which needs to be mapped to `<cfgov_path>/static/code.json`. The only way they'd be redundant is if we changed all the urls to be in the form of  `consumerfinance.gov/static/code.json`, which I don't think we want to do.

chosak: Ack, you're right. I totally misread that. Thanks for the correction.

cfgov/apache/conf/magic
```diff
@@ -0,0 +1,385 @@
+# Magic data for mod_mime_magic Apache module (originally for file(1) command)
```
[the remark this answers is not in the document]
wpears: Nope, I'm now mapping it from the default location in ansible.

cfgov/apache/conf.d/redirects.conf
```diff
@@ -0,0 +1,1030 @@
...
+Redirect permanent /complaintdatabase/technical-documentation  https://cfpb.github.io/api/ccdb/
+Redirect permanent /complaintdatabase/technical-documentation/ https://cfpb.github.io/api/ccdb/
+
+# New document releases
+Redirect permanent /f/201501_cfpb_list-consumer-reporting-agencies.pdf /f/201604_cfpb_list-of-consumer-reporting-companies.pdf
+Redirect permanent /f/201601_cpfb_list-of-consumer-reporting-companies.pdf /f/201604_cfpb_list-of-consumer-reporting-companies.pdf
+
+# redirects associated with the v1 project
+
+    Redirect permanent /pressreleases/ /about-us/newsroom/
```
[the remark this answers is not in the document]
wpears: Nope, fixed.

cfgov/apache/conf.d/alias.conf
```diff
@@ -0,0 +1,13 @@
+# Bing site auth
+Alias /BingSiteAuth.xml ${STATIC_PATH}/BingSiteAuth.xml
+Alias /code.json ${STATIC_PATH}/code.json
+Alias /data.json ${STATIC_PATH}/data.json
+Alias /favicon.ico ${STATIC_PATH}/favicon.ico
+Alias /foia/quarterly ${STATIC_PATH}/foia/quarterly
+# Facebook site auth
+Alias /lln8595c61g9qnvuwvtlcwo1k6kem8.html ${STATIC_PATH}/lln8595c61g9qnvuwvtlcwo1k6kem8.html
+Alias /robots.txt ${STATIC_PATH}/robots.txt
+Alias /static/f/ ${STATIC_PATH}/nemo/_/f/
```
chosak: We need to add entries for the Akamai SureRoute test objects, as implemented over in ansible#316.

wpears: Added

cfgov/apache/conf.d/wsgi.conf
```diff
@@ -0,0 +1,20 @@
+ServerName consumerfinance.gov
+
+WSGIApplicationGroup %{GLOBAL}
+WSGIDaemonProcess django  processes=${APACHE_PROCESS_COUNT} threads=15 display-name=%{GROUP} python-path=${CFGOV_CURRENT}/venv/lib/python2.7/site-packages:${CFGOV_CURRENT}/venv/lib64/python2.7/site-packages:${CFGOV_CURRENT} home=${CFGOV_CURRENT}
```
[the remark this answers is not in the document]
rosskarchner: If we got he mod-wsgi-express route, this file won't be relevant in the docker world.

PR #5123, opened 2019-07-23
Featured menu content fix

virginiacc:
Changes wrapper element for featured menu content module's text from `<p>` to `<div>` to prevent HTML validation issues when text contains paragraph tags. Addresses GHE issue 3302

## Changes

- Updates wrapper element in `featured-menu-content.html`

## Testing

1. Check out branch and navigate to http://localhost:8000/
2. Verify that HTML validation errors no longer appear for the  `.m-featured-menu-content` sections 

### Before
9ac0db80-344c-11e9-91c6-3478cf552742

### After
Screen Shot 2019-07-23 at 7 26 33 AM


Scotchester:
If this is a rich text field that will add its own `div.rich-text` wrapper, we don't need this HTML element wrapping it in our template at all, do we?

virginiacc:
@Scotchester Good point -- wrapper element removed in https://github.com/cfpb/cfgov-refresh/pull/5123/commits/6726b1ab5ff9b9f5e3c9878048a4ea3ef04b40c4

PR #5124, opened 2019-07-23
Ensure that related_posts returns specific pages

willbarton:
The PR fixes a bug where events show up as related posts, they would show up with the published date instead of the event start date, despite template logic that looked for the state date of events. 

So, a listing of related events on the site right now (on this page for example: https://www.consumerfinance.gov/about-us/events/archive-past-events/town-hall-baton-rouge-la-preventing-elder-financial-exploitation/) looks like this:

image

These are published dates. After this change this same listing looks like:

image

This change adds a call to Wagtail's `specific` property on each returned page in the `RelatedPosts` organism that will fetch the actual model class and fields for the page.

## Checklist

### DIFF
--- a/cfgov/v1/atomic_elements/organisms.py
+++ b/cfgov/v1/atomic_elements/organisms.py
@@ -329,7 +329,7 @@ class RelatedPosts(blocks.StructBlock):
         specific_categories = value['specific_categories']
         limit = int(value['limit'])
         queryset = AbstractFilterPage.objects.live().exclude(
-            id=page.id).order_by('-date_published').distinct()
+            id=page.id).order_by('-date_published').distinct().specific()
 
         for parent in related_types:  # blog, newsroom or events
             # Include children of this slug that match at least 1 tag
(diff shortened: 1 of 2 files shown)


cfgov/v1/atomic_elements/organisms.py
```diff
@@ -360,8 +360,13 @@ def match_all_topic_tags(queryset, page_tags):
 
             related_items[parent.title()] = related_queryset[:limit]
 
-        # Return items in the dictionary that have non-empty querysets
-        return {key: value for key, value in related_items.items() if value}
+        # Return specific pages for each queryset in the dictionary that is not
+        # empty
+        return {
+            key: [v.specific for v in value]
```
chosak: Is there some way of using [`.specific()`] on the querysets that are pulled back above, to reduce the number of database queries? As written this will do one DB lookup for each page that is returned in the queryset.

willbarton: I just completely overlooked `specific()` on the Wagtail `PageQueryset`! Changing to that now.

willbarton: Fixed in ef9cfa9!

PR #5130, opened 2019-07-23
Stop storing ZIP codes as integers

higs4281:
Our EventPage model stored ZIP codes as an IntegerField, which removed
significant leading zeros. This creates a string version and restores
missing zeros.

The migration doesn't need a user because the new field doesn't generate
a page revision, so the page doesn't need to be published for the fix
to show up.

## Testing
* Pull in recent prod data
* Check this page locally to see a bad ZIP code: <http://localhost:8000/about-us/events/archive-past-events/field-hearing-on-arbitration/>
* Run the migration in this branch with `./cfgov/manage.py migrate`
* Recheck the page above. It should have a 5-digit ZIP code.


willbarton:
@higs4281 do you want to make a migration to remove the `venue.zip` field entirely after the data migration?

higs4281:
> Works as advertised! I do think we could remove the `venue_zip` field entirely.

Removed!

PR #5133, opened 2019-07-23
add django-dotenv, and compensate for the removal of drama-free-django runtime features

rosskarchner:
These changes compensate for the removal of the "runtime" customizations in Drama Free Django, proposed [here](https://github.com/cfpb/drama-free-django/pull/25). This means that cfgov-refresh needs to do a *little* more work:

- dfd will no longer populate environment variables from a JSON file. To deal with that, this PR adds [django-dotenv](https://github.com/jpadilla/django-dotenv) **and** directly supports the old 'environment.json' files, so that they continue to work while we make the transition.
- dfd will no longer add 'static.in' subdirectories to the STATICFILES_DIRS list, so that functionality has been added to cfgov.settings.base, triggered by a 'DJANGO_STATICFILES_IN' environment variable. You can see that in action in docker/drama-free-django/_build.sh
- the scripts in  docker/drama-free-django/ have been adapted to these changes.

## Additions

- django-dotenv
- DJANGO_STATICFILES_IN functionality in cfgov.settings.base

## Testing

1. `docker/drama-free-django/build.sh && docker/drama-free-django/test.sh` should work as expected

## Todos

- [ ] Merge https://github.com/cfpb/drama-free-django/pull/25
- [ ] docker/drama-free-django/_build.sh  is currently set to install drama-free-django's magic-removal-branch. That should be fixed before merging.
- [ ] The ansible side of this effort is not quite done yet.

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


cfgov/cfgov/wsgi.py
```diff
@@ -10,6 +10,24 @@
 import os
 
 
+import dotenv
+
+this_dir = os.path.dirname(os.path.abspath(__file__))
+
+
+def initialize_environment():
+    environment_candidates = [
```
hkeeler: Why do we need these 3 possible options?  Seems like a chance to have a `.env` in more than one place, and mistake which one is actually being used.  Seems worth of a comment and/or some docs describing when to use these different locations.

cfgov/cfgov/wsgi.py
```diff
@@ -10,6 +10,24 @@
...
+        os.path.join(this_dir, '../.env'),
+        os.path.join(this_dir, '../../.env'),
+        os.path.join(os.getcwd(), '.env')
+    ]
+
+    for candidate in environment_candidates:
+        if os.path.exists(candidate):
```
hkeeler: What about when none of the candidates are found?  Is that OK?  In a Docker world, I'm guessing you wouldn't use a `.env` file at all, and would just pass in envvars right to the container?

Perhaps a log statement saying which `.env` file is being used, or none at all?

cfgov/manage.py
```diff
@@ -2,8 +2,26 @@
 import os
 import sys
 
+import dotenv
+
+this_dir = os.path.dirname(os.path.abspath(__file__))
+
+
+def initialize_environment():
```
hkeeler: Same questions as in `wsgi.py`.

docker/drama-free-django/_test.sh
```diff
@@ -58,7 +60,8 @@ grep -Eho \
     | xargs touch
 
 # Now that we're sure that we have webfont files, we can test collectstatic.
-./venv/bin/django-admin collectstatic
+#./venv/bin/django-admin collectstatic
```
hkeeler: Probably don't need this commented out code here.

docker/drama-free-django/_build.sh
```diff
@@ -29,7 +29,7 @@ yum install -y gcc git python27
 source /opt/rh/python27/enable
 
 pip install -U pip
-pip install -U git+https://github.com/cfpb/drama-free-django.git
+pip install -U git+https://github.com/cfpb/drama-free-django.git@magic-removal-branch
```
hkeeler: Once https://github.com/cfpb/drama-free-django/pull/25 is merged, don't forget to revert this.  :)

cfgov/cfgov/settings/base.py
```diff
@@ -264,6 +265,11 @@
     PROJECT_ROOT.child('templates', 'wagtailadmin')
 ]
 
+if 'DJANGO_STATICFILES_IN' in os.environ:
+    pattern = os.path.join(
+        os.environ['DJANGO_STATICFILES_IN'],
```
chosak: What's the argument over making this configurable vs. just hardcoding use of our current `static.in` as the one place we always place these files? We already document this location, for example [here](https://cfpb.github.io/cfgov-refresh/installation/#webfonts).

cfgov/cfgov/wsgi.py
chosak: Agree, it'd be nice to standardize on one location for .env. Can we standardize on using the repo root (`../.env`), which would also work out to the DFD root during deployments?

cfgov/cfgov/wsgi.py
```diff
@@ -10,6 +10,24 @@
...
+                dotenv.read_dotenv(candidate)
```
chosak:
```suggestion
            dotenv.read_dotenv(candidate)
```

cfgov/cfgov/wsgi.py
```diff
@@ -10,6 +10,24 @@
...
+                return
```
chosak:
```suggestion
            return
```

cfgov/cfgov/settings/base.py
rosskarchner: right now, I don't think the cfgov-refresh version of static.in gets included in the build artifact?

I guess we *could* make it relative, though--   it will always be a sibling to the cfgov directory.

cfgov/cfgov/wsgi.py
rosskarchner: That makes sense to me.

cfgov/cfgov/settings/base.py
chosak: Is there any reason why not to modify the Jenkins build to use static.in? In my branch modifying the Jenkins job to use the Docker-based build (docker-dfd-build), I use that and it works well. We don't use it in the non-Docker-based build currently.

rosskarchner: I don't see a reason not to do that-- I think they just came about at different times, and for different reasons, and nobody ever connected the dots.

cfgov/cfgov/wsgi.py
rosskarchner: it now only looks for '.env' as a sibling of 'cfgov'

cfgov/cfgov/wsgi.py
rosskarchner: now, instead of searching for candidates, it tries to read ../../env. If it isn't there, the [dotenv library will issue a warning](https://github.com/jpadilla/django-dotenv/blob/master/dotenv.py#L64)

cfgov/manage.py
rosskarchner: addressed above

docker/drama-free-django/_test.sh
rosskarchner: fixed!

cfgov/cfgov/wsgi.py
rosskarchner: I've removed that whole function

chosak:
I note your comment:

> The ansible side of this effort is not quite done yet.

Would it make sense to push back loading of .env and instead alter this PR to load from environment.json, essentially pulling in [this code](https://github.com/cfpb/drama-free-django/blob/113003e842ab93059b8f15de2828d24051f7e009/no_drama/build_skel/lib/sitecustomize.py#L13) from DFD?

This would accomplish "stop using DFD at runtime" and allow us to proceed more incrementally, working as-is with the current Ansible code that generates environment.json files. In a subsequent PR (or multiple PRs) we could then further alter cfgov-refresh and the Ansible code to instead generate .env files.

rosskarchner:
> This would accomplish "stop using DFD at runtime" and allow us to proceed more incrementally, working as-is with the current Ansible code that generates environment.json files. In a subsequent PR (or multiple PRs) we could then further alter cfgov-refresh and the Ansible code to instead generate .env files.

That's an interesting idea-- I've been pondering a different approach to breaking the dependencies (having ansible deploy both the json and the .env)-- but since this exists now and the ansible changes don't yet, your approach makes more sense.

rosskarchner:
@chosak I just pushed an implementation of that

chosak:
@rosskarchner, thanks, I'll take another look at this PR.

One thing: don't we need to make at least some Ansible changes before this can go live, e.g. setting the [`DJANGO_STATIC_ROOT`](https://github.com/cfpb/cfgov-refresh/blob/49aa3e23981e983db99f1788d24e542a390a72d1/cfgov/cfgov/settings/base.py#L244) environment variable to the right location?

rosskarchner:
@chosak what do you think about changing this default, which I don't think we actually use anywhere (outside of Docker experiments, maybe) https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/base.py#L244

chosak:
@rosskarchner that seems sensible to me, you mean, to `/srv/cfgov/static`? That seems like it'd be good to add to this PR.

rosskarchner:
Yeah, that's what I meant

cfgov/cfgov/settings/base.py
```diff
@@ -264,6 +265,8 @@
     PROJECT_ROOT.child('templates', 'wagtailadmin')
 ]
 
+static_in = REPOSITORY_ROOT.child('static.in')
+STATICFILES_DIRS += [str(d) for d in static_in.listdir(filter=DIRS)]
```
chosak: If you add this logic into the base settings, can you also remove [this duplicative line](https://github.com/cfpb/cfgov-refresh/blob/49aa3e23981e983db99f1788d24e542a390a72d1/cfgov/cfgov/settings/local.py#L16) from local settings?

chosak:
```suggestion
STATICFILES_DIRS += map(str, static_in.listdir(filter=DIRS))
```

cfgov/cfgov/wsgi.py
```diff
@@ -23,6 +29,17 @@ def initialize_new_relic():
         newrelic.agent.initialize(new_relic_config_file)
 
 
+envfile_path = os.path.join(this_dir, '../../.env'),
+environmentdotjson_path = os.path.join(this_dir, '../../environment.json')
+
+if os.path.exists(envfile_path):
+    dotenv.read_dotenv(envfile_path)
```
chosak: Do we want to use `override=True` here (see [docs](https://github.com/jpadilla/django-dotenv#usage))? That would make the .env file override any existing variables already in the environment. I think we do want that behavior, and that's how environment.json works now.

cfgov/manage.py
```diff
@@ -1,11 +1,24 @@
...
 
+import dotenv
+
 
 if __name__ == '__main__':
+    this_dir = os.path.dirname(os.path.abspath(__file__))
+    envfile_path = os.path.join(this_dir, '../.env')
+    environmentdotjson_path = os.path.join(this_dir, '../environment.json')
+    if os.path.exists(envfile_path):
+        dotenv.read_dotenv(envfile_path)
```
chosak: See above comment re `override=True`.

docker/drama-free-django/_test.sh
```diff
@@ -45,6 +45,8 @@ export DJANGO_SETTINGS_MODULE=cfgov.settings.production
 # commands do nothing.
 mkdir -p static.in/0/fonts
 
+export DJANGO_STATICFILES_IN=/tmp/current/static.in
```
chosak: Is this a legacy addition that can now be removed since we're using local static.in?

cfgov/cfgov/wsgi.py
rosskarchner: I'd say I'm *mildly* uncomfortable with that-- when I read about it, it seemed like an elegant way to recognize that if a variable has been set through some other mechanism, it reflects somebody's intent.

That said, I don't feel too strongly. I'll go ahead and make the change.

chosak:
One other thought about this PR: if you were to remove [these changes](https://github.com/cfpb/cfgov-refresh/pull/5133/files#diff-0f4009d2f136f6ab763f9dda4871ae37) in _build.sh, and [these changes](https://github.com/cfpb/cfgov-refresh/pull/5133/files#diff-300249821afbbfcb945e6fbe141d0bb8) in _test.sh, could this PR be merged as-is and have things continue to work using the existing master version of drama-free-django?

Essentially, the variables would get loaded again, with no change. That'd let us move ahead with this change even while the DFD changes are still under review.

Another thing: the docs [currently suggest using autoenv to load the variables, [here](https://cfpb.github.io/cfgov-refresh/installation/#autoenv-module) and [here](https://cfpb.github.io/cfgov-refresh/installation/#set-up-your-environment). Can we change that if we want to move to auto-loading?

rosskarchner:
> This is only an annoyance but it sure feels like it could be a real annoyance for people running management commands constantly. Is there anything we can do about this?

Do you know what that's for?

We could:
- find a way to eliminate that logic -- django-dotenv doesn't interpret bash files, it expects every line to just be a variable.
- preprocess the .env file, reading it into memory, stripping out the invalid lines, and then using `parse_dotenv` instead of `read_dotenv`.

chosak:
> Do you know what that's for?

Yes. It's because when running using a local virtualenv, the database host and Elasticsearch host are both `localhost`, but in the existing docker-compose setup, these are `postgres` and `elasticsearch`, respectively. Perhaps there's some better way of eliminating that logic, like defaulting them to `localhost` but letting docker-compose override them?

wpears:
Those commands (specifically activate-virtualenv.sh) could just be moved into a separate file that gets invoked. `.env` was just being overloaded and I agree that it's probably better to keep it as a pure envvar file.

Also the lines at the bottom of a file are overrides for local development. Seems like those could be made the default and a modified envvar file could be injected during deploy. (or they could just be included in the separate script that runs activate-virtualenv)

rosskarchner:
> One other thought about this PR: if you were to remove [these changes](https://github.com/cfpb/cfgov-refresh/pull/5133/files#diff-0f4009d2f136f6ab763f9dda4871ae37) in _build.sh, and [these changes](https://github.com/cfpb/cfgov-refresh/pull/5133/files#diff-300249821afbbfcb945e6fbe141d0bb8) in _test.sh, could this PR be merged as-is and have things continue to work using the existing master version of drama-free-django?
> 
> Essentially, the variables would get loaded again, with no change. That'd let us move ahead with this change even while the DFD changes are still under review.

One wrinkle:  these changes don't help with `/venv/bin/django-admin collectstatic`

That's kind of the reason my [initial inclination](https://github.com/cfpb/drama-free-django/pull/24) was a seperate DFD pull request, just for adding the manage.py wrapper (that PR would need to be updated to use the runpy.run_path approach)

> 
> Another thing: the docs [currently suggest using autoenv to load the variables, [here](https://cfpb.github.io/cfgov-refresh/installation/#autoenv-module) and [here](https://cfpb.github.io/cfgov-refresh/installation/#set-up-your-environment). Can we change that if we want to move to auto-loading?

I think so?

chosak:
> One wrinkle: these changes don't help with /venv/bin/django-admin collectstatic

Hm. Well, frankly, the original thought behind adding the collectstatic call to _test.sh was because I wanted to test collectstatic on Travis (#5110). But because of some bugginess of that approach, and a hesitancy to add it as a potential blocker to all cfgov-refresh PRs, that got rolled back (#5119). So that code isn't being invoked by anything right now.

So if it would make things simpler to just remove that call to collectstatic, maybe just for now, we could just do that here. I'll also take a look at your open DFD PR as well, of course.

rosskarchner:
While chipping away at getting this working under mod_wsgi, I discovered that variables set via Apache's SetEnv were not available to the wsgi.py script.  After tinkering with [some work-arounds](https://stackoverflow.com/a/16291949/104365), I decide that a simpler solution was to just add a new, production WSGI file, that unambiguously sets DJANGO_SETTINGS_MODULE to cfgov.settings.production.  Deployment will need to be updated to use cfgov/cfgov/production_wsgi.py. We are already working on changing that, since DFD will no longer include the current /current/wsgi.py

cfgov/cfgov/wsgi.py
rosskarchner: I think I found the reason no to use override=True: if .env takes precedence, then we can't set environment variables in docker-compose.

hkeeler: >I think I found the reason not to use override=True: if .env takes precedence, then we can't set environment variables in docker-compose.

docker-compose itself has an [`env_file`](https://docs.docker.com/compose/environment-variables/#the-env_file-configuration-option) attribute, and it supports the same `KEY=value` format as django-dotenv.  So, maybe the better option for compose, is to **not** load the `.env` file in-container, and let compose inject those envvars itself?

We'll have to do something similar in a production environment.  We don't want to bake in env-specific `.env` files at image build time, and we won't be able to map in a `.env` at runtime, so the app will have to support just having those envvars already there at container start.  I _think_ that's fine in the current logic, since it doesn't fail if neither `.env` nor `environment.json` are present.

Also, is the plan to update [`.env_SAMPLE`](https://github.com/cfpb/cfgov-refresh/pull/5133/files#diff-eb951766c3e0a8d12740b6236ded5a9f) to the `KEY=value` style syntax?  Or do we need it to be an actual shell script for other purposes?

rosskarchner: I've removed override=True, so variables set via Docker, Linux, Compose, or tox will always take precedence.

PR #1909, opened 2016-04-15
Removing pagination validation client side code

sebworks:
This is part of a move toward using the browsers native data types and removing client side validation code; It is makes sense to do so where browsers have done a decent job standardizing the validation experience.

## Removals

- Removing pagination validation client side code and reference from `common.js`.

## Changes

- Adding a error handling block to `query.py` to handle non int values. 
- Adding attributes to the input to invoke the numeric keyboard in IOS.

## Testing

- Run `gulp build` and visit `http://localhost:8000/about-us/blog/?page=yyyy`.

## Review

- @KimberlyMunoz 
- @anselmbradford 
- @jimmynotjim 
- @kurtw 

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #5135, opened 2019-07-23
Removes unneeded attributes on pagination

anselmbradford:
[This PR](https://github.com/cfpb/cfgov-refresh/pull/1909) added `pattern="[0-9]*"` and `inputmode="numeric"` to pagination inputs. However, this fails HTML validation. Additionally, `type="number"` should infer the intent both these attributes are providing.

> When inputmode is unspecified (or is in a state not supported by the user agent), the user agent should determine the default virtual keyboard to be shown. Contextual information such as the input type or pattern attributes should be used to determine which type of virtual keyboard should be presented to the user.

Screen Shot 2019-07-23 at 3 07 12 PM

Screen Shot 2019-07-23 at 3 06 44 PM

## Removals

- `pattern` and `inputmode` attributes from the pagination input.

## Testing

1. Visit https://validator.w3.org/nu/?doc=https%3A%2F%2Fwww.consumerfinance.gov%2Factivity-log%2F (Note: there are other errors in the style tags and featured content modules).

### DIFF
--- a/cfgov/agreements/jinja2/agreements/search.html
+++ b/cfgov/agreements/jinja2/agreements/search.html
@@ -96,8 +96,6 @@
                                         type="number"
                                         min="1"
                                         max="999"
-                                        pattern="[0-9]*"
-                                        inputmode="numeric"
                                         value="{{ page.number }}">
                                 of {{ page.paginator.num_pages }}
                             </label>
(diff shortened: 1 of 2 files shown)


Scotchester:
Looks like we have to add back the `pattern` attribute to get the finger-friendly numeric keyboard on iOS and Android, and swallow the fact that this will continue to report as a validation failure.

Scotchester:
Also, please update the markup pattern in capital-framework.

anselmbradford:
Closed as a wontfix

PR #5137, opened 2019-07-23
Update to capital-framework@11.0.1 | cf-tables: Migrate table vars

anselmbradford:
Migrates cf-table vars added in https://github.com/cfpb/capital-framework/pull/951

## Removals

- Remove `gulp-bless`

## Changes

- Update to capital-framework@11.0.1
- Update to gulp-terser@1.2.0
- Update to @babel/core@7.5.5
- Update to @babel/preset-env@7.5.5
- `@thead-text` migrated to `@table-head-text`
- `@thead-bg` migrated to `@table-head-bg`


Scotchester:
Pushed a commit that should resolve the Bless issue.

A weird thing happened with Yarn, though. Instead of running `yarn remove gulp-bless` the first time, I accidentally ran `yarn remove gulp-less`, uninstalling Less, not Bless. When I ran `yarn add less@4.0.1` to put it back, it did not restore the packages that it had deleted from the offline cache, so I went through and manually restored them, but I only noticed this _after_ I had then run `yarn remove gulp-bless`, so I had to look for each of the packages that had been removed from the offline cache in `yarn.lock` and make sure they weren't still needed.

I think I restored all the ones that needed to be kept, but just in case, could you double-check for me that none of the removed tarballs are still needed?

It worked to remove `node_modules`, re-run `yarn` and `gulp`, so I think everything should be OK.

anselmbradford:
@Scotchester This was failing Travis for some reason, so I've re-generated the cache and pushed an update. This looks good to me otherwise!

anselmbradford:
Note: gulp-bless is removed here because of https://github.com/BlessCSS/bless/issues/109

anselmbradford:
This likely will fix some/all of the lodash security vulnerabilities reported in https://github.com/cfpb/cfgov-refresh/pull/5133

PR #5138, opened 2019-07-23
Ensure matching of original link string when parsing links

willbarton:
This change ensures that we match the original link string when we replace it by using regular expressions to extract it rather than BeautifulSoup.

Originally we were using BeautifulSoup to find all `a` tags, and then rendering the BeautifulSoup tag back to HTML and replacing that string with our marked-up replacement.

The problem is that BeautifulSoup doesn't guaratee attribute order, so when it's time to replace the original string the BeautifulSoup-rendered copy wouldn't necessary exist in the HTML string because the attribute order might've changed.

This change gets around that problem by using regular expresisons to match the original string and not engaging BeautifulSoup until it's time to add the markup.

To see this in action, take a look at the "Additional Reports" on http://localhost:8000/data-research/hmda/:
image

In production these do not get the external link icon class (https://www.consumerfinance.gov/data-research/hmda/):

image

## Checklist


cfgov/core/utils.py
```diff
@@ -24,6 +24,9 @@
 
 LINK_ICON_TEXT_CLASSES = 'a-link_text'
 
+A_TAG = re.compile(r'<a[^>]+>.+?</a>')
+IMG_TAG = re.compile(r'<img[^>]+>')
```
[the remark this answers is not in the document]
willbarton: This is a regular expression to match image tags.

cfgov/core/utils.py
```diff
@@ -24,6 +23,9 @@
 
 LINK_ICON_TEXT_CLASSES = 'a-link_text'
 
+A_TAG = re.compile(r'<a[^>]*>.+?</a>')
+IMG_TAG = re.compile(r'<img[^>]*>')
```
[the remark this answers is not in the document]
willbarton: This is a regular expression to match image tags, not an image tag itself.

cfgov/core/tests/test_middleware.py
```diff
@@ -122,3 +122,15 @@ def test_non_default_encoding(self):
...
         self.assertEqual(parsed, s)
+
+    def test_external_link_with_attribute(self):
+        s = '<a href="https://somewhere/foo" data-thing="something">Link</a>'
+        output = parse_links(s)
+        self.assertIn('external-site', output)
+        self.assertIn('cf-icon-svg', output)
+
+    def test_external_link_with_img(self):
+        s = '<a href="https://somewhere/foo"><img src="some.png"></a>'
```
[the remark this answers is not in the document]
willbarton: This is a unittest.

cfgov/core/utils.py
```diff
@@ -67,11 +69,10 @@ def format_file_size(bytecount, suffix='B'):
...
 
 
-def get_link_tags(soup):
+def get_link_tags(html):
     tags = []
-    for a in soup.find_all('a', href=True):
-        if not is_image_tag(a):
-            tags.append(a)
+    for a in A_TAG.findall(html):
+        tags.append(a)
```
chosak: Now that the image-related logic has been removed, could this be simplified into a one-liner?

```py
return A_TAG.findall(html)
```

cfgov/core/utils.py
```diff
@@ -128,6 +134,9 @@ def add_link_markup(tag):
         tag.contents.append(BeautifulSoup(svg_icon(icon), 'html.parser'))
         return str(tag)
 
+    elif is_image_tag(tag):
```
chosak: I might be misreading this but, as written, this seems like it won't properly skip adding an icon to `<a href="external or download"></a>` links, because the `elif` here is against the `if icon` check, not the other `elif`s above. See my note about the test above: what is the desired behavior here?

Could you also update the `add_link_markup` docstring above (and maybe add a one-line comment here) to explain why this branch is here?

cfgov/core/tests/test_middleware.py
```diff
@@ -122,3 +122,15 @@ def test_non_default_encoding(self):
...
+        output = parse_links(s)
+        self.assertIn('external-site', output)
+        self.assertIn('cf-icon-svg', output)
```
chosak: Is this actually the desired behavior, to add a link icon to image links?

cfgov/core/utils.py
willbarton: Yes, it could!

cfgov/core/tests/test_middleware.py
willbarton: You're right — it's probably not.

cfgov/core/tests/test_middleware.py
```diff
@@ -122,3 +123,33 @@ def test_non_default_encoding(self):
...
         self.assertEqual(parsed, s)
+
+    def test_external_link_with_attribute(self):
+        s = '<a href="https://somewhere/foo" data-thing="something">Link</a>'
+        output = parse_links(s)
+        self.assertIn('external-site', output)
+        self.assertIn('cf-icon-svg', output)
+
+    def test_external_link_with_img(self):
+        s = '<a href="https://somewhere/foo"><img src="some.png"></a>'
```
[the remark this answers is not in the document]
willbarton: This is a unittest.

cfgov/core/tests/test_middleware.py
willbarton: Fixed in c654196.

cfgov/core/utils.py
willbarton: Ok, that's fixed in c654196.

willbarton:
I've adjusted the regular expression so that it will match other external links on the page, including the USA.gov and OIG links at the bottom:

image

cfgov/core/utils.py
```diff
@@ -24,6 +24,9 @@
 
 LINK_ICON_TEXT_CLASSES = 'a-link_text'
 
+A_TAG = re.compile(r'<a [^>]*?>.+?(?=</a>)</a>(?s)(?i)')
+IMG_TAG = re.compile(r'<img[^>]*>(?i)')
```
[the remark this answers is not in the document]
willbarton: This is a regular expression.

PR #5141, opened 2019-07-24
Start new Mouseflow capture when first expandable in a group is clicked

Scotchester:
Background: https://GHE/CFGOV/platform/issues/2908

Our analytics team would like to start a new Mouseflow capture once the first expandable in an expandable group is clicked, so that the screenshot shows more of the page, which provided better heatmap data.

Console logging left in for now to aid in testing. **On hold** until this can be tested in Mouseflow and logging statements are removed.

## Additions

- Event listener added to all expandable targets in an expandable group that triggers a new Mouseflow capture when the first one is clicked (listeners are removed after that first click)
- `removeEventListenerFromSelector` and `removeEventListenerFromElem` functions added along side their `add` counterparts in analytics module utils

## Testing

1. Pull branch
1. `gulp scripts`
1. Load a page with an expandable group, like http://localhost:8000/consumer-tools/getting-an-auto-loan/plan-to-shop-for-your-auto-loan/
1. Open the console
1. Click an expandable and see the logging statement
1. Click another one (or the same one) and see no logging statements (as the event listeners were removed)

## Notes

- The add/remove event listener utils aren't really analytics-specific. Should we move them out of that module into a more generic location?

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/cfgov/unprocessed/apps/analytics-gtm/js/util/analytics-util.js
+++ b/cfgov/unprocessed/apps/analytics-gtm/js/util/analytics-util.js
@@ -26,6 +26,33 @@ function addEventListenerToElem( elem, event, callback ) {
 }
 
 /**
+ * Query a selector and remove listeners from returned elements.
+ * @param {string} selector - A dom selector.
+ * @param {string} event - An event string, probably a "MouseEvent."
+ * @param {Function} callback - The event handler.
+ */
+function removeEventListenerFromSelector( selector, event, callback ) {
+  const elems = document.querySelectorAll( selector );
+  for ( let i = 0, len = elems.length; i < len; i++ ) {
+    removeEventListenerFromElem( elems[i], event, callback );
+  }
+}
+
+/**
+ * Check if an element exists on the page, and if it does, remove listeners.
+ * @param {[type]}   elem     [description]
+ * @param {[type]}   event    [description]
+ * @param {Function} callback [description]
+ */
+function removeEventListenerFromElem( elem, event, callback ) {
+  if ( elem ) {
+    elem.removeEventListener( event, callback );
+  } else {
+    analyticsLog( `${ elem } doesn't exist!` );
+  }
+}
+
+/**
  * Log a message to the console if the `debug-gtm` URL parameter is set.
  * @param {string} msg - Message to load to the console.
  */
@@ -113,6 +140,8 @@ function getQueryParameter( key ) {
 module.exports = {
   addEventListenerToSelector,
   addEventListenerToElem,
+  removeEventListenerFromSelector,
+  removeEventListenerFromElem,
   analyticsLog,
   Delay,
   getQueryParameter,
--- a/cfgov/unprocessed/js/routes/on-demand/expandable-group.js
+++ b/cfgov/unprocessed/js/routes/on-demand/expandable-group.js
@@ -2,4 +2,45 @@
    Scripts for Expandable Group organism.
    ========================================================================== */
 
-require( 'cf-expandables/src/Expandable' ).init();
+import {
+  addEventListenerToSelector,
+  analyticsLog,
+  removeEventListenerFromSelector
+} from '../../../apps/analytics-gtm/js/util/analytics-util';
+
+import Expandable from 'cf-expandables/src/Expandable';
+Expandable.init();
+
+
+/**
+ * Tells Mouseflow to stop capturing, if it already is, and start a new capture.
+ */
+function stopStartMouseflow() {
+  console.log( 'stopStartMouseflow' );
+  if ( window.mouseflow ) {
+    // Stop any in-progress heatmap capturing.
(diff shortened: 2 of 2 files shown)


cfgov/unprocessed/js/routes/on-demand/expandable-group.js
```diff
@@ -2,4 +2,45 @@
...
+
+import Expandable from 'cf-expandables/src/Expandable';
+Expandable.init();
+
+
+/**
+ * Tells Mouseflow to stop capturing, if it already is, and start a new capture.
+ */
+function stopStartMouseflow() {
+  console.log( 'stopStartMouseflow' );
```
anselmbradford: Why not use `analyticsLog( 'stopStartMouseflow' );` here and elsewhere where there are console.log statements?

anselmbradford: Ohh I guess this is still a draft.

cfgov/unprocessed/apps/analytics-gtm/js/util/analytics-util.js
```diff
@@ -25,6 +25,33 @@ function addEventListenerToElem( elem, event, callback ) {
...
+  const elems = document.querySelectorAll( selector );
+  for ( let i = 0, len = elems.length; i < len; i++ ) {
+    removeEventListenerFromElem( elems[i], event, callback );
+  }
+}
+
+/**
+ * Check if an element exists on the page, and if it does, remove listeners.
+ * @param {[type]}   elem     [description]
+ * @param {[type]}   event    [description]
```
anselmbradford: event and callback can be the same descriptions as those set in `removeEventListenerFromSelector`.

anselmbradford:
Instead of creating `cfgov/unprocessed/js/routes/on-demand/expandable-group.js`, we should probably move that to a file in `https://github.com/cfpb/cfgov-refresh/tree/master/cfgov/unprocessed/apps/analytics-gtm/js` and then tell Kelley to create a tag in GTM for all pages that runs on all pages, as is done for `accessibility-interactions-listeners.js` and `tab-navigation-tracking.js`. (For testing, you could actually add your code into either of those, which already run on all pages.)

cfgov/unprocessed/js/routes/on-demand/expandable-group.js
Scotchester: Yeah, I just left those in for ease of testing locally that the code paths are doing what we want, and then we'll test with Mouseflow itself next week.

Scotchester:
> Instead of creating `cfgov/unprocessed/js/routes/on-demand/expandable-group.js`, we should probably move that to a file in `https://github.com/cfpb/cfgov-refresh/tree/master/cfgov/unprocessed/apps/analytics-gtm/js` and then tell Kelley to create a tag in GTM for all pages that runs on all pages, as is done for `accessibility-interactions-listeners.js` and `tab-navigation-tracking.js`. (For testing, you could actually add your code into either of those, which already run on all pages.)

`cfgov/unprocessed/js/routes/on-demand/expandable-group.js` already exists to trigger initialization of expandables on pages that have expandable groups. If we can isolate this code to such pages, shouldn't we do that, rather than run it on all pages?

cfgov/unprocessed/js/routes/on-demand/expandable-group.js
```diff
@@ -2,4 +2,45 @@
...
+  removeEventListenerFromSelector(
+    Expandable.prototype.ui.target,
+    'click',
+    handleExpandable
+  );
+  const waitForTransition = window.setTimeout( stopStartMouseflow, 500 );
+}
+
+addEventListenerToSelector(
+  Expandable.prototype.ui.target,
```
Scotchester: It occurs to me that this selector should be made more specific to only expandable targets _within an expandable group_, because we don't necessarily want this happening for other one-off expandables, like the menu.

anselmbradford:
> cfgov/unprocessed/js/routes/on-demand/expandable-group.js already exists to trigger initialization of expandables on pages that have expandable groups. If we can isolate this code to such pages, shouldn't we do that, rather than run it on all pages?

That makes sense. We may want to move the mouseflow stuff to an analytics or mouseflow module though and call methods on it in `expandable-group.js`. Like would it make sense to add `stopStartMouseflow` to https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/js/modules/Analytics.js?

Scotchester:
Closing per comments in GHE issue.

PR #5143, opened 2019-07-24
Redo smoke tests to include the current top site URLs

higs4281:
We want to automate the list of URLs as much as possible, so we're pursuing
an automated top-20 list and a compiled list of megamenu links.

We'll pull the full list from a json file with a standing S3 URI, which we can update programmatically. The S3 json file is current as of ~~July 2019~~ January 2020.

If there's any connectivity issues or the S3 file is otherwise not available, we'll fall back to a default 
list current as of ~~July 2019~~ January 2020.


cfgov/scripts/tests/test_smoke_tests.py
```diff
@@ -102,10 +134,17 @@ def test_allowed_timeouts(self, mock_get):
...
         http_smoke_test.check_urls('pro1')
-        self.assertEqual(mock_get.call_count, len(http_smoke_test.SHORT_RUN))
+        self.assertEqual(
+            mock_get.call_count,
+            len(http_smoke_test.SHORT_RUN)
+        )
 
     @mock.patch('scripts.http_smoke_test.requests.get')
-    def test_http_fail_request_error(self, mock_get):
+    @mock.patch('scripts.http_smoke_test.get_full_list')
```
schbetsy: Looks like `mock_list` isn't needed for this one.

cfgov/scripts/tests/test_smoke_tests.py
```diff
@@ -36,55 +36,87 @@ def test_home_page_test_failure(self, mock_get):
...
 
     @mock.patch('scripts.http_smoke_test.requests.get')
-    def test_http_success_full(self, mock_get):
+    @mock.patch('scripts.http_smoke_test.get_full_list')
+    def test_http_success_full(self, mock_list, mock_get):
+        mock_list.return_value = http_smoke_test.FULL_RUN
         mock_response = mock.Mock()
         mock_response.status_code = 200
         mock_get.return_value = mock_response
         http_smoke_test.check_urls('pro1', full=True)
```
schbetsy: Let's check the return value of the smoke tests, at least on the tests that check the "success" path.

```suggestion
         result = http_smoke_test.check_urls('pro1', full=True)
         self.assertTrue(result)
```

cfgov/scripts/http_smoke_test.py
```diff
@@ -38,6 +38,7 @@
 ALLOWED_TIMEOUTS = 1
 FULL = False
 BASE = 'https://www.consumerfinance.gov'
+S3_URI = 'http://files.consumerfinance.gov.s3.amazonaws.com/build/smoketests/smoketest_urls.json'  # noqa
```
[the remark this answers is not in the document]
higs4281: This is a great idea. I realized the short list wasn't useful to anyone, so I refactored to remove that and add an option for passing a custom URL list.

cfgov/scripts/http_smoke_test.py
```diff
@@ -108,37 +121,46 @@
...
 ]
 
+FULL_RUN = sorted(set(TOP20 + APPS))
+
+
+def get_full_list():
+    try:
+        url_data = requests.get(S3_URI).json()
+    except Exception as e:
+        logger.info(
```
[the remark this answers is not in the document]
higs4281: Agreed.

cfgov/scripts/http_smoke_test.py
```diff
@@ -108,37 +121,46 @@
...
+    '/your-story/',
 ]
 
+FULL_RUN = sorted(set(TOP20 + APPS))
+
+
+def get_full_list():
+    try:
+        url_data = requests.get(S3_URI).json()
+    except Exception as e:
```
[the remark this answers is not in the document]
higs4281: I was anticipating possible network or s3 outages that might be hard to anticipate, and I think our fallback list of URLs is robust enough that failing to get the S3 list is not crucial.

Since this job can stop deployments, I thought we should err on the side of letting a potentially temporary S3 issue pass rather than stop deployments to debug S3.

cfgov/scripts/tests/test_smoke_tests.py
higs4281: True! Refactored.

cfgov/scripts/tests/test_smoke_tests.py
higs4281: Yes!

higs4281:
@schbetsy Thank you for your careful review.

I've refactored a bit to remove the unused SHORT_RUN routine and add the ability to pass a custom list of URLs to test. I also updated out most-popular URLs list to be current with Jan. 2, 2020, and bump the length to 25, as we did in hubcap/wiki.

cfgov/scripts/http_smoke_test.py
```diff
@@ -108,39 +117,61 @@
...
     '/about-us/innovation/',
-    '/about-us/contact-us/',
-    '/eregulations/',
-    '/eregulations/1026',
+    '/activity-log/',
+    '/your-story/',
 ]
 
+# call `set` on the combined list to weed out dupes
+FULL_RUN = sorted(set(TOP + APPS))
```
schbetsy: Could we change this to something like `FALLBACK_URLS`? I think `FULL_RUN` is a misleading name for this list, since it will only be used in cases where something has gone wrong.

higs4281: Good point.

schbetsy: 👍

higs4281:
@Scotchester I refactored the static routine to unify the URL format with http-smoke-tests and to increase test coverage.

PR #5144, opened 2019-07-25
Specify Python version in Docker via build ARG

hkeeler:
This PR is the first step in creating a production-ready Docker configuration for cfgov-refresh.  Specifically, this PR refactors how we specify the Python version used in our `python2` and `python3` services.  Currently, both services use the same image with both Python 2 and 3 installed.  In some ways this in convenient, but also means you have near-duplicate code that's tied to the Python version.  Going forward, you can always just refer to `python`, and know you're using the version that's right for that specific service.

## Additions

- ~~Automatically (but optionally) perform Django database migrations at container startup.  This was not the primary goal of this change, but made it much easier to test these changes while frequently destroying and rebuilding the whole stack over and over.~~

    ~~This change includes a new `INIT_DB` environment variable, set to `ON` for `python2` and `OFF` for `python3` services.  This was needed because Django migrations fail when multiple instances are running at the same time.~~

    ~~This _could_ be moved into a follow-up PR if we feel it violates the _no scope creep_ rule.  😬~~

- Adds a new `APP_HOME` environment variable, hardcoded to `/src/cfgov-refresh` in the `Dockerfile`.  I found myself referencing this directory a lot, so seemed useful.

## Removals

None

## Changes

- Refactors main `Dockerfile` to specify the Python version via a build [`ARG`](https://docs.docker.com/engine/reference/builder/#arg), resulting in two very similar but distinct containers, each using just the version of Python they need.
- Switches Python installs from default CentOS `python` for 2.7 and EPEL-installed `python36` to SCL-based `python27` and `rh-python36`.  This makes the installs very consistent, and has the added benefit of tools like `pip` and `python-devel` coming along for the ride.
- Introduces Docker multi-stage builds.  This initially just segregates the local dev-specific bits from the base dependencies needed by all cfgov images.  This is a building block for future changes.  See #5051 for details on this model.
- Stops `extend-environment.sh` from adding wildcard expansion string to `PYTHONPATH` when `develop-apps` is empty.

## Testing

1. I have manually tested the `docker-compose` workflow, confirming that both `python2` and `python3` services serve the cfgov site at http://localhost:8000 and http://localhost:8333 as expected.
1. It would be helpful if cfgov-refresh devs could take this for a spin and make sure this setup works for their day-to-day development workflow.

## Todos

- [ ] Update documentation
- [ ] Test satellite apps more
- ~~[ ] Confirm there are no race conditions between when `postgres` container starts and Django migrations are executed.~~
- ~~[ ] Make `INIT_DB` truly optional in `docker-compose.yml`.  It's currently on for python2 and off for python3.~~

## Checklist


hkeeler:
@chosak and I chatted about the _`initial-data.sh` at container startup_ functionality, and we decided to take that out since it'd be a change to current developer workflow, and there's likely other things we need to consider before introducing a change like that.  So, I've pulled that out, and we'll come back to that sometime soon in a future PR.

Dockerfile
```diff
@@ -1,30 +1,56 @@
...
+    yum -y install \
+        gcc \
+        httpd \
+        mailcap \
+        postgresql10 \
+        postgresql10-devel \
+        which
+
+# Specify SCL-based Python version.
+# Curretnly used options: python27, rh-python36
```
chosak:
```suggestion
# Currently used options: python27, rh-python36
```

docker-entrypoint.sh
```diff
@@ -0,0 +1,20 @@
...
+# NOTE: By using --login, it guarantees /etc/profile is always sourced, unlike the
+# non-login, non-interactive shell you get by default with `docker run`.
+
+set -e
+
+echo "Using '$(python --version 2>&1)' from '$(which python)'"
+
+# FIXME: Could we replace this with a Docker Compose env_file?
+# SEE: https://docs.docker.com/compose/compose-file/#env_file
+if [ -f "${APP_HOME}/.env" ]; then
```
chosak: This doesn't seem to work properly for me. I don't have an `APP_HOME` environment variable in the container. Does the use of [`ENV`](https://docs.docker.com/engine/reference/builder/#env) in the Dockerfile actually persist at container runtime?

wpears: I believe APP_HOME is only set in cfgov-base and what's running via compose is cfgov-develop.

wpears: Then again, I have APP_HOME set after `docker-compose exec python2 bash --login`

chosak: I should have specified, I followed the [documentation](https://cfpb.github.io/cfgov-refresh/running-docker/#access-a-containers-shell) that suggests using `docker exec -it cfgov-refresh_python2_1 bash`.

docker-entrypoint.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash --login
```
wpears: Perhaps we don't need to force the login shell if we're going to manually `source /etc/profile` below? (though I'd assume we'd want to bring lines 9-15 above the `Using python...` echo in that case).

docker-compose.yml
```diff
@@ -9,54 +11,53 @@ services:
...
-            context: ./
-            dockerfile: ./Dockerfile
-        image: python
+        build:
+            context: .
+            dockerfile: Dockerfile
+            target: cfgov-develop
+            args:
+                scl_python_version: python27
         ports:
```
rosskarchner: Any reason for not including the environment variables for connecting to postgres?

hkeeler: I considered it, but since [`.env_SAMPLE` already has `DATABASE_URL`](https://github.com/cfpb/cfgov-refresh/blob/master/.env_SAMPLE#L65) defaulting to the Docker setup, and that file has to be sourced in the container anyway, _and_ it's used by non-Docker setups too, it felt like it'd add to confusion as to why just that one envvar was setup in compose.  In fact, that's why I removed the `ES_HOST` envvar from compose, since it was [duplicated in `.env_SAMPLE`](https://github.com/cfpb/cfgov-refresh/blob/master/.env_SAMPLE#L92) as well.

docker-entrypoint.sh
hkeeler: I added the `source /etc/profile` after doing the `--login` bit, so yeah, that may no longer be necessary.  Let me try pulling that out and see if/how that changes the environment.

hkeeler: Confirmed.  Removed the `--login`, which always felt a little weird.

Dockerfile
```diff
@@ -1,30 +1,56 @@
...
+WORKDIR ${APP_HOME}
+
+# Install common OS packages
+RUN yum -y install \
+        centos-release-scl \
+        epel-release \
+        https://download.postgresql.org/pub/repos/yum/10/redhat/rhel-7-x86_64/pgdg-centos10-10-2.noarch.rpm && \
+    yum -y install \
+        gcc \
+        httpd \
```
chosak: Is this forward-looking for a time when this container can run Apache as well, or is it needed if we're only still using runserver?

hkeeler: You're right, that can probably be taken out.  Confirming now...

hkeeler: Removed.

docker-entrypoint.sh
hkeeler: Actually, the `--login` _is_ needed.  Without this, the `source /etc/profile` fails.  Unfortunately, it does not give an error message.  I'd like to understand why, but I don't want to hold up the PR any longer, so I just put the `--login` back in for now.

PR #5145, opened 2019-07-25
Move root-required bits of drama-free-django build into Dockerfile

hkeeler:
The current drama-free-django Docker-based build tool (see [`docker/drama-free-django`](https://github.com/cfpb/cfgov-refresh/tree/master/docker/drama-free-django)) works well in most cases, but we've had issues in CI/CD environments where the files written back to the host environment cannot be cleaned up since they're owned by `root`.  This change moves the the `root`-required bits into a new `Dockerfile` where all the tools can be installed, and allows the DFD-related processes to run in the container whichever user you'd like.

## Notes

- When running the container as a user that exists on the host, but not in the container, you may notice a warning similar to:

    ```
    /usr/bin/id: cannot find name for user ID 502
    ```

    This is not anything to worry about.  It simply means the `uid`/`gid` don't match any users/groups setup _in_ the container.

## Checklist


docker/drama-free-django/_build.sh
```diff
@@ -62,13 +49,14 @@ no-drama build "${build_args[@]}"
 echo "{}" > ./dfd_env.json
 
 # This is used by DFD to set Django's settings.STATIC_ROOT.
-echo '{"static_out": "../../../static"}' > ./dfd_paths.json
+# Q: Why do we need to override the default?
+# echo '{"static_out": "../static"}' > ./dfd_paths.json
```
chosak: This is needed because [the default drama-free-django behavior](https://github.com/cfpb/drama-free-django/blob/113003e842ab93059b8f15de2828d24051f7e009/no_drama/build_skel/lib/dfd.py#L17) is to set its `static_out` variable to `'static'`. This is used to set Django [`STATIC_ROOT`](https://docs.djangoproject.com/en/2.2/ref/settings/#std:setting-STATIC_ROOT), which specifies where collectstatic puts its files. The default behavior would have this command collect staticfiles to a `static` relative path within the DFD deploy.

We instead want these files to go to `'../../../static/'`, in practice going from, say, `/srv/cfgov/versions/20190712095508/current` to `/srv/cfgov/static`.

@rosskarchner's open PR #5133 currently includes [a change](https://github.com/cfpb/cfgov-refresh/pull/5133/files#diff-ac27c5919c46548eada7a7388e726c59L244) that would set the default static root to `/srv/cfgov/static` as desired, but that'll only work once https://github.com/cfpb/drama-free-django/pull/25 or something like it removes the DFD behavior that overwrites the static root at runtime.

docker/drama-free-django/_test.sh
```diff
@@ -16,15 +17,11 @@ if [ ! -d "$artifact_volume" ]; then
...
-# Install runtime requirements.
-yum install -y centos-release-scl
-yum install -y python27
-
-source /opt/rh/python27/enable
 
 # Extract the artifact in /tmp.
-cp "$artifact_volume/$artifact_filename" /tmp
-cd /tmp
+mkdir -p $dfd_test_dir
```
chosak: Not a big deal, but I'm curious why you made a new directory for this. I figured that working in /tmp was fine since the container is ephemeral.

docker/drama-free-django/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+# Must be world writable since alternate uid:gid may be patched in at `docker run` time.
+RUN mkdir -p ${DFD_DIR} && chmod 777 ${DFD_DIR}
+WORKDIR ${DFD_DIR}
+
+# Install dependencies
+RUN yum install -y centos-release-scl && \
+    curl -sL https://rpm.nodesource.com/setup_10.x | bash - && \
+    curl -sL https://dl.yarnpkg.com/rpm/yarn.repo | tee /etc/yum.repos.d/yarn.repo && \
+    yum install -y ${SCL_PYTHON_VERSION} gcc git nodejs which yarn && \
```
chosak: Is the inclusion of `which` here leftover from debugging?

docker/drama-free-django/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    echo "source scl_source enable ${SCL_PYTHON_VERSION}" > /etc/profile.d/scl_python.sh && \
+    source /etc/profile && \
+    pip install -U pip && \
+    pip install -U git+https://github.com/cfpb/drama-free-django.git
+
+COPY _build.sh _test.sh docker-entrypoint.sh ./
+
+ENTRYPOINT [ "./docker-entrypoint.sh"]
```
chosak:
```suggestion
ENTRYPOINT ["./docker-entrypoint.sh"]
```

docker/drama-free-django/test.sh
```diff
@@ -1,3 +1,11 @@
...
 
-docker run -v `pwd`:/cfgov centos:6 /cfgov/docker/drama-free-django/_test.sh
+set -e
+
+docker build -t cfgov-dfd-builder docker/drama-free-django
+
+docker run \
+  --rm \
+  -u $(id -u):$(id -g) \
+  -v $(pwd):/cfgov:cached \
```
chosak: Why is `:cached` used here in the test script but not in the build script? Is this Mac-specific, as documented [here](https://docs.docker.com/docker-for-mac/osxfs-caching/#cached) -- is it a no-op when run on Linux systems?

docker/drama-free-django/test.sh
```diff
@@ -1,3 +1,11 @@
...
+  cfgov-dfd-builder ./_test.sh
```
chosak: This shouldn't hold up this PR, but, if we are going to maintain this DFD testing capability, I wonder if it's worth entertaining the idea of a distinct Dockerfile just for that purpose. We don't really need everything in the "cfgov-dfd-builder" image just to run the DFD image, and it would be nice to actually determine what is needed. But probably thinking more about that should wait until/if we want to think about migrating Ansible code here.

docker/drama-free-django/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM centos:7
```
chosak: The current versions of the build and test scripts use `centos:6`(also documented [here](https://github.com/cfpb/cfgov-refresh/tree/master/docker/drama-free-django)). I did this deliberately to try to match the current setup. Is there a particular reason to change that version here?

docker/drama-free-django/_build.sh
hkeeler: Makes sense.  I'll change it back.

docker/drama-free-django/_test.sh
hkeeler: The issue was that by having it in `/tmp`, it resulted in directories being created in the `/` root directory, and the non-`root` user didn't have permission to create those directories.  An alternative would be to open up permissions on `/`, but this seems like a better option.

docker/drama-free-django/Dockerfile
hkeeler: Yes, I can remove that.  I find it strange that CentOS doesn't have `which` in its base distribution, but yeah.

docker/drama-free-django/test.sh
hkeeler: Unfortunately, it fails currently on Linux.  I'll remove it.

docker/drama-free-django/test.sh
hkeeler: Yeah, it does feel a little awkward to have the two scripts that both build the same image, but my assumption was that _generally_ when you'd run `test.sh`, you would have run `build.sh` just before it, and so the `docker build` part would be fully cached and be pretty instant, but I didn't want to make that a requirement to running `test.sh`, so it's duplicated  in both scripts.

As for a separate image, yeah we could.  It just seemed like it'd be better to only maintain one Dockerfile that could do both.  But if there's a scenario where we'd be running _just_ `test.sh`, maybe that starts to make more sense.

docker/drama-free-django/Dockerfile
hkeeler: That was not intentional.  Will fix.

hkeeler:
@chosak, I think I've covered most of your comments.
 
> * Is there somewhere sensible to put your note above about the ID warning? Maybe in the [DFD README](https://github.com/cfpb/cfgov-refresh/blob/master/docker/drama-free-django/README.md)? Or are people not even going to notice that?

Seems like a good place.  I added a _Notes_ section with those details.

> * There are a bunch of warnings during the Yarn build:
>   ```
>   warning Skipping preferred cache folder "/.cache/yarn" because it is not writable.
>   warning Selected the next writable cache folder in the list, will be "/tmp/.yarn-cache-501".
>   ```
>   
>   
>   I don't know enough about Yarn to know what this means, but these don't appear during our normal frontend builds. Is this something to be concerned about?

This was occurring because Docker defaults the home directory to `/` for users it doesn't know about, and then `yarn` didn't have permissions to create it's cache in it's preferred location (`/.cache/yarn`).  My preferred solution would be to just not have a cache, since it doesn't make sense in a Docker build context, but `yarn` doesn't support that [because reasons](https://github.com/yarnpkg/rfcs/pull/53).  So, instead I just added a `HOME` envvar right into the `Dockerfile`, pointed to a directory it _could_ write to, and that made it happy.

> * We get similar warnings during the DFD build about pip:
>   ```
>   WARNING: The directory '/.cache/pip/http' or its parent directory is not owned by the current user and the cache has been disabled. Please check the permissions and owner of that directory. If executing pip with sudo, you may want sudo's -H flag.
>   ```

Now `pip` does support disabling cache with either the `--no-cache-dir` arg or the `PIP_NO_CACHE_DIR` envvar.  I went the `PIP_NO_CACHE_DIR` route so it'd apply to _all_ the `pip install`s...and after some wrangling, it works!  No more warnings package manager warnings!

**Fun fact:** `PIP_NO_CACHE_DIR` originally disabled caching when set to a `0`/`false`/`off`/`no`, but [_that_ was confusing](https://github.com/pypa/pip/issues/5735), so they [fixed it](https://github.com/pypa/pip/pull/5884) by **_also_** supporting `1`/`true`/`on`/`yes`.  That's right, this envvar disables the cache with  [either truthy **or** falsy value](https://github.com/pypa/pip/pull/5884/files#diff-da07b5eee61e8b55be71a57ff546de02R102-R114)!

docker/drama-free-django/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM centos:6
+
+ENV SCL_PYTHON_VERSION python27
+
+# Disables pip cache, which reduces build time, and suppresses warnings when run a non-root.
```
chosak:
```suggestion
# Disables pip cache, which reduces build time, and suppresses warnings when run as non-root.
```

chosak:
@hkeeler follow-up question about this PR. One concern I had that motivated the initial implementation as a single script was worrying about caching out-of-date versions of dependencies. Specifically with [this line](https://github.com/cfpb/cfgov-refresh/blob/ef52570d709dbb831a607e7ab2a77619438ceb1c/docker/drama-free-django/Dockerfile#L22), if and when new commits are merged to drama-free-django master, that line of the Dockerfile won't be re-run, will it? Locally I've encountered "Using cache" during the build step. How does Docker determine which lines to re-run?

Would it be a good idea to use [`docker build --no-cache`](https://docs.docker.com/engine/reference/commandline/build/) in the build and test scripts?

chosak:
I've opened https://github.com/cfpb/cfgov-refresh/pull/5173 to add `--no-cache`.

PR #5146, opened 2019-07-26
Expand tags inside a link that don't get an icon

willbarton:
This change refactors the way `` tags were being matched in links to prevent icons being added for external links. Instead of matching an `img` tag name in a `for` loop, it uses BeautifulSoup's CSS `select()` method. 

(As a personal aside, I'm much happier with this as a way to filter the links.)

This change also adds additional elements that can be contained within a link that do not get icons. That list is currently, `img`, `svg`, `div`, `h1`, `h2`, `h3`, `h4`, `h5`, and `h6`.

This is intended to fix this issue on production:

image
https://www.consumerfinance.gov/equifax-settlement/

On localhost, without the image file itself, this looks like:

image

http://localhost:8000/equifax-settlement/

With this change:

image

http://localhost:8000/equifax-settlement/

## Checklist

### DIFF
--- a/cfgov/core/utils.py
+++ b/cfgov/core/utils.py
@@ -24,8 +24,22 @@ LINK_ICON_CLASSES = 'a-link a-link__icon'
 
 LINK_ICON_TEXT_CLASSES = 'a-link_text'
 
-A_TAG = re.compile(r'<a [^>]*?>.+?(?=</a>)</a>(?s)(?i)')
-IMG_TAG = re.compile(r'<img[^>]*>(?i)')
+# Regular expression to match <a> links in HTML strings
+A_TAG = re.compile(
+    # Match an <a containing any attributes
+    r'<a [^>]*?>'
+    # And match everything inside
+    r'.+?'
+    # As long as it's not a </a>, then match '</a>'
+    '(?=</a>)</a>'
+    # Make '.' match new lines, ignore case
+    r'(?s)(?i)'
+)
+
+# If a link contains these elements, it should *not* get an icon
+ICONLESS_LINK_CHILD_ELEMENTS = [
+    'img', 'svg', 'div', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6'
+]
 
 
 def append_query_args_to_url(base_url, args_dict):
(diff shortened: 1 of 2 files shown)


cfgov/core/utils.py
```diff
@@ -24,8 +24,22 @@
...
-A_TAG = re.compile(r'<a [^>]*?>.+?(?=</a>)</a>(?s)(?i)')
-IMG_TAG = re.compile(r'<img[^>]*>(?i)')
+# Regular expression to match <a> links in HTML strings
+A_TAG = re.compile(
+    # Match an <a containing any attributes
+    r'<a [^>]*?>'
+    # And match everything inside
+    r'.+?'
+    # As long as it's not a </a>, then match '</a>'
+    '(?=</a>)</a>'
```
Scotchester:
```suggestion
    r'(?=</a>)</a>'
```

cfgov/core/utils.py
```diff
@@ -24,8 +24,22 @@
...
+    # Make '.' match new lines, ignore case
+    r'(?s)(?i)'
+)
+
+# If a link contains these elements, it should *not* get an icon
+ICONLESS_LINK_CHILD_ELEMENTS = [
+    'img', 'svg', 'div', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6'
```
anselmbradford: Possibly we could future-proof this by including [`picture`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture), or wait till which time we actually were to use that element.

anselmbradford: Actually, this may not be applicable since it contains an `img`, but not sure if the whole thing was wrapped in a link what gets the icon.

anselmbradford: I'd say disregard and we can add it later if needed.

willbarton: Heh, okay. @Scotchester and I also discussed making this a whitelist instead of a blacklist — which makes more sense to me. It'd be a trivial change based on this, I think, but I want some decision-making and documentation about what kind of links should get the icon and what kind should not before that.

PR #5156, opened 2019-08-02
Allow HMDA Explorer lender IDs to contain dashes

contolini:
We use JS to allow only numbers to be entered into the Lender ID field. Unfortunately, lender ID numbers also contain a dash. This PR adds the dash key to the list of permitted keycodes for that field.

## Testing

1. Go to https://www.consumerfinance.gov/data-research/hmda/explore, change the filters to "all records", open the lender section and try entering a dash into the Lender ID field. It should prevent you from doing so.
1. Pull down this branch.
1. Go to http://0.0.0.0:8000/data-research/hmda/explore and try entering a dash into the Lender ID field. It should now allow it.

## Screenshots

Before:

Screen Shot 2019-08-02 at 2 51 24 PM

After:

Screen Shot 2019-08-02 at 2 51 43 PM


contolini:
The HTML, CSS and JS for this page is a bunch of minified inlined mumbo-jumbo because we are in the process of retiring it. The diff for this PR is unreadable but literally all I did was add a number to an array:

Screen Shot 2019-08-02 at 2 49 54 PM

schbetsy:
The validation message is a bit misleading now:

Screen Shot 2019-08-02 at 3 14 34 PM

But since this tool is imminently going to be retired (🤞),  this doesn't seem like a dealbreaker to me.

contolini:
Oooh good call I'll change the message to "Numbers and dashes only, please!"

PR #5159, opened 2019-08-07
Fix iRegs notification typo

niqjohnson:
Fix a typo and missing margin in the iRegs notification

## Changes

- Fixed the typo: "ammendments" is now "amendments"

## Testing

1. Go to any previous or future version of any regulation, like the previous version of Reg C at http://localhost:8000/policy-compliance/rulemaking/regulations/1003/2018-01-01/
2. Make sure "amendments" is spelled correctly in the notification box and that the notification box is separated from the view links by 30px
3. Hit "view this version" to go to the first page of the regulation
4. Make sure "amendments" is spelled correctly in the notification box (the box was already separated from the content by the correct amount, so no need to test that)

## Screenshots

Current | Fixed
---- | ----
current | fixed

## Checklist

### Browsers

- [x] Chrome on desktop
- [x] Firefox
- [x] Safari on macOS
- [x] Edge
- [x] Safari on iOS
- [x] Chrome on Android

### Accessibility

- [x] Keyboard friendly
- [x] Screen reader friendly

### Other

- [x] Is useable without CSS
- [x] Is useable without JS

### DIFF
--- a/cfgov/regulations3k/jinja2/regulations3k/browse-regulation.html
+++ b/cfgov/regulations3k/jinja2/regulations3k/browse-regulation.html
@@ -164,7 +164,7 @@
         {%- endfor %}
 
         {% if requested_version != current_version %}
-        <div class="block block__flush-regs3k u-mb60">
+        <div class="block block__flush-regs3k">
             <div class="m-full-width-text
                         m-notification
                         m-notification__visible
--- a/cfgov/regulations3k/jinja2/regulations3k/browse-regulation.html
+++ b/cfgov/regulations3k/jinja2/regulations3k/browse-regulation.html
@@ -71,10 +71,10 @@
                         <div class="h4 m-notification_message">This version is not the current regulation.</div>
                         <p class="m-notification_explanation">
                         {% if requested_version.effective_date < current_version.effective_date %}
-                            You are viewing a previous version of this regulation with ammendments that went into effect on
+                            You are viewing a previous version of this regulation with amendments that went into effect on
                         {% endif %}
                         {% if requested_version.effective_date > current_version.effective_date %}
-                            You are viewing a future version of this regulation with ammendments that will go into effect on
+                            You are viewing a future version of this regulation with amendments that will go into effect on
                         {% endif %}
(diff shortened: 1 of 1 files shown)


cfgov/regulations3k/jinja2/regulations3k/browse-regulation.html
```diff
@@ -164,7 +164,7 @@ <h1>{{ page.title }}</h1>
         {%- endfor %}
 
         {% if requested_version != current_version %}
-        <div class="block block__flush-regs3k">
+        <div class="block block__flush-regs3k u-mb60">
```
anselmbradford: Margin bottom of regular ol' `block` should be 60px as set [here](https://github.com/cfpb/capital-framework/blob/822b932571e28dac0a1b834c4ba9901596240595/packages/cf-layout/src/cf-layout.less#L356). This appears to be an issue with `block__flush-regs3k`, can the issue being solved here be solved by removing that class and just having `block`?

niqjohnson: Yeah, the styles on this notification are getting a little messy. I tried to remove `block__flush-regs3k` and just leave `block`, but that causes too much space between the bottom of the intro and the top of the notification (`block__flush-regs3k` adjust the top margin of blocks, too). And then it snowballs: fixing *that* extra space means the spacing gets thrown off on states of the page when the notification isn't present, and on from there.

If keeping `u-mb60` feels like it's only making the styles messier, I'll remove that change, keep this PR to just the typo fix—which is more critical—and spend more time fixing the margins in a separate PR.

anselmbradford: Yeah let's fix the root issue. Would it be acceptable to have `block block__flush-top` on the notification? I know that has a larger top-margin, but the block above has two redundant block__flush-top classes added, which should be looked at anyway.

niqjohnson: @anselmbradford: Fixing the root issue makes sense to me. I've removed the style fixes from this PR, so now it's just the "amendment" typo that's addressed here. I'll open a separate issue for fixing the margins.

PR #5161, opened 2019-08-07
Create featured-content-module.js

anselmbradford:
Components shouldn't reach upward into their parent's dom. This creates a FeaturedContentModule that includes a VideoPlayer, as opposed to a VideoPlayer that's aware of FeaturedContentModule classes.

## Additions

- Adds `FeaturedContentModule` and `featured-content-module`.

## Changes

- Makes VideoPlayer dispatch `onPlay` and `onStop` events.

## Testing

1. Pull branch
1. `./frontend.sh`
1. Visit a bunch of different video/FCM pages:
   a. Standard static image FCM: http://localhost:8000/practitioner-resources/your-money-your-goals/
   b. Video FCM: http://localhost:8000/prepaid-rule/
   c. Standalone video: http://localhost:8000/data-research/prepaid-accounts/issuer-instructions/
   d. Event video: localhost:8000/about-us/events/archive-past-events/fall-2018-bcfp-advisory-committee-meetings-washington-dc/

## Notes

- This is spun off work from re-working the video player module.

PR #5162, opened 2019-08-08
Re-work video module to use composition instead of inheritance

anselmbradford:
This PR re-works the video player module to remove the inheritance relationship between `VideoPlayer` and `YoutubePlayer` and instead creates a utility module `youtube-api` that's composed into the `VideoPlayer` in an attempt to simplify the video player codebase and solve https://[GHE]/CFGOV/platform/issues/3456 

## Changes

- Remove inheritance of YoutubePlayer and VideoPlayer and use VideoPlayer directly.
- Include video `<iframe …>` directly in the template instead of creating it dynamically via JS.
- Set default export on js-loader utility.
- Separate out FeaturedContentModule script from VideoPlayer.

## Testing

1. Pull branch
1. `./frontend.sh`
1. Visit and test play a bunch of different video/FCM pages:
   a. Standard static image FCM: http://localhost:8000/practitioner-resources/your-money-your-goals/
   b. Video FCM: http://localhost:8000/prepaid-rule/
   c. Standalone video: http://localhost:8000/data-research/prepaid-accounts/issuer-instructions/
   d. Event video: http://localhost:8000/about-us/events/archive-past-events/fall-2018-bcfp-advisory-committee-meetings-washington-dc/

## Notes

- FCM supersedes https://github.com/cfpb/cfgov-refresh/pull/5161


cfgov/unprocessed/js/modules/youtube-api.js
```diff
@@ -0,0 +1,86 @@
...
+   ========================================================================== */
+
+import jsLoader from './util/js-loader';
+
+const IMAGE_URL = 'https://img.youtube.com/vi/%video_id%/maxresdefault.jpg';
+const SCRIPT_API = 'https://www.youtube.com/iframe_api';
+
+/**
+ * The "onYouTubeIframeAPIReady" function is automatically called when the
+ * YouTube iFrame API is done loading.
```
Scotchester: This is trivial, but YouTube calls it the "IFrame API" (capital I). If you care, might be nice to use that throughout here. (And also ensure it's "YouTube" with a capital 'T' everywhere.)

cfgov/unprocessed/js/modules/youtube-api.js
```diff
@@ -0,0 +1,86 @@
...
+ * See https://developers.google.com/youtube/iframe_api_reference
+ * @param {Function} callback - function to call when the API is ready.
+ */
+function attachAPIReadyCallback( callback ) {
+  window.onYouTubeIframeAPIReady = callback;
+}
+
+/**
+ * Embed the YouTube iFrame API.
+ * @param {Function} [callback] - function to call when the script is loaded.
```
Scotchester:
```suggestion
 * @param {Function} callback - function to call when the script is loaded.
```

cfgov/unprocessed/js/modules/youtube-api.js
```diff
@@ -0,0 +1,86 @@
...
+function fetchImageURL( videoId ) {
+  if ( !videoId ) {
+    throw new Error( 'No Video ID provided!' );
+  }
+
+  return IMAGE_URL.replace( '%video_id%', videoId );
+}
+
+/**
+ *
```
Scotchester: Could use a short summary of what the function does in this JSDoc.

cfgov/unprocessed/js/organisms/VideoPlayer.js
```diff
@@ -0,0 +1,190 @@
+/* ==========================================================================
+   Base Video Player Class
+   ========================================================================== */
+
+// Required modules.
+import { checkDom, setInitFlag } from '../modules/util/atomic-helpers';
+import youTubeAPI from '../modules/youtube-api';
+import EventObserver from '../modules/util/EventObserver';
```
Scotchester: Should be moved up a line to placate the https://eslint.org/docs/rules/sort-imports warning.

cfgov/unprocessed/js/organisms/VideoPlayer.js
```diff
@@ -0,0 +1,190 @@
...
+let _videoId;
+let _playBtnDom;
+let _closeBtnDom;
+let _iframeDom;
+let _player;
+
+let _playBtnClickedHandlerBinded;
+let _closeBtnClickedHandlerBinded;
+let _playVideoBinded;
+let _stopVideoBinded;
```
Scotchester: That's quite a lot of `lets`. How do we feel about comma-separating them? Our linter doesn't seem to mind. [Our guidance on variable declaration](https://github.com/cfpb/development/blob/master/standards/javascript.md#variable-declaration) is somewhat confusing.

cfgov/unprocessed/js/organisms/VideoPlayer.js
```diff
@@ -0,0 +1,190 @@
...
+
+  _imageLoad();
+
+  youTubeAPI.attachAPIReadyCallback( _videoAPIReady );
+  youTubeAPI.embedVideoScript();
+
+  return this;
+}
+
+function _videoAPIReady() {
```
Scotchester: Please add JSDoc.

cfgov/unprocessed/js/organisms/VideoPlayer.js
```diff
@@ -0,0 +1,190 @@
...
+function _videoPlayerReadyHandler() {
+  // Add events.
+  _playBtnDom.addEventListener( 'click', _playBtnClickedHandlerBinded );
+  _closeBtnDom.addEventListener( 'click', _closeBtnClickedHandlerBinded );
+
+  // The video has loaded.
+  _dom.classList.add( `${ BASE_CLASS }__loaded` );
+}
+
+function _videoStateChangeHandler( event ) {
```
Scotchester: Please add JSDoc.

cfgov/unprocessed/js/organisms/VideoPlayer.js
```diff
@@ -0,0 +1,190 @@
...
+  if ( event.data === window.YT.PlayerState.ENDED ) {
+    _stopVideoBinded();
+  }
+}
+
+/**
+ * HANDLE THUMBNAIL IMAGE.
```
Scotchester: Hmm... there are four functions, it appears, relating to this. Is it worth breaking them out into their own file?

If not, I'd like to see at least some extra blank lines or comment separators or something to make it clearer where this section begins and ends.

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -954,7 +954,7 @@ class Meta:
         classname = 'block__flush'
 
     class Media:
-        js = ['video-player.js']
+        js = ['featured-content-module.js']
```
Scotchester: Out of scope for this PR, but it would be nice if we could figure out how to not load this JS if a given FCM doesn't have a video in it, only a static image.

cfgov/unprocessed/js/modules/youtube-api.js
anselmbradford: This is correct with the brackets. Those denote [optional arguments in jsdocs](https://devhints.io/jsdoc). Where this is called, [here](https://github.com/cfpb/cfgov-refresh/pull/5162/files#diff-9efea6479fb7879a333f4d6718297e33R71), the callback is not supplied. We could remove it entirely, but it is useful for debugging and has been carried over from the prior codebase.

cfgov/unprocessed/js/organisms/VideoPlayer.js
anselmbradford: Personally I'm against comma-separated variable declarations, and I thought the linter did mind, but maybe I didn't enforce it due to both being in the codebase. 

Consider this structure:

```
let a,
b,
c;
```
This makes it (1) unnecessarily harder to find within a code search where (e.g.) `b` is first defined, and (2) makes it easier to produce syntax errors when deleting:

```
let a,
b,
```

Or re-arranging:

```
let a,
c;
b,
```

Single line variable declarations are cleaner for these reasons, in my opinion. It's not our job to do what the minifier can do.

If lengthy variable declarations are bothersome, they may indicate the functionality they define could in part be spun off to their own module.

Scotchester: Those are great points!

cfgov/unprocessed/js/modules/youtube-api.js
Scotchester: Ah, whoops! That makes sense.

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -103,6 +104,11 @@
     }
 }
 
+.o-video-player__loaded .o-video-player_play-btn {
+    opacity: 1;
```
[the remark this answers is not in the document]
anselmbradford: Updated in https://github.com/cfpb/cfgov-refresh/pull/5162/commits/868a228a15a0764b43eb6f4e9609c93693d39de6

I changed it to a link and button in the markup that are alternatively shown and hidden with the `u-hidden` class instead of using opacity.

cfgov/unprocessed/js/modules/youtube-api.js
anselmbradford: Fixed

cfgov/v1/atomic_elements/organisms.py
anselmbradford: Fixed in https://github.com/cfpb/cfgov-refresh/pull/5162/commits/868a228a15a0764b43eb6f4e9609c93693d39de6 to bail out when there is no video present.

cfgov/unprocessed/js/organisms/VideoPlayer.js
anselmbradford: Added additional code comment section below the four image methods.

cfgov/unprocessed/js/organisms/VideoPlayer.js
anselmbradford: Added

cfgov/unprocessed/js/organisms/VideoPlayer.js
anselmbradford: Added

cfgov/unprocessed/js/modules/youtube-api.js
anselmbradford: Added

cfgov/unprocessed/js/organisms/VideoPlayer.js
anselmbradford: Moved

anselmbradford:
Updated this so there will be a play "button" link that opens the video in a new window when JavaScript is turned off or before the script has initialized.

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -94,15 +96,24 @@
...
-        left: 0.09375em;
-
-        // Ensuring play icon is vertically centered.
-        vertical-align: bottom;
+        // Size and position visually centers the play arrow.
+        position: absolute;
+        left: 55%;
+        top: 50%;
+        transform: translate( -50%, -50% );
+        font-size: 30px;
```
Scotchester: This is already inherited from the button element itself and isn't needed.
```suggestion
```

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -94,15 +96,24 @@
...
-        position: relative;
-        left: 0.09375em;
-
-        // Ensuring play icon is vertically centered.
-        vertical-align: bottom;
+        // Size and position visually centers the play arrow.
+        position: absolute;
+        left: 55%;
+        top: 50%;
+        transform: translate( -50%, -50% );
```
Scotchester: Any particular reason for changing the positioning method? It seems to end up in the same place, but it takes one more rule.

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -70,12 +70,14 @@
 .o-video-player_play-btn {
```
Scotchester: Looks like we need to add this for the button to be the correct size when it's an `a` element.

```suggestion
.o-video-player_play-btn {
    box-sizing: border-box;
```

Scotchester: Not sure what's up with it showing a deleted space before the selector. Maybe don't use the UI to accept this suggestion and do it manually.

cfgov/unprocessed/js/organisms/FeaturedContentModule.js
```diff
@@ -0,0 +1,69 @@
...
+ */
+function init() {
+  setInitFlag( _dom );
+
+  const videoPlayerDom = _dom.querySelector( `.${ VideoPlayer.BASE_CLASS }` );
+
+  // If we don't have a video in this FCM, bail out.
+  if ( videoPlayerDom === null ) {
+    return this;
+  }
```
Scotchester: This is helpful, but my point was that we shouldn't even request and download this JS. Still think it's out of scope for this PR, though.

cfgov/jinja2/v1/_includes/macros/video-player.html
```diff
@@ -58,7 +58,16 @@
...
+                        src="{{ video_url }}"
+                        frameborder="0"
+                        allow="autoplay"
+                        allowfullscreen="true"
+                        scrolling="no">
+                </iframe>
+            </div>
             <div class="o-video-player_video-actions-container">
                 <div class="o-video-player_video-actions">
                     <button class="a-btn o-video-player_close-btn" href="/">
```
Scotchester: Another thing that's out of scope for this PR: I was testing with a screen reader to make sure that it didn't read both the fallback `a` play button and the JS-enabled `button` play button. It doesn't, but unfortunately, I discovered that I must've not tested this updated FCM thoroughly in the first place. Definitely some weirdness happening. But let's not let that hold up this work.

cfgov/unprocessed/css/organisms/video-player.less
anselmbradford: The other positioning didn't work the same for both a `a` link and `button`. This way it does work the same for both.

cfgov/unprocessed/css/organisms/video-player.less
anselmbradford: Fixed

cfgov/unprocessed/css/organisms/video-player.less
anselmbradford: Fixed

PR #5173, opened 2019-08-14
Use --no-cache for drama-free-django Docker build

chosak:
The Docker-based process to build and test drama-free-django uses a Dockerfile that includes a reference to the DFD GitHub repository.

If `docker build` is used without `--no-cache`, it's possible that changes to that repo wouldn't get included even if the image is rebuilt.

This change adds `--no-cache` to ensure that the image is completely rebuilt each time.

## Testing

Follow the instructions [here](https://github.com/cfpb/cfgov-refresh/tree/master/docker/drama-free-django) and observe that the image gets completely rebuilt.

## Checklist

- [X] PR has an informative and human-readable title
- [X] Changes are limited to a single goal (no scope creep)
- [X] Code can be automatically merged (no conflicts)
- [X] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [X] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated
- [X] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #5174, opened 2019-08-14
Don't cache the dfd install in the dfd docker build

wpears:
This addresses the problem andy described in https://github.com/cfpb/cfgov-refresh/pull/5173, but with a **COOL HACK** that allows us to use most of the cache while always getting fresh code from drama-free-django.

## Testing:
Follow the instructions [here](https://github.com/cfpb/cfgov-refresh/tree/master/docker/drama-free-django) two times in a row. You'll note that for at least the second attempt, the first 9 steps should use the cache, but no others (depending on if you've run these commands before the first 8 images will likely be cached on your first run as well).

### DIFF
--- a/docker/drama-free-django/Dockerfile
+++ b/docker/drama-free-django/Dockerfile
@@ -30,7 +30,7 @@ RUN yum install -y centos-release-scl && \
 # Set cache-busting arg, a unique value should be provided when the build is invoked
 ARG CACHEBUST
 
-RUN pip install -U git+https://github.com/cfpb/drama-free-django.git
+RUN source /etc/profile && pip install -U git+https://github.com/cfpb/drama-free-django.git
 COPY _build.sh _test.sh docker-entrypoint.sh ./
 
 ENTRYPOINT ["./docker-entrypoint.sh"]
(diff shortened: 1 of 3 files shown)


chosak:
This is indeed a **COOL HACK**. One thing, though. What about the [other dependencies](https://github.com/cfpb/cfgov-refresh/blob/610f207c3f5f68941d5427adc8af9882b61c6252/docker/drama-free-django/Dockerfile#L23) -- how often do these versions change? I'm thinking things like git or gcc change so infrequently that it's unlikely to matter. But what about node, or yarn? Is it worth caching _some_ of these?

wpears:
Not often enough to matter for the added time savings, imo. Also, if we upgrade, say, node throughout the project, we'll upgrade it here too and the cache won't persist since we've changed the dockerfile

wpears:
Hold! A bug exists :)

hkeeler:
This is indeed a **COOL HACK**.  :)  I like it.

I do share @chosak's concern about yarn and node versions getting stale.  As-is, they'll never (very infrequently) get updated on our internal CI/CD servers.  Are we OK with that?  I have a feeling it'll _eventually_ bite us.  Perhaps we need some internal docs on how to manually run this with `--no-cache` in case the build starts to break.

Overall, 👍 though.

PR #5177, opened 2019-08-19
new docker stage that builds and collects frontend assets

rosskarchner:
This adds an additional stage to the Dockerfile that installs Node 10, yarn, runs the front-end build, and runs collectstatic in production mode (putting the resulting files in /var/www/html/static )

This will not change life for docker-compose users, since the python containers are [pinned to to the cfgov-dev stage](https://github.com/cfpb/cfgov-refresh/blob/docker-multistage/docker-compose.yml#L25)

## Additions

- cfgov-build docker stage

## Testing

`docker build . --build-arg scl_python_version=rh-python36 --target cfgov-build -t tmp && docker run -it tmp:latest bash` should run the whole build, and drop you in a shell session where you can inspect the results. You can see the built (and post-processed) static files in /var/www/html/static

## Notes

-

## Todos

-

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM centos:7 AS cfgov-base
 
-SHELL ["/bin/bash", "--login", "-c"]
+SHELL ["/bin/bash", "--login", "-o", "pipefail", "-c"]
 
 # Stops Python default buffering to stdout, improving logging to the console.
 ENV PYTHONUNBUFFERED 1
@@ -62,25 +62,19 @@ RUN curl --silent --location https://dl.yarnpkg.com/rpm/yarn.repo | tee /etc/yum
 RUN yum install -y nodejs yarn  && \
     yum clean all && rm -rf /var/cache/yum
 
-WORKDIR /src/cfgov-refresh
 
-COPY scripts scripts
-COPY frontend.sh .
-COPY package.json .
-COPY yarn.lock .
-COPY babel.config.js .
-COPY gulpfile.js .
+COPY cfgov/ ./cfgov
+COPY config/ ./config
 COPY gulp ./gulp
-COPY config ./config
-COPY jest.config.js .
-COPY cfgov/ ./cfgov/
 COPY static.in ./static.in
+COPY scripts ./scripts
+
+COPY frontend.sh gulpfile.js jest.config.js package.json yarn.lock /src/cfgov-refresh/
 
 ENV DJANGO_SETTINGS_MODULE=cfgov.settings.production
+ENV DJANGO_STATIC_ROOT=/var/www/html/static
 ENV ALLOWED_HOSTS='["*"]'
 
 RUN sh ./frontend.sh production
 
-RUN DJANGO_STATIC_ROOT=/var/www/html/static cfgov/manage.py collectstatic
-
-ENTRYPOINT []
+RUN cfgov/manage.py collectstatic
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,3 @@
-# Base image inherited by all cfgov-refresh child images
 FROM centos:7 AS cfgov-base
 
 SHELL ["/bin/bash", "--login", "-c"]
@@ -53,3 +52,34 @@ COPY extend-environment.sh /etc/profile.d/extend-environment.sh
 RUN pip install --no-cache-dir -r requirements/local.txt
 
 CMD ["python", "./cfgov/manage.py", "runserver", "0.0.0.0:8000"]
+
+FROM cfgov-develop as cfgov-build
+
+# add node and yarn repos
+RUN curl -sL https://rpm.nodesource.com/setup_10.x | bash -
+RUN curl --silent --location https://dl.yarnpkg.com/rpm/yarn.repo | tee /etc/yum.repos.d/yarn.repo
+
+RUN yum install -y nodejs yarn  && \
+    yum clean all && rm -rf /var/cache/yum
+
+WORKDIR /src/cfgov-refresh
+
+COPY scripts scripts
+COPY frontend.sh .
+COPY package.json .
+COPY yarn.lock .
+COPY babel.config.js .
+COPY gulpfile.js .
+COPY gulp ./gulp
+COPY config ./config
+COPY jest.config.js .
+COPY cfgov/ ./cfgov/
+COPY static.in ./static.in
+
+ENV DJANGO_SETTINGS_MODULE=cfgov.settings.minimal_collectstatic
+ENV ALLOWED_HOSTS='["*"]'
+ENTRYPOINT []
+
+RUN sh ./frontend.sh production
+
+RUN DJANGO_STATIC_ROOT=/var/www/html/static cfgov/manage.py collectstatic
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,10 +76,11 @@ COPY jest.config.js .
 COPY cfgov/ ./cfgov/
 COPY static.in ./static.in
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -53,3 +52,35 @@ COPY extend-environment.sh /etc/profile.d/extend-environment.sh
...
+FROM cfgov-develop as cfgov-build
+
+# add node and yarn repos
+RUN curl -sL https://rpm.nodesource.com/setup_10.x | bash -
+RUN curl --silent --location https://dl.yarnpkg.com/rpm/yarn.repo | tee /etc/yum.repos.d/yarn.repo
+
+RUN yum install -y nodejs yarn  && \
+    yum clean all && rm -rf /var/cache/yum
+
+WORKDIR /src/cfgov-refresh
```
hkeeler: This is already set above.

https://github.com/cfpb/cfgov-refresh/pull/5177/files#diff-3254677a7917c6c01f55212f86c57fbfR8-R10

Dockerfile
```diff
@@ -53,3 +52,35 @@ COPY extend-environment.sh /etc/profile.d/extend-environment.sh
...
+COPY jest.config.js .
+COPY cfgov/ ./cfgov/
+COPY static.in ./static.in
+
+ENV DJANGO_SETTINGS_MODULE=cfgov.settings.production
+ENV ALLOWED_HOSTS='["*"]'
+
+RUN sh ./frontend.sh production
+
+RUN DJANGO_STATIC_ROOT=/var/www/html/static cfgov/manage.py collectstatic
```
hkeeler: Why is `DJANGO_STATIC_ROOT` not an `ENV`?

Dockerfile
```diff
@@ -53,3 +52,35 @@ COPY extend-environment.sh /etc/profile.d/extend-environment.sh
...
+# add node and yarn repos
+RUN curl -sL https://rpm.nodesource.com/setup_10.x | bash -
+RUN curl --silent --location https://dl.yarnpkg.com/rpm/yarn.repo | tee /etc/yum.repos.d/yarn.repo
+
+RUN yum install -y nodejs yarn  && \
+    yum clean all && rm -rf /var/cache/yum
+
+WORKDIR /src/cfgov-refresh
+
+COPY scripts scripts
```
hkeeler: Perhaps sort these alphabetically?

Dockerfile
```diff
@@ -53,3 +52,35 @@ COPY extend-environment.sh /etc/profile.d/extend-environment.sh
 RUN pip install --no-cache-dir -r requirements/local.txt
 
 CMD ["python", "./cfgov/manage.py", "runserver", "0.0.0.0:8000"]
+
+FROM cfgov-develop as cfgov-build
+
+# add node and yarn repos
+RUN curl -sL https://rpm.nodesource.com/setup_10.x | bash -
+RUN curl --silent --location https://dl.yarnpkg.com/rpm/yarn.repo | tee /etc/yum.repos.d/yarn.repo
```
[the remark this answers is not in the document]
rosskarchner: I'll give that a try-- I think those were both pasted from official docs

Dockerfile
```diff
@@ -53,3 +52,35 @@ COPY extend-environment.sh /etc/profile.d/extend-environment.sh
...
+COPY static.in ./static.in
+
+ENV DJANGO_SETTINGS_MODULE=cfgov.settings.production
+ENV ALLOWED_HOSTS='["*"]'
+
+RUN sh ./frontend.sh production
+
+RUN DJANGO_STATIC_ROOT=/var/www/html/static cfgov/manage.py collectstatic
+
+ENTRYPOINT []
```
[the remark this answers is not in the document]
rosskarchner: Probably unneccessary, but it made it easier to test in the early phases (before the entrypont was copied over). I'll remove that

Dockerfile
wpears: +1 to alphabetically, but I'd also think it's worthwhile to include these all in one `COPY` (space-separated), then they'll only be a single layer

Dockerfile
```diff
@@ -53,3 +52,35 @@ COPY extend-environment.sh /etc/profile.d/extend-environment.sh
...
 
 CMD ["python", "./cfgov/manage.py", "runserver", "0.0.0.0:8000"]
+
+FROM cfgov-develop as cfgov-build
+
+# add node and yarn repos
+RUN curl -sL https://rpm.nodesource.com/setup_10.x | bash -
+RUN curl --silent --location https://dl.yarnpkg.com/rpm/yarn.repo | tee /etc/yum.repos.d/yarn.repo
+
+RUN yum install -y nodejs yarn  && \
```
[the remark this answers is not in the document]
wpears: I'd say we're okay leaving this how it is, I agree 10.x is fine and the way the yarn repo is added above, you get `stable` by default, which is probbbbbbably fine for a command line tool.

In a production image, though, I'd think we'd want yarn and node's install in a separate stage from from the actual frontend build, so that it wouldn't be rerun on every code change

Dockerfile
rosskarchner: Because it's only used once?

Dockerfile
rosskarchner: Yeah, I only recently learned that was possible-- I'll make that change

Dockerfile
hkeeler: Ah.  OK.  I thought it was used when the app was running too.

Dockerfile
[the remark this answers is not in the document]
hkeeler: Does that really matter, though?  Yes, the `COPY`s and forward would rerun in this stage, but the expensive operations (nodejs/yarn install here) would remain cached.

Dockerfile
rosskarchner: that's in theory possible-- so maybe I will set it with ENV. In general, I think this image is just for producing assets that will get copied to the production image

hkeeler: Oh right, the production image won't actually inherit from this one.  You _could_ move it into `cfgov-base`, then it'd be available in both.  Might be nice so you don't have to set it twice, or remember to change it in two places.  Or we could wait to figure that out when we do the full Apache work since I'm guessing we'll have other `ENV`s that this could build off of.

rosskarchner: so we set it in base, but override it in deveop?

hkeeler: That doesn't really work either, though, since `-build` inherits from `-develop`.  You'd just end up re-resetting it.

I say just keep it as-is for this PR, and we'll probably have a better idea of how all the stages will fit together when we do that Apache work.

Dockerfile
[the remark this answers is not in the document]
wpears: Ahhh, that's a good point! Yeah so I guess this should work well then going forward.

Dockerfile
rosskarchner: removed!

Dockerfile
rosskarchner: OK

Dockerfile
rosskarchner: I was kind of defeated in my attempt to put them all on one line-- everything I tried to do with directories, ended up with the *contents* of the directories being dropped into /src/cfgov-refresh. Any ideas are welcome, but in the meantime I just copied the directories separately from the single files.

Dockerfile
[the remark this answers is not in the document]
rosskarchner: done!

PR #5188, opened 2019-08-23
Fix Python 3-incompatible code in Related Posts

chosak:
This change changes a use of [`iteritems()` to `items()`](https://python-future.org/compatible_idioms.html#iterating-through-dict-keys-values-items) to make the related posts template compatible with both Python 2 and Python 3. 

Currently the related posts module doesn't work in Python 3 because of this. I've added a test case to cover this behavior.

In Python 2, going from `iteritems()` (an iterator) to `items()` (a list) could in theory be slightly slower, but in practice the related posts module only has a maximum of 9 items, so this shouldn't have a big impact on performance.

## Testing

See new unit test. Also visit [any page](http://localhost:8000/about-us/blog/protect-yourself-credit-card-fraud-register/) locally that has a related posts module in the sidebar.

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/molecules/related-posts.html
+++ b/cfgov/jinja2/v1/_includes/molecules/related-posts.html
@@ -94,7 +94,7 @@
             </h2>
         </header>
     {% endif %}
-    {% for post_type, post_type_list in posts.iteritems() %}
+    {% for post_type, post_type_list in posts.items() %}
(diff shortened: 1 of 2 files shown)


richaagarwal:
This looks good!  I'm wondering if there's any way to move this logic into the back-end so that it can be more easily tested (and not reliant on checking CSS classes) but I understand if that's not a straightforward thing to do.

PR #5196, opened 2019-08-27
Add multiple route options to reducer

el-mapache:
This pull request add support for multiple routes in the route option reducer. Each `routeOptionView` gets passed the index of the route it is responsible for.

## Additions

- Add a route factory function to build new route data objects

## Removals

-

## Changes

- Since the tool will only allow for two routes for now, explicitly initialize
each view and expandable in `index.js`
- Update reducer to handle multiple route objects

## Testing

1. Just unit tests at this point

## Notes

- I don't love all the manual selecting of `expandables` and `views` in the `index.js` file, but we only need two route options for the tool and this is the simplest way to ensure that all the data goes to the right places and all the JS objects get initialized correctly

## Todos

-

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/cfgov/unprocessed/apps/youth-employment-success/js/reducers/route-option-reducer.js
+++ b/cfgov/unprocessed/apps/youth-employment-success/js/reducers/route-option-reducer.js
@@ -1,18 +1,20 @@
 import { actionCreator, assign } from '../util';
 
 const initialState = {
-  transportation: '',
-  daysPerWeek: '',
-  miles: '',
-  dailyCost: ''
+  routes: []
 };
-const actionTypes = {
+
+const actionTypes = Object.freeze( {
+  ADD_ROUTE_OPTION: 'ADD_ROUTE_OPTION',
   UPDATE_TRANSPORTATION: 'UPDATE_TRANSPORTATION',
   UPDATE_MILES: 'UPDATE_MILES',
   UPDATE_DAILY_COST: 'UPDATE_DAILY_COST',
   UPDATE_DAYS_PER_WEEK: 'UPDATE_DAYS_PER_WEEK'
-};
+} );
 
+const addRouteOptionAction = actionCreator(
+  actionTypes.ADD_ROUTE_OPTION
+);
 const updateTransportationAction = actionCreator(
   actionTypes.UPDATE_TRANSPORTATION
 );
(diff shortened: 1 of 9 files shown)


cfgov/unprocessed/apps/youth-employment-success/js/reducers/route-option-reducer.js
```diff
@@ -26,6 +28,46 @@ const updateDailyCostAction = actionCreator(
...
+/**
+ *
+ * @param {array} routes the array of routes in the reducer's state
+ * @param {number} routeIndex the index of the route object we want to update
+ * @param {object} data object describing the updates we want to make
+ *  to the route object's proeprties
+ * @returns {object} the next state of the reducer
+ */
+function updateRouteData( routes, routeIndex, data ) {
+  return {
```
anselmbradford: If you have a big block of logic you're returning I might store it as an object and then return that, which I think is a little clearer when parsing over the code, like:

```
const nextState = {
  routes: routes.map( ( route, index ) => {
    if ( index !== routeIndex ) {
      return route;
    }

    return assign( route, data );
  } )
};

return nextState;
```

cfgov/unprocessed/apps/youth-employment-success/js/route-option-view.js
```diff
@@ -43,29 +44,35 @@ function RouteOptionFormView( element, { store } ) {
    * @param {object} updateObject object with DOM event and field name
    */
   function _setQuestionResponse( { name, event } ) {
-    const method = actionMap[name];
+    const action = actionMap[name];
+    const { target: { value }} = event;
```
anselmbradford: TRIVIAL: should be `const { target: { value } } = event;` looks like

cfgov/unprocessed/apps/youth-employment-success/js/route.js
```diff
@@ -0,0 +1,18 @@
+import { assign } from './util';
+
+/**
+ * Factory function for creating a route object. Designed to be used in conjunction
+ * with the routeOptionReducer
+ * @param {object} route optional object describing the values of the route object
```
anselmbradford: You can enclose the parameter name in brackets to designate to jsdocs that it's optional (see [here](https://devhints.io/jsdoc)). So should be:

```
* @param {Object} [route] optional object describing the values of the route object.
```

cfgov/unprocessed/apps/youth-employment-success/js/route-option-view.js
el-mapache: I wonder why the linter didn't catch this! 😠

cfgov/unprocessed/apps/youth-employment-success/js/route.js
el-mapache: oh cool, thanks!

PR #5203, opened 2019-08-29
Add Apache httpd hosting option to Docker setup

hkeeler:
This PR attempts to simulate what a production-like Docker setup _could_ look like.  It extends the current dev-centric Docker setup to include the option for hosting the app in Apache using the recently-added [`cfgov/apache`](https://github.com/cfpb/cfgov-refresh/tree/master/cfgov/apache) configs.

## Additions

- Adds optional Apache-based hosting in Docker.

## Changes

- Refactors `Dockerfile` to have just two stages, `dev` for local development, and `prod`.  The default remains `dev`, so the current dev workflow should not change.
- The Docker Compose setup has been split up, with shared bits in `docker-compose.yml`, default dev-centric setup in `docker-compose.override.yml`, and prod-like setup in `docker-compose.prod.yml`.
- `.dockerignore` now ignores files by default, and files/directories must be _added_ to be included in a `Dockerfile`'s `COPY`/`ADD`.

    **Note:** This is not set in stone.  We may change this to just have two separate compose files if this is too confusing.

## Testing

1. Manual testing by switching between Django (dev) and Apache httpd (prod) Compose files.

    1. Wipe out any existing Compose-provisioned containers.

            docker-compose down

    1. Start the dev stack, and test everything works, while watching the compose logs for errors.

            docker-compose up --build

    1. Switch to the prod stack, and retest.

            docker-compose -f docker-compose.yml -f docker-compose.prod.yml up --build

    1. Switch _back_ to the dev stack, and re-retest.

            docker-compose up --build

1. Follow the above-mentioned procedure, use the smoke test scripts.

## Notes

-

## Todos

- [x] Try to refactor `wsgi.conf` so it is usable in both Docker and Ansible.  There is currently a separate `wsgi.conf.docker` file in this PR.
- [ ] Remove cfgov-sandbox setup, which doesn't really apply to a Docker setup.
- [x] Consider symlinking to apache configs rather than copying them around at `docker build`-time.
- [x] Decide whether we want to keep the current Compose setup.
- [x] Document how to use this new setup.

## Checklist


cfgov/apache/conf.d/redirects.conf
```diff
@@ -1019,12 +1019,10 @@ RedirectMatch permanent (?i)^/practitioner-resources/servicemembers/protecting/?
...
 RedirectMatch (?i)(.*)morgage(.*) $1mortgage$2
 
-
-#######################
-# Everything below this point is diverging from the production configs at
-# https://GHE/CFPB/conf.d/blob/master/redirects.conf
-
 Redirect /assets https://s3.amazonaws.com/files.consumerfinance.gov/a/assets
+
+<If "-T osenv('APACHE_COMMON_LOGIN_ALIAS')">
```
wpears: Can kill section due to it aging out, see conversation on a-c-f

wpears:
Cool! I like the rewrite rule stuff for the  upload conditionals. I think your approach to DocumentRoot looks fine to me (maybe even better ^^;)

cfgov/apache/conf.d/redirects.conf
hkeeler: Will do.

rosskarchner:
If you pull in or cherry pick https://github.com/cfpb/cfgov-refresh/commit/867ec696b151aa857f229dfc0a4ff11c8fd82456 it'll add some documentation for using the prodlike image.

cfgov/apache/conf.d/uploads.conf
```diff
@@ -0,0 +1,9 @@
+RewriteEngine On
```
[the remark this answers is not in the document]
hkeeler: Should we move it into `httpd.conf` so it's clearer that it's just generally on?

hkeeler:
> If you pull in or cherry pick [867ec69](https://github.com/cfpb/cfgov-refresh/commit/867ec696b151aa857f229dfc0a4ff11c8fd82456) it'll add some documentation for using the prodlike image.

Awesome!  Will do once we decide if this PR will stand as-is, or if we're going to split it up between Docker and Apache changes.

cfgov/apache/conf.d/headers.conf
```diff
@@ -10,6 +10,12 @@ Header always set X-Frame-Options SAMEORIGIN
 Header always set X-XSS-Protection: "1; mode=block"
 Header always set X-Content-Type-Options: nosniff
 
+# Add X-Forwarded headers when APACHE_HTTPS_FORWARDED_HOST envvar is set
+<If "-n osenv('APACHE_HTTPS_FORWARDED_HOST')">
+        RequestHeader set X-Forwarded-Proto https
+        RequestHeader set X-Forwarded-Host "${APACHE_HTTPS_FORWARDED_HOST}"
```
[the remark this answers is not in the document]
hkeeler: Good question.  I'll dig into that and get back to you on that one.

cfgov/apache/conf.d/uploads.conf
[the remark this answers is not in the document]
chosak: Sure, that sounds sensible too.

hkeeler:
> This is great! Is it worth splitting out the "pure Apache in cfgov-refresh" bits into a separate PR from the Docker-related bits?

Good question.  I don't have a strong opinion on that one.  It has been nice having them on the same branch to confirm each Apache change works in each env, but now that the Docker side has stabilized, I think we _could_ pull them apart if we feel it'd be easier to review.

cfgov/apache/conf/httpd.conf
```diff
@@ -279,6 +276,9 @@ AddDefaultCharset UTF-8
 #EnableMMAP off
 EnableSendfile on
 
+# Enable mod_rewrite globally
```
chosak: This might be unnecessary, but for consistency with the rest of this file, could we enclose this in an `<IfModule rewrite_module></IfModule>`?

docker-compose.override.yml
```diff
@@ -0,0 +1,23 @@
...
+services:
+
+    python2:
+        build:
+            target: cfgov-dev
+        ports:
+            - "8000:8000" # Django
+        volumes:
+            - ./:/src/cfgov-refresh
+            - ./develop-apps:/src/develop-apps
```
chosak: Something feels wrong about this line, although I realize you haven't changed it in this PR. The [`extend-environment.sh`](https://github.com/cfpb/cfgov-refresh/blob/688a818670a86c4f34d230f6bfd82f4eaccfb110/extend-environment.sh#L3) script adds apps that are in /src/cfgov-refresh/develop-apps/ to the path, not apps that are in /src/develop-apps. And since the entire source tree gets added under /src/cfgov-refresh anyway, isn't it unnecessary to add another volume specifically for develop-apps? Perhaps this line (and the similar one below) can be removed?

hkeeler: Agreed.  That _seems_ broken.  I'll check it out.

cfgov/apache/conf.d/headers.conf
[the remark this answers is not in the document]
hkeeler: I believe this works as-is because it is injecting the `${APACHE_HTTPS_FORWARDED_HOSTS}` via [HTTPD config substitution](https://httpd.apache.org/docs/current/configuring.html#syntax), while the previous version used [`RequestHeader`](https://httpd.apache.org/docs/current/mod/mod_headers.html#requestheader)'s [ap_expr](https://httpd.apache.org/docs/current/expr.html) expression parser.  The latter is much more powerful, but since we only need to inject an envvar, this way seems simpler.

Dockerfile
```diff
@@ -9,72 +13,90 @@ ENV APP_HOME /src/cfgov-refresh
...
+        ${SCL_PYTHON_VERSION} && \
+    yum clean all && rm -rf /var/cache/yum && \
+    echo "source scl_source enable ${SCL_PYTHON_VERSION}" > /etc/profile.d/enable_scl_python.sh && \
     source /etc/profile && \
-    pip install --upgrade pip setuptools
+    pip install --no-cache-dir --upgrade pip setuptools
 
-# Copy over our requirement files to install
+# Disables pip cache. Reduces build time, and suppresses warnings when run as non-root.
+# NOTE: MUST be after pip upgrade.  Build fails otherwise due to bug in old pip.
```
chosak:
```suggestion
# NOTE: MUST be after pip upgrade. Build fails otherwise due to bug in old pip.
```

Dockerfile
```diff
@@ -9,72 +13,90 @@ ENV APP_HOME /src/cfgov-refresh
...
+    yarn cache clean && \
+    yum -y remove nodejs yarn && \
+    yum clean all && rm -rf /var/cache/yum && \
+    rm -rf \
+        cfgov/apache/www \
+        cfgov/unprocessed \
+        node_modules && \
+    ln -s ${SCL_HTTPD_ROOT}/etc/httpd/modules ${APACHE_SERVER_ROOT}/modules && \
+    mkdir -p cfgov/f/ && chown apache:apache cfgov/f/ && \
+    mkdir /tmp/eregs_cache && chown apache:apache /tmp/eregs_cache
```
chosak: Request: comment here indicating that this is the local of the Django file-based cache?

Dockerfile
```diff
@@ -9,72 +13,90 @@ ENV APP_HOME /src/cfgov-refresh
...
+    cfgov/manage.py collectstatic && \
+    yarn cache clean && \
+    yum -y remove nodejs yarn && \
+    yum clean all && rm -rf /var/cache/yum && \
+    rm -rf \
+        cfgov/apache/www \
+        cfgov/unprocessed \
+        node_modules && \
+    ln -s ${SCL_HTTPD_ROOT}/etc/httpd/modules ${APACHE_SERVER_ROOT}/modules && \
+    mkdir -p cfgov/f/ && chown apache:apache cfgov/f/ && \
```
chosak: Request: comment here indicating that this is the location for file uploads?

docs/running-docker.md
```diff
@@ -101,3 +101,99 @@ For `docker` commands, `[CONTAINER]` is the container name displayed with `docke
...
+docker build . --build-arg scl_python_version=rh-python36 -t your-desired-image-name
+```
+
+**Note:** The `scl_python_version` build arg specifies which
+[Python Software Collection](https://www.softwarecollections.org/en/scls/?search=python)
+version you'd like to use. We've tested this against `python27` and `rh-python36`.
+
+### Docker Compose
+
+You can also launch the full cf.gov stack locally via `docker-compose`.  This setup is
```
chosak:
```suggestion
You can also launch the full cf.gov stack locally via `docker-compose`. This setup is
```

docs/running-docker.md
```diff
@@ -101,3 +101,99 @@ For `docker` commands, `[CONTAINER]` is the container name displayed with `docke
...
+a nice way to test out new Apache config changes.  It includes volumes that mount your
```
chosak:
```suggestion
a nice way to test out new Apache config changes. It includes volumes that mount your
```

docs/running-docker.md
```diff
@@ -101,3 +101,99 @@ For `docker` commands, `[CONTAINER]` is the container name displayed with `docke
...
+
+    ```bash
+    docker-compose rm -sf python3 python2
+    docker-compose up --build python3 python2
+    ```
+
+
+## How does it work?
+
+The production image extendes the development image. If you look at the `Dockerfile`, this is spelled out by the line:
```
chosak:
```suggestion
The production image extends the development image. If you look at the `Dockerfile`, this is spelled out by the line:
```

docs/running-docker.md
```diff
@@ -101,3 +101,99 @@ For `docker` commands, `[CONTAINER]` is the container name displayed with `docke
...
+FROM cfgov-dev as cfgov-prod
+```
+
+Both 'cfgov-dev' and 'cfgov-prod' are called "[build stages](https://docs.docker.com/develop/develop-images/multistage-build/)". That line means, "create a new stage, starting from cfgov-dev, called cfgov-prod".
+
+From there, we:
+
+- Install SCL-based Apache HTTPD, and the `mod_wsgi` version appropriate for our chosen `scl_python_version`.
+- Run frontend.sh, Django's collectstatic command, and then *uninstall* node and yarn.
+- Set the default command on container startup to `httpd -d ./cfgov/apache -D FOREGROUND`, which runs apache using
```
chosak:
```suggestion
- Set the default command on container startup to `httpd -d ./cfgov/apache -D FOREGROUND`, which runs Apache using
```

rosskarchner:
Since I can't push to @hkeeler's fork, I can't do the doc fixes

docker-compose.override.yml
hkeeler: Discussed with @chosak that we don't even need a separate volume here.  It's been removed, and confirmed that `PYTHONPATH` is still built correctly for `develop-apps`.

cfgov/apache/conf/httpd.conf
hkeeler: Fixed.

PR #5207, opened 2019-08-30
WIP: Prepaid agreements search 

virginiacc:
Adds ability to search and download prepaid agreement files.

## Additions

- Prepaid agreements app

## Testing

1. Check out branch, run migrations and gulp, and load agreements data
   ```
   python cfgov/manage.py runscript import_prepaid_agreements_data
   yarn run gulp
   ```
2. Turn on `PREPAID_AGREEMENTS_SEARCH` flag in admin
3. Navigate to [localhost:8000/prepaid-agreements](http://localhost:8000/prepaid-agreements) 

## Notes
This is an initial PR for work that is still in progress. Some outstanding issues:

- We are still refining the relationship between the filters and search form. Currently the filters are narrowed to only display values represented in the results when a search term is submitted, but not when a filter is applied.

- Submitting the search and filter forms currently involves a page reload. We intend to add Javascript form submission based on the existing eRegs search page code in a future PR.

- Tests are in progress, particularly for the search/filter functionality that is still being finalized, but we would still appreciate feedback in the meantime

## Todos

-

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


cfgov/unprocessed/css/search.less
```diff
@@ -0,0 +1,110 @@
+.search_wrapper {
```
virginiacc: The CSS in this file is pulled from the eregs app's `search.less` with some omissions. In a future PR I'll update eregs so both apps import this shared code, but I want to limit the external impacts of this initial PR to make it easier to review.

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
```diff
@@ -0,0 +1,100 @@
...
+        'href': '/data-research/prepaid-agreements/',
+        'title': 'Prepaid account agreements'
+    }]%}
+    {{ breadcrumbs.render(breadcrumb_items) }}
+
+    <div class="content_wrapper search_form">
+        <div class="block block__sub search_intro">
+            <h1>Prepaid account agreements database</h1>
+            <div class="lead-paragraph u-readable-width">
+                Find prepaid product agreements from more than 80 issuers. You can search agreements by issuer name, product name, prepaid account type, and other critera.
```
chosak: Are there any plans to make this "80" dynamic? Will this get out of date over time?

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
```diff
@@ -0,0 +1,134 @@
...
+    {{ product.name }} | Consumer Financial Protection Bureau
+{%- endblock title %}
+
+{% block hero %}
+<div class="wrapper content_wrapper">
+
+    {%- import 'molecules/breadcrumbs.html' as breadcrumbs with context -%}
+    {% set breadcrumb_items = [{
+        'href': '/prepaid-agreements/',
+        'title': 'Prepaid Product Agreements Database'
```
chosak: The title of the new index page shows up as "Prepaid Agreements Search" in the browser title bar (from the `<title>` tag), "Prepaid account agreements database" on the page itself, and "Prepaid Product Agreements Database" here in the link from the detail page. Could these be made consistent?

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
```diff
@@ -0,0 +1,100 @@
...
+{%- endblock css %}
+
+{# BODY items #}
+
+{% block content_main_modifiers -%}
+    content__1-3 prepaid-agreements prepaid-agreements-search
+{%- endblock %}
+
+{% block content_main scoped %}
+    {%- import 'molecules/breadcrumbs.html' as breadcrumbs with context -%}
```
chosak: Is this index page eventually intended to live under /data-research/prepaid-agreements/, instead of at /prepaid-agreements/?

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
```diff
@@ -0,0 +1,100 @@
...
+    {{ breadcrumbs.render(breadcrumb_items) }}
+
+    <div class="content_wrapper search_form">
+        <div class="block block__sub search_intro">
+            <h1>Prepaid account agreements database</h1>
+            <div class="lead-paragraph u-readable-width">
+                Find prepaid product agreements from more than 80 issuers. You can search agreements by issuer name, product name, prepaid account type, and other critera.
+            </div>
+            <p>Note that we are unable to do a full-text search of agreement files.</p>
+            <p>You may also choose to download all prepaid agreements as a <a href="https://files.consumerfinance.gov/a/assets/prepaid-agreements/bulk.zip">single compressed file</a> (1.3 GB, updated nightly).
```
chosak: Similarly, will this "1.3 GB" figure get out of date over time?

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
```diff
@@ -0,0 +1,134 @@
...
+            <dt><h2>Other relevant parties</h2></dt>
+            <dd>
+            {% if product.other_relevant_parties %}
+                {% if product.other_relevant_parties | length > 500 %}
+                    <div data-qa-hook="expandable"
+                         class="o-expandable">
+                        <div class="o-expandable_content">
+                {% endif%}
+                {{ product.other_relevant_parties | replace('\n', '; ') }}
+                {% if product.other_relevant_parties | length > 500 %}
```
anselmbradford: How to test showing this condition?

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
```diff
@@ -0,0 +1,100 @@
...
+                                   placeholder="Enter search term"
+                                   value="{{ query }}">
+                            <a class="a-btn a-btn__link" href="{{ remove_url_parameter(request, { 'q': [query] }) }}">{{ svg_icon('remove') }} Clear</a>
+                        </div>
+                    </div>
+                    <button class="a-btn flex-fixed">Search</button>
+                </div>
+            </form>
+            <p class="u-mt30 u-mb15">
+                <a href="">View frequently used terms</a>
```
chosak: What will this link do?

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
```diff
@@ -0,0 +1,134 @@
...
+                        </div>
+                        <button class="a-btn a-btn__link
```
anselmbradford: We should avoid modifying the expandable pattern without the use of a modifier class because it makes it harder to change the base expandable pattern in the future because this pattern is basically undocumented. Does this spot require `a-btn a-btn__link` be added?

cfgov/prepaid_agreements/scripts/import_prepaid_agreements_data.py
```diff
@@ -0,0 +1,71 @@
+from __future__ import unicode_literals
```
chosak: Not a criticism, just curious why this was implemented as a script to be run with `manage.py runscript` instead of as a Django management command to be run with e.g. `manage.py import_prepaid_agreements_data`.

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
```diff
@@ -0,0 +1,134 @@
...
+                {{ default_text }}
+            {% endif %}    
+            </dd>         
+        </dl>
+
+{% set agreements = product.prepaidagreement_set.all() %}
+{% if agreements | length %}
+
+    <div class="block u-mt45">
+        <h2 class="h3"> Download most recent agreement files</h2>
```
anselmbradford: Space before Download needed?

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
```diff
@@ -0,0 +1,100 @@
...
+        <div class="block block__sub search_intro">
+            <h1>Prepaid account agreements database</h1>
+            <div class="lead-paragraph u-readable-width">
+                Find prepaid product agreements from more than 80 issuers. You can search agreements by issuer name, product name, prepaid account type, and other critera.
+            </div>
+            <p>Note that we are unable to do a full-text search of agreement files.</p>
+            <p>You may also choose to download all prepaid agreements as a <a href="https://files.consumerfinance.gov/a/assets/prepaid-agreements/bulk.zip">single compressed file</a> (1.3 GB, updated nightly).
+        </div>
+
+        <div class="search_wrapper u-mt30">
```
anselmbradford: Do you need the margin utilities here and the next link or should `block` classes be used?

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
```diff
@@ -0,0 +1,100 @@
...
+                    {% if filter in valid_filters %}
+                        {% for val in filters[filter] %}
+                            <input type="hidden" name="{{ filter }}" value="{{ val }}">
+                        {% endfor %}
+                    {% endif %}
+                {% endfor %}
+                </div>
+
+                <h3 class="h4 u-mb5">Search within</h3>
+                <div class="layout-row u-mb30">
```
anselmbradford: Do you need `u-mb30` here or can `block` classes be used?

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
```diff
@@ -0,0 +1,100 @@
...
+                        <div class="m-btn-inside-input input-contains-label">
+                            <label for="searchText" class="input-contains-label_before input-contains-label_before__search">
+                                {{ svg_icon('search') }}
+                                <span class="u-visually-hidden">The term to search for</span>
+                            </label>
+                            <input id="searchText" 
+                                   type="text"
+                                   autocomplete="off"
+                                   class="a-text-input"
+                                   name="q"
```
anselmbradford: Ensure `q` doesn't collide with another search field on the page.

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
```diff
@@ -0,0 +1,100 @@
...
+                                   placeholder="Enter search term"
+                                   value="{{ query }}">
+                            <a class="a-btn a-btn__link" href="{{ remove_url_parameter(request, { 'q': [query] }) }}">{{ svg_icon('remove') }} Clear</a>
+                        </div>
+                    </div>
+                    <button class="a-btn flex-fixed">Search</button>
+                </div>
+            </form>
+            <p class="u-mt30 u-mb15">
```
anselmbradford: Can `block` classes be used here?

cfgov/prepaid_agreements/jinja2/prepaid_agreements/search_results.html
```diff
@@ -0,0 +1,174 @@
+{% import 'atoms/checkbox.html' as checkbox %}
+{% import 'atoms/radio-button.html' as radio %}
+{% import 'atoms/tag.html' as tag %}
+{% import 'search_result_item.html' as search_item %}
+{% import 'molecules/pagination.html' as pagination with context %}
+{% from 'organisms/expandable.html' import expandable with context %}
+
+<aside class="content_sidebar content__flush-top-on-small content__flush-sides-on-small filters">
+    <h3>Narrow results by...</h3>
```
anselmbradford: TRIVIAL: looks like technically it should be an ellipsis `…`

cfgov/prepaid_agreements/views.py
```diff
@@ -0,0 +1,119 @@
...
+        'query': search_term or '',
+        'active_filters': active_filters,
+        'valid_filters': valid_filters,
+        'search_field': search_field,
+    })
+
+
+def detail(request, product_id):
+    return render(request, 'prepaid_agreements/detail.html', {
+        'product': PrepaidProduct.objects.get(id=product_id)
```
chosak: This will trigger a `PrepaidProduct.DoesNotExist` exception if the requested ID does not exist in the database, causing a 500 when we probably want a 404 instead.

One approach would be to use the built-in [`get_object_or_404`](https://docs.djangoproject.com/en/1.11/topics/http/shortcuts/#get-object-or-404) like:

```py
get_object_or_404(PrepaidProduct, pk=product_id)
```

This will cause a 404 instead of raising an unhandled exception if the ID is invalid.

Another approach would be to replace this detail view entirely with the built-in Django [`DetailView`](https://docs.djangoproject.com/en/1.11/ref/class-based-views/generic-display/#detailview):

```py
class PrepaidProductDetailView(DetailView):
    model = PrepaidProduct
    template_name = 'prepaid_agreements/detail.html'

# then in urls.py
    ...
    url(r'^detail/(?P<pk>\d+)/$', PrepaidProductDetailView.as_view(), ...),
    ....

File: cfgov/prepaid_agreements/urls.py [hunk 17]
```diff
@@ -0,0 +1,9 @@
+from django.conf.urls import url
+
+from prepaid_agreements.views import detail, index
+
+
+urlpatterns = [
+    url(r'^$', index, name='index'),
```
chosak: The names `index` and `detail` here either should be changed to `prepaid_agreements_index` and `prepaid_agreements_detail` OR this urls.py file should be included [in the main file](https://github.com/cfpb/cfgov-refresh/pull/5207/files#diff-e45111e649c8f0371d5367b7351f1dd8R238) with a [namespace](https://docs.djangoproject.com/en/1.11/topics/http/urls/#url-namespaces) ([example](https://github.com/cfpb/cfgov-refresh/blob/48c3dfe6733e25dd584bc606f4ce73d8b9f93d0b/cfgov/cfgov/urls.py#L231)). Otherwise this will be defining the one single "index" URL for the entire site.

cfgov/prepaid_agreements/jinja2/prepaid_agreements/search_results.html
```diff
@@ -0,0 +1,174 @@
...
+
+    <div class="database-disclaimer">
+        <div class="o-well">
+            <h4>Database disclaimer</h4>
+            <p>We will display the prepaid agreements in this database as the respective issuers submitted them. The CFPB is not responsible for the content of the agreements, including any discrepancies between an agreement as presented in this database and the agreement as offered to the public, or for any omissions or other errors in the agreement as submitted by the issuer.</p>
+            <p>The agreements on file will have general terms and conditions, pricing, and fee information. They are not specific to an individual's account information.</p>
+            <p>If you have questions about the agreements themselves, contact the prepaid issuer directly.</p>
+        </div>
+    </div>
+</div>
```
anselmbradford: TRIVIAL: Add ending newline.

cfgov/unprocessed/apps/prepaid-agreements/css/main.less
```diff
@@ -0,0 +1,119 @@
+@import (reference) '../../../css/main.less';
+@import (less) '../../../css/search.less';
+@import (less) './search-bar.less';
+
+.u-readable-width {
+    max-width: 41.875rem;
```
anselmbradford: Add comment that this magic number matches the width set in cf-layout [here](https://github.com/cfpb/capital-framework/blob/0b15295c00ba1972676b6f4c86f3b40dcf45289d/packages/cf-layout/src/cf-layout.less#L252).

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
```diff
@@ -0,0 +1,134 @@
...
+        {% set agreement = agreements | first  %}
+        <div class="block block__sub block__flush-top">
+            <p class="u-mb5"><span>Agreement effective date:</span> {{ agreement.effective_date }}</p>
+            <p class="break-word">
```
anselmbradford: Should this be `u-break-word`?

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
```diff
@@ -0,0 +1,134 @@
...
+        </div>
+    </div>
+
+    {% if agreements | length > 1 %}
+    <div class="block u-mt45">
+        <h2 class="h3"> Download older agreements</h2>
+        {% for agreement in agreements[1:] %}
+            <div class="block block__sub {{ 'block__flush-top' if loop.index == 1 else '' }}">
+                <p class="agreement-effective-date"><span>Agreement effective date:</span> {{ agreement.effective_date }}</p>
+                <p class="break-word">
```
anselmbradford: Should this be `u-break-word`?

cfgov/unprocessed/apps/prepaid-agreements/css/main.less
```diff
@@ -0,0 +1,119 @@
...
+        display: block;
+        margin-left: 0;
+    }
+
+    dt {
+        letter-spacing: 0;
+        font-weight: normal;
+        text-transform: none;
+        font-size: inherit;
+        .heading-4;
```
anselmbradford: Let's do `.heading-4();` to better show this is a mixin.

cfgov/unprocessed/apps/prepaid-agreements/css/main.less
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    padding-top: 0 !important;
+
+    .filters {
+        &_applied {
+            padding: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
+        }
+        .o-multiselect_choices {
+            display: none;
```
anselmbradford: Would be more properly set up as a modifier on `o-multiselect`.

cfgov/unprocessed/apps/prepaid-agreements/css/main.less
```diff
@@ -0,0 +1,119 @@
...
+        .content_main {
+            width: 67% !important;
+        }
+    });
+
+    .search_wrapper {
+        label.h2 {
+            display: block;
+            max-width: none;
+            margin-bottom: unit( @grid_gutter-width * 2/3 / @base-font-size-px, em );
```
anselmbradford: Should be `2/3` here?

cfgov/unprocessed/apps/prepaid-agreements/css/main.less
```diff
@@ -0,0 +1,119 @@
...
+        margin-bottom: unit( @grid_gutter-width / 6 / @base-font-size-px, em );
+    }
+
+    dd {
+        margin-bottom: unit( @grid_gutter-width * 2/3 / @base-font-size-px, em );
+    }
+    
+    .o-expandable_content__collapsed {
+        // shows approx. 4 lines of content
+        max-height:  unit( @base-line-height-px * 4 / @base-font-size-px, em ) !important;
```
anselmbradford: Add comment of why `!important` is necessary.

cfgov/prepaid_agreements/jinja2tags.py
```diff
@@ -0,0 +1,36 @@
...
+    Discards should be a dict of lists:
+         {param: [values]}
+    """
+    query = request.GET.copy()
+    params = dict(iterlists(query))
+    for key in discards:
+        if key in params:
+            params[key] = [item for item in params[key]
+                           if item not in discards[key]]
+    querystring = urlencode(params, 'utf-8')
```
chosak: This line causes a 500 due to a `UnicodeEncodeError` if the params contain a non-ASCII character, for example with URLs like:

- http://localhost:8000/prepaid-agreements/?q=debit&issuer_name=unicod%C3%AB
- http://localhost:8000/prepaid-agreements/?prepaid_type=Travel&search_field=all&q=unicod%C3%AB

This can be repeated by entering a search term with non-ASCII characters while filtering.

cfgov/unprocessed/apps/prepaid-agreements/css/search-bar.less
```diff
@@ -0,0 +1,125 @@
...
+
+.respond-to-max( @bp-sm-max, {
+    .a-select {
+        margin-bottom: unit( @grid_gutter-width / 2 / @base-font-size-px, em );
+        width: 100%;        
+    }
+});
+
+ // select
+@select-height:     @size-ii;
```
anselmbradford: TRIVIAL: Space after colon not needed.

cfgov/unprocessed/apps/prepaid-agreements/css/search-bar.less
```diff
@@ -0,0 +1,125 @@
...
+    background-color: @gray;
+    flex-direction: column;
+    justify-content: center;
+
+    select {
+        padding: unit( 4px / @base-font-size-px, em )
+                 unit( 6px / @base-font-size-px, em );
+
+        // Save room for the drop-arrow
+        padding-right: unit( @select-height * 1.1 / @base-font-size-px, em );
```
anselmbradford: Possibly comment what the `1.1` magic number is for.

cfgov/prepaid_agreements/models.py
```diff
@@ -0,0 +1,40 @@
...
+
+class PrepaidAgreement(models.Model):
+    product = models.ForeignKey(PrepaidProduct, on_delete=models.CASCADE)
+    effective_date = models.DateField(blank=True, null=True)
+    compressed_files_url = models.TextField(blank=True, null=True)
+    bulk_download_path = models.TextField(blank=True, null=True)
+    filename = models.TextField(blank=True, null=True)
+
+    def __str__(self):
+        return 'IFL-' + str(self.pk)
```
chosak: What is the purpose of this `IFL-`? I don't see anywhere that it appears in the frontend. If this is needed, would it be possible to add a comment to explain?

cfgov/unprocessed/js/routes/prepaid-agreements/single.js
```diff
@@ -0,0 +1,13 @@
...
+
+const multiSelectsSelector = `.${ Multiselect.BASE_CLASS }`;
+
+if ( document.querySelector( multiSelectsSelector ) ) {
+  instantiateAll( multiSelectsSelector, Multiselect );
+}
+
+Expandable.init();
+
+
```
anselmbradford: TRIVIAL: Remove trailing newlines.

cfgov/prepaid_agreements/models.py
```diff
@@ -0,0 +1,40 @@
...
+    other_relevant_parties = models.TextField(blank=True, null=True)
+    status = models.TextField(blank=True, null=True)
+    withdrawal_date = models.DateField(blank=True, null=True)
+
+    def __str__(self):
+        return self.name
+
+    @property
+    def latest_agreement(self):
+        return self.prepaidagreement_set.order_by('-pk').first()
```
chosak: Is the assumption that the data being loaded is always ordered in the input file with the latest agreements last? In other words, this doesn't use the `effective_date` field of the agreement, but the order that it was loaded.

Additionally, is this `order_by` clause needed if `PrepaidAgreement` already has a default ordering in its `class Meta`?

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
```diff
@@ -0,0 +1,134 @@
...
+                    {{ agreement.filename }}
+                </a>
+            </p>
+        </div>
+    </div>
+
+    {% if agreements | length > 1 %}
+    <div class="block u-mt45">
+        <h2 class="h3"> Download older agreements</h2>
+        {% for agreement in agreements[1:] %}
```
chosak: Can you provide an example link to test a card with multiple agreements?

cfgov/prepaid_agreements/jinja2/prepaid_agreements/search_result_item.html
```diff
@@ -0,0 +1,38 @@
...
+            <h5 class="h4 u-mb5">Download most recent agreement files</h5>
+            {% set latest_agreement = result.latest_agreement %}
+            <p class="u-break-word">
+                <a href="{{ latest_agreement.compressed_files_url }}">
+                    {{ latest_agreement.filename }}
+                </a>
+            </p>
+            <div class="block block__sub">
+                <h5 class="h4 u-mb5">View all agreement files</h5>
+            <p><a href="detail/{{result.id}}">View all agreements for {{ result.name }}</a></p>
```
chosak: Formatting: could you please put spaces around `{{ result.id }}`?

Django's documentation [discourages](https://docs.djangoproject.com/en/1.11/ref/models/instances/#get-absolute-url) hardcoding model instance URLs like this, instead recommending implemeting a `get_absolute_url` method on the model class that `reverse`s the URL name specified in urls.py.

cfgov/cfgov/settings/base.py
```diff
@@ -89,6 +89,7 @@
     'wellbeing',
     'search',
     'paying_for_college',
+    'prepaid_agreements',
```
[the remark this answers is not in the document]
willbarton: `isort` will also sort for you:

```
isort --recursive cfgov
```

cfgov/prepaid_agreements/scripts/import_prepaid_agreements_data.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+S3_PATH = 'https://files.consumerfinance.gov/a/assets/prepaid-agreements/'
+METADATA_FILENAME = 'prepaid_metadata.json'
+
+
+def import_products_data(products_data):
+    for item in products_data:
+        pk = item['product_id'].replace('PRODUCT-', '')
+        product = PrepaidProduct.objects.filter(pk=pk).first()
```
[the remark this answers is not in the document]
richaagarwal: The product might not exist yet if it's a new product -- but yes I like the `update_or_create` option for this -- thanks for the suggestion!

cfgov/prepaid_agreements/models.py
richaagarwal: Yes, I'll add some docstrings here (and elsewhere more generally) to explain this... but basically the data is a little messy and there exist many agreements with the same `effective_date`, so we decided to use the ID as a proxy for which was uploaded most recently, since on Salesforce's end the IDs simply increment.  We're not generating our own IDs here, so they reflect the order in which they were created in Salesforce.  

Good point about the order_by - it's likely redundant and I can remove it.

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
```diff
@@ -0,0 +1,134 @@
...
+                    <div data-qa-hook="expandable"
+                         class="o-expandable">
+                        <div class="o-expandable_content">
+                {% endif%}
+                {{ product.other_relevant_parties | replace('\n', '; ') }}
+                {% if product.other_relevant_parties | length > 500 %}
+                        </div>
+                        <button class="a-btn a-btn__link
+                                        o-expandable_header
+                                        o-expandable_target">
```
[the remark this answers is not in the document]
virginiacc: Fixed in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/7371cd90fac1928b166a496101f3bb76ed858b96.

cfgov/unprocessed/apps/prepaid-agreements/css/main.less
virginiacc: Good point -- updated in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/7371cd90fac1928b166a496101f3bb76ed858b96. One thought I had in adding this is that it would be useful to set this value as a variable in cf-layout and/or move the utility class to cf-utility -- we use the number multiple times in various cfgov-refresh less files.

cfgov/unprocessed/js/routes/prepaid-agreements/single.js
virginiacc: Updated in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/7371cd90fac1928b166a496101f3bb76ed858b96

cfgov/unprocessed/apps/prepaid-agreements/css/search-bar.less
virginiacc: Updated in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/7371cd90fac1928b166a496101f3bb76ed858b96

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
virginiacc: Fixed in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/7371cd90fac1928b166a496101f3bb76ed858b96

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
virginiacc: Updated in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/7371cd90fac1928b166a496101f3bb76ed858b96

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
virginiacc: `q` is the standard field name we use for in-page search -- we use `query` for the main search in the header.

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
virginiacc: Updated in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/7371cd90fac1928b166a496101f3bb76ed858b96

cfgov/prepaid_agreements/jinja2/prepaid_agreements/search_results.html
virginiacc: Updated in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/7371cd90fac1928b166a496101f3bb76ed858b96

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
virginiacc: Thanks for catching that -- updated in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/7371cd90fac1928b166a496101f3bb76ed858b96

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
virginiacc: Fixed in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/7371cd90fac1928b166a496101f3bb76ed858b96

cfgov/unprocessed/apps/prepaid-agreements/css/main.less
virginiacc: Fixed in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/7371cd90fac1928b166a496101f3bb76ed858b96

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
virginiacc: This will link to a page with general info on searching the database / definitions of some of the metadata field terminology used on this page.

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
virginiacc: @chosak Yes, sorry for the lack of detail in the testing instructions! The [Apple Cash card](http://localhost:8000/prepaid-agreements/detail/46953/) has multiple agreements.

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
virginiacc: We don't expect this number to change significantly, so at this point it's probably not necessary to make it dynamic. We can revisit later if this changes, though.

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
virginiacc: @chosak Yes, we moved the prepaid urls under /data-research in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/64f2aa21bd21a6135639605695f90bd0385a7ce7.

cfgov/prepaid_agreements/views.py
richaagarwal: Thanks! This is fixed in https://github.com/cfpb/cfgov-refresh/commit/e5aea1f71c2ed02ba173aa6a5c7be1c7ed8795c8

cfgov/prepaid_agreements/scripts/import_prepaid_agreements_data.py
```diff
@@ -0,0 +1,71 @@
...
+        effective_date = item['effective_date']
+        if effective_date and effective_date != 'None':
+            effective_date = datetime.strptime(
+                effective_date, "%m/%d/%Y").date()
+        else:
+            effective_date = None
+
+        product_id = item['product_id'].replace('PRODUCT-', '')
+        product = PrepaidProduct.objects.get(pk=product_id)
+        url = S3_PATH + item['agreements_files_location']
```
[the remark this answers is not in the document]
richaagarwal: I thought about this but it ended resulting in passing the S3_PATH around to the templates in a few different places, which felt suboptimal.  It doesn't seem harmful to have an additional field to store this - the data is still portable in that we store the filename independently of the full URL as well. Additionally, triggering the Jenkins job to run - should the S3 path ever change - would be fairly trivial.

cfgov/cfgov/settings/base.py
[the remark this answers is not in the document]
richaagarwal: I can add it there, but is there a reason those are not currently in alphabetical order?  I'm having trouble understanding what value this configuration adds.

cfgov/prepaid_agreements/scripts/import_prepaid_agreements_data.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def run(*args):
+    source_url = S3_PATH + METADATA_FILENAME
+    resp = requests.get(url=source_url)
+    data = resp.json()
+    last_updated = datetime.strptime(
+        data['data_last_updated'], '%Y-%m-%d %H:%M:%S')
+    # Only import data if the data has been updated in the past 24 hours
+    if datetime.now() - last_updated <= timedelta(hours=24):
```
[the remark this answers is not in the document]
richaagarwal: Thanks for pointing this out!  It would probably be ideal if the timezone was passed along as well, but for the time being I've checked with the data team and it's an eastern timezone, so I've updated the script to process it as such.

cfgov/prepaid_agreements/urls.py
richaagarwal: Thanks for pointing this out! Fixed in https://github.com/cfpb/cfgov-refresh/commit/418c21c8291dad2f5ba17c285c4d1a6b838aab4c

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
richaagarwal: @sonnakim any thoughts here?

cfgov/cfgov/settings/base.py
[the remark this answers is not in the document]
chosak: The .isort.cfg file is used to configure how isort works [when we run it as part of tox](https://github.com/cfpb/cfgov-refresh/blob/d4b9b329e167b78a6060ae4238f930069829500c/tox.ini#L93). Part of the import ordering config is to group imports a certain way: first built-in Python things, then Django/Wagtail, then other third-party packages, and finally our own "first-party" packages in a separate section. If `prepaid_agreements` isn't included in that file, then imports would get sorted with the other third-party packages as opposed to with our own.

You correctly point out that [the current list](https://github.com/cfpb/cfgov-refresh/blob/48c3dfe6733e25dd584bc606f4ce73d8b9f93d0b/.isort.cfg#L13-L27) in that file isn't ordered; as far as I know that order shouldn't matter and so that should be fixed for better maintainability.

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
sonnakim: Yes - the outcome of this is sort of awaiting clearance, but let's unify for now: 

- Title tag: Prepaid Product Agreements Database
- H1: Prepaid product agreements database
- Detail page breadcrumb: Prepaid Product Agreements Database

@richaagarwal @virginiacc

cfgov/prepaid_agreements/views.py
```diff
@@ -0,0 +1,121 @@
...
+        'issuer_name',
+        'other_relevant_parties',
+        'name',
+        'program_manager',
+        'prepaid_type',
+    ]
+    if search_field and search_field[0] in search_fields:
+        search_fields = [search_field[0]]
+    products = products.annotate(
+        search=SearchVector(
```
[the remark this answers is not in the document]
richaagarwal: Hmm, yeah I just ran into this while writing tests for the search.   I'm a little conflicted on what to do - this feels like a test configuration issue, and usage of `SearchVector` seems like the way to go here, but I could also rewrite these queries to not be dependent on Postgres.  I'd prefer the former but that would involve a change in how we all run tests.  Curious what your thoughts are and if other BEWDs have a preference here -- @willbarton @higs4281 @schbetsy ?

richaagarwal:
Just an FYI we're in the middle of responding to feedback and adding tests so I'll post another comment here when it's ready to review again.

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
virginiacc: This should now be standardized as @sonnakim describes.

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
virginiacc: @anselmbradford That's a good point. This is a variation on the read more expandable pattern that's used on Ask CFPB, so in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/e7bee16fd73c52305cd044fe57535b5923ec0df7 I added another modifier (`.o-expandable__read-more-toggle`) for this version and moved its CSS to `read-more.less`. I also renamed the existing pattern to be slightly more descriptive (`.o-expandable__read-more-mobile`). (Totally open to other suggestions for approach or naming, though!)

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
virginiacc: This element was removed in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/2fcf0566ddcee6cb65a655addcf68ba1f4a5dc11.

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
virginiacc: Updated in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/eb20d39e7e59f558d5d3b9dfca9f91d38ac66070.

cfgov/unprocessed/apps/prepaid-agreements/css/main.less
virginiacc: Added to new location in `read-more.less` in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/487905184e94449340d7604af8f7cfed4d18201c

cfgov/prepaid_agreements/jinja2tags.py
virginiacc: @chosak Thanks for finding that issue! It should be fixed in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/60fbeca62dac8d65e736a0042c977a2c03d51193 (although the code may need some future cleanup).

cfgov/unprocessed/apps/prepaid-agreements/css/main.less
virginiacc: @anselmbradford That's a good point. I'm working on a couple of multiselect usability tweaks, so I can include a modifier for this in that PR.

cfgov/cfgov/settings/base.py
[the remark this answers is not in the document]
richaagarwal: Ooh ok, thanks for the detailed explanation!

I added it in this commit - https://github.com/cfpb/cfgov-refresh/pull/5207/commits/0d02553ca3dbc3d7f349cf500c3a283a9700e020 - alphabetizing it in the first part of the list that appears alphabetized.

cfgov/prepaid_agreements/views.py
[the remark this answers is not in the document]
richaagarwal: Based on feedback in a BEWD meeting today, I'll be leaving this code as-is and the tests will now need to be run against Postgres locally by passing in a TEST_DATABASE_URL like so: 

TEST_DATABASE_URL=postgres://django@localhost/cfgov tox -e fast

I will update the documentation.

cfgov/prepaid_agreements/models.py
```diff
@@ -0,0 +1,35 @@
...
+
+    class Meta:
+        ordering = ['name']
+
+
+class PrepaidAgreement(models.Model):
+    product = models.ForeignKey(PrepaidProduct, on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
richaagarwal: Thanks for the suggestion!  This has been updated.

cfgov/prepaid_agreements/models.py
richaagarwal: I've removed the order_by since it's redundant, though slightly worried about readability. 

I've also updated this to be based on the created_date instead of the agreement ID.  Unfortunately, effective_dates are not reliable on their own since issuers sometimes upload duplicates.

richaagarwal:
I am still getting through remaining unaddressed comments (& updating documentation), but am open to any additional feedback! -- @chosak @higs4281 @schbetsy @willbarton 

Also, I'm not sure why Travis is failing - if anyone has insights there.

richaagarwal:
I should also note that I don't have tests yet for the scripts added here - I thought about removing them from the PR, but the import script is needed for local testing.

The `write_prepaid_agreements_data_to_csv` script is possibly more in flux, but it's nice to have on this branch and it doesn't harm anything -- so if it's OK with people I will open a subsequent PR that adds tests / finalizes these two scripts, along with links to a corollary Jenkins PR.

Just don't want to block this current PR more than needed.

cfgov/prepaid_agreements/views.py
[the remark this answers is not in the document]
chosak: Is your intention to recommend that developers always use a local Postgres when running tests, or to suggest using Postgres only when you want to run these new specific tests?

If the former, it would be better to update [the existing `settings.test` code](https://github.com/cfpb/cfgov-refresh/blob/fb96d5ef92df6068a4659911fc350da6734aaa1e/cfgov/cfgov/settings/test.py#L4-L17) which currently defaults to a Sqlite DB, although this introduces some complexity see below).

Thinking about this some more, an even better option might be to leave the default test DB as Sqlite but mark these Postgres-specific tests with `@unittest.skipUnless` (like [this example](https://github.com/django/django/blob/193c109327c5216cab1d4eed6bfdff24629b09a3/tests/postgres_tests/__init__.py#L9) in the Django code) so that they only execute if being run against Postgres. This would have the advantage of making it easier to run tests without requiring users to set up Postgres, but still let us test against Postgres on Travis. The docs also [already suggest](https://cfpb.github.io/cfgov-refresh/python-unit-tests/#running-tests) that you can override `TEST_DATABASE_URL` if you want to test locally against Postgres.

Back to your specific proposal for `TEST_DATABASE_URL`, one problem is that we can't hardcode `localhost` as the host, because that won't work for people running tests in the Docker setup, where the Postgres host is named `postgres`. So this value would have to be different for the two different methods of running, which probably means adding `TEST_DATABASE_URL` to the [.env_SAMPLE](https://github.com/cfpb/cfgov-refresh/blob/master/.env_SAMPLE) as we have `DATABASE_URL`.

Because of the additional complexity this would add, I would prefer enabling the tests so that they run only when run against Postgres.

cfgov/prepaid_agreements/views.py
```diff
@@ -0,0 +1,121 @@
+from django.contrib.postgres.search import SearchVector
+from django.core.paginator import InvalidPage, Paginator
+from django.db.models import Q
+from django.shortcuts import render
+
+from prepaid_agreements.models import PrepaidProduct
+
+
+def validate_page_number(request, paginator):
```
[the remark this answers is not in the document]
virginiacc: Added in https://github.com/cfpb/cfgov-refresh/pull/5207/commits/ad685a4c942c8728191e513d482fdadf11a14243

cfgov/prepaid_agreements/views.py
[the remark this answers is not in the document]
richaagarwal: @chosak I didn't realize I could skip them like that -- that's perfect, thanks for the suggestion.  I've implemented this in https://github.com/cfpb/cfgov-refresh/commit/34ae7b516896972c1e6ddd0d6311d9eded8a9018

cfgov/prepaid_agreements/scripts/import_prepaid_agreements_data.py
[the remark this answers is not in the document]
richaagarwal: Updated this to use `update_or_create` in https://github.com/cfpb/cfgov-refresh/commit/0db7144a4cd4b8f0eacfd144423b9322c59e17e6

cfgov/jinja2/v1/_includes/atoms/tag.html
```diff
@@ -21,10 +21,11 @@
 
 {%- set behavior = 'data-js-hook=behavior_' ~ options.behavior if options.behavior else '' -%}
 {%- set value = 'data-value=' ~ options.value if options.value else '' -%}
+{%- set href = 'href=' ~ options.href if options.href else '' -%}
 
 <div class="a-tag"
      {{ value }}
      {{ behavior }}>
-    {{ options.label }} {{ svg_icon('remove') }}
+    {{ options.label }} {% if href %}<a {{href}}>{% endif %}{{ svg_icon('remove') }}{% if href %}</a>{% endif %}
```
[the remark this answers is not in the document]
virginiacc: This seems to work as expected for me -- on eregs, where tags are generated without an href option, the [link isn't added to tags](http://localhost:8000/policy-compliance/rulemaking/regulations/search-regulations/results/?regs=1026&q=term)

cfgov/prepaid_agreements/jinja2/prepaid_agreements/detail.html
virginiacc: Here's a product that meets the condition: [Global Cash Card Mastercard](http://localhost:8000/data-research/prepaid-accounts/search-agreements/detail/49606/)

PR #5208, opened 2019-09-03
YES tool — Hides checkboxes in a no-js environment

el-mapache:
~To check for supported JS environments, I'm checking for `classList`. The code makes use of passing multiple arguments to `classList.add`, which is not supported by IE 11. If there is a more standard feature to check for,  I'm happy to use that!~

Now using the `no-js` class modernizr adds to the `html` element.

## Additions

- Add `no-js` class to `main` element
- Add `u-js-only` classes to each check box.

## Removals

-

## Changes

-

## Testing

1. Either load the page using IE 11 or lower, or remove the check for `classList` on line 37 of this project's `index.html` file.

2. Each of the 'I'm not sure' checkboxes in the image below should be hidden:
Screen Shot 2019-08-26 at 4 45 31 PM

## Notes

-

## Todos

-

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/cfgov/jinja2/v1/youth_employment_success/index.html
+++ b/cfgov/jinja2/v1/youth_employment_success/index.html
@@ -34,7 +34,9 @@
 {% block javascript scoped %}
     {{ super() }}
     <script async>
-      if ( document.body.parentElement.className.indexOf( 'no-js' ) === -1 ) {
+      var htmlEl = document.documentElement;
+
+      if (htmlEl.className.indexOf('no-js') === -1) {
         !function(){
           {# Include site-wide JavaScript. #}
           var s = [
--- a/cfgov/unprocessed/apps/youth-employment-success/js/index.js
+++ b/cfgov/unprocessed/apps/youth-employment-success/js/index.js
@@ -29,7 +29,7 @@ const routeOptionForms = expandables.map( ( expandable, index ) => {
 } );
 
 /* only initialize the first form, the other gets initialized when
-   the user clicks 'add another option' button */
+  the user clicks 'add another option' button */
 routeOptionForms[0].init();
 routeOptionToggleView(
   document.querySelector( `.${ OPTION_TOGGLE_CLASSES.BUTTON }` ), {
--- a/cfgov/unprocessed/apps/youth-employment-success/js/input-view.js
+++ b/cfgov/unprocessed/apps/youth-employment-success/js/input-view.js
@@ -73,8 +73,16 @@ function InputView( element, props = {} ) {
    */
   function _bindEvents() {
     const { events = {}} = props;
+    const eventHandlers = [];
 
-    Object.entries( events ).forEach( ( [ event, handler = noop ] ) => {
+    for ( const event in events ) {
+      if ( events.hasOwnProperty( event ) ) {
+        const handler = events[event] || noop;
+        eventHandlers.push( [ event, handler ] );
+      }
+    }
+
+    eventHandlers.forEach( ( [ event, handler ] ) => {
       const handlerCache = eventsMap[event] || [];
       const delegate = eventHandler( handler );
       handlerCache.push( delegate );
@@ -89,7 +97,17 @@ function InputView( element, props = {} ) {
    * Unbind all event handlers from the input
    */
   function _unbindEvents() {
-    Object.entries( eventsMap ).forEach( ( [ event, handlers ] ) => {
+    const events = [];
+
+    for ( const event in eventsMap ) {
+      if ( eventsMap.hasOwnProperty( event ) ) {
+        const handlers = eventsMap[event];
+
+        events.push( [ event, handlers ] );
+      }
+    }
+
+    events.forEach( ( [ event, handlers ] ) => {
       handlers.forEach( handler => _dom.removeEventListener( event, handler ) );
     } );
   }
--- a/cfgov/jinja2/v1/youth_employment_success/route-questions/average-total-cost.html
+++ b/cfgov/jinja2/v1/youth_employment_success/route-questions/average-total-cost.html
@@ -32,7 +32,7 @@
   }}
   {{ 
     render_checkbox({
-      'class': 'block__sub-micro',
+      'class': 'u-js-only block__sub-micro',
       'label': "I'm not sure, add this to my to-do list to look up later",
       'name': "yes-miles-unsure",
       'disabled': true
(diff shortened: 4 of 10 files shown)


cfgov/unprocessed/apps/youth-employment-success/js/index.js
```diff
@@ -24,7 +23,7 @@ const routeOptionForms = expandables.map( ( expandable, index ) => {
 } );
 
 /* only initialize the first form, the other gets initialized when
-   the user clicks 'add another option' button */
+  the user clicks 'add another option' button */
```
anselmbradford: Did the autofix on the linter do this? Seemed better aligned, but fine if the linter did it.

cfgov/jinja2/v1/youth_employment_success/index.html
```diff
@@ -7,7 +7,7 @@
 
 {% block content %}
 
-<main class="content content__1-3" role="main">
+<main class="no-js content content__1-3" role="main">
```
anselmbradford: Is the `no-js`/`js` class on the `html` element not working how you want it?

el-mapache: I don't know how to add it to the `html` element when I'm extending an existing layout 😬 . All the content for the tool exists under the `main` element, so it seemed like an ok solution!

cfgov/unprocessed/apps/youth-employment-success/js/index.js
el-mapache: Yeah the linter would have done this, I didn't touch the documentation layout

el-mapache:
@anselmbradford `classList.add` is supported in IE11, but passing multiple arguments to it, e.g. `add('u-mt0', 'u-hidden')` is not. [See this link](https://caniuse.com/#search=classlist`)

el-mapache:
Having said that, I think we need to check for another feature to determine if we should disable js, thoughI'm not sure which.

cfgov/jinja2/v1/youth_employment_success/index.html
anselmbradford: You have the functionality of having it added to the html element already. (`js` class appears on `html`, or `no-js` in things like IE10). However, I'm wondering if it's not working how you need it to be working?

anselmbradford: Oh I see you've answered it below.

anselmbradford:
> classList.add is supported in IE11, but passing multiple arguments to it, e.g. add('u-mt0', 'u-hidden') is not.

I'm not totally clear… don't you need to support IE11? If you do need to support IE11, I would say refactor to add the classes individually. If not, then the structure you have is fine, but you're right only checking classList on document.body will pass in IE11.

el-mapache:
I guess I'm confused....doesn't the current codebase resort to serving just the no-js scripts in the case of IE 11? Or do all of your tools / apps / etc support IE 11?

anselmbradford:
> I guess I'm confused....doesn't the current codebase resort to serving just the no-js scripts in the case of IE 11? Or do all of your tools / apps / etc support IE 11?

IE11 should be supported. IE10 and below get served no-js, because of the check in https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/jinja2/v1/modernizr.html#L19-L21 (which isn't IE-specific, but checks features that legacy IE doesn't support, so is largely intended for those browsers).

el-mapache:
@anselmbradford Ok! I made some changes and tested this in VirtualBox, I _think_ the behavior is correct now.

cfgov/jinja2/v1/youth_employment_success/index.html
```diff
@@ -34,7 +34,10 @@
 {% block javascript scoped %}
     {{ super() }}
     <script async>
-      if ( document.body.parentElement.className.indexOf( 'no-js' ) === -1 ) {
+      var htmlEl = document.documentElement;
+
+      if (htmlEl.className.indexOf('no-js') < 0) {
```
anselmbradford: It looks like the feature detection check has been removed, so I'm still not clear why this is needed at all? You should be able to target styles via the no-js on the `html` element with, e.g.:

```
.no-js .my-custom-class { /* thing to do when in the no-js state */ }

.js .my-custom-class { /* thing to do when in the js state */ }
```

el-mapache: Don't we not want to load the custom JS for this tool in the case of an unsupported environment?

anselmbradford: Oh wait, I see `document.documentElement` IS the `html` element, so this should be good. I was thinking you were still targeting `main`. This is probably a good change to do overall. I was thinking we targeted the parentElement of body for compatibility reasons, but it doesn't look like we need to https://developer.mozilla.org/en-US/docs/Web/API/Document/documentElement#Browser_compatibility

Why check if the index is less than zero versus checking if it's -1?

cfgov/jinja2/v1/youth_employment_success/index.html
```diff
@@ -34,7 +34,10 @@
...
+        console.log('HELLO')
```
anselmbradford: Remove console log.

cfgov/unprocessed/apps/youth-employment-success/js/input-view.js
```diff
@@ -73,8 +73,16 @@ function InputView( element, props = {} ) {
    */
   function _bindEvents() {
     const { events = {}} = props;
+    const eventHandlers = [];
 
-    Object.entries( events ).forEach( ( [ event, handler = noop ] ) => {
+    for(let event in events) {
```
anselmbradford: Run linter autofix to fix spacing.

cfgov/jinja2/v1/youth_employment_success/index.html
el-mapache: I guess slightly less characters? Originally I didn't have the `htmlEl` as a separate variable, everything was inline, and I think the resultant line was longer than 80 characters 😄

PR #5212, opened 2019-09-04
Save a full list of housing counselor results

schbetsy:
The Office of Innovation needs the Bureau to retain a copy of each day's list of HUD-approved housing counselors.

## Additions

- Add a step to the `hud_generate_json` management command that takes the full set of results and saves them to a .zip file.
- Add a test for this functionality
- Update public documentation of the Housing Counselor data processing

## Testing

1. `cfgov/manage.py hud_generate_json jsons --archive-file-name ./test-archive.zip`

## Checklist

### DIFF
--- a/cfgov/housing_counselor/results_archiver.py
+++ b/cfgov/housing_counselor/results_archiver.py
@@ -1,0 +1,17 @@
+import json
+from zipfile import ZIP_DEFLATED, ZipFile
+
+
+json_file_name = 'HUD_approved_housing_counselors.json'
+
+
+# By request from stakeholders in the Office of Innovation, save a
+# list of the full results we receive from the HUD API, in case of
+# future enforcement actions.
+
+# As an MVP, we're saving the complete JSON response, zipped. As a
+# future enhancement, we could consider improving the format of the
+# data, e.g. as CSV.
+def save_list(counselors, path):
+    with ZipFile(path, 'w', ZIP_DEFLATED, allowZip64=True) as zip_file:
+        zip_file.writestr(json_file_name, json.dumps(counselors))
--- a/cfgov/housing_counselor/tests/management/test_results_archiver.py
+++ b/cfgov/housing_counselor/tests/management/test_results_archiver.py
@@ -1,0 +1,26 @@
+import json
+import os
+import shutil
+from tempfile import mkdtemp
+from unittest import TestCase
+from zipfile import ZipFile
+
+from housing_counselor.results_archiver import json_file_name, save_list
+
+
+class TestHousingCounselorResultsArchiver(TestCase):
+    def setUp(self):
+        self.temp_directory = mkdtemp()
+        self.addCleanup(shutil.rmtree, self.temp_directory)
+
+    def test_save_list_creates_zip_file(self):
+        content = [
+            {"fake": 1, "housing": "counselor", "json": 2},
+            {"fake": 3, "housing": "data", "json": 4},
+        ]
+        zip_path = os.path.join(self.temp_directory, 'archive.zip')
+
+        save_list(content, zip_path)
+        with ZipFile(zip_path) as zip_file:
+            zip_content = json.loads(zip_file.read(json_file_name))
+            self.assertEqual(zip_content, content)
--- a/cfgov/housing_counselor/management/commands/hud_generate_json.py
+++ b/cfgov/housing_counselor/management/commands/hud_generate_json.py
@@ -10,6 +10,7 @@ from housing_counselor.generator import generate_counselor_json
 from housing_counselor.geocoder import (
     GazetteerZipCodeFile, GeocodedZipCodeCsv, geocode_counselors
 )
+from housing_counselor.results_archiver import save_list
 
 
 logger = logging.getLogger(__name__)
@@ -24,10 +25,13 @@ class Command(BaseCommand):
                             help='CSV containing zipcode lat/lngs')
         parser.add_argument('--zipcode-gazetteer-file',
                             help='Census Gazetteer zipcode file')
+        parser.add_argument('--archive-file-name',
+                            help='Archive file output path', required=True)
 
     def handle(self, *args, **options):
         zipcode_csv_file = options['zipcode_csv_file']
         zipcode_gazetteer_file = options['zipcode_gazetteer_file']
+        archive_file_name = options['archive_file_name']
 
         if zipcode_csv_file:
             zipcodes = GeocodedZipCodeCsv.read(zipcode_csv_file)
@@ -42,6 +46,9 @@ class Command(BaseCommand):
 
         # Retrieve counselors from the HUD website.
         counselors = fetch_counselors()
+
+        # Save the full list, for archive purposes
+        save_list(counselors, archive_file_name)
 
         # Standardize formatting of counselor data.
         counselors = clean_counselors(counselors)
--- a/docs/housing-counselor-tool.md
+++ b/docs/housing-counselor-tool.md
@@ -62,9 +62,10 @@ This step calls the [`hud_generate_json`](https://github.com/cfpb/cfgov-refresh/
 which performs the following steps:
 
   1. fetch agency listings from HUD
-  2. clean the results
-  3. fill in any missing latitude and longitude values
-  4. create files of the 10 nearest results for each U.S. ZIP code
+  2. save a copy of the full results, for our records
+  3. clean the results
+  4. fill in any missing latitude and longitude values
+  5. create files of the 10 nearest results for each U.S. ZIP code
 
 
 #### Fetch agency listings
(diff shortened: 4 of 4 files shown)


cfgov/housing_counselor/enforcement_list_saver.py
```diff
@@ -0,0 +1,24 @@
...
+# future enforcement actions.
+
+# As an MVP, we're saving the raw JSON response, zipped. As a future
+# enhancement, we could consider improving the format of the data,
+# e.g. as CSV.
+def save_list(counselors, destination_path):
+    with open(json_file_path, 'w') as f:
+        f.write(json.dumps(counselors))
+
+    if not destination_path:
```
chosak: Is there an advantage to making this argument optional, instead of making it required? It'd simplify the code to make it always required, and avoid the need to hardcode a default that won't be used in practice.

cfgov/housing_counselor/tests/management/test_enforcement_list_saver.py
```diff
@@ -0,0 +1,18 @@
+import os
+from unittest import TestCase
+
+from housing_counselor.enforcement_list_saver import save_list
+
+
+class TestEnforcementListSaver(TestCase):
+    file_path = './zip_destination.zip'
+
+    def tearDown(self):
```
chosak: A potentially cleaner alternative to this would be to use [`tempfile.TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory), like:

```py
with TemporaryDirectory() as tempdir:
    zipfile = os.path.join(tempdir, 'archive.zip')
    save_list(content, zipfile)
    self.assertTrue(os.path.exists(zipfile))
```

This takes care of cleanup and guarantees that the file won't exist at the start of the test, avoiding the need for the `assertFalse` call.

cfgov/housing_counselor/enforcement_list_saver.py
schbetsy: If the call to the `hud_generate_json` management command doesn't include a value for the `--full-file-name` argument, then the `full_file_name` variable ends up having a value of `None`. In order to use the `save_list` function with an optional parameter and a default value, I'd have to call the method like this:

```python
        if full_file_name:
            save_list(counselors, full_file_name)
        else:
            save_list(counselors)
```
Is that what you're suggesting? I'm probably misunderstanding your comment.

cfgov/housing_counselor/tests/management/test_enforcement_list_saver.py
schbetsy: `TemporaryDirectory` isn't available in [Python 2.7](https://docs.python.org/2.7/library/tempfile.html#tempfile.TemporaryDirectory), unfortunately.  I've pushed an update that uses a lower-level method in the same library. Does that look better?

cfgov/housing_counselor/enforcement_list_saver.py
chosak: Is there a good reason to make the `--full-file-name` parameter optional? Adding `required=True` to its definition would make it required, preventing the script from running if it isn't provided.

If saving the full archive isn't optional -- meaning that it always gets saved regardless -- then I think it's reasonable to require an output filename.

schbetsy:
1. I see your point. I've pushed an update renaming the file to use the word "archive" rather than "enforcement". However, I've left my explanatory comment mentioning enforcement above the `save_list` method, because I think it's important for future developers to know why the feature was added (so they can know how to maintain it). (cc @Scotchester, this seems like a compromise in the discussion we had privately)

2. My use of the word "raw" was a mistake. I've updated my comment to say "save the _complete_ JSON response". It is important to our stakeholders that this be a raw, unprocessed archive of the HUD results. They just want the list to be complete.

cfgov/housing_counselor/tests/management/test_enforcement_list_saver.py
chosak: Ah, that is too bad. Unfortunately, in that case, the `os.remove` calls probably should be in the test `tearDown`, as you had originally. Otherwise, if the assert statement fails, the test will abort, and those commands won't get run. If they're in `tearDown`, they'll happen no matter what.

Another option that's compatible with Python 2.7 is to use `addCleanup`:

```py
class Tests(TestCase):
    def setUp(self):
        self.tempdir = tempfile.mkdtemp()
        self.addCleanup(shutil.rmtree, self.tempdir)

    def testSomething(self):
        zipfile = os.path.join(self.tempdir, 'archive.zip')
        ...
```

chosak:
>  It is _not_ important to our stakeholders that this be a raw, unprocessed archive of the HUD results; they just want the list to be complete.

:+1: with your edit to add _not_ to that sentence 😄

cfgov/housing_counselor/enforcement_list_saver.py
schbetsy: Added an update to give that parameter a better name and make it required. Thanks for the suggestion.

cfgov/housing_counselor/tests/management/test_enforcement_list_saver.py
schbetsy: Thanks for the suggestion. Updated to include this.

schbetsy:
I'd love to get this merged and deployed today if possible, so we can start collecting daily data before our stakeholder's deadline early next week. Any other review?

Scotchester:
@schbetsy I'll trust Andy's review :)

cfgov/housing_counselor/results_archiver.py
```diff
@@ -0,0 +1,21 @@
...
+# As an MVP, we're saving the complete JSON response, zipped. As a future
+# enhancement, we could consider improving the format of the data,
+# e.g. as CSV.
+def save_list(counselors, destination_path):
+    with open(json_file_path, 'w') as f:
+        f.write(json.dumps(counselors))
+
+    zip = ZipFile(destination_path, 'w', ZIP_DEFLATED, allowZip64=True)
+    zip.write(json_file_path)
+    zip.close
```
chosak:
```suggestion
    zip.close()
```

This is missing the parenthesis, so isn't actually calling the `close` method. Another way of doing this would be to use a context manager like

```py
with ZipFile(...) as zf:
    zf.write(json_file_path)
```

Additionally, [`zip`](https://docs.python.org/3/library/functions.html#zip) is a built-in method in Python so it might be better to use a different variable name.

cfgov/housing_counselor/results_archiver.py
```diff
@@ -0,0 +1,21 @@
...
+# By request from stakeholders in the Office of Innovation, save a
+# list of the full results we receive from the HUD API, in case of
+# future enforcement actions.
+
+# As an MVP, we're saving the complete JSON response, zipped. As a future
+# enhancement, we could consider improving the format of the data,
+# e.g. as CSV.
+def save_list(counselors, destination_path):
+    with open(json_file_path, 'w') as f:
+        f.write(json.dumps(counselors))
```
chosak: One more minor suggestion: won't this code unnecessarily leave behind the HUD_approved_housing_counselors.json file on disk? You could use [`ZipFile.writestr`](https://docs.python.org/2/library/zipfile.html#zipfile.ZipFile.writestr) to directly write the JSON with something like this, avoiding the need to generate that temporary uncompressed file:

```py
zip.writestr(json_file_path, json.dumps(counselors))
```

cfgov/housing_counselor/tests/management/test_results_archiver.py
```diff
@@ -0,0 +1,19 @@
...
+class TestHousingCounselorResultsArchiver(TestCase):
+    def setUp(self):
+        self.temp_directory = mkdtemp()
+        self.addCleanup(shutil.rmtree, self.temp_directory)
+
+    def test_save_list_creates_zip_file(self):
+        content = '{"fake": 1, "housing": "counselor", "json": 2}'
+        zip_path = os.path.join(self.temp_directory, 'archive.zip')
+
+        save_list(content, zip_path)
```
chosak: I noticed that, although this test works, it doesn't exactly do what it seems to at first. The `save_list` method is expecting a list, which it then `json.dumps` and writes, and this test is instead passing in a string already. The call to `json.dumps` will work, it'll just write it as a JSON string (`'"{\\"fake\\": 1, \\"housing\\": \\"counselor\\", \\"json\\": 2}"'`).

Not a huge thing, but if you also wanted to test the JSON dumping functionality, you could re-read the file and confirm the contents:

```py
content = {"fake": 1, "housing": "counselor", "json": 2}

...

with ZipFile(zip_path) as zf:
    self.assertEqual(
        json.loads(zf.read('HUD_approved_housing_counselors.json')),
        content
    )
```

cfgov/housing_counselor/results_archiver.py
schbetsy: Thanks for that. I missed that part of the ZipFile documentation. Updated.

cfgov/housing_counselor/results_archiver.py
schbetsy: Nice. Added.

cfgov/housing_counselor/tests/management/test_results_archiver.py
schbetsy: I originally wrote this test to be implementation-neutral, i.e., If we decide on a different way of saving the archive file, the test would still pass. But that wasn't very useful, nor was it as flexible as I'd hoped. I've updated it to use your suggestion.

PR #5214, opened 2019-09-04
[WIP] Add redirects for broken links

cwdavies:
Add redirects for broken links, that includes policy-compliance pages and the old Ask CFPB search URL

### DIFF
--- a/cfgov/ask_cfpb/views.py
+++ b/cfgov/ask_cfpb/views.py
@@ -218,4 +218,4 @@ def redirect_ask_search(request, language='en'):
                         '+', '_')
                     return redirect_to_tag(tag, language)
 
-        return redirect('/ask-cfpb/search')
+        return redirect('/ask-cfpb/search/')
--- a/cfgov/ask_cfpb/views.py
+++ b/cfgov/ask_cfpb/views.py
@@ -163,7 +163,7 @@ def redirect_ask_search(request, language='en'):
     else:
         facets = request.GET.getlist('selected_facets')
         if not facets or not facets[0]:
-            return redirect('/ask-cfpb/search')
+            return redirect('/ask-cfpb/search/')
 
         def redirect_to_category(category, language):
             if language == 'es':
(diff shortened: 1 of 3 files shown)


cfgov/ask_cfpb/views.py
```diff
@@ -163,7 +163,7 @@ def redirect_ask_search(request, language='en'):
     else:
         facets = request.GET.getlist('selected_facets')
         if not facets or not facets[0]:
-            raise Http404
+            return redirect('/ask-cfpb/search')
```
Scotchester: @richaagarwal or @higs4281 Do you know if there was a specific reason that this was set to raise a 404 instead of redirecting to the basic search page with no query? Any objection to changing that? (Here and L221 below.)

```suggestion
            return redirect('/ask-cfpb/search/')
```

@cwdavies Adding the trailing slash will save the app the work of adding it after the request (here and L221 below).

cfgov/cfgov/urls.py
```diff
@@ -95,39 +96,96 @@ def empty_200_response(request, *args, **kwargs):
...
+    # Temporarily serve Wagtail OAH journey pages at `/process/` urls
+    # TODO: change to redirects after 2018 homebuying campaign
     url(r'^owning-a-home/process/(?P<path>.*)$',
         lambda req, path: ServeView.as_view()(
             req, 'owning-a-home/{}'.format(path or 'prepare/')
         ),
     ),
     # END TODO
 
+    url(r'^how-to-prepare-your-finances-for-buying-a-home/',
```
chosak: Is there any particular reason why these are being added as redirects here instead of in Apache?

cwdavies: I added redirects into urls.py as there were already redirects in this file, and I could easily verify the redirects were working locally using a Docker container. When I found out there was a redirects.conf for Apache, I put the redirects into redirects.conf. Would you like me to remove the redirects I added from the urls.py file?

chosak: Yes, thanks. Generally if we are creating permanent redirects that don't require any Python logic, they should be placed in the Apache redirects file. Sometimes we'll put redirects into the urls.py temporarily, or initially behind a feature flag that gets removed later, and neglect to move that now-permanent redirect to Apache.

cwdavies: Thanks for clarifying and taking time to review. I will remove permanent redirects that don't require any Python logic from urls.py.

Scotchester:
Superseded by #5461.

PR #5223, opened 2019-09-09
Fix download on rural or underserved tool

wpears:
## Changes

- Uses the `table` tag rather than the `.table` class for the download selector. Looks like this was changed in the move to wagtail.

## Testing

1. Pull branch, build.
2. Enter an address, hit "Check addresses"
3. Click "Download results"

The resulting file should be populated with data. Following these steps on cf.gov will result in a file with only headers.

### DIFF
--- a/cfgov/unprocessed/apps/rural-or-underserved-tool/js/common.js
+++ b/cfgov/unprocessed/apps/rural-or-underserved-tool/js/common.js
@@ -400,7 +400,7 @@ function generateCSV() {
   }
 
   // loop through each row
-  [].slice.call( DT.getEls( 'table tbody tr td' ) )
+  [].slice.call( DT.getEls( '.rout-results-table tbody tr td' ) )
     .forEach( _loopHandler );
 
   return theCSV;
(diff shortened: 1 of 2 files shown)


cfgov/unprocessed/apps/rural-or-underserved-tool/js/common.js
```diff
@@ -400,7 +400,7 @@ function generateCSV() {
   }
 
   // loop through each row
-  [].slice.call( DT.getEls( '.table tbody tr td' ) )
```
anselmbradford: This was meant to be `.o-table`. It was migrated with the original [markup](https://github.com/cfpb/rural-or-underserved-test/blob/master/src/index.html#L746), but then the table got migrated to `o-table` during the [course of the PR](https://github.com/cfpb/cfgov-refresh/pull/5092/commits/e9a816701e19fc68c9251fbb01ee121cd436bc8b) and this was missed in the update. Do we want to target `table` or `.o-table` (which is assigned to the parent div in scrollable tables). 

OR do we want to give the address tables a custom class so that if this page has other tables in the future they won't get pulled in?

wpears: I'd think a custom class or id would be best

anselmbradford: Probably a custom class, because there are multiple tables.

PR #5228, opened 2019-09-11
Ask CFPB: Clean up unused portions of ask pages

anselmbradford:
## Removals

- Remove unused ask category page.
- ~~Remove references to nonexistent `ask-search-bar.html` template in .po file.~~
- Remove empty `content` section of Ask results page.

## Testing

1. Pull branch, run migrations.
2. Create a new Ask Results Page type and see that there is only a title, no empty content area.

## Screenshots

Before:
Screen Shot 2019-09-11 at 12 42 14 PM

After:
Screen Shot 2019-09-11 at 12 57 04 PM


cfgov/locale/es/LC_MESSAGES/django.po
```diff
@@ -188,25 +188,6 @@ msgstr "Sí"
 msgid "optional"
 msgstr "opcional"
 
-#: v1/_includes/ask/ask-search-bar.html
```
virginiacc: I think most (maybe all) of these translations are still needed, they're just no longer scoped to the ask-search-bar template.

anselmbradford: So would I just be deleting line 191, or updating it to the `ask-search.html` template?

anselmbradford: Oh wait, I see this is an autogenerated file and that's just a comment. Running https://github.com/cfpb/cfgov-refresh/blob/master/docs/translation.md#2-run-the-makemessages-management-command-to-add-the-string-to-the-portable-object-file I see a whole bunch changed. I'll revert this change and leave it for another PR.

PR #5240, opened 2019-09-13
Refactor ask search to include icon and responsive styling

anselmbradford:
## Removals

- Remove `input-with-btn__bar` class that was setting input height to 35px.

##  Changes

- Refactor ask search to include icon and responsive styling

## Testing

1. Log into local Wagtail.
1. Create a new sublanding page.
1. Add a "Full width text" 
1. Add a "Well with ask search"
1. Fill in the options and Preview the page. Check at mobile and desktop.
1. Check search fields in Ask CFPB search (/ask-cfpb/) at mobile and desktop.

## Screenshots

Desktop

Before (bottom)
After (top)

Screen Shot 2019-09-13 at 5 18 12 PM

Mobile

Before (bottom)
After (top)

Screen Shot 2019-09-13 at 5 18 20 PM

### DIFF
--- a/cfgov/unprocessed/css/organisms/ask-search.less
+++ b/cfgov/unprocessed/css/organisms/ask-search.less
@@ -3,10 +3,6 @@
    Ask Search organism
    ========================================================================== */
 
-.o-search-bar_input {
-    max-width: 500px;
-}
-
 // TODO: Reconcile this with the input with button in Capital Framework.
 .o-form__input-w-btn_btn-container {
 
--- a/cfgov/unprocessed/css/organisms/ask-search.less
+++ b/cfgov/unprocessed/css/organisms/ask-search.less
@@ -5,11 +5,11 @@
 
 // TODO: Reconcile this with the input with button in Capital Framework.
 .o-form__input-w-btn_btn-container {
-
     // Only attach the button to the input at tablet/desktop sizes.
     .respond-to-min( 480px, {
         border-left: 0;
-        & .a-btn {
+
+        .a-btn {
             border-top-left-radius: 0;
             border-bottom-left-radius: 0;
         }
--- a/cfgov/unprocessed/css/organisms/ask-search.less
+++ b/cfgov/unprocessed/css/organisms/ask-search.less
@@ -3,6 +3,19 @@
    Ask Search organism
    ========================================================================== */
 
-.input-with-btn__bar .input-with-btn_btn button {
-    height: 35px;
+.o-search-bar_input {
+    max-width: 500px;
+}
+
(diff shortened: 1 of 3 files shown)


cfgov/unprocessed/css/organisms/ask-search.less
```diff
@@ -3,6 +3,15 @@
...
+
+    // Only attach the button to the input at tablet/desktop sizes.
+    .respond-to-min( 480px, {
+        border-left: 0;
+        & .a-btn {
+            border-top-left-radius: 0;
+            border-bottom-left-radius: 0;
+        }
+    } );
+}
```
Scotchester: Trivial: Add blank line at EOF.

cfgov/unprocessed/css/organisms/ask-search.less
```diff
@@ -3,6 +3,15 @@
...
-.input-with-btn__bar .input-with-btn_btn button {
-    height: 35px;
-}
+// TODO: Reconcile this with the input with button in Capital Framework.
+.o-form__input-w-btn_btn-container {
+
+    // Only attach the button to the input at tablet/desktop sizes.
+    .respond-to-min( 480px, {
+        border-left: 0;
+        & .a-btn {
```
Scotchester: Trivial: The ampersand here isn't necessary, I don't think. And I think we usually put a line between standard properties and nested selectors.

```suggestion

        .a-btn {
```

cfgov/unprocessed/css/organisms/ask-search.less
```diff
@@ -3,6 +3,15 @@
    Ask Search organism
    ========================================================================== */
 
-.input-with-btn__bar .input-with-btn_btn button {
-    height: 35px;
-}
+// TODO: Reconcile this with the input with button in Capital Framework.
+.o-form__input-w-btn_btn-container {
+
```
Scotchester:
```suggestion
```

cfgov/unprocessed/css/organisms/ask-search.less
anselmbradford: I'll remove it since this block is so small, but we may want to favor including it for clarity. I just got tripped up with https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/css/pages/ask.less#L239, because I didn't pick up right away that it was actually `.ask-cfpb-page__answer .o-search-bar_input { … }` and not `.o-search-bar_input { … }`. The indent should show that, but when every rule viewable in a window is at the same indent it's not as clear. If we included `&` it would always be clear there's a placeholder for the parent class.

Scotchester: That's a good point!

PR #5249, opened 2019-09-17
Disable acceptance tests temporarily in Travis

willbarton:
This disables the acceptance tests that depend on WebDriver and Chrome. Once those issues are resolved we can revert this change, but this will ensure we don't have to keep disabling Travis to merge PRs.

@anselmbradford suggested running `test:unit` as a way to do this.

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [ ] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [x] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #5257, opened 2019-09-24
Move cfgov/unprocessed to folder to repository root

chosak:
Currently many input files for the frontend build live under cfgov/unprocessed. These files are processed during the build, generating output that lives in cfgov/static_built. Once the build happens, the unprocessed files should no longer be needed to run cf.gov.

The cf.gov build and deploy process packages up everything under cfgov/ into the deployable zipfile, including the unprocessed folder. This ships unnecessary files to the web servers, makes the deploy archives significantly larger than they need to be, and makes the build and deploy take longer because of the additional files.

This PR moves cfgov/unprocessed to the repository root at /unprocessed, and attempts to update all references to unprocessed files, including in documentation.

## Testing

I've verified that the frontend build works and that all backend and frontend tests pass. Browser acceptance tests are currently broken/disabled for this repository (see #5249) and so it's not possible to test those right now.

I've generated a deployable zipfile using `docker/deployable-zipfile/build.sh` and verified that it works properly using our internal Ansible process.

By way of comparison, a deployable zipfile generated with master is 13,641 files, 205 MB uncompressed, 130 MB compressed. A zipfile generated after this change is only 4,517 files, 110 MB uncompressed, 97 MB compressed. This is because the unprocessed folder contains 8,320 files and is 121 MB (I'm not sure why these numbers don't add up exactly).

## Notes

I started working on this change as part of the (now-closed) ShipItBetter effort so wanted to push it up to get feedback/acceptance if FEWDs or others thing this would be a good change.

The one piece of this I'm not sure about is the npm-packages-offline-cache subdirectories of the unprocessed apps. Is there anything extra that needs to be done with this?

## Checklist


anselmbradford:
High-level this seems like a fine change to me. One possible change is naming it `cfgov-unprocessed`, if that makes organizational sense to do so.

Scotchester:
This seems like real nice win for our deployment efficiency!

While I don't want to delay that win, I do wonder if we should consider alternatives that might allow us to maintain the more traditional Django structure of having the static asset source files in the app folder. For example, could we just ignore the `unprocessed` folder when building the deploy artifact?

chosak:
> maintain the more traditional Django structure of having the static asset source files in the app folder

This is a good point @Scotchester.

> could we just ignore the unprocessed folder when building the deploy artifact?

Yes, that would require modifying [this line](https://github.com/cfpb/cfgov-refresh/blob/69fd37433c06a0809158d71650125d72180b7333/cfgov/deployable_zipfile/create.py#L87) to ignore that folder when the source tree is copied over. Ideally that'd be provided as an argument to the existing command-line script, to avoid hardcoding it.

As this PR is already out-of-date, and will continue to become even more so as unprocessed apps are developed, I'm inclined to close this in favor of ignoring as a less disruptive solution.

PR #5265, opened 2019-09-26
Set up Voluntary Diversity Assessment Form

Scotchester:
We need a new form on the website for entities to voluntarily register for a diversity assessment. This commit starts to set that up.

GHE issue: CFGOV/platform#3557

## Additions

- New `diversity_inclusion` app
- Basic Django form setup for the voluntary diversity assessment registration form
- A bare-bones template

## Testing

1. Pull branch
1. Run server
1. Visit http://localhost:8000/about-us/diversity-and-inclusion/voluntary-assessment-form/

## Todos

- [x] Add in custom layout markup
- [x] Add in the rest of the content surrounding the form
- [x] Replace sidebar content
- [x] Update Python unit tests

## Checklist


cfgov/diversity_inclusion/views.py
```diff
@@ -0,0 +1,28 @@
+from django.core.urlresolvers import reverse
+from django.http import HttpResponse, HttpResponseRedirect
+from django.shortcuts import render
+from django.views.generic import TemplateView
+
+from forms import VoluntaryAssessmentForm
+
+
+class GetName(TemplateView):
```
chosak: Could this be simplified with use of a [`FormView`](https://docs.djangoproject.com/en/1.11/ref/class-based-views/generic-editing/#formview) instead, something like:

```py
class GetName(FormView):
    template_name = 'diversity_inclusion/voluntary-assessment-form.html'
    form_class = VoluntaryAssessmentForm
    success_url = reverse_lazy('diversity_inclusion:form_submitted')
```

cfgov/cfgov/urls.py
```diff
@@ -381,6 +381,10 @@ def empty_200_response(request, *args, **kwargs):
         r'^(?i)consumer-tools/financial-well-being/',
         include('wellbeing.urls')
     ),
+    url(
+        r'^(?i)about-us/diversity-and-inclusion/',
+        include('diversity_inclusion.urls')
```
chosak: Could this [`include`](https://docs.djangoproject.com/en/1.11/ref/urls/#include) specify a namespace, e.g. `namespace='diversity_inclusion`? This allows you to keep the d-i URLs simple (like "form_submitted") while avoiding conflicts with other apps, letting you reference them with`diversity_inclusion:form_submitted`.

cfgov/cfgov/settings/base.py
```diff
@@ -94,6 +94,7 @@
     'treemodeladmin',
     'housing_counselor',
     'hmda',
+    'diversity_inclusion'
```
chosak: Don't forget to add this to the list of "first party" apps in [.isort.cfg](https://github.com/cfpb/cfgov-refresh/blob/69fd37433c06a0809158d71650125d72180b7333/.isort.cfg#L12). It's unfortunate that we have to maintain that list manually.

cfgov/diversity_inclusion/forms.py
```diff
@@ -0,0 +1,86 @@
+from __future__ import division
+
+from django import forms
+
+
+# Form input attributes for Capital Framework compatibility.
+# Technique copied from data_research/forms.py
```
chosak: Could this be centralized somewhere, to avoid duplication?

cfgov/diversity_inclusion/forms.py
```diff
@@ -12,75 +12,62 @@
...
-    'rows': None,
-    'cols': None,
-}
-textarea_attrs.update(text_input_attrs)
-
-
 class VoluntaryAssessmentForm(forms.Form):
     institution_name = forms.CharField(
-        max_length=250,
+        label='Institution name',
```
Scotchester: Setting this label explicitly isn't necessary, given the field name of `instutition_name`, but if you want to for clarity and consistency, that's cool.

cfgov/diversity_inclusion/jinja2/diversity_inclusion/voluntary-assessment-form.html
```diff
@@ -12,132 +12,217 @@
...
-    <div class="block">
-        <header class="m-slug-header">
-            <h2 class="a-heading">About this questionnaire</h2>
-        </header>
+{% block content_main -%}
+    <div class="block block__flush-top">
+        <h1>Voluntary diversity assessment registration form</h1>
+        <div class="lead-paragraph">
+            <p>
+                The Consumer Financial Protection Bureau’s (Bureau) D&I Assessment website allows financial institutions to submit D&I Assessments.
```
Scotchester: Diversity & Inclusion should be spelled out the first time here, right? Also, unless lawyers demand it, `(Bureau)` feels unnecessarily formal and of limited value in this setting.

cfgov/diversity_inclusion/jinja2/diversity_inclusion/voluntary-assessment-form.html
```diff
@@ -12,132 +12,217 @@
...
+                                {{ form.tax_id.label }}
+                            </label>
+                            {{ form.tax_id }}
+                        </div>
+                    </div>
+                </div>
+            </div>
+            <h2>Contact information</h2>
+            <p>
+                Please provide the name and contact information for the individual who will be your institution’s POC.
```
Scotchester: Maybe a trivial thing, but given the major context shift here since the `POC` abbreviation was introduced in the introductory text, I would spell it out here.

cfgov/diversity_inclusion/jinja2/diversity_inclusion/voluntary-assessment-form.html
```diff
@@ -12,132 +12,217 @@
...
+                            <small class="a-label_helper">(optional)</small>
+                        </label>
+                        {{ form.contact_phone_alt }}
+                    </div>
+                </div>
+            </div>
+            <div class="o-form_group">
+                <input type="hidden"
+                       name="csrfmiddlewaretoken"
+                       value="{{ csrf_token }}">
```
Scotchester: Trivial: These attributes are short enough to leave on line line, IMO.

cfgov/diversity_inclusion/jinja2/diversity_inclusion/voluntary-assessment-form.html
```diff
@@ -12,132 +12,217 @@
...
+                <div class="m-btn-group">
+                    <input class="a-btn"
+                       title="Submit your diversity assessment registration form"
+                       type="submit"
+                       value="Submit">
```
Scotchester: Trivial: Please align attribute names.

cfgov/diversity_inclusion/jinja2/diversity_inclusion/voluntary-assessment-form.html
```diff
@@ -12,132 +12,217 @@
...
+                    </div>
+                </div>
+            </div>
+            <div class="o-form_group">
+                <input type="hidden"
+                       name="csrfmiddlewaretoken"
+                       value="{{ csrf_token }}">
+                <div class="m-btn-group">
+                    <input class="a-btn"
+                       title="Submit your diversity assessment registration form"
```
Scotchester: I don't believe the `title` does anything in this context, but I could be wrong.

cfgov/diversity_inclusion/jinja2/diversity_inclusion/voluntary-assessment-form.html
niqjohnson: Fixed in 612d440

cfgov/diversity_inclusion/jinja2/diversity_inclusion/voluntary-assessment-form.html
niqjohnson: Fixed in 612d440

cfgov/diversity_inclusion/jinja2/diversity_inclusion/voluntary-assessment-form.html
niqjohnson: Fixed in 612d440

cfgov/diversity_inclusion/jinja2/diversity_inclusion/voluntary-assessment-form.html
niqjohnson: Hmmm, good question. I copied what was in https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/wellbeing/jinja2/wellbeing/home.html#L931, but I'm not really sure what `title` does here, either. I can remove it if it seems useless.

cfgov/diversity_inclusion/jinja2/diversity_inclusion/voluntary-assessment-form.html
niqjohnson: Fixed in 612d440

cfgov/diversity_inclusion/forms.py
niqjohnson: These label names might still be tweaked a little, so I'm going to leave the `label` for now for consistency and just in case it changes to something that wouldn't be great in code.

cfgov/diversity_inclusion/jinja2/diversity_inclusion/voluntary-assessment-form.html
Scotchester: Ah, I think I did that there so it would show up on hover if someone tried to click the button on that form before all required fields were completed. I think it's unnecessary in this case, but I will remove it when I make some edits to address Andy's feedback.

cfgov/diversity_inclusion/forms.py
Scotchester: We could define `text_input_attrs` anywhere and have it imported into these two forms, but what would be the appropriate place for that definition?

chosak: Maybe something like core/widgets.py?

cfgov/diversity_inclusion/jinja2/diversity_inclusion/voluntary-assessment-form.html
```diff
@@ -0,0 +1,189 @@
...
+                               for="{{ form.contact_phone_alt.id_for_label }}">
+                            {{ form.contact_phone_alt.label }}
+                            <small class="a-label_helper">(optional)</small>
+                        </label>
+                        {{ form.contact_phone_alt }}
+                    </div>
+                </div>
+            </div>
+            <div class="o-form_group">
+                <input type="hidden" name="csrfmiddlewaretoken" value="{{ csrf_token }}">
```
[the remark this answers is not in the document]
Scotchester: Hidden input; does not need label.

cfgov/diversity_inclusion/views.py
Scotchester: Thanks for the suggestion! Changed in https://github.com/cfpb/cfgov-refresh/pull/5265/commits/a82f1df9e65bb0be103c0aeecfe68b70229b1de1.

cfgov/cfgov/urls.py
Scotchester: Done in https://github.com/cfpb/cfgov-refresh/pull/5265/commits/a82f1df9e65bb0be103c0aeecfe68b70229b1de1.

cfgov/cfgov/settings/base.py
Scotchester: Done in https://github.com/cfpb/cfgov-refresh/pull/5265/commits/a82f1df9e65bb0be103c0aeecfe68b70229b1de1.

Scotchester: Done in a82f1df.

cfgov/diversity_inclusion/jinja2/diversity_inclusion/voluntary-assessment-form.html
```diff
@@ -0,0 +1,190 @@
...
+                <div class="m-btn-group">
+                    <input class="a-btn"
+                           type="submit"
+                           value="Submit">
```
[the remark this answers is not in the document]
Scotchester: Submit button does not need a label.

cfgov/diversity_inclusion/jinja2/diversity_inclusion/voluntary-assessment-form.html
Scotchester: Done in https://github.com/cfpb/cfgov-refresh/pull/5265/commits/a5984a4de355a7039e340103d72a9e04bcb19935.

Scotchester:
This PR is ready for final review!

cfgov/diversity_inclusion/jinja2/diversity_inclusion/privacy.html
```diff
@@ -0,0 +1,49 @@
...
+    {{ super() }}
+    <meta name="robots" content="noindex">
+{% endblock %}
+
+{% block title -%}
+    Privacy Act Statement
+{%- endblock %}
+
+{% block desc -%}
+    TODO: Come back to this
```
schbetsy: Should these TODOs in the templates still be here?

Scotchester: Good catch! Updated in a2b6197.

schbetsy:
I was able to submit the form with invalid data by entering "b@c" in the email field. This was confusing because it didn't show any validation messages to indicate that it failed or why.

Scotchester:
> I was able to submit the form with invalid data by entering "b@c" in the email field. This was confusing because it didn't show any validation messages to indicate that it failed or why.

I came across this in my testing, too. I guess this is the expected behavior for Django's built-in `EmailField` validation, so I'm not sure we should try to do anything about it?

If someone wanted to enter a fake email address, anyway, it'd be trivial to come up with one that passed a stricter validation.

schbetsy:
The problem would be solved if we either (a) matched the front-end regex validation to the back-end validation (or are they both built-in? If so, I wonder why they're different from each other), or (b) allowed there to be a validation error message to let the user know what failed.

Scotchester:
Sorry, @schbetsy, I misinterpreted your first comment. I now see that we were missing output of Django validation errors! I have added a commit that displays those.

I don't know that it's worth worrying about matching the Django validation to the browser's built-in validation, though. I suspect that the built-in browser validation is being true to the RFC-whatever spec of an email address by allowing email addresses of `user@gtld` (like, `postmaster@com`, which the operator of the .com TLD could theoretically allow), though in reality, that is not helpful at all, because no one should do that. So I wouldn't want to make Django's email validation less strict to abide by that, but we can't make the browser's validation more strict. We could change the `input` from `type="email"` to `type="text"` and validate it on the front end with JavaScript, but that seems like an unnecessary amount of work for an edge case. Would you agree?

schbetsy:
I realized after I commented that showing form validation errors is definitely the right way to go.  The front-end validator only works for people with JavaScript enabled, so we definitely shouldn't rely on it as the main validation solution.

LGTM!

PR #5291, opened 2019-10-11
PFC model update

higs4281:
This PR sets the stage for our college-cost decoder, an open version of
the dynamic disclosures tool built to help students at for-profit colleges.

Its intent is to establish our model changes in the cfgov database so that we can
iterate on importing and using new data points and implementing a new tool design.

### Included
- Three new page models to accommodate the open tool, a guide-quiz format and static legacy pages.
- New fields on the `school` model to hold data we will collect from the College Scorecard api.
- Blocks for the quizzes.
- Initial templates and front-end resources for the new pages.

### Not included
- Code to harvest the new data from Scorecard
- Final versions of templates and front-end code

### Testing
- Import the branch and run the new migration against a recent prod db snapshot:
```bash
./cfgov/manage.py migrate
```
- Add a child page to the PFC placeholder (http://localhost:8000/admin/pages/12466/) and choose **Student loan quiz page** 
- In the admin, note the Guided Quiz block available on the page, and try adding one. You can see a populated version of the guided quiz internally at [DEV5]/paying-for-college/choose-a-student-loan/
- You can also try adding a College costs page or a Repaying student debt page. They should render, but they are not fully functional.


cfgov/paying_for_college/jinja2/paying-for-college/quiz-answers.html
```diff
@@ -0,0 +1,44 @@
...
+   answer.answer_response:      Response explanation (right or wrong)
+
+   ========================================================================== #}
+
+
+<div class="quiz__answers m-form-field m-form-field__radio">
+  {% for answer in value.answers %}
+  <div class="quiz__answer">
+    <div class="quiz__answer-choice">
+      <input type="radio" value="some value" id="{{ value.situation_id }}" class="a-radio" name="test">
```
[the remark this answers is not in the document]
mistergone: This is a placeholder. Labels are not present because of an issue with element IDs.

cfgov/paying_for_college/jinja2/paying-for-college/college-costs.html
```diff
@@ -0,0 +1,231 @@
...
+                    </div>
+                </div>
+                <div class="FINANCIAL-ITEM">
+                    <div class="form-group_item">
+                        <label class="form-label" for="costs__roomboard">
+                            Room and board
+                        </label>
+                    </div>
+                    <div class="aid-form_input-wrapper">
+                        <input type="text" id="costs__room-board" name="costs__roomboard" data-financial="roomBoard" autocorrect="off" value="0">
```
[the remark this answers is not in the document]
mistergone: This is placeholder code. Labels are not present because of an issue with element IDs.

cfgov/paying_for_college/models/blocks.py
```diff
@@ -0,0 +1,57 @@
...
+class GuidedQuiz(blocks.StructBlock):
+    situation = blocks.RichTextBlock(
+        features=[
+            'ol', 'ul', 'bold', 'italic', 'link', 'image', 'document-link',
+        ],
+        blank=True,
+        required=False,
+    )
+    questions = blocks.ListBlock(
+        blocks.StructBlock([
```
Scotchester: Is there supposed to be a field to indicate what the correct answer for a question is?

cfgov/paying_for_college/models/blocks.py
```diff
@@ -0,0 +1,57 @@
+from wagtail.wagtailcore import blocks
```
chosak: I propose moving this file up a level, so it lives at cfgov/paying_for_college/blocks.py. I know we haven't followed this convention, but there's no reason that block types need to live in the models submodule. Moving it out would let you import the blocks (say, for testing) without importing all of Django's model code, which in turn loads settings, etc.

cfgov/paying_for_college/models/pages.py
```diff
@@ -1,4 +1,56 @@
...
+    edit_handler = TabbedInterface([
+        ObjectList(content_panels, heading='General Content'),
+        ObjectList(CFGOVPage.sidefoot_panels, heading='Sidebar'),
+        ObjectList(CFGOVPage.settings_panels, heading='Configuration'),
+    ])
+    objects = CFGOVPageManager()
+
+    def get_template(self, request):
+        for block in self.content:
+            block.value['situation_id'] = block.id
```
chosak: What is this logic doing? I believe that StreamField blocks should already have a unique ID. Could you use [this trick](https://github.com/wagtail/wagtail/pull/3593#issuecomment-315027698) with `{% include_block %}` to leverage it instead of defining a new variable?

cfgov/paying_for_college/models/pages.py
```diff
@@ -1,4 +1,56 @@
...
+        return 'paying-for-college/{}.html'.format(
+            self.slug)
```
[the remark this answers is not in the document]
chosak: +1 for having a design discussion about this, if you like @higs4281.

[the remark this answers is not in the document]
higs4281: I would love to discuss. We were trying to avoid type proliferation, but I can see the downsides.

cfgov/paying_for_college/models/blocks.py
higs4281: The "quiz" is less a test or competition than it is an attempt to educate in an engaging way, so we don't have a right or wrong flag, and we're not keeping score. 
The "reveal" of right or wrong would come in the explanations for each question.

cfgov/paying_for_college/models/blocks.py
higs4281: √

cfgov/paying_for_college/models/pages.py
higs4281: That's a great suggestion @chosak -- I'll give that a try.

cfgov/paying_for_college/models/blocks.py
higs4281: The quiz block might be more likely to be reused if it had more gamelike elements, but it's not the intent here, and the YES project was looking at this same strategy and might pick it up.

cfgov/paying_for_college/models/blocks.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class QuizAnswers(blocks.StructBlock):
+    answers = blocks.ListBlock(
+        blocks.StructBlock([
+            ('choice_letter', blocks.CharBlock(
+                max_length=20, required=False)),
+            ('answer_choice', blocks.CharBlock(max_length=500)),
+            ('answer_response', blocks.RichTextBlock(
```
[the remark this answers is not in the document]
higs4281: Will do.

higs4281:
> * Consider putting `quiz-answers.html` and `quiz.html` into a `blocks/` subfolder.

√

cfgov/paying_for_college/models/blocks.py
```diff
@@ -0,0 +1,57 @@
+from wagtail.wagtailcore import blocks
+
+
+class QuizAnswers(blocks.StructBlock):
+    answers = blocks.ListBlock(
+        blocks.StructBlock([
```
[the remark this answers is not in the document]
higs4281: I like this idea, @Scotchester , but I have had no luck basing QuizAnswers directly on ListBlock, which requires a child_block argument to base all the elements on. I can't find examples of anyone extending ListBlock in this way.

I hid the stuttered "Answers" label in the admin, but it's a less than ideal solution, because an annoying bare colon still displays. Wagtail's [docs](https://docs.wagtail.io/en/v1.13.4/topics/streamfield.html#basic-block-types) say ListBlock labels default to an empty string, but that doesn't appear to be true.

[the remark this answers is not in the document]
chosak: Would this work?

```py
class QuizAnswer(blocks.StructBlock):
    ...

class QuizQuestion(blocks.StructBlock):
    ...
    answers = blocks.ListBlock(
        QuizAnswer(),
        template='paying-for-college/blocks/quiz-answers.html'
    )

class GuidedQuiz(blocks.StructBlock):
   ...
   questions =  blocks.ListBlock(QuizQuestion())
```
[Here's](https://gist.github.com/chosak/1b20fc3a63316d1854433248ad1157b7) the change as a diff. It looks like this in the admin:

image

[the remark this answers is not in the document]
higs4281: @chosak Thank you, sir. And why didn't I think of that?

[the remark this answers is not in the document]
higs4281: @chosak Thanks again for lighting the way.

In adjusting the models I learned that it does no good to assign templates to [children of StructBlock and ListBlock](https://docs.wagtail.io/en/v1.13.4/topics/streamfield.html#boundblocks-and-values) -- the templates are ignored. This cleared up much confusion about how templates were working for these models.

cfgov/paying_for_college/models/pages.py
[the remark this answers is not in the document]
higs4281: I created three page types with explicit template assignments to handle our page needs.

chosak:
@higs4281 would you mind rebasing this against master? I see a lot of unrelated commits and changed files that make it hard to re-review.

cfgov/paying_for_college/models/pages.py
higs4281: Given the template behavior of ListBlock children, I'm not sure where we'll need to bring in the block ID, or whether we'll need it at all in the end. 

I would like to leave this variable in place for this PR, so that it's available as the front-end adjusts to the new model structures. My goal will be to ditch it or try to use the {% include_block %} method.

higs4281:
@chosak rebased.

higs4281:
On hold while I clean up rebase issues.

cfgov/paying_for_college/jinja2/paying-for-college/blocks/quiz.html
```diff
@@ -0,0 +1,66 @@
...
+      {% include_block question_block.subtitle %}
+    </h2>
+    <div class="quiz__question-content">
+      <div class="rich-text">
+        {% include_block question_block.question | richtext %}
+      </div>
+      <div class="quiz__answers m-form-field m-form-field__radio">
+      {% for answer_block in question_block.answers %}
+        <div class="quiz__answer-choice">
+          <input type="radio" value="some value" id="{{ answer_block.id }}" class="a-radio" name="test">
```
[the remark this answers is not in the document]
higs4281: This is proof-of-concept markup for the quiz structure. Real front-end code to follow.

higs4281:
Reviewers: This PR has more markup and front-end assets than originally intended, but my  absences and our timeline forced a unification.

higs4281:
@chosak @Scotchester I think all issues raised have been addressed.

cfgov/paying_for_college/jinja2/paying-for-college/choose-a-student-loan.html
```diff
@@ -0,0 +1,58 @@
...
+{% endblock %}
+
+{% block javascript scoped %}
+    {{ super() }}
+    <script async>
+      if ( document.body.parentElement.className.indexOf( 'no-js' ) === -1 ) {
+        !function(){
+          {# Include site-wide JavaScript. #}
+          var s = [
+            '{{ static("apps/paying-for-college/js/quiz.js") }}'
```
chosak: Why is this JavaScript file being included here in the template, instead of defined as part of the `page_js` [here](https://github.com/cfpb/cfgov-refresh/blob/c8841386ab6afb47b9f5f12e8db995576c780478/cfgov/paying_for_college/models/pages.py#L94), like secondary-navigation.js? A similar comment applies to the other templates which include JS like this. It would be better if we standardize on one method for including Page-specific JS files.

cfgov/paying_for_college/models/pages.py
```diff
@@ -1,4 +1,137 @@
...
+        ObjectList(content_panels, heading='General Content'),
+        ObjectList(CFGOVPage.sidefoot_panels, heading='Sidebar'),
+        ObjectList(CFGOVPage.settings_panels, heading='Configuration'),
+    ])
+    objects = CFGOVPageManager()
+
+    def get_template(self, request):
+        for block in self.content:
+            block.value['situation_id'] = block.id
+        return 'paying-for-college/choose-a-student-loan.html'.format(
```
chosak: Is this `.format()` needed any longer, now that this has been split into multiple page types?

cfgov/paying_for_college/models/pages.py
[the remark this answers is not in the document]
chosak: Thanks, @higs4281. This implementation duplicates a lot between the three page types: basically everything instead of the content StreamField blocks and the template. What would you think about introducing an [`abstract`](https://github.com/wagtail/wagtail/blob/e844200f27fe6a2089175b488a283bc27331c4c6/wagtail/contrib/routable_page/models.py#L138) parent page type for these, with three concrete models that inherit from it and only define the template code, and content field? This would reduce a lot of duplication. For the content field you could even do something like we have [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/models/browse_filterable_page.py#L19), for example:

```py
class PayingForCollegePage(CFGOVPage):
    header = ...

    ...

    class Meta:
        abstract = True

class PayingForCollegeContent(StreamBlock):
    full_width_text = organisms.FullWidthText()
    info_unit_group = organisms.InfoUnitGroup()
    expandable_group = organisms.ExpandableGroup()
    expandable = organisms.Expandable()
    well = organisms.Well()
    raw_html_block = blocks.RawHTMLBlock(label='Raw HTML block')

class RepayingStudentDebtPage(PayingForCollegePage):
    content = StreamField(PayingForCollegeContent, blank=True)
    template_name = 'paying-for-college/repaying-student-debt.html'

class CollegeCostsPage(PayingForCollegePage):
    content = StreamField(PayingForCollegeContent, blank=True)
    template_name = 'paying-for-college/college-costs.html'

class StudentLoanQuizContent(PayingForCollegeContent):
    guided_quiz = GuidedQuiz()

class StudentLoanQuizPage(PayingForCollegePage):
    content = StreamField(StudentLoanQuizContent, blank=True)
    
    def get_template(self):
        ...
```

Unfortunately Django model inheritance doesn't let you redefine a field on a parent model, the way vanilla Python class inheritance does; this prevents you from defining the base `content` on the parent and overriding as needed in the child.

higs4281:
@chosak All excellent suggestions. Working them up.

cfgov/paying_for_college/jinja2/paying-for-college/choose-a-student-loan.html
higs4281: Agreed.

cfgov/paying_for_college/models/pages.py
higs4281: Missed that. Thanks for the catch.

cfgov/paying_for_college/jinja2/paying-for-college/choose-a-student-loan.html
higs4281: @chosak @Scotchester 
Assets added to the `page_js` page value (which gets [rolled into](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/models/base.py#L315-L316) the page's `media` value) [must be available](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_layouts/base.html#L305-L309) at the `/static/js/routes/on-demand/` asset path. The on-demand namespace appears to be reserved for sitewide js modules, such as header.js, footer.js and exapandable.js, so it doesn't seem to be the right spot for page-specific js.

The template method we used to grab assets from the /apps/ namespace was copied from [other spots](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/regulations3k/jinja2/regulations3k/search-regulations.html#L216-L232) where app-specific js was needed. 

Is there a better or more standard way to deliver the app-namespaced js?

chosak:
@higs4281 thanks for the thorough explainer. Does component-level mean that the page_js appears on all pages on the site ("shared sitewide components"), or only pages _of the specific type_?

Also pinging @anselmbradford since he might be the most familiar with some of this code.

higs4281:
@chosak Component-level only targets one page type, but most "on-demand" choices are single-file js components that are used across the site. Would we want to start populating on-demand with   sets of js files that only apply to one page?

Scotchester:
@higs4281 

Yeah, this is a tricksy situation, for sure. That may be the reason why it was done the way it was on [the regs3k example you cited](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/regulations3k/jinja2/regulations3k/search-regulations.html#L216-L232). And maybe that's the best way to go for now, until the full path/slug for these pages settles down?

anselmbradford:
@higs4281 We have a fourth type, which is app assets, such as https://github.com/cfpb/cfgov-refresh/tree/master/cfgov/unprocessed/apps/owning-a-home. This allows an app to set it's own dependencies for a particular part of the site, which are independent from the wider site dependencies. These type of assets are used in templates that extend the base templates and are only used for a specific section, but aren't tied to a URL, like https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/owning-a-home/explore-rates/index.html#L22-L36. I think this is how the regs3k example Scott linked is set up too.

higs4281:
Thanks, @anselmbradford 

@chosak The "fourth type" @anselmbradford referred to is the method we're using.

chosak:
@higs4281 awesome. It'd be nice to remove some of the duplication in that code in the templates, but that's not a job for this already aged PR. Well done.

PR #5303, opened 2019-10-18
Allow any InfoUnitGroup format on LearnPages

chosak:
Currently LearnPages can only be configured with InfoUnitGroups in the 25/75 format. This is enforced via a custom block named InfoUnitGroup2575Only. We want to loosen these restrictions to allow for other configurations (50/50 and 33/33/33).

This change both modifies the LearnPage to add a new InfoUnitGroup option but also includes a data migration that should move any existing InfoUnitGroup2575Onlys to regular InfoUnitGroup.

I'm purposefully not removing InfoUnitGroup2575Only yet to be more conservative and roll out the change in two steps.

## Testing

The best way to test this is to create or identify a page that has an IUG2575Only, run the migration, and see that it's now a plain ol' IUG.

## Screenshots

I've modified how IUG2575Only appears in the Wagtail admin to discourage anyone from adding one once this change goes through:

image

## Checklist


cfgov/v1/atomic_elements/organisms.py
```diff
@@ -165,7 +165,7 @@ class InfoUnitGroup2575Only(InfoUnitGroup):
     )
 
     class Meta:
-        label = 'Info unit group'
+        label = 'IUG 25/75 (deprecated)'
```
anselmbradford: Can we add a `TODO` comment here that describes what state we have to reach to be able to remove this bit?

cfgov/v1/models/learn_page.py
```diff
@@ -104,6 +104,8 @@ def meta_image(self):
 class LearnPage(AbstractFilterPage):
     content = StreamField([
         ('full_width_text', organisms.FullWidthText()),
+        ('info_unit_group', organisms.InfoUnitGroup()),
+        # TODO: InfoUnitGroup2575Only is deprecated and will be removed.
```
anselmbradford: Oh maybe ignore my earlier comment, but could we elaborate here on what will need to happen to remove this? Unless it's coming in like the very next PR.

chosak: Yes, my intention is to open a subsequent PR that essentially just removes the definition of and any code references to InfoUnitGroup2575Only. That will generate another schema migration for LearnPage but won't require a data migration since we're taking care of that in this RP.

PR #5307, opened 2019-10-23
YES tool — Adds sidebar content to YES tool

el-mapache:
So the site maintains a consistent visual appearance, and the user can easily view other related resources, adds the standard sidebar on the YES transportation tool page.

## Additions

- Adds sidebar links and styles

## Screenshots

Screen Shot 2019-10-22 at 5 59 50 PM


el-mapache:
@anselmbradford The difference is that our menu is hardcoded, where as the other one you mention is, I think, handled via wagtail. So that will need to get updated separately by a content person.

PR #5311, opened 2019-10-24
YES tool — Reorganize to-do list structure in review section

el-mapache:
The information layout was hard to parse in the previous designs of the review section, so they have been altered to be a bit clearer.

## Changes

- Todo lists in review section now appear below the option details
- Alerts now appear above option details, and mirror the alerts in the form
- Alert styles normalized

## Testing

1. Specs pass
2. Todo lists in the review section should each have a default item


cfgov/unprocessed/apps/youth-employment-success/css/main.less
```diff
@@ -218,6 +218,11 @@ input:-moz-ui-invalid {
   display: none;
 }
 
+.highlight {
+  background-color: #f9f9f9;
+  padding: .75em;
```
anselmbradford: Perhaps make this a calculated value so the intended pixel size is apparent:

```
unit( 12px / @base-font-size-px, em );
```

PR #5322, opened 2019-10-30
Remove CodeCov integration

Scotchester:
CodeCov does not work reliably in this repo.


Scotchester:
Aside from Coveralls, [CodeClimate](https://codeclimate.com/quality/) at least has the appearance of being a decent option,

Scotchester:
I do understand @willbarton's POV. If CodeCov's not there annoying us, do we ever get back around to adding something else back in?

Let's put this on hold until the transition to GitHub Actions is complete. I would prefer not to implement running a different coverage tool in Travis for just another two weeks.

willbarton:
We don't current require any code coverage for PRs to be merged, so broken coverage reporting isn't a blocker to PRs. I do feel like outage at broken code coverage reporting in some PRs makes me check what the code coverage is. So, I'd like us to have an alternative coverage reporting solution in place before we remove something and forget about it.

chosak:
CodeCov seems to have started to post comments to PR, for example [here](https://github.com/cfpb/cfgov-refresh/pull/5492#issuecomment-582592815). This makes it harder to ignore, and I still can't interpret these reports to do anything useful with them. Now that this is starting to generate GitHub notifications, can we revisit the idea of removing this, or can it be configured to stop commenting?

willbarton:
I am in favor. I just feel very strongly that we should have an alternative.

willbarton:
It looks like we might be able to disable the Codecov comments per #5498.

chosak:
> It looks like we might be able to disable the Codecov comments per #5498.

Unfortunately this doesn't seem to have worked. @contolini's PR #5499 was opened after that change was made (see on [his branch](https://github.com/contolini/cfgov-refresh/blob/887fb3d914cf0a1daee5f008817acede31e7610c/codecov.yml#L6)), but codecov still [posted a comment](https://github.com/cfpb/cfgov-refresh/pull/5499#issuecomment-584284883).

willbarton:
@chosak I don't think that's quite right. @contolini's first commit was 5d685c63ebb38b04f65033bfb4a3f704e66d9676 , and `codecov.yml` at that commit was: https://github.com/cfpb/cfgov-refresh/blob/5d685c63ebb38b04f65033bfb4a3f704e66d9676/codecov.yml

chosak:
@willbarton I'm happy to be corrected! And it looks like @wpears's subsequent PR #5500 didn't get a comment, so 🎉

willbarton:
LET'S REMOVE CODECOV!

PR #5323, opened 2019-10-30
Paying for college: Fix the https URL of S3 resources

higs4281:
In September we changed some http URLs used in code to https.
Most of the changes were good as-is, but one related to an S3 resource made
a Jenkins job fail, because the S3 URL was not designed to be used with https.

In this case, the old URL was:
`http://files.consumerfinance.gov.s3.amazonaws.com/pb/paying_for_college/csv/validated_program_data/[SLUG]`

Using that URL with https raises a certificate error when a Jenkins job requests it.

The https form that we can use (as pointed out by @rosskarchner) is:
`https://s3.amazonaws.com/files.consumerfinance.gov/pb/paying_for_college/csv/validated_program_data/[SLUG]`

## Testing
To load an S3 resource locally and confirm that the S3 file is fetched using https without error, run
```bash
./cfgov/manage.py load_programs --s3 true CFPB_EDMC_20191017.csv
```

You should get an echo saying:
```bash
0 programs created. 2 programs updated.
```

### DIFF
--- a/cfgov/paying_for_college/disclosures/scripts/load_programs.py
+++ b/cfgov/paying_for_college/disclosures/scripts/load_programs.py
@@ -217,7 +217,7 @@ def load(source, s3=False):
     updated_programs = 0
     FAILED = []  # failed messages
     if s3:
-        s3_url = ('https://s3.amazonaws.com/files.consumerfinance.gov'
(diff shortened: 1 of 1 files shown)


higs4281:
> Can't we also just do ...

Yup, that works, and is simpler.

PR #5324, opened 2019-10-30
Only deliver video JS to event pages when needed

Scotchester:
Event pages are currently loading `video-player.js` regardless of whether or not they need it, resulting in a console error if there is not actually a video module on the page.

This PR makes the loading of that JS conditioned on whether or not the page is in a state which would have a video module.

Fixes GHE/CFGOV/platform#3569

## Additions

- Adds `event_state` property to `EventPage` that makes it easy to check what the current state of an event is (past, present, or future) at the current time. This is also added as a context variable to the `event.html` template.
- New testing requirement of freezegun for unit testing things that call `datetime.now()` with a fixed date.

## Changes

- Modifies `EventPage`'s `page_js` property to check the current state of the event page and only add `video-player.js` if the page is in a state that would expect to have a video player module on it.
- Makes the `start_dt` field on `EventPage` a required field. This greatly simplifies some of the logic that was previously needed to cover the scenario in which it wasn't populated. In practice, and event is never created without knowing what the expected date is.
- The `_convert_date` private method from the `v1.jinja2tags.datetimes` module was moved into a public util module so that it could be accessed both by the Jinja tag and by the EventPage prop, as well as any future use.

## Removals

- Removes page title tests that seemed to provide no value, simply testing standard Wagtail behavior.

## Testing

1. Reproduce the problem on `master` with a fresh database dump by loading up http://localhost:8000/about-us/events/cfpb-symposium-section-1071-dodd-frank-act/ and looking in the console
1. Pull branch
1. Reload the page and see that the console error disappears
1. Fiddle with start and end dates and video fields (Livestream URL and YouTube URL) to check all states of an event and confirm that they load videos when expected

## Notes

- ~~I took the lazy route of importing the `when()` method from the `v1.jinja2tags.datetimes` module. If you would strongly prefer refactoring so that that method lives in a different module that is imported both as a Jinja2 tag and in the `EventPage` `page_js` property definition, please say so.~~
  - Actually, I think the better thing to do would be to determine `event_state` (which is what `when()` returns) in a class property, then add that to the context for easy reference in a template, eliminating the need for `when()` at all. _[Yep, did this.]_

## Checklist


cfgov/v1/models/learn_page.py
```diff
@@ -373,9 +375,35 @@ class EventPage(AbstractFilterPage):
...
+
+        if self.live_stream_date:
+            start = convert_date(
+                self.live_stream_date,
+                'America/New_York'
+            )
+        else:
+            start = convert_date(self.start_dt, 'America/New_York')
+
+        if datetime.now(timezone('America/New_York')) > start:
```
schbetsy: So if an event doesn't have an end date, it will stay in "present" state forever, right? Do you foresee that ever being an issue? How did we previously deal with events that had no end time?

PR #5328, opened 2019-11-04
YES Tool — Conditionally hide the days per week copy

el-mapache:
So the user doesn't get confused by situational text, remove the `Based on the {n} days a week you will make this trip` text when the user indicates that their average cost of getting to work is calculated on a monthly basis.

## Changes

- Hide this text when the user indicates their average cost to get to work is
a monthly cost

## Testing

1. Specs should pass
2. In a route option form, select the `monthly` radio button choice in the `average cost` fieldset

### DIFF
--- a/cfgov/jinja2/v1/youth_employment_success/line-item.html
+++ b/cfgov/jinja2/v1/youth_employment_success/line-item.html
@@ -2,7 +2,9 @@
 <div class="content-l m-yes-line-item u-clearfix">
   <p class="content-l_col content-l_col-2-3 line-item line-item__content-l h4">
     {{ values.label | safe }}
-    {{ values.helperText | safe if values.helperText else '' }}
+    <span class="line-item__content-l">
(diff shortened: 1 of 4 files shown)


cfgov/jinja2/v1/youth_employment_success/line-item.html
```diff
@@ -2,7 +2,9 @@
 <div class="content-l m-yes-line-item u-clearfix">
   <p class="content-l_col content-l_col-2-3 line-item line-item__content-l h4">
     {{ values.label | safe }}
-    {{ values.helperText | safe if values.helperText else '' }}
+    <span class="line-item__content-l helper">
```
anselmbradford: `helper` should probably be something less generic, or at least `u-helper`.

el-mapache: I actually don't need this class, removing! 😬

PR #5332, opened 2019-11-05
Remove hmda `decimal-align-right` class

anselmbradford:
Fixes https://[ghe]/CFGOV/platform/issues/3595
Not worth having a dedicated app asset directory for this one subtle CSS class.

## Removals

- Remove hmda `decimal-align-right` class and associated app asset files.

## Testing

1. Visit https://www.consumerfinance.gov/data-research/hmda/historic-data/ and compare the results table to http://localhost:8000/data-research/hmda/historic-data/ from this branch.

## Screenshots

First line in both images is with this class removed; second line is with this class:
Screen Shot 2019-11-05 at 10 36 50 AM
Screen Shot 2019-11-05 at 10 37 02 AM


anselmbradford:
Maybe the ability to right-align a column is something we need in `o-table`.

Scotchester:
> Maybe the ability to right-align a column is something we need in `o-table`.

[We have it!](https://cfpb.github.io/capital-framework/components/cf-tables/#right-aligned-cells)

The issue here is that the columns are too wide for it to look normal in a full-width table with only 3 columns that have small amounts of content in them.

Scotchester:
Seeing as the only table that uses this has no problematic 8-digit numbers, and never will (if I understand it correctly), perhaps we can proceed with merging this PR, @contolini? With the changes in this PR, the commas on https://www.consumerfinance.gov/data-research/hmda/historic-data/ are all still aligned enough to the naked eye.

contolini (approved):
There are [scenarios](https://www.consumerfinance.gov/data-research/hmda/historic-data/?geo=al&records=originated-records&field_descriptions=labels) where the alignment is more visible but it doesn't look bad, just different, so I'm cool with this change.

PR #5338, opened 2019-11-07
Move psycopg2 into libraries.txt

willbarton:
Because we now have a model using [a PostgreSQL-specific field](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/paying_for_college/models/disclosures.py#L11), we need to ensure we're always installing psycopg2.

This is especially important for a project like Teacher's Digital Platform, that relies on installing cfgov-refesh and [uses `libraries.txt`](https://github.com/cfpb/teachers-digital-platform/blob/master/install_cfgov_refresh.sh#L9-L12) for purpose of dependencies. ([Here is a failure](https://travis-ci.org/cfpb/teachers-digital-platform/jobs/608305892#L1063) illustrating the current problem.)

Even though this is an immediate problem for TDP, I think conceptually the fix that seems most correct is to move the requirement here, in cfgov-refresh. The alternative would be to have TDP also install `postgres.txt` and leave cfgov-refresh alone.

## Checklist


chosak:
There are [several other references to the postgres.txt file](https://github.com/cfpb/cfgov-refresh/search?q=postgres.txt&unscoped_q=postgres.txt) that will need to be removed as well, for example in deployment.txt or the Snyk configuration.

Also, does use of this field mean that it's not possible to use SQLite anymore? More specifically, that it's not possible to `manage.py migrate` a full cf.gov database against SQLite? If so, [this part of the documentation](https://github.com/cfpb/cfgov-refresh/blob/8b1ddc85c507687f0deb7563cd00e9d094ac0ee1/docs/installation.md#install-postgres) should be changed to remove references to it. There are [some other references to SQLite](https://github.com/cfpb/cfgov-refresh/search?q=sqlite&unscoped_q=sqlite) as well that would also need to be removed. More generally, though, if a cf.gov database can no longer be initialized with SQLite, that might cause problems with some Jenkins jobs that try to do cf.gov things locally; one that comes to mind is the ratechecker validation job, see scripts/ratechecker_load.py#L83 in the internal cf.gov Jenkins job repo.

willbarton:
I will note, SQLite still works for running the tests (I haven't looked into how that `JSONField` might be tested, but the tests pass), I just thing it needs to be in place.

We might want to look more into that though — what the effect of using Django's PostgreSQL-specific fields have on our use of SQLite in tests, Jenkins, and various other little places might be and what we need to change about that (I **do not** think we should avoid using the PostgreSQL fields).

willbarton:
Note to self: modify the Snyk config first next time.

higs4281:
Late to the show, but tox tests run without issue for the person who introduced a postgres-only field!

PR #5340, opened 2019-11-07
YES tool — Updates validation, fixes reducer splice logic

el-mapache:
This PR fixes the following issues around which alert notifications should be displayed:

* When `averageCost` is entered but `per month` or `per day` is not, the yellow incomplete alert
should be displayed.

* When `averageCost` is entered and `per day` is selected, but `daysPerWeek` is empty, the yellow incomplete alert should be displayed.

* When a to-do list item is selected, and all other data is filled in, and the user moves to a different transportation type, the green success alert `looks good, but with todos` should show.

## Additions

- Adds validations for all transportation types other than `drive`, to capture the behavior of the `averageCost` and associated fields

- Adds `isEmpty` method to determine if a value is empty, rather than relying on a straight boolean check (`!value`)

## Changes

* Updates `updateActionPlanItems`. Performing a `splice` of a non-existent index is the same as -1,1, meaning the first element of an array will always be removed. Now we check that the item to be removed from the `actionPlanItems` array is present.

## Testing

1. Specs, and the notes above.

### DIFF
--- a/cfgov/unprocessed/apps/youth-employment-success/js/validators/route-option.js
+++ b/cfgov/unprocessed/apps/youth-employment-success/js/validators/route-option.js
@@ -36,6 +36,34 @@ function isFieldInActionPlan( fieldName, routeTodoList ) {
 }
 
 /**
+ * Determine is a value can be considered empty. It is empty if the value is:
+ *
+ * null
+ * undefined
+ * an empty string
+ * an empty array
+ * an empty object
+ *
+ * @param {*} value The value to be checked
+ * @returns {Boolean} If this value is empty or not
+ */
+function isEmpty( value ) {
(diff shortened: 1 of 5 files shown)


cfgov/unprocessed/apps/youth-employment-success/js/validators/route-option.js
```diff
@@ -35,6 +35,42 @@ function isFieldInActionPlan( fieldName, routeTodoList ) {
   return false;
 }
 
+/**
+ * Determine is a value can be considered empty. It is empty if the value is:
+ *
+ * null
+ * undefined
+ * an empty string
+ * and empty array
```
anselmbradford: `and` --> `an` ?

PR #5342, opened 2019-11-08
Fix the ability to run tox tests in Docker

willbarton:
Setting `PIP_NO_CACHE_DIR` to `true` breaks our use of venv-update in tox. This means it's not possible to successfully run the Python unit tests in the Docker container.

Since our primary use of Docker at the moment is for local development, so it seems prudent to preserve the ability to do all aspects of development, including run the test. The cost of longer build time and some warnings seems worth that to me.

We should revisit this when/if we deploy Docker containers for deployment.

## Testing

1. `docker-compose build python2 python3`
2. `docker-compose up`

Then, in another terminal:

3. `docker-compose exec python3 bash`
4. `tox`

Prior to this change you would see the following error when tox tries to create the virtual envs:

```
  File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/virtualenv_support/pip-9.0.1-py2.py3-none-any.whl/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/virtualenv_support/pip-9.0.1-py2.py3-none-any.whl/pip/commands/install.py", line 272, in run
    with self._build_session(options) as session:
  File "/opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/virtualenv_support/pip-9.0.1-py2.py3-none-any.whl/pip/basecommand.py", line 69, in _build_session
    if options.cache_dir else None
  File "/src/cfgov-refresh/.tox/lint-py36/lib64/python3.6/posixpath.py", line 78, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not int
```

After this change, `tox` should run successfully.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,10 +32,6 @@ RUN yum -y install \
     source /etc/profile && \
     pip install --no-cache-dir --upgrade pip setuptools
 
-# Disables pip cache. Reduces build time, and suppresses warnings when run as non-root.
-# NOTE: MUST be after pip upgrade. Build fails otherwise due to bug in old pip.
-ENV PIP_NO_CACHE_DIR true
-
 # Install python requirements
 COPY requirements requirements
 


Scotchester:
@willbarton Did you see this error on the Python 3 tests?

```
ModuleNotFoundError: No module named 'pip._internal.req.constructors'
```

willbarton:
@Scotchester this seems related to venv-update and the tox updates I tried to make here. I've backed those out.

Scotchester:
tox runs for me now, but only the lint-py27 environment runs correclty. I get some errors trying to install deps in the other environments: `error: command 'gcc' failed with exit status 1`

PR #5344, opened 2019-11-08
Update enforcement actions

wpears:
@cfarm 
Addresses JIRA 3716

## Additions

- Adds `update_enforcement_actions.py` which loops through all `DocumentDetailPage` objects, grabbing just the enforcement action pages. It then grabs the sidefoot stream data, removes the inline blob categories, and sets the new "Show Categories" option.
- Adds a block to to the RelatedMetadata molecule to allow editors to trigger whether they want to display page categories (added by default by this script).
 - Updates the related-metadata template to use the page categories when this is selected
 - Adds the new enforcement action categories that will be assigned to pages in a separate script after this migration is applied

## Testing

1. Pull branch
2. `docker-compose exec python2 /bin/bash`
3. `./cfgov/manage.py migrate`
4. `./cfgov/manage.py runscript update_enforcement_categories`
5. Check the sidebars of various enforcement actions to make sure that the category is updated and displayed
6. Check in the wagtail admin to note the new "show categories" block and that the page category is fed from the configuration pane.


cfarm:
instructions for step 2, "Run update-enforcement-blobs.py script locally"

1. `docker-compose exec python2 /bin/bash`
1. `./cfgov/manage.py runscript update_enforcement_blobs`

note filename in PR description is wrong, should be underscores not hyphens

cfarm:
THIS PR APPROVED BY CAT FARMAN 12-17-2019 STAMPED SUBMITTED APPROVED SIGNED .

PR #5347, opened 2019-11-12
Paying for College: Fix default import to named import

anselmbradford:
[closest](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/js/modules/util/dom-traverse.js#L105) is one of several exports from the dom-traverse, module, so we need to specify it. Otherwise all methods are imported under the `closest` object.

Screen Shot 2019-11-12 at 12 28 27 PM

## Changes

- Change default import to named import.

## Testing

1. @higs4281 Not sure how this is used, but likely is not working correctly on master.


mistergone:
This is code that's not used on production and is still in early development. It got yanked into `master` because some of my early work on P4C made it into one of Bill's PRs.

You can close this PR without worrying about merging - it's not currently used and my error won't appear in future versions of this code. :)

anselmbradford:
@mistergone …or we can merge it, right? It shows up as a warning in webpack when we build the assets

mistergone:
> @mistergone …or we can merge it, right? It shows up as a warning in webpack when we build the assets

Oh, yeah, then go for it!

I've reviewed this PR and approve! 👍

PR #5350, opened 2019-11-13
YES tool — Reorg js files to be more logical + understandable

el-mapache:
This PR is part of the cleanup week effort to document our work and put a bow on everything! Here, I've reorganized the JS files to hopefully be more logical and clear. Also, I've fixed a bug where totals were not calculating when `0` was entered as an input to certain fields.

## Additions

- Fix bug where totals don't calculate when days per week is filled in

## Changes

- Reorganize all the views, and other JS files, to make the structure of the app more apparent


anselmbradford:
Looks like the test needs to be updated (run `gulp test:unit`).

Also, if it makes sense here, run `gulp lint:tests` and fix the error there, which looks like it would be updating [this](https://github.com/cfpb/cfgov-refresh/blob/master/test/unit_tests/apps/youth-employment-success/js/validators/route-option-spec.js#L19-L24) from
```
  it( 'validates undefined data', () => {
    expect( validate( {
      ...data,
      earned: undefined
    } ) ).toBeFalsy();
  } );
```

to

```
  it( 'validates undefined data', () => {
    let UNDEFINED;
    expect( validate( {
      ...data,
      earned: UNDEFINED
    } ) ).toBeFalsy();
  } );
```

el-mapache:
Whoops! Thanks, thought I ran the tests right before I pushed, but clearly I did not 😄

PR #5334, opened 2019-11-06
Add new ContactExpandableGroup block to LearnPages

chosak:
This change introduces a new StreamField block named ContactExpandableGroup, and adds it as an option on LearnPages. See GHE/el-camino#21 for background on motivations for this change; this will make it easier to support [the site Contact Us page](https://www.consumerfinance.gov/about-us/contact-us/). Ping @sonnakim.

This branch has been deployed to our internal DEV3 testing server, and can be viewed in action at DEV3/about-us/contact-us/.

## Additions

- New `ContactExpandableGroup` organism that renders similarly to regular `ExpandableGroup`s, except that the content gets auto-filled from `Contact` snippets instead of from manual user input.

## Changes

- A new option has been added to `ExpandableGroup`, which wraps the entire organism in a `<div>` with an `id` based on the slugified header. This allows for in-page linking to a specific `ExpandableGroup`. I've left this disabled by default. Question for @anselmbradford @Scotchester: any issues with adding the plain `<div>` to all EGs even if this is disabled?
- I've modified the `Contact` snippet so that it always sorts by `heading` by default, which seems reasonable and prevents us from needing to manually specify it each time.

## Testing

1. See new unit tests, run `tox`.
2. Try creating a LearnPage and adding a `ContactExpandableGroup` block.

## Screenshots

New "Contact expandable group" block available on LearnPages:

image

Wagtail editing interface:

image

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #5352, opened 2019-11-13
Add unique slugs to expandable group headings

chosak:
This change generalizes some behavior that was introduced in #5334, adding an 'id' attribute to the div wrapping ExpandableGroups, based on a slugified version of the EG's heading (if specified).

This makes it easier to generate in-page links to ExpandableGroups, and is now turned on by default for all existing EGs.

A new `slugify_unique` Jinja filter has been created to support this behavior.

To test, run a local server, create a page that supports EGs (like LearnPage) and add one or more. Inspect page source to see the generated 'id' attribute. Note that if you generate multiple EGs with the same heading, their 'id' attributes will be different.

Ping @contolini.

## Todos

One concern here: should we worry that a slugified EG heading might conflict with some _other_, non-EG element on the page with the same `id`?

## Checklist


Scotchester:
> One concern here: should we worry that a slugified EG heading might conflict with some _other_, non-EG element on the page with the same `id`?

It's certainly possible. The solution implemented here is clever, but the possibility of that sort of conflict is again why it would be beneficial to implement automatic heading IDs as middleware against the full page markup.

chosak:
> The solution implemented here is clever, but the possibility of that sort of conflict is again why it would be beneficial to implement automatic heading IDs as middleware against the full page markup.

Thanks @Scotchester; I agree that would be a better holistic solution. Hypothetically, how would you propose to implement that? Would elements render with some kind of class or other attribute indicating that they need an id? Or an `id="to-be-filled-in"` attribute that gets replaced?

Scotchester:
> Edit to add: maybe I'm overthinking this, and it should be as simple as applying an `id` attribute to all `<h1..6>` elements?

That's what I had in mind :) Pseudocode:

```python
h_tags = H_TAG_REGEX.findall(html)
h_ids - []

for h in h_tags:
    if it has an id attr:
        h_ids.append(h.id)
    else:
        slugified = h.text|slugify
        while slugified in h_ids:
            slugified = slugified + '-1'
        h.id = slugified
        h_ids.append(slugified)
```

PR #5353, opened 2019-11-13
webpack: Don't extract comments in build

anselmbradford:
Terser webpack plugin extracts license code comments in code into a `.LICENSE` bundle. We don't use this, so best to turn it off.

## Changes

- Turn off extracting comments in terser minification step.
- Add `minimize` flag to build config, per terser documentation examples.

## Testing

1. Run `gulp build` on `master` and see that there are `.LICENSE` files generated.
1. Pull this branch and run `gulp build` and see that there are no such files.

@contolini


contolini:
👍 🎉

PR #5357, opened 2019-11-19
PFC - adjust School model to serve cohort comparisons

higs4281:
Our new paying-for-college open tool envisions offering school comparisons
by cohort -- schools in the same state, in the same degree-granting class,
or of the same public/private/for-profit control type.

This requires some new model fields and methods to gather new data points
and calculate rankings, and adjustments to our `update_colleges` script.

~~The code [tests] for processing cohort comparisons will follow shortly.
I wanted some product-owner review of my percentile ranking, for which I'd prefer to avoid requiring the heavy `scipy` package.~~
cohort comparison code is tested and reviewed by product owner.

## Testing
- Load a recent db dump and run `./cfgov/manage.py migrate`

#### update_via_api
You can run the `update_via_api` job against a single school to see the new fields and update script in action, if you have our ED API key in your env. Ping me if you'd like to add that.  

```bash
./cfgov/manage.py update_via_api --school_id 155317
```
And then see the results via the API at:
<http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/api/school/155317/>

You *could* run the whole update by omitting the `--school_id` arg. This refreshes every school in our database. But the routine is paced to avoid throttling by College Scorecard, so it can take more than an hour, depending on connection speeds. This job is generally only run once a year, in the Fall.

#### process_cohorts

This job is intended as an annual second step to the long `update_via_api` job. It pre-calculates a matrix of cohort rankings for every school so that they don't have to be calculated on the fly. The whole job takes 6 minutes or so, but you can run it against one school as shown below. The example is for Texas Tech.

Run the job: 
```bash
./cfgov/manage.py shell
from paying_for_college.disclosures.scripts.process_cohorts import run
run(single_school=229115)
```

See the results:
<http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/api/school/229115/>

Since this script resides in an interior app directory, the `runscript` utility can't run it. 
We'll add a management command to run it in a later PR.

## Note
A few new API endpoints are duplicates, in order to be backward-compatible for the sake of the for-profit version of the tool, which is still in use.


cfgov/paying_for_college/disclosures/scripts/process_cohorts.py
```diff
@@ -0,0 +1,109 @@
+import datetime
+import sys
+
+import localflavor
+import numpy as np
```
chosak: I'd like to do a more thorough review of this PR, but one quick thing that jumps out: Numpy is a pretty heavyweight package. Is there any lighter method of doing this processing that doesn't require it?

higs4281: I mistakenly thought we had numpy in our requirements. Now that you've set me straight, I'll pursue a lighter option. I'm only using it for one [percentile ranking calculation](https://github.com/cfpb/cfgov-refresh/blob/e38b9867db577ce7575e7e157e18a7aa68a20ab4/cfgov/paying_for_college/disclosures/scripts/process_cohorts.py#L54-L58).

higs4281: numpy requirement removed. I tested the replacement method against 400 random schools, and the ranking results were the same as the numpy method.

cfgov/paying_for_college/models/disclosures.py
```diff
@@ -359,7 +394,8 @@ def get_cohort(self, cohort):
         """
         cohorts = {
             'degrees_highest': School.objects.filter(
-                degrees_highest=self.degrees_highest),
+                degrees_highest=self.degrees_highest).exclude(
+                pk=FAKE_SCHOOL_PK),
```
chosak: Why is the fake school excluded from the degrees cohort, but not the state or control one? Similarly, should these also filter on `operating=True`?

This makes me wonder if it would be worth creating a custom manager method to filter out this fake school, so this doesn't need to be repeated everywhere:

```py
class SchoolQuerySet(models.QuerySet):
    def operating(self):
        return self.filter(operating=True).exclude(pk=FAKE_SCHOOL_PK)

class School(models.Model):
    ...

    objects = SchoolQuerySet.as_manager()
```

This consolidates all logic around the "fake school" in one place, instead of having to worry about it everywhere. Then, anywhere you do anything with schools, you could use `School.objects.operating()....` to start the query and be guaranteed to have all valid schools.

(I realize I just suggested a very similar thing on https://github.com/cfpb/cfgov-refresh/pull/5367#discussion_r349640239, so maybe I'm just trying out my shiny new hammer.)

cfgov/paying_for_college/tests/test_models.py
```diff
@@ -81,16 +83,16 @@ def test_loan_quiz_template(self):
...
-        self.assertEqual(
-            self.debt_page.get_template(HttpRequest()),
-            'paying-for-college/repaying-student-debt.html')
-
     def test_college_costs_template(self):
         self.assertEqual(
             self.college_costs_page.get_template(HttpRequest()),
             'paying-for-college/college-costs.html')
 
+    def test_debt_page_template(self):
```
chosak: Snce the templates are now all hardcoded since there are different page types, are these tests testing anything significant beyond whether Wagtail's `get_template` method works right?

cfgov/paying_for_college/disclosures/scripts/api_utils.py
```diff
@@ -99,52 +192,19 @@
...
-    'earnings.8_yrs_after_entry.percent_greater_than_25000',
-    'earnings.9_yrs_after_entry.mean_earnings',
-    'earnings.9_yrs_after_entry.percent_greater_than_25000',
-    'earnings.10_yrs_after_entry.working_not_enrolled.mean_earnings',
-    'earnings.10_yrs_after_entry.median',
-    'earnings.10_yrs_after_entry.percent_greater_than_25000'
 ]
 
 
+def api_school_query(school_id, fields):
```
chosak: I see a test for this method but then it isn't referenced anywhere else in the code. Is it a placeholder for things to come?

cfgov/paying_for_college/tests/test_models.py
```diff
@@ -10,8 +10,10 @@
 from django.test import TestCase
 from django.utils import timezone
 
+# from wagtail.wagtailcore import blocks
```
chosak:
```suggestion
```

cfgov/paying_for_college/tests/test_models.py
```diff
@@ -10,8 +10,10 @@
...
 import requests
 from paying_for_college.apps import PayingForCollegeConfig
+# from paying_for_college.blocks import GuidedQuiz
```
chosak:
```suggestion
```

cfgov/paying_for_college/models/disclosures.py
```diff
@@ -214,6 +226,8 @@ class School(models.Model):
     url = models.TextField(blank=True)
     degrees_predominant = models.TextField(blank=True)
     degrees_highest = models.TextField(blank=True)
+    program_count = models.IntegerField(blank=True, null=True)
+    program_most_popular = JSONField(blank=True, null=True)
```
chosak: Maybe the cow is out of the barn at this point, but is it worth doing some kind of check for Postgres on tests that touch this behavior, a la [this code](https://github.com/cfpb/cfgov-refresh/blob/e0663d69a87d4af239edecf6c91d5e8fdca52394/cfgov/prepaid_agreements/tests/test_views.py#L41) for the prepaid agreements use of Postgres search?

cfgov/paying_for_college/models/disclosures.py
```diff
@@ -187,6 +191,14 @@ def __str__(self):
         )
 
 
+def format_for_null(value):
```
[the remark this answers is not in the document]
higs4281: It's not standardized nicely, but we are trying hard not to affect the original tool, as it is still in heavy use and will stop students from enrolling if it breaks.

The format_for_null fixed a problem we had with the original API, which was errantly delivering the text string of "None" instead of the json value of null in some cases.

cfgov/paying_for_college/disclosures/scripts/update_colleges.py
```diff
@@ -68,77 +108,59 @@ def update(exclude_ids=[], single_school=None):
...
-                        school.zip5 = fix_zip5(str(school.zip5))
-                        school.save()
-                else:
-                    sys.stdout.write('-')
-                    sys.stdout.flush()
-                    NO_DATA.append(school)
+            sys.stdout.write('.')
+            sys.stdout.flush()
+            if data.get('school.operating') == 0:
+                school.operating = False
```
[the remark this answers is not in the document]
higs4281: Great idea.

cfgov/paying_for_college/models/disclosures.py
higs4281: Thank you for the catches on cohort exclusions.

The fake-school exclusion is tricky, because we need the fake school to show up in search and API calls, so that it can be used in demos and displayed in the tool, but we don't want it to throw off any cohort calculations, slight as the effects might be. 

It's a pretty obvious fake: Example University from Example Town, TX, url: example.com 
It does have a kicking mascot: the Fighting Exemplars!

It's also likely to be short-lived, as we hope not to need it after a demo period. So I'm not sure we want to embed it too deeply in code.

cfgov/paying_for_college/disclosures/scripts/api_utils.py
higs4281: This has a useful method for exploring the scorecard API, which we've needed to do a lot lately since the Dept. of ED just unveiled a batch of potentially valuable endpoints. It has also helped identify endpoints that mostly deliver nulls, which is hard to plan for.

cfgov/paying_for_college/models/disclosures.py
higs4281: I would vote not to support testing with other databases, but I'm unreasonably opinionated on this.

cfgov/paying_for_college/tests/test_models.py
higs4281: fixed.

cfgov/paying_for_college/tests/test_models.py
higs4281: fixed.

cfgov/paying_for_college/tests/test_models.py
higs4281: Yes, the template tests are not very valuable at this point.

cfgov/paying_for_college/models/disclosures.py
```diff
@@ -239,16 +253,21 @@ class School(models.Model):
...
         decimal_places=3,
         blank=True, null=True,
-        help_text="LOAN DEFAULT RATE AT 3 YRS")
+        help_text="LOAN DEFAULT RATE AT 5 YRS")
     median_total_debt = models.DecimalField(
         max_digits=7,
         decimal_places=1,
         blank=True, null=True,
-        help_text="MEDIAN STUDENT DEBT")
+        help_text="MEDIAN STUDENT DEBT 10 YRS AFTER ENROLLING")
```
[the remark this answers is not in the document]
higs4281: Great observation!

As the owner of the intent, I sure hope this wasn't intentional.
The total debt's precision seems appropriate, but the monthly is out of hand.
max_digits=7, decimal_places=2 would be more appropriate for a monthly value.

These are the max/min values for schools in our db:
median total debt: $39,000.0/$1,090.0
median monthly debt: $552.812416409/$15.946512012 (for those who like to budget precisely)

I tested migrating this change, and the values survive properly.

cfgov/paying_for_college/models/disclosures.py
higs4281: Oh, my. The models don't need any cohort-calculation code. That's all handled now in the process_cohorts script. Removing (and removing the fake-school issue there).

higs4281:
@chosak 
Many thanks for the thorough review. I think I've addressed the issues raised so far.

cfgov/paying_for_college/disclosures/scripts/update_colleges.py
```diff
@@ -54,12 +94,12 @@ def update(exclude_ids=[], single_school=None):
...
-        print("Updating {0}".format(base_query[0]))
+        base_query = School.objects.exclude(
+            pk__in=exclude_ids).filter(pk=single_school)
+        logger.info("Updating {}".format(base_query[0]))
     else:
-        base_query = School.objects.filter(operating=True)
-        if exclude_ids:
-            base_query = base_query.exclude(pk__in=exclude_ids)
+        base_query = School.objects.filter(
+            operating=True).exclude(pk__in=exclude_ids)
```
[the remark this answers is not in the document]
higs4281: agreed!

cfgov/paying_for_college/tests/test_models.py
chosak: You could move at least some of these (except for StudentLoanQuizPage, which has the block ID logic) from `get_template()` methods to a `template` property on the model:

```py
class CollegeCostsPage(PayingForCollegePage):
    template = 'paying-for-college/college-costs.html'
```

This takes care of the coverage. 😀

higs4281: My tunnel vision is showing. Thank you!

higs4281:
@chosak 
I *think* I've addressed all of your concerns. We wanted to get this code merged so front-end work can proceed before the holiday. If any of your concerns were not addressed, I'll gladly revisit.
Thanks again for the eagle-eyed review.

cfgov/paying_for_college/models/disclosures.py
chosak: Now that this has been merged, the cfgov-refresh tests fail if you don't have a `TEST_DATABASE_URL` environment variable set to point to a Postgres DB. The [current default](https://github.com/cfpb/cfgov-refresh/blob/10a0d58b666ee95c721dd37ff2f78b9bafa26249/cfgov/cfgov/settings/test.py#L7) is to try to use SQLite if this isn't defined, which causes several JSONField-related tests to fail with errors like "InterfaceError: Error binding parameter 14 - probably unsupported type." (To replicate, try running tox with that environment variable unset.) Travis [explicitly uses Postgres](https://github.com/cfpb/cfgov-refresh/blob/10a0d58b666ee95c721dd37ff2f78b9bafa26249/.travis.yml#L42) when running the tests, so this problem doesn't show up there.

We should either change the settings file and the docs ([1](http://cfpb.github.io/cfgov-refresh/installation/#install-postgres), [2](http://cfpb.github.io/cfgov-refresh/python-unit-tests/#running-tests)) to remove any reference of supporting SQLite, and require Postgres; OR mark the relevant tests with [`skipUnless`](https://github.com/cfpb/cfgov-refresh/blob/e0663d69a87d4af239edecf6c91d5e8fdca52394/cfgov/prepaid_agreements/tests/test_views.py#L56). I probably have a weak preference for the latter, but don't feel very strongly either.

higs4281: I'll mark relevant tests. Thanks for the follow-up.

PR #5335, opened 2019-11-06
Support testing of alternate home page designs

chosak:
This change adds the ability to use a new `NEW_HOME_PAGE` feature flag to toggle between the existing home page designs (both English and Spanish) to a new (placeholder) design.

Ping @contolini for your review.

## Testing

To test, run a local server and log in to Wagtail. Visit the flags area at http://localhost:8000/admin/flags/ and hit the "NEW FLAG" button to define a new flag named "NEW_HOME_PAGE". Click "ADD A CONDITION" to create a condition with name "parameter" and expected value "nhp".

Visiting either the English or Spanish home pages with a query string parameter of ?nhp=True should now activate the new (placeholder) design. Test with these pages:

- http://localhost:8000/ - standard home page
- http://localhost:8000/?nhp=True - new placeholder home page
- http://localhost:8000/es/ - standard Spanish home page
- http://localhost:8000/es/?nhp=True - new placeholder Spanish home page

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [x] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [x] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #5359, opened 2019-11-19
Homepage: Add initial carousel organism

anselmbradford:
Adds an initial front-end architecture for a carousel organism, for cycling through a sequence of blocks of content.

## Additions

- Adds Carousel organism JavaScript, CSS, and jinja template.
- Adds `single.js` script for scripts only on the homepage route.

## Testing

2. Enable alternate homepage design (via instructions in https://github.com/cfpb/cfgov-refresh/pull/5335)
3. Visit http://localhost:8000/?nhp=True and see there's a barebones carousel that you can click forward and backward through.

## Screenshots

carousel

### DIFF
--- a/cfgov/unprocessed/css/organisms/carousel.less
+++ b/cfgov/unprocessed/css/organisms/carousel.less
@@ -1,0 +1,31 @@
+.o-carousel {
+  width: 100%;
+
+  // Override default unordered list padding.
+  & ul {
+    padding: 0;
+  }
+
+  &_item {
+    border: 1px solid red;
+    height: 100px;
+    width: 100%;
+    padding: 30px;
+  }
+
+  &_item:nth-child(1) {
+    background: orange;
+  }
+
+  &_item:nth-child(2) {
+    background: greenyellow;
+  }
+
+  &_item:nth-child(3) {
+    background: cyan;
+  }
+
+  &_item:nth-child(4) {
+    background: forestgreen;
+  }
+}
--- a/cfgov/jinja2/v1/_includes/organisms/carousel.html
+++ b/cfgov/jinja2/v1/_includes/organisms/carousel.html
@@ -1,0 +1,28 @@
+{# ==========================================================================
+
+   carousel.render()
+
+   ==========================================================================
+
+   Description:
+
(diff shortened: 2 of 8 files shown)


anselmbradford:
@contolini

cfgov/unprocessed/css/organisms/carousel.less
```diff
@@ -0,0 +1,31 @@
...
+
+  // Override default unordered list padding.
+  & ul {
+    padding: 0;
+  }
+
+  &_item {
+    border: 1px solid red;
+    height: 100px;
+    width: 100%;
```
contolini: Down the road, how do you envision image aspect ratios being handled? Would we maintain a fixed height, center the image in the container, and overflow the sides? i.e.

0NrWT

contolini: In the same vein, we'll have to reuse these images for the previews by somehow cropping them into little squares. We could use the same technique as above. Loading the full-size images in the previews would preload them for us.

anselmbradford: That's a good question. One direction might be to follow the approach used in https://cfpb.github.io/design-system/components/featured-content-module

anselmbradford:
> This works well on the English homepage, but doesn't work properly on the Spanish one -- the panels appear, but the JS isn't loaded. Is that because the single.js file isn't loaded against that Spanish URL?

Ah, yes, that's correct. How do you think we should handle that? We can create a duplicate `single.js` in `/es/`, which is a shame, but I was also thinking this would be in Wagtail at some point and Carousel.js would move to the `on-demand` directory.

chosak:
@anselmbradford per our conversation offline, I just pushed a commit that moves the homepage JS to a new `routes/on-demand/home-page.js` file, referenced by our [custom, but infrequently used](https://github.com/cfpb/cfgov-refresh/search?l=Python&q=page_js&type=) [`page_js`](https://github.com/cfpb/cfgov-refresh/blob/5badef39b1e49269f1b37058f3d6478d7a74baf4/cfgov/v1/models/base.py#L297) property.

PR #5360, opened 2019-11-20
Homepage: Carousel no-js state

anselmbradford:
Hides carousel by default on page load and then shows it when the JS has initialized. Shows the carousel stacked in the no-js state.

## Changes

- Hides carousel by default till JS is initialized.
- Stacks carousel in no-js state.

## Testing

2. Load http://localhost:8000/?nhp=True and see that Carousel is hidden till JS is initialized.
3. Inspect the page and change the `js` class on the `html` element to `no-js` and see that the carousel items stack on top of each other and the buttons disappear.

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/carousel.html
+++ b/cfgov/jinja2/v1/_includes/organisms/carousel.html
@@ -13,7 +13,7 @@
 
 {% macro render() %}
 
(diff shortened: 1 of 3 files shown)


anselmbradford:
@contolini

contolini:
Smart and works great! 🎉

PR #5368, opened 2019-11-21
Homepage: Add image/text content to carousel

anselmbradford:
## Changes

- And initial image and text layout to carousel items and thumbnails.

## Testing

2. Visit http://localhost:8000/?nhp=True and click around carousel buttons.

## Screenshots
Screen Shot 2019-11-21 at 6 17 16 PM

## Notes

- This sets an arbitrary fixed height on the main item and thumbnails. It might be better if all thumbnails and the main item set their height to the highest item/thumbnail so if the text is too tall it doesn't break the layout. Anyone see an easy way to do this?


anselmbradford:
> Should that right edge also be highlighted?

Yes, I think it should. However, this is the default focus state of a button showing up. I think we'll design a focus state that will replace that. If it is a state that is similar to the default one we'll need to change the stacking order (z-index) of the buttons when they are selected. So, important to note, but not something we need to address yet in this PR.

PR #5372, opened 2019-11-25
Handle missing carousel organism on home page

chosak:
This commit fixes a bug with the prototype carousel created for a new home page design. When running the current production home page, there's no carousel element, so this causes a JS console error.

This change handles the case where there is no carousel organism, while still properly handling the case when there is one.

@contolini

## Testing

To test: `gulp clean && gulp build`, visit http://localhost:8000.

## Checklist


anselmbradford:
Long term we'll have it as on-demand attached to a block in Wagtail and the check won't be necessary at that point, right?

Scotchester:
> Long term we'll have it as on-demand attached to a block in Wagtail and the check won't be necessary at that point, right?

Yeah, when you have a Wagtail block for it, the JS file can be set as the block's `Meta` `media`.

PR #5375, opened 2019-11-25
Fix iRegs wayfinder JS errors on odd paragraphs

Scotchester:
Certain paragraphs in iRegs have odd `data-label` values, e.g., the first five paragraphs on 
https://www.consumerfinance.gov/policy-compliance/rulemaking/regulations/1002/Interp-9/. This commit puts certain statements containing methods that were assuming other formats for that attribute inside conditionals to prevent those methods from erroring on the unexpected input.

/cc @contolini, @cfarm 

## Testing

1. Visit https://www.consumerfinance.gov/policy-compliance/rulemaking/regulations/1002/Interp-9/
1. Open the dev tools console and start scrolling down
1. Observe errors popping in, e.g., `Uncaught TypeError: Cannot read property 'split' of undefined`
1. Pull this branch
1. `yarn run gulp scripts:apps`
1. Visit http://localhost:8000/policy-compliance/rulemaking/regulations/1002/Interp-9/
1. Open the dev tools console and start scrolling down
1. Observe no errors!

## Checklist

### Browsers

- [x] Chrome on desktop
- [ ] Firefox
- [x] Safari on macOS
- [x] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### DIFF
--- a/cfgov/unprocessed/apps/regulations3k/js/permalinks-utils.js
+++ b/cfgov/unprocessed/apps/regulations3k/js/permalinks-utils.js
@@ -122,13 +122,16 @@ const getCommentMarker = label => {
   let commentParagraphID = '';
 
   const splitCurrentParagraph = label.split( 'Interp' );
+  /* A guard clause just in case any comment paragraphs have an invalid ID that
+     doesn't have 'Interp' in it */
+  if ( splitCurrentParagraph.length < 2 ) {
+    return '';
+  }
   if ( splitCurrentParagraph !== null ) {
     commentedParagraphID = splitCurrentParagraph[0]
       .split( '-' );
-    if ( splitCurrentParagraph[1] ) {
-      commentParagraphID = splitCurrentParagraph[1]
-        .split( '-' );
-    }
+    commentParagraphID = splitCurrentParagraph[1]
+      .split( '-' );
   }
   if ( commentedParagraphID !== null ) {
     commentedSection = commentedParagraphID[0];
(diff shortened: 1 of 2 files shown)


niqjohnson:
@Scotchester: Whew, this one turned out to be tricker than I thought. I pushed an update that I *think* will handle the issue. Basically, interpretation paragraphs should have a `data-label` with "Interp" in it to be valid; the examples you saw here didn't have that in the `data-label`, so the wayfinder code that transforms the `data-label` into the marker text in the wayfinder was choking. I added a guard clause that makes the wayfinder just show "Comment" as the marker for any interpretation paragraphs with invalid `data-label`s now. That seems to be working for all of the commentary sections I spot checked.

At first I tried to get fancy and modify the code to figure out what the correct maker should be for paragraphs with invalid IDs like this. I can definitely imagine other IDs out there that are invalid in different ways, though, so that seemed like more work than it was worth. Instead I think the best way is to show the generic "Comment" as the marker in the wayfinder and then change the invalid ID to a valid one in the Regdown when we see cases like this.

Also, I know there are other ways than a guard clause to do this, so feel free to change to another method if you have any other ideas. I didn't look at any JS unit tests to see if we need to cover this case yet, either.

niqjohnson:
@Scotchester: OK, I just added a unit test for the guard clause, but I have no idea if I did it right, so definitely feel free to redo that, too.

Scotchester:
Hahah, wow, I really didn't test this PR very thoroughly, did I? 😆 Sorry, @niqjohnson, and thanks for picking up the slack. Your changes are working great for me!

PR #5378, opened 2019-11-26
Fix webpack mode warning

anselmbradford:
## Changes

- Adds missing `mode` configuration to webpack config.
- Move webpack mode ~~and minimization~~ settings to constants.

## Testing

1. `gulp build` on `master` shows a warning for `mode`.
2. `gulp build` on this branch does not.
3. If you have `export NODE_ENV=development && gulp clean && gulp build` the code should be unminimized. If you `export NODE_ENV=production && gulp clean && gulp build` the code should be minimized.

## Screenshots

Fixes this warning
Screen Shot 2019-11-26 at 11 46 43 AM

### DIFF
--- a/config/webpack-config.js
+++ b/config/webpack-config.js
@@ -9,7 +9,11 @@ const TerserPlugin = require( 'terser-webpack-plugin' );
 
 // Constants
 const COMMON_BUNDLE_NAME = 'common.js';
-const WEBPACK_MODE = 'production';
+
+/* This sets the default mode for webpack configurations to satisfy the need
+   of webpack to have a `mode` set.
+   This value gets overridden when NODE_ENV=development. */
+const WEBPACK_MODE_DEFAULT = 'production';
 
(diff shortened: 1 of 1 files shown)


config/webpack-config.js
```diff
@@ -10,6 +10,11 @@ const TerserPlugin = require( 'terser-webpack-plugin' );
 // Constants
 const COMMON_BUNDLE_NAME = 'common.js';
 
+/* This sets the default mode for webpack configurations to satisfy the need
+   of webpack to have a `mode` set.
+   This value gets overridden when NODE_ENV=development. */
```
Scotchester:
```suggestion
   This value gets overridden when NODE_ENV=development.
   See the `if ( envvars.NODE_ENV === 'development' )` block below. */
```

anselmbradford: Not super-ly a fan of the literal code in the comment since it becomes two things that need to be kept in sync. But if it helps 👍

PR #5381, opened 2019-11-26
Move u-visually-hidden markup to aria-label attribute

anselmbradford:
We use a `u-visually-hidden` hack to hide content. In many locations this is used for screenreaders and the content is better moved to an `aria-label` attribute intended for this purpose. 

## Changes

- Move instances of `u-visually-hidden` content meant for screenreaders to `aria-label` attribute.

## Testing

1. You can check these in https://wave.webaim.org/extension/ and/or using your favorite screenreader testing steps.

## Screenshots

Before:
Screen Shot 2019-11-26 at 2 32 53 PM

After:
Screen Shot 2019-11-26 at 2 32 46 PM

## Notes

- Moving the `u-visually-hidden` content that's inside a label or heading to `aria-label` brings up an empty heading or empty label error, so those have been left in place.


anselmbradford:
Hold till this can be looked at closer for a better solution.

cfgov/jinja2/v1/_includes/organisms/post-preview.html
```diff
@@ -145,8 +145,8 @@
                 <time class="calendar-icon"
                       data-month="{{ event.start_dt | date('%b') }}"
                       data-day="{{ event.start_dt | date('%-d') }}"
-                      datetime="{{ event.start_dt | date('%c') }}">
-                    <span class="u-visually-hidden">{{ event.start_dt | date('%b %-d, %Y') }}</span>
+                      datetime="{{ event.start_dt | date('%c') }}"
+                      aria-label="{{ event.start_dt | date('%b %-d, %Y') }}">
```
Scotchester: In this case, It's just a visual flourish that's repeating the main date slug that preceded it, so let's hide this one from screen readers. Since we changed the layout of these previews to put the image on the right, it makes less sense to have this at all anymore, but removing it is a discussion for another day.

```suggestion
                      aria-label="{{ event.start_dt | date('%b %-d, %Y') }}">
                      aria-hidden="true">
```

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
```diff
@@ -69,11 +69,12 @@ <h3 class="h4 u-mb5">Search within</h3>
                     </div>
                     <div class="flex-all">
                         <div class="m-btn-inside-input input-contains-label">
-                            <label for="searchText" class="input-contains-label_before input-contains-label_before__search">
+                            <label for="searchText"
+                                   class="input-contains-label_before input-contains-label_before__search"
+                                   aria-label="The term to search for">
```
Scotchester: I think in this case you would apply the `aria-label` to the `input`, not the `label`.

cfgov/regulations3k/jinja2/regulations3k/regulations3k-secondary-navigation.html
```diff
@@ -43,18 +43,12 @@
             Table of Contents
         </span>
         <span class="o-expandable_header-right o-expandable_link">
-            <span class="o-expandable_cue o-expandable_cue-open">
-                <span class="u-visually-hidden">
-                    {{ _('Show') }}
-                    {{ expandable_cue_additional_text }}
-                </span>
+            <span class="o-expandable_cue o-expandable_cue-open"
+                  aria-label="{{ _('Show') }} {{ expandable_cue_additional_text }}">
```
Scotchester: I'm not sure if these will be read, since `span` is not an interactive element. https://developer.paciellogroup.com/blog/2017/07/short-note-on-aria-label-aria-labelledby-and-aria-describedby/

Given the reported inconsistency between browser/screenreader combos, might be safer to leave these as-is.

cfgov/jinja2/v1/_includes/organisms/post-preview.html
anselmbradford: This would have only `aria-hidden` and not `aria-label` then.

Scotchester: Yes, that's right.

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
anselmbradford: I don't think so?

> The aria-label attribute is used to define a string that labels the current element. Use it in cases where a text label is not visible on the screen.
([ref](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute))

This label only contains an icon, so this `aria-label` is labeling that icon.

cfgov/regulations3k/jinja2/regulations3k/regulations3k-secondary-navigation.html
anselmbradford: Okay I'll revert it. Perhaps the svg_icon shortcode should provide a means to pass in a `title`, per the instructions in inline SVG on https://css-tricks.com/accessible-svgs/

cfgov/prepaid_agreements/jinja2/prepaid_agreements/index.html
anselmbradford: Thinking about this further, I think you're right. We'd `aria-label` the `input` and `aria-hidden="true"` the `label` in case that would be read.

cfgov/jinja2/v1/_includes/organisms/post-preview.html
anselmbradford: Fixed

anselmbradford:
@Scotchester Ready for review again. Thanks!

cfgov/jinja2/v1/_includes/organisms/post-preview.html
Scotchester: Oh, I see, I messed up my suggestion 😆

cfgov/regulations3k/jinja2/regulations3k/regulations3k-secondary-navigation.html
Scotchester: Yeah, I think that'd be a good idea.

PR #5382, opened 2019-11-26
Fix accessibility issue with alternative text in image link

anselmbradford:
## Changes

- Adds text to `alt` attribute in images that are links in the form explainers.

## Testing

1. http://localhost:8000/owning-a-home/loan-estimate/ show images in the brand footer at the bottom that have alt text equal to the links that appear below them. Production has `alt=""` for the same images.

## Screenshots

Before:
Screen Shot 2019-11-26 at 4 03 40 PM

After:
Screen Shot 2019-11-26 at 4 04 34 PM


Scotchester:
The problem with this approach is that a screen reader is going to read those links duplicatively. EIt'll go,

"Link: Learn what comes next"
"Link: Learn what comes next"
"Link: Explore interest rates"
"Link: Explore interest rates"
"Link: Learn how"
"Link: Learn how" (doesn't provide enough context already)

Having duplicate links next to each other is another violation. I think I wold suggest `aria-hidden="true"` to hide the links around the images.

anselmbradford:
> Edit: Hmm... that idea might not work for a focusable element. I guess the "proper" fix would be to wrap the whole chunk in one <a> element? You get into some text styling stickiness with that though.

I don't think we can do that because it has a line of non-link text between the image and text:
```
<image link>
Plain text
<link>
```

What's the argument that this PR is not correct? Consider the scenario where you have this:

```
<a href="/somewhere">Learn what comes next</a>
<p>Ready to request Loan Estimates?</p>
<a href="/somewhere">Learn what comes next</a>
```

Would you wrap the entire thing in a link or hide the first link? I don't think so. I think repeating the link would be the correct thing as it describes what is there. Having a redundant link may be an issue, but it is a separate content issue to that of the accessibility fix here. The repetition of link content seems like an annoyance, whereas the lack of image link alt text seems like a barrier.

Scotchester:
> I don't think we can do that because it has a line of non-link text between the image and text:

You could, it just requires some additional styling gyrations, like we were just talking about in chat regarding the homepage cards.

> What's the argument that this PR is not correct?

The argument that this PR is not ideal ("not correct" is maybe too strong) is the way screenreaders will read these links that I laid out in my previous comment:

"Link: Learn what comes next"
"Link: Learn what comes next"
"Link: Explore interest rates"
"Link: Explore interest rates"
"Link: Learn how"
"Link: Learn how"

I'd say for both the first and last pairs of links, it's not at all going to be clear to a screenreader user that those links are, in fact, repeated links to the same destination. That may be fixable by changing the link text or adding `aria-label` attributes, though, and if that were done, I agree that a repeated link annoyance is better than a lack of alt text on an image-only link.

anselmbradford:
@Scotchester I added an `image_label` for the images with the heading text of the linked page or document. Does that work?

Scotchester:
> @Scotchester I added an `image_label` for the images with the heading text of the linked page or document. Does that work?

No, I don't think it does. Sorry if I was not clear, but I don't think that we want the text that's read for the image link to be different than the text that's read for the text link. We want them to be the same, and the the text needs to be specific enough that, when hearing the text repeated, it's clear that the two links go to the same place. In the case of something like "Link: Learn how; Link: Learn how", it seems probable that many users would assume those are links to learn how to do two different things, appearing after two different intro paragraphs. (And "Learn how" is bad link text even if it wasn't repeated. Learn how to what?)

anselmbradford:
@Scotchester I've added an `aria-label` in https://github.com/cfpb/cfgov-refresh/pull/5382/commits/6a8fafb2dd7c341dec30d375f242bb3b632609e8 where the `aria-label` in the link matches the `alt` of the image.

PR #5383, opened 2019-11-26
Move new home page content into Wagtail

chosak:
This set of changes moves the content for the new home page design into configuration for the Wagtail `HomePage` page type.

Serving the new content from hardcoded values is still supported via the existing `NEW_HOME_PAGE` feature flag. A new `NEW_HOME_PAGE_IN_WAGTAIL` feature flag can be used to instead serve the content from Wagtail. If both flags are enabled, the Wagtail one takes precedence.

There are several changes here related to Wagtail InlinePanels and the django-modelcluster package that powers them. In the version of Wagtail that we are using (1.13), [nested InlinePanels](https://github.com/wagtail/wagtail/issues/1952) are not supported at all. In versions >2.0, nested InlinePanels are supported via a workaround
detailed on https://github.com/wagtail/wagtail/issues/5511. Full support for nested InlinePanels will be available in Wagtail starting in version 2.8, see https://github.com/wagtail/wagtail/issues/5566.

It turns out that we can use the nested InlinePanel workaround on Wagtail 1.13, but to do so we need to increment the version of django-modelcluster from 3.1 to 4.4. Unfortunately, Wagtail 1.13
requires django-modelcluster<4.0 even though [4.4 still supports it](https://github.com/wagtail/django-modelcluster/releases), along with Python 2 (support for both of these is not removed until django-modelcluster 5.0). So we want to force use of django-modelcluster==4.4, even though it technically conflicts. Unfortunately, our use of tox_pip_extensions_ext_venv_update prevents this (it doesn't allow you to specify any conflicting versions, even though pip does out of the box; see https://github.com/Yelp/venv-update/issues/105 on this including a comment from an old colleague, hi @acrewdson). So, I've (hopefully temporarily) disabled that until we are on Wagtail 2.

After all that you may be wondering why use nested InlinePanels at all. The alternative to this would be to use StreamFields, but this felt like a worse implementation to me: harder migrations, worse user experience in the Wagtail editor, etc. I welcome any feedback on this choice.

These changes include a new Django management command to automatically populate the Wagtail fields from the hardcoded values. After running the new v1 migration on your database, run

```
cfgov/manage.py migrate_home_page_content
```

to update both the English and Spanish pages. This creates an (unauthored) revision for the pages, but does not publish it.

@contolini 

## Screenshots

New tabs when editing the home page:

image

Carousel tab:

image

InfoUnits tab, including a nested InlinePanel for the links:

image

Cards tab:

image

## Todos

- Still no tests :(

## Checklist


cfgov/v1/models/home_page.py
```diff
@@ -273,6 +199,9 @@ class HomePage(CFGOVPage):
         ObjectList(content_panels, heading='General Content'),
         ObjectList(CFGOVPage.sidefoot_panels, heading='Sidebar'),
         ObjectList(CFGOVPage.settings_panels, heading='Configuration'),
+        ObjectList([
+            InlinePanel('carousel', min_num=4, max_num=4, label="Carousel"),
```
Scotchester: Naming this `carousel_items` might make the call in the template (`{{ carousel.render( page.carousel.all() ) }}` :arrow_right: `{{ carousel.render( page.carousel_items.all() ) }}`) a bit more clear. The former sorta implies that you can have multiple carousels.

cfgov/jinja2/v1/_includes/organisms/carousel.html
```diff
@@ -9,35 +9,34 @@
...
-        <p><a href="{{ item.link.url }}">{{ item.link.text }}</a></p>
+        <h2>{{ item.heading }}</h2>
+        <p>{{ item.text }}</p>
+        <p><a href="{{ item.link_url }}">{{ item.link_text }}</a></p>
     </div>
     <div class="o-carousel_item-visual">
         <img class="o-carousel_item-img"
-             src="{{ item.image.url }}"
-             alt="{{ item.image.alt_text }}"
+             {% set original_image = image( item.image, 'original' ) %}
```
Scotchester: I'd suggest moving this up one line, to not break up the `` tag. (Below on the thumbnail, too.)

cfgov/jinja2/v1/_includes/organisms/carousel.html
```diff
@@ -48,15 +47,18 @@ <h2>{{ item.title }}</h2>
...
             <img class="o-carousel_thumbnail-img"
-                src="{{ item.image.thumbnail_url }}"
-                alt="{{ item.image.alt_text }}">
+                {% set thumbnail_image = image( item.image, 'fill-64x64' ) %}
+                src="{{ thumbnail_image.url }}"
+                alt="{{ item.image.alt }}">
         </div>
-        <h6 class="o-carousel_thumbnail-text">{{ item.title }}</h6>
+        <h6 class="o-carousel_thumbnail-text">
+            {{ item.thumbnail_heading | default( item.heading, true ) }}
```
Scotchester: I always forget about this `default` filter, and it's super handy!

cfgov/v1/models/home_page.py
chosak: Originally I did try to name this carousel_items. Unfortunately, that conflicts with [this similar reference](https://github.com/wagtail/wagtail/blob/b9c470df37ee971ad6d60827e631e36897b168e6/wagtail/tests/testapp/models.py#L194) in the Wagtail test app, which caused our tests to break since we include that app when we run them. Looking at this again, though, I might be able to get around that by changing the ParentalKey from Page to CFGOVPage. I'll give that a try.

chosak:
@Scotchester @virginiacc I've significantly expanded this PR to include moving all new home page content into Wagtail. See updated description above.

I should also note that these changes include the minor frontend change we've discussed internally to [restore an empty `alt` tag on images](https://github.com/cfpb/cfgov-refresh/pull/5383/files#diff-58c0cf3da5e7e11896b654441858b91cR50) if no alt text exists.

cfgov/v1/models/home_page.py
chosak: This change worked. By adding the `carousel_items` reference to CFGOVPage instead of Page, it no longer conflicts with the Wagtail test app.

cfgov/v1/models/home_page.py
```diff
@@ -289,8 +68,21 @@ class HomePage(CFGOVPage):
...
+        ], heading='Carousel'),
+        ObjectList([
+            InlinePanel(
+                'info_units', min_num=6, max_num=6, label="Info Unit"
+            ),
+        ], heading='Info Units'),
+        ObjectList([
+            FieldPanel('card_heading'),
+            InlinePanel('cards', min_num=3, max_num=3, label="Card"),
+        ], heading='Cards')
```
Scotchester: Suggest putting these before Configuration to keep the Configuration tab consistently on the far right for all page types.

cfgov/jinja2/v1/home-page/index_new.html
```diff
@@ -1,16 +1,15 @@
...
 
 {% import "organisms/card-group.html" as card_group with context %}
+{% import "organisms/carousel.html" as carousel with context %}
 
 {% block title -%}
     {{ _('Consumer Financial Protection Bureau') }}
 {%- endblock %}
 
 {% block content_main %}
 {% block carousel %}
```
Scotchester: Not directly related to this PR, but does having a specific named `block` for this serve any purpose? The info unit and card sections don't have their own named blocks.

cfgov/v1/models/home_page.py
Scotchester: One more comment, I guess. while the Carousel would need a lot of work to support something other than exactly four items, perhaps we should be more flexible on the info units and cards? /cc @sonnakim

cfgov/v1/models/home_page.py
```diff
@@ -381,6 +401,133 @@ class HomePageExcludedUpdates(models.Model):
...
+    ))
+    body = models.TextField(max_length=160, help_text=(
+        "160 characters maximum (including spaces)."
+    ))
+    link_text = models.CharField(max_length=35, help_text=(
+        "35 characters maximum (including spaces). "
+        "Lead with a verb, and be specific."
+    ))
+    # TODO: Change this to use a URLField that also allows relative links.
+    # https://code.djangoproject.com/ticket/10896
```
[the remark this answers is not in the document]
chosak: Either of those could work. There are some packages out there that do this ([for example](https://pypi.org/project/django-url-or-relative-url-field/)) as well.

I am not sure of the likelihood of using absolute URLs at all here. For example, we might want to link off-site ([order brochures](https://pueblo.gpo.gov/CFPBPubs/CFPBPubs.php), or to a YouTube video), in which case we need to support absolute URLs. On the other hand, if we allow that, editors will likely enter URLs that include https://www.consumerfinance.gov/, which we don't want (because it'd break links on other servers/locally).

The downside of the page chooser is that it's almost certain that we will want the link text to differ from the page title, which the page chooser doesn't easily support.

The use cases we probably need are:

1. Other Wagtail pages, in which case use a page chooser, but specify alternate text.
2. Other non-Wagtail pages on the site, in which case we want only relative links, never using e.g. https://www.consumerfinance.gov/. (To me the complexity of supporting both this and case 1 eliminate any UX improvement of the page chooser.)
3. Pages off-site, in which case we always want full URLs.

I'm not really sure what the best UX is here -- maybe a URL field that only supports absolute URLs if the domain isn't one of ours? Either way, using a plain CharField for now should be sufficient for the first implementation. Is there a more appropriate comment I can put for the TODO here to better sum this up?

cfgov/v1/models/home_page.py
chosak: I thought of this; it'd be nice, for example, to enforce that the number of cards or info units is always divisible by 3. I'm not sure that's easy to do in Wagtail because of how it relies on Django's built-in [formset validation](https://docs.djangoproject.com/en/2.2/topics/forms/formsets/#validating-the-number-of-forms-in-a-formset).

cfgov/jinja2/v1/home-page/index_new.html
chosak: This might be leftover from previous implementations. We can remove if it's not needed.

cfgov/v1/models/home_page.py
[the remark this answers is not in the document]
Scotchester: > The downside of the page chooser is that it's almost certain that we will want the link text to differ from the page title, which the page chooser doesn't easily support.

I still don't see this as a downside. You already have to have a separate field for the link text.

I suppose it'd be good to get a UX professional's take on it, but IMO, the best thing would be:

1. A page chooser for internal Wagtail pages
2. A `RegExField` for anything else, that validates that if the URL contains `consumerfinance.gov` it errors and requires you to make it relative
3. A link text field for either case

cfgov/v1/models/home_page.py
Scotchester: I don't even know that it's necessary to make it divisible by 3. Allow anything from some reasonable minimum (2?) to some reasonable maximum (6?) We could use CSS to change the layout to 2-column if it's an even number, which would eliminate the worst-looking case of 4 items in a 3-column grid.

cfgov/v1/models/home_page.py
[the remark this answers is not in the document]
chosak: Thanks for the writeup @Scotchester. I'm going to merge this as-is and open an internal issue to solicit further UX feedback on the chooser.

PR #5384, opened 2019-11-26
Homepage: Set focus styles on carousel thumbnails

anselmbradford:
## Changes

- Set focus state on carousel thumbnails.
- Vertical-align the slide text content to the top.
- Remove unneeded `display: table-row` rule.

## Testing

1. Pull branch, `gulp clean && gulp build`
2. Visit http://localhost:8000/?nhp=True
3. Click into the carousel and then use the tab key to navigate to the thumbnails. Note that they get a focus border around them. Use the spacebar to "click" them. Use shift+tab key to go in the reverse order.

## Screenshots

Screen Shot 2019-11-26 at 5 04 45 PM

### DIFF
--- a/cfgov/unprocessed/css/organisms/carousel.less
+++ b/cfgov/unprocessed/css/organisms/carousel.less
@@ -32,10 +32,6 @@
 
   &_item {
     box-sizing: border-box;
-
-    &:not( .u-hidden ) {
-      display: table-row;
-    }
   }
 
   &_item-text {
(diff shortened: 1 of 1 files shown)


anselmbradford:
@contolini

anselmbradford:
> One minor curious thing: there's one tab stop between "About Us" and the left arrow in the carousel. Where is that tab stop? I don't see anything highlighted.

Hmm I'll check into it!

anselmbradford:
> One minor curious thing: there's one tab stop between "About Us" and the left arrow in the carousel. Where is that tab stop? I don't see anything highlighted.

Oh this is probably the tab-trigger, https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/js/modules/TabTrigger.js#L42-L43. We'll want to look at this when working on the new menu.

PR #5389, opened 2019-12-02
Homepage: mobile carousel items

anselmbradford:
## Changes

- Add mobile carousel appearance.

## Testing

1. Pull branch and build.
2. Visit http://localhost:8000/?nhp=True and resize the page and see that the thumbnails lose their selected state and show links.
3. Focus state should be the whole button at desktop when tabbing. It should be the image and links at mobile size.

## Screenshots

Screen Shot 2019-12-02 at 11 27 21 AM
Screen Shot 2019-12-02 at 11 38 07 AM

## Notes

- I originally used the same markup for both mobile and desktop, however, to get the focus behavior to work correctly when tabbing through the content we'd need to set the tabindex to ignore the thumbnail text and image links at desktop sizes, these would then need to be removed via JavaScript from the mobile size. I opted instead to duplicate the markup, which is unfortunate, but I think it is more maintainable. The original commit can be viewed at https://github.com/cfpb/cfgov-refresh/commit/5d9c3c87c7f18571d051f94c9d426ed871ef57cf

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/carousel.html
+++ b/cfgov/jinja2/v1/_includes/organisms/carousel.html
@@ -32,11 +32,7 @@
     <div class="o-carousel_item-text">
         <h2>{{ item.title }}</h2>
         <p>{{ item.body }}</p>
-        <p>
-            <a href="{{ item.link.url }}">
-                {{ item.link.text }}
-            </a>
-        </p>
+        <p><a href="{{ item.link.url }}">{{ item.link.text }}</a></p>
     </div>
     <div class="o-carousel_item-visual">
         <img class="o-carousel_item-img"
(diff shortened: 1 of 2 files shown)


cfgov/jinja2/v1/_includes/organisms/carousel.html
```diff
@@ -58,8 +58,32 @@ <h2>{{ item.title }}</h2>
 </button>
 {% endmacro %}
 
+{#
+    Since desktop uses a button and mobile uses links,
+    we duplicate the markup here but use a div instead of a button.
+#}
+{% macro render_carousel_thumbnail_mobile( item, is_selected=false ) %}
+<div class="o-carousel_thumbnail{% if is_selected %} o-carousel_thumbnail-selected{%endif%}">
```
chosak:
```suggestion
<div class="o-carousel_thumbnail{% if is_selected %} o-carousel_thumbnail-selected{% endif %}">
```

cfgov/jinja2/v1/_includes/organisms/carousel.html
```diff
@@ -58,8 +58,32 @@ <h2>{{ item.title }}</h2>
...
+        <div class="o-carousel_thumbnail-visual">
+            <a href="{{ item.link.url }}">
+                <img class="o-carousel_thumbnail-img"
+                    src="{{ item.image.thumbnail_url }}"
+                    alt="{{ item.image.alt_text }}">
+            </a>
+        </div>
+        <p class="o-carousel_thumbnail-text">
+            <a href="{{ item.link.url }}">
+                {{ item.title }}
```
chosak: Per @sonnakim's style guide, there's a separate "thumbnail heading" entry (30 characters maximum) separate from the regular "heading" (45 characters maximum). I realize I need to add that to the stub content. Should I do that here or in a separate PR?

anselmbradford: Separate PR please

anselmbradford:
> Also, the highlighting of menu items at desktop is a little unexpected. I wouldn't expect the currently selected tab to turn from green to grey on hover; I'd expect it to stay green. It's especially surprising if you click to change tabs, because the green doesn't appear until you mouse away.

This'd be a question for @huetingj Maybe we do a darker green over something on hover?

huetingj:
👆 We looked at this during team time today. Would like the selected item's bar to stay CFPB green at all times. So, stay green when a user hovers over it as well as when a user hovers over the other non-selected items. @Scotchester @virginiacc  — I heard one of you may be helping us with FEWD work while @anselmbradford is out. Would you mind making this change? Thank you!!

PR #5393, opened 2019-12-02
Homepage: Add analytics code to carousel

anselmbradford:
## Changes

- Adds analytics code to carousel interactions.

## Testing

1. Pull branch
2. Add console.log statements above the `Analytics.sendEvent` entries in this PR with the value of the `label`.
3. `gulp clean && gulp build`
4. Visit http://localhost:8000/?nhp=True and click the next/previous buttons and/or the thumbnails in the carousel. You should see outputs like:

Screen Shot 2019-12-02 at 6 45 46 PM

Once all slides are seen you will see "All Items Seen"

## Notes

- Once all slides are seen the "All Items Seen" event will fire each time the slide is changed since there's no logic to reset the count. I'm not sure if this is the desired behavior or not.

### DIFF
--- a/cfgov/unprocessed/js/organisms/Carousel.js
+++ b/cfgov/unprocessed/js/organisms/Carousel.js
@@ -26,7 +26,7 @@ function Carousel( element ) { // eslint-disable-line max-lines-per-function
 
   // Track how many unique slides the user has seen in the carousel.
   let _slideVisitCount = 1;
-  const ALL_SLIDES_VISITED = 15;
+  const ALL_SLIDES_VISITED = Math.pow( 2, _items.length ) - 1;
   let _allSlidesVisited = false;
 
   /**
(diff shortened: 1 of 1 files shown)


kelleyholden:
@anselmbradford to your note above - I would say we would rather it didn't fire each time if there is a way to build that logic in, but if that's not an option we can let it run and see if it's necessary to update.

cfgov/unprocessed/js/organisms/Carousel.js
```diff
@@ -80,17 +89,61 @@ function Carousel( element ) {
...
+     * We use a bitmask to see if all slides have been visited:
+     * 0001 = item one (1 decimal).
+     * 0010 = item two (2 decimal).
+     * 0100 = item three (4 decimal).
+     * 1000 = item four (8 decimal).
+     * 1111 = all visited (15 decimal).
+     *
+     * We need to map 0, 1, 2, 3 item indices to bitwise digits to 1, 2, 4, 8.
+     */
+    _slideVisitCount |= [ 1, 2, 4, 8 ][newCurrItemIndex];
```
chosak: Would this be more future-proof to a variable number of slides?

```suggestion
    _slideVisitCount |= Math.pow( 2, newCurrItemIndex );
```

Then you'd also want to remove the hardcoded `ALL_SLIDES_VISITED` and instead compare against `Math.pow( 2, _items.length ) - 1` (or something similar).

Scotchester: That makes sense. I'll try it after we sort out this other stuff.

Scotchester: Updated! Great suggestion; thanks.

PR #5410, opened 2019-12-09
Remove old enforcement categories

wpears:
Depends on #5344 

# Removals

- Removes old enforcement categories after switching to new ones in #5344

## Testing

1. Pull branch
2. `docker-compose exec python2 /bin/bash`
3. `./cfgov/manage.py migrate`
4. Note that in the enforcement filter control, there are only the two new categories


cfarm:
> Note that in the enforcement filter control, there are only the two new categories

Does this refer to the front end here?
Screen Shot 2019-12-30 at 6 13 11 PM

Or do you mean the admin?

If the former, I already see only the two new categories in the filter UI, just when checking out this branch, and without running a migration: http://localhost:8333/policy-compliance/enforcement/actions/

Is that expected? If so, how else can I test/confirm that this branch works and the old categories are gone?

If the latter, please share a link to the admin page.

wpears:
@cfarm Yeah looks like since the remaining category options are just a subset of the current options (because we added the new ones without removing the old ones last PR) that the migration isn't actually required for you to see the results of this update in both the admin and the enforcement pages. So to test, just checking out the branch and see the 2 rather than the 5 options should suffice.

The migration is still needed though from a cleanliness perspective, since we're perturbing the model and if we don't commit the migration now, someone else will have to puzzle over it when making a migration.

PR #5413, opened 2019-12-09
New homepage: prevent tabbing to links in hidden carousel slides

virginiacc:
When tabbing through the home page, focus is temporarily lost after the active slide's link because links in the inactive slides appear in the tab order. This ensures that focus goes directly from the active slide's link to the first of the slideshow thumbnail buttons. (GHE 63)

## Additions

- Hidden class for carousel items

## Changes

- Include visibility as well as opacity in the slide transitions

## Testing

1. Check out branch
2. Navigate to [home page](http://localhost:8000/?nhp=True)
3. Tab through the page and verify that focus goes directly from link in active tab to first of 4 carousel slide buttons
4. Click through carousel and verify that slide transitions work as expected

### Browsers

- [x] Chrome on desktop
- [x] Firefox
- [x] Safari on macOS
- [x] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [x] Keyboard friendly
- [ ] Screen reader friendly


virginiacc:
@chosak Thanks for reviewing! I'll merge this now, but maybe we can open another issue to consider the general carousel tab order.

PR #5426, opened 2019-12-16
Includes script to update oaa status from a csv

wpears:
@cfarm
Closes jira 3995
## Additions

- csv of urls and target statuses
- A script that applies these statuses to each page with matching urls, replacing the blob status in wagtail (getting us to parity with statuses in enforcement pages).

## Testing

1. Pull branch
2. ./cfgov/manage.py runscript update_oaa_status
3. Child pages of /administrative-adjudication-proceedings/administrative-adjudication-docket/ should have their sidebar statuses updated


cfarm:
this needs to be linted to fix the failing travis test

also I found at least one page that was in the csv, that didn't get the update locally that should have, and the status is missing for it now: http://localhost:8333/administrative-adjudication-proceedings/administrative-adjudication-docket/american-express-travel-related-services-company/

cfarm:
NOT APPROVED ❌ 🚫 🚯

higs4281:
Drive-by observation: 
The csv (from Excel?) has a BOM that will defeat URL matching for the first entry.
When I run it through the code the first URL comes out like this, and won't be found:
```python
{
    '\xef\xbb\xbfhttps://www.consumerfinance.gov/administrative-adjudication-proceedings/administrative-adjudication-docket/capital-one-bank/': 
    'Expired/Terminated/Dismissed'
}
```

wpears:
@cfarm That action just doesn't have a status in general, see https://www.consumerfinance.gov/administrative-adjudication-proceedings/administrative-adjudication-docket/american-express-travel-related-services-company/

so that's unrelated to this PR.

cfarm:
but it's in the script? 

https://github.com/cfpb/cfgov-refresh/pull/5426/files#diff-31311c4bc7cdf38b0ba44cc541dda987R17

wpears:
> but it's in the script?
> 
> https://github.com/cfpb/cfgov-refresh/pull/5426/files#diff-31311c4bc7cdf38b0ba44cc541dda987R17

Good point! Further investigation reveals there are 4 such cases, I'll update them in content manually

PR #5428, opened 2019-12-19
Mega Menu: Add variations styles and JavaScript

anselmbradford:
## Changes

- Adds styles and JavaScript for mega menu variations.

## Testing

2. Enable the `MEGA_MENU_VAR_1` flag and reload the homepage and see that the menu is the variation 1 with a caret by the menu items.
3. Enable the `MEGA_MENU_VAR_2` flag (and disable `MEGA_MENU_VAR_1`) and reload the homepage and see that the menu is the variation 2 without a caret by the menu items.

## Note
- This duplicates the MegaMenu.js code, which inflates the common.js bundle a bit. When we finalize the menu changes this will drop back down.


cfgov/unprocessed/js/organisms/Header.js
```diff
@@ -1,7 +1,23 @@
...
+import MegaMenuVar1 from '../organisms/MegaMenuVar1';
+import MegaMenuVar2 from '../organisms/MegaMenuVar2';
+
+if ( window.cfpb && window.cfpb.megaMenuVar1 ) {
+  MegaMenu = MegaMenuVar1;
+} else if ( window.cfpb && window.cfpb.megaMenuVar2 ) {
+  MegaMenu = MegaMenuVar2;
+} else {
+  MegaMenu = MegaMenuOrig;
+}
```
chosak: Do we need to import all three even if only one will ever be shown, and its selection is fixed in the page HTML?

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -274,6 +274,20 @@
 
     {% block javascript scoped %}
 
+    {# TODO: Remove when 2019/2020 Mega Menu modifications are finalized. #}
+    {% if flag_enabled('MEGA_MENU_VAR_1', request) %}
+      <script>
+        window.cfpb = window.cfpb || {};
```
chosak: Is this an existing technique (using `window.cfpb` to stash configuration variables), or a new approach?

I wonder (ping @willbarton) how hard it would be to generate a list of enabled flags, and stash that in the DOM. So we'd have another [django-flags](https://github.com/cfpb/django-flags) call like `get_enabled_flags()` and then we could do something like:

```html
<script>
  if (!window.cfpb) {
    window.cfpb = {
      flags: {{ get_enabled_flags() | json }}
    }
  }

   // Then to test a flag you could do something like:
  if (window.cfpb.flags.MEGA_MENU_VAR_1) {
    ...
  }
</script>
```

If we made this part of the base template then FEWDs could more easily generally test for flags without having to hardcode each one in the template.

One downside might be that it might be inefficient to test flag state for all flags at once.

cfgov/unprocessed/js/organisms/Header.js
anselmbradford: [Conditional imports](https://stackoverflow.com/questions/36367532/how-can-i-conditionally-import-an-es6-module#answer-46543835) are available, but it wraps the response in a Promise, which means we'd have to move all the code that follows in this file conditionally into the Promise block in some fashion. Given we'll strip these out (hopefully) soonish, I opted for the simpler importing them all. It inflates the JS a bit, but not terrible.

cfgov/jinja2/v1/_layouts/base.html
anselmbradford: > Is this an existing technique (using window.cfpb to stash configuration variables), or a new approach?

New approach. I thought I should namespace it for safety.

anselmbradford:
I've pushed up the variation 1 and variation 2 menus.

**Note: The green horizontal line below the menu gets converted to the gray 1px line on production with this change.** It's probably not worth putting in the logic to retain that as we transition to a new menu.

anselmbradford:
Also, the second variation is only closed when re-clicking the same menu item.

anselmbradford:
Var 1
Screen Shot 2019-12-31 at 4 08 05 PM

Var 2
Screen Shot 2019-12-31 at 4 07 45 PM

sonnakim:
@anselmbradford (sorry, overdue comment - this reflects what we discussed during yesterday's post-scrum). Regarding Scott's comments, I'd like to keep rolloff to close on var. 1 so we can test how it performs against click to close. For var. 2, I agree with Scott that one should be able to click anywhere outside the menu to close it. I don't feel it's crucial to fix the first three items for testing (visual jank, icon alignment, pipes vs. borders), so I'll leave it up to you whether you want to fix before merging.

anselmbradford:
> @anselmbradford (sorry, overdue comment - this reflects what we discussed during yesterday's post-scrum). Regarding Scott's comments, I'd like to keep rolloff to close on var. 1 so we can test how it performs against click to close. For var. 2, I agree with Scott that one should be able to click anywhere outside the menu to close it. I don't feel it's crucial to fix the first three items for testing (visual jank, icon alignment, pipes vs. borders), so I'll leave it up to you whether you want to fix before merging.

Cool, the menu closing is fixed.

PR #5433, opened 2020-01-02
Carousel markup and IE11 fixes

anselmbradford:
## Changes

- Fixes incorrectly closed paragraph tags.
- Adds `overflow: visible` IE11 hack to show the top border of the thumbnails.
- Converts `ul` and `li` markup into `div` and `span`s.
- Sets image `alt` attribute for mobile carousel links to be same as accompanying link text.
- Sets image requested from wagtail as `fill-48x48`, instead of `fill-50x50`.

## Testing

1. Pull down this branch and `gulp clean && gulp build`
2. Test in IE11 in Sauce labs or similar and see when the screen size is such that the thumbnails are stacked, there is a horizontal line between them. Other browsers should be unchanged. 
3. WAVE accessibility plugin should only show 1 error, instead of 5.

## Screenshots

Before:
Screen Shot 2020-01-02 at 4 39 53 PM

After:
Screen Shot 2020-01-02 at 4 38 53 PM


Scotchester:
How about we take this opportunity to change those `p`s to something more semantically correct? :)

anselmbradford:
> How about we take this opportunity to change those ps to something more semantically correct? :)

Sure, what did you have in mind?

Scotchester:
> Sure, what did you have in mind?

Probably needs to be a `span` to be a valid child of `button`, and actually, now noticing that the `li`s inside the `buttons` are also a problem. To do this all valid and stuff, should go more like this:

```
ul
  li
    button
      img (no div) (or no img and make it a background-image on the button)
      span
```

But all that seems a bit out of scope for this PR.

anselmbradford:
What about converting the list to divs in the meantime? We wouldn't need to swap nodes then.

```
div
  button
    div
      div
        img
      span
```

Scotchester:
> What about converting the list to divs in the meantime? We wouldn't need to swap nodes then.
> 
> ```
> div
>   button
>     div
>       div
>         img
>       span
> ```

You're still only allowed to have [phrasing content](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content) within a `button`, though, so those `div`s are still not allowed.

anselmbradford:
`span`s then and we may need to add `display: block` to some:

```
div
  button
    span
      span
        img
      span
```

?

Scotchester:
Yeah, seems like that would work fine, though I question whether the image needs to be wrapped in its own wrapper at all. Also, we're asking Wagtail for a 50px image but only rendering it at 48px.

anselmbradford:
> Yeah, seems like that would work fine, though I question whether the image needs to be wrapped in its own wrapper at all.

Yeah the original purpose of that was to have a wrapper for cropping purposes as we didn't know what those image thumbnails would be.

> Also, we're asking Wagtail for a 50px image but only rendering it at 48px.

Hmm not sure how that ended up like that. Any preference for 50px or 48px? I guess it would be easier to change the Wagtail request to `fill-48x48`.

Scotchester:
> Hmm not sure how that ended up like that. Any preference for 50px or 48px? I guess it would be easier to change the Wagtail request to `fill-48x48`.

Yeah, I think that'd be the most straightforward fix there, guaranteed not to impact layout.

anselmbradford:
Okay pushed up these changes:

> - Converts `ul` and `li` markup into `div` and `span`s.
> - Sets image `alt` attribute for mobile carousel links to be same as accompanying link text.
> - Sets image requested from wagtail as `fill-48x48`, instead of `fill-50x50`.

Flagging @chosak on that last one, if there are any issues with that change.

Scotchester:
> * Sets image `alt` attribute for mobile carousel links to be same as accompanying link text.

We don't want to do this. That will read the same thing twice, which is not helpful. Like the large images, the `alt` attribute here should be empty.

anselmbradford:
> We don't want to do this. That will read the same thing twice, which is not helpful. Like the large images, the alt attribute here should be empty.

Seems accurate though? It's two links to the same thing next to each other, so reading them twice is describing what's there. I was thinking fix the errors till the whole thing is clickable is preferable than leaving it.

Scotchester:
> > We don't want to do this. That will read the same thing twice, which is not helpful. Like the large images, the alt attribute here should be empty.
> 
> Seems accurate though? It's two links to the same thing next to each other, so reading them twice is describing what's there. I was thinking fix the errors till the whole thing is clickable is preferable than leaving it.

Gah, right. Sorry, I forgot that it was still two separate links. Hmm, what about adding `aria-hidden="true"` to the link around the image?

anselmbradford:
> Gah, right. Sorry, I forgot that it was still two separate links. Hmm, what about adding aria-hidden="true" to the link around the image?

Sure, pushing it up.

Scotchester:
@anselmbradford I gave you bad advice here. My goldfish memory couldn't even remember that we had the same thing come up not two months ago: https://github.com/cfpb/cfgov-refresh/pull/5382

Really should go the distance and wrap the entire thing in a link.

anselmbradford:
Oh haha oops. I still think the duplication is preferable to the error. While potentially irritating, it does describe the markup.

PR #5446, opened 2020-01-10
Move focus to header on return to top button

anselmbradford:
## Changes

- Set focus on header element when return to top button is clicked.

## Testing

2. Resize window to mobile size.
3. Scroll to bottom of the page and click "Back to top" button.
4. Press tab to see that the focus is in the header, not the footer (compare to behavior on production).

## Notes

- ~Ideally the focus should be BEFORE the skip link, so that the first click of the tab key ends up on the skip link. However, to move the focus from the footer, we need to move focus to another focusable element. The skip link is the first of these on the page. We don't want the skip link to show (I don't think?) when returning to the top as the user has not yet hit the tab key, so we set it to the skip link and then blur the `activeElement`. The consequence of this is that the next hit of the tab key will not be on the skip link, as it'll have already moved past it.~

### DIFF
--- a/cfgov/unprocessed/js/modules/footer-button.js
+++ b/cfgov/unprocessed/js/modules/footer-button.js
@@ -61,15 +61,8 @@ function _scrollToTop() {
  *  Move focus to the top of the page.
  */
 function _setFocus() {
-  /* To move the focus from the footer we need to move it to another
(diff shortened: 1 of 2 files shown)


anselmbradford:
Updated with a better method (thanks @Scotchester !)

PR #5456, opened 2020-01-14
Mark up Key Terms lists with DefinedTerm schema

Scotchester:
Using [DefinedTerm](http://schema.org/DefinedTerm) and [DefinedTermSet](http://schema.org/DefinedTermSet) schema markup should improve the performance and display of our content in search results.

Addresses https://GHE/CFGOV/platform/issues/3644

## Changes

- Updates template that outputs Key Terms on portal pages to use `dl` markup with `DefinedTerm` schema attributes

## Testing

1. Visit http://localhost:8000/consumer-tools/auto-loans/answers/key-terms/ and inspect the markup for a term
2. Compare with https://www.consumerfinance.gov/consumer-tools/auto-loans/answers/key-terms/ – there should be no visible difference

## Notes

- There's a small text formatting bug in IE that will be fixed when cfpb/design-system#422 is merged and pulled into cfgov-refresh.

## Todos

- ~~Get third party confirmation that these are the correct schema bits to use~~

## Checklist

### Browsers

- [x] Chrome on desktop
- [x] Firefox
- [x] Safari on macOS
- [x] Edge
- [x] Internet Explorer 11
- [x] Safari on iOS
- [x] Chrome on Android

### DIFF
--- a/cfgov/jinja2/v1/ask-cfpb/see-all.html
+++ b/cfgov/jinja2/v1/ask-cfpb/see-all.html
@@ -11,18 +11,23 @@
       {{ page.title }}
     </h1>
     {% if page.glossary_terms %}
-        <div class="terms">
+        <dl class="terms" itemscope itemtype="http://schema.org/DefinedTermSet">
         {% for term in page.glossary_terms %}
             <div class="term {{'' if loop.index == 1 else 'block block__border-top block__padded-top'}}"
-                 id="{{ term.anchor(page.language) }}">
-            <h2>{{ term.name(page.language) }}</h2>
-                {{ term.definition(page.language) | richtext }}
-                {% if term.answer_page_en %}
-                    <p class="term-link"><a href="{{ term.answer_page_url(page.language) }}"> {{ _('Read more') }} </a></p>
-                {% endif %}
+                 id="{{ term.anchor(page.language) }}"
+                 itemscope
+                 itemtype="http://schema.org/DefinedTerm">
+                <link itemprop="url" href="{{ page.full_url }}#{{ term.anchor(page.language) }}">
+                <dt class="term_name" itemprop="name">{{ term.name(page.language) }}</dt>
+                <dd class="term_definition" itemprop="description">
+                    {{ term.definition(page.language) | richtext }}
+                    {% if term.answer_page_en -%}
+                        <a class="term_link" href="{{ term.answer_page_url(page.language) }}">{{ _('Read more') }}</a>
+                    {%- endif %}
+                </dd>
             </div>
         {% endfor %}
-        </div>
+        </dl>
     {% else %}
         {% if not page.portal_category %}
         <p class="lead-paragraph u-mt0">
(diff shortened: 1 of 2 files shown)


Scotchester:
Taking this PR out of Draft status. Despite our third party's current ambivalence, as discussed with @kelleyholden last week, this still seems like a potentially useful thing to do, with no downsides that we're aware of. @niqjohnson, could you review?

cfgov/jinja2/v1/ask-cfpb/see-all.html
```diff
@@ -11,18 +11,23 @@ <h1>
...
-                 id="{{ term.anchor(page.language) }}">
-            <h2>{{ term.name(page.language) }}</h2>
-                {{ term.definition(page.language) | richtext }}
-                {% if term.answer_page_en %}
-                    <p class="term-link"><a href="{{ term.answer_page_url(page.language) }}"> {{ _('Read more') }} </a></p>
-                {% endif %}
+                 id="{{ term.anchor(page.language) }}"
+                 itemscope
+                 itemtype="http://schema.org/DefinedTerm">
+                <link itemprop="url" href="{{ page.full_url }}#{{ term.anchor(page.language) }}">
```
niqjohnson: I ran this through https://validator.w3.org/, and the validator shows an error with this `link` element here. If I'm reading the error right, I *think* the problem is that only `dt` and `dd` elements are allowed as children of a `div` in a `dl`. But I'm not sure that there's any other way to get that `link` in here, and that seems like a pretty harmless validation error, so I'm fine leaving it as is.

cfgov/jinja2/v1/ask-cfpb/see-all.html
```diff
@@ -11,18 +11,23 @@ <h1>
...
+                <dt class="term_name" itemprop="name">{{ term.name(page.language) }}</dt>
+                <dd class="term_definition" itemprop="description">
+                    {{ term.definition(page.language) | richtext }}
+                    {% if term.answer_page_en -%}
+                        <a class="term_link" href="{{ term.answer_page_url(page.language) }}">{{ _('Read more') }}</a>
```
niqjohnson: I don't think we need to worry about this for this first pass, but I wonder how we should treat this link in the DefinedTerm schema. It doesn't quite feel like part of the `description` to me, but I'm not sure what else to do with it. I'm fine leaving it for now and refactoring later if needed.

cfgov/jinja2/v1/ask-cfpb/see-all.html
Scotchester: Good catch! I tested moving it inside the `<dt>` element and that eliminated the validation errors. You might wonder if having the `url` nested inside the `name` would be a problem, but according to Google's [Structured Data Testing Tool](https://search.google.com/structured-data/testing-tool), the properties are still understood correctly. Fixed in dd68f18eec29e9291824cbbc1c222154b50c194a.

cfgov/jinja2/v1/ask-cfpb/see-all.html
Scotchester: Hmm... I see what you mean, but I don't think that there is a good alternative that's immediately obvious.

niqjohnson: Agreed. I'm good leaving this as-is for now.

PR #5457, opened 2020-01-14
Mega Menu: Fix top padding. Fix brochure icon. Hide menu column header.

anselmbradford:
##  Changes

- Fix top padding above the overview link, so that the hover has correct height.
- Change order brochure link icon to be a document instead of a complaint.
- Hide redundant menu item group header.

## Testing

2. Enable `MEGA_MENU_VAR_1` flag.
3. Visit homepage and resize desktop to view extra large and regular desktop sizes and see that hover for overview link is equal height on the top and bottom. Also see that second column in menu item groups has a hidden heading. Also see that the Order Brochures icon has changed to a document.

## Screenshots

Screen Shot 2020-01-14 at 3 27 15 PM
Screen Shot 2020-01-14 at 3 27 22 PM


anselmbradford:
@chosak Thanks for catching that. Pushed up a fix.

Screen Shot 2020-01-14 at 3 48 27 PM

PR #5468, opened 2020-01-17
Redirect legacy Ask URLs with uppercase categories

higs4281:
Some of our broken internal links originate as knowledgebase-era
querystring-based category searches, which we routed to category pages
in 2016. But Some links request uppercase category names with spaces, and
the redirects break down.

This change slugifies the requested category so that the redirects can complete.

## Local testing

Two legacy URLs that currently return 404:
- [http://localhost:8000/askcfpb/search/?selected_facets=category_exact:Prepaid Cards](http://localhost:8000/askcfpb/search/?selected_facets=category_exact:Prepaid%20Cards)
- [http://localhost:8000/askcfpb/search/?selected_facets=category_exact:Student Loans](http://localhost:8000/askcfpb/search/?selected_facets=category_exact:Student%20Loans)

After pulling in this branch, the URLs above should relevant to portal pages under
**/consumer-tools/**.

## Note
You may need to clear your browser cache after switching to the PR branch to get your browser to follow the fixed redirect.

### DIFF
--- a/cfgov/ask_cfpb/tests/test_views.py
+++ b/cfgov/ask_cfpb/tests/test_views.py
@@ -426,6 +426,17 @@ class RedirectAskSearchTestCase(TestCase):
             result.get('location'),
             '/ask-cfpb/search/')
 
+    def test_redirect_search_uppercase_facet(self):
+        """Handle odd requests with uppercase, spaced category names."""
+        category_querystring = 'selected_facets=category_exact:Prepaid Cards'
+        request = HttpRequest()
+        request.GET = QueryDict(category_querystring)
+        result = redirect_ask_search(request)
+        self.assertEqual(
+            result.get('location'),
+            '/ask-cfpb/category-prepaid-cards/'
+        )
+
     def test_redirect_search_no_query(self):
         request = HttpRequest()
         request.GET['q'] = ' '
--- a/cfgov/ask_cfpb/views.py
+++ b/cfgov/ask_cfpb/views.py
@@ -214,7 +214,8 @@ def redirect_ask_search(request, language='en'):
             if category_facet in facet:
                 category = facet.replace(category_facet, '')
                 if category:
-                    return redirect_to_category(category, language)
+                    slug = slugify(category)  # handle uppercase and spaces
+                    return redirect_to_category(slug, language)
 
         for facet in facets:
             if audience_facet in facet:
--- a/cfgov/cfgov/urls.py
+++ b/cfgov/cfgov/urls.py
@@ -437,91 +437,91 @@ urlpatterns = [
 
 # Ask CFPB category and subcategory redirects
 category_redirects = [
-    url(r'^ask-cfpb/category-auto-loans/(.*)$',
+    url(r'^(?i)ask-cfpb/category-auto-loans/(.*)$',
         RedirectView.as_view(
             url='/consumer-tools/auto-loans/',
             permanent=True)),
-    url(r'^ask-cfpb/category-bank-accounts-and-services/(.*)$',
+    url(r'^(?i)ask-cfpb/category-bank-accounts-and-services/(.*)$',
         RedirectView.as_view(
             url='/consumer-tools/bank-accounts/',
             permanent=True)),
-    url(r'^ask-cfpb/category-credit-cards/(.*)$',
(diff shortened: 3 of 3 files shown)


Scotchester:
This works great for the oldest format with the `selected_facets` and such, but I think we also need redirects from the `/ask-cfpb/category-Debt Collection/` paths, in case those URLs are cached anywhere. Could you add that to this PR, @higs4281?

higs4281:
> This works great for the oldest format with the `selected_facets` and such, but I think we also need redirects from the `/ask-cfpb/category-Debt Collection/` paths, in case those URLs are cached anywhere. Could you add that to this PR, @higs4281?

I added case insensitivity, but I think trying to capture spaces may be a point of diminishing returns. I haven't seen any 404 lists with such URLs. I've asked analytics whether it's an issue.

I think the free-form, query-string-based URLs invited variations, but the 2016 category URLs were simple and canonical.

higs4281:
Update: Jabari did a broader 404 link search and found only facet-based category URLs going awry.

Scotchester:
Hmm... that's odd that Jabari didn't see any except facet-based URLs, because in the January 13 report, it's indicating that some category URLs are indeed a problem:

Snippet of a spreadsheet with three columns—Page URL, Broken Link URL, and HTTP Code—and 6 rows showing that Broken Link URLs on Page URLs returned HTTP Codes of 404

So, for example, if some server somewhere cached the redirect to https://www.consumerfinance.gov/ask-cfpb/category-Student Loans/, then that would still result in a 404 unless we add a rule that can redirect from that formation to `/ask-cfpb/category-student-loans/` (or straight to `/consumer-tools/student-loans/`).

higs4281:
Maybe I'm misreading the spreadsheet, but all the source pages above begin with fact-based URLs, which will resolve correctly with this PR.

Currently they result in an errant 2016-style URL, but that won't be produced anymore. 
After this PR, where would a person find the bad 2016-style links?

Scotchester:
> Maybe I'm misreading the spreadsheet, but all the source pages above begin with facet-based URLs, which will resolve correctly with this PR.
> 
> Currently they result in an errant 2016-style URL, but that won't be produced anymore.
> After this PR, where would a person find the bad 2016-style links?

Try loading one of those facet-based URLs on your branch: 
 https://www.consumerfinance.gov/askcfpb/search/?selected_facets=category_exact:Debt Collection

Scotchester:
Sorry, I'm an idiot. I tested the live URL instead of the URL on my localhost.

Scotchester:
> Currently they result in an errant 2016-style URL, but that won't be produced anymore.
After this PR, where would a person find the bad 2016-style links?

But this is still something I'm concerned about. If someone has a cached redirect (which I gather is a thing that can happen with permanent redirects), they could still end up on something like https://www.consumerfinance.gov/ask-cfpb/category-Student Loans/.

higs4281:
I think you're right that a browser, revisiting a malformed facet-based URL with a permanent redirect, could return the cached 404 response. But in that case any code we write is ignored until the cache expires or is broken.

If/when the browser cache goes away, the request will get passed to our new code, and the visitor's browser should never encounter a possibly cached bad intermediary URL with spaces.

All that said, browser cache behavior often baffles me, and I agree we should do all we can to get visitors to a helpful page.

It bugs me that, because of how Ask evolved, we send 2014-era category requests to 2016-era category requests in order to get to 2019-era portal pages. Maybe we could save a hop and solve your concern at the same time by redirecting facet-based requests straight to portal pages, and attempt to catch as many query strings as we can.

higs4281:
"Pages" and "links" are misrepresented on the spreadsheet above. The source pages for our six bad Ask links are blog pages that are not listed.

The "Page URL" column are actually the bad links on the blog pages and the "broken link URL" are not links but the redirects generated by our code.

In this case, the sure-fire fix is to go to the six original pages and fix the links. Then browser cache is taken out of the equation, because a visitor's browser will never see the bad links again. All 6 pages are now fixed. (Page IDs 2010, 2779, 2829, 2851, 2786, 4721.)

higs4281:
In the interests of fixing current 404s, I'm merging this and leaving hop-elimination to another PR.

PR #5471, opened 2020-01-22
New mega menu backend implementation

chosak:
This commit adds a new backend implementation of the site mega menu, asa new Django app named mega_menu. This implementation can be used instead of the existing one by enabling the `MEGA_MENU_BACKEND_V2` feature flag.

This new implementation supports both the current mega menu design as well as the two variations currently being considered and tested. These two variations can be tested by enabling the `MEGA_MENU_VAR_1` or `MEGA_MENU_VAR_2` feature flags, respectively.

The major differences between the new implementation and the old one include:

1. Uses far fewer database queries. Instead of making at least one database query for each menu link pointing at a Wagtail Page, all page information is retrieved in a single query. This optimization allows us to remove a dedicated Django cache that currently exists to mitigate these queries.
2. An independent menu is now maintained for each supported language, meaning that the Spanish menu is now editable in Wagtail.
3. The data passed to the frontend is now decoupled from the database schema on the backend, hopefully allowing for greater flexibility and experimentation in future.

The new menu content can be migrated from existing menu content by running this new Django management command:

```
cfgov/manage.py migrate_menu_content
```

The new menu content can then be edited in the Wagtail admin under a new "Mega menu (v2)" section. This area contains a "Preview" button that can be used to easily test the menu without having to navigate to a page. Each menu has its own preview page, for example:

- http://localhost:8000/admin/mega_menu/menu/preview/en/
- http://localhost:8000/admin/mega_menu/menu/preview/es/

Currently the "Preview" button only shows the saved version of the menu; in future it would be nice to have this work like Wagtail page previews, and allow for previewing of unsaved edits.

## Screenshots

New Wagtail admin area. Note "Mega menu (v2)" in sidebar, and list of menus to edit, including edit, delete, and preview buttons:

image

Editing a menu. Note the preview button at the bottom:

image

Preview view, with placeholder content:

image

## Testing

These testing instructions assume you are using a production dump:

1. Check out this branch, and apply its migrations to your local database using `cfgov/manage.py migrate`.
2. Migrate mega menu content from the existing tables into the new tables by running `cfgov/manage.py migrate_menu_content`.
3. In the Wagtail admin, navigate to http://localhost:8000/admin/mega_menu/menu/ and confirm that you have data for both the English and Spanish menus.
4. Try clicking the "Preview" button for both of these -- note that this still uses the existing backend and design because you haven't enabled anything new with feature flags.

Restart your local dev server with the Django Debug Toolbar enabled to observe the improvement in database queries. For example, use `ENABLE_DEBUG_TOOLBAR=1 ./runserver.sh`. Navigate back to the English preview view at http://localhost:8000/admin/mega_menu/menu/preview/en/, and, in the DJDT sidebar, you should see something like 119 queries.

5. Now, enable the new backend by going to http://localhost:8000/admin/flags/create/ and creating a flag named `MEGA_MENU_BACKEND_V2`. Enable it for all requests at http://localhost:8000/admin/flags/MEGA_MENU_BACKEND_V2/.
6. Go back to http://localhost:8000/admin/mega_menu/menu/preview/en/, and you should instead see something like 10 queries. The visual appearance and behavior of the menu should be the same.
7. Also visit the Spanish preview page at http://localhost:8000/admin/mega_menu/menu/preview/en/ and confirm that the menu looks and acts the same.
8. You can also visit the site homepage at http://localhost:8000/ with and without the flag to compare the number of database queries (151 vs 46).

You can further test both the existing and new backends with the two proposed mega menu designs by setting the `MEGA_MENU_VAR_1` and `MEGA_MENU_VAR_2` flags. In both cases, the design and behavior should be the same.

CFPB devs can also play with a version of this branch on our internal DEV4 server.

## Todos

This is a big change but there are several TODOs that I have kept out of this PR to keep its scope as small as possible:

- Update the documentation at http://cfpb.github.io/cfgov-refresh/mega-menu/ to reflect the new editing experience.
- Add better help text and form labels in the Wagtail menu editor.
- Create a nicer/more compact link chooser to reduce the duplication of (page, text, link) in the Wagtail menu editor (this might have to wait until we are on Wagtail 2+ to be able to leverage packages like [wagtail-generic-chooser](https://github.com/wagtail/wagtail-generic-chooser)).
- Move existing mega menu templates from jinja2/v1/_includes to the mega_menu app (if FEWDs agree this makes sense).
- Make the "Preview" button on the menu edit screen preview unsaved edits, the way the preview button on the Wagtail page editor works (requires some JS code).
- Add documentation of the expected menu content format to the frontend template.
- Remove legacy menu code and related infrastructure (the Django `default_fragment_cache` cache, which will no longer be needed after this change).

## Checklist


cfgov/cfgov/settings/base.py
```diff
@@ -96,6 +96,7 @@
     'hmda',
     'youth_employment',
     'diversity_inclusion',
+    'mega_menu.apps.MegaMenuConfig',
```
chosak: Somehow I never learned that the preferred format for entries in [`settings.INSTALLED_APPS`](https://docs.djangoproject.com/en/1.11/ref/settings/#std:setting-INSTALLED_APPS) is the dotted Python path to an application configuration class, not the a package containing an application.

cfgov/cfgov/settings/base.py
```diff
@@ -802,7 +804,12 @@
 # which we don't want this logic to be run.
 PARSE_LINKS_EXCLUSION_LIST = [
     # Wagtail admin pages, except preview and draft views
-    r'^/admin/(?!pages/\d+/(edit/preview|view_draft)/)',
+    (
+        r'^/admin/(?!'
+        r'pages/\d+/(edit/preview|view_draft)/|'
+        r'mega_menu/menu/preview/\w+/'
```
chosak: This ensures that the menu preview screen gets our link parsing logic applied.

cfgov/jinja2/v1/_includes/molecules/featured-menu-content.html
```diff
@@ -31,8 +31,7 @@
     {% set img_src = featured_image.url if featured_image.url else '' %}
 {% else %}
     {% set featured_image = value.image %}
-    {% set img_src = static(featured_image.url) if featured_image.url else '' %}
-
+    {% set img_src = featured_image.url if featured_image.url else '' %}
```
chosak: This logic was unused, and was a leftover of the days before the mega menu was in Wagtail. I made this change to support the existing menu design with the new backend; when the menu content contains an image URL, we shouldn't assume that it'll be in Django static files. It never is because featured menu images are Wagtail image uploads, not static files.

cfgov/mega_menu/frontend_conversion.py
```diff
@@ -0,0 +1,148 @@
+class FrontendConverter:
+    def __init__(self, menu, request=None):
```
chosak: This code may seem to be generating arbitrarily named and structured Python dicts for the frontend, and indeed it is, mainly to ensure compatibility with both the legacy and proposed frontend code. In future we can hopefully simplify and rename quite a bit of this (for example avoiding the use of `external_link` when passing menu URLs to the frontend). I wanted to avoid having to alter the existing menu frontend code as part of this PR.

cfgov/mega_menu/frontend_conversion.py
```diff
@@ -0,0 +1,148 @@
...
+                url = link['external_link']
+                link_text = link['link_text']
+            else:
+                text = ''
+                url = other_link['url']
+                link_text = other_link['text']
+
+            html = '<p>'
+
+            if text:
```
chosak: This logic is copied from [how the existing menu does this](https://github.com/cfpb/cfgov-refresh/blob/b97c47871d3658e49eba284690de0b47df34a90c/cfgov/v1/models/menu_item.py#L111-L129), and will be unnecessary if/when we move to a menu with simpler links that don't require this markup.

cfgov/jinja2/v1/_includes/organisms/mega-menu-var-2.html
```diff
@@ -193,6 +223,7 @@
                 {% endif %}
             </div>
             {% endif %}
+            {% endif %}
```
anselmbradford: Is this double `endif` correct? If so, should the inside block be indented one level?

cfgov/jinja2/v1/_includes/organisms/mega-menu-var-2.html
```diff
@@ -248,6 +296,7 @@
                     </ul>
                 </div>
                 {% endif %}
+                {% endif %}
```
anselmbradford: Same double `endif` question as above.

cfgov/mega_menu/templates/modeladmin/mega_menu/menu/edit.html
```diff
@@ -0,0 +1,22 @@
...
+        <ul>
+            <li class="actions">
+                {% block form_actions %}
+                    {{ block.super }}
+                {% endblock %}
+            </li>
+            <li class="preview">
+                <a href="{{ view.preview_url }}"
+                    class="button icon icon-view"
+                    target="_blank" rel="noopener noreferrer"
```
anselmbradford: Maybe move `rel` to its own line like the other attributes. Also, maybe move the `>` to the last line of content, not on its own, but I don't really care. Also (trivial), we line up the attributes indents elsewhere (so `href` and `class` would be left-aligned), but there's an argument to be made for the indent you have too.

cfgov/jinja2/v1/_includes/organisms/mega-menu-var-2.html
chosak: It is. The first `endif` is from [this block](https://github.com/cfpb/cfgov-refresh/blob/3f564b25bd5228122304ee931c70fb84ecfada23/cfgov/jinja2/v1/_includes/organisms/mega-menu-var-2.html#L211-L225), one of the hardcoded ones you added. I could and maybe should have indented but didn't want to add so many more diffs to these files, especially since they're still just proposed variations.

cfgov/jinja2/v1/_includes/organisms/mega-menu-var-2.html
chosak: See answer above.

chosak:
> One question: should clicking the menu preview button open in a new tab? Currently it doesn't and you have to use the browser back button to get back to the admin.

Thanks, that's a good catch. As mentioned above under TODOs, there's more I'd like to do around the preview page after this PR, but this would be a simple fix I can include here.

chosak:
@anselmbradford it turns out that, while it is possible to add `target="blank"` to custom ModelAdmin buttons, it isn't yet possible to add the best practice `rel="noopener noreferrer`. I've opened a new Wagtail PR at https://github.com/wagtail/wagtail/pull/5790 to add support for this.

cfgov/cfgov/settings/base.py
higs4281: You are not alone here.

cfgov/mega_menu/frontend_conversion.py
```diff
@@ -0,0 +1,148 @@
...
+        if featured_links:
+            featured_link = featured_links[0]
+
+            value = {'link': self.make_link(featured_link)}
+
+            value.update({
+                'body': featured_link['body'],
+                'image': {
+                    'url': (
+                        featured_link['image'].get_rendition('original').url
```
schbetsy: When I try to add a new menu item, I get an error here. `image` is not a required field, but this line assumes it is present.

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
```diff
@@ -255,3 +255,9 @@ <h2 class="{{ base_class ~ '_heading' }}">
 </nav>
 
 {% endmacro %}
+
+
+{# Render the menu if this entire template is rendered with menu_items. #}
```
schbetsy: I don't quite understand this. Does this if statement prevent the menu template from rendering in the case where there is no mega menu created in Wagtail? 

I tried deleting the entire (Spanish) menu in wagtail to see this case, and there was an error instead.

cfgov/mega_menu/jinja2tags.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def select_menu_for_context(context):
+    language = context.get('language')
+
+    if not language:
+        request = context['request']
+        language = get_language_from_request(request, check_path=True)
+
+    return Menu.objects.get(language=language)
```
schbetsy: This line generates an error for all of the other language pages linked in the top bar of the site (Chinese, Vietnamese, Korean, etc): `mega_menu.models.DoesNotExist: Menu matching query does not exist.`

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
chosak: Good feedback; I realize this comment might not be that clear. Let me try to explain what I mean by this, and maybe together we can come up with better wording.

In the existing mega menu implementation, the template [mega-menu.html](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/mega-menu.html) (and its two proposed variations, [mega-menu-var-1.html](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/mega-menu-var-1.html) and [mega-menu-var-2.html](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/mega-menu-var-2.html)) can't be rendered by Django on its own, that is, using the standard process of loading the template and then calling some kind of `template.render(context)` on it. This is because, in master, the template only contains Jinja macros; it doesn't contain any content at the top level. Currently the top-level entry point is [this](https://github.com/cfpb/cfgov-refresh/blob/3a4a2b9264b2c386b0a2aa0b7ae06138f1f8296f/cfgov/jinja2/v1/_includes/organisms/mega-menu.html#L232) `mega_menu()` macro, which has to be explicitly invoked.

Currently we do that in the header template [header.html](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/header.html). First we use Jinja's [`import`](https://jinja.palletsprojects.com/en/2.11.x/templates/#import) to import that `mega_menu` macro from the template file [here](https://github.com/cfpb/cfgov-refresh/blob/3a4a2b9264b2c386b0a2aa0b7ae06138f1f8296f/cfgov/jinja2/v1/_includes/organisms/header.html#L106), and then we invoke that macro [here](https://github.com/cfpb/cfgov-refresh/blob/3a4a2b9264b2c386b0a2aa0b7ae06138f1f8296f/cfgov/jinja2/v1/_includes/organisms/header.html#L114), passing it the data to render.

What I'd like to do instead, as part of this PR, is be able to "just" render the mega menu template itself, from Python, without having to invoke a specific macro in it. In this PR I expose a new `mega_menu` function to Jinja that renders the whole menu, and invoke it [here](https://github.com/cfpb/cfgov-refresh/blob/1dd2688815ee69f6565ae0dcf325946352af9091/cfgov/jinja2/v1/_includes/organisms/header.html#L94) (this line is now calling [this](https://github.com/cfpb/cfgov-refresh/blob/1dd2688815ee69f6565ae0dcf325946352af9091/cfgov/mega_menu/jinja2tags.py#L51) Jinja function, not a macro from a template). And I want my function to be able to "just" call `template.render(context)` to render the menu, on this line, [here](https://github.com/cfpb/cfgov-refresh/blob/1dd2688815ee69f6565ae0dcf325946352af9091/cfgov/mega_menu/jinja2tags.py#L43). But in order to do that, I have to change the template as I've done here, with this logic:

```jinja
{# Render the menu if this entire template is rendered with menu_items. #}
{% if menu_items %}
    {{ mega_menu( menu_items, language | default( null ) ) }}
{% endif %}
```

What this does is, if the template is `render`ed with a context containing a `menu_items` variable, it will call that top-level macro, to invoke it.

Writing all this out, I fear it seems very complex, but the main idea was to simplify the point where the HTML frontend calls the backend, and let the single new `mega_menu` call in Jinja do all the work. Does that make any sense?

schbetsy: That makes some sense. Are you planning on simplifying the templates to remove the `if` conditions after we migrate to this new mega menu back-end? (I'm assuming there will be a lot of cleanup involved after the switch)

chosak: You're right that there will be a lot of cleanup, but honestly I hadn't thought that much about the template. One reasonable way to do it might be to remove the `if` as you suggest and then add some documentation to the top of the template about how to render it -- that you need to provide `menu_items` and, more importantly, what structure those items need to be. That structure can also hopefully be simplified once we move to the new backend.

cfgov/mega_menu/jinja2tags.py
chosak: Good catch, thanks. I was being too clever with this. I didn't properly handle pages in other languages besides English and Spanish. I've pushed a proposed change in ad0173161; to test, visit e.g http://localhost:8000/language/vi/.

cfgov/mega_menu/frontend_conversion.py
chosak: Thanks, I've made featured link body and image required in 98cf65bb4.

These are required in the current menu design, but in both proposed menu design variations, featured links will be simplified to just text and URL. When I implemented this I originally wasn't planning on supporting the current design, so this is something that I missed.

cfgov/jinja2/v1/_includes/organisms/mega-menu-var-1.html
```diff
@@ -136,6 +136,29 @@
             <div class="{{ _classes( nav_depth, '-lists' ) }}">
                 {{ _nav_list( nav_depth, nav_item.nav_groups ) }}
 
+                {% if flag_enabled( 'MEGA_MENU_BACKEND_V2' ) and nav_item.featured_links %}
```
virginiacc: It looks like `other_links` will only be displayed if there are also `featured_links` -- I'm not sure if that's an intentional part of the design?

cfgov/mega_menu/jinja2tags.py
```diff
@@ -0,0 +1,57 @@
...
+    language = context.get('language')
+
+    if not language:
+        request = context['request']
+        language = get_language_from_request(request, check_path=True)
+
+    try:
+        return Menu.objects.get(language=language[:2])
+    except Menu.DoesNotExist:
+        return Menu.objects.get(language=settings.LANGUAGE_CODE[:2])
```
virginiacc: It looks like this would error if there's no English menu, although I'm not sure that's a use case we need to support?

chosak: :+1: I figured this was a reasonable thing to assume, that'd there always be at least one menu defined corresponding to the default language. I've added a new test in 874b2058d that verifies that the code raises `Menu.DoesNotExist` if there isn't a default menu.

cfgov/jinja2/v1/_includes/organisms/mega-menu-var-1.html
chosak: Thanks, nice catch. Fixed in 634a09feb.

.isort.cfg
```diff
@@ -18,6 +18,7 @@ known_first_party=
     data_research,
     jobmanager,
     legacy,
+    mega_menu,
     prepaid_agreements,
     regulations3k,
     search,
```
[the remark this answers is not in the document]
chosak: Yeah, every time we add a new app I remember this file and wonder if there's a better way to do this, like could we automatically include any folder names under cfgov/? Pinging @willbarton as our resident isort export.

chosak:
> Just wondering: Will the featured-content image chooser go away at some point, since it's not being used in the new configuration?

Thanks for the review @higs42819 (and everyone else). Yes, if and when we move to one of the new designs that doesn't have the images, we can remove the code that backs it. For now, until that decision is made and implemented, I wanted to retain backwards compatibility, so left it in.

PR #5484, opened 2020-02-03
Mega menu variation 1: hide caret if no items

chosak:
Currently proposed mega menu variation 1 always shows the caret even if the menu has no child items, as would happen with the Spanish menu.

This change hides the up/down carets if no children exist.

## Testing

This can't be tested on the Spanish menu without using #5471, but can be tested in English by deleting all contents from one of the submenus.

First, enable the proposed menu variation 1:

1. Go to http://localhost:8000/admin/flags/create/ and create a feature flag named `MEGA_MENU_VAR_1`.
1. Enable it at http://localhost:8000/admin/flags/MEGA_MENU_VAR_1/ by clicking "ENABLE MEGA_MENU_VAR_1 FOR ALL REQUESTS".
1. Visiting a page on the site like http://localhost:8000/ should show the proposed variation 1, including down and up carets when menus are opened.

Now, delete all content from one of the menus to see how this change looks:

1. Edit the "Consumer Tools" menu at http://localhost:8000/admin/v1/menuitem/edit/2/ and delete all menu items by clicking the trash can icon to the right of COLUMN 1, COLUMN 2, COLUMN 3, COLUMN 4, and NAV FOOTER:
   image
1. Now revisit a page and notice that the Consumer Tools item no longer shows a caret, but does function as expected:
   menu-no-items-no-caret

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/mega-menu-var-1.html
+++ b/cfgov/jinja2/v1/_includes/organisms/mega-menu-var-1.html
@@ -257,8 +257,10 @@
              if has_children else ''
            }}>
               {{ text }}
+              {% if has_children %}
               <span class="{{- _classes( nav_depth, '-link-icon-closed' ) -}}">{{ svg_icon( 'down' ) }}</span>
               <span class="{{- _classes( nav_depth, '-link-icon-open' ) -}}">{{ svg_icon( 'up' ) }}</span>
+              {% endif %}
         </a>
         {% if has_children %}
             {{ _nav_level( nav_depth | int + 1, nav_item, url, text ) }}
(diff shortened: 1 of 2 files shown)


anselmbradford:
We'll also need to adjust the styles so that the space allocated for the caret is removed when the caret is not present. See how on "Consumer Tools" the gap on the right is wider than the gap on the left. Is this something we want to adjust in this PR?

chosak:
@anselmbradford ah, right. Sure -- is it a change in [here](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/css/organisms/mega-menu-var-1.less) somewhere? You should have permission to push to this branch if you want/have time to take a look.

PR #5475, opened 2020-01-24
Enforcement actions metadata

cfarm:
in progress new struct block for Enforcement Actions detail pages metadata that appears in sidebar

right now, this is also in a new custom page type. which we may or may not keep, to avoid cluttering all document detail pages with this new metadata field, and to use for page-level metadata

Screen Shot 2020-01-23 at 7 05 58 PM

## Additions

-

## Removals

-

## Changes

-

## Testing

1.

## Notes

-

## Todos

-

## Checklist

- [ ] PR has an informative and human-readable title
- [ ] Changes are limited to a single goal (no scope creep)
- [ ] Code can be automatically merged (no conflicts)
- [ ] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [ ] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated
- [ ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Internet Explorer 9, 10, and 11
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS
- [ ] Flexible from small to large screens
- [ ] No linting errors or warnings
- [ ] JavaScript tests are passing

PR #5500, opened 2020-02-10
Enforcement action page

wpears:
@cfarm 
Alternative implementation of https://github.com/cfpb/cfgov-refresh/pull/5475.

This creates a new page type EnforcementActionPage, which stores an enforcement action's metadata at the page level. 
Screen Shot 2020-02-13 at 12 03 39 PM

Categories & topics are always shown, so I didn't keep the "show categories" check since I figured it was crufty (however, maybe it's useful as a reminder to set in in the configuration section?)

Also, I removed the `date_filed` section because that should also be set in the configuration section. (we could also have a "show date filed" reminder here if we do for categories and topics).

Other pieces of the sidebar, like "related links" are still set in the sidebar tab


cfarm:
@chosak When reviewing this with Wyatt, we had a question about whether putting the fields that are in the new Metadata tab in Configuration is 1) technically acceptable and 2) more usable for editors. Can you let us know what you think of the feasibility or advisability, code-wise?

cfarm:
> Let's make the required fields required in the UI so it's obvious to the editor that the template requires these fields 💃 👍 🤔 🆗

@chosak what's the deal with the required fields and nullification business. Wyatt wants to know. pls help us we're but 2 half-time front ends

cfarm:
@schbetsy @willbarton y'all are our helpers now! any thoughts on the questions I tagged Andy on above?

wpears:
Seems like plenty of other fields in our models are nullable, so I'm assuming it's fine to make these nullable as well, change reflected in latest commit.

Something that I think would be the best possible UX would be if categories, tags(listed as "topics" in the template), and date_filed were settable from the metadata pane but whose values were bound to the same underlying variables in the configuration pane. Not really sure what this would look like in code, and I wouldn't want to change the underlying AbstractFilterPage without good reason, so maybe the current regime (unique metadata fields set in the metadata tab and categories, tags, date_filed set in Configuration) is acceptable.

chosak:
>  putting the fields that are in the new Metadata tab in Configuration is 1) technically acceptable and 2) more usable for editors

Placement of fields into tabs is arbitrary from a technical POV; any field can go anywhere (except maybe the title field, which comes from the base Page class). In terms of editor usability, I think it'd be better UX to have a separate tab instead of putting these into the Configuration tab. See also the "Wagtail fundamentals" page in Hubcap, which says:

> To make it easier for users to edit specific parts of a page, the Wagtail page editing interface provides a tabbed layout that groups similar page content and settings together. For consumerfinance.gov, most pages are configured with three tabs following this general setup:

> - General content: the main content area of the page
> -Sidebar / Footer / "Sidefoot": the secondary content area of the page, appearing as a sidebar or page footer depending on the page layout and browser width
> -Configuration: settings controlling page behavior and other information that does not visually appear on the page
>
> Some page types use an alternate set of tabs to organize content; for example, EventPages have separate tabs about an event's venue and agenda. In general, though, the last tab is always used for page configuration.

In terms of nullability,

> Seems like plenty of other fields in our models are nullable, so I'm assuming it's fine to make these nullable as well

If a field is required, it should be `null=False`, and, if a text field, also `blank=False`. Were you running into a specific issue with doing this to make fields required?

wpears:
>If a field is required, it should be null=False, and, if a text field, also blank=False. Were you running into a specific issue with doing this to make fields required?

I believe I was getting errors around providing a default value with `blank=False` and `null=False`. AFAICT, the `blank=False` bit is the only part necessary to make a `CharField` required, but I had to set `null=True` in order to do not set default values for these required fields.

It's possible that this is related to some intermediate database state where I've added data without required fields and so the migration is choking on that once I try to switch it to required. Does this sound likely?

chosak:
> It's possible that this is related to some intermediate database state where I've added data without required fields and so the migration is choking on that once I try to switch it to required. Does this sound likely?

Yes, this happens when you first have a model state with an optional field and then make it non-optional. Since these models haven't been merged/deployed yet, a good strategy is to delete migrations from previous commits and recreate a new one.

schbetsy:
Superseded by #5523

PR #5510, opened 2020-02-19
Re-adds gulp-bless to split stylesheets for IE

contolini:
`gulp-bless` was removed due to a [CSS-parsing bug](https://github.com/cfpb/cfgov-refresh/pull/5137#issuecomment-517743389). This PR fixes the bug by patching its upstream dependency and adding it to package.json as a [selective resolution](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/).

## Additions

- `gulp-bless`
- `resolutions` section `package.json`

## Testing

1. Pull down branch
1. `./frontend.sh`
1. View http://0.0.0.0:8000/ in legacy versions of IE and the carousel should load.

## Screenshots

Screen Shot 2020-02-19 at 1 15 19 PM


anselmbradford:
At some point we should drop support here for IE9 so this dependency is not needed. How do we decide that?

PR #5511, opened 2020-02-19
Revert mobile menu in variation 1 mega menu

anselmbradford:
## Changes

- Revert mobile menu in variation 1 mega menu to that on production.

## Testing

1. Pull branch and build `gulp clean && gulp build`
2. Run site and go to mobile menu and see that you can click into the levels.

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/mega-menu-var-1.html
+++ b/cfgov/jinja2/v1/_includes/organisms/mega-menu-var-1.html
@@ -284,6 +284,7 @@
              if has_children else ''
            }}>
               {{ text }}
+              {{ svg_icon('right') }}
               {% if has_children %}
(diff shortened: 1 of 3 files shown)


chosak:
This works for me, and properly shows only 2 levels when run with the new mega menu backend flag (MEGA_MENU_BACKEND_V2), which is now turned on in production.

The sidebar links don't show up nicely now, unfortunately:

image

I realize we have an outstanding design decision about what to do with these links (i.e. putting them under a separate section under Consumer Tools); for now could we just not show the icon on mobile?

anselmbradford:
@chosak I've pushed up a change for the mobile links so they at least don't look blown up. I'd say we merge this and then iterate on that design.

Screen Shot 2020-02-19 at 5 29 32 PM

PR #5518, opened 2020-02-21
Mega menu: Fix focus state

anselmbradford:
When hovering, the focus state rectangle was too short. This fixes that by moving the hover from a box-shadow to an absolutely positioned `:after` selector.

## Changes

- Changes `box-shadow` hover to an `:after` hover.
- Adjusts height of featured link area to better match specs.

## Testing

1. Pull branch and build.
2. Test var 1 menu.
3. When clicking a menu item, the focus rectangle should be offset outside 2px around all edges. Hover should work across states of hover, active, and focus and combinations of each.

## Screenshots

Screen Shot 2020-02-21 at 5 51 54 PM

### DIFF
--- a/cfgov/unprocessed/css/organisms/mega-menu-var-1.less
+++ b/cfgov/unprocessed/css/organisms/mega-menu-var-1.less
@@ -618,18 +618,29 @@ body {
 
                 &:hover {
                     padding-bottom: ( @grid_gutter-width / 2) - 6px;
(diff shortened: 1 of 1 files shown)


anselmbradford:
@contolini

contolini:
Dang nice attention to detail 👍 :shipit: 

The focus state on the new mm is a huge improvement over the old one.

PR #5519, opened 2020-02-24
Mega menu: Adjustments to featured links

anselmbradford:
Removes DS list classes, since more needed to be overridden than was being applied. It was only used for setting the left padding value to zero. 

## Removals

- Code comment to move menu-specific variable to DS, which doesn't make sense.

## Changes

- Removes DS list classes.
- Moves featured link icons into their associated links.

## Testing

1. Pull branch and build and visit homepage. 
2. Expand menu and visit featured links.
3. Resize page to see at responsive sizes.
4. compare to beta.consumerfinance.gov

## Screenshots

hover

## Todos

- External link utility needs adjusting to work again.


anselmbradford:
@contolini

huetingj:
@anselmbradford let's change the featured item's star icon to black on hover to fall more in line with the hover behavior of the links below it. thank you!

contolini:
> let's change the featured item's star icon to black on hover to fall more in line with the hover behavior of the links below it

It'd also be dope if the entire featured card was clickable (star included) but we can save that for a future PR.

contolini:
> let's change the featured item's star icon to black on hover to fall more in line with the hover behavior of the links below it

It'd also be dope if the entire featured card was clickable (star included) but we can save that for a future PR.

chosak:
@anselmbradford pushed icon fix in 580a14a, but now icon is too close to text:
image

PR #5491, opened 2020-02-05
Form explainers in Wagtail

virginiacc:
Adds the ability to create form explainers in Wagtail rather than as hard-coded pages in cfgov-refresh, like the existing [Loan Estimate](https://www.consumerfinance.gov/owning-a-home/loan-estimate/) and [Closing Disclosure](https://www.consumerfinance.gov/owning-a-home/closing-disclosure/) pages.

This PR focuses on enabling the Wagtail explainer functionality. Much of the front end code here is copied directly from the existing hard-coded explainers (including the Jinja template, CSS, and JS), with minor modifications to remove the tab functionality and accommodate Wagtail data instead of JSON. A future PR will add front end enhancements that can be evaluated against the baseline of these copied files.

### New additions
- Form explainer app with
  - Form explainer Wagtail page type 
  - Form explainer Wagtail block
  - Form explainer Jinja page type template
- Custom CSS for the form explainer block admin display

### Copied files
- Existing [form explainer template](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/owning-a-home/_templates/form-explainer.html) copied into explainer app. Modifications include: 
  - removal of [tabs](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/owning-a-home/_templates/form-explainer.html#L23-L40)
  - accepts Wagtail data instead of imported JSON
  - adds numbered (`.image-map_overlay__category-1`) instead of named (`.image-map_overlay__checklist`) category classes
  - main template code is no longer wrapped in render macro
- Existing [form explainer less file](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/apps/owning-a-home/css/form-explainer.less) is copied into the `apps/form-explainer/css` directory. Modifications include:
  - removal of [tab styles](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/apps/owning-a-home/css/form-explainer.less#L211-L249)
  - numbered (`.image-map_overlay__category-1`) instead of named (`.image-map_overlay__checklist`) category classes
- Existing [FormExplainer js](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/apps/owning-a-home/js/form-explainer/form-explainer.js) is copied into the `apps/form-explainer/js` directory. Modifications include:
[...]

PR #5523, opened 2020-02-25
Add new EnforcementActionPage with metadata

schbetsy:
Currently, we use the Document Detail page type to create pages about Enforcement Actions in Wagtail, and use the Related Metadata streamfield block to store important information about the action. This means that the metadata (which should be present on all enforcement actions) might be missing or malformed, and is not easily searchable or filterable.

This PR adds an Enforcement Action page type, which includes a new tab in the wagtail interface called "Metadata". It contains database fields for the required metadata. This will allow us to filter on these values in the Enforcement Actions filtered list page in a future update.

## Additions

- `EnforcementActionPage` page type
- `EnforcementActionStatus` class to allow an Enforcement Action to have an arbitrary number of statuses
- Templates and styling for the new page type

## Testing

1. Running this branch, log in to the [Wagtail admin](http://localhost:8000/admin).
2. Create a new page of the Enforcement Action Page type.
    1. Be sure to fill in the information in the metadata panel
    2. Publish the page
3. Visit the page you created, see that the metadata are displayed correctly in the sidebar of the published page.

## Screenshots
The Metadata panel on the published Enforcement Action page:
Screen Shot 2020-02-27 at 12 31 52 PM

The Metadata panel on the editor's view in Wagtail:
Screen Shot 2020-02-27 at 2 50 17 PM

## Todos

- ~~Styling for enforcement action status in sidebar (@wpears)~~
- ~~Make the migration Wagtail 2 compatible (Betsy)~~
- ~~Require at least one Category and Enforcement Status~~
- Require at least one Topic
- ~~Require `date_filed` field~~

In a future update, we will migrate existing Enforcement Actions (currently stored in `DocumentDetailPage`) to the new `EnforcementActionPage`

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


schbetsy:
@wpears 
Since the `date_filed` field is inherited from the parent class (`AbstractFilterPage`, which is not actually abstract), we can't override it in our `EnforcementActionPage` class. So, we can't just make `date_filed` required on this model. We'll have to either (1) make the field required for all page types that inherit from `AbstractFilterPage`, which will affect 4000-ish existing pages, or (2) create a new field (`filed_date`? yuck ) on the `EnforcementActionPage` and make it required.

I believe I read that they're (eventually) considering updating how the `date_filed` field works (e.g. updating to "initial date filed", then "last updated" instead). So another option could be to leave the `date_filed` field optional for now, and architect the whole date-tracking feature later when we have the requirements in hand.

wpears:
(Woops I didn't actually send this comment 😅)

Let's just leave `date_filed` as optional. I'll make sure the template doesn't error with a missing date (or maybe we SHOULD make the template error with a missing `date_filed` as a hacky actually-this-field-is-required signal to be seen when the content editor previews the changes)

So either way I think we do nothing on the back end for `date_filed` right now but I'm neutral on whether we error in the template or not if it's missing.

schbetsy:
Let's leave `date_filed` as optional in the template, so we don't have user-facing errors. We can note in the documentation that the field shouldn't be left blank.

wpears:
Looks great to me! Questions likely worth some discussion among backends:
  - Is creating new page types like this something we should be doing/doing more of?
  - The `AbstractFilterPage` isn't really abstract..! Should it be?
  - There was some necessary duplication between the `EnforcementActionPage` and the `AbstractFilterPage` (specifically wrt to moving the `date_filed` field to our new Metadata tab, which required redefining the `settings_panel` in the derived class. Is this gross? Is there a better way? Should we care about this?
  - How does/should testing work here? Seems like `AbstractFilterPage` tests are pretty limited, should we take a similarly limited approach with this new page type?

willbarton:
> Is creating new page types like this something we should be doing/doing more of?

I think it generally depends on the need. I don't think we're opposed to doing so, but it should be done with consideration (@cfpb/back-end-admins for any further opinions, because I don't want to speak for everyone).

> The AbstractFilterPage isn't really abstract..! Should it be?

Not really... mostly because I shudder at the thought of the data migrations. Plus, given that Wagtail's `Page`, `CFGOVPage`, and everything above `AbstractFilterPage` is not abstract, I think it makes sense to keep it concrete — it's more in line with Wagtail page inheritance assumptions (IMHO... I'd love @chosak's thoughts). It'd be kinda weird to have an in-between abstract model.

I think it should possibly be renamed to something like`BaseFilterPage` or `CFGOVFilterPage`.

`AbstractFilterPage` is only really unfortunate because `Abstract` in the model name implies a particular type of Django model that `AbstractFilterPage` is not. 

> There was some necessary duplication between the EnforcementActionPage and the AbstractFilterPage (specifically wrt to moving the date_filed field to our new Metadata tab, which required redefining the settings_panel in the derived class. Is this gross? Is there a better way? Should we care about this?

There's not a better way to do that that I'm aware of... although, that we need to redefine the settings panel here suggests that maybe we should consider how we handle "settings" vs "metadata" on filterable pages more generally. Is this the only page type that will have a "metadata" tab?

> How does/should testing work here? Seems like AbstractFilterPage tests are pretty limited, should we take a similarly limited approach with this new page type?

Coverage for `AbstractFilterPage` looks decent (I haven't looked any closer than that though, and I know a *lot* of its filtering functionality lives in weird places elsewhere), but for *this* change, if you're not adding methods/functionality to your new enforcement models, I don't think there's anything to unittest.

willbarton:
>>There was some necessary duplication between the EnforcementActionPage and the AbstractFilterPage (specifically wrt to moving the date_filed field to our new Metadata tab, which required redefining the settings_panel in the derived class. Is this gross? Is there a better way? Should we care about this?
> There's not a better way to do that that I'm aware of... although, that we need to redefine the settings panel here suggests that maybe we should consider how we handle "settings" vs "metadata" on filterable pages more generally. Is this the only page type that will have a "metadata" tab?

Thinking a little more about this, if we're uncomfortable with the way it's done here, we could break apart the list of panels such that each subclass of `AbstractFilterPage` could compose its panels as it wants to without redefining them wholesale. But I feel like that might be a bit too much for this particular case. I think this is fine for now, personally.

Scotchester:
> > Is creating new page types like this something we should be doing/doing more of?
> 
> I think it generally depends on the need. I don't think we're opposed to doing so, but it should be done with consideration (@cfpb/back-end-admins for any further opinions, because I don't want to speak for everyone).

I think it makes a lot of sense in a case like this, where this new type of page has a discrete set of unique fields. I'm all for it 👍

chosak:
I agree that we should have more page types. I do think it would be better if our abstract page types were actually abstract, as it'd simplify the database, but that'd be a larger change than this PR.

The one piece that I do think would be good to address is the template duplication, specifically the content in [enforcement-action/index.html](https://github.com/cfpb/cfgov-refresh/pull/5523/files#diff-5310b8efa9402cbdd43fcc07da4983c2). There's a lot of boilerplate here seemingly copied from other page templates. I made a similar comment [here](https://github.com/cfpb/cfgov-refresh/pull/5491#discussion_r386615976) on @virginiacc's FormExplainerPage PR. Duplicating, for example, all of [this code](https://github.com/cfpb/cfgov-refresh/pull/5523/files#diff-5310b8efa9402cbdd43fcc07da4983c2R14-R36) feels like technical debt to me. Is there some way of abstracting out some base page template(s) and then only customizing as necessary?

wpears:
>Duplicating, for example, all of this code feels like technical debt to me. Is there some way of abstracting out some base page template(s) and then only customizing as necessary?

I agree this makes sense, I'm going to open a ticket about this and link back to this convo, it may be that the work to do here would best be done in one larger/organized effort (eg where "base" templates can be better defined, etc.) If that stalls out though I'll just do the work to limit the duplication in this case only.

PR #5524, opened 2020-02-25
Improve homepage carousel hierarchy

contolini:
It's unclear to screen readers that the carousel is a set of featured items. Grouping them under a visually hidden heading and improving the hierarchy helps distinguish it.

https://www.w3.org/WAI/tutorials/carousels/structure/

## Changes

- Changes `h5` above carousel to an `h2` and makes its visually hidden but available to screen readers.
- Changes `Featured` heading to read `Featured content`.

## Testing

1. Pull down branch.
1. Load homepage.
1. Everything should visually look the same as https://www.consumerfinance.gov/ except small screens should now read `Featured content` above the featured content.

## Screenshots

Screen Shot 2020-02-25 at 12 57 31 PM

## Notes

- Not sure if "Featured content" is the correct language. "Featured items", "Showcase items", etc. might be more appropriate.

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/carousel.html
+++ b/cfgov/jinja2/v1/_includes/organisms/carousel.html
@@ -27,7 +27,7 @@
 <section class="o-carousel_item">
     <div class="o-carousel_item-wrapper">
         <div class="o-carousel_item-text">
-            <h2>{{ item.title }}</h2>
+            <h3 class="h2">{{ item.title }}</h3>
             <p>{{ item.body }}</p>
             <a class="a-link a-link__jump" href="{{ item.link_url }}">
                 <span class="a-link_text">{{ item.link_text }}</span>
@@ -85,14 +85,14 @@
 {% endmacro %}
 
 {% macro render( value ) %}
-<div class="o-carousel u-hidden">
-    <h5>{{ _( 'Featured' ) }}</h5>
+<div class="o-carousel u-hidden" aria-labelledby="o-carousel_heading">
+    <h2 class="h5" id="o-carousel_heading" aria-label="Featured content">{{ _( 'Featured' ) }}</h2>
     <div class="o-carousel_navigator">
 
-        <button class="o-carousel_btn o-carousel_btn-prev a-btn" aria-label="Previous showcase item">
+        <button class="o-carousel_btn o-carousel_btn-prev a-btn" aria-label="Previous featured content item">
             {{ svg_icon( 'left' ) }}
         </button>
-        <button class="o-carousel_btn o-carousel_btn-next a-btn" aria-label="Next showcase item">
+        <button class="o-carousel_btn o-carousel_btn-next a-btn" aria-label="Next featured content item">
             {{ svg_icon( 'right' ) }}
         </button>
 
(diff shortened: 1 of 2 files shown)


cfgov/jinja2/v1/_includes/organisms/carousel.html
```diff
@@ -27,7 +27,7 @@
 <section class="o-carousel_item">
     <div class="o-carousel_item-wrapper">
         <div class="o-carousel_item-text">
-            <h2>{{ item.title }}</h2>
+            <h3 class="h2">{{ item.title }}</h3>
```
contolini: It feels hacky using classes like this to restyle headings but it feels better than adding CSS rules. LMK if there's a better approach.

anselmbradford: I think this is the hacky approach we use elsewhere, so :+1:

cfgov/unprocessed/css/organisms/carousel.less
```diff
@@ -209,9 +209,10 @@
       display: none;
     }
 
-    // Hide "featured" heading at desktop size.
-    & h5 {
-      display: none;
+    // Hide "featured content" heading at desktop size but keep it available
+    // to screen readers.
+    & h2 {
```
anselmbradford: How about we make this `.h5` instead of `h2` to better convey the appearance?

cfgov/jinja2/v1/_includes/organisms/carousel.html
```diff
@@ -85,14 +85,14 @@ <h2>{{ item.title }}</h2>
 {% endmacro %}
 
 {% macro render( value ) %}
-<div class="o-carousel u-hidden">
-    <h5>{{ _( 'Featured' ) }}</h5>
+<div class="o-carousel u-hidden" aria-labelledby="o-carousel_heading">
+    <h2 class="h5" id="o-carousel_heading">{{ _( 'Featured content' ) }}</h2>
```
anselmbradford: We should check with UX whether this text should be changed. We could have `aria-label` on the `h2` alternatively.

anselmbradford: @sonnakim

sonnakim: @anselmbradford works for me — either changing the H2, or just the aria label, to add the word "content"  👍

huetingj: We talked during team time about changing the aria label to "featured content"

chosak: I think because this uses Django's translation framework (the usage of `_( )`), you'll need to change [the text translation file](https://github.com/cfpb/cfgov-refresh/blob/bc4c5941d49c1ab1c50f51af9d8a67815ac84e7f/cfgov/locale/es/LC_MESSAGES/django.po#L82-L83) to add a Spanish translation for this phrase, and then run [this step](https://cfpb.github.io/cfgov-refresh/translation/#4-run-the-compilemessages-management-command-to-compile-the-machine-object-file) to recompile the binary translation file.

contolini: @chosak You raise a good point that we need to ensure we translate aria labels.

PR #5525, opened 2020-02-25
Remove cfpb-chart-builder from the landing map dependencies

JeffreyMFarley:
Fixed up `package.json` and `yarn.lock` so that chart-builder is no longer included.


flacoman91:
The only issue I see with this is here:
https://github.com/cfpb/cfgov-refresh/blob/ccdb-dev/cfgov/unprocessed/apps/ccdb-landing-map/js/TileMap.js#L5

We still reference the chart builder function here which contains the svg paths for the tiles i believe.

JeffreyMFarley:
Yeah, that's weird, because it looks fine
Screen Shot 2020-02-25 at 2 29 00 PM

I'll investigate further

sephcoster:
In order for this to install in offline mode, you'll also want to add the following to your `.yarnrc` file before running `yarn`, then restoring the file so the correct deps are checked-in:
```
yarn-offline-mirror-pruning true
--install.offline false
```

JeffreyMFarley:
I can also check in the offline packages

JeffreyMFarley:
> The only issue I see with this is here:
> https://github.com/cfpb/cfgov-refresh/blob/ccdb-dev/cfgov/unprocessed/apps/ccdb-landing-map/js/TileMap.js#L5
> 
> We still reference the chart builder function here which contains the svg paths for the tiles i believe.

It was finding `cfpb-chart-builder` in the root project and using the values from there.

flacoman91:
Should we leave in the cfpb-chart-builder dependency and remove the highcharts dependency?  
or remove both since the parent cfgov-refresh should have it already?

JeffreyMFarley:
> Should we leave in the cfpb-chart-builder dependency and remove the highcharts dependency?
> or remove both since the parent cfgov-refresh should have it already?

I don't really like the "shadow dependency".  What I think I will do, is
1. Wait for you to finish your work with linting and testing
1. Merge
1. Copy the `paths` from chart builder into a `constants.js`

Scotchester:
Hi folks,

Could I please request that your PR titles be a little less ambiguous? For someone who doesn't have the ambient context that you're working on a development branch, it could be confusing. Also, these PR titles may show up in the autogenerated changelog after this dev branch is merged to master, depending on how that merge is executed.

Thanks for considering!

PR #5528, opened 2020-02-25
Remove extra borders from FCMs on some page types

Scotchester:
The Featured Content Module component has provided its own border for a long time. We apparently haven't noticed that it gets doubled up on Browse, Browse Filterable, and Newsroom Pages, probably because FCMs are rarely used on those pages types. This commit removes the class in the markup for those page type templates that adds the extra border.

## Removals

- `block__border` class that was on the wrapper for FCMs on the Browse, Browse Filterable, and Newsroom templates

## Checklist

### Browsers

- [x] Chrome on desktop
- [ ] Firefox
- [x] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android


anselmbradford:
Should these templates be pulling in a fcm template of some sort? Why the redundancy?

niqjohnson:
It looks like that same extra `block__border` deal is also going on in:

https://github.com/cfpb/cfgov-refresh/blob/96f80bec517fa56d685ff6f84fb7620af697e221/cfgov/paying_for_college/jinja2/paying-for-college/choose-a-student-loan.html#L19-L28

I'm not sure what that template is used for, though, so I can't find a page to test to see for sure if we need to remove the `block__border` class there, too.

Scotchester:
> Should these templates be pulling in a fcm template of some sort? Why the redundancy?

They do pull in the FCM template, whose Less includes the border, but they do that _inside_ a `div.block.block__border`.

Scotchester:
> It looks like that same extra `block__border` deal is also going on in:
> 
> https://github.com/cfpb/cfgov-refresh/blob/96f80bec517fa56d685ff6f84fb7620af697e221/cfgov/paying_for_college/jinja2/paying-for-college/choose-a-student-loan.html#L19-L28
> 
> I'm not sure what that template is used for, though, so I can't find a page to test to see for sure if we need to remove the `block__border` class there, too.

Good catch, @niqjohnson! I don't believe this template is in use yet, and it's meant to replace the currently legacy Choose a Student Loan page. I will remove that one, as well.

PR #5532, opened 2020-02-26
Add basic `watch` gulp task

ascott1:
This is a proof of concept that adds a basic watch task for uncompiled Less and JS during local development. It watches files in the src directories for changes and runs the 'modern' gulp tasks when they change. This will not run commands to update the application specific styles/scripts, the on demand atomic components, or legacy browser support.


gulp/tasks/scripts.js
```diff
@@ -252,3 +252,9 @@ gulp.task( 'scripts',
     'scripts:ondemand'
   )
 );
+
+gulp.task( 'scripts:watch', function() {
+  gulp.watch( configScripts.src, gulp.parallel(
+    'scripts:modern'
+  ) );
```
Scotchester: While this passes the linter, I think we would prefer the arguments to `gulp.watch` start on their own lines when one is a function.

```suggestion
  gulp.watch(
    configScripts.src,
    gulp.parallel( 'scripts:modern' )
  );
```

gulp/tasks/styles.js
```diff
@@ -235,3 +235,9 @@ gulp.task( 'styles',
     'styles:ondemand'
   )
 );
+
+gulp.task( 'styles:watch', function() {
+  gulp.watch( `${ configStyles.cwd }/**/*.less`, gulp.parallel(
+    'styles:modern'
+  ) );
```
Scotchester: Same here.

```suggestion
  gulp.watch(
    `${ configStyles.cwd }/**/*.less`,
    gulp.parallel( 'styles:modern' )
  );
```

Scotchester:
Not sure why this was marked "[WIP]", but it seems good to go to me. Just be sure to edit the title before merging.

PR #5483, opened 2020-01-31
Homepage: Make cards fully clickable at mobile size

anselmbradford:
## Changes

- Make cards on homepage fully clickable at mobile size.

## Testing

1. Pull branch and build. Visit homepage and resize the page to see the behavior of the cards at desktop and mobile.
2. Use the tab key to tab through content and see where the focus falls on the cards.

## Screenshots

Mobile card with focus rectangle:

Screen Shot 2020-01-31 at 3 17 55 PM

PR #5538, opened 2020-02-27
Update Card pattern and fix ask cards

anselmbradford:
The /ask-cfpb/ cards were broken by https://github.com/cfpb/cfgov-refresh/pull/5483
This PR updates the card pattern to support both the homepage and ask card layouts.

See accompanying Design System update https://github.com/cfpb/design-system/pull/428

## Changes

- Update Card pattern and fix Ask CFPB cards.
- Adds hover state to card heading.

## Testing

2. Visit homepage and compare cards to production.
3. Visit /ask-cfpb/ and compare to https://web.archive.org/web/20200105175022/https://www.consumerfinance.gov/ask-cfpb/

## Screenshots

Screen Shot 2020-02-27 at 5 19 11 PM

## Notes

- This uses CSS grid now as this is perfectly suited for a grid layout, however, to avoid complex IE hacks, the cards are laid out vertically in IE.


anselmbradford:
@contolini

cfgov/unprocessed/css/pages/ask.less
```diff
@@ -328,19 +328,6 @@
                 .respond-to-range( @bp-sm-min, @bp-sm-max, {
                     .h3();
```
anselmbradford: The heading changes from `h2` to `h3` at smaller sizes, should this be part of the wider pattern? Currently this only happens in the /ask-cfpb/ cards.

huetingj: :+1: that makes sense to me. Changing it from h2 to h3 falls in line with how we typically handle headings at small screen size. I guess Ask CFPB is the only implementation of this as of now? (The only other cards I can think of are the homepage cards, and those don't have headings.)

anselmbradford: Yes it's ask and the homepage. By default implementing this across the cards would mean the icon on the homepage cards would go from `h2` to `h3` also when the screen size is smaller.

huetingj: oh right— i think that would be fine but would you mind posting a screenshot?

anselmbradford: h2
Screen Shot 2020-02-27 at 6 30 29 PM

h3
Screen Shot 2020-02-27 at 6 30 37 PM


It's pretty subtle.

anselmbradford: err sorry, that's a shoddy screenshot because it shows the boxes different sizes. Maybe best if I demo for you.

huetingj: It's fine! When I download the screenshots and look, it's easy to see. I think h3 looks good. Let's go for it! Thanks again for bringing this up.

anselmbradford: Cool! This responsive sizing is now is https://deploy-preview-428--cfpb-design-system.netlify.com/design-system/components/cards as well

cfgov/unprocessed/css/organisms/card-group.less
```diff
@@ -1,47 +1,42 @@
...
+        padding: unit( 30px / @base-font-size-px, em );
         background: @green-20;
     }
+}
+
+// Hack for Edge/IE to convert grid to block layout.
+// See https://jeffclayton.wordpress.com/2015/04/07/css-hacks-for-windows-10-and-spartan-browser-preview/
+// Necessitated because of https://stackoverflow.com/questions/46743722/items-in-css-grid-in-ie-superimposed-on-each-other
+_:-ms-lang(x),.o-card-group_cards {
+    display: block !important;
```
virginiacc: I'm not necessarily opposed, but I am wondering whether it makes sense to convert the cards to grid if that necessitates block display on all IE. It looks like IE11 is one of the most commonly used browsers on our site, so we'd be serving the stacked display to a fairly large group of users. Is there a workaround we could use instead of block?

cfgov/jinja2/v1/ask-cfpb/landing-page.html
```diff
@@ -37,48 +37,47 @@ <h2 class="u-show-on-mobile u-mt30">
...
+                                    {{ partially_styled_link.render({
+                                        'plain_text': page.question,
+                                        'underlined_text': _('Read answer'),
+                                        'url': page.url })
+                                    }}
+                                    </li>
+                                {% endfor %}
+                            </ul>
+
+                            <div class="content-l_footer__equal-height">
```
virginiacc: Should this have the class `m-card_footer` instead?

cfgov/jinja2/v1/ask-cfpb/landing-page.html
```diff
@@ -37,48 +37,47 @@ <h2 class="u-show-on-mobile u-mt30">
...
-                                    <div class="content-l_footer__equal-height">
-                                        <p class="short-desc u-mb0">
-                                            <a href="{{ portal.url }}">
-                                               {{ _('Learn more about') }} {{ portal.title|lower }}
-                                            </a>
-                                        </p>
-                                    </div>
-
-                                </article>
+            <div class="u-hide-on-mobile question-categories">
```
virginiacc: One thing I'm wondering (probably outside the scope of this PR) is whether, instead of being swapped out for a list of links, the Ask cards could reduce to their header links on mobile? We could probably use flex & `row-reverse` to flip the icons to the right and match the existing links' layout...

cfgov/jinja2/v1/ask-cfpb/landing-page.html
anselmbradford: Yes! Fixed in https://github.com/cfpb/cfgov-refresh/pull/5538/commits/4cc683577077f1295120ea49eabda10a04ab785b

cfgov/jinja2/v1/ask-cfpb/landing-page.html
anselmbradford: I wondered this too. This should probably be hashed out in https://github.com/cfpb/design-system/ in the future.

cfgov/unprocessed/css/organisms/card-group.less
anselmbradford: It's a good point, IE11 is about 10% of visitors. My thinking here was that this is the kind of layout grid was made for so we get a lot of flexibility with a little code. We certainly could expand on the fallbacks for IE in the above block, but I didn't think it looked terrible as is, so that extra maintenance burden may not be worth it. Flexbox is an option too, but I was worried about unexpected wrapping across the breakpoints.

Screen Shot 2020-03-03 at 5 43 04 PM

Of course it matters if the stakeholders aren't happy with that change, but I'm not sure who has say on this page? cc @schaferjh

virginiacc: I agree that it would be really nice to use grid here and I don't think that the Ask cards look too bad stacked either. I wonder if there's more visual impact with the three columns of cards stacking on the home page, though.

anselmbradford: How about this? I put in a floated layout fallback in https://github.com/cfpb/cfgov-refresh/pull/5538/commits/6cd9abd3a6d3f50bae90609066a3d84dcf792d4f

The only issue is the boxes aren't equal height and there's slightly more margin on the right. I'm not sure how else to do a fallback without putting each row into its own container. What do you think?

Screen Shot 2020-03-05 at 1 37 48 PM

Screen Shot 2020-03-05 at 1 37 09 PM

Screen Shot 2020-03-05 at 1 37 30 PM

virginiacc: This definitely looks better to me. I defer to @huetingj, but I would be fine with merging this to get the fix out for Ask. We can always revisit the fallback in the design system if necessary.

huetingj: :+1: from me to merging this side-by-side layout.

PR #2481, opened 2016-10-21
StreamField migration utilities

willbarton:
This PR includes some utility functions adapted from [0011_hero_refactor_data.py](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/migrations/0011_hero_refactor_data.py) that will hopefully make StreamField data migrations more straight-forward in the future. These utilities were reviewed as part of #2450, but they're broken into their own PR here.

I have added some documentation around Wagtail migrations based on the previous work @chosak and @kurtrwall have done, the discussion in GHE#323, and the way I've modified it here. 

## Changes

- Refactored `get_page_stream_data` and `set_page_stream_data` to handle either page objects or revision objects, renamed them `get_stream_data` and `set_stream_data`, and moved them to `v1.util.migrations`.

## Additions

- Added the `show_category` field on the ItemIntroduction organism
- Added `migrate_stream_field` and `migrate_page_types_and_fields` utility functions
 
## Testing

- Run `tox` to run the tests for the migration and migration utility functions.

## Review

- @chosak
- @cfpb/cfgov-backends 

## Checklist

* [x] Changes are limited to a single goal (no scope creep)
* [x] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [x] Passes all existing automated tests
* [x] New functions include new tests
* [x] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [x] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #5539, opened 2020-02-28
Add support for multiple leads in Bureau Structure

chosak:
Currently the Bureau Structure blocks only allow for at most two leads per division and only one lead in each office. This change modifies the schema so that both divisions and offices can have a configurable number of leads.

See https://www.consumerfinance.gov/about-us/the-bureau/bureau-structure/ for the current production code that uses these blocks.

## Testing

You'll need to run the new migration in this PR before testing the code. Run a local server and edit the Bureau Structure page, and try adding multiple people to different roles.

## Screenshots

|||
|-|-|
|image|image|

## Notes

@willbarton @Scotchester I've tagged you because of the change to the data migration utility; see the commit message in 10de39e for a little more detail.

## Todos

~~To keep the scope of this small, I didn't mess with the frontend layout very much. I leveraged the existing styles that wrap each person in `o-bureau-structure_role`, and, although the visual design seems correct, it feels weird to have the office name getting displayed as part of the first role, as opposed to a distinct element. Is there a (constrained) improvement that can be made to this? @anselmbradford @contolini~~ Fixed below thanks to suggestion from @anselmbradford.

## Checklist


cfgov/v1/util/migrations.py
```diff
@@ -77,17 +77,11 @@ def get_stream_data(page_or_revision, field_name):
     revision """
     if is_page(page_or_revision):
         field = getattr(page_or_revision, field_name)
-        stream_block = field.stream_block
-        stream_data = stream_block.get_prep_value(field)
+        return field.stream_data
```
willbarton: This is a nice simplification — I went back through the history to see if I could figure out why we weren't just using `field.stream_data` to begin with, and the [PR that last refactored this still](https://github.com/cfpb/cfgov-refresh/pull/2481) doesn't [yield any good reasons](https://github.com/cfpb/cfgov-refresh/pull/2481/commits/5a49cb65662ee03456bd72c08d9e95ad698062fa) on the commit that made the change.

Looks good to me!

Scotchester: I agree, I like this change!

willbarton: [Right now our documentation says](https://cfpb.github.io/cfgov-refresh/migrations/#wagtail-specific-considerations) to have a schema migration to add the new fields, data migration to migrate, and then another schema migration to remove the old fields. Should the docs be updated with this change?

Scotchester: I don't think this change impacts the need to do it that way, does it?

chosak: It's complicated: in general best practice with data migrations is to always do them in multiple steps (because then you don't break the database out from under an older version of the code), although in practice we rarely do it that way. That documentation page could definitely be cleaned up; for example it mentions moving data "from the old field to the new field" but I think it means to talk about StreamField _blocks_.

One key thing about this change, is that it's more than a simplification but actually necessary for this to work. The issue is that if you load `stream_data` into Python, Wagtail strips out blocks that aren't currently defined in the model field. So if your code alters a StreamField to replace one block with another, and the data migration code uses our existing `get_stream_data`, you don't have a way to get the old block data to migrate it to the new block. Instead you have to use `.stream_data` to bypass the Python conversion.

Thanks for the reviews; at some point I'd like to revisit SF data migrations but wanted to do the minimum here to make this PR possible.

PR #5529, opened 2020-02-26
Remove unused Hallo.js plugins

willbarton:
This PR removes two Hallo.js plugins: the ask-tips Hallo plugin (which isn't being used anymore as of #4982 and #5024) and the edit-html (which was removed for users in #5055).

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [ ] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [x] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #5541, opened 2020-03-02
Wagtail 2: Support both Draftail & Hallo, fix Hallo in table cell RTEs

Scotchester:
When upgrading to Wagtail 2, we want to move as much of the rich text editing as possible to Draftail, but there is at least one corner case: our bolted-on ability to use the rich text editor when editing cells in the `AtomicTableBlock` (which is an exension of Wagtail's default
`TableBlock`).

In this particular case, it was easier/faster to maintain support for using Hallo as the table cell editor than figure out how to call in Draftail instead.

We still want to convert those table cell editors to Draftail, but it's something we can put on the backburner to get to Wagtail 2 faster.

See https://GHE/CFGOV/platform/issues/3707

## Additions

- Ability to enable the Hallo.js editor on any rich text field (this may not strictly be necessary, but it seemed prudent to leave in for now)

## Changes

- Updates `table-block.js` to fix the loading of Hallo RTEs when editing a table cell
- Hallo JS and CSS files are included on every editor page (not sure if it's possible to limit to pages where an `AtomicTableBlock` is present)

## Testing

1. Pull branch
1. With Wagtail 1.13 installed/running, ensure that Hallo still functions correctly everywhere on http://localhost:8000/admin/pages/8264/edit/
1. `pip install wagtail==2.3`
1. Reload http://localhost:8000/admin/pages/8264/edit/
1. See that Draftail works as expected in the RTFs outside of the tables
1. Open a table cell for editing and see that it loads Hallo
1. Test editing table cells with various Hallo features
1. Preview and ensure that the changes you made are all visible on the rendered page

## Technical notes:

- In order to ensure that the table cell editor sets up Hallo the way Wagtail 2 does, adjustments were made to `table-block.js` by comparing [the version of Wagtail core's `hallo-bootstrap.js` that we originally cribbed from in 2016](https://github.com/wagtail/wagtail/blob/6c9d202057babfb1fbe59a1bd88a9f4b5273a846/wagtail/wagtailadmin/static_src/wagtailadmin/js/hallo-bootstrap.js) to [the version currently in Wagtail master](https://github.com/wagtail/wagtail/blob/6feb29d54a4ea458d6453786a8d8d02b2b417929/wagtail/admin/static_src/wagtailadmin/js/hallo-bootstrap.js). This amounted to changing a class name, adding a data attribute, and changing event bindings from `bind` to `on`. I also had to manually list the plugins we wanted to be available in the Hallo toolbar, because the original `halloPlugins` system default was not working.
- I also had to make `hallo.js`, all Hallo plugin JavaScript files, and `hallo.css` available to all editor pages in the admin. I'm not sure if there is a way to limit those to pages where an `AtomicTableBlock` is added.
- Includes commit originally from #5529 by @willbarton: Fix removed expand_db_html argument – The `for_editor` argument was removed from `expand_db_html` at this change for Wagtail 2.0: https://github.com/wagtail/wagtail/commit/3e1ab7e74f5b182b8f591d5b1ed2dc0f556812e2

## Checklist

### Browsers

- [x] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android


higs4281:
Admin pages run without errors, but I don't get the rich-text controls inside a table cell in 2.3. 
Here's how the rich-text controls appear when in 1.13:

cell_wag113


And here's what I'm seeing in 2.3:

cell_wag23

The link works correctly, but I can't edit it via the hallo widget.

I ran migrations after loading 2.3 and ran the frontend script.

higs4281:
My editor anomaly was browser cache, not a Draftail issue.

cfgov/v1/wagtail_hooks.py
```diff
@@ -86,9 +87,31 @@ def log_page_deletion(request, page):
 
 @hooks.register('insert_editor_js')
 def editor_js():
-    js_files = [
-        'js/table-block.js',
-    ]
+    js_files = ['js/table-block.js']
+
+    if wagtail.VERSION >= (2, 0):
```
willbarton: Could we have a comment here explaining why we're inserting the hallo js files?

Scotchester: You got it!

PR #5542, opened 2020-03-02
Mega menu: Adjust tab padding

anselmbradford:
## Changes

- Adjust tab padding so caret isn't so tight on the right.
  
## Testing

1. Pull branch and build. Tab should have more space on the right and now be equal in space on the left and right.

## Screenshots

Screen Shot 2020-03-02 at 6 03 52 PM


anselmbradford:
@contolini

contolini:
The padding looks good but it causes the mm items to break to a second line between ~ 930px and 900px.

mm-carets

Can we skim a couple pixels off somewhere else to compensate?

PR #5552, opened 2020-03-06
Creates base page template and page_type variable

wpears:
Creates a base template that accounts for varying behavior
between learn-page, document-detail, and enforcement-action.
This base template is extended by these child pages and the
page_type variable is set to trigger different template
behavior.

@chosak Was this what you were after? It's likely that other page_types can use this same base template as well.


chosak:
@wpears thanks for keeping this dialogue going.

Creation of a `page_type` variable shouldn't be necessary because Wagtail pages always pass the [`page`](https://docs.wagtail.io/en/stable/topics/pages.html#template-context) variable to the template context, allowing for checks against e.g. [`Page.specific_class`](https://docs.wagtail.io/en/stable/reference/pages/model_reference.html#wagtail.core.models.Page.specific_class).

More generally, it should be possible to completely remove logic like [this](https://github.com/cfpb/cfgov-refresh/blob/ff9713456466fca46e07f0d27a71ca105c79a383/cfgov/jinja2/v1/_layouts/page.html#L15-L17) or [this](https://github.com/cfpb/cfgov-refresh/blob/ff9713456466fca46e07f0d27a71ca105c79a383/cfgov/jinja2/v1/_layouts/page.html#L45-L48) if we standardize on using [`{% include_block %}`](https://docs.wagtail.io/en/stable/topics/streamfield.html#template-rendering) to render blocks instead of loading templates manually (we have an internal issue about this at platform#378).

Ideally the page template could look something this:

```html
{% block content_main %}
  {% for block in page.header %}
    <div class="block">
      {% include_block block %}
    </div>
  {% endfor %}

  {% for block in page.content %}
    <div class="block">
      {% include_block block %}
    </div>
  {% endfor %}
{% endblock %}

{% block content_sidebar %}
  {% for block in page.sidefoot %}
    <div class="block">
      {% include_block block %}
    </div>
  {% endfor %}
{%- endblock %}
```

or, if we create a custom [`StreamBlock`](https://docs.wagtail.io/en/stable/topics/streamfield.html#streamblock) implementation that handles the right `block` classes, even just:

```html
{% block content_main %}
{% include_block page.header %}
{%  include_block page.body %}
{%- endblock %}

{% block content_sidebar %}
{% include_block page.sidefoot %}
{% endblock %}
```

This would greatly simplify our page templates.

For this to work, though, all of our blocks would have to consistently know how to render themselves, meaning they'd all need a `class Meta: template` definition, and we'd have to figure out how to fit [this logic](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/templates/render_block.html) in somewhere.

PR #5553, opened 2020-03-06
Add Jenkins multibranch builds and preview environments

hkeeler:
This introduces Jenkins [multibranch](https://jenkins.io/doc/book/pipeline/multibranch/) builds, with optional preview environments per-PR.  This first iteration introduces a standard Docker build/scan/push/deploy type workflow, but _could_ be used for other workflows we'd like to run on our internal Jenkins servers in the future.

## Additions

- `Jenkinsfile` -  A Jenkins pipeline script designed for [multibranch](https://jenkins.io/doc/book/pipeline/multibranch/) builds.  It currently focuses on the `cfgov-prod` Docker setup; building, scanning, pushing the Docker image, and optionally deploying the stack to our development hosting platform.
- `docker-stack.yml` - A [Docker Swarm](https://docs.docker.com/engine/swarm/) compatibile Compose file.
- [Docker Secrets](https://docs.docker.com/engine/swarm/secrets/) - The `cfgov-prod` image will now translate any Secrets into environment variables at container startup.  This allows envvars such as `PGPASSWORD` to be registered as Secrets and looked up by the container runtime instead of putting them right in the Compose file.  See [`Dockerfile:77`](https://github.com/cfpb/cfgov-refresh/compare/master...hkeeler:jenkinsfile#diff-3254677a7917c6c01f55212f86c57fbfR77) for details.
- [Docker `HEALTHCHECK`](https://docs.docker.com/engine/reference/builder/#healthcheck) - The `cfgov-prod` image now includes a `HEALTHCHECK`, which polls the root URL of the app from within the container, letting the Docker scheduler know when the app is _really_ up, and when if it has crashed or is experiencing degraded response times.  This is a requirement of our image vulnerability scanner.

## Changes

- The Apache `Listen` port is no longer hardcoded to `80`, and is now set via the `APACHE_PORT` environment variable.
- The `cfgov-prod` variant of container now runs as `apache`, not `root`.  This is also a requirement of our image vulnerablity scanner.

## Testing

- [x] Confirm the current Docker dev setup continues to work as expected.
- [x] Confirm the Jenkins pipeline build and deploys as expected.

## Notes

- This should not affect the current development Docker setup at all.
- ~~Jenkins multibranch builds update Github commit status.  Once this is merged, you will see a new **continuous-integration/jenkins/pr-merge** check on each PR.  I will make sure this check does not block PRs from being merged.~~ (See: https://github.com/cfpb/cfgov-refresh/pull/5553#issuecomment-601446387)
- This `Jenkinsfile` abstracts the details of deploying to our internal systems via [Shared Libraries](https://jenkins.io/doc/book/pipeline/shared-libraries/).  These shared libraries can be found on our internal GHE repo **app-ops/app-ops-jenkins-shared-libraries**.

## Todos

- [x] Add short description of the `Jenkinsfile` and `docker-stack.yml` to the docs.
- [x] Refactor Ansible playbook to account for `APACHE_PORT` change.
- [x] Make sure **continuous-integration/jenkins/pr-merge** check does not block PR merges.
- [x] Re-enable **Scan Image** step.

## Checklist


hkeeler:
The non-Docker Ansible setup for this PR is now up at **ansible-consumer-finance#345**.  This PR is dependent on _that_ PR.

thephillipsequation:
>Jenkins multibranch builds update Github commit status. Once this is merged, you will see a new >continuous-integration/jenkins/pr-merge check on each PR. I will make sure this check does not >block PRs from being merged.

@hkeeler why wouldn't we want to enforce this check before merge?

Dockerfile
```diff
@@ -38,27 +40,21 @@ ENV PIP_NO_CACHE_DIR true
 
 # Install python requirements
 COPY requirements requirements
-
-RUN yum -y install gcc && \
-    pip install -r requirements/local.txt -r requirements/deployment.txt && \
-    yum -y remove gcc && \
```
thephillipsequation: I see we were previously installing and uninstalling gcc, why have we decided to leave gcc in the image now?

hkeeler: Good question.  I moved the `gcc` install up with the main `yum install` in an attempt to speed up the build...and it does _a little_.  The install/uninstall was originally under a single `RUN` to try to slim the image size, but the image is already pretty massive, so it's not much of a win on that front.

hkeeler:
> @hkeeler why wouldn't we want to enforce this check before merge?

We may eventually, but this is pretty experimental at this stage.  I don't want to hold up someone's PR because of an issue on the Jenkins side.

hkeeler:
We're still working on getting the standard Jenkins multibranch GitHub integration approved for use.  In the meantime, we've installed the [Disable GitHub Multibranch Status](https://plugins.jenkins.io/disable-github-multibranch-status/) plugin, which will let Jenkins continue to listen for changes, but **not** write back to the GitHub's Commit Status API.  This means for the time being, there won't be a **continuous-integration/jenkins/pr-merge** check on PRs as previously advertised.  If/when it is approved, we'll bring it back.

Dockerfile
```diff
@@ -75,30 +71,41 @@ ENV DJANGO_SETTINGS_MODULE cfgov.settings.production
...
+COPY --chown=apache:apache . .
+
+RUN yum -y install nodejs yarn  && \
+    yum clean all && rm -rf /var/cache/yum && \
+    chown -R apache:apache ${APP_HOME} ${SCL_HTTPD_ROOT}/usr/share/httpd ${SCL_HTTPD_ROOT}/var/run
+
+# Remove files flagged by image vulnerability scanner
+# FIXME: Find out if these packages can be remove altogether
+RUN cd /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/ && \
+    rm -f ndg/httpsclient/test/pki/localhost.key sslserver/certs/development.key
```
willbarton: So, `django-sslserver` probably doesn't need to be installed, as it is only used to add a `runserver` analog `runsslserver` for testing with SSL locally. But it [comes from `local.txt` above](https://github.com/cfpb/cfgov-refresh/pull/5553/files#diff-3254677a7917c6c01f55212f86c57fbfR43), and I'm not sure how we might want to ensure it doesn't get this far. We want to retain it for local development usage.

ndg-httpsclient is a requirement for our Akamai edgegrid-python package and we can't really do much about it, I don't think.

hkeeler: @willbarton, thanks for checking into this.  I don't think we need to change anything now.  I'll just take out the `FIXME` comment for now.  Over time if we start having a bunch of one-off _fixes_ to make the vulnerability scanner happy, we can adjust, perhaps tuning/whitelisting on the scanner instead.

PR #5554, opened 2020-03-06
Fix construct_whitelister_element_rules hook is depricated

cwdavies:
Support both construct_whitelister_element_rules hook in Wagtail 1.13 and register_rich_text_features in Wagtail 2.0
https://docs.wagtail.io/en/stable/releases/2.0.html#construct-whitelister-element-rules-hook-is-deprecated

## Additions

- Support register_rich_text_features hook

## Changes
- Only `span` element with attributes will be supported in Wagtail 2.3

## Testing

1. tox
```
...
Ran 1531 tests in 73.012s

OK (skipped=10)
Destroying test database for alias 'default'...

  lint-py36: commands succeeded
  unittest-py36-dj111-wag113-slow: commands succeeded
  unittest-py36-dj111-wag23-slow: commands succeeded
  congratulations :)
```

## Checklist

### Browsers

- [x] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android


cfgov/v1/wagtail_hooks.py
```diff
@@ -365,21 +366,39 @@ def hide_snippets_menu_item(request, menu_items):
...
+    def blockquote_feature(features):
+
+        # register a feature 'blockquote'
+        # which whitelists the <blockquote> element
+        features.register_converter_rule('editorhtml', 'blockquote', [
+            WhitelistRule('blockquote', allow_without_attributes),
+        ])
+
+        # add 'blockquote' to the default feature set
+        features.default_features.append('blockquote')
```
Scotchester: Per discussion in scrum just now, we realized that the `whitelister_element_rules` _only_ apply Hallo, and since we _probably_ don't want to allow any of these tags (pending investigation) within our lone remaining use of Hallo (table cells), we can probably wipe this whole section for Wagtail 2.

```suggestion
```

cfgov/v1/wagtail_hooks.py
```diff
@@ -365,21 +366,39 @@ def hide_snippets_menu_item(request, menu_items):
...
-# Override list of allowed tags in a RichTextField
-@hooks.register('construct_whitelister_element_rules')
-def whitelister_element_rules():
-    allow_html_class = attribute_rule({
-        'class': True,
-        'itemprop': True,
-        'itemscope': True,
-        'itemtype': True,
-    })
+if wagtail.VERSION <= (1, 14):
```
Scotchester: Since it's a `<=` comparison, this can maybe be slightly clearer by checking for Wagtail 1.13.

```suggestion
if wagtail.VERSION <= (1, 13):
```

cfgov/v1/wagtail_hooks.py
```diff
@@ -365,21 +366,49 @@ def hide_snippets_menu_item(request, menu_items):
...
+# https://docs.wagtail.io/en/stable/releases/
+# 2.0.html#construct-whitelister-element-rules-hook-is-deprecated
+if wagtail.VERSION >= (2, 0):
+    @hooks.register('register_rich_text_features')
+    def register_span_feature(features):
+        allow_html_class = attribute_rule({
+            'class': True,
+            'itemprop': True,
+            'itemscope': True,
+            'itemtype': True,
```
Scotchester: These properties are only for use with schema markup. Because Hallo will only be used in table cells in Wagtail 2+, we should have no need for these attributes, because there's almost no chance we'll use schema markup within a table cell.

```suggestion
```

cfgov/v1/wagtail_hooks.py
```diff
@@ -365,21 +366,49 @@ def hide_snippets_menu_item(request, menu_items):
...
+                        'table', 'tr', 'th', 'td', 'tbody', 'thead', 'tfoot',
+                        'col', 'colgroup']
+
+        return {tag: allow_html_class for tag in allowed_tags}
+
+
+# construct_whitelister_element_rules are depricated in Wagtail 2.0
+# https://docs.wagtail.io/en/stable/releases/
+# 2.0.html#construct-whitelister-element-rules-hook-is-deprecated
+if wagtail.VERSION >= (2, 0):
```
Scotchester: Let's make this an `elif` so it doesn't always check twice, and make the comment more specific.

```suggestion
elif wagtail.VERSION >= (2, 0):
    # The construct_whitelister_element_rules hook was deprecated in Wagtail 2,
    # so we'll use register_rich_text_features instead.
    # Only applies to Hallo editors, which only remain in our custom
    # AtomicTableBlock table cells.
```

cfgov/v1/wagtail_hooks.py
```diff
@@ -365,21 +366,49 @@ def hide_snippets_menu_item(request, menu_items):
...
+        # register a feature 'span'
+        # which whitelists the <span> element
+        features.register_converter_rule('editorhtml', 'span', [
+            WhitelistRule('span', allow_without_attributes),
+        ])
+
+        # add 'span' to the default feature set
+        features.default_features.append('span')
+
+        return {tag: allow_html_class for tag in allowed_tags}
```
Scotchester: This line shouldn't be needed with the new mechanism for registering these features.

```suggestion
```

cfgov/v1/wagtail_hooks.py
```diff
@@ -365,21 +366,49 @@ def hide_snippets_menu_item(request, menu_items):
...
+
+
+# construct_whitelister_element_rules are depricated in Wagtail 2.0
+# https://docs.wagtail.io/en/stable/releases/
+# 2.0.html#construct-whitelister-element-rules-hook-is-deprecated
+if wagtail.VERSION >= (2, 0):
+    @hooks.register('register_rich_text_features')
+    def register_span_feature(features):
+        allow_html_class = attribute_rule({
+            'class': True,
```
Scotchester: Let's be 100% sure that the `id` attribute will be allowed, too:

```suggestion
            'class': True,
            'id': True,
```

cfgov/v1/wagtail_hooks.py
```diff
@@ -365,21 +366,49 @@ def hide_snippets_menu_item(request, menu_items):
...
+            'itemscope': True,
+            'itemtype': True,
+        })
+
+        allowed_tags = ['span']
+
+        # register a feature 'span'
+        # which whitelists the <span> element
+        features.register_converter_rule('editorhtml', 'span', [
+            WhitelistRule('span', allow_without_attributes),
```
Scotchester: We need to use `allow_html_class` here to allow those.

```suggestion
            WhitelistRule('span', allow_html_class),
```

cfgov/v1/tests/test_wagtail_hooks.py
```diff
@@ -206,6 +207,28 @@ def test_tag_param_returns_resources_tagged_that_way(self):
...
+      itemprop="step"
+      itemscope=""
+      itemtype="http://schema.org/HowToSection">
+    <span class="schema-container" itemprop="itemListElement">
+    </span>
+</span>
+        '''
+        output_html = Whitelister.clean(input_html)
+        self.assertHTMLEqual(input_html, output_html)
+
```
Scotchester: Craig, I might be missing something, but this test seems duplicative of the one just below it.

cwdavies: The content of `input_html` is different in `test_whitelist_hooks` depending on `if wagtail.VERSION < (2, 0)` or `if wagtail.VERSION > (2, 0)`.
`input_html` contains only `<span>` for Wagtail 2.x

Scotchester: Ah, of course, now I see! Thanks for explaining.

cfgov/v1/tests/test_wagtail_hooks.py
```diff
@@ -206,6 +207,28 @@ def test_tag_param_returns_resources_tagged_that_way(self):
...
 
+@unittest.skipIf(wagtail.VERSION < (2, 0), "No need to test in Wagtail 1+")
+class TestWhitelistOverride(SimpleTestCase):
+    # Borrowed from https://github.com/wagtail/wagtail/blob/master/wagtail
+    # /core/tests/test_whitelist.py
+
+    def test_whitelist_hooks(self):
+        """Test that Whitelister does not strip scan element and attributes.
+
+        The allowed element <scan> and attributes are added in v1.wagtail_hooks.
```
Scotchester: Two tiny typos here.

```suggestion
        """Test that Whitelister does not strip span element and attributes.

        The allowed element <span> and attributes are added in v1.wagtail_hooks.
```

cwdavies: Good catch. I will replace `scan` with `span` in comments.

PR #5556, opened 2020-03-06
Header JS breakpoint logic cleanup

anselmbradford:
Refactors breakpoint code to remove some unneeded things and share some things.

`get-viewport-dimensions.js` utility is removed and I called `window.innerWidth` in the only place where this was used as this should be supported in our target browsers.

## Removals

- Removes "modern browser" checks in viewport size check and replaces with `window.innerWidth` 

## Changes

- Combines `isInMobile()` and `isInDesktop()` methods into one `isIn(…)` method.
- Replaces `get-viewport-dimensions.js` utility with `window.innerWidth` and removes "modern browser" checks.
- Removes `is` prefixing logic for internal breakpoint keys.
- 

## Testing

2. `gulp clean && gulp build` and responsive behavior of the menu and global search should be as before. Bureau structure page should also have unchanged responsive behavior https://www.consumerfinance.gov/about-us/the-bureau/bureau-structure/


anselmbradford:
@contolini

cfgov/unprocessed/js/modules/util/breakpoint-state.js
```diff
@@ -22,40 +21,58 @@ function _inBreakpointRange( breakpointRange, width ) {
...
+
 /**
- * Whether currently in the desktop view.
- * @returns {boolean} True if in the desktop view, otherwise false.
+ * Checks whether the current breakpoint is in a particular breakpoint group.
+ * @param {string} breakpointGroup - Breakpoint group names.
+ * @returns {boolean} True if in the breakpoint group, otherwise false.
  */
-function isInDesktop() {
+function isIn( breakpointGroup ) {
```
contolini: Harping on semantics again but `isIn` isn't super descriptive and because we're not passing any arguments to the private function `getBreakpoint()` below, this function only tests the current viewport's width. So I'd vote for a name like `viewportIsIn` or `isInBreakpointRange`.

test/unit_tests/js/modules/util/breakpoint-state-spec.js
```diff
@@ -41,29 +47,45 @@ describe( 'breakpoint-state', () => {
...
   } );
 
-  describe( '.isInDesktop()', () => {
+  describe( '.isIn()', () => {
     it( 'should determine whether inside desktop breakpoint threshold', () => {
-      expect( breakpointState.isInDesktop() ).toBe( true );
+      expect( isIn( DESKTOP ) ).toBe( true );
+
+      /* TODO:
+         Mock getBreakpoint() to return a small breakpoint size,
```
contolini: Instead of mocking you could try [this technique](https://github.com/cfpb/cfgov-refresh/blob/fe13404b973d4d8eca996b8fd50ab66f1d1df677/test/unit_tests/js/modules/BreakpointHandler-spec.js#L9-L16).

cfgov/unprocessed/js/modules/util/breakpoint-state.js
anselmbradford: Updated to `viewportIsIn`

test/unit_tests/js/modules/util/breakpoint-state-spec.js
anselmbradford: Works great! Updated in https://github.com/cfpb/cfgov-refresh/pull/5556/commits/ac6359c9536dd5b90d54f0150378127e5667e195

contolini:
LGTM :shipit:

PR #5557, opened 2020-03-07
Correct width of mega menu horizontal rule

contolini:
The mm's horizontal rule should have a fluid width instead of a fixed width to avoid hanging off the side of the viewport. Currently, every page of cf.gov has a long bar on the right side of the screen for widths between 900px and 1260px.

## Testing

1. Pull down and build.
1. Set your browser width to 1000px and compare consumerfinance.gov to localhost.

## Screenshots

Before:

Screen Shot 2020-03-06 at 8 45 03 PM

@anselmbradford

### DIFF
--- a/cfgov/unprocessed/css/organisms/mega-menu.less
+++ b/cfgov/unprocessed/css/organisms/mega-menu.less
@@ -529,9 +529,9 @@ body {
             position: absolute;
             bottom: 0;
             left: -@grid_gutter-width;
+            right: -@grid_gutter-width;
             display: block;
             z-index: 20;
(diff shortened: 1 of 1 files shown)


anselmbradford:
How about this instead? (Fixes left alignment issue too.) Uses calc, which we may have to look at in IE.

Screen Shot 2020-03-09 at 4 36 30 PM

contolini:
@anselmbradford Yeah if that fixes the left alignment issue, let's do it! `calc()` should be fine IMHO as long as it falls back to a width that is reasonable for unsupported browsers.

anselmbradford:
New PR --> https://github.com/cfpb/cfgov-refresh/pull/5561

PR #5561, opened 2020-03-10
Mega menu: correct width of mega menu horizontal rule and responsive sizes

anselmbradford:
Supersedes https://github.com/cfpb/cfgov-refresh/pull/5557

## Changes

- Adjusts width and relative positioning of desktop menu.

## Testing

1. Pull down and build.
3. View at 1261px or greater.
4. Open the menu and check that as it animates down that the horizontal gray rule fully extends to the end of the menu on the left and right side (open one of the center menu items, such as data & research, to make this easier to see). Also note that the gap on the left and right sides of the menu and the edge of the window are the same width.
2. Resize your browser width to 1000px and repeat the above and compare consumerfinance.gov to localhost.

## Screenshots

Screen Shot 2020-03-10 at 11 54 56 AM

Screen Shot 2020-03-10 at 11 55 31 AM

## Todo

We need to break from the xtra-large desktop size to the desktop size earlier as there is a small window where the horizontal scrollbar appears.


anselmbradford:
@contolini

contolini:
Kills two birds and great use of `calc()`! :shipit:

PR #5562, opened 2020-03-10
Add aria labels to mega menu link columns

contolini:
When tabbing through links in the mega menu, screen readers lack the context of the headings above the link columns. This PR adds aria labels to the lists of links.

## Additions

- Aria labels to the unordered lists of links in the mega menu
- Invisible heading "Featured" above the featured links list, if present

## Testing

1. Pull down branch and load the site locally.
1. Command + F5 to enable voiceover.
1. When tabbing to auto loans in the mm, voiceover should state "List, Money Topics" after reading the link text. When tabbing over to Mortgages in the next column, it should say "List, Money Topics Continued".
1. Compare the above step to what happens at https://www.consumerfinance.gov/

## Notes

-

## Todos

-

## Checklist

### Browsers

- [x] Chrome on desktop
- [x] Firefox
- [x] Safari on macOS
- [x] Edge
- [x] Safari on iOS
- [x] Chrome on Android

### Accessibility

- [x] Keyboard friendly
- [x] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

@anselmbradford


cfgov/jinja2/v1/_includes/organisms/header.html
```diff
@@ -102,6 +96,12 @@
             </a>
             {% endif %}
 
+            {% import 'molecules/global-header-cta.html' as global_header_cta with context %}
+            {{ global_header_cta.render( true, language ) }}
+
+            {% import 'molecules/global-search.html' as global_search with context %}
+            {{ global_search.render() }}
+
```
contolini: Unrelated to the mm, I moved the search form and CTA molecule below the logo to fix the tab order of the page. Moving them doesn't affect the header's layout because they're both floated right.

contolini:
> For the menu sections (e.g. Data & Research) that don't have headings is the behavior as expected?

Hmmm you raise a good question. I left the aria labels blank because there are no group headings but IMHO we should use the overview text as labels.

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
```diff
@@ -47,20 +47,28 @@
...
+            aria-label="{{ aria_group_title | safe }}">
             {{ group.group_title | safe }}
         </h4>
         {% endif %}
-        <ul>
+        <ul
+            {% if aria_group_title %}
+            aria-labelledby="{{ aria_group_title | slugify ~ '-menu' }}"
+            {% else %}
+            aria-label="{{ nav_title | safe }}"
```
contolini: @anselmbradford I changed it so that if there's no link group title whatsoever it falls back to the section's nav title (e.g. Practitioner Resources).

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
```diff
@@ -47,18 +47,27 @@
...
+            id="{{ aria_group_title | slugify ~ '-menu' }}"
+            {% if group.title_hidden %} aria-hidden="true" {% endif %}
+            aria-label="{{ aria_group_title }}">
             {{ group.title }}
         </h4>
         {% endif %}
-        <ul>
+        <ul
+            {% if aria_group_title %}
+            aria-labelledby="{{ aria_group_title | slugify ~ '-menu' }}"
```
anselmbradford: We end up this:

Screen Shot 2020-03-13 at 12 45 10 PM

Is that correct? We have a visible item that is labeled by an aria-hidden item. Will screenreaders handle that?

anselmbradford: It seems like `aria-labelledby` maybe shouldn't ever point to the `-continued-` instance, but should point to the one that appears in the first column.

contolini: This is admittedly a little weird. Originally I was printing "Populations Served continued" to the text node and just using `aria-labelledby` for every link group column with no `aria-label`. Buuuuuut then I realized we're using the invisible `h4`s to maintain proper padding above the heading-less columns and by appending "continued" causes a line wrap on smaller viewports. Here's a screenshot with the invisible header revealed in red:

Screen Shot 2020-03-13 at 1 34 42 PM

Notice that "Servicemembers & Veterans" is pushed lower than "Adult Financial Education".

So I left the text node as "Populations Served" and added an `aria-label` with the proper text for screenreaders. It looks dumb but the `aria-labelledby` on the `ul` overrides the `aria-hidden` property.

Buuuuut on second thought we can remove the `aria-hidden` logic because we now never want the heading hidden to screenreaders. Problem solved. 👍

PR #5563, opened 2020-03-10
Mega menu: Remove top margin on ul elements

anselmbradford:
Removes margin added by default `ul` element.

## Changes

- Fixes issue where menu items had too much top margin.

## Testing

1. Pull branch and build. 
2. Compare localhost to production. The mobile links should have equal margin/padding. On desktop, the featured link area should be slightly higher.

## Screenshots

Before:
Screen Shot 2020-03-10 at 1 26 05 PM

After:
Screen Shot 2020-03-10 at 1 26 18 PM


anselmbradford:
@contolini

anselmbradford:
@sonnakim by default this will move the featured link box up slightly. Is this okay? Seems like actually a better alignment.

sonnakim:
@anselmbradford actually, would you mind posting before and after screenshots without any of the inspect dropdowns showing?

anselmbradford:
Before:
Screen Shot 2020-03-10 at 3 32 29 PM

After:
Screen Shot 2020-03-10 at 3 32 40 PM

anselmbradford:
So I realized this puts the box in alignment with the top of the headings, but it means the links in hte fourth column are out of line with the other columns, so not sure which is better?

sonnakim:
The featured link top alignment looks great. :+1:  @anselmbradford and I discussed and realized we can't consistently make 4th column links line up with links in the other columns, because of line wrapping. So, approving this as is.

PR #5581, opened 2020-03-16
Tests: Replace complex tox dependency factors with "current" and "future" factors

willbarton:
This is a proposal to replace our version-specific factors for tox environments with a simpler `current` and `future` nomenclature.

This means to run the unittests on the current production version of Python, Django, and Wagtail, one would use:

```
tox -e unittest-current
```

To run them against upcoming versions of Python, Django, and Wagtail:

```
tox -e unittest-future
```

I've also added a convenience `tox -e unittest` (🎩 @higs4281) that runs identically to `unittest-current`.

I've removed the `fast` and `slow` factors here and environments here; everything will run "slow," i.e. with migrations. As part of this, I've also removed the `NoMigrations` `MIGRATION_MODULE` and its dependencies. The acceptance test env will now run with full migrations (again, 🎩 @higs4281).

The intent here is to simplify the name of our test environments and make the naming version-independent.

## Checklist

### DIFF
--- a/docs/python-unit-tests.md
+++ b/docs/python-unit-tests.md
@@ -60,33 +60,32 @@ You can select specific environments using `-e`.
 Running `tox` by itself is the same as running:
 
 ```sh
-tox -e lint -e unittest-py36-dj111-wag113-slow
+tox -e lint -e unittest
 ```
 
 These default environments are:
 
-- `lint`, which runs our [linting](#linting) tools
-- `unittest-py27-dj111-wag113-slow`, our "slow" (with migrations) environment 
-  for running unit tests on Python 2.7 with Django 1.11 and Wagtail 1.13.
-- `unittest-py36-dj111-wag113-slow`, our "slow" (with migrations) environment 
-  for running unit tests on Python 3.6 with Django 1.11 and Wagtail 1.13.
+- `lint`, which runs our [linting](#linting) tools. We require this 
+  environment to pass in CI.
+- `unittest`, which runs unit tests against the current production 
+  versions of Python, Django, and Wagtail. We require this environment to 
+  pass in CI.
+
+In addition, we also have this environment:
+
+- `unittest-future`, which runs unit tests against upcoming versions of 
+  Python, Django, and Wagtail. We do not require this environment to pass in 
+  CI.
  
 By default this uses a local SQLite database for tests. To override this, you
 can set the `TEST_DATABASE_URL` environment variable to a database connection
 string as supported by [dj-database-url](https://github.com/kennethreitz/dj-database-url).
 
-If you haven't changed any Python dependencies and you don't need to test 
-all migrations, you can run our "fast" environments that skip migrations:
-
-```sh
-tox -e unittest-py36-dj111-wag113-fast
-```
-
 If you would like to run only a specific test, or the tests for a specific app, 
 you can provide a dotted path to the test as the final argument to any of the above calls to `tox`:
 
 ```sh
-tox -e unittest-py36-dj111-wag113-fast regulations3k.tests.test_regdown
+tox -e unittest regulations3k.tests.test_regdown
 ```
 
 ### Linting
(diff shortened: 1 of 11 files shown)


docs/python-unit-tests.md
```diff
@@ -60,33 +60,29 @@ You can select specific environments using `-e`.
 Running `tox` by itself is the same as running:
 
 ```sh
-tox -e lint -e unittest-py36-dj111-wag113-slow
+tox -e lint -e unittest-current
```
higs4281: Could we make this environment plain "unittest" and assume it's current?

willbarton: Absolutely, It'd just involve creating a new `testenv` that uses the `unittest-current` configuration. [That's what `lint` is](https://github.com/cfpb/cfgov-refresh/pull/5581/files#diff-b91f3d5bd63fcd17221b267e851608e8R160-R167)

willbarton: Done!

tox.ini
```diff
@@ -121,43 +108,27 @@ commands=
...
-commands=
-    {env:TRAVIS_CHECK_MIGRATIONS:echo "Skipping Travis-only step"}
+[future]
+# Configuration necessary for upcoming versions of Python, Django, and 
+# Wagtail
+# Note: This is not an env will not run if invoked.
+basepython=python3.6
+deps=
+    Django>=2.2,<2.3
+    wagtail>=2.3,<2.4
```
higs4281: Is it worth noting here that this section will need to be manually changed when we move the chains (unlike the 'current' configs)?

willbarton: Yes, it probably is.

tox.ini
```diff
@@ -121,43 +108,27 @@ commands=
...
-#   tox -e unittest-py{36}-dj{111}-wag{23}-slow
-#
-# To run unit tests.
-setenv=
-    DJANGO_SETTINGS_MODULE=cfgov.settings.test
-
+[current]
+# Configuration necessary for the production versions Python, Django and 
+# Wagtail
+# Note: This is not an env will not run if invoked.
```
chosak: Could you rephrase this comment (and below)? I'm not entirely sure what this means.

willbarton: Certainly! I've also renamed all of these config sections to `-config` to ensure that the fact that they're configuration values rather than testenvs is clearer.

tox.ini
willbarton: I've taken a stab at this in 3f28e5e. What do you think?

higs4281:
If we drop the fast/slow routine, we should be able to drop this file:

- cfgov/cfgov/settings/test_nomigrations.py

... and these code blocks:

- https://github.com/cfpb/cfgov-refresh/blob/91beb6e620372a21e3475a25420ef21e7fc86b34/cfgov/core/utils.py#L174-L186
- https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/test.py#L68-L111

chosak:
> If we drop the fast/slow routine...

@higs4281 that's a great idea, but I think we'll need to keep [this call](https://github.com/cfpb/cfgov-refresh/blob/ed714a7d7f553aac8e308dbcd9bcca6a8d917b24/cfgov/cfgov/test.py#L76-L78) to the initial data script, in some form.

higs4281:
> > If we drop the fast/slow routine...
> 
> @higs4281 that's a great idea, but we might need to keep [this call](https://github.com/cfpb/cfgov-refresh/blob/ed714a7d7f553aac8e308dbcd9bcca6a8d917b24/cfgov/cfgov/test.py#L76-L78) to the initial data script, in some form.

Ah, thank you!

tox.ini
willbarton: And I've fixed to missing "and", which is critical to that sentence.

willbarton:
@higs4281 @chosak right now our [acceptance test settings use the no-migrations settings](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/test_acceptance.py#L1) (and thus the [no-migrations `MIGRATIONS_MODULES`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/test_nomigrations.py#L6)). 

Should that just run with migrations, or should we keep `NoMigrations` for that?

willbarton:
I've pushed a change to remove all the no-migrations options and for the acceptance test settings to run migrations.

chosak:
> Should that just run with migrations, or should we keep NoMigrations for that?

I suppose if we think that running with migrations always is fast enough, then it's better to be consistent than to introduce a difference in how the acceptance tests work.

Scotchester:
The acceptance tests will soon have no option to run them in a way that involves Python.

PR #5585, opened 2020-03-16
Mega Menu: Remove unused classes and add missing ones

anselmbradford:
Unused `o-mega-menu_content-grid__three-col` class, which is no longer needed as there will always be four columns.

## Removals

- Unused `o-mega-menu_content-grid__three-col` class.
- Unneeded `m-list` classes.
- Remove `additional_suffix` argument from `_classes` helper.

## Changes

- Adds missing `o-mega-menu_content-item` class to menu `li` items.

## Testing

1. Pull branch and build
2. Compare to production. The only difference should be that the featured items at mobile get separator lines:

## Screenshots

Before:
Screen Shot 2020-03-16 at 3 52 30 PM

After:
Screen Shot 2020-03-16 at 3 52 25 PM

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
+++ b/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
@@ -55,7 +55,7 @@
         {% set aria_group_title = group.title ~ ' continued' if group.title_hidden else group.title %}
         <h4 class="h5
                   {{ base_class ~ '_group-heading' }}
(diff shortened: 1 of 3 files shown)


anselmbradford:
@contolini

contolini:
I was about to comment on the broken classes but then I pulled your latest commit and saw you beat me to it! Looks good 🎉

PR #4228, opened 2018-06-05
Replace disabled links with headings in mega menu

cfarm:
Fixes https://GHE/CFGOV/platform/issues/2593 and https://GHE/CFGOV/platform/issues/2607

Headings are brought back into the mega menu to add: 
- navigable landmarks to the page for screen reader users
- additional context for sighted mobile users when navigating through the submenus.

## Additions

- a new `<h2>` heading element is added to the mobile menu and is hidden from sighted users, but read aloud by screen readers. This heading is used as an aria label and landmark to describe the purpose of the `<nav>` element. it replaces the previous way we described the `<nav>` element which was another aria attribute. 
- benefits of this change: 
  - screen reader users still hear a landmark/label for nav element, and also now can get to the navigation by navigating through the page headings. 
  - gives a parent heading element to the new `<h3>` headings that I added for overview text, and to the existing `<h4>` elements (see details below)

## Removals

-

## Changes

- Overview links are only link elements if they have a real value, otherwise they default to heading elements (`<h3>`)
- `<h3>` and `<h4>` elements are now nested inside an `<h2>` (see "Additions" above)
- brought back the `<h4>` element for the "Money Topics", "Guides", etc text. Since these bits of text describe the content after them, it makes sense to keep them as headings. See https://GHE/CFGOV/platform/issues/2593#issuecomment-165887 for more detail and research findings on this approach.
- CSS was updated to accommodate the above markup changes while keeping the design the same.
- The only visible design difference: on the mobile mega menu, the overview text for the Consumer Tools and Practitioner Resources menus is now visible as regular text (instead of a disabled link).

## Testing

1. check out branch, run `gulp styles`
1. `./runserver.sh` and visit http://localhost:8000/
1. optional:  navigate through the mega menu at small and large screen sizes with a mouse
1. optional:  navigate using keyboard: tab to move through links

### Testing with desktop voiceover
Most importantly: test using voiceover on desktop. follow instructions in the Web Accessibility Audit, Section B, "How to test" or do the following short version:
  1. cmd+F5 to trigger voiceover
  1. with focus on the webpage (tab into the web content in Chrome or click on the page), use keyboard shortcut *control+option+U* to trigger the Rotor interface
  1. use the right arrow key to get to the "Headings" list in the Rotor
  1.  use the down arrow key to select "2. Main menu" (it should be first in the list of Headings) and hit enter. this will bring you to that element on the page
  1. *control+option+A* to read the mega menu elements in order
[...]

PR #5587, opened 2020-03-17
Mega Menu: Remove h2 heading wrapper around trigger button

anselmbradford:
Seems kind of weird that we wrap the hamburger button in an `h2` and reset the default styles for `h2`. I think this was probably originally added in anticipation of more content being in this part of the menu that didn't show up.

**I also wondered if there's an accessibility reason to have heading content in a `nav`, but it seems to work the same with or without with voiceover. Anyone else know of any reason the heading wrapper may be needed?**

## Removals

- Remove h2 `o-mega-menu_heading` heading wrapper around mobile trigger button.

## Changes

- Adds h2 heading wrapper around top-level menu links.

## Testing

1. Pull branch and build.
2. Visit the mobile version of the site and expand/collapse the hamburger menu. It should work just like production.


anselmbradford:
@contolini

Scotchester:
The heading was added in #4228. I'd take a look at the description of that PR and maybe discuss with @cfarm, but it seems to me like the logic for including it was sound at the time. Possibly some of the structural changes in the new MM change the situation though?

contolini:
Per [W3C](https://www.w3.org/TR/wai-aria-practices/#menubutton) I'm okay removing it. We need to add `role="menu"` to the [wrapper](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/mega-menu.html#L111) though.

anselmbradford:
Are you aware of any resources saying we should include the menu in the headings list in the rotor? I visited a few sites with a hamburger menu (washingtonpost, nytimes, whitehouse, usps) and none of them include the menu in the headings list. It's clearest on the news sites where the headings list is just a list of news article headings.

contolini:
Per [W3C](https://www.w3.org/TR/wai-aria-practices/#menubutton) I'm okay removing it. We need to add `role="menu"` to the [wrapper](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/jinja2/v1/_includes/organisms/mega-menu.html#L111) though.

Scotchester:
> Are you aware of any resources saying we should include the menu in the headings list in the rotor? I visited a few sites with a hamburger menu (washingtonpost, nytimes, whitehouse, usps) and none of them include the menu in the headings list. It's clearest on the news sites where the headings list is just a list of news article headings.

From a quick survey of some search results for "accessible mega menu", it seems like some folks recommend using them (without saying they are _necessary_), while others don't. Seems like a judgement call, and one that we've already made. I'd like Cat to weigh in before reversing that.

[Sidenote: I checked out the homepage's heading structure in the WAVE toolbar and some of the recent changes have introduced some issues that we should probably address.]

cfarm:
I didn't see @anselmbradford address this, which is from [my PR](https://github.com/cfpb/cfgov-refresh/pull/4228) that introduced the h2

> gives a parent heading element to the new <h3> headings that I added for overview text, and to the existing <h4> elements (see details below)

that's the main reason for the change and other decisions were to make sure that the heading hierarchy is sensible which makes it accessible for screen reader users. having heading elements gives screen readers natural landmarks to read aloud to users.

if there are still h3s in this menu then I would not approve this change. that's why the h2 was added, because those were showing up in the headings list on screen readers but the hierarchy made no sense without knowing the context and making the menu navigable from the heading list too.

anselmbradford:
> gives a parent heading element to the new h3 headings that I added for overview text, and to the existing h4 elements

This makes sense, but the h2 is not a parent of the h3s and h4s. It's enclosing the hamburger menu item (aka the "menu trigger"), which is a sibling of the main menu content. It seems like the h2 should be surrounding the top-level overview link, not the menu trigger. So "Consumer Tools" in this image:

Screen Shot 2020-04-20 at 12 16 14 PM

cfgov/unprocessed/css/organisms/mega-menu.less
```diff
@@ -365,6 +365,13 @@ body {
         margin-top: 0;
     }
 
+    // Reset default h2 heading styles.
+    h2 {
+        font-size: 1em;
+        line-height: 22px;
```
Scotchester, line 371: Can we make this unitless?

```suggestion
        line-height: 1.375;
```

PR #5589, opened 2020-03-18
Cards: Add visually hidden heading content to homepage cards

anselmbradford:
WAVE was showing an error about empty headings for the homepage cards. These cards have heading elements that contain only icons, but are in heading content for consistency with the cards on https://www.consumerfinance.gov/ask-cfpb/ . This PR adds textual content to these headings but visually hides it.

## Changes

- Add visually hidden heading content to homepage cards.

## Testing

1. WAVE should show only one error on the homepage, not four.


anselmbradford:
@contolini

PR #5596, opened 2020-03-20
Mega Menu: Remove `o-mega-menu_content-grid` container

anselmbradford:
We had the structure:
```
.o-mega-menu_content-wrapper
  .o-mega-menu_content-grid
```
Which just adds padding. So these containers can be combined. One markup change is that the "back" button was a child of the `-wrapper` container and has been moved to be a sibling.

## Removals

- Removes `o-mega-menu_content-grid` container.

## Testing

1. Pull branch and build.
2. Mega menu layout should be unchanged from production at all screen sizes.


anselmbradford:
@contolini

PR #5605, opened 2020-03-26
Mega Menu: Add menu state helper utililities

anselmbradford:
Adds utilities for creating a hover bar and a focus rectangle for the menu link hover and focus states, which can be fine-tuned independent of each other.

## Changes

- Adds hover state bar in :before pseudo-element.
- Adds focus state rectangle in :after pseudo-element.
- Adds utilities for desktop and mobile link types different states. 

## Testing

1. Pull branch and build.
2. Visting site and interact with menu links across desktop and mobile. 
3. Tab through the menu items and hover at the same time and note appearance.
4. Click an internal page link and note current link appearance, by default and in the hover and focus states.

## Screenshots

menu-states

### DIFF
--- a/cfgov/unprocessed/css/organisms/mega-menu.less
+++ b/cfgov/unprocessed/css/organisms/mega-menu.less
@@ -153,6 +153,201 @@ body {
 */
 
 @menu_max_width_px: 1200px;
+@menu_link_hover_bar_px: 5px;
+
+//
+// ELEMENT STATES
+//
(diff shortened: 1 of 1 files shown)


anselmbradford:
@contolini

anselmbradford:
Featured menu item (the item in the box) doesn't have a "current" appearance. It appears when clicking that it doesn't get a `__current` from the backend, should it?

PR #5613, opened 2020-04-03
Mega Menu: Perform ems conversion of breakpoints

anselmbradford:
The CSS converts the breakpoint values in pixels to ems, which then convert back to greater or less pixels based on deviations of the base font size from 16px. The JavaScript stores the breakpoints in pixels and doesn't change when the font size is customized. This meant that the breakpoints honored in the CSS and JS differed, leading to bugs in transitioning the menu state from the mobile to the desktop menu and vice versa. This PR makes the JS convert its breakpoint values to ems and back to make the same adjustments as the CSS, if the base font size differs from 16px.

## Changes

- Perform em conversation on JavaScript breakpoints when base font size has been customized.

## Testing

1. Menu should function as usual when resizing the window between desktop and mobile.
2. Go into your browser settings and adjust the font size (in Chrome, Chrome > Preferences… > and type "font" in the search and select Customize fonts and change the size).
3. Menu text should be larger (or smaller), but it should break to the mobile menu at some point without being broken open.

## Screenshots

Before:
Screen Shot 2020-04-03 at 12 10 18 PM

After:
Screen Shot 2020-04-03 at 12 10 02 PM

## Notes

- Menu items will likely be stacked at desktop because our page's container uses pixels and not ems and therefore is a consistent width, while the menu is not.

Screen Shot 2020-04-03 at 12 09 52 PM


anselmbradford:
@contolini

cfgov/unprocessed/js/modules/util/breakpoint-state.js
```diff
@@ -4,14 +4,38 @@
...
+                             _getBodyBaseFontSize() === 16 ? false : true;
+  if ( useEmsConversation ) {
+    /* 16 = base font size without adjustments.
+       The CSS converts breakpoints to ems, which then change the width of the
+       pixel width of the breakpoint. In JavaScript, the breakpoints are defined
+       in pixels, so we first convert them to ems using the 16px base font size
+       and then multiply them by any adjustments set by customizations of the
+       font size in the user's browser. */
+    breakpointRangeMin = ( breakpointRangeMin / 16 ) * _getBodyBaseFontSize();
+    breakpointRangeMax = ( breakpointRangeMax / 16 ) * _getBodyBaseFontSize();
```
contolini, line 34: L33 and L34 aren't covered by our tests (`gulp test:unit --specs=js/modules/util/breakpoint-state-spec.js`). It'd be cool if we could mock a non-standard font size in the test file by doing something like `document.body.fontSize = '22px';` but I haven't had any luck figuring it out. I would think the process would be similar to what we do with `windowResizeTo()`?

cfgov/unprocessed/js/modules/util/breakpoint-state.js
```diff
@@ -4,14 +4,38 @@
...
  */
 function _inBreakpointRange( breakpointRange, width ) {
-  const min = breakpointRange.min || 0;
-  const max = breakpointRange.max || Number.POSITIVE_INFINITY;
+  let breakpointRangeMin = breakpointRange.min;
+  let breakpointRangeMax = breakpointRange.max;
+
+  // Whether the user has set a custom size for the font in their browser.
+  const useEmsConversation = isNaN( _getBodyBaseFontSize() ) ||
+                             _getBodyBaseFontSize() === 16 ? false : true;
```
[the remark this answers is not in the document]
anselmbradford: How about this https://github.com/cfpb/cfgov-refresh/pull/5613/commits/d2fde157800cb11448d9ec025326decacfb45868 ?

The reason the NaN is here is that the tests set the `fontSize` to an empty string. We need to use em conversation if the `fontSize` is not 16 or if it doesn't exist. I adjusted it to return -1 if it's empty, but we still need to do two checks—that it is greater than 1 and that it is not 16. See a better way?

cfgov/unprocessed/js/modules/util/breakpoint-state.js
anselmbradford: Okay I reworked the tests in https://github.com/cfpb/cfgov-refresh/pull/5613/commits/b7765a444475da5a9a6fd7ba25a778d49ce0c29d

PR #5614, opened 2020-04-03
Add CampaignPage page type with new JumboHero block

Scotchester:
There is a desire for a page type that has a sidebar-less template with a larger hero and an alternate features box, similar to our homepage design from Spring 2016 – Fall 2019. This commit adds a new `CampaignPage` page type to do that, supported by a new option on the `Hero` molecule to render a "jumbo" hero.

See:
- https://GHE/CFGOV/platform/issues/3573#issuecomment-231301
- https://GHE/CFPB/Consumer-Facing-Iteration/issues/57

## Additions

- New `CampaignPage` page type
- New `is_jumbo` option on `Hero` molecule to, uh, make it jumbo
- New `richtext_isempty` filter to patch the inability to simply test the non-emptiness of Draftail RichTextFields with `if rich_text_field` as before

### Local
1. Pull branch
1. Migrate
1. `yarn gulp styles`
1. Run server
1. Create a new Campaign Page and enter test content

### Docker stack
1. Wait for it to be deployed
1. Visit `/sssu/` on the super secret Docker stack domain
1. Play with it at different screen sizes

## Screenshots

A campaign page laid out with new content for the Start Small, Save Up landing page

## Todos

- [x] Write tests for the `richtext_isempty` filter
- [x] Rework migrations to pull out the `Hero` modification into its own, combine the two `CampaignPage` migrations
- [x] ~~Consider hiding the jumbo option on non-`CampaignPage`s, or validation to prevent it from being selected on any other page type~~ Solved by creating a new `JumboHero` block.
- Consider making the Less enhancements for heroes in the Design System code, if it can be done before the deadline to get the new campaign page live (and if not, do a quick follow-up PR in design-system and update in cfgov-refresh)
- In a separate PR, look for more places to apply the `richtext_isempty` template filter
- ~~In a separate PR, remove the legacy home_hero code that's still hanging around, which I noticed when working on this~~ Taken care of by #5618 

## Checklist

### Browsers

- [x] Chrome on desktop
- [x] Firefox
- [x] Safari on macOS
- [x] Edge
- [x] Safari on iOS
- [x] Chrome on Android

### Accessibility

- [x] Keyboard friendly
- [x] Screen reader friendly

### Other

- [x] Is useable without CSS
- [x] Is useable without JS


cfgov/v1/models/campaign_page.py
```diff
@@ -0,0 +1,53 @@
...
+from wagtail.core.blocks import StreamBlock, StructBlock
+from wagtail.core.fields import StreamField
+from wagtail.core.models import PageManager
+from wagtail.search import index
+
+from v1 import blocks as v1_blocks
+from v1.atomic_elements import molecules, organisms
+from v1.feeds import FilterableFeedPageMixin
+from v1.models.base import CFGOVPage
+from v1.util.filterable_list import FilterableListMixin
```
higs4281, line 13:
```suggestion
from wagtail.core.blocks import StreamBlock
from wagtail.core.fields import StreamField
from wagtail.core.models import PageManager

from v1.atomic_elements import molecules, organisms
from v1.models.base import CFGOVPage
```

I think these can go.

Scotchester: Yeah, I lazily left in a bunch of stuff I copied in from a page model that I used as an example.

willbarton:
I will add, I kinda do like the page-specific specialty blocks being located with their page model. I'm just not sure how that fits in with how we're currently organizing our blocks (perhaps because it's not obvious to me how we're currently organizing our blocks).

Scotchester:
@willbarton That is an excellent question. I thought this might be the first instance of a page-specific block in `v1` (as opposed to the various subapps that implement special blocks), but I found some prior art that I inadvertently emulated: https://github.com/cfpb/cfgov-refresh/blob/090d741d46f63ff9b086b0b77a596882e1095acc/cfgov/v1/models/sublanding_filterable_page.py#L16-L30

I also think I like keeping page-specific blocks with the page model, which I guess is self-evident, since I instinctively implemented it that way :)

cfgov/core/templatetags/richtext.py
```diff
@@ -0,0 +1,23 @@
...
+    for the comparison.
+    """
+
+    return any([
+        value is None,
+        value == '',
+        value == '<p></p>',
+        hasattr(value, 'source') and value.source == '',
+        hasattr(value, 'source') and value.source == '<p></p>',
+    ])
```
higs4281, line 23: I suggested some of this code, but now I think we could avoid some repetition:

```python
blank_values = ['', None, '<p></p>']
return any([
    value in blank_values,
    hasattr(value, 'source') and value.source in blank_values,
])
```

This would also cover the case of value.source being None, Not sure that case exists, but can't hurt to check.

Scotchester: Done in 8c1196a, thanks for the suggestion!

Scotchester:
Thanks to @chosak for the consultation, leading to the refactoring of how the jumbo hero is created in the latest commit.

This is ready for final review.

chosak:
I have one suggestion for an alternate way to handle the template; see https://github.com/chosak/cfgov-refresh/commit/63eb902ab4c84c932247b841d38be0bcbcb6f237. This approach removes the need for a separate jumbo-hero.html file by having Hero and JumboHero use the same hero.html. JumboHero then always sets `is_jumbo` through use of a [custom value class](https://docs.wagtail.io/en/v2.8/topics/streamfield.html#custom-value-class-for-structblock). The advantage of this is that it keeps the frontend module as a single thing, which can be configured in one consistent way. A frontend developer coming to hero.html in the future can see in a single place how our hero module works; we can have one file to keep in sync with the design system; and we keep the ability to (hypothetically now but maybe for real soon) test hero.html and have it test the full range of hero designs.

Scotchester:
You know, I wondered if something like that was possible! I just didn't
look very hard for how to do it. Thanks for pointing it out!

Scotchester:
@chosak Please feel free to push that commit to this branch :)

cfgov/core/templatetags/richtext.py
```diff
@@ -0,0 +1,22 @@
...
+    Draftail RichTextBlocks passed in require accessing the `source` attribute
+    for the comparison.
+    """
+
+    blank_values = [None, '', '<p></p>']
+
+    return any([
+        value in blank_values,
+        hasattr(value, 'source') and value.source in blank_values,
+    ])
```
higs4281, line 22: Sorry to keep coming back to this, but it occurred to me that our rich text is pretty messy, and we might want to catch cases where the rich text just has spaces, such as `'<p>     </p>'` or `'  '`.

If so, this would handle them: 

```suggestion
    blank_values = [None, '', '<p></p>']

    return any([
        value.replace(' ', '') in blank_values,
        hasattr(value, 'source') and value.source.replace(' ', '') in blank_values,
    ])
```

Scotchester: Good call!

higs4281: Sorry for the bum steer (can't call `replace` on None).

Not as clean, but I think this would do if if you want to catch spaces.

```python
    blank_values = ['', '<p></p>']

    return any([
        value is None or value.replace(' ', '') in blank_values,
        hasattr(value, 'source') and value.source is None or value.source.replace(' ', '') in blank_values,  # noqa: 501
    ])
```

cfgov/core/templatetags/richtext.py
```diff
@@ -0,0 +1,37 @@
...
+        return (
+            value.source is None
+            or value.source.replace(' ', '') in blank_values
+        )
+
+    # This is a RichTextField
+    return (
+        value is None
+        or value.replace(' ', '') in blank_values
+    )
```
higs4281, line 37: These comments are helpful.

PR #5617, opened 2020-04-06
Cards: Update design-system to 0.1.x

anselmbradford:
## Changes

- Updates design-system to 0.1.x
- Removes card component less files from cfgov-refresh, which are delivered through cfpb-layout.

## Testing

1. Pull branch and run `./frontend.sh`.
2. Cards on homepage and on http://localhost:8000/ask-cfpb should be unchanged from production.


anselmbradford:
@wpears Maybe we could put some instructions in the wiki or something on how to gracefully update packages? Not sure what I'm doing wrong.

contolini:
> Maybe we could put some instructions in the wiki or something on how to gracefully update packages?

Yes plz. I think most people edit `.yarnrc`, update packages and then reset `.yarnrc`.

anselmbradford:
> Yes plz. I think most people edit .yarnrc, update packages and then reset .yarnrc. I'm not sure what the proper process is.

I use [`yarn-check`](https://www.npmjs.com/package/yarn-check) for the updates, which obviously misses a step.

Scotchester:
> > Maybe we could put some instructions in the wiki or something on how to gracefully update packages?
> 
> Yes plz. I think most people edit `.yarnrc`, update packages and then reset `.yarnrc`.

That should definitely not be necessary, right?

If memory serves, doesn't it come down to needing to use `yard add package` for everything and not manually updating `package.json`?

contolini:
> If memory serves, doesn't it come down to needing to use yard add package for everything and not manually updating package.json?

Ah, cool, I remember it being flaky but I guess that's not the case anymore.

wpears:
I've been testing this a bit, I think the truly foolproof command is:

`yarn cache clean && yarn add w@x.y.z`

I think if you have a package in your global yarn cache you won't get one generated in the yarn offline cache local to the project. This is imo a completely backwards design decision which I still need to validate. If it is indeed true, then I was thinking of just adding a small command to the package.json to smooth this a bit.

wpears:
Just added a few docs in dev tips: https://github.com/cfpb/cfgov-refresh/pull/5632

anselmbradford:
Thanks @wpears !

anselmbradford:
Review plz!

PR #5618, opened 2020-04-06
Homepage: Remove unused component modifications

anselmbradford:
Three component modifiers are unused since the latest refresh of the cf.gov homepage.

## Removals

- Remove unused `home.less` file that contained three modifiers for components on the homepage.
- Remove unused `home-hero` component.

## Testing

1. Pull branch and it should build. 
2. Site should be unchanged.


Scotchester:
Thanks for noticing these! When working on #5614, I also noticed that there is a bunch of `home-hero` stuff laying around still, too. I took a simpler approach to extending the main Hero component, so all of the home hero stuff can go away. Mind removing that in this PR, too?

anselmbradford:
@Scotchester Removed it in https://github.com/cfpb/cfgov-refresh/pull/5618/commits/40e7e2be0fb1992d7fbca92a3b751dca827a3117

PR #5627, opened 2020-04-07
Use Postgres as the default test database

higs4281:
We now use a Postgres-specific Django db field, JSONField, and to test this feature 
we can't use SQLite. So we need to specify our test database as Postgres, 
rather than providing that option.

This adds an uncommented line to our `.env_SAMPLE` file to specify Postgres
for local development and testing.

Documentation is updated to note that we're dropping support for SQLite and
to change setups so that Django ~~can create~~ creates test Postgres databases.

**UPDATE**
We don't need to "specify" Postgres; we can remove options for SQLite and let Django use Postgres by default. So we don't need a `TEST_DATABASE_URL`.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -155,22 +155,14 @@ Then create the database, associated user, and schema for that user:
 
 ```bash
 dropdb --if-exists cfgov && dropuser --if-exists cfpb
-createuser cfpb && createdb -O cfpb cfgov
+createuser --createdb cfpb && createdb -O cfpb cfgov
 psql postgres://cfpb@localhost/cfgov -c 'CREATE SCHEMA cfpb'
 ```
 
-If you absolutely need to use SQLite, you'll need to update your `.env` file
-to comment out the line that specifies Postgres as the db:
+We don't support using an SQLite database, because we use database fields 
+that are specific to Postgres. The `--createdb` flag above allows Django to 
+create temporary Postgres databases when running unit tests.
 
-```bash
-# export DATABASE_URL=postgres://cfpb@localhost/cfgov
-```
-
-And then uncomment the line that tells Django to use SQLite:
-
-```bash
-export DATABASE_URL=sqlite:///db.sqlite3
-```
 
 #### Run the setup script
 
--- a/tox.ini
+++ b/tox.ini
@@ -104,6 +104,7 @@ setenv=
     LC_ALL=en_US.UTF-8
     WAGTAIL_SHARING_HOSTNAME=content.localhost
     DJANGO_SETTINGS_MODULE=cfgov.settings.test
+    TEST_DATABASE_URL={env:TEST_DATABASE_URL:postgres://cfpb:cfpb@localhost/cfgov_test}
 deps=
     -r{toxinidir}/requirements/libraries.txt
     -r{toxinidir}/requirements/test.txt
--- a/tox.ini
+++ b/tox.ini
@@ -95,7 +95,7 @@ commands=
 changedir=
     {toxinidir}/cfgov
 passenv=
-    TRAVIS TRAVIS_JOB_ID TRAVIS_BRANCH TEST_DATABASE_URL TEST_RUNNER
+    TRAVIS TRAVIS_JOB_ID TRAVIS_BRANCH DATABASE_URL TEST_RUNNER
 setenv=
     GOVDELIVERY_ACCOUNT_CODE=fake_account_code
     DJANGO_ADMIN_USERNAME=admin
@@ -104,7 +104,7 @@ setenv=
     LC_ALL=en_US.UTF-8
     WAGTAIL_SHARING_HOSTNAME=content.localhost
     DJANGO_SETTINGS_MODULE=cfgov.settings.test
-    TEST_DATABASE_URL={env:TEST_DATABASE_URL:postgres://cfpb:cfpb@localhost/cfgov_test}
+    DATABASE_URL={env:DATABASE_URL:postgres://cfpb:cfpb@localhost/cfgov}
 deps=
     -r{toxinidir}/requirements/libraries.txt
     -r{toxinidir}/requirements/test.txt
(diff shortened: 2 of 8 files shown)


willbarton:
This doesn't really work as-is for those of us who use Docker and don't have PostgreSQL installed locally. I'm trying to see what the best way to set this up to use the Docker Postgres container might be.

willbarton:
Should we also [remove the SQLite configuration](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/test.py#L11-L17) altogether?

Scotchester:
> Should we also [remove the SQLite configuration](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/cfgov/settings/test.py#L11-L17) altogether?

Seems like it, if we're done with SQLite and there's no going back.

cwdavies:
Should we remove SQLite configuration from teachers-digital-plaform as well?
Otherwise I would need to fix the following:
django.db.utils.NotSupportedError: Renaming the 'teachers_digital_platform_activityspecialpopulation' table while in a transaction is not supported on SQLite < 3.26 because it would break referential integrity. Try adding `atomic = False` to the Migration class.

chosak:
+1 for removing [the default SQLite configuration](https://github.com/cfpb/cfgov-refresh/blob/82961b6679c0defedb2f2d10dcd0bade428e5549/cfgov/cfgov/settings/test.py#L4-L17), and in fact removing that entire configuration section completely.

If you don't define a test database, Django will try to create one next to whatever is configured as the default one. [From the docs:](https://docs.djangoproject.com/en/3.0/topics/testing/overview/#the-test-database)

> The default test database names are created by prepending test_ to the value of each NAME in DATABASES.

If we remove any alteration of `settings.DATABASES` in settings.test, wouldn't this then avoid any need to define anything related to this in .env? Django will just run the tests against whatever database backend you have configured. Then we don't need `TEST_DATABASE_URL` in tox at all and don't have to worry about being in Docker or not.

It might be nice to leave the _ability_ to customize the database for test purposes, should we ever want to do that:

```py
# in settings/test.py
TEST_DATABASE_URL = os.getenv('TEST_DATABASE_URL')
if TEST_DATABASE_URL:
    DATABASES = {'default': dj_database_url.parse(TEST_DATABASE_URL)}
```

higs4281:
> It might be nice to leave the _ability_ to customize the database

I'm not sure how I'd use it, but I see no harm in leaving this ability intact.

higs4281:
@andy I was just tumbling to the fact that we don't need a `TEST_DABASE_URL` now.
But if we leave it in settings, perhaps a commented line in `.env` (or whatever `.env` becomes) would be friendly for showing how to specify a custom test database.

higs4281:
> ... those of us who use Docker and don't have PostgreSQL installed

I propose that we require installing Postgres if you want to test locally.

higs4281:
> Should we remove SQLite configuration from teachers-digital-plaform as well?

I don't think we need to enforce Postgres tests in satellites, if they have test suites that can run stand-alone using SQLite. TDP may not fit that model, since it imports cfgov.

willbarton:
> I propose that we require installing Postgres if you want to test locally.

I'm not sure, but I don't think we required the TDP devs to install Postgres locally when we set them up. I also would prefer not to have to manage a locally running Postgres as well as managing a running Docker. Managing a locally running database server is one of the primary reasons I run Docker (the other is Elasticsearch). There's no reason for me to have both going at the same time.

But, the `tox.ini` and `docker-compose.yml` changes I pushed take care of making it possible to use Docker Postgres for tox. So I'm happy.

tox.ini
```diff
@@ -104,6 +104,7 @@ setenv=
     LC_ALL=en_US.UTF-8
     WAGTAIL_SHARING_HOSTNAME=content.localhost
     DJANGO_SETTINGS_MODULE=cfgov.settings.test
+    TEST_DATABASE_URL={env:TEST_DATABASE_URL:postgres://cfpb:cfpb@localhost/cfgov_test}
```
chosak, line 107:
```suggestion
    TEST_DATABASE_URL={env:TEST_DATABASE_URL:postgres://cfpb:cfpb@localhost/cfgov}
```

Using the `_test` suffix here sets the default Django database to `cfgov_test` but then it creates a test database named `test_cfgov_test`. This could just be `cfgov` which would use a test database name of `test_cfgov`.

More generally, though, would it be better to instead remove this line, add `DATABASE_URL` to `passenv` a few lines up (replacing `TEST_DATABASE_URL` which is already there), and have Travis/GHA just use `DATABASE_URL`? Then we'd be consistent everywhere:

1. For local development, tests run against whatever your `DATABASE_URL` is, consistent with your local server.
2. For Travis/GHA, tests also run against `DATABASE_URL`.

Thinking about this more, this would then remove the need to maintain the `TEST_DATABASE_URL` special case entirely, and perhaps we could totally eliminate that variable. This would make the workflow: if you want to run tests against another database, just set `DATABASE_URL`, e.g. `DATABASE_URL=foo tox`. I'm not sure it's worth keeping the complexity of the separate `TEST_DATABASE_URL` in this case.

willbarton: Just getting caught up... and I agree, I like your ideas to simplify!

willbarton:
One thing with running tests against Postgres that @cwdavies noticed is that we can't run `tox` simultaneously with the same database.

tox.ini
higs4281: @chosak @willbarton I've removed references to `TEST_DATABASE_URL` so that we rely on `DATABASE_URL` for everything.

higs4281:
> parallel

Hmm. `tox --parallel auto` runs for me, but the tests actually take *longer*, and there is no progress reporting .

Maybe we should find out how much use `parallel` mode was getting?

willbarton:
@higs4281 I actually meant two separate `tox` runs in two shells.

docs/installation.md
```diff
@@ -155,22 +155,14 @@ Then create the database, associated user, and schema for that user:
 
 ```bash
 dropdb --if-exists cfgov && dropuser --if-exists cfpb
-createuser cfpb && createdb -O cfpb cfgov
+createuser --createdb cfpb && createdb -O cfpb cfgov
```
willbarton, line 158: If we were to set a password for the "cfpb" user, then both local Postgres and Docker Postgres will have the same URL (Docker currently has `cfpb` as the password, local has no password).

PR #5628, opened 2020-04-07
CCDB Landing page: IE11 remove node patch

flacoman91:
Fixing an ie11 bug.  Object doesn't support .remove()


cfgov/unprocessed/apps/ccdb-landing-map/js/TileMap.js
```diff
@@ -204,6 +204,7 @@ export function tooltipFormatter() {
  * Draw a legend on a chart.
  * @param {Object} chart A highchart chart.
  */
+// eslint-disable-next-line max-lines-per-function, require-jsdoc
```
JeffreyMFarley, line 207: :chef's kiss:

PR #5632, opened 2020-04-09
Note the usage of yarn add for new deps

wpears:
Adds some notes in the docs about the preferred js package installation command and various quirks.


docs/development-tips.md
```diff
@@ -54,3 +54,13 @@ The structure looks like this:
 !!! note
     After running `gulp build` the site's assets are copied over to `cfgov\static_built`,
     ready to be served by Django.
+
+### TIP: Installing new frontend dependencies
+- Use `yarn add new_dep@se.m.ver` to install new dependencies.
+- If you can't do this for some reason or are looking to freshen all dependencies,
+  you will need to edit `.yarnrc`, commenting out the two included flags before
```
anselmbradford, line 61: Do we want to be more explicit here? 

```
…, temporarily commenting out the `--install.pure-lockfile true` and `--install.offline true` flags before
```

docs/development-tips.md
```diff
@@ -54,3 +54,13 @@ The structure looks like this:
...
+  proceeding with your installation or update.
+- In the rare but observed case that `yarn add new_dep@se.m.ver` doesn't add
+  every needed package to the offline cache, you likely need to first run
```
anselmbradford, line 64: And someone opening a PR would know this was the case because github actions would fail on installing dependencies? In https://github.com/cfpb/cfgov-refresh/pull/5617 I ran the updates and built the project and all seemed good till it go to the PR and you fixed it. Is there something I would see locally before running checks on the PR that indicates everything didn't install?

wpears: Following the `yarn add` workflow, that situation shouldn't be encountered, but you could certainly test locally by running `rm -rf node_modules && yarn cache clean && yarn`

wpears: heh pushed a solution to this in b118cf2

package.json
```diff
@@ -98,7 +98,7 @@
   "scripts": {
     "enforce": "enforce-node-version",
     "preinstall": "node scripts/preinstall.js",
-    "postinstall": "yarn run enforce && node scripts/yarn/apps-install",
+    "postinstall": "yarn run enforce && node scripts/yarn/apps-install && ./scripts/check-npm-cache.sh",
```
anselmbradford, line 101: Should we be consistent with the paths here? So one of these:

`./scripts/yarn/apps-install && ./scripts/check-npm-cache.sh`
`./scripts/yarn/apps-install/index.sh && ./scripts/check-npm-cache.sh`

PR #5636, opened 2020-04-09
Improve print styling and strip current path from in-page anchor links

Scotchester:
Addressing https://JIRA/browse/INTAKE-4238

## Changes

- When printing a cf.gov page:
  - Link `href`s are now output after the link text
  - In-page anchor links are unstyled
  - Several unhelpful-in-print elements are no longer displayed:
    - Footer navigation links
    - Email signup and RSS subscription blocks
    - Video player play buttons
  - Video players will print with "Please visit this webpage to view the above video." appended
- Python link parsing middleware updated to strip the current path from in-page anchor links. See the commit message on c001112 for details.

## Removals

- Unused data-qa-hooks were removed from the RSS subscription block template (along with the one browser test shared object file that referenced them, but was never used in any actual 
tests)

## Testing

1. Pull branch
1. `gulp styles`
1. Load http://localhost:8000/about-us/blog/guide-coronavirus-mortgage-relief-options/
1. Print preview (or add `.print` to the `html` element and view in a small screen layout)
1. Verify the changes described above

## Checklist

### Browsers

- [x] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Internet Explorer 9 and 11


cfgov/unprocessed/css/print.less
```diff
@@ -28,25 +28,38 @@
     - cfgov-print
 */
 
+.respond-to-print({
```
wpears, line 31: Have you considered adding in any more print-appropriate base styles on body? eg `background: none` & `color: black` (which would also apply to the link color on line 48).

Also curious about your thoughts on `.o-header`, `.o-feedback`, `.content_sidebar`, and `.related-questions` being `display: none`d on print

cfgov/core/utils.py
```diff
@@ -112,6 +117,13 @@ def add_link_markup(tag):
 
     class_attrs = tag.attrs.setdefault('class', [])
 
+    if request_path is not None:
```
willbarton, line 120: Non-blocking, but I think I'd prefer comment above this line with the TODO from the docstring, just so it's in context.

niqjohnson:
Also, I chatted with @stephanieosan about this. It sounds like El Camino might be taking on print styles longer term, so I wanted to tag you, @stephanieosan, in case you or anyone from your team wanted to take a look at these initial improvements.

cfgov/unprocessed/css/organisms/footer.less
```diff
@@ -202,6 +209,10 @@
                 padding-right: @grid_gutter-width;
             }
         } );
+
+        .respond-to-print({
```
anselmbradford, line 213: Should we aggregate all the print styles at the bottom instead of sprinkling them throughout?

ielerol:
@niqjohnson Super Regular is also going to be doing some print style work as part of the reports-as-webpages projects, so we can create nice-looking PDF versions of the reports. We'll be focused on the needs of reports specifically, but improving the print typography is definitely on the agenda and should for the most part be applicable across the site. We could probably do some incremental improvements on how printed URLs are handled as well, like the line breaks and external links. 

I'm not sure when El Camino might be picking up print style work, but we can coordinate with them too.

cfgov/unprocessed/css/print.less
```diff
@@ -28,25 +28,38 @@
...
+    }
+
+    .o-video-player_play-btn {
+        display: none !important;
+    }
+
+    .o-video-player::after {
+        // Output href in parentheses for links with a protocol already there
+        content: " Please visit this webpage to view the above video.";
+    }
```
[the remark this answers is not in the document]
Scotchester: Brilliant suggestion!

cfgov/unprocessed/css/print.less
```diff
@@ -28,25 +28,37 @@
     - cfgov-print
 */
 
+.respond-to-print({
+    a::after {
```
anselmbradford, line 32: Are we moving to using double colons now? Should we do this everywhere?

cfgov/unprocessed/css/print.less
```diff
@@ -28,25 +28,37 @@
...
+    }
+
+    a[href^="#"] {
+        // Style in-page anchor links (like page TOCs) to not look like links
+        color: @dark-gray;
+        border-bottom-width: 0;
+    }
+
+    .o-video-player_play-btn {
+        display: none !important;
```
anselmbradford, line 53: It would be helpful to add a code comment above any `!important` to note why it's required.

schaferjh:
I agree that this does not need to turn into a massive effort and it's great to see these improvements. A lot of what @niqjohnson mentions would be good to examine in the future. However, I am curious if we might be able to axe the topic tags and "join the conversation" paragraph within this PR? They are both unique to viewing the content on a website.

cfgov/unprocessed/css/print.less
Scotchester: These are good questions, but ones that I think we should save for later, like some of the others that were mentioned later in the discussion.

cfgov/core/utils.py
Scotchester: Good call. This was harder to do on an earlier iteration where the added code was not a single hunk of stuff, and I didn't think about it after I refactored.

Scotchester: Done in 249daf180177650e822290053d4cce0c34824002

cfgov/unprocessed/css/organisms/footer.less
Scotchester: Good question. I was following the lead of the other `respond-to-` media query mixins in the file, which seems like the right path for now. If we want to refactor all of those mixins later, I'm open to it.

cfgov/unprocessed/css/print.less
Scotchester: I admit to never feeling like I fully understood pseudostuff, but I think it's used here because it's creating a pseudo_element_ within the `a`, as opposed to something like `:hover` or `:focus` that is a pseudo_class_ on the element?

I didn't actually give it any thought; the example code I found used the double colon.

anselmbradford: It just comes down to if we want IE8 to recognize this style or not. If we're past that, this is fine, if not, it should be `a:after`.

anselmbradford: Though I guess that would only matter for a print preview, since this is print styles, so double is probably the way to go.

cfgov/unprocessed/css/print.less
Scotchester: :+1: Done in f63b2f9351845a02dda5c6db5ee609c02be8a1db

cfgov/unprocessed/css/print.less
Scotchester: Ah, I still have some browser testing to do, but forgot that IE8 wouldn't recognize it. Probably worth changing, then.

Scotchester: Done in 831968d

cfgov/unprocessed/css/organisms/footer.less
anselmbradford: For what it's worth, in [mega-menu.less](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/css/organisms/mega-menu.less#L709) the queries are aggregated together in distinct sections of the file and I find it a lot easier to work with. You can look at one screen size and go to one section of the less file to make edits in most cases. The disadvantage is if a selector is deleted, it may need to be deleted in each media query block. However, that's worth it for the one-to-one relationship of one code block to one media state, in my opinion.

Scotchester: Yeah, interesting. Seems like it'd be a good thing to discuss during a FEWD bi-weekly.

Scotchester:
@schaferjh Updated with your request!

Once this passes its checks, we'll move forward with merging and deploying this today. I'll compile the list of other ideas that have come up here and put them in an issue somewhere for further consideration. Nicholas and I will also be using some of our Weekender time today to hold a discussion about print styles.

PR #5638, opened 2020-04-10
CCDB Landing Map improvements

flacoman91:
We were missing some of the latest work from CCDB-UI, so this PR migrates:
- Tile map binning work from ccdb-ui
- Responsive styling work


flacoman91:
closing this. I need to fix this PR.  IE11 is not playing nice with any of these changes

PR #5640, opened 2020-04-10
Add Django 2.1 support to urls

cwdavies:
In Django 2.1, it appears that (?i) placed anywhere inside the URL regex will cause it to ignore case. However, reverse() fails unless the (?i) follows the $ at the end of the regex.

## Additions

Change urls.py to use:
* auth_views.LogoutView
* auth_views.PasswordChangeView
* auth_views.PasswordChangeDoneView
to fix the following:
* AttributeError: module 'django.contrib.auth.views' has no attribute 'logout'
* AttributeError: module 'django.contrib.auth.views' has no attribute 'password_change'
* AttributeError: module 'django.contrib.auth.views' has no attribute 'password_change_done'
* ValueError: Non-reversible reg-exp portion: '(?i' in Django 2.1

## Testing

1. tox -e unittest-future

## Checklist

- [x] Code follows the standards laid out in the [CFPB development guidelines(https://github.com/cfpb/development)


cfgov/cfgov/urls.py
```diff
@@ -371,25 +371,25 @@ def empty_200_response(request, *args, **kwargs):
...
         ask_autocomplete, name='ask-autocomplete-en'),
     re_path(r'^(?P<language>es)/obtener-respuestas/api/autocomplete/$',
         ask_autocomplete, name='ask-autocomplete-es'),
 
     re_path(r'^_status/', include_if_app_enabled('watchman', 'watchman.urls')),
 
     re_path(
-        r'^(?i)consumer-tools/financial-well-being/',
+        r'^consumer-tools/financial-well-being/',
         include('wellbeing.urls')
```
willbarton, line 389: We should ensure that `wellbeing/urls.py` has similar changes so that those URLs remain case-insensitive.

cfgov/cfgov/urls.py
```diff
@@ -371,25 +371,25 @@ def empty_200_response(request, *args, **kwargs):
...
     ),
     re_path(
-        r'^(?i)about-us/diversity-and-inclusion/',
+        r'^about-us/diversity-and-inclusion/',
         include((
             'diversity_inclusion.urls',
             'diversity_inclusion'),
```
willbarton, line 395: We should ensure that `diversity_inclusion/urls.py` has similar changes so that those URLs remain case-insensitive.

cfgov/cfgov/urls.py
cwdavies: Added in PR https://github.com/cfpb/cfgov-refresh/pull/5649

cfgov/cfgov/urls.py
cwdavies: Also addressed in PR https://github.com/cfpb/cfgov-refresh/pull/5649

cwdavies:
> I think this makes sense, but I wonder, is this documented anywhere that `(?i)` breaks `reverse` if it's at the beginning of the RE? Is there a test or a check we could add for this?
> 
> Clearly this is a pattern we use a lot, so I think (a) we'd want to socialize that this is a thing now, and (b) we want to make sure no leading-`^(?i)` regexes are created before we are testing against Django 2.1 (which it sounds like is where this breaks)?

I found some Django documentation about `(?i)` here:
https://docs.djangoproject.com/en/3.0/releases/2.1/#features-removed-in-2-1

I believe when you run tox using Django 2.1 or 2.2 there will be tests that fail as `(?i)` breaks `reverse`

willbarton:
@cwdavies Alrighty. Clicking through to the Django 1.11 deprecations, there's this:

> Using regular expression groups with iLmsu# in url() is deprecated. The only group that’s useful is (?i) for case-insensitive URLs, however, case-insensitive URLs aren’t a good practice because they create multiple entries for search engines, for example. An alternative solution could be to create a handler404 that looks for uppercase characters in the URL and redirects to a lowercase equivalent.

It sounds like this might not be a bad idea generally, to replace our use of `(?i)` in our URL patterns and use a 404 handler to lower-case URLs. This is probably out of scope for our Django upgrade work, but it would inform the other discussions we're having around URL handling.

As for tests — my concern is that if any new URLs get added *before* we have Django 2.1 support, we're going to have to edit them again.

PR #5649, opened 2020-04-13
Fix wellbeing and diversity_inclusion so URLs are case-insensitive

cwdavies:
Add (?i) to wellbeing and diversity_inclusion so URLs are case-insensitive

## Checklist

### DIFF
--- a/cfgov/cfgov/urls.py
+++ b/cfgov/cfgov/urls.py
@@ -385,11 +385,11 @@ urlpatterns = [
     re_path(r'^_status/', include_if_app_enabled('watchman', 'watchman.urls')),
 
     re_path(
-        r'^consumer-tools/financial-well-being/',
+        r'^(?i)consumer-tools/financial-well-being/',
         include('wellbeing.urls')
     ),
     re_path(
-        r'^about-us/diversity-and-inclusion/',
+        r'^(?i)about-us/diversity-and-inclusion/',
         include((
             'diversity_inclusion.urls',
             'diversity_inclusion'),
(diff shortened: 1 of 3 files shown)


cfgov/cfgov/urls.py
```diff
@@ -385,11 +385,11 @@ def empty_200_response(request, *args, **kwargs):
     re_path(r'^_status/', include_if_app_enabled('watchman', 'watchman.urls')),
 
     re_path(
-        r'^consumer-tools/financial-well-being/',
+        r'^(?i)consumer-tools/financial-well-being/',
```
willbarton, line 388: Wait, but don't these need to be at the end of the line? My impression from #5640 was that we'd need to add this to the individual URL patterns in [`wellbeing/urls.py`](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/wellbeing/urls.py)?

cfgov/cfgov/urls.py
```diff
@@ -385,11 +385,11 @@ def empty_200_response(request, *args, **kwargs):
...
         include('wellbeing.urls')
     ),
     re_path(
-        r'^about-us/diversity-and-inclusion/',
+        r'^(?i)about-us/diversity-and-inclusion/',
```
willbarton, line 392: Same as above, doesn't this need to be added to the patterns in [`diversity_inclusion/urls.py `](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/diversity_inclusion/urls.py) instead of here?

cwdavies: Agreed and fixed

cfgov/cfgov/urls.py
cwdavies: Agreed and fixed

PR #5666, opened 2020-04-16
Bump node engine version

wpears:
Bumps minimum node version to 10.x (which is minimum LTS).


anselmbradford:
Prolly should do this for [all the app asset packages](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/unprocessed/apps/) too.

PR #5671, opened 2020-04-17
Adds REFRESH_DB param and updates refresh-data script to add conditional

thephillipsequation:
cfgov-refresh's docker-stack.yml currently runs the refresh-data.sh every time. This is unnecessary considering the postgres service does not redeploy when just app code changes.

This causes a couple issues:

Data modifications are lost between deploys
It makes the deploys slow. It is an expensive operation to refresh the database, pulling down the big DB archive and applying it to the running container.

## Additions

- REFRESH_DB parameter in the Jenkinsfile that defaults to false
- Conditional logic added to the refresh-data.sh script

## Removals

-

## Changes

-

## Testing

Run an adhoc deploy with refresh db = true
Check for django wagtail table in postgres (wagtailcore_page suggested)
Run an adhoc deploy with refresh db = false
Check for django wagtail table in postgres (wagtailcore_page suggested)
## Notes

-

## Todos

-

## Checklist

### Browsers

- [ ] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


docker-stack.yml
```diff
@@ -65,7 +66,18 @@ services:
...
+            if django-admin inspectdb wagtailcore_page | grep \"db_table = \'wagtailcore_page\'\" && $REFRESH_DB == false;
+              then 
+                echo 'Skipping DB refresh as db already exists'
+                httpd -d cfgov/apache -D FOREGROUND
+            elif $REFRESH_DB == true;
+              then
+                ./refresh-data.sh && httpd -d cfgov/apache -D FOREGROUND
+            else
+                ./refresh-data.sh && httpd -d cfgov/apache -D FOREGROUND
+            fi "
```
higs4281, line 80: If both elif and else result in the same action, couldn't you just use `else` ?

```suggestion
            else
                ./refresh-data.sh && httpd -d cfgov/apache -D FOREGROUND
            fi "
```

Also, I wonder if using psql could be a simpler test of database existence:

```
psql -l | grep cfgov
```

thephillipsequation: modified the logic a bit to make the conditional more explicit

refresh-data.sh
```diff
@@ -8,6 +8,17 @@
...
 
+if django-admin inspectdb wagtailcore_page | grep "db_table = 'wagtailcore_page'" && $REFRESH_DB == false;
+    then 
+        echo 'Skipping DB refresh as db already exists'
+        exit 1
+    elif $REFRESH_DB == true;
+        then
+            continue
+    else
+        continue
```
willbarton, line 19: Both macOS's `bash` and `bash` in the `python` container complain about this test the use of `continue`:

```
./refresh-data.sh: line 15: ==: command not found
./refresh-data.sh: line 19: continue: only meaningful in a `for', `while', or `until' loop
```

This also seems superfluous, as if the original test for the database falls through, the script will continue regardless.

```suggestion
```

willbarton:
s> This is unnecessary considering the postgres service does not redeploy when just app code changes.

I have one other thought — it might be worth having the `./refresh-data.sh` script run migrations even if set not to refresh the database. I think there are pros and cons to doing that automatically though. I'd be curious what @chosak thinks.

chosak:
@willbarton it seems like your comment is out of date since the changes to refresh-data.sh were removed in the latest commits here? Regardless, I vote that a script named refresh-data.sh should, indeed, refresh data. 😄

refresh-data.sh
```diff
@@ -8,6 +8,7 @@
 
 set -e
 
+
```
chosak, line 11:
```suggestion
```

thephillipsequation: removed the line :)

docker-compose.prod.yml
```diff
@@ -24,4 +24,4 @@ services:
             - ./cfgov/apache/conf/:/src/cfgov-refresh/cfgov/apache/conf/
             - ./cfgov/apache/conf.d/:/src/cfgov-refresh/cfgov/apache/conf.d/
             - ./cfgov/apache/conf.modules.d/:/src/cfgov-refresh/cfgov/apache/conf.modules.d/
-        
+
```
Scotchester, line 27: Mind reverting the changes to this file for a cleaner history?

thephillipsequation: sure I can do that

Dockerfile
```diff
@@ -71,16 +71,17 @@ ENV DJANGO_SETTINGS_MODULE cfgov.settings.production
 ENV DJANGO_STATIC_ROOT ${STATIC_PATH}
 ENV ALLOWED_HOSTS '["*"]'
 
+
+# See .dockerignore for details on which files are included
+COPY --chown=apache:apache . .
+
```
Scotchester, line 77: This change seems unrelated to the purpose of this PR? Doesn't seem like it would be harmful, though.

docker-stack.yml
```diff
@@ -65,7 +66,20 @@ services:
...
         - com.docker.lb.network=${LB_NETWORK}
-    command: /bin/bash -c './refresh-data.sh && httpd -d cfgov/apache -D FOREGROUND'
+    command: >
+                bash -c "
+                    
+                psql $${DATABASE_URL} -c 'SELECT COUNT(*) FROM auth_user'
+                
+                django_tables_exist=$$?
+                
+                echo $$django_tables_exist
```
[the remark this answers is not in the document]
thephillipsequation: removed

Dockerfile
thephillipsequation: sorry this leaked in from a bad commit from a separate branch, I've reverted it!

PR #5672, opened 2020-04-20
Add schema_json field to CFGOVPage for structured data

Scotchester:
This field can be used to store structured data about a page, which is then output in the `<head>`, between `<script type="application/ld+json">` and `</script>` tags.

Such structured data can be used by search engines to provide "rich" 
results when these pages are returned by a search.

Learn more at: https://developers.google.com/search/docs/guides/intro-structured-data

Part of addressing https://JIRA/browse/INTAKE-4248

## Additions

- `schema_json` field added to `CFGOVPage` model with output in the `<head>` of a page

## Testing

1. Pull branch
1. Visit any page in the browser, view the source, and verify that you cannot find `<script type="application/ld+json">`
1. Open up the page in the Admin and go to its Configuration tab
1. Enter valid JSON into the Schema JSON field and publish the page
1. Reload the rendered page and look for the `<script type="application/ld+json">` in the source, with the JSON you entered

## Notes

- When the field is empty, the word `null` appears in it. This doesn't get output on the page, but it's still somewhat weird. Any advice would be welcome.
- Also hoping to get some help from the Wagtail Slack on retaining line breaks that are entered into the field.

## Todos

- [x] Add tests for new context variable

## Checklist


cfgov/v1/models/base.py
```diff
@@ -220,6 +240,9 @@ def get_context(self, request, *args, **kwargs):
                 path=Value(request.path, output_field=models.CharField())) \
             .filter(path__regex=F('url_pattern'))
 
+        if self.schema_json:
```
willbarton, line 243: image
You'll probably get a test coverage failure because this conditional isn't covered.

cfgov/v1/models/base.py
```diff
@@ -220,6 +240,9 @@ def get_context(self, request, *args, **kwargs):
...
+            context['schema_json'] = mark_safe(json.dumps(self.schema_json))
```
chosak, line 244: Isn't there a potential security risk if someone were to enter a closing script tag into this field?

```
</script><script>my malicious code</script><script>
```

Why does this need to be `mark_safe`? Can it contain HTML characters? Could [one of these](https://docs.djangoproject.com/en/3.0/ref/utils/#module-django.utils.html) be used instead?

cfgov/v1/models/base.py
Scotchester: > # Todos
> - [ ] Add tests for new context variable

😜

cfgov/v1/models/base.py
Scotchester: > Isn't there a potential security risk if someone were to enter a closing script tag into this field?

Good question! I thought about that, but JSONField validates that only valid JSON be entered, so I don't think it'd be possible to exploit this.

> Why does this need to be `mark_safe`? Can it contain HTML characters? Could [one of these](https://docs.djangoproject.com/en/3.0/ref/utils/#module-django.utils.html) be used instead?

I used `mark_safe` to prevent the rendering of the quotation marks encoded as `&#39;` HTML character entities. It doesn't look like any of the functions at the link you referenced would perform that same function. I again think it should be pretty safe to use, since the field validates that it be valid JSON, but if you can think of a reason why it might not be safe, I suppose we could just do a find-and-replace filter.

cfgov/v1/models/base.py
```diff
@@ -77,6 +81,21 @@ class CFGOVPage(Page):
             'Maximum size: 4096w x 4096h.'
         )
     )
+    schema_json = JSONField(
+        null=True,
+        blank=True,
+        verbose_name='Schema JSON',
+        help_text=mark_safe(
```
chosak, line 88: Instead of adding this lengthy help text and having it get squished in the sidebar, you could use a MultiFieldPanel with a HelpPanel (if a bit too spaced out):

image

cfgov/v1/models/base.py
chosak: Here's a hack that works: `{"hack": "</script><script>alert('hacked!')</script>"}`. This is valid JSON but also triggers the script if entered into the field.

cfgov/v1/models/base.py
chosak, line 88: Instead of adding this lengthy help text and having it get squished in the sidebar, you could use a MultiFieldPanel with a HelpPanel (if a bit too spaced out):

image

cfgov/v1/models/base.py
Scotchester: Well, shoot. Guess I'll try the find and replace method.

Scotchester: In another case of just not looking hard enough until my smarter colleague gently nudges me harder… https://jinja.palletsprojects.com/en/2.11.x/templates/#tojson

cfgov/v1/models/base.py
Scotchester: Eh, I don't mind calling less attention to the field. It should be avoided most of the time.

PR #5677, opened 2020-04-22
Override new .u-hidden class on sidenav and read-more "expandables"

Scotchester:
cfpb-expandables 0.1.0 added the `.u-hidden` utility class to expandable 
content on init, in order to prevent focus from entering collapsed 
expandables. This was inadvertently hiding the content that should 
always be visible at certain screen sizes on these two special types of 
"expandables", secondary navigation and read-more.

## Additions

- `display: block !important` overrides added to sidenav and read-more `o-expandable_content` at the screen sizes where they should always been open

## Checklist

### Browsers

- [x] Chrome on desktop
- [ ] Firefox
- [ ] Safari on macOS
- [ ] Edge
- [ ] Safari on iOS
- [ ] Chrome on Android

### DIFF
--- a/cfgov/unprocessed/css/organisms/read-more.less
+++ b/cfgov/unprocessed/css/organisms/read-more.less
@@ -43,19 +43,16 @@
 */
 
 .o-expandable__read-more-mobile {
-    .o-expandable_target__collapsed {
+    .o-expandable_target__collapsed,
+    .o-expandable_content__collapsed:after {
         display: none;
     }
 
-    .o-expandable_content__collapsed {
+    .o-expandable_content {
         // These two !important values override basic expandable styling,
         // because these don't function like expandables on sm+ screens.
         display: block !important;
         max-height: 100% !important;
-
-        &:after {
-            display: none;
-        }
     }
 
     @media only screen and (max-width: @bp-xs-max) {
--- a/cfgov/unprocessed/css/organisms/read-more.less
+++ b/cfgov/unprocessed/css/organisms/read-more.less
@@ -48,6 +48,9 @@
     }
 
     .o-expandable_content__collapsed {
+        // These two !important values override basic expandable styling,
+        // because these don't function like expandables on sm+ screens.
+        display: block !important;
         max-height: 100% !important;
 
         &:after {
@@ -55,7 +58,7 @@
         }
     }
 
-    @media only screen and (max-width:@bp-xs-max) {
+    @media only screen and (max-width: @bp-xs-max) {
         .o-expandable_target__collapsed {
             display: block;
 
--- a/cfgov/unprocessed/css/organisms/secondary-navigation.less
+++ b/cfgov/unprocessed/css/organisms/secondary-navigation.less
@@ -94,6 +94,9 @@
         }
 
         .o-expandable_content {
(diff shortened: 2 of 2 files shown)


cfgov/unprocessed/css/organisms/secondary-navigation.less
```diff
@@ -94,6 +94,9 @@
         }
 
         .o-expandable_content {
+            // These two !important values override basic expandable styling,
+            // because these don't function like expandables on med+ screens.
+            display: block !important;
```
anselmbradford, line 99: Maybe instead of this we should add a separate block so this and the "read more" edit are consistent:

```
.o-expandable_content__collapsed {
            // These two !important values override basic expandable styling,
            // because these don't function like expandables on med+ screens.
            display: block !important;
}

.o-expandable_content {
            max-height: 100% !important;
            padding: 0;
…
}
```

Scotchester: I'm not sure I follow. Did your code block get messed up?

anselmbradford: Like create a new block that targets the `.o-expandable_content__collapsed`, not `o-expandable_content`. That will still target the same element on the page, but that way both of the edits you have in this PR will be overriding the same class.

Scotchester: Ah! Sounds good.

Scotchester: Also realizing now that it needs to be targeted at `@bp-sm-min` to avoid _also_ taking precedence on mobile devices.

Scotchester: Updated in c00bfab.

cfgov/unprocessed/css/organisms/read-more.less
```diff
@@ -43,19 +43,19 @@
...
+    .o-expandable_content__collapsed:after {
         display: none;
     }
 
-    .o-expandable_content__collapsed {
+    .o-expandable_content {
+        // These two !important values override basic expandable styling,
+        // because these don't function like expandables on sm+ screens.
+        display: block !important;
         max-height: 100% !important;
```
anselmbradford, line 55: Is it an issue that this wasn't moved, like the `:after` was?

cfgov/unprocessed/css/organisms/secondary-navigation.less
anselmbradford: You're targeting `o-expandable_content` instead of `o-expandable_content__eollapsed` in the override, because the [JS targets the content class](https://github.com/cfpb/design-system/blob/master/packages/cfpb-expandables/src/Expandable.js#L57). That makes sense.

cfgov/unprocessed/css/organisms/read-more.less
anselmbradford: Seems to work fine.

Scotchester: Yeah, I just tried it and tested the functionality and it worked /shrug

PR #5680, opened 2020-04-22
Fix deprecation warnings

cwdavies:
When tox is run with `python -Wd` there are a multitude of deprecation warnings logged
The following changes are made to fix many of these deprecations:

1. Change `assertRaisesRegexp` to `assertRaisesRegex`
1. Replace `assertRegexMatch` with `assertRegex`
1. Change `assertEquals` to `assertEqual`
1. The `warn` method is deprecated, use `warning` instead
1. The `modelcluster.tags` module has been moved to `modelcluster.contrib.taggit`
1. The `unescape` method is deprecated and will be removed in 3.5, use `html.unescape()` instead.
1. The `imp` module is deprecated in favour of `importlib`
1. Change to `is_safe_url` used in `login_with_lockout` in `v1\view\__init__.py` for Django 2.2
1. Change to `_post_feedback` in `v1\tests\handlers\blocks\test_feedback.py` to fix:
`TypeError: Cannot encode None as POST data`

## Additions

- `handle_404_error` in `cfgov/urls.py`
- `test handle_404_error` in `cfgov/tests/test_urls.py`
- `test_handle_404_error_case_insensitive_redirect` in `cfgov/tests/test_urls.py`

## Removals

- Django 1.10 imports

## Testing

Update `tox.ini` to add `-Wd` option to python command line as follows:
`python -b -Wd -m coverage run --source='.' manage.py test {posargs}`

1. tox -e unittest-current
2. tox -e unittest-future

## Notes

- There are far too many deprecation warnings to fix all of them in this Pull Request

## Checklist


cfgov/agreements/urls.py
```diff
@@ -1,14 +1,11 @@
-from agreements.views import index, issuer_search
-
+from django.urls import re_path
 
-try:
-    from django.urls import re_path
-except ImportError:
-    from django.conf.urls import url as re_path
```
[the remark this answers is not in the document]
cwdavies: Agreed and added `try / except` to handle `re_path` import that does not exist in Django 1.11.

willbarton:
@cwdavies could you look at the test coverage failures? It's probably only failing because you touched those files and they don't have full coverage, not because of changes you made specifically.

But I think it'd be good to add coverage as we find places like that.

cfgov/cfgov/urls.py
```diff
@@ -617,5 +617,13 @@ def handle_error(code, request, exception=None):
                             "HTTP Error %s." % str(code), status=code)
 
 
-handler404 = partial(handle_error, 404)
+# Handle case-insensitive URLs
+# Using (?i) in url() patterns is deprecated in Django 2.1
+def handle_404_error(code, request, exception=None):
+    if request.path != request.path.lower():
+        return redirect(request.path.lower())
+    return handle_error(code, request, exception)
```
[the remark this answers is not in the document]
cwdavies: Adding a `test_handle_404_error` in `test_urls.py` to increase test coverage

cfgov/cfgov/tests/test_urls.py
```diff
@@ -133,12 +133,20 @@ def test_handle_error_attribute_error(self, mock_render):
...
         )
 
+    @mock.patch('cfgov.urls.render')
+    def test_handle_404_error(self, mock_render):
+        request = self.factory.get('/test')
+        urls.handle_404_error(404, request)
+        mock_render.assert_called_with(
+            request, '404.html', context={'request': request}, status=404
+        )
+
```
willbarton, line 149: Could you add another test of the case-insensitivity handling in `urls.handle_404_error`? I think something like:

```python
    def test_handle_404_error(self):
         request = self.factory.get('/TEst')
         response = urls.handle_404_error(404, request)
         self.assertRedirects(response, '/test')
```

cfgov/cfgov/tests/test_urls.py
```diff
@@ -133,12 +133,30 @@ def test_handle_error_attribute_error(self, mock_render):
...
+    def test_handle_404_error_case_insensitive_redirect(self):
+        request = self.factory.get('/TEst')
+        response = urls.handle_404_error(404, request)
+        self.assertRedirects(
+            response,
+            '/test',
+            target_status_code=302,
```
willbarton, line 156:
```suggestion
            target_status_code=301,
```

cfgov/cfgov/urls.py
```diff
@@ -617,5 +617,13 @@ def handle_error(code, request, exception=None):
                             "HTTP Error %s." % str(code), status=code)
 
 
-handler404 = partial(handle_error, 404)
+# Handle case-insensitive URLs
+# Using (?i) in url() patterns is deprecated in Django 2.1
+def handle_404_error(code, request, exception=None):
+    if request.path != request.path.lower():
+        return redirect(request.path.lower())
```
willbarton, line 624: Per @kelleyholden's suggestion in GHE/CFGOV/platform/issues/3748#issuecomment-249062, these should be 301s:

```suggestion
        return redirect(request.path.lower(), permanent=True)
```

cwdavies: Django comes with an optional redirects application. It lets you store redirects in a database and handles the redirecting for you. It uses the HTTP response status code 301 Moved Permanently by default.

willbarton: @cwdavies the Django redirects application does the same thing our Wagtail redirect application does. But that's not really relevant here.

We need to make sure we're sending permanent, 301 redirects for case corrections. By default [Django's `redirect()` function has `permanent=False`](https://docs.djangoproject.com/en/2.2/topics/http/shortcuts/#redirect), which is a 302. By passing `permanent=True`, it will be a 301.

cfgov/cfgov/tests/test_urls.py
cwdavies: As I could not get assertRedirects to work with target_status_code=301. I added:
self.assertEqual(response.status_code, 301)
self.assertIsInstance(response, HttpResponsePermanentRedirect)

cwdavies:
https://docs.djangoproject.com/en/stable/ref/contrib/redirects/

Django comes with an optional redirects application. It lets you store redirects in a database and handles the redirecting for you. It uses the HTTP response status code 301 Moved Permanently by default.

https://docs.djangoproject.com/en/stable/ref/request-response/#django.http.HttpResponsePermanentRedirect

class HttpResponsePermanentRedirect¶
Like HttpResponseRedirect, but it returns a permanent redirect (HTTP status code 301) instead of a “found” redirect (status code 302).

cfgov/cfgov/tests/test_urls.py
cwdavies: Done

cfgov/cfgov/tests/test_urls.py
```diff
@@ -133,12 +134,32 @@ def test_handle_error_attribute_error(self, mock_render):
...
+        urls.handle_404_error(404, request)
+        mock_render.assert_called_with(
+            request, '404.html', context={'request': request}, status=404
+        )
+
+    def test_handle_404_error_case_insensitive_redirect(self):
+        request = self.factory.get('/TEst')
+        response = urls.handle_404_error(404, request)
+        self.assertEqual(response.status_code, 301)
+        self.assertIsInstance(response, HttpResponsePermanentRedirect)
```
willbarton, line 155: I think this is unnecessary, given that the status code is 301. A user's browser won't care which Django response object it is. I think it'd be best to restore the `assertRedirects` below instead, to make sure that `/TEst/` redirects to `/test/` — that it redirects from mixed case to lower case.

cwdavies: A 301 redirect means that the page has permanently moved to a new location. - This is requested and implemented
A 302 redirect means that the move is only temporary. - This is not requested for case-insensitive URLs

When I was reading: https://docs.djangoproject.com/en/stable/ref/contrib/redirects/
I tried changing `MIDDLEWARE` in `cfgov/settings/base.py` to add:
`django.contrib.redirects.middleware.RedirectFallbackMiddleware`
Currently the following redirect middleware is being used:
`wagtail.contrib.redirects.middleware.RedirectMiddleware`


This seems like something we should pair on.
When I uncomment the self.assertRedirects line I see the following:
```
FAIL: test_handle_404_error_case_insensitive_redirect (cfgov.tests.test_urls.HandleErrorTestCase)
  File "cfgov-refresh/cfgov/cfgov/tests/test_urls.py", line 160, in test_handle_404_error_case_insensitive_redirect
    fetch_redirect_response=True
  File "cfgov-refresh/.tox/unittest-current/lib/python3.6/site-packages/django/test/testcases.py", line 294, in assertRedirects
    % (response.status_code, status_code)
AssertionError: 301 != 302 : Response didn't redirect as expected: Response code was 301 (expected 302)
```
I have tried setting `fetch_redirect_response=False` and `fetch_redirect_response=True`

cwdavies: Resolved using
`self.assertEqual(response.get('location'), '/test')`

willbarton:
@cwdavies I figured out the right incantation and pushed a change to the tests to use `assertRedirects`. I think this is ready to go!

PR #5698, opened 2020-05-01
Fix the iRegs wayfinder to account for newly hidden expandable content

niqjohnson:
The iRegs wayfinder broke when we started adding `u-hidden` classes to collapsed expandable content in https://github.com/cfpb/design-system/pull/406 and added to cfgov-refresh in #5617. The `u-hidden` class gives those elements a `display: none`, which means their positions can't be correctly calculated for the map of all paragraphs on the page that powers the wayfinder. That was causing the wayfinder to always show the ID of the last interpretation paragraph on the page no matter where you scrolled.

This change fixes that by toggling the `u-hidden` class off while the map of paragraph positions is being constructed, then toggling it back on again after the map is done.

## Changes

- Changed `updateParagraphPositions` to only get the positions of *visible* paragraphs instead of all paragraphs (i.e. only paragraphs that aren't in a `u-hidden` container)

## Testing

Note that, because we have an iRegs service worker that caches the JS that powers the wayfinder, you might have to test in an incognito window or hard refresh some pages to get the new JS.

1. Visit any iRegs page with inline interpretations (like http://localhost:8000/policy-compliance/rulemaking/regulations/1026/18/, which has a *bunch* of paragraphs if you want to test on a big page). Without expanding any of the interps, scroll up and down the page to make sure the wayfinder updates correctly.
2. Expand some interps and verify that the wayfinder still shows the correct paragraph IDs as you scroll up and down the page. Also try collapsing some interps and checking that the wayfinder still looks good.
3. Visit an iRegs page without inline interpretations (any page in Supplement I of any reg will work) and make sure the wayfinder still works there.

## Notes

Yeesh, my JavaScript is so rusty, so definitely let me know if anyone can think of a better way to do this.

## Checklist

### Browsers

- [x] Chrome on desktop
- [x] Firefox
- [x] Safari on macOS
- [x] Edge
- [x] Safari on iOS
- [ ] Chrome on Android

### Accessibility

- [x] Keyboard friendly
- [x] Screen reader friendly

### Other

- [x] Is useable without CSS
- [x] Is useable without JS

### DIFF
--- a/cfgov/unprocessed/apps/regulations3k/js/permalinks-utils.js
+++ b/cfgov/unprocessed/apps/regulations3k/js/permalinks-utils.js
@@ -1,3 +1,5 @@
+import { closest } from '../../../js/modules/util/dom-traverse';
+
 // Array that tracks paragraph positions
 let paragraphPositions;
 const regs3kMainContent = document.querySelector( '.regulations3k' );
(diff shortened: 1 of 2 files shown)


contolini:
If the service worker is being a PITA check these two boxes when testing locally:

Screen Shot 2020-05-01 at 1 38 56 PM

niqjohnson:
@Scotchester, @anselmbradford: I refactored this to get rid of the method I added that toggled the `u-hidden` class off and on during `updateParagraphPositions`. Instead I thought of what seems like the much more sensible approach of having `updateParagraphPositions` update the positions of *only* visible paragraphs (i.e. paragraphs that aren't in a `u-hidden` container).

PR #5709, opened 2020-05-06
Re-enable venv-update and resolve dependencies for Django 2.2

willbarton:
This PR does two major things:

1. Re-enables `tox_pip_extensions_ext_venv_update` in tox, which will automatically update our local test envs with dependency changes, and will enforce dependency resolution, and updates dependencies based on what they collectively require to achieve a set of dependencies that match.
2. Pins Wagtail to `<2.6` in `unittest-future`. Wagtail 2.5 is the first version to support Django 2.2, so this satisfies `venv-update`'s requirement for matching version ranges.

There is one other minor change as well:

- `OPTIONAL_APPS` in settings has been removed as part of the dependency/app updates. There's no practical change here, as our satellite apps have been non-optional for a while now.

And there's one thing to highlight:

- `venv-update` is not exactly under *active* development (it doesn't seem inactive, but it just seems like many of our own libraries — maintained as Yelp needs to), but it requires `pip` `>=18.1`. This is a couple years old at this point, and I'd hope not to keep it there long-term. I do think for now the benefit of using `venv-update` is worth the limited risk in pinning an older `pip` for now.

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -92,25 +92,17 @@ INSTALLED_APPS = (
     "diversity_inclusion",
     "mega_menu.apps.MegaMenuConfig",
     "form_explainer.apps.FormExplainerConfig",
+
+    # Satellites
+    "comparisontool",
+    "retirement_api",
+    "ratechecker",
+    "countylimits",
+    "complaint_search",
+    "ccdb5_ui",
+    "mptt",
+    "teachers_digital_platform",
 )
-
-OPTIONAL_APPS = [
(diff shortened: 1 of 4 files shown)


Scotchester:
Not certain if there's a more extensive way to test this, but `tox` runs successfully for me on this branch (including this branch after rebasing on master).

higs4281:
> * but it requires `pip` `>=18.1`.

`<=18.1` ?

willbarton:
>> * but it requires `pip` `>=18.1`.

> `<=18.1` ?

Indeed!

PR #5715, opened 2020-05-07
Strip extraneous characters at end of URLs

cwdavies:
Update `handle_404_error` to remove the last character from the URL if that character is in the `extraneous_char_list`

## Additions

- test_handle_404_error_strip_extraneous_chars

## Removals

- URLs that end in %20) have these two characters removed

## Testing

1. tox -e unittest-current cfgov.tests.test_urls.HandleErrorTestCase

## Checklist


willbarton:
Do you think we could do this with a regular expression matching any of those characters multiple times at the end of a string (i.e. `[!#$%&()*+,-.:;<=>?@\[\]^_`{|}~]+$`) and then using `re.sub()` on the `request.path`? That would generalize this to more than just the last 1-2 characters of the string.

Scotchester:
> Do you think we could do this with a regular expression matching any of those characters multiple times at the end of a string and then using `re.sub()` on the `request.path`? That would generalize this to more than just the last 1-2 characters of the string.
> 
> Something like this (pseudocode, I haven't tried it):
> 
> ```python
> extraneous_char_re = re.compile(r'[!#$%&()*+,-.:;<=>?@\[\]^_`{|}~]+$')
> request.path = re.sub(extraneous_char_re, '', request.path)
> ```

I had a similar thought, Will. Or if we wanted to avoid regex, we could `while` through basic string comparisons of the final character until that final character is not one of the extraneous characters.

I was testing something to that effect yesterday afternoon, but ran into some weirdness with how browsers convert spaces to %20 that I couldn't resolve before the end of the day.

willbarton:
I haven't tested, but I'd assume that `re.sub` is more performant than a `while` loop over the path.

Scotchester:
> I haven't tested, but I'd assume that `re.sub` is more performant than a `while` loop over the path.

Probably true.

cfgov/cfgov/urls.py
```diff
@@ -617,11 +617,20 @@ def handle_error(code, request, exception=None):
...
+    extraneous_char_list = [
+        '!', '#', '$', '%', '&', '(', ')', '*', '+', ',', '-', '.', ':', ';',
+        '<', '=', '>', '?', '@', '[', ']', '^', '_', '`', '{', '|', '}', '~'
+    ]
     if request.path != request.path.lower():
         return redirect(request.path.lower(), permanent=True)
+    if request.path[-2:] == " )":
+        return redirect(request.path[:-2:], permanent=True)
+    if request.path[-1] in extraneous_char_list:
+        return redirect(request.path[:-1:], permanent=True)
```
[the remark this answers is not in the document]
cwdavies: Thanks for the feedback and I removed the final colons.

cfgov/cfgov/urls.py
```diff
@@ -617,11 +618,19 @@ def handle_error(code, request, exception=None):
...
 # Using (?i) in url() patterns is deprecated in Django 2.1
 def handle_404_error(code, request, exception=None):
+    url_length = len(request.path)
+    extraneous_char_re = re.compile(r'[ !#$%&()*+,-.:;<=>?@\[\]^_`{|}~]+$')
+
     if request.path != request.path.lower():
         return redirect(request.path.lower(), permanent=True)
+
+    request.path = re.sub(extraneous_char_re, '', request.path)
+    if len(request.path) < url_length:
```
willbarton, line 632: `url_length` is probably unnecessary. Instead of overwriting `request.path` with the `re.sub` (which, I know, was my suggestion), you could just do something like:

```python
stripped_path = re.sub(extraneous_char_re, '', request.path)
if stripped_path != request.path:
   …
```

cwdavies: Done

cfgov/cfgov/urls.py
```diff
@@ -618,18 +619,39 @@ def handle_error(code, request, exception=None):
...
+            # through the CommonMiddleware that performs that task.
+            if path[-1] != '/':
+                path += '/'
+            # If it resolves, redirect to it.
+            resolve(path)
+            return redirect(path, permanent=True)
+        except Resolver404:
+            # If the new path didn't resolve, pass out of here onto the return
+            # statement below to raise the 404 for the original path.
+            pass
```
chosak, line 653: Could also use `try...except...else` here like:

```py
try:
    # Is the cleaned URL valid?
    resolve(path)
except Resolver404:
    # If not, return the original 404 error.
    return handle_error(code, request, exception)
else:
    # If so, redirect to the cleaned URL.
    return redirect(path, permanent=True)
```

cfgov/cfgov/urls.py
```diff
@@ -618,18 +619,39 @@ def handle_error(code, request, exception=None):
...
+    # Check for and remove extraneous characters at the end of the path.
+    extraneous_char_re = re.compile(
+        r'[`~!@#$%^&*()\-_–—=+\[\]{}\\|;:\'‘’"“”,.…<>? ]+$'
+    )
+    path = re.sub(extraneous_char_re, '', path)
+
+    # If the path has changed, try resolving the new path.
+    if path != request.path:
+        try:
+            # Add trailing slash if not present, because resolve() doesn't go
```
chosak, line 643: We have some rare URLs that don't end in a slash, say `/sitemap.xml`. If someone accesses `/sitemap.xml!`, this logic will try resolving `sitemap.xml/` and fail, instead of redirecting to `/sitemap.xml`.

Would it be better to try the resolution twice, first without a slash, and second with?

Additionally, the addition of the trailing slash is governed by [`settings.APPEND_SLASH`](https://docs.djangoproject.com/en/3.0/ref/settings/#append-slash); we do happen to have this set to true, but I wonder if it would be better to check that setting here for consistency.

Scotchester: Yesss good call on the filenames at root.

I thought about checking the setting, and you're right, it's probably a good idea. Thanks for saving me from my laziness :)

cfgov/cfgov/urls.py
Scotchester: Awesome! Forgot you could do that.

Scotchester:
@cwdavies and @willbarton Retreating to just doing a single redirect if the path changed at all. Reasons are detailed in the https://github.com/cfpb/cfgov-refresh/pull/5715/commits/8b495e049a945d22e01e46f3b98832012bdcc1af commit message. Let me know if you have questions or concerns.

willbarton:
@Scotchester nope, that sounds reasonable.

Scotchester:
(Note: Still need to write more tests before merging.)

Scotchester:
New commit pushed with what I think are adequate tests. Ready for final review, @cwdavies @chosak @willbarton @higs4281.

PR #5722, opened 2020-05-12
Fix newly surfaced Python linting errors

willbarton:
It looks like an update to flake8 is catching more linting errors now, which is good! This fixes the new ones it's found.

## Checklist

### DIFF
--- a/cfgov/cfgov/urls.py
+++ b/cfgov/cfgov/urls.py
@@ -83,12 +83,16 @@ urlpatterns = [
         template_name='rural-or-underserved/index.html'),
         name='rural-or-underserved'),
 
-    re_path(r'^documents/(?P<document_id>\d+)/(?P<document_filename>.*)$',
+    re_path(
+        r'^documents/(?P<document_id>\d+)/(?P<document_filename>.*)$',
         DocumentServeView.as_view(),
-        name='wagtaildocs_serve'),
+        name='wagtaildocs_serve'
+    ),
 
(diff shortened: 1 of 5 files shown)


cfgov/cfgov/urls.py
```diff
@@ -83,12 +83,16 @@ def empty_200_response(request, *args, **kwargs):
...
-        name='wagtaildocs_serve'),
+        name='wagtaildocs_serve'
+    ),
 
-    re_path(r'^home/(?P<path>.*)$',
-        RedirectView.as_view(url='/%(path)s', permanent=True)),
+    re_path(
+        r'^home/(?P<path>.*)$',
+        RedirectView.as_view(url='/%(path)s', permanent=True)
+    ),
```
Scotchester, line 95: This feels like it's venturing into autoformatting territory, but I suppose I don't mind that.

willbarton: Well, what flake8 wants is:

```python
re_path(r'^home/(?P<path>.*)$',
        RedirectView.as_view(url='/%(path)s', permanent=True)),
```

For the second argument to line up with the first argument inside the `(` when it's on the same line.

The way we preserve the single 4-char indent of the `RedirectView` is to move the pattern down a line and indent the same way.

Scotchester: Yeah, I get it, it's just interesting that it seems like something Black would handle, if we had implemented that yet. And I guess could possibly come into conflict with Black (not sure without reviewing its rules).

PR #5728, opened 2020-05-13
Update jest and adjust tests to handle latest JSDOM

anselmbradford:
## Changes

- Update `jest` to `26.0.1` from `24.9.0`.
- Update `jest-cli` to `26.0.1` from `24.9.0`.
- Update `babel-jest` to `26.0.1` from `24.9.0`.
- Prefixes `window` on references to global `location` references.
- Changes how `window.location.assign` is mocked in tests to handle changes in JSDOM.
- Changes how body font-size is set in tests to handle changes in JSDOM getComputedStyle method.
- Refactor `recent-notices-spec` to utilize simulate event utility, re-require the tested script before each test, and reset the xhr after each test.

## Testing

1. Pull branch and run `./frontend.sh` and run `gulp test:unit`. It should pass.


anselmbradford:
Coverage failure is because I edited repay.js, but there is currently NO coverage for it. Not planning to add a test for it in the scope of this PR.

test/util/simulate-event.js
```diff
@@ -5,7 +5,17 @@
...
+  let event;
+
+  if ( eventType === 'click' ) {
+    event = new MouseEvent( 'click', {
+      bubbles: true,
+      cancelable: true,
+      view: window
+    } );
+  } else {
+    event = document.createEvent( 'Event', eventOption.currentTarget );
```
contolini, line 17: This doesn't need to be changed to `window.document`, does it?

anselmbradford: Wouldn't hurt in the tests, since JSDOM is finicky

PR #5731, opened 2020-05-19
Add ability to force breadcrumbs for homepage children

willbarton:
This change adds a checkbox to all CFGOVPage pages that, if checked when they are direct children of the home page, will cause their children to include that page in its breadcrumbs.

This is necessary for some top-level pages that are not available through the mega menu.

This is deployed to UCP. Please see GHE/CFGOV/platform/issues/3778 for example URLs to review.

## Screenshots

image

image

## Notes

This removes the special-casing for `owning-a-home`. We will need to manually check the box for that page when this deploys.

## Checklist


cfgov/v1/models/base.py
```diff
@@ -189,28 +203,8 @@ def get_breadcrumbs(self, request):
...
-                                '/owning-a-home/sources')):
-                            ancestor_url = ancestor_url.replace(
-                                'owning-a-home', 'owning-a-home/process')
-                        breadcrumbs.append({
-                            'title': ancestor.title,
-                            'href': ancestor_url,
-                        })
-                    return breadcrumbs
-                # END TODO
+                if ancestor.specific.force_breadcrumbs:
```
chosak, line 206: Could we add [`.specific()`](https://docs.wagtail.io/en/v2.7/reference/pages/queryset_reference.html#wagtail.core.query.PageQuerySet.specific) to the `get_ancestors()` call above instead of doing `.specific` on each one? Doing it this way means an extra database query per ancestor, whereas using `.specific()` will use only one query per ancestor page type.

cfgov/v1/models/base.py
```diff
@@ -189,28 +203,8 @@ def get_breadcrumbs(self, request):
...
+                    return [ancestor for ancestor in ancestors[i:]]
```
chosak, line 207: Can this be simplified to `ancestors[i:]`, and the line below to `ancestors[i + 1:]`?

willbarton: Done in 76da944!

cfgov/v1/models/base.py
willbarton: Done in 76da944!

PR #5734, opened 2020-05-19
Fix feedback export from admin

willbarton:
Fix the use of`call_command` in the admin feedback export by passing the the page slugs to `call_command` instead of looking up the slugs first and passing the page objects.

The one side-effect of this is to requires the hard-coded slugs to exist and be unique (where previously if they didn't, the lookup in `_get_pages_from_slugs` would ignore `DoesNotExist` and `MultipleObjectsReturned`). This is consistent with the behavior of the management command, and I think that's okay — if we want to allow for ignorable errors, I think we should do it there instead of in the admin, so that the behavior is consistent.

## Checklist

### DIFF
--- a/cfgov/v1/admin_forms.py
+++ b/cfgov/v1/admin_forms.py
@@ -4,8 +4,6 @@ from io import BytesIO, StringIO
 from django import forms
 from django.core.management import call_command
 
-from wagtail.core.models import Page
-
 
 class CacheInvalidationForm(forms.Form):
     url = forms.URLField(required=False,
@@ -26,7 +24,8 @@ class ExportFeedbackForm(forms.Form):
         'bah': ('owning-a-home',),
     }
 
-    def get_filename_dates(self):
+    @property
+    def filename_dates(self):
         return '{}_to_{}'.format(*(
             self.cleaned_data[k].strftime('%Y%m%d')
             for k in ('from_date', 'to_date')
@@ -49,30 +48,17 @@ class ExportFeedbackForm(forms.Form):
         all_pages = []
 
         for name, page_slugs in self.export_files.items():
-            pages = self._get_pages_from_slugs(page_slugs)
-
             # Generate one CSV for each entry in export_files defined above.
-            yield self.generate_zipfile_csv(name, pages)
+            yield self.generate_zipfile_csv(name, page_slugs)
 
-            all_pages.extend(pages)
+            all_pages.extend(page_slugs)
 
         # Generate a final CSV containing all pages not already exported.
         yield self.generate_zipfile_csv('other', all_pages, exclude=True)
 
-    @staticmethod
-    def _get_pages_from_slugs(page_slugs):
-        pages = []
-
-        for page_slug in page_slugs:
-            try:
-                pages.append(Page.objects.get(slug=page_slug))
-            except (Page.DoesNotExist, Page.MultipleObjectsReturned):
-                continue
-
-        return pages
-
     def generate_zipfile_csv(self, name, pages, exclude=False):
         csv_contents = StringIO()
+        csv_filename = f'feedback_{name}_{self.filename_dates}.csv'
 
         call_command(
             'export_feedback',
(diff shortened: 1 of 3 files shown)


chosak:
@willbarton the idea behind calling the management command code was to be DRY. With the proposed changes in this PR, the code would have two independent places where the export queryset is constructed (the form where you've changed it and [the command itself](https://github.com/cfpb/cfgov-refresh/blob/1470f789df6881bb78385d6e887aa23e053a3449/cfgov/v1/management/commands/export_feedback.py#L67-L94)), and it would be better to have only one. One option would be to create some new, central place for this logic to live (e.g. v1.feedback.FeedbackExporter) and have _that_ be called by both the form and the command.

>  Is there stuff we can clean up in the management command where it tries to check if a page is a page or a slug before performing the lookup?

As written the management command always assumes that the 'pages' argument it gets [here](https://github.com/cfpb/cfgov-refresh/blob/1470f789df6881bb78385d6e887aa23e053a3449/cfgov/v1/management/commands/export_feedback.py#L67) is always a set of Page objects. The conversion from string arguments passed on the command line to Page objects was done through a custom argparse type [here](https://github.com/cfpb/cfgov-refresh/blob/1470f789df6881bb78385d6e887aa23e053a3449/cfgov/v1/management/commands/export_feedback.py#L40). I understand that this has broken somehow with the code changes made for the newer versions of Django. My guess is that [this change](https://github.com/cfpb/cfgov-refresh/commit/f57e7b45926cab1fb45c13e9e8a7f284a39e4bee#diff-b27f250e71113c6c6fe5303cdc55e91dL90) that changes from passing the pages as `pages=pages`  to `*pages` invokes the argparse logic when we don't want it to.

If we could find a fix for invoking the management command properly, would you be open to leaving the current organization? In the commit message for the change mentioned above it says "Django 2.1+ handles positional arguments differently than Django 2.0 in `call_command`." -- is there a specific place in the Django release notes / tickets that mentions this change?

willbarton:
> With the proposed changes in this PR, the code would have two independent places where the export queryset is constructed (the form where you've changed it and the command itself), and it would be better to have only one.

I thought about that, but I'm not sure I feel strongly that assembling a queryset is particularly repetitive. Maybe this kind of query is better suited to being on the `Feedback` object manager?

> If we could find a fix for invoking the management command properly, would you be open to leaving the current organization? In the commit message for the change mentioned above it says "Django 2.1+ handles positional arguments differently than Django 2.0 in call_command." -- is there a specific place in the Django release notes / tickets that mentions this change?

This is the change that broke the previous behavior from my debugging: https://github.com/django/django/pull/9701/files

I believe from the [`call_command` documentation](https://docs.djangoproject.com/en/3.0/ref/django-admin/#django.core.management.call_command) the fix for passing the pages as a keyword argument would be to add a `dest` to the argparse argument in the management command.

willbarton:
In particular [this comprehension](https://github.com/django/django/pull/9701/files#diff-860fce37924469764af399caaa365e00R123-R126):

```python
    parse_args += [
         '{}={}'.format(min(opt.option_strings), arg_options[opt.dest])
          for opt in parser._actions if opt.required and opt.dest in options
     ]
```

Results in a value error when `pages` are not given as positional arguments:

```
ValueError: min() arg is an empty sequence
```

willbarton:
@chosak I have an alternative approach in 4630e66 that uses `call_command` and instead of looking up the page slugs in the admin form it passes the slugs to the management command.

willbarton:
@chosak I've settled on just passing page slugs to `call_command` (which makes the most sense to me).

@higs4281 @cwdavies @Scotchester this is ready for review, I think.

cfgov/v1/admin_forms.py
```diff
@@ -49,40 +48,26 @@ def generate_zipfile_csvs(self):
...
-
-        return pages
-
     def generate_zipfile_csv(self, name, pages, exclude=False):
         csv_contents = StringIO()
+        csv_filename = f'feedback_{name}_{self.filename_dates}.csv'
 
         call_command(
             'export_feedback',
+            # pages=pages,
```
Scotchester, line 65: Can this comment be removed?

willbarton: Removed in 90eabd1.

willbarton:
I'm going to merge this. @chosak I'm happy to iterate further if you have anything concerns about how this ended up.

PR #5738, opened 2020-05-19
Move search to the top of the Ask landing page

niqjohnson:
Move the Ask search to the top of the Ask landing page. See [GHE]/CFGOV/platform/issues/3779 for more details.

## Additions

- Added the `label_level` parameter to the `ask-search` template. The `ask-search` template is used in a number of places. Previously, all of those places either didn't show the label (using `show_label=False`) or situated the form under an `h2`. Now that the search form is at the top of the Ask landing page, it needed an `h2` to be semantically correct. The new `label_level` parameter lets you set the heading level in the label to something other than `h3` if you need it.

## Changes

- The Ask search now appears at the top of the Ask landing page.
- Previously, the Spanish search button label, "búsqueda," was cut off (see the [current Spanish Ask landing page](https://www.consumerfinance.gov/es/obtener-respuestas/)); the search button now has a width of `auto` to allow it to break out of its container on Ask pages where the form's `max-width` is set to fix that issue.

## Testing

1. Visit http://localhost:8000/ask-cfpb/ and make sure the Ask search appears at the top of the Ask landing page, as in the screenshots.
2. Visit http://localhost:8000/es/obtener-respuestas/ and make sure the Ask search also appears at the top of the Spanish Ask landing page *and* that "búsqueda" isn't cut off in the submit button.
3. If you have Elasticsearch running locally, make sure the typeahead suggestions still work as they do in production.
4. Check the Ask search at a few screen sizes. The label should drop to `h4` size when all of the other headings on the page make their one-step drop (at the medium breakpoint, I think).
5. Visit some other pages where the Ask search is used (like http://localhost:8000/ask-cfpb/search/, http://localhost:8000/askcfpb/316/, and http://localhost:8000/complaint/) to confirm that nothing has changed on those pages.

## Screenshots

Small screens | Large screens
---- | ----
ask-landing-small | ask-landing-large

## Notes

- @anselmbradford and @virginiacc, I added you as reviewers since you've worked on the Ask search template before. Let me know if you spot anything that can be improved, but no worries if you don't have time to review this (I'll pester @Scotchester 😄 ).

## Todos

- I think the Spanish submit button text, "búsqueda," should be capitalized. But I see that it's also not capitalized in the global search's submit button, so I'm deeming that fix outside the scope of this PR.

## Checklist

### Browsers

- [x] Chrome on desktop
- [x] Firefox
- [x] Safari on macOS
- [x] Edge
- [x] Safari on iOS
- [x] Chrome on Android

### Accessibility

- [x] Keyboard friendly
- [x] Screen reader friendly

### Other

- [x] Is useable without CSS
- [x] Is useable without JS

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/ask-search.html
+++ b/cfgov/jinja2/v1/_includes/organisms/ask-search.html
@@ -8,17 +8,18 @@
 
    Output an ask search bar when given:
 
-   ask_query     The previous search query string.
+   ask_query      The previous search query string.
 
-   language:     page's language.
+   language:      page's language.
 
-   show_label:   Boolean; whether to show form label.
+   show_label:    Boolean; whether to show form label.
 
-   autocomplete: Boolean; whether to allow autocomplete.
+   autocomplete:  Boolean; whether to allow autocomplete.
 
-   placeholder: String; text to enter for the search input's placeholder value.
+   placeholder:   String; text to enter for the search input's placeholder value.
 
-   label_level:  String; heading level to use for the label.
+   is_subsection: Boolean; whether to mark up the form label with an h3 to fit
+                  into an exisitng section or with an h2 to be its own section
 
    ========================================================================== #}
 
@@ -28,13 +29,15 @@
     show_label=True,
     autocomplete=True,
     placeholder='',
-    label_level='h3'
+    is_subsection=True
 ) %}
 <div class="o-search-bar">
     <form method="get" action="{{ _('/ask-cfpb/search/') }}">
         {% if show_label %}
         <label for="o-search-bar_query">
-            <{{ label_level }}>{{ _('Search for your question') }}</{{ label_level }}>
+            <{{ 'h3' if is_subsection else 'h2' }}>
+                {{ _('Search for your question') }}
+            </{{ 'h3' if is_subsection else 'h2' }}>
         </label>
         {% endif %}
 
--- a/cfgov/jinja2/v1/_includes/organisms/ask-search.html
+++ b/cfgov/jinja2/v1/_includes/organisms/ask-search.html
@@ -18,6 +18,8 @@
 
    placeholder: String; text to enter for the search input's placeholder value.
 
+   label_level:  String; heading level to use for the label.
+
    ========================================================================== #}
 
 {% macro render(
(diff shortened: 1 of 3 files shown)


cfgov/jinja2/v1/_includes/organisms/ask-search.html
```diff
@@ -18,20 +18,23 @@
...
    ========================================================================== #}
 
 {% macro render(
     ask_query='',
     language='en',
     show_label=True,
     autocomplete=True,
-    placeholder=''
+    placeholder='',
+    label_level='h3'
```
anselmbradford, line 31: Instead of allowing an arbitrary number of heading levels into this component, let's be more atomic and limit the number of variations this component allows. Let's make this a boolean, `is_compact`, or similar and then where you set the heading levels you'd check the boolean:

```
{% if is_compact %}
<h3>{{ _('Search for your question') }}</h3>
{% else %}
<h2>{{ _('Search for your question') }}</h2>
{% endif %}
```
Also, maybe just check it around the tags so you don't repeat the heading text twice, but I'll leave that up to you.

Scotchester: @niqjohnson Ans and I had a conversation around those lines before he posted his comment. Where we landed was that the thing probably needs a significant overhaul, and moving of some of the parts of it into fields on the Wagtail block (e.g., if we had a `HeadingBlock` in there, this template logic would be unnecessary), but given the "urgency" of this request, I would not hold up a short-term solution for future proofing.

The idea behind Ans's specific suggestion was to attempt to limit the possible proliferation of more variations on the component in the interim between now and doing that overhaul.

niqjohnson: @anselmbradford, @Scotchester: I fixed this in fb44e25. I changed the string `label_level` to the boolean `is_subsection`. Hopefully `is_subsection` as the name for that property captures the intent: is the form rendered inside an existing section of a page that already has its own `h2` or should it be a top-level section on the page that needs an `h2`. I can change the label if you can think of a better name.

Also, I used a ternary to handle the template logic, but I can change that to a regular if/else block if that seems confusing.

anselmbradford: LGTM 👍 Nice job!

PR #5759, opened 2020-06-03
Revise cohort calculations in school comparisons

higs4281:
The initial design of our public-facing college enrollment disclosure tool
envisioned graphs that compare how a school ranks among a cohort of schools.

The initial cohorts were grouped three ways:
1. Highest degree cohort -- mainly so that community colleges aren't compared with 4-year schools
2. State cohort, so that a student could see how a school rates among other schools in the same state
3. Control cohort -- whether a school was public or private

The API currently delivers these pre-calculated values for every school.
Example for GWU: https://www.consumerfinance.gov/paying-for-college2/understanding-your-financial-aid-offer/api/school/131469/

We now want to always group schools by highest degree, before building cohort comparisons.
As a result, the API will change the cohort values that are delivered. 

- cohortRankByHighestDegree – This will now reflect the base, national cohort
- cohortRankByState – The base cohort, narrowed to the school's state
- cohortRankByControl – The base cohort, narrowed to the school's type (public or private)   

## Testing

Check out this branch and run the `refresh-data.sh` script, which will run the branch's migration, or run 'migrate' yourself. Then you can update local GWU data cohort values.

I'll be adding a management command to run this script in a future PR.

For now, you can test it with the Django shell.

```bash
./cfgov/manage.py shell

from paying_for_college.disclosures.scripts.process_cohorts import run
run(single_school=131469)
```

Compare the results of the GWU API call above with the same
call locally (http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/api/school/131469/)
It helps to have a browser plugin (like JSON Viewer in Chrome) to view the API's json responses.

You should see some small adjustments to cohort values. They may vary slightly from the screen shots below, depending on your local data.

New cohorts (local) | Cohorts (prod)
:---------------: |  :---------------:
cohorts_local | cohorts_prod


cfgov/paying_for_college/disclosures/scripts/process_cohorts.py
```diff
@@ -72,29 +65,43 @@ def rank_by_metric(school, cohort, metric):
...
         if count % 500 == 0:  # pragma: no cover
-            print(count)
-        state_cohort = STATE.get(school.state) if school.state else None
-        control_cohort = CONTROL.get(
-            school.control) if school.control else None
-        degree_cohort = DEGREES_HIGHEST.get(
-            school.degrees_highest) if school.degrees_highest else None
+            logger.info(count)
+        # degree_cohort is the default, national base cohort
+        degree_cohort = DEGREE_COHORTS.get(school.degrees_highest)
```
chosak, line 83:
```suggestion
        degree_cohort = DEGREE_COHORTS.get(school.degrees_highest, [])
```

The [`HIGHEST_DEGREES`](https://github.com/cfpb/cfgov-refresh/blob/6b3d002fa796473cc9d71ec1e2378421dbcdb025/cfgov/paying_for_college/models/disclosures.py#L39) dict only contains values 0-4, and so `DEGREE_COHORTS` gets initialized with only those 5 keys. But there are  many schools (for example 488785) with a different `degrees_highest`, which makes this line return `None`. Then down below when `degree_cohort` is used, it is unexpectedly null.

For example:

```
(cfgov-refresh) ~/Projects/cfgov-refresh $ PYTHONPATH=cfgov DJANGO_SETTINGS_MODULE=cfgov.settings.local python -c 'import django; django.setup(); from paying_for_college.disclosures.scripts.process_cohorts import run; run(488785)'
  File "<string>", line 1, in <module>
  File "/Users/chosaka/Projects/cfgov-refresh/cfgov/paying_for_college/disclosures/scripts/process_cohorts.py", line 85, in run
    s for s in degree_cohort
TypeError: 'NoneType' object is not iterable
```

cfgov/paying_for_college/disclosures/scripts/nat_stats.py
```diff
@@ -18,7 +19,7 @@ def get_bls_stats():
     try:
         with open(BLS_FILE, 'r') as f:
             data = json.loads(f.read())
-    except FileNotFoundError:
+    except builtins.FileNotFoundError:
```
chosak, line 22: Curious why this is needed. Isn't `FileNotFoundError` valid on its own? For example, [Django seems to use it that way frequently.](https://github.com/django/django/search?q=filenotfounderror&unscoped_q=filenotfounderror)

cfgov/paying_for_college/disclosures/scripts/update_ipeds.py
```diff
@@ -19,7 +19,7 @@
 PFC_ROOT = Path(__file__).ancestor(3)
 # LATEST_YEAR specifies first year of academic-year data
 # So 2015 would fetch data for 2015-2016 cycle
-LATEST_YEAR = datetime.datetime.now().year - 1
+LATEST_YEAR = datetime.datetime.now().year - 2
```
chosak, line 22: 😬 This scares me a little bit. Is there some assumption about which year this code will be running?

cfgov/paying_for_college/disclosures/scripts/update_ipeds.py
```diff
@@ -189,16 +186,16 @@ def create_school(id, data):
...
     for field in data:
         if field == 'alias':
-            ALIAS = data['alias']
+            _alias = data['alias']
         else:
             setattr(school, field, data[field])
     school.zip5 = school.zip5[:5]
     school.save()
-    create_alias(ALIAS, school)
+    create_alias(_alias, school)
```
chosak, line 194: Won't this code fail if the first branch of the `if` never gets hit, and `_alias` (or previously `ALIAS`) is undefined?

cfgov/paying_for_college/disclosures/scripts/update_ipeds.py
```diff
@@ -189,16 +186,16 @@ def create_school(id, data):
...
 
 
 def process_missing(missing_ids):
-    """Create missing school and alias objects and dump csv of additions"""
+    """Create missing school and alias objects and dump csv of additions.   """
```
chosak, line 198:
```suggestion
    """Create missing school and alias objects and dump csv of additions."""
```

cfgov/paying_for_college/disclosures/scripts/update_colleges.py
```diff
@@ -60,8 +60,8 @@ def update_programs(api_data, school):
                 'program_name': entry['title'],
                 'cip_code': cip_code,
                 'completers': entry['counts']['titleiv'],
-                'level_code': level_code,
-                'level': PROGRAM_LEVELS.get(int(level_code)),
+                'level': level_code,
```
chosak, line 63: So `level` became `level_name`, and `level_code` became `level`? Is there some data migration needed, or is this not needed because this data isn't live yet?

cfgov/paying_for_college/tests/test_scripts.py
```diff
@@ -669,7 +688,7 @@ def test_get_bls_stats(self):
 
     def test_get_bls_stats_failure(self):
         m = mock_open()
-        m.side_effect = FileNotFoundError
+        m.side_effect = builtins.FileNotFoundError
```
chosak, line 691: Same comment above re: `FileNotFoundError`.

cfgov/paying_for_college/disclosures/scripts/process_cohorts.py
```diff
@@ -72,29 +65,43 @@ def rank_by_metric(school, cohort, metric):
...
+            and s.state
+            and s.state == school.state
+        ]
+        if not school.control:
+            control_cohort = None
+        elif school.control == 'Public':
+            control_cohort = [
+                s for s in degree_cohort if s.control == school.control
+            ]
+        else:
```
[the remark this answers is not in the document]
higs4281: Good point. We are indeed consolidating. Our intention is to offer comparisons among public schools and among private schools. We do not want to isolate for-profit private schools as a cohort.
I'll make that clearer in comments.

cfgov/paying_for_college/disclosures/scripts/process_cohorts.py
```diff
@@ -58,7 +50,8 @@ def rank_by_metric(school, cohort, metric):
     """
     Return a school's percentile rank among a cohort by 3 metrics.
 
-    Possible cohorts are by state, by highest degree, or by control type.
+    Possible cohorts are by degree (the base), by state or by control type.
+    All cohorts are first grouped by highest degree.
     Possible metrics are grad rate, repayment rate, or median total debt.
     """
     values = [getattr(s, metric) for s in cohort if getattr(s, metric)]
```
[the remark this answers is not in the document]
higs4281: Great catch! Thank you. Fixed to allow zero.

cfgov/paying_for_college/disclosures/scripts/process_cohorts.py
```diff
@@ -58,7 +50,8 @@ def rank_by_metric(school, cohort, metric):
     """
     Return a school's percentile rank among a cohort by 3 metrics.
 
-    Possible cohorts are by state, by highest degree, or by control type.
+    Possible cohorts are by degree (the base), by state or by control type.
+    All cohorts are first grouped by highest degree.
```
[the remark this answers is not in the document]
higs4281: Agreed.

cfgov/paying_for_college/disclosures/scripts/update_ipeds.py
higs4281: Scary but true.  
We initially thought we'd run the annual IPEDS update late in the year, in hopes of snagging data from the most recent academic year, but found that updates were never posted in time. In practice, we stay closer to the latest data by running the IPEDS script early in the year. That means that, in 2020, during the 2019-2020 academic year (for which there is no data), we need to specify current-year minus 2 to get the latest data, from the 2018-2019 academic year.

cfgov/paying_for_college/disclosures/scripts/update_ipeds.py
higs4281: Excellent point. It's never happened, but that logic is bad.  
Fixed.

cfgov/paying_for_college/disclosures/scripts/update_colleges.py
higs4281: Data collection evolved in a confusing way here, but there's no data migration needed.  

The following is surely TMI, but here goes:  

The name swapping is driven by the need to keep our API backward-compatible for the original tool, which used program data provided by schools, while dealing with a new source for program data -- CollegeScorecard.

#### The evolution 
- The former db column `level_code` was intended to reflect a program's numeric level, but was never populated or used. In our data [from schools](https://cfpb.github.io/college-costs/csv-spec/), the numeric "program_level" CSV value got stored as the db value "Program.level" and was delivered via our API to the tool as `level`.
- Now that program data is available from CollegeSorecard, as opposed to getting it directly from schools, we want to have the level's name available in our API, while leaving our API var `level` untouched. So we're renaming the empty db column "level_code" to "level_name" for that purpose. 
- The db column "level" is left alone, and we map CollegeScorecard's numerical `level` value  to that. The `update_iped` script's use of the var `level_code` was contributing to the confusion, so I renamed it `level` to match Scorecard and our db value.

cfgov/paying_for_college/disclosures/scripts/nat_stats.py
higs4281: This surprised me too. Faulty linter.
Fixed.

cfgov/paying_for_college/tests/test_scripts.py
higs4281: Fixed.

cfgov/paying_for_college/disclosures/scripts/process_cohorts.py
```diff
@@ -72,29 +65,43 @@ def rank_by_metric(school, cohort, metric):
...
+        elif school.control == 'Public':
+            control_cohort = [
+                s for s in degree_cohort if s.control == school.control
+            ]
+        else:
+            control_cohort = [
+                s for s in degree_cohort if s.control != 'Public'
+            ]
         for metric in ['grad_rate', 'repay_3yr', 'median_total_debt']:
             if not getattr(school, metric):
```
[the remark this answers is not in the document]
higs4281: Fixed

cfgov/paying_for_college/disclosures/scripts/process_cohorts.py
higs4281: Many thanks for this catch. 

We'd like to classify all schools with higher post-graduate programs as graduate-degree-granting, so I added a function to add such schools to our level '4' cohort:

```python
def get_grad_level(school):
    """Consider degrees higher than graduate level '4' as graduate degrees."""
    if int(school.degrees_highest) > 4:
        return '4'
    else:
        return school.degrees_highest
```

higs4281:
@chosak Your close review is much appreciated. All comments addressed.

cfgov/paying_for_college/disclosures/scripts/process_cohorts.py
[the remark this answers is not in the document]
chosak: Wouldn't this line still hit if `getattr(school, metric)` is 0?

cfgov/paying_for_college/disclosures/scripts/update_ipeds.py
chosak: Thanks for the explanation. Would it be worth adding a comment here so that in future this is easier to find?

higs4281:
Both your "minor" things are spot-on. I fixed the metric test and improved the IPEDS year note. I also changed that "LATEST_YEAR" var to "DATA_YEAR" to, I hope,  avoid confusion about what it represents.

PR #5765, opened 2020-06-04
Add timeout to Docker stack deployment in Jenkinsfile

imuchnik:
Adding timeout to docker deploy step

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -107,7 +107,9 @@ pipeline {
             } 
             steps {
                 script {
-                    dockerStack.deploy(env.STACK_NAME, 'docker-stack.yml')
+                    timeout(time: 10, unit: 'MINUTES') {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -107,7 +107,9 @@ pipeline {
             } 
             steps {
                 script {
-                    dockerStack.deploy(env.STACK_NAME, 'docker-stack.yml')
+                    timeout(time: 10, unit: 'MINUTES') {
```
Scotchester, line 110: I can't seem to easily see stage timings going back that far, but at least one successful job in the recent past took 15 minutes total, so I think it's plausible that the deployment stage could sometimes run longer than 10. How about we go with 15 here?

```suggestion
                    timeout(time: 15, unit: 'MINUTES') {
```

PR #5767, opened 2020-06-05
Report in subapp

wpears:
❗ Note: This PR is for the Reports as Webpages project and is not going into Master! If merged, it will go into the reports-as-webpages branch. ❗

This more or less gathers all the report code into `research_reports` and `unprocessed/apps/research-reports`.

I left the `sidenav-toc` organism intact as it seemed more duplicative to have to copy the sidenav layout + the sidenav-toc over to `research_reports` but this could certainly be done.


cfgov/research_reports/models.py
```diff
@@ -1,3 +1,142 @@
...
+    def child_form(cls):
+        return ReportForm
+
+
+class ReportForm(WagtailAdminPageForm, metaclass=ReportMetaclass):
+    pass
+
+
+class Report(CFGOVPage):
+    base_form_class = ReportForm
```
schbetsy, line 94: What does this do?

wpears: It's a workaround that allows nested inline panels. See https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/models/home_page.py#L20-L24 (and the linked issues for more gory details).

PR #5769, opened 2020-06-05
Mega Menu: remove GTM ignore attribute

anselmbradford:
The FlyoutMenu that the MegaMenu uses has the concept of "triggers", which trigger a transition. In the case of the menu, that is expanding and collapsing the menu. At desktop, the triggers are the top-level links that open and close the menu. Additionally, at mobile, the triggers are the hamburger menu and the "back" button. These links have their default link behavior prevented, so the `data-gtm_ignore` attribute is added so they aren't recorded in Google Tag Manager like other links. 

https://github.com/cfpb/cfgov-refresh/pull/2835 added logic so this attribute is only added if the site is in a mobile breakpoint at initialization time. As noted in that PR, it does not factor in resizing, so the page could be loaded at the mobile size and resized to desktop and the links would have the attribute, OR it could be loaded at desktop size and resized to mobile and the links would not. I'm unclear what was intended by this PR since the links at desktop have their default behavior prevented just like in mobile, so I'm reverting it here.

## Removals

- Removes logic to have `data-gtm_ignore` attribute added only if the page is in mobile size at page load.

## How to test this PR

1. Pull branch, build page, visit any page and check that any link with a `data-js-hook="behavior_flyout-menu_trigger"` attribute also have a `data-gtm_ignore="true"` attribute.


anselmbradford:
@kelleyholden Wants the GTM ignore totally removed, since it was just there for when we had the hover-to-open menu behavior (which fired a ton of events). I'll be updating this PR to remove it.

PR #5771, opened 2020-06-09
Mega Menu: Make triggers the same

anselmbradford:
The mega menu has a trigger and an "alt-trigger" for the back button in the mobile menu. These are set up differently and don't quite have the same events or markup. Each trigger should probably be the same, and the menu should maintain a list of triggers instead of hardcoding references to each. 

## Changes

- Remove `alt-trigger` behavior element and replace with a list of triggers within a flyout menu. 
- Enable disabled alt trigger unit test.

## How to test this PR

1. Pull branch and build and test that the menu works like production. That it can be expanded/collapsed on desktop and mobile and via the keyboard.


anselmbradford:
> Is the point to allow an unlimited number of triggers? If so, should all of those [0] and [1] array indices be hard-coded in MegaMenuMobile.js?

Oh instead of doing the look up repeatedly. Sure I can do that.

anselmbradford:
I'll cleanup the look ups in another PR.

Scotchester:
> > Is the point to allow an unlimited number of triggers? If so, should all of those [0] and [1] array indices be hard-coded in MegaMenuMobile.js?
> 
> Oh instead of doing the look up repeatedly. Sure I can do that.

Yeah, caching them as constants would be better, but I was more getting at whether or not we should be requesting specific indices at all, instead of just looping through all of them?

anselmbradford:
> Yeah, caching them as constants would be better, but I was more getting at whether or not we should be requesting specific indices at all, instead of just looping through all of them?

Oh gotcha. So the FlyoutMenu can support arbitrary numbers of triggers, but the specific things that use it (mega menu, search) would likely set their specific trigger references. In the case of mobile mega menu it is setting the focus, clicking, etc. on specific triggers, so would be call those out directly as opposed to iterating over the available ones.

Scotchester:
> > Yeah, caching them as constants would be better, but I was more getting at whether or not we should be requesting specific indices at all, instead of just looping through all of them?
> 
> Oh gotcha. So the FlyoutMenu can support arbitrary numbers of triggers, but the specific things that use it (mega menu, search) would likely set their specific trigger references. In the case of mobile mega menu it is setting the focus, clicking, etc. on specific triggers, so would be call those out directly as opposed to iterating over the available ones.

Ah, OK, that makes sense.

PR #5781, opened 2020-06-11
VideoPlayer consolidation and optional thumbnails

chosak:
This PR encompasses a set of changes around our VideoPlayer module and related modules that include it:

1. VideoPlayers now are configured with YouTube video IDs (like [dQw4w9WgXcQ](https://www.youtube.com/watch?v=dQw4w9WgXcQ)) instead of YouTube embed URLs (like https://www.youtube.com/embed/dQw4w9WgXcQ).
2. VideoPlayers now support an optional thumbnail image which gets displayed before/after the video plays, instead of using the default YouTube video thumbnail. See internal el-camino#247.
3. The FeaturedContent organism now uses the VideoPlayer directly instead of using its own set of custom fields for video details (most of these fields were unused, see below).

EventPage usage of the VideoPlayer template has been updated to be consistent with these changes; however the support for alternate thumbnail images has not been added there.

This PR includes multiple Wagtail data migrations to properly alter existing video data. These migrations required some improvements to our [Wagtail StreamField migration](https://cfpb.github.io/cfgov-refresh/migrations/) code.

As part of this work, and to support easier debugging of video playback, I modified the JS code to support multiple VideoPlayers and FeaturedContent modules per page; previously the JS assumed a single module of each type.

A final minor improvement from this PR is that video playback will now work properly on Wagtail preview/draft views; previously our JS didn't work in those cases.

## How to test this PR

Running a local server with a production dump, here are some example pages to test:

1. VideoPlayer: http://localhost:8000/coronavirus/mortgage-and-housing-assistance/director-message-housing-resources/. Compare with behavior [in prod](https://www.consumerfinance.gov/coronavirus/mortgage-and-housing-assistance/director-message-housing-resources/).
2. FeaturedContent with video: http://localhost:8000/about-us/diversity-and-inclusion/self-assessment-financial-institutions/. Compare with behavior [in prod](https://www.consumerfinance.gov/about-us/diversity-and-inclusion/self-assessment-financial-institutions/).
3. EventPage with video: http://localhost:8000/about-us/events/archive-past-events/cfpb-financial-well-being-research-conference/. Compare with behavior [in prod](https://www.consumerfinance.gov/about-us/events/archive-past-events/cfpb-financial-well-being-research-conference/).

Experiment with setting custom thumbnails for these videos to see how they appear. For the EventPage changes, you can experiment with setting a live video ID for a live event to verify that it works properly.

(You can also use [the Wagtail Block Inventory tool](http://localhost:8000/admin/inventory/) to find other example pages with these modules.)

I've also added two new template debug views for VideoPlayer and FeaturedContent:

- http://localhost:8000/admin/template_debug/v1/video_player/
- http://localhost:8000/admin/template_debug/v1/featured_content/

## Screenshots

Configuring a VideoPlayer:

image

Video section in FeaturedContent:

image

New fields on EventPage:

image
image

## Notes and todos

- I did some analysis of our FeaturedContent modules to make sure that we could safely transition this to using VideoPlayer. That analysis can be found [here](https://nbviewer.jupyter.org/gist/chosak/09f98263908f9d1335c6ce6605f12512) for anyone interested.
- This work exposed a few TODOs around FeaturedContent: for example, the body should be plain text, not rich text. These changes will be tackled in future PRs.
- If/when this is merged, I will update our internal Hubcap documentation around embedding videos.

## Checklist

  - PR titles are used to generate the change log in [releases](../../releases),
    so good ones make that easier to scan.
  - Consider prefixing with the feature/section you're addressing,
    e.g., "Mega Menu: fix layout bug" or "Paying for College: Update content on Repay tool"

#### Browser testing

- [x] Firefox
- [x] Chrome
- [ ] Safari
- [ ] Edge 18
- [ ] Safari on iOS
- [ ] Chrome on Android

[Further guidance on browser support](https://github.com/cfpb/development/blob/master/guides/browser-support.md)

#### Accessibility

- [ ] Screen reader friendly
- [x] Does not introduce new errors or warnings in
      [WAVE](https://wave.webaim.org/extension/)

#### Other

- [ ] Is useable without CSS
- [x] Is useable without JS


cfgov/jinja2/v1/_includes/organisms/video-player.html
```diff
@@ -1,23 +1,132 @@
...
+   video_id:        YouTube video ID, e.g. "dQw4w9WgXcQ".
 
-   value.video_url: A string containing the YouTube embed URL to use.
+   button_pos:      String used to denote the position of the play button.
+                    Possible values: 'center (default), bottomRight'
+
+   thumbnail_url:   Optional alternate image URL to show before video is played
+                    or after it is finished.
+
+   is_fcm:          Boolean. True if this video is being rendered as part of a
```
chosak, line 19: @anselmbradford @Scotchester is there some better way to do this, so that the VP just renders properly depending on whether or not it's contained within an FCM? If not, is there some more appropriate name for this setting, like `display_at_fixed_size`?

cfgov/jinja2/v1/_includes/organisms/featured-content.html
```diff
@@ -6,94 +6,54 @@
...
-                                <a href="{{ link.url }}">
-                                    {{ link.text }}
-                                </a>
-                            </li>
-                        {% endif %}
-                    {% endfor %}
-                {% endif %}
-            </ul>
+
+        {{ value.body | richtext }}
```
anselmbradford, line 34: `richtext` required? 😢

chosak: For now, because the FCM body is a rich text field for some reason. @huetingj and @sonnakim endorsed changing this to plain text, and I'd like to do that subsequent to this PR.

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -846,30 +846,112 @@ def get_context(self, value, parent_context=None):
...
-                       "on the YouTube video page.",
-        }
+            'invalid': "The YouTube video ID is in the wrong format.",
+        },
+        help_text=mark_safe(YOUTUBE_ID_HELP_TEXT)
     )
+    thumbnail_image = images_blocks.ImageChooserBlock(
+        required=False,
+        help_text=mark_safe(
+            'Optional thumbnail image to show before or after the video plays.'
```
niqjohnson, line 881: Should we add a second sentence here to explain what happens if you don't set this image? Something like "If the thumbnail image is not set here, the video player will show the thumbnail that was set in YouTube instead"?

chosak: Thanks for the suggestion @niqjohnson. I've added a commit that adds this sentence. It looks like this in Wagtail:

image

cfgov/jinja2/v1/_includes/organisms/featured-content.html
```diff
@@ -6,94 +6,54 @@
...
+
+        {% if value.links %}
+        <ul class="m-list m-list__links">
+            {% for link in value.links %}
+                <li class="m-list_item">
+                    <a href="{{ link.url }}">{{ link.text | default( link.url ) }}</a>
```
anselmbradford, line 40: Link list [here in the DS](https://cfpb.github.io/design-system/components/lists#link-list) has the link as an element of `m-list`, with `class="m-list_link"` on the link. Does adding that screw anything up?

cfgov/jinja2/v1/_includes/organisms/featured-content.html
```diff
@@ -6,94 +6,54 @@
...
 
-    {% if value.video.url and
-          value.video.id %}
-        {% import 'macros/video-player.html' as video_player with context %}
-        {% do value.video.update({'is_fcm': true}) %}
-        {{ video_player.render(value) }}
-    {% elif value.image.upload %}
+    {% if value.video and value.video.video_id %}
+        {% with value = value.video %}
+            {% do value.update({'is_fcm': true}) %}
```
anselmbradford, line 49: TRIVIAL: I think we tend to add spaces inside parentheses.

cfgov/jinja2/v1/_includes/organisms/video-player.html
```diff
@@ -1,23 +1,132 @@
...
 
    ==========================================================================
 
    Description:
 
-   Create a video player for the Wagtail Video Player module when given:
+   Builds Video markup when given:
 
-   value: Object defined from a StreamField block.
+   video_id:        YouTube video ID, e.g. "dQw4w9WgXcQ".
```
anselmbradford, line 11: 😆 🤣

cfgov/jinja2/v1/_includes/organisms/video-player.html
```diff
@@ -1,23 +1,132 @@
...
 
-   value.video_url: A string containing the YouTube embed URL to use.
+   button_pos:      String used to denote the position of the play button.
+                    Possible values: 'center (default), bottomRight'
```
anselmbradford, line 14: TRIVIAL: other code comments here have an ending period.

cfgov/jinja2/v1/_includes/organisms/video-player.html
anselmbradford: Yes I would agree this has a code smell. Without digging into it I can't fully answer your question, but `display_at_fixed_size` sounds good. If you haven't already I would render the video player outside of a FCM, but with this flag set and see what happens. It should be able to be rendered with this set but outside of a FCM. If it doesn't then there's an incorrect tight coupling happening here.

cfgov/jinja2/v1/_includes/organisms/video-player.html
```diff
@@ -1,23 +1,132 @@
...
+        Exclude the origin parameter on preview views to avoid breaking video
+        playback when the preview request host (from the Wagtail Site) doesn't
+        match the URL serving the video.
+    #}
+    {% set video_url =
+        'https://www.youtube.com/embed/' ~
+        video_id ~
+        '?enablejsapi=1' ~
+        (
+            ('&origin=' ~ request.scheme ~ '://' ~ request.get_host())
```
anselmbradford, line 45: Would `request.scheme` ever not be `https`?

cfgov/jinja2/v1/_includes/organisms/video-player.html
```diff
@@ -1,23 +1,132 @@
...
+    <div class="o-video-player
+                o-video-player__youtube
+                {{- ' o-featured-content-module_visual' if is_fcm else '' }}"
+         data-id="{{ video_id }}"
+         data-src="{{ video_url }}"
+         {{- ' data-custom-thumbnail' if thumbnail_url else '' }}>
+        {{ caller() | safe if caller else '' }}
+        <div class="o-video-player_video-container
+                    {{- ' o-video-player_video-container__flexible' if not is_fcm else '' }}
+                    show-on_video-playing">
```
anselmbradford, line 59: Ideally, this should be a utility since it's not atomic `u-show-on_video-playing` . Probably a fix for another time, unless you want to adjust it here.

cfgov/jinja2/v1/_includes/organisms/video-player.html
```diff
@@ -1,23 +1,132 @@
...
+                        'linkedin_title': 'Consumer Financial Protection Bureau',
+                        'linkedin_text': 'Check out this video from the @CFPB',
+                        'twitter_text': 'Check out this video from the @CFPB',
+                        'email_title': page.title if page else 'Consumer Financial Protection Bureau'
+                    } ) }}
+                {% endif %}
+            </div>
+        </div>
+        <div class="o-video-player_image-container
+                    hide-on_video-playing">
```
anselmbradford, line 89: Ideally this should be a utility `u-hide-on_video-playing`

cfgov/unprocessed/js/modules/youtube-api.js
```diff
@@ -7,22 +7,48 @@ import jsLoader from './util/js-loader';
...
+  _callbacks.push( callback );
+
+  /**
+   * The "onYouTubeIframeAPIReady" function is automatically called when the
+   * YouTube IFrame API is done loading.
+   * See https://developers.google.com/youtube/iframe_api_reference
+   */
+  function onYouTubeIframeAPIReady() {
+    _scriptLoaded = true;
+    _callbacks.forEach( function( cb ) {
```
anselmbradford, line 34: Probably could be arrow function `_callbacks.forEach( cb => {`

cfgov/unprocessed/js/organisms/FeaturedContentModule.js
```diff
@@ -22,46 +19,52 @@ let _videoPlayer;
...
-  _dom = checkDom( element, BASE_CLASS );
+  const _dom = checkDom( element, BASE_CLASS );
 
-  this.init = init;
+  let _videoPlayer = null;
 
-  return this;
-}
+  /**
+   * @returns {FeaturedContentModule|undefined} An instance.
```
anselmbradford, line 27: Eventually we should update these init methods to just return the instance instead of undefined if the instance has already been initialized. I put in undefined so you'd have a way to know if something was already initialized, but thinking about it further I think it should always just return the instance. This should happen with all code structures like this across the codebase at once, so we should probably leave it alone here.

cfgov/unprocessed/js/organisms/FeaturedContentModule.js
```diff
@@ -22,46 +19,52 @@ let _videoPlayer;
...
+    _dom.classList.remove( 'video-playing' );
+  }
 
-/**
- * Event handler for when video has begun playing.
- */
-function _videoPlayHandler() {
-  _dom.classList.add( 'video-playing' );
-}
+  this.init = _init;
```
anselmbradford, line 65: `_init` is being set to public here so shouldn't have an underscore prefix.

cfgov/unprocessed/js/organisms/VideoPlayer.js
```diff
@@ -33,174 +20,163 @@ let _stopVideoBinded;
...
-    imageURL = youTubeAPI.fetchImageURL( _videoId );
+  const _dom = checkDom( element, BASE_CLASS );
+
+  const _videoId = _dom.getAttribute( 'data-id' );
+  const _showCustomThumbnail = _dom.hasAttribute( 'data-custom-thumbnail' );
+
+  const _iframeDom = _dom.querySelector( `.${ BASE_CLASS }_iframe` );
+  const _imageDom = _dom.querySelector( `.${ BASE_CLASS }_image` );
+  const _closeBtnDom = _dom.querySelector( `.${ BASE_CLASS }_close-btn` );
+  const _playBtnDom = _dom.querySelector( `button.${ BASE_CLASS }_play-btn` );
```
anselmbradford, line 31: Hm this should probably be a consistent query to that of the close button. So it should omit `button` here or the close button should include `button`.

cfgov/unprocessed/js/organisms/VideoPlayer.js
```diff
@@ -33,174 +20,163 @@ let _stopVideoBinded;
...
+    _player.addEventListener( 'onStateChange', _videoStateChangeHandler.bind( this ) );
+  }
 
-  /* On page load we show a play link that links directly to the video, so that
-     the user can still access the video with no JavaScript.
-     We need to hide the link and show the play button for the embedded video. */
-  _playLinkDom.classList.add( 'u-hidden' );
-  _playBtnDom.classList.remove( 'u-hidden' );
+  /**
+   * Event handler for the when the video is ready.
```
anselmbradford, line 72: Typo `for the when`

cfgov/jinja2/v1/_includes/organisms/featured-content.html
chosak: Nice find. It turns out that this does pretty significantly change the behavior of these links, both on desktop and mobile.

The font weight gets increased to 500, and on mobile these links get a bigger touch target and horizontal lines separating them.

[Page with single link](https://www.consumerfinance.gov/consumer-tools/auto-loans/), desktop and mobile:

|Before|After|
|-|-|
|image|image|
|image|image|

[Page with multiple links](https://www.consumerfinance.gov/consumer-tools/credit-reports-and-scores/consumer-reporting-companies/), on desktop and mobile:

|Before|After|
|-|-|
|image|image|
|image|image|

(I fixed the markup to remove that highlighted space before the download icon as well.)

Ping @huetingj for your thoughts on making this change.

anselmbradford: We can keep it without the link class, but should document that pattern in DS, IMO.

chosak: Per team discussion, I've added this change to this PR.

cfgov/unprocessed/js/organisms/FeaturedContentModule.js
chosak: Thanks, I made similar changes to the public methods in VideoPlayer as well.

cfgov/jinja2/v1/_includes/organisms/video-player.html
chosak: When running with localhost or in some Docker container accessed over HTTP, it might not be.

cfgov/jinja2/v1/_includes/organisms/video-player.html
chosak: Other examples in cfgov-refresh seem to use a dash after `u-show-on`, like `u-show-on-mobile` or `u-show-on-stacked`. Should there be an underscore anywhere?

cfgov/unprocessed/js/organisms/VideoPlayer.js
chosak: I guess that `button.` is included here because [the next line](https://github.com/cfpb/cfgov-refresh/blob/c280e5d0b0f10f3ad74508e471565e8960740527/cfgov/unprocessed/js/organisms/VideoPlayer.js#L32) refers to the same classes for a different element type:

```html
  const _playBtnDom = _dom.querySelector( `button.${ BASE_CLASS }_play-btn` );
  const _playLinkDom = _dom.querySelector( `a.${ BASE_CLASS }_play-btn` );
```

Is this correct? Might make sense to add `button.` to the play button, then, but then why not have it on the other selectors?

cfgov/jinja2/v1/_includes/organisms/featured-content.html
Scotchester: Ah, good, was just about to comment on this. Yes, the After here is definitely the correct pattern.

cfgov/jinja2/v1/_includes/organisms/video-player.html
chosak: So currently (slightly modified by this branch), [the video player template](https://github.com/cfpb/cfgov-refresh/blob/fd74edddc78452b1f161ac4abfbe25ef6dbb2667/cfgov/jinja2/v1/_includes/macros/video-player.html#L52) adds `o-featured-content-module_visual` to `<div class="o-video-player">` if this boolean is true. That FCM style ([from the DS](https://github.com/cfpb/design-system/blob/1cf848f8df72f56252bbb5d99a2117f26238816d/packages/cfpb-layout/src/organisms/featured-content-module.less#L36)) then sets `position: absolute; top: 0; right: 0` to the top-level organism, which causes it to be rendered in the upper-right of the page.

So it seems like the FCM relies on the VP template having a certain class to make it render on the right side, vs the FCM itself applying that class. Is there some way of pulling that out?

cfgov/jinja2/v1/_includes/organisms/video-player.html
anselmbradford: Ah true, yeah so this is saying `video-playing` is an element of the `show-on` block, which doesn't exist, right? So yes I would agree it should be `u-show-on-video-playing`.

cfgov/jinja2/v1/_includes/organisms/video-player.html
anselmbradford: Make dashes, not underscore.

cfgov/unprocessed/js/organisms/VideoPlayer.js
anselmbradford: Yeah I'd just add it so the buttons are consistent and not worry about the rest above it. I don't know for sure, but narrowing down the selector by element type may make it microscopically faster to look up.

cfgov/jinja2/v1/_includes/organisms/video-player.html
chosak: Fixed in latest commit.

cfgov/unprocessed/js/organisms/VideoPlayer.js
chosak: Fixed in latest commit.

Scotchester:
@chosak Would you please add tests for the missing coverage?

cfgov/jinja2/v1/_includes/organisms/featured-content.html
```diff
@@ -6,94 +6,56 @@
...
-   See [GHE]/CFPB/hubcap/wiki/Featured-Content-Module
+   See https://cfpb.github.io/design-system/components/featured-content-module
 
    Create a featured content molecule when given:
 
-   value:                 Object defined from a StreamField block.
+   value.heading:   Heading text.
 
-   value.heading:         Heading text.
+   value.body:      Body text (rich text).
```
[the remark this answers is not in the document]
chosak: I promise I won't forget. We'll need to edit this file anyway to remove the `| richtext`.

cfgov/jinja2/v1/_includes/organisms/featured-content.html
```diff
@@ -6,94 +6,56 @@
...
-                                </a>
-                            </li>
-                        {% endif %}
-                    {% endfor %}
-                {% endif %}
-            </ul>
+
+        {{ value.body | richtext }}
+
+        {% if value.links %}
```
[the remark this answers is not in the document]
chosak: I'd prefer not to because it breaks the separation between the backend Wagtail implementation and this (in theory) standalone template.

cfgov/jinja2/v1/_includes/organisms/featured-content.html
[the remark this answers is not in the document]
Scotchester: Right, I'm just saying it would be easy to remove the `| richtext` and forget to remove this parenthetical in the opening comment. But not a huge deal.

cfgov/jinja2/v1/_includes/organisms/featured-content.html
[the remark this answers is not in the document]
Scotchester: I maintain that this separation comes at the cost of clarity for developers looking at the template, but it's not a hill I'm gonna die on.

cfgov/jinja2/v1/events/_macros.html
```diff
@@ -117,44 +117,25 @@ <h2 class="event-venue_heading">
...
+                        video_id=event.live_video_id,
+                        button_pos='bottomRight'
+                    ) }}
                 {% elif event_state == 'past' %}
-                    {% if event.youtube_url %}
+                    {% if event.archive_video_id %}
                         {# DISPLAY VIDEO RECORDING #}
-                        {% import 'macros/video-player.html' as video_player with context %}
-                        {# TODO: Figure out what's happening with this image_url, if anything. #}
-                        {% set image_url  = event.flickr_url %}
```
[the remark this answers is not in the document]
chosak: Mostly nothing. Technically it was possible to set an EventPage's flickr_url to something that would then be used for the video thumbnail, but only in cases when JS is disabled. I looked through existing events with flickr_url and it didn't seem worth keeping this functionality.

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -846,30 +846,114 @@ def get_context(self, value, parent_context=None):
...
+    def clean(self, value):
+        cleaned = super().clean(value)
+
+        errors = {}
+
+        if not cleaned['video_id']:
+            if getattr(self.meta, 'required', True):
+                errors['video_id'] = ErrorList([
+                    ValidationError('This field is required.'),
+                ])
```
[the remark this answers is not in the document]
chosak: This is needed because of https://github.com/wagtail/wagtail/issues/2665. We want to be able to make VideoPlayers optional sub-blocks of other blocks (like FeaturedContent). But we still want `video_id` to be required anytime a VideoPlayer is defined.

We have to define `video_id` with `required=False`, otherwise you'd never be able to leave VideoPlayer blank. But if VideoPlayer is defined like `VideoPlayer(required=True)`, then in that case we do want `video_id` to act as if it is required. That's what this logic does.

cfgov/v1/atomic_elements/atoms.py
```diff
@@ -112,6 +112,33 @@ class Button(Hyperlink):
...
+        # TODO: This duplicates the logic in v1.jinja2tags.image_alt_value,
+        # which cannot be called here because of a circular import. It would
+        # be better to deprecate the image_alt_value tag in favor of using
+        # this logic wherever we use ImageBasic atoms.
+        alt = self.get('alt')
+        if alt:
+            return alt
+
+        upload = self.get('upload')
+        if upload:
```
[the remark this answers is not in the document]
chosak: Technically this isn't necessary because [`alt` can't be null](https://github.com/cfpb/cfgov-refresh/blob/7c95ee016c20480a6085031ce9685fb063309cb8/cfgov/v1/models/images.py#L13). But I agree this will add some clarity to the code, so I'll add it.

cfgov/v1/atomic_elements/organisms.py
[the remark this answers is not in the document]
Scotchester: What about wrapping the VideoPlayer in a ListBlock or StreamBlockj with a max of one item when used in a place like the FCM? Then we could make `video_id` required, right?

cfgov/v1/atomic_elements/atoms.py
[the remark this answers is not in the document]
Scotchester: NOt sure if this is a null vs. blank vs. none thing that I still can't seem to wrap my head around, but the `alt` field definitely does not have to be _populated,_ and if it's not populated, then we have no reason to return it.

cfgov/v1/models/learn_page.py
```diff
@@ -339,15 +339,15 @@ class EventPage(AbstractFilterPage):
         help_text='Check if this event will be streamed live. This causes the '
                   'event page to show the parts necessary for live streaming.'
     )
-    live_stream_url = models.URLField(
-        "URL",
+    live_video_id = models.CharField(
```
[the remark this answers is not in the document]
chosak: Unfortunately we'd only be able to consolidate the two uses in EventPage, since the use in VideoPlayer is of a different type (a Wagtail StreamField block type vs. a Django field type). Since these two are right near each other, I'm inclined to let it be until we need it [in a third place](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

cfgov/v1/atomic_elements/atoms.py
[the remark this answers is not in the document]
chosak: We want to return it because it should be the one place that defines the logic for what an image's alt text is. We're not just returning the value, we're saying that we defer the logic for the alt text to that method. And, because of the way CFGOVImage is defined, that can never be null and will always default to the empty string.

Actually thinking about this some more, and your other question [here](https://github.com/cfpb/cfgov-refresh/pull/5781#discussion_r442510301), I think I'd prefer to back this out to

```py
alt = self.get('alt')
if alt:
    return alt

upload = self.get('upload')
if upload:
    return upload.alt

return None
```

(See latest commit.)

The logic should be:

1. If the block has an `alt` defined, return it.
2. Otherwise, return the image's `alt` <-- this defers to the image logic about handling alt text, which encapsulates falling back to the empty string.
3. In the case where `upload` is undefined, its alt text should be undefined. We shouldn't  ever be putting a block like this into a template.

cfgov/v1/atomic_elements/atoms.py
```diff
@@ -112,6 +112,33 @@ class Button(Hyperlink):
...
+
+        if upload:
+            return upload.get_rendition('original').url
+
+    @property
+    def alt_text(self):
+        # TODO: This duplicates the logic in v1.jinja2tags.image_alt_value,
+        # which cannot be called here because of a circular import. It would
+        # be better to deprecate the image_alt_value tag in favor of using
+        # this logic wherever we use ImageBasic atoms.
```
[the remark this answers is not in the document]
chosak: For renditions there's no need for `image_alt_value` since we can already rely on [`rendition.alt`](https://github.com/cfpb/cfgov-refresh/blob/7c95ee016c20480a6085031ce9685fb063309cb8/cfgov/v1/models/images.py#L109), which calls [`CFGOVImage.alt`](https://github.com/cfpb/cfgov-refresh/blob/7c95ee016c20480a6085031ce9685fb063309cb8/cfgov/v1/models/images.py#L109), which returns the right thing or the empty string if left empty. See other comment below.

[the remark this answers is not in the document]
Scotchester: If that's the case, I'm not 100% sure why we needed `image_alt_value` in the first place. Feels like we need to revisit/reboot our image handling, in general.

cfgov/jinja2/v1/_includes/organisms/video-player.html
chosak: @Scotchester given how unwieldy this PR is already, and some of the questions you raise, I'm inclined to punt on doing this refactoring right now. Perhaps we can revisit with @anselmbradford after this gets merged in.

cfgov/v1/atomic_elements/organisms.py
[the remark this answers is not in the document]
chosak: I suppose, but what would the advantage be? The user experience in Wagtail would be significantly worse, both visually and because the max count isn't respected on the frontend, and it'd make the block storage less obvious.

chosak:
@Scotchester I believe I've addressed all of your (excellent) comments.

> Would you please add tests for the missing coverage?

I've added missing backend tests. As things stand we don't have any JS tests for VP and FCM, and the existing JS tests for the YouTube API were pretty minimal to start with. I'd need some help writing new tests for all of these changes. If you're available to help with that, that'd be great, otherwise I'd propose moving ahead with existing testing in the interest of getting this large PR in.

cfgov/v1/atomic_elements/organisms.py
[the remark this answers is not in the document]
Scotchester: The advantage would be being able to require `video_id` everywhere that a VideoPlayer block is added to a page, including within FCMs. It's arguable whether that UX improvement is offset by the issues you note with that approach, but given that it would also eliminate this custom clean complexity, I would argue that it's better.

Scotchester:
> I've added missing backend tests. As things stand we don't have any JS tests for VP and FCM, and the existing JS tests for the YouTube API were pretty minimal to start with. I'd need some help writing new tests for all of these changes. If you're available to help with that, that'd be great, otherwise I'd propose moving ahead with existing testing in the interest of getting this large PR in.

I don't think I have the time to help with this soon. If Ans or Chris do, great, but if not I'm OK with moving forward without them.

chosak:
@Scotchester thanks; in that case could I please get an approval?

PR #5790, opened 2020-06-16
Downgrade terser-webpack-plugin

anselmbradford:
## Changes

- `terser-webpack-plugin` to handle issue in our staging environments where node 12 is not supported.

## How to test this PR

1. Check that GitHub actions pass and this deploys in mattermost.


anselmbradford:
https://github.com/cfpb/cfgov-refresh/pull/5784 made this unnecessary

PR #5793, opened 2020-06-16
Processing Reports as Webpages - first pass

schbetsy:
❗ Note: This PR is for the Reports as Webpages project and is not going into Master! If merged, it will go into the reports-as-webpages branch. ❗

## Additions

- New tab in Wagtail editor view on Reports, called Report Upload. This will separate fields the user has to actively edit (e.g. uploading the document) from ones they don't (e.g. report contents that get parsed automatically from the document)
- New automated report process that currently only works in a local environment. (A version that works on Build is coming up soon). 
    - When a user saves a page, a wagtail hook triggers the `parse_report_file` method.
    - `parse_report_file` checks whether to parse the report, and what environment cf.gov is running in
    - If it's running on local or a DEV server (🤞 haven't tested on dev servers' yet, but theoretically it should work?), run the management command (`parse_research_report`) locally
    - `parse_research_report` uses Pandoc to parse the report file, then saves the results into the `footnotes` field for now

## Changes

- In the template: mark `page.footnotes` as `safe` so it renders that section as html
- Set all report content fields as not required. Otherwise the user is prevented from saving the page to kick off the report processing.

## How to test this PR

1. Run this branch, refresh database or migrate
2. sign in to wagtail admin
3. create a page of type `Report`
4. add a title, upload a word document of a report, check the "Process report" checkbox, press save. (wait ~5 seconds)
5. see that the html version of the document was inserted into the footnotes field
6. preview the page if you want

## Notes and todos

Future PRs:
- make it work on Beta and Prod
- actually parse the report into the correct report fields


cfgov/research_reports/models.py
```diff
@@ -47,6 +51,10 @@ def get_researchers():
     ])
 
 
+def _get_deploy_environment():
+    return getattr(settings, 'DEPLOY_ENVIRONMENT', None)
```
wpears, line 55: This isn't set my default in the local docker setup. This could be set or the `getattr` call could default to `'local'` instead of `None`

wpears, line 55:
```suggestion
    return getattr(settings, 'DEPLOY_ENVIRONMENT', 'local')
```

wpears: `DEPLOY_ENVIRONMENT` isn't set in local docker, we can use the above to provide local when it's unset

PR #5796, opened 2020-06-17
Deprecate Wagtail welcome page

chosak:
This page can currently only be seen by those users with Django admin permission (developers and admins). Those users don't always get sent to the page upon logging in, but only when logging in with certain `?next=/something/` parameters in the login URL.

Besides that, the content and design of this page has some issues, so we've decided to remove it. This PR does so.

See platform#3800 for background.

## How to test this PR

1. Login and logout and confirm that the normal flow works correctly.
2. When logged out, try to visit a page like http://localhost:8000/admin/template_debug/jobmanager/job_listing_list/. You should get redirected to http://localhost:8000/login/?next=/admin/template_debug/jobmanager/job_listing_list/, and then, upon logging in, go to the right page.
3. When logged out, try to visit a page like http://localhost:8000/admin/pages/319/edit/. You should get redirected to http://localhost:8000/login/?next=/admin/pages/319/edit/, and then, upon logging in, get redirected to http://localhost:8000/admin/.  <--  this is the case affected by this PR! Previously this is how you would end up at /login/welcome/. This is actually a bug, I think, as there's no reason why we shouldn't get to the right page, but I am deliberately not addressing that in this PR.

## Checklist

  - Consider prefixing with the feature/section you're addressing, e.g., "Mega Menu: fix layout bug" or "Paying for College: Update content on Repay tool"


cfgov/templates/wagtailadmin/login.html
```diff
@@ -25,9 +25,9 @@ <h1>Log in to consumerfinance.gov</h1>
             {% block login_form %}
             {% csrf_token %}
 
-            {% url 'user_welcome' as home_url %}
```
chosak, line 28: I removed this because we have no `user_welcome` URL and so this was always null, and thus never used.

cfgov/cfgov/settings/base.py
```diff
@@ -439,7 +439,7 @@
 LOGIN_FAIL_TIME_PERIOD = os.environ.get("LOGIN_FAIL_TIME_PERIOD", 120 * 60)
 # number of failed attempts
 LOGIN_FAILS_ALLOWED = os.environ.get("LOGIN_FAILS_ALLOWED", 5)
-LOGIN_REDIRECT_URL = "/login/welcome/"
+LOGIN_REDIRECT_URL = "/admin/"
 LOGIN_URL = "/login/"
```
Scotchester, line 443: Are these two settings even needed anymore?

cfgov/templates/wagtailadmin/login.html
```diff
@@ -25,9 +25,9 @@ <h1>Log in to consumerfinance.gov</h1>
...
-            <input type="hidden" name="next" value="{{ next|default:home_url }}" />
-
+            {% if next %}
+            <input type="hidden" name="next" value="{{ next }}"/>
```
Scotchester, line 29: Can we get rid of this override template entirely? If I'm right that the alternate `h1` text for build vs. other servers is the only thing we're still customizing here, it hardly seems worth it to me.

But, if we want to stick with it for now, why not revert these lines to what currently exists in this file in Wagtail 2.5.2?

```suggestion
            {% url 'wagtailadmin_home' as home_url %}
            <input type="hidden" name="next" value="{{ next|default:home_url }}" />
```

cfgov/v1/tests/views/test_login_views.py
```diff
@@ -63,7 +63,7 @@ def test_check_permissions_no_auth(self):
     def test_check_permissions_next_url_does_not_exist(self):
         self.client.login(username='admin', password='admin')
         response = self.client.get('/login/check_permissions/?next=/badurl/')
-        self.assertRedirects(response, '/login/welcome/')
+        self.assertRedirects(response, '/admin/')
 
     def test_check_permissions_next_permissions_problem(self):
```
Scotchester, line 68: Shouldn't these two tests and the one above (that I can't select to comment on) be covered by stock Wagtail test coverage?

cfgov/templates/wagtailadmin/login.html
chosak: This is a nice thought. It would be good to get rid of this override, but I don't think we can right now, for at least two reasons:

1. The default template [POSTs to the `wagtailadmin_login` view](https://github.com/wagtail/wagtail/blob/ce6b2fe97b6fada507741c6221493c6d4759afe8/wagtail/admin/templates/wagtailadmin/login.html#L29), and we override that to [instead POST to our own `cfpb_login` view](https://github.com/cfpb/cfgov-refresh/blob/1b7b9b64d70753f74487bf5f848cbed9d198cc21/cfgov/templates/wagtailadmin/login.html#L24). (It would also be good to remove this view and find other, less custom, ways to do our password/lockout validation, for example something like [django-axes](https://github.com/jazzband/django-axes).)
2. We make use of [`form.non_field_errors`](https://github.com/cfpb/cfgov-refresh/blob/1b7b9b64d70753f74487bf5f848cbed9d198cc21/cfgov/templates/wagtailadmin/login.html#L9) to show messages about [failed logins](https://github.com/cfpb/cfgov-refresh/blob/1b7b9b64d70753f74487bf5f848cbed9d198cc21/cfgov/v1/auth_forms.py#L106) and [expired passwords](https://github.com/cfpb/cfgov-refresh/blob/1b7b9b64d70753f74487bf5f848cbed9d198cc21/cfgov/v1/auth_forms.py#L121). The default template [doesn't allow such customization](https://github.com/wagtail/wagtail/blob/ce6b2fe97b6fada507741c6221493c6d4759afe8/wagtail/admin/templates/wagtailadmin/login.html#L17-L22), although it does support use of Django messages, so potentially we could use those instead.

Modifying our form to use the default next URL from Wagtail upstream seems reasonable, though. I've added a commit that does that.

cfgov/v1/tests/views/test_login_views.py
chosak: Unfortunately, no, because the [`check_permissions`](https://github.com/cfpb/cfgov-refresh/blob/1b7b9b64d70753f74487bf5f848cbed9d198cc21/cfgov/v1/views/__init__.py#L129) view is something custom to cfgov-refresh (although, like our custom login view, it would be good to look at deprecating this).

cfgov/cfgov/settings/base.py
chosak: Yes, not least because we refer to them in our custom [`check_permissions`](https://github.com/cfpb/cfgov-refresh/blob/1b7b9b64d70753f74487bf5f848cbed9d198cc21/cfgov/v1/views/__init__.py#L129) view. These differ from the [Django default values](https://docs.djangoproject.com/en/3.0/ref/settings/#login-redirect-url) for these settings.

cfgov/locale/es/LC_MESSAGES/django.po
```diff
@@ -518,10 +518,6 @@ msgstr ""
 msgid "Delete"
 msgstr ""
 
-#: templates/welcome.html:7
```
higs4281, line 521: Thorough.

chosak:
> Would reeeeeeally like to get the broken next redirect bug working, though. (I assume it's related to our custom check_permissions?

Yes. I would also like to remove this custom check entirely.

PR #5803, opened 2020-06-22
Replace use of "whitelist" with "allowlist"

willbarton:
This change replaces our use of the term "whitelist" with "allowlist" where we can. It also changes a comment to prefer the term "blocklist" to "blacklist".

"Allowlist" and "blocklist" have a clarity of meaning for which "whitelist" and "blacklist" rely on metaphor, and metaphor can encode in thought and language both historic and current forms of domination. This is an attempt to rid our codebase of two of those metaphors.

## Checklist


cfgov/regulations3k/scripts/ecfr_importer.py
```diff
@@ -543,10 +543,10 @@ def ecfr_to_regdown(part_number, file_path=None):
     DIV9 is an appendix
     DIV9 element whose HEAD starts with 'Supplement I' is an interpretation
 
-    To avoid mischief, we make sure the part number is on a whitelist.
+    To avoid mischief, we make sure the part number is on a allowlist.
```
higs4281, line 546:
```suggestion
    To avoid mischief, we make sure the part number is on an allowlist.
```

PR #5816, opened 2020-06-24
Add GitHub Actions workflow for their CodeQL analysis feature

Scotchester:
Let's see if this is useful, particularly for static analysis looking for potential security issues.

Documentation about this feature: https://help.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/automatically-scanning-your-code-for-vulnerabilities-and-errors

## Checklist


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,34 @@
+name: "Code scanning - action"
+
+on:
+  pull_request:
+  schedule:
+    - cron: '0 8 * * *'
```
higs4281, line 6: We generate a lot of PRs. Do we need a cron?

willbarton: I kind of agree — I think the benefit of doing this in GitHub is to run it on PRs.

Scotchester: It was part of the template for this action, but yeah, I don't really see how new issues could creep in under-the-radar. I'll remove it.

willbarton:
I'd still like to see it fail. Can we insert some code that *should* fail it? @laminfaty?

Scotchester:
> I'd still like to see it fail. Can we insert some code that _should_ fail it? @laminfaty?

Maybe some of the examples screenshotted in this GitHub help doc could give us some inspiration for inserting a fake error into our code: https://help.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/managing-alerts-from-code-scanning

willbarton:
@Scotchester let's not make testing of failures block getting this PR in. Instead, I think maybe that's a follow-up task that's necessary before we make the check required for PRs.

Scotchester:
I think we may need to add the `push` trigger back to this workflow. I can't seem to get the results I saw before anymore, and I think it may be related to how the UI is built around branch-based data?

Scotchester:
Here's what the default screen looks like. Presumably it's referring to the main branch:

PR #5817, opened 2020-06-25
Fix middleware test by starting with clean state

higs4281:
I'm not sure what changed, but our translation middleware tests are failing in master because of a state issue. 

~~If we start the test with a clean slate by deactivating all translations, the test can't be tripped up by another test.~~

Since the test is checking deactivation, it would make more sense to explicitly set the language first, rather than first deactivating.


cwdavies:
Thanks for fixing translation in the middleware

PR #5819, opened 2020-06-26
Allow filtering of Enforcement Actions by Enforcement Status

wpears:
- Adds status to the filterable list control and FilterForm
  - Updates statuses to be url friendly
  - Includes a script to update old statuses to the friendly format
  - adds status to the special-cased filterable_list util logic

Closes <jira>/INTAKE-3864

## Testing

- Pull & migrate
- Check the "Filter Enforcement Statuses" checkbox on the enforcement actions filter page in wagtail
- run the included `resave_enforcement` script with `./manage.py runscript resave_enforcement`
- note the inclusion of statuses in the filterable list control on the enforcement actions page
- Filtering should work as expected
- Statuses should still display as expected on any given enforcement action page

Once this is merged, we'll need to do the above steps in prod as well

Screen Shot 2020-06-25 at 5 16 21 PM


caheberer:
Overall, I think the design looks good, but two questions:

@ielerol maybe you can weigh in?
1. Considering in our other filter instances on cf.gov "Topic" and "Date range" are stacked on the right, do you think it's beneficial for users to stick with that convention? (So in our case putting "Status" below "Category")
2. I'm wondering if we want to alphabetize the items for both category and status? This doesn't seem to be standard in many of our filters, but in this instance, I don't see a strong reason for not doing it.

wpears:
@caheberer What do you think about this:
Screen Shot 2020-06-29 at 9 02 28 AM
Preserves the normal positioning of Topic and the date ranges while cutting the rows down to two

Could also reverse the order of topics and statuses, which I just personally find more appealing:
Screen Shot 2020-06-29 at 9 06 24 AM

caheberer:
After chatting with Wyatt about the complexities of having status below category, I'm now onboard with the second option, status to the right of category. @ielerol agree with that?

ielerol:
@caheberer @wpears yes, that second option with category-status-topic looks good to me. Also big thumbs up to alphabetizing the category and status lists.

caheberer:
@wpears one last thing I saw, when betsy shared this during review, it looked like there wasn't much bottom padding to the filter box, according to the design system, that should be 30px.

wpears:
Talking with Candice, it looks like 15px is currently implemented sitewide, so Candice will bring it up to El Camino that the design manual isn't reflecting our implementation.

cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
```diff
@@ -46,10 +46,27 @@
...
+                {% if value.statuses %}
+                    <div class="content-l_col
+                                content-l_col-1-2">
+                        <div class="o-form_group">
+                            <fieldset class="o-form_fieldset">
+                                <legend class="a-legend">
+                                   Status
+                                </legend>
+                                <ul class="m-list m-list__unstyled">
+                                {% for slug, name in form.statuses.field.choices %}
```
schbetsy, line 62: If I erroneously check the "Filter Enforcement Statuses" checkbox on a different filterable list page (e.g. the Blog page), this line errors because `form.statuses` doesn't exist. Do you think we should put a guard on this section, even though that situation shouldn't happen?

cfgov/v1/forms.py
```diff
@@ -262,6 +269,7 @@ def get_query_strings(self):
             'categories__name__in',  # categories
             'tags__slug__in',        # topics
             'authors__slug__in',     # authors
+            'statuses__status__in',  # statuses
```
schbetsy, line 272: I'm trying to think of a situation where this-- having a different number of query parameters for this form class-- might cause a problem. Maybe if the `status` checkbox is checked on a form that's not an EnforcementActionsFilterPage? I haven't dug into it.

cfgov/jinja2/v1/_includes/molecules/enforcement-action-metadata.html
```diff
@@ -54,7 +54,7 @@ <h3 class="h4">
         {% for action_status in page.statuses.all() %}
             <div>
                 <h6>{{_text(action_status.institution)}}</h6>
-                {{_text(action_status.status)}}
+                {{_text(action_status.get_status_display())}}
```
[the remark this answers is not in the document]
wpears: [It's a special django instance method](https://docs.djangoproject.com/en/3.0/ref/models/instances/#django.db.models.Model.get_FOO_display)

cfgov/v1/forms.py
wpears: Well the `EnforcementActionsFilterForm` is only used on instances of `EnforcementActionsFilterPage`, so I think this is safe.

cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
wpears: Fixed in 5cb5d9

cfgov/jinja2/v1/_includes/molecules/enforcement-action-metadata.html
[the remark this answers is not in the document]
schbetsy: Huh. Thanks.

cfgov/v1/migrations/0226_update_enforcement_action_data.py
```diff
@@ -0,0 +1,42 @@
...
+
+        if not page.live:
+            continue
+        if 'policy-compliance/enforcement/actions' not in url:
+            continue
+        if page.has_unpublished_changes:
+            continue
+
+        existing_status = enforcement_status.status
+        enforcement_status.status = get_value(existing_status)
```
schbetsy, line 26: This PR updated the `EnforcementActionStatus` model to use different underlying values for the `status` field, while the display names stayed the same (new values are listed in `v1/util/ref.py`). Now, an object that had the status value `Post Order/Post Judgment` in the database should have `post-order-post-judgment` instead.

Seems to me that a data migration is the simplest and most reliable way to make that update, but I'd like a second opinion. @chosak or anyone else, does this seem reasonable to you?

cfgov/v1/migrations/0226_update_enforcement_action_data.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def get_value(label):
+    for status in enforcement_statuses:
+        if label == status[1]:
+            return status[0]
+
+
+def forward(apps, schema_editor):
+    for enforcement_status in EnforcementActionStatus.objects.all():
```
chosak, line 14: Per the documentation [here](https://docs.djangoproject.com/en/3.0/topics/migrations/#data-migrations), best practice when accessing Django model classes in a migration is to use something like this:

```py
EnforcementActionStatus = apps.get_model('v1', 'EnforcementActionStatus')
```

instead of importing at the top of the module as you do here. The reason for this is that you want to make sure that the code is accessing the "historical" model state at the time this migration was written.

cfgov/v1/migrations/0226_update_enforcement_action_data.py
chosak: A data migration seems reasonable to me.

One potential performance optimization would be to do something like `EnforcementActionStatus.objects.prefetch_related('action')` -- this would retrieve all status pages from the DB in a single query, instead of doing one page query per status in the loop.

Thinking further, is there any reason why you don't want to bulk update _all_ EASes, including any related to unpublished pages or pages elsewhere in the tree? That could maybe be accomplished with a simpler set of commands using [`bulk_update`](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#bulk-update), something like (untested):

```py
for before, after in (
    ('Expired/Terminated/Dismissed', 'expired-terminated-dismissed'),
    ('Pending Litigation', 'pending-litigation'),
    ('Post Order/Post Judgment', 'post-order-post-judgment'),
):
    statuses = EnforcementActionStatus.objects.filter(status=before)
    for status in statuses:
        status.status = after

     # optionally include batch_size here if there are many statuses.
    EnforcementActionStatus.objects.bulk_update(statuses, ['status'])
```

cfgov/v1/migrations/0226_update_enforcement_action_data.py
schbetsy: Gooood to know. Thanks for the explanation.

In that case, would it also make sense to remove the reference to `v1.util.ref.enforcement_statuses`? I could just... make a dictionary that's local to this file that's a mapping of the values I want to update?

cfgov/v1/migrations/0226_update_enforcement_action_data.py
schbetsy: Great feedback, thanks. I'll try this out.

schbetsy: We talked it through and decided that `bulk_update` is a good option here, so I updated my migration to use it. Thanks for the excellent suggestion.

cfgov/v1/migrations/0226_update_enforcement_action_data.py
```diff
@@ -0,0 +1,31 @@
...
+from django.db import migrations
+
+
+def forward(apps, schema_editor):
+    EnforcementActionStatus = apps.get_model('v1', 'EnforcementActionStatus')
+    for existing, updated in (
+        ('Expired/Terminated/Dismissed', 'expired-terminated-dismissed'),
+        ('Pending Litigation', 'pending-litigation'),
+        ('Post Order/Post Judgment', 'post-order-post-judgment'),
+    ):
```
wpears, line 11: This is probably unnecessary, but I like the idea of importing the list of statuses & labels from `v1/util/ref.py` and flipping the order in this for loop to `updated, existing`. That was we aren't duplicating the raw data/it's impossible to get bitten by a typo here (even though it looks typo-free to me).

schbetsy: Theoretically, if we want this data migration to continue to work correctly in the future, we would need to *not* refer to `v1/util/ref.py`, since that file might change.

I feel pretty confident there are no typos and all records get updated correctly. When I tested locally, the total number of `EnforcementActionStatus` objects in the database was equal to the number of results that come up when searching for `expired-terminated-dismissed`, `pending-litigation`, and `post-order-post-judgment`.

```
cfgov=> select count( distinct action_id ) from v1_enforcementactionstatus;
 count
-------
   245
```

and

Screen Shot 2020-07-14 at 10 37 12 AM

PR #5822, opened 2020-06-29
Adjust notification tester and comments

higs4281:
We've had to check our PFC notifications a lot in the past few days. These changes will make that easier.

Originally we got back json responses from schools, but they stopped doing that, and the absence was obscuring test success.

## Testing

The script can only be fully tested from a productions server, because otherwise your outbound IP may be blocked.  
However, you can test it locally against httpbin, and you can see it time out for requests to schools that use a blocklist.

To test:

```bash
./cfgov/manage.py shell
```

```python
from paying_for_college.disclosures.scripts.notification_tester import *
print(send_test_notifications(url=BINPOST))
print(send_test_notifications())
```

- The httpbin request should succeed, and the response's 'origin' field should reflect your outbound IP.
- The bare `send_test_notifications()` should return at least one time-out message; one may succeed.

### DIFF
--- a/cfgov/paying_for_college/disclosures/scripts/notification_tester.py
+++ b/cfgov/paying_for_college/disclosures/scripts/notification_tester.py
@@ -18,53 +18,62 @@ BINGET = "https://httpbin.org/get"
 OID = '9e0280139f3238cbc9702c7b0d62e5c238a835d0'
 ERRORS = 'INVALID: test notification via Python'
 REPORT = ('URL is {}\nOK is {}\nReason is {}\n'
-          'Status is {}\nTime sent is {}\nHeaders are {}\n'
-          'Origin is {}')
+          'Status is {}\nTime sent is {}')
 
 
-def send_test_notification(url, oid, errors):
+def send_test_notifications(url=None, oid=OID, errors=ERRORS):
     """
-    Send fake notifiations to various endpoints to help troubleshoot.
+    Send fake notifiations to school endpoints to help troubleshoot.
 
-    You can send a test notification to a school's endpoint to confirm
+    You can send test notifications to school endpoints to confirm
     reception, or send to httpbin to check the outbound IP (origin),
     headers, and the data payload.
     Examples:
-    - print(send_test_notification(EDMC_PROD, OID, ERRORS))
-    - print(send_test_notification(BPI_PROD, OID, ERRORS))
-    - print(send_test_notification(BINPOST, OID, ERRORS))
+    - print(send_test_notifications())
+    - print(send_test_notifications(url=BINPOST))
     """
     prod_endpoints = {
         'edmc': Contact.objects.get(name='EDMC').endpoint,
         'bpi': Contact.objects.get(name='Bridgepoint Education').endpoint
     }
-    if url in prod_endpoints:
-        _url = prod_endpoints[url]
+    if not url:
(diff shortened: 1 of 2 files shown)


cfgov/paying_for_college/disclosures/scripts/notification_tester.py
```diff
@@ -18,53 +18,62 @@
...
-          'Status is {}\nTime sent is {}\nHeaders are {}\n'
-          'Origin is {}')
+          'Status is {}\nTime sent is {}')
 
 
-def send_test_notification(url, oid, errors):
+def send_test_notifications(url=None, oid=OID, errors=ERRORS):
     """
-    Send fake notifiations to various endpoints to help troubleshoot.
+    Send fake notifiations to school endpoints to help troubleshoot.
```
chosak, line 26:
```suggestion
    Send fake notifications to school endpoints to help troubleshoot.
```

cfgov/paying_for_college/disclosures/scripts/notification_tester.py
```diff
@@ -18,53 +18,62 @@
...
+"""
+curl -v https://httpbin.org/get
+curl -v -X POST https://httpbin.org/post
+
+export XDATA="oid='f38283b5b7c939a058889f997949efa566c616c5'&errors='INVALID: test notification via curl'&time='2020-06-24T18:36:09.922690+00:00'". # noqa
+export BPI="https://sissecureservices.bpiedu.com/routingmanager/api/cfpb"
+export EDMC="https://exml.dcedh.org/cfpb"
+
+curl -v -X POST --data "$XDATA" --url "$BPI"
+curl -v -X POST --data "$XDATA" --url "$EDMC"
```
chosak, line 78:
```suggestion
curl -v --data "$XDATA" --url "$BPI"
curl -v --data "$XDATA" --url "$EDMC"
```

Minor: `-X POST` is not needed if you use `--data` as it POSTs by default

cfgov/paying_for_college/disclosures/scripts/notification_tester.py
```diff
@@ -18,53 +18,62 @@
...
+# curl_examples
+"""
+curl -v https://httpbin.org/get
+curl -v -X POST https://httpbin.org/post
+
+export XDATA="oid='f38283b5b7c939a058889f997949efa566c616c5'&errors='INVALID: test notification via curl'&time='2020-06-24T18:36:09.922690+00:00'". # noqa
+export BPI="https://sissecureservices.bpiedu.com/routingmanager/api/cfpb"
+export EDMC="https://exml.dcedh.org/cfpb"
+
+curl -v -X POST --data "$XDATA" --url "$BPI"
```
chosak, line 77: With the BPI endpoint I get a `certificate has expired` error if I don't also pass `-k`, which is surprising because the browser reports that https://sissecureservices.bpiedu.com/ has a good certificate.

If I use `-k`, I generate a 500 error with an error: `Message:String was not recognized as a valid DateTime. Inner:`.

I'm not able to connect to the EDMC endpoint from my local network (as you suggest might happen).

higs4281: I get the same certificate error when using curl locally, but when using requests, I don't get the error.

Many thanks for the review.

PR #5825, opened 2020-06-29
Paying for College: Add program codes to API

higs4281:
Now that we have program-level data from College Scorecard, we can
provide a detailed list of available programs to the financial-path tool
so that students can find their specific degree.

## How to test this PR

With a recent production database dump, check out the branch and check the
school API endpoint for George Washington University.

<http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/api/school/131469/>

You should see a new value called programCodes with entries like this:

```python
{
    "graduate": [
        {
            "code": "501-5",
            "name": "Area Studies",
            "level": "Master's degree",
            "salary": 51700,
        },
        {
            "code": "909-5",
            "name": "Public Relations, Advertising, and Applied Communication",
            "level": "Master's degree",
            "salary": 61800,
        }
    ]
    "undergrad": []
}
```

## Note

We have only fetched this detailed data for GWU.

- [x] Project documentation has been updated:  
[GHE]/paying-for-college/dynamic-disclosures-documentation/wiki/API-documentation#json-result-1


cfgov/paying_for_college/models/disclosures.py
```diff
@@ -389,6 +390,22 @@ def as_json(self):
     def __str__(self):
         return self.primary_alias + " ({})".format(self.school_id)
 
+    @property
+    def program_codes(self):
```
willbarton, line 394: Could we have some tests for this method? I suspect coverage is passing because it's accessed in a call to `as_json` below, but I just want to make sure it's covered so that we don't break it sometime in the future.

higs4281: √

higs4281: @willbarton Test added.

PR #5550, opened 2020-03-05
Turn on new mega menu everywhere

chosak:
This change turns on the new mega menu (previously referred to as "variation 1"), by removing the `MEGA_MENU_VAR_1` flag and switching to the new code everywhere.

Ping @contolini.

## Testing

`gulp clean && gulp build` to rebuild the frontend. Then visit any page and test out the new mega menu.

CFPB folks can also test drive this at DEV4.

## Notes

Git is showing the diff as changes to the existing files even though what I actually did was mostly replace the existing design with the -var-1 files.

## Todos

1. The nonresponsive pages (like http://localhost:8000/paying-for-college/) show the menu properly, but it's at a more narrow width for some reason:
   narrow
2. Similarly, on Spanish pages the menu appears narrower because there's no caret; should we add spacing here to keep this consistent?
   image
3. A minor bug that users are unlikely to encounter: if you open the menu at desktop width, then narrow your browser to mobile width, and then try to open the menu there, you get this error in the console:
   image
4. The external link icon on "Order Free Brochures" is too close to the text; that is to be fixed in a separate PR.
   image

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [x] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #5843, opened 2020-07-13
Mega Menu: Fix issue with updating menu state

anselmbradford:
https://github.com/cfpb/cfgov-refresh/pull/5770 neglected the case where the user resized without ever opening a menu, which meant `_activeMenu` was null and was throwing an error when its state was updated. This PR adds logic to check that `_activeMenu` has a value before resetting its state.

Additionally, the new mega menu added in https://github.com/cfpb/cfgov-refresh/pull/5550 does not use `type` in the `_updateMenuState` method any more, so that parameter has been removed in this PR.

## Changes

- Checks `_activeMenu` value before resetting its state.
- Removes unused `type` parameter from `_updateMenuState` method.

## How to test this PR

1. Pull branch and build.
2. View the site at desktop size and resize to mobile without opening the menu. There should be no JS error (this will produce an error on production currently).
3. View the site at desktop size, click the menu, and resize to mobile. There should be no JS error.

### DIFF
--- a/cfgov/unprocessed/js/organisms/MegaMenuDesktop.js
+++ b/cfgov/unprocessed/js/organisms/MegaMenuDesktop.js
@@ -75,7 +75,7 @@ function MegaMenuDesktop( baseClass, menus ) {
     this.dispatchEvent( 'triggerClick', { target: this } );
     const menu = event.target;
     if ( menu.isAnimating() ) { return; }
-    _updateMenuState( menu, event.type );
+    _updateMenuState( menu );
   }
 
   /**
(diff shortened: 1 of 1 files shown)


anselmbradford:
> Is it expected that I see the mobile menu open for a flash and animate closed when resizing to under 900px?

Yes. Back in the day of the original release of the new menu we determined it was not worth fixing this, though I can take a look if it might be easier to fix with the new menu. Separate issue to this PR though.

> Is it expected that if I have the menu open on desktop and resize to under 900px that it does not "remain" open (i.e., open the mobile version of the menu to the same place)?

Yes. The state of the menu resets when moving between screen sizes. Retaining state across both sizes would need to be a new feature, and would increase the complexity and testing scenarios since the desktop and mobile menus operate in different ways.

PR #5844, opened 2020-07-14
Remove the comparisontool URL includes

higs4281:
In order to let Wagtail take over serving the paying-for-college
landing page, we need to remove the URL entry that sent PFC requests 
to comparisontool.

We need to retain three downstream comparisontool pages for now, so
this patch adds explicit URL entries for those pages.

## Testing
After checking out the branch, make sure these pages render properly:
- http://localhost:8000/paying-for-college/ should show the new wagtail landing page:

pfc_new

These three pages should be unchanged, and show a header with pencils:
- http://localhost:8000/paying-for-college/repay-student-debt/
- http://localhost:8000/paying-for-college/choose-a-student-loan/
- http://localhost:8000/paying-for-college/manage-your-college-money/

pfc_old

### DIFF
--- a/cfgov/cfgov/urls.py
+++ b/cfgov/cfgov/urls.py
@@ -269,7 +269,27 @@ urlpatterns = [
         'transcripts'),
         namespace='transcripts')),
 
-    re_path(r'^paying-for-college/', include('comparisontool.urls')),
+    re_path(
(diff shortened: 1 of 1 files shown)


higs4281:
changing branch name

PR #5865, opened 2020-07-21
Pin cryptography temporarily to 2.9.2

willbarton:
This change temporarily pins cryptography to 2.9.2. Akamai's edgegrid-python has an open-ended dependency on PyOpenSSL, which in turn has an open-ended dependency on cryptography. cryptography 3.0 currently doesn't build in our build environment. This pins the last release that  does until we can solve that problem.

## Checklist


willbarton:
I've confirmed this branch builds successfully in Jenkins.

PR #5870, opened 2020-07-22
Add fake bad script to test code scan PR check

Scotchester:
We're trying to see what a failing code scan PR check looks like. Currently, this repo reports 14 warnings, but no errors. Hoping that opening this fake PR with code that should throw an error will show us a failing check.

This file is used as an example in [GitHub's code scanning docs](https://help.github.com/en/github/finding-security-vulnerabilities-and-errors-in-your-code/managing-alerts-from-code-scanning) of something that will trigger a code scan error (not just a warning).

### DIFF
--- a/spec-main/security-warnings-spec.ts
+++ b/spec-main/security-warnings-spec.ts
@@ -1,0 +1,232 @@
+import { expect } from 'chai';
+import * as http from 'http';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as url from 'url';
+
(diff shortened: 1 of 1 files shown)


willbarton:
image

This looks like we expect, right?

Scotchester:
> image
> 
> This looks like we expect, right?

Yes and no. Per the conversation I opened at github/codeql-action#122, it is working as intended but the UX could be better. But we've shown what we intended to show, so I think this can be closed.

PR #5871, opened 2020-07-23
Optionally enable Wagtail's Postgres full text search support and index PageRevisions

willbarton:
The upgrade to 2.5 removed an obstacle to indexed admin searching because it removed the naming clash between Wagtail and Haystack over the `update_index` command.

This version also makes available Postgres' full-text search as an option (using the "fulltext" backend), but not the default, so that we can do further iteration on Wagtail's search without disrupting current admin and external link search expectations.

This PR also adds a proxy model for `PageRevision` so that it can be searched as well. Because it's a separate Django model, and not a descendent of `Page`,  it will not clutter any `Page` searches, and would have to be explicitly searched.

## How to test this PR

To give it a try, check out the branch, migrate and index:

```
./cfgov/manage.py migrate
./cfgov/manage.py wagtail_update_index
./cfgov/manage.py shell
```

Then perform a querying the "fulltext" backend:

```python
from v1.models import IndexedPageRevision
from wagtail.core.models import Page
from wagtail.search.backends import get_search_backend

search_query = "sandbox"
page_results = Page.objects.search(search_query, backend="fulltext")
revision_results = get_search_backend("fulltext").search(search_query, IndexedPageRevision
```

## Checklist


cfgov/cfgov/settings/base.py
```diff
@@ -108,6 +110,32 @@
...
+    # An upcoming effort to overhaul search within consumerfinance.gov and
+    # Wagtail should address these issues. In the meantime, Postgres full text
+    # search with the custom search_fields defined on our models is available
+    # with the "fulltext" backend defined below.
+    'default': {
+        'BACKEND': 'wagtail.search.backends.db',
+    },
+    'fulltext': {
+        'BACKEND': 'wagtail.contrib.postgres_search.backend',
+    },
```
willbarton, line 136: There's some additional discussion in the PR that added the external link search, https://github.com/cfpb/cfgov-refresh/pull/4540, about this.

willbarton:
@higs4281 You are correct — I've corrected the text above in the PR.

willbarton:
Tagging @cfpb/back-end-admins for review here because I'm making a change to our Model Bakery bakes models in this repo:

https://github.com/cfpb/cfgov-refresh/pull/5871/files#diff-c3c4d577f5e4d155f04d914b5e05af7e
https://github.com/cfpb/cfgov-refresh/pull/5871/files#diff-0cf61f1e9bb958eff38a3fae3fd25a3fR8-R9

I think the comments and docstrings are explanatory, but in short, for Wagtail search not to cause occasional Model Bakery failures when creating Page models, the content type cannot be random. By default Bakery uses a random content type. This change creates a custom baker and configures it to be the default that always uses the real content type.

cfgov/v1/management/commands/search_pagerevisions.py
```diff
@@ -0,0 +1,60 @@
...
+from django.urls import reverse
+
+from wagtail.core.models import Page
+from wagtail.search.backends import get_search_backend
+
+from v1.models import IndexedPageRevision
+
+
+class Command(BaseCommand):
+    help = 'Search for a particular string within page revisions'
```
higs4281, line 13: It might be more helpful to say "search for words" within page revisions. A string might imply that you can search for partial words, or that an entire phrase is being searched for. Results I see return cases where any of the words entered appear, whether they're found together or not.

PR #5876, opened 2020-07-24
Validate our static assets in CI

willbarton:
We've had a `validate-assets` tox environment for ages that collects our assets and then runs Django's `collectstatic` on them. Because we use `ManifestStaticFileStorage` in production, if static assets are missing, this `collectstatic` will fail.

Because our webfonts are not public, we have not been able to run this in CI.  I've pulled in [a workaround for this] (https://github.com/cfpb/cfgov-refresh/pull/5110/files#diff-300249821afbbfcb945e6fbe141d0bb8R32-R55) created by @chosak in another context in #5110.

I think the `frontend.sh` file is right place for this to live, but I'm open to moving it.

This PR is in draft so we can see it run.

## Checklist

### DIFF
--- a/frontend.sh
+++ b/frontend.sh
@@ -89,7 +89,7 @@ fake_fonts() {
 
     # 1. Use grep to find all .woff and .woff2 files referenced in CSS files.
     # 2. Reduce these to a list of unique webfont filenames.
-    # 3. Touch each filename in static.in/0/fonts, causing it to be created as
+    # 3. Touch each filename in static.in/fake-fonts/fonts, causing it to be created as
     # an empty file if it doesn't exist already.
     grep -Eho \
         '[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12}.woff2?' \
--- a/frontend.sh
+++ b/frontend.sh
@@ -63,11 +63,42 @@ install() {
   fi
 }
 
-
 # Run tasks to build the project for distribution.
 build() {
   echo "Building project…"
   yarn run gulp build
+}
+
+# Fake our font files, if necessary, in public CI
+fake_fonts() {
+  if [ ! -d static.in/cfgov-fonts/fonts ]; then
+    echo "Faking font files…"
+    # We want to test Django collectstatic, but we might not have our webfont 
+    # files if this script is running somewhere public. Because cfgov-refresh
+    # uses the Django ManifestStaticFilesStorage backend, the collectstatic 
+    # command will fail if any referenced files are missing.
+    #
+    # Builds generated with webfont files place them in a static.in/0/fonts
+    # subdirectory. Builds generated without these files don't have this 
+    # directory.
+    #
+    # If we don't have the webfont files, we create empty files with the same 
+    # name to allow collectstatic to run successfully. If the files already 
+    # exist, these commands do nothing.
+    mkdir -p static.in/fake-fonts/fonts
+
+    # 1. Use grep to find all .woff and .woff2 files referenced in CSS files.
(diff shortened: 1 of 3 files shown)


Scotchester:
I'm not a huge fan of adding that `fake_fonts` script to `front-end.sh`, at least in this way. `front-end.sh` is currently assumed to always be:

1. Installing front-end deps (if necessary)
1. Running the front-end build

The new `fake_fonts` command does neither of those things.

(I realize that the `init` and `build` paths right above where you added the `fake_fonts` path contradict the above assertion, but I'm not aware of any scenarios in which we actually use those paths.)

I see two preferable options:

1. Create a dedicated file for the `font_files` script and run it on the same line in `tox.ini` where you added `{toxinidir}/frontend.sh fake_fonts`
2. Create a new environment option in `front-end.sh`, maybe `ci`, and run `front-end.sh ci` instead of `front-end.sh production` in `[testenv:validate-assets]`.

frontend.sh
```diff
@@ -63,19 +63,50 @@ install() {
...
 
+# Fake our font files, if necessary, in public CI
+fake_fonts() {
+  echo "Faking font files if necessary…"
+  # We want to test Django collectstatic, but we might not have our webfont 
+  # files if this script is running somewhere public. Because cfgov-refresh
+  # uses the Django ManifestStaticFilesStorage backend, the collectstatic 
+  # command will fail if any referenced files are missing.
+  #
+  # Builds generated with webfont files place them in a static.in/0/fonts
```
chosak, line 80: In the deployable zipfile case, I used `static.in/0/fonts` because our artifacts happen to use numbered subfolders of `static.in`. [Our documentation recommends](https://cfpb.github.io/cfgov-refresh/installation/#webfonts) putting the webfonts locally under `static.in/cfgov-fonts/fonts`. Would it make sense to use that subfolder instead?

Then, this script could check if the folder exists, and do nothing if so (which would be consistent with the "if necessary" in the logging statement above).

I guess one downside of doing this is that it might confuse people if they run this code to generate fake fonts and later wonder why the fonts don't look correct. So maybe it'd be better to more explicitly put the fake files under `static.in/fake-fonts/fonts` -- but still only do that if the real fonts don't exist in a sibling directory.

willbarton: @chosak I've taken your suggestion to use `static.in/fake-fonts/fonts` only if `static.in/cfgov-fonts/fonts` does not exist.

willbarton:
@Scotchester I've taken option 2, and added a "ci" option for `frontend.sh` that does the font-faking.

frontend.sh
```diff
@@ -63,19 +63,55 @@ install() {
...
+    # directory.
+    #
+    # If we don't have the webfont files, we create empty files with the same 
+    # name to allow collectstatic to run successfully. If the files already 
+    # exist, these commands do nothing.
+    mkdir -p static.in/fake-fonts/fonts
+
+    # 1. Use grep to find all .woff and .woff2 files referenced in CSS files.
+    # 2. Reduce these to a list of unique webfont filenames.
+    # 3. Touch each filename in static.in/0/fonts, causing it to be created as
```
higs4281, line 92:
```suggestion
    # 3. Touch each filename in static.in/fake-fonts/fonts, causing it to be created as
```

higs4281:
If I remove the `cfgov-fonts/fonts/` directory from `static.in/cfgov-fonts/fonts` and run `./frontend.sh ci`, should I expect to see a `static.in/fake-fonts/fonts` directory generated?

willbarton:
> If I remove the cfgov-fonts/fonts/ directory from static.in/ and run ./frontend.sh ci, should I expect to see a static.in/fake-fonts/fonts/ directory generated?

@higs4281 yes, you should.

PR #5885, opened 2020-07-27
Turn autocomplete off on multiselect's typeahead input

niqjohnson:
Right now, there's no autocomplete attribute set on the multiselect's typeahead text input, which means that a browser's autocomplete interface can pop up and obscure the multiselect checkboxes until you dismiss the autocomplete suggestions. Since the multiselect's text input is meant as a typeahead way to filter the options to find what you're looking for, turning the autocomplete off doesn't take away from the intent of that field.

See GHE/CFGOV/platform/issues/3568 for more details.

---

## Additions

- `autocomplete="off"` attribute to the multiselect's typeahead text input

## How to test this PR

1. In production (or locally on the main branch), visit a page with a multiselect. Any filterable list page with the topics field will do (the blog landing page, for example).
2. Enter a term in the multiselect text input but don't check any of the suggested items in the multiselect's checklist interface, and submit the form (hit "apply filters").
3. Refresh the page or visit another page with the same kind of multiselect.
4. Focus in the multiselect and note that the browser's autocomplete UI appears on top of the multiselect's checklist UI.
5. Repeat the same steps in this branch and make sure the autocomplete UI no longer pops up when focusing in the typeahead field.

## Screenshots

Production | This PR
---- | ----
production | pr

## Checklist

### DIFF
--- a/cfgov/unprocessed/js/molecules/Multiselect.js
+++ b/cfgov/unprocessed/js/molecules/Multiselect.js
@@ -149,12 +149,12 @@ function Multiselect( element ) { // eslint-disable-line max-statements
     } );
 
     _searchDom = create( 'input', {
-      className:      BASE_CLASS + '_search ' + TEXT_INPUT_CLASS,
-      type:           'text',
-      placeholder:    _placeholder || 'Choose up to five',
-      inside:         _headerDom,
-      id:             _name,
-      'autocomplete': 'off'
+      className:    BASE_CLASS + '_search ' + TEXT_INPUT_CLASS,
+      type:         'text',
+      placeholder:  _placeholder || 'Choose up to five',
+      inside:       _headerDom,
+      id:           _name,
+      autocomplete: 'off'
     } );
 
     _fieldsetDom = create( 'fieldset', {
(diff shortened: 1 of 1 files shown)


cfgov/unprocessed/js/molecules/Multiselect.js
```diff
@@ -149,11 +149,12 @@ function Multiselect( element ) { // eslint-disable-line max-statements
...
-      type:        'text',
-      placeholder: _placeholder || 'Choose up to five',
-      inside:      _headerDom,
-      id:          _name
+      className:      BASE_CLASS + '_search ' + TEXT_INPUT_CLASS,
+      type:           'text',
+      placeholder:    _placeholder || 'Choose up to five',
+      inside:         _headerDom,
+      id:             _name,
+      'autocomplete': 'off'
```
anselmbradford, line 157: Do you need the quotes around `autocomplete`?

niqjohnson: I guess not? I removed them in 4f12e8f, and everything still works fine. Not sure why some of the values are quoted and some aren't elsewhere in the file, but I guess it's better to stay consistent within a group of attributes.

anselmbradford: Yeah the other ones should be removed too. Quotes are only needed if it has a dash or space if I recall. I think the linter probably will note that too. At any rate, for the scope of this PR having them the same within a group is fine.

PR #5886, opened 2020-07-27
Improve front-end GitHub Actions workflow

Scotchester:
These changes should make our front-end CI checks more efficient.

## Changes

- Only run if a `.js` file or `package.json` is modified
- Also test the build
- Stop setting up Python and installing unnecessary dependencies for the main job
- Only install the one Python dependency we need for checking coverage

## How to test this PR

1. Wait for the checks to run on it!

## Notes and todos

- Would be nice if we had a tool for tracking our JS coverage that didn't require Python at all.

## Checklist


Scotchester:
Hmm… problem whose solution is not readily apparent: The existing required frontend lint and test checks will apparently never pass, because they are not run at all. Need to figure out if it's possible to only require a check if its workflow is run. Second best option would be returning an immediate success if the workflow is unnecessary.

willbarton:
This action might be a way forward for this: https://github.com/dorny/paths-filter#conditional-execution

willbarton:
Using dorny/paths-filter seems to work! This will help with some, but not all, of our checks, but it will also give us a pattern to use as we get more checks into GitHub Actions.

willbarton:
With `dorny/paths-filter`, I think this works. I've rebased it, and I think we should get it merged but I'd like a second opinion. @cfpb/back-end-admins @cfpb/front-end-admins

.github/workflows/backend.yml
```diff
@@ -37,35 +37,48 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - uses: dorny/paths-filter@v2
```
anselmbradford, line 40: I think I discovered in https://github.com/cfpb/consumerfinance.gov/pull/6645 that I don't think it's enough to do `@v2` and get the latest, but you'd have to do `@v2.10.2` https://github.com/dorny/paths-filter/releases

.github/workflows/backend.yml
```diff
@@ -80,22 +93,34 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
+      - uses: dorny/paths-filter@v2
```
anselmbradford, line 96: `dorny/paths-filter@v2.10.2`

.github/workflows/frontend.yml
```diff
@@ -12,32 +13,36 @@ jobs:
         gulp_cmd:
           - lint --ci
           - test:unit --ci --headless
+          - build
 
     steps:
       - uses: actions/checkout@v2
+      - uses: dorny/paths-filter@v2
```
anselmbradford, line 20: `dorny/paths-filter@v2.10.2`

.github/workflows/frontend.yml
```diff
@@ -52,22 +57,33 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
+      - uses: dorny/paths-filter@v2
```
anselmbradford, line 60: `dorny/paths-filter@2.10.2`

PR #5890, opened 2020-07-28
Link Spanish pages to Spanish social media

chosak:
When used in its non-sharing configuration, our social media molecule links to the Bureau's English Facebook (cfpb) and Twitter (cfpb) accounts. On Spanish pages, these links should go to the appropriate Spanish Facebook (CFPBespanol) and Twitter (CFPBespanol) accounts.

We always import this social-media template `with context` so it should have the proper `language` variable set.

## How to test this PR

To test, run a local server and visit these pages, and observe the links in the footer. These should be English:

- http://localhost:8000/
- http://localhost:8000/about-us/blog/receive-your-unemployment-benefits-options/

And these will be Spanish:

- http://localhost:8000/es/
- http://localhost:8000/about-us/blog/opciones-recibir-beneficios-desempleo/

## Notes and todos

There's some additional potential work to be done here when the social media molecule is used for sharing, for example around the default text, but that is not covered here.

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/molecules/social-media.html
+++ b/cfgov/jinja2/v1/_includes/molecules/social-media.html
@@ -63,6 +63,7 @@
 {% set linkedin_title = ( value.linkedin_title or blurb ) | urlencode %}
 {% set linkedin_text = ( value.linkedin_text or '' ) | urlencode %}
 {% set twitter_text = ( value.twitter_text or blurb ) | urlencode %}
+{% set language = language | default( 'en' ) %}
 
 <div class="m-social-media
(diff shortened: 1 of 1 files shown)


chosak:
> I wonder if there might be a better way to do this that doesn't rely on this extra template logic – something using the translation framework, perhaps?

Yes, I agree. There's some other complexity here too like [this code](https://github.com/cfpb/cfgov-refresh/blob/83df08aff98a5b04821c74565a22cf72835771bf/cfgov/jinja2/v1/_includes/molecules/social-media.html#L70-L83) that puts both languages in the template but then we use CSS to hide one based on the page `lang`. Here I was going for the minimal fix -- thanks for the review.

PR #5893, opened 2020-07-29
Reports as Webpages: Connect Jenkins API

schbetsy:
🚧  This is a Reports as Webpages PR, not going into cfgov-refresh yet 🚧 

---

Two changes here:

1. On Build and Production, use the Jenkins API to trigger the report processing job remotely
    - I've opened a PR to our ansible repo to supply values for the given environment variables in the Build and Production envs.
    - Unfortunately there's not a good way to locally test this change. I tested by using my personal Jenkins credentials and setting `RESEARCH_REPORT_JENKINS_JOB` to a test job I put up on our test jenkins server. 
    - Once this branch is merged in to the main branch and deployed to Build, it'll be testable there.

2. When saving the draft of a report, associate a user with the revision, to make revision history more understandable. Previously the associated user was `None`.
    - You can test this change by running the report processing locally and seeing that the revision created by the process is called `bot.user`.

### DIFF
--- a/cfgov/research_reports/management/commands/parse_research_report.py
+++ b/cfgov/research_reports/management/commands/parse_research_report.py
@@ -1,5 +1,6 @@
 import re
 
+from django.contrib.auth.models import User
 from django.core.management.base import BaseCommand
 
 import pypandoc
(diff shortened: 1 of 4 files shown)


cfgov/research_reports/management/commands/parse_research_report.py
```diff
@@ -24,6 +25,17 @@ def document_as_beautifulsoup(report_page):
     return BeautifulSoup(output, 'html.parser')
 
 
+def save_page(report_page):
+    print(" *** saving the page... ***")
+    report_page.process_report = False
+    try:
+        bot_user = User.objects.get(username='bot.user')
```
chosak, line 32: This doesn't really affect us because we don't customize the Django user model, but I've read that best practice is to do something like [`User = get_user_model()`](https://docs.djangoproject.com/en/2.2/topics/auth/customizing/#referencing-the-user-model) instead of directly importing `User`. We do this in a bunch of places, [for example](https://github.com/cfpb/cfgov-refresh/blob/4d41e1ff660fee379b99d2a1099a5f7732e7b0fd/cfgov/core/tests/test_management_inactive_users.py#L25).

schbetsy: Thanks for clearing that up! I added a commit to make this change.

PR #5898, opened 2020-07-30
Fix more issues with Jenkins multibranch build notifications

thephillipsequation:
## Changes

- Prevent more instances of `null` in the output
- Further standardization and consistency

## Checklist


Scotchester:
> My only recommended change is not to hard code the name `master` for the GIT_BRANCH.
> As another change would have to be made to the Jenkinsfile when the `master` branch is renamed.

Good thought, Craig, but I'm not sure if that can be avoided. I don't know that we have a way to test for whether or not the branch being built is generically the primary branch.

cwdavies:
> Couple questions:
> 
> 1. Does the `if/else` need to be wrapped in a `script` wrapper, as you did for the success path?
> 2. I'm noticing now that we inconsistently say "PR" or "Branch". The notification you replaced in the previous PR said "PR", but you replaced it with "Branch" in both cases. In _this_ PR, you're using "PR" if it's not `master`, but "Branch" if it is. I'm honestly not sure which is correct – whether these notifications apply to automatic PR builds, or the automatic master branch build, or both.
> 
> One other minor comment: In both the success and failure cases, if we are in the `else` path, we know that the branch name is master, so we could simplify the output to not ask for `$env.GIT_BRANCH` again.

That's a great suggestion Scott as a solution. Has this simplification been tried and does it work?

cwdavies (dismissed):
I recommend using something like this to support the branch rename to main:
`if (env.GIT_BRANCH != 'master' or env.GIT_BRANCH != 'main') {`

Jenkinsfile
```diff
@@ -134,16 +134,23 @@ pipeline {
     post {
         success {
             script {
-                if (env.GIT_BRANCH != 'master') {
-                    notify("${NOTIFICATION_CHANNEL}", ":white_check_mark: Branch $env.GIT_BRANCH PR $env.CHANGE_URL deployed by $env.CHANGE_AUTHOR_EMAIL via $env.BUILD_URL and available at https://$env.CFGOV_HOSTNAME.")
+                if (env.GIT_BRANCH != 'master' or env.GIT_BRANCH != 'main') {
+                    notify("${NOTIFICATION_CHANNEL}", ":white_check_mark: PR $env.CHANGE_URL deployed by $env.CHANGE_AUTHOR_EMAIL via $env.BUILD_URL and available at https://$env.CFGOV_HOSTNAME.")
```
Scotchester, line 138: I think we want this to be just `CHANGE_AUTHOR`. It's currently just spitting out `null`s.

```suggestion
                    notify("${NOTIFICATION_CHANNEL}", ":white_check_mark: PR $env.CHANGE_URL deployed by $env.CHANGE_AUTHOR via $env.BUILD_URL and available at https://$env.CFGOV_HOSTNAME.")
```

Scotchester (dismissed):

Jenkinsfile
thephillipsequation: I've made this fix but I think we may wind upw ith this spitting out null as well, if so I'll just simplify the notification

alexm118 (approved):

PR #5900, opened 2020-07-31
Add section level three to reports-as-webpages

wpears:
Closes GHE/CFPB/super-regular/issues/110

Into the reports-as-webpages branch

Adds a class for the level three section header of reports-as-webpages. Pretty unfancy/bloated code (more or less just copies&nests the subsection stuff). Could possibly set the header&body fields in separately from each class definition, but this works for now.

Also, renames the `sub_header` and `sub_body` fields on the subsection to just `header` and `body`, which is also how they're defined in the section level three.


cfgov/research_reports/jinja2/index.html
```diff
@@ -48,6 +48,19 @@ <h3 class="report-header">
       </h3>
 
       {{ subsection.body | safe}}
+
+      {% for section3 in subsection.children -%}
+        <h4 class="report-header">
+          <a
+            style="color:black;"
```
anselmbradford, line 55: Do you want an inline style here?

wpears: Fixed in 8c9ccc8f

cfgov/research_reports/models.py
```diff
@@ -40,10 +40,19 @@ def sec(request, page):
...
+                'title': subsection.header,
+                'body': subsection.body,
                 'url': format('#section-{}.{}', i, j),
-                'numbering': '' if is_appendix else format('{}.{} ', i, j)
+                'numbering': '' if is_appendix else format('{}.{} ', i, j),
+                'children': [{
+                    'title': section3.header,
+                    'body': section3.body,
+                    'url': format('#section-{}.{}.{}', i, j, k),
+                    'numbering': '' if is_appendix else
```
schbetsy, line 51: Are Appendix headers just not going to be numbered, or will we be adding that later?

cfgov/research_reports/models.py
```diff
@@ -40,10 +40,19 @@ def sec(request, page):
...
+                                 format('{}.{}.{} ', i, j, k),
+                } for k, section3 in enumerate(
+                    subsection.report_sections_level_three.all().order_by('pk')
```
schbetsy, line 54: The whole `get_report_parts` method is getting a little gnarly. I'll think about how to make it cleaner.

cfgov/research_reports/models.py
```diff
@@ -82,14 +91,30 @@ class ReportSection(ClusterableModel):
                          related_name='report_sections')
 
 
-class ReportSubSection(models.Model):
-    sub_header = models.CharField(max_length=200)
-    sub_body = models.TextField(blank=True)
+class ReportSubSection(ClusterableModel):
+    header = models.CharField(max_length=200)
+    body = models.TextField(blank=True)
```
schbetsy, line 96: 👍

cfgov/research_reports/models.py
wpears: Right, they won't be numbered (for subheaders and header-level-3)

cfgov/research_reports/models.py
```diff
@@ -82,14 +91,30 @@ class ReportSection(ClusterableModel):
...
+    ]
     section = ParentalKey('ReportSection',
                           on_delete=models.CASCADE,
                           related_name='report_subsections')
 
 
+class ReportSectionLevelThree(models.Model):
+    header = models.CharField(max_length=200)
+    body = models.TextField(blank=True)
+    section = ParentalKey('ReportSubSection',
```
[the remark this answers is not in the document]
wpears: Done in 404a6d70e2

PR #5905, opened 2020-08-01
Style current selection as black & fix ToC alignment

wpears:
PR into reports-as-webpages branch

Closes /CFPB/super-regular/issues/128
Closes GHE/CFPB/super-regular/issues/129

Just minor styling updates as described in the linked issues

### DIFF
--- a/cfgov/unprocessed/apps/research-reports/css/main.less
+++ b/cfgov/unprocessed/apps/research-reports/css/main.less
@@ -1,3 +1,5 @@
+@import (reference) '../../../css/main.less';
+
 .research-report {
   max-width: unit( (670 + 30) / 16, em);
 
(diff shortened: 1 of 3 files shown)


cfgov/unprocessed/apps/research-reports/css/main.less
```diff
@@ -22,12 +22,12 @@
   height: 100%;
   & > ul {
     a {
-      padding-left: 0.75em;
+      padding-left: 1.85em;
```
anselmbradford, line 25: These should probably use the `unit(…)` helper so it's easier to see the desired pixel output at the base font size. E.g. see https://github.com/cfpb/design-system/blob/38096726d51473b98424db3cac76325b0fc462a9/packages/cfpb-core/src/base.less#L346

cfgov/unprocessed/apps/research-reports/css/main.less
```diff
@@ -22,12 +22,12 @@
...
+      text-indent: -1.1em;
     }
   }
-  .m-nav-link.current-section {
-    color: #101820;
-    border-color: #20aa3f;
+  h3, h4 {
+    padding-left: 19px;
```
anselmbradford, line 30: Should this be ems?

wpears: They need to be the same so that everything is left aligned. I used the unit helper for this as well though

schbetsy:
Code looks good to me. 

I don't have an eye for these things, but how do you feel about the wrapping for the Appendix title here?
Screen Shot 2020-08-04 at 4 53 59 PM

wpears:
This is a good catch, yeah let me revisit the code, both for long subsection headers and for that appendix indent

wpears:
@schbetsy fixed in 9e5b420437ab

PR #5908, opened 2020-08-03
Reports as webpages: initial test

schbetsy:
This is an initial feature set for testing the Reports as Webpages process in our Build environment. Collaboration with @wpears.

## Additions

- New django app called `research_reports` for the reports models and code
- New page type, `ResearchReportPage`, with associated child models, e.g. `ReportSection`, `ReportAuthor`
- Jinja templates for displaying `ResearchReportPage`s
- Front-end code for creating and displaying a table of contents in the left-hand sidebar
- New management command for fetching a Word report, converting its contents to HTML, and putting its report data into a `ResearchReportPage` instance
- New management command for calling the Jenkins API to kick off processing a report

## Todos

Reports as Webpages code is not a finished feature. Here's an incomplete list of additional things we'll be adding in future PRs:

- Code to convert all remaining Report content from Pandoc-produced HTML to `ResearchReportPage` data fields, including:
    - Content sections, subsections, and third-level sections
    - Appendices
    - Figures
    - Data tables
- Models and front-end handling of a third level of report content
- Python tests
- Error handling, including scenarios where: 
    - Jenkins is down (when on a server where connection to Jenkins is enabled)
    - Pandoc isn't installed (when running in a location where local processing is enabled)
    - The provided Word report isn't parseable as a report
- Content protections to ensure original Word reports aren't discoverable by the public

## How to test this PR

1. Install Pandoc in your local environment: `brew install pandoc`
1. Get a sample report: [GHE]/CFPB/super-regular/wiki/Reports-as-Webpages:-sample-reports-for-testing
2. Get a recent database dump. Run migrations on this branch. `./frontend.sh`.
3. Run the branch, log in to the Wagtail Admin
4. Create a new Research Report Page
5. On the "Report Upload" tab, enter a Title manually. Using the "Report file" file chooser, select your sample report. Check the "Process report" checkbox.
6. Click the "Save draft" button. Within 10 seconds, the page banner should indicate that it was updated.
7. Visit the "Report Content" tab, see that the following content from the sample report was inserted into the page fields:
    - Report type, title, subtitle, authors, and footnotes.
8. Press the "Preview" button, see the rendered version of the report, including interactive sidebar table of contents

## Screenshots
Screen Shot 2020-08-03 at 1 20 19 PM
Screen Shot 2020-08-03 at 1 20 58 PM

## Checklist


cfgov/research_reports/models.py
```diff
@@ -0,0 +1,235 @@
...
+
+get_report_sections = get_report_parts()
+get_report_appendices = get_report_parts(True)
+
+
+def get_researchers():
+    return dict([
+        (r.title, r.url) for r in
+        SublandingFilterablePage.objects.get(pk=4833).get_children()
+    ])
```
chosak, line 74: Can we find some other way to do this that doesn't rely on a hardcoded page ID that only exists in production dumps? For example, what about doing a bulk lookup of DocumentDetailPages by researcher name? Is it possible for pages to have these names but exist elsewhere in the tree? It'd be nice if we had a ResearcherPage to help limit this query.

cfgov/research_reports/models.py
```diff
@@ -0,0 +1,235 @@
...
+from modelcluster.models import ClusterableModel
+
+from v1.models import SublandingFilterablePage
+from v1.models.base import CFGOVPage
+
+
+alpha_map = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+
+
+def get_report_parts(is_appendix=False):
```
chosak, line 26: I know you mention tests in your TODOs -- those would really help understand how this code works.

cfgov/research_reports/models.py
```diff
@@ -0,0 +1,235 @@
...
+
+def _get_deploy_environment():
+    return getattr(settings, 'DEPLOY_ENVIRONMENT', 'local')
+
+
+class ReportSection(ClusterableModel):
+    header = models.CharField(max_length=200, blank=True)
+    is_appendix = models.BooleanField(default=False)
+    body = models.TextField(blank=True)
+    panels = [
```
chosak, line 85: Minor style thing, but Wagtail convention [seems to be](https://docs.wagtail.io/en/v2.9.3/topics/pages.html#an-example-wagtail-page-model) to put panel definitions below all of the model fields. I find that helpful to separate out database fields from other attributes on the class.

cfgov/research_reports/models.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+class ReportMetaclass(WagtailAdminModelFormMetaclass):
+    @classmethod
+    def child_form(cls):
+        return ReportForm
+
+
+class ReportForm(WagtailAdminPageForm, metaclass=ReportMetaclass):
+    pass
```
chosak, line 134: Because this form class hasn't been customized at all, are these needed?

cfgov/research_reports/models.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+class ResearchReportPage(CFGOVPage):
+    # Report Upload Fields
+    report_file = models.ForeignKey(
+        'wagtaildocs.Document',
```
chosak, line 140: If the original reports aren't intended to be downloadable / linked elsewhere on the website, is there an advantage to using a Wagtail Document for this as opposed to a plain Django [`FileField`](https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.FileField)? The latter would still allow for uploading to S3, but would allow for uploading to a different location, and wouldn't stick the uploaded reports among the other Wagtail document views.

cfgov/research_reports/models.py
```diff
@@ -0,0 +1,235 @@
...
+        FieldPanel('footnotes')
+    ]
+
+    sidefoot_panels = CFGOVPage.sidefoot_panels
+
+    # Tab handler interface
+    edit_handler = TabbedInterface([
+        ObjectList(upload_panels, heading='Report Upload'),
+        ObjectList(content_panels, heading='Report Content'),
+        ObjectList(sidefoot_panels, heading='Sidebar'),
```
chosak, line 207:
```suggestion
        ObjectList(CFGOVPage.sidefoot_panels, heading='Sidebar'),
```

Doing this allows for deletion of the `sidefoot_panels` definition above.

cfgov/research_reports/models.py
```diff
@@ -0,0 +1,235 @@
...
+                # TODO: show a friendly error message to the user
+                pass
+
+    def get_context(self, request, *args, **kwargs):
+        context = super(ResearchReportPage, self) \
+            .get_context(request, *args, **kwargs)
+        context.update({
+            'get_report_sections': get_report_sections,
+            'get_report_appendices': get_report_appendices,
+            'get_researchers': get_researchers
```
chosak, line 233: These methods get called in the template to generate variables; would it be simpler to just call those here and pass the variables in directly?

cfgov/research_reports/models.py
```diff
@@ -0,0 +1,235 @@
...
+            elif deploy_env == 'local':
+                management.call_command('parse_research_report', self.id)
+            else:
+                # deployed to another environment where s3 isn't configured
+                # TODO: show a friendly error message to the user
+                pass
+
+    def get_context(self, request, *args, **kwargs):
+        context = super(ResearchReportPage, self) \
+            .get_context(request, *args, **kwargs)
```
chosak, line 229:
```suggestion
        context = super().get_context(request, *args, **kwargs)
```

Python 3!

cfgov/research_reports/jinja2/index.html
```diff
@@ -0,0 +1,114 @@
...
+{% set researchers = get_researchers() %}
+
+
+{% macro report_part(sections, is_appendix) %}
+  {% for section in sections -%}
+    <h2 class="report-header">
+      <a
+        id="{{section.url[1:]}}"
+        href="{{section.url}}">
+          {{('Appendix ' if is_appendix else '') + section.numbering + section.title}}
```
chosak, line 33: Nitpicky thing, but @anselmbradford has asked me many times to use more spacing in Jinja templates. This is just one example from this PR, and feel free to ignore, especially since I can't find any documentation on our Jinja standards.

```suggestion
          {{ ( 'Appendix ' if is_appendix else '' ) + section.numbering + section.title }}
```

cfgov/research_reports/jinja2/index.html
```diff
@@ -0,0 +1,114 @@
...
+{% endif %}
+
+{% set researchers = get_researchers() %}
+
+
+{% macro report_part(sections, is_appendix) %}
+  {% for section in sections -%}
+    <h2 class="report-header">
+      <a
+        id="{{section.url[1:]}}"
```
chosak, line 31: What's the motivation for the `[1:]` here?

cfgov/research_reports/management/commands/parse_research_report.py
```diff
@@ -0,0 +1,90 @@
...
+
+# To invoke this command from the cfgov-refresh root:
+#     cfgov/manage.py parse_research_report [report_page_id]
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument('report_page_id', type=int)
+
+    def handle(self, *args, **options):
+        report_id = options['report_page_id']
+        print('received id as argument: ', id)
```
chosak, line 86:
```suggestion
        print('received id as argument: ', report_id)
```

requirements/libraries.txt
```diff
@@ -35,6 +35,7 @@ openpyxl==2.5.8
 psycopg2==2.7.3.2
 pyelasticsearch==0.6.1
 python-dateutil==2.7.3
+pypandoc==1.5
```
chosak, line 38:
```suggestion
pypandoc==1.5
python-dateutil==2.7.3
```

cfgov/research_reports/jinja2/index.html
```diff
@@ -0,0 +1,114 @@
...
+        href="{{section.url}}">
+          {{('Appendix ' if is_appendix else '') + section.numbering + section.title}}
+      </a>
+    </h2>
+
+    {{ section.body | safe }}
```
chosak, line 37: Just to confirm: these need `| safe`  because they contain raw HTML that needs to go verbatim into the page?

schbetsy: Yes. Without `| safe`, the html content would show up as html instead of as content.

requirements/libraries.txt
schbetsy: Fixed in fb68748672adc3af6a50c70501c76809190e9982.

cfgov/research_reports/management/commands/parse_research_report.py
schbetsy: Fixed in fb68748672adc3af6a50c70501c76809190e9982.

cfgov/research_reports/management/commands/parse_research_report.py
```diff
@@ -0,0 +1,90 @@
+import re
+
+from django.contrib.auth import get_user_model
+from django.core.management.base import BaseCommand
+
+import pypandoc
+from bs4 import BeautifulSoup
+from research_reports.models import ReportAuthor, ResearchReportPage
```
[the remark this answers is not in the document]
schbetsy: Fixed in fb68748672adc3af6a50c70501c76809190e9982.

cfgov/research_reports/models.py
schbetsy: I meant to ask about this. @wpears Do we even want to allow sidefoot content? Seems like that's not part of the design, so could we leave out that tab entirely?

schbetsy:
|  Is this expected given the in-progress state of this code?

Yes. Parsing of report content is listed in the To Do section of this PR. The purpose of this merge is to get the basic workflow (uploading reports, calling the Jenkins API, running the process on Jenkins, seeing updates in the Wagtail admin) working on Build.

cfgov/research_reports/models.py
schbetsy: I see. I had grouped things by which tab they appeared on in the Wagtail admin, since that organization was helpful for me, but I can move them to match the standard.

cfgov/research_reports/models.py
schbetsy: This was Wyatt's response when I asked him about that: https://github.com/cfpb/cfgov-refresh/pull/5767/files#r436171282

cfgov/research_reports/models.py
wpears: Yeah I think we can nix the sidefoot

cfgov/research_reports/jinja2/index.html
wpears: The `section.url` variable always starts with a `#` which we don't want to pass through to the id. Perhaps this would be more cogent if the variable name was `fragment`?

cfgov/research_reports/models.py
wpears: Yeah this logic could likely be modified to not target the researchers parent page directly, as you intimated above, this was mostly done to limit the query.

cfgov/research_reports/management/commands/parse_research_report.py
```diff
@@ -0,0 +1,90 @@
...
+
+report_type_target = 'Report [Tt]ype: '
+title_target = 'Title: '
+subtitle_target = 'Subtitle: '
+authors_target = 'Authors: '
+blank = ''
+
+
+def document_as_beautifulsoup(report_page):
+    print(" *** downloading the report file... ***")
```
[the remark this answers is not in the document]
schbetsy: This file will end up changing a lot before Reports as Webpages is done. The print statements are still helpful for debugging for now, but they won't make it into the end product ☀️

cfgov/research_reports/jinja2/index.html
chosak: Yes, it might be clearer if the `url` variable had a different name (e.g. `section_id`) and was defined without the hash, and then the hash gets appended where it needs to be?

cfgov/research_reports/jinja2/index.html
chosak: It would be nice if there was some kind of validation against improper HTML, for example stripping out `<script>` tags or the like. I realize that might add some extra complexity, though, depending on how careful you think we need to be about the input content.

cfgov/research_reports/models.py
chosak: Oooh. Past me was a little too optimistic. It turns out that, even with that workaround, nested inline panel support is still incomplete and doesn't always work as expected. For example, you can manually create nested inline items, but subsequently reordering them won't work. There are [a few open Wagtail issues](https://github.com/wagtail/wagtail/search?q=nested+inlinepanel&type=Issues) about this. For your purposes this might not matter if you are creating the items programmatically once and then never altering them manually.

cfgov/research_reports/models.py
```diff
@@ -0,0 +1,235 @@
...
+        default=False,
+        help_text=mark_safe(
+            'If this is checked, when you press "save", the system will '
+            'read in the report document and use its contents to overwrite '
+            'the fields in the "Report Content" tab.'
+            '<ul class="help">'
+            '    <li>&bull; If you uploaded a new report file for '
+            'processing, check this box.</li>'
+            '    <li>&bull; If you manually edited fields in the '
+            '"Report Content" tab, do not check this box</li>'
```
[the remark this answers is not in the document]
schbetsy: Fixed in 109a63e.

cfgov/research_reports/models.py
schbetsy: Fixed in 109a63e.

cfgov/research_reports/models.py
schbetsy: Sweet. Done in b89e19f5ce6cc12.

cfgov/research_reports/models.py
```diff
@@ -0,0 +1,235 @@
...
+        'wagtaildocs.Document',
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name='+',
+    )
+    process_report = models.BooleanField(
+        default=False,
+        help_text=mark_safe(
+            'If this is checked, when you press "save", the system will '
```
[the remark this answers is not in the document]
schbetsy: Good idea. I'll make that change. Also, after this goes to Build, we'll be getting feedback from users and designers, so we'll have more data to further hone the messaging.

cfgov/research_reports/jinja2/index.html
schbetsy: Interesting. What case are you hoping to protect against? I think the only ways that HTML could get into these models are (a) from our `parse_research_report` command, which is based on Pandoc, which wouldn't include a `<script>` tag, (b) entered manually by a content editor (who wouldn't have a reason to include a `<script>` tag), or (c) entered maliciously by someone unauthorized.  I felt okay using `| safe` because (a) is the intended use case and should be the overwhelming majority of where content comes from. 

Are you picturing a custom method to scan HTML and strip out undesired tags? And when are you picturing that happening--at the parsing step or on page load?

chosak: Those are great questions. My comment came from a general feeling that putting fields behind `| safe` is generally something we should only do if we are really sure that it's safe to do so. In terms of case (b), there's also the possibility of something inadvertently getting included, either by Pandoc or by a content editor making a mistake. Maybe the risk is low enough -- after all, we already allow [RawHTMLBlocks](https://docs.wagtail.io/en/v2.9.3/topics/streamfield.html#rawhtmlblock) on pages, which does the same thing.

If we did want an extra layer of certainty about Pandoc output, either now or in future, we could do some post-processing of the soup, for example:

```py
# Given the soup generated when BS runs on the Pandoc output
>>> set(tag.name for tag in soup.find_all())
{'td', 'table', 'sup', 'u', 'thead', 'p', 'hr', 'h5', 'h2', 'ol', 'em', 'br', 'strong', 'tbody', 'th', 'section', 'h1', 'img', 'li', 'a', 'tr'}
# Check to make sure 'script' isn't in that list...
```

But as you mention, this wouldn't protect against tags entered in the editor unless this also happened on page save validation.

cfgov/research_reports/models.py
```diff
@@ -0,0 +1,235 @@
...
+
+def get_researchers():
+    return dict([
+        (r.title, r.url) for r in
+        SublandingFilterablePage.objects.get(pk=4833).get_children()
+    ])
+
+
+def _get_deploy_environment():
+    return getattr(settings, 'DEPLOY_ENVIRONMENT', 'local')
```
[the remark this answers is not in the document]
schbetsy: Done in 12d5d71bcf963736d019e34b5cb3a6a57e357be1.

cfgov/research_reports/models.py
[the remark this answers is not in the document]
schbetsy: Ignore me. I'll push the migration for this one soon. 🤦

cfgov/research_reports/jinja2/index.html
schbetsy: Thanks for that idea. I think that can be a future enhancement, rather than going into this PR, so I'll add it to our backlog.

cfgov/research_reports/models.py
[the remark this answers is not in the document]
schbetsy: Done in 43be6a5b30778e5f6b7a38ba47969bebdcf47d5d.

cfgov/research_reports/models.py
schbetsy: Yeah, we've discussed this several times before, and I hadn't gotten to that item of my backlog. I had thought I wouldn't tackle that issue yet, but now I realize that it might affect UX, so I probably should. Looking into it now. 👀

cfgov/research_reports/models.py
schbetsy: @wpears Is this code still doing what you intended? I couldn't discern a difference in how the nested panels work when it's removed. And we're on Wagtail 2.9 now, so that comment that references Wagtail 2.8+ should no longer apply.

wpears: Yeah, that's probably vestigial at this point, if you've found that things still work as expected with it removed, let's remove it.

cfgov/research_reports/models.py
wpears: Fixed in df7ee34e64732c7e48a8a3bbb838e6cd5d4372d9

wpears:
Closing for PR hygiene. Let's of course keep the branch though for when this project gets restarted.

PR #5916, opened 2020-08-05
Add wagtailmedia app and make AudioPlayer block available on LearnPage

Scotchester:
We will soon have a need to upload audio files in Wagtail and add players for them on-page. This PR makes that possible by introducing the wagtailmedia package and an `AudioPlayer` block (currently restricted to `LearnPage`s).

## Additions

- wagtailmedia package
- `AudioPlayer` block in `LearnPage`'s `content` StreamField

## How to test this PR

1. Pull branch
1. Migrate
1. Run server
1. Go to the Media library and upload an MP3 file
1. Create or edit a `LearnPage` and add an audio block, choosing the MP3 you uploaded

## Checklist

#### Browser testing

- [x] Firefox
- [x] Chrome
- [x] Safari
- [x] Safari on iOS
- [x] Chrome on Android

#### Accessibility

- [x] Screen reader friendly

#### Other

- [x] Is useable without CSS
- [x] Is useable without JS


cfgov/v1/models/learn_page.py
```diff
@@ -117,6 +117,7 @@ class LearnPage(AbstractFilterPage):
         ('call_to_action', molecules.CallToAction()),
         ('email_signup', organisms.EmailSignUp()),
         ('video_player', organisms.VideoPlayer()),
+        ('audio_player', v1_blocks.AudioPlayer()),
```
niqjohnson, line 120: Can we also make the `AudioPlayer` block an option inside of full-width text blocks on learn pages, too (without the `.block` container)? Or maybe just inside of the full-width text block and not as a top level block? The latest mockups for the podcast page I've seen have the audio player surrounded by a heading and links that don't need to be separated from the audio player by our 60px block margins.

cfgov/v1/blocks.py
```diff
@@ -163,6 +166,29 @@ class Meta:
...
+            '''
+        else:
+            return ''
+
+        return format_html(player_code, format_html_join(
+            '\n', "<source{0}>",
+            [[flatatt(s)] for s in value.sources]
+        ))
+
+
```
niqjohnson, line 191: Should we give this an icon? I wish there were an option other than the same icon we use for the video player block, but I couldn't find anything more suitable in the Wagtail UI icon set.

```suggestion
class Meta:
        icon = 'media'
```

Scotchester: Whoops, yes! An oversight that I noticed but forgot to correct. Thanks!

cfgov/v1/models/learn_page.py
Scotchester: I'm working on implementing an alternate approach that @niqjohnson and I discussed yesterday: including heading, body, and additional details fields in the audio player block.

niqjohnson:
@Scotchester: Sorry, one more thing I just noticed. Is there any way to suppress the listing of media files in the Wagtail dashboard?

media-listing

I can't see it being that useful *and* it pushes the list of recent page edits further down. If not, no biggie.

Scotchester:
Hokay, this PR has been refactored pretty significantly to now provide an `AudioPlayer` organism block, which includes optional heading and body text fields to appear before the audio player widget itself, and an additional details field to appear after it. Everything's rebased up, migration is recreated, tests added, and ready for a hopefully-final review!

Scotchester:
I've encountered an issue with the audio player in Safari that I don't think is a blocker for this PR, but I could use some BEWD help to sort out soon. Safari interprets the audio as a "live broacast", resulting in this sort of player without a scrubbable timeline: 

HTML audio player in Safari with Live Broadcast indicator

This old but still-regularly-added-to SO question describes a number of possible solutions, all relating to headers that are a bit over my head: https://stackoverflow.com/questions/1995589/html5-audio-safari-live-broadcast-vs-not

Scotchester:
> * I can add audio blocks to a LearnPage and have them render, but I can't get an mp3 or m4a to play on the page.
>   They do play for me in the media admin.

Concerning! It works for me locally, in every browser, and on the Docker stack. Perhaps something specific to your environment?

> * I wonder why the duration field is required? Whatever I enter, it is ignored, and when I play the audio file in the media admin the player reports the true duration whatever I enter. On the LearnPage, I'm not seeing any duration value.

Annoying! Yes, this is a quirk of wagtailmedia that we unfortunately can't control. I don't really see the point of requiring it, either, though it will come in handy for future automatic podcast feed creation. I would like to maybe raise an issue/PR in wagtailmedia about it later.

> * If thumbnails are for video only, it would be nice to note that, say in the audio_file block help text.

Agreed! I can add a note about that.

> * A bunch of django.po changes seem to have hitched a ride on this PR

Intentional! Want to translate the unsupported-browser string. Adding it also brings along a lot of comment updates in that file (line number indications for where the translated string is found) that should probably have been done long ago. But I forgot to do the final part of the process that creates the `.mo` file, too, so I need to add that here.

higs4281:
> Perhaps something specific to your environment?

It appears so. 

My browsers were blocking the S3 stream, citing CSP. Adding a setting fixed it locally, but not sure this is necessary. 

```python
CSP_MEDIA_SRC = (
    "s3.amazonaws.com",
)
```

👍 on other issues.

cfgov/unprocessed/js/organisms/AudioPlayer.js
```diff
@@ -0,0 +1,87 @@
...
+ * @classdesc Initializes a new AudioPlayer organism for the purpose of sending
+ * analytics events for the native HTML `audio` element.
+ *
+ * @param {HTMLNode} element
+ *   The DOM element within which to search for the organism.
+ * @returns {AudioPlayer} An instance.
+ */
+function AudioPlayer( element ) { // eslint-disable-line max-lines-per-function
+  const _dom = checkDom( element, BASE_CLASS );
+  const label = _dom.dataset.title;
```
anselmbradford, line 22: Move this inside `init`

cfgov/unprocessed/js/organisms/AudioPlayer.js
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * @returns {Carousel} An instance.
```
anselmbradford, line 25: `{Carousel}` --> `{AudioPlayer}`

Scotchester:
Closing and reopening due to Snyk hang.

PR #5921, opened 2020-08-06
Update openpyxl from 2.5.8 to 3.0.3

cwdavies:
In order to prepare to support Wagtail 2.10 openpyxl version needs to be newer than 2.5.8

Refer to https://openpyxl.readthedocs.io/en/stable/changes.html#id9 for the bug fixes

### DIFF
--- a/requirements/libraries.txt
+++ b/requirements/libraries.txt
@@ -28,7 +28,7 @@ jsonfield==2.0.2
 lxml==4.2.5
 Markdown==3.2.1
 ntplib==0.3.4
-openpyxl==2.6.0
+openpyxl==3.0.3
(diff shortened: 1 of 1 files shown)


requirements/libraries.txt
```diff
@@ -28,7 +28,7 @@ jsonfield==2.0.2
 lxml==4.2.5
 Markdown==3.2.1
 ntplib==0.3.4
-openpyxl==2.5.8
+openpyxl==2.6.0
```
Scotchester, line 31:
```suggestion
openpyxl==3.0.3
```

cwdavies: Agreed and updated to the suggested change.

PR #5923, opened 2020-08-07
Remove Document Serve View

alexm118:
Removes the DocumentServeView in preference of [WAGTAILDOCS_SERVE_METHOD](https://docs.wagtail.io/en/v2.9.2/reference/settings.html#documentsl)

---

## Removals

- DocumentServeView 

## How to test this PR

1.

## Notes and todos

-

## Checklist


docker-stack.yml
```diff
@@ -80,10 +85,9 @@ services:
                 if [ $$django_tables_exist -gt 0 ] || [ $$REFRESH_DB == 'true' ]; then
                     ./refresh-data.sh 
                     ./cfgov/manage.py update_index
-                    httpd -d cfgov/apache -D FOREGROUND
-                else
-                    httpd -d cfgov/apache -D FOREGROUND
-                fi"
+                fi
+                
+                httpd -d cfgov/apache -D FOREGROUND"
```
alexm118, line 90: Saw some cleanup could be done to this, we don't really need an else block if its the same as the bottom of the if block.

Dockerfile
```diff
@@ -113,6 +113,8 @@ COPY --from=cfgov-build --chown=apache:apache ${CFGOV_PATH}/static.in ${CFGOV_PA
 RUN yum clean all && rm -rf /var/cache/yum && \
     chown -R apache:apache ${APP_HOME} ${SCL_HTTPD_ROOT}/usr/share/httpd ${SCL_HTTPD_ROOT}/var/run
 
+ENV PATH="/opt/rh/${SCL_PYTHON_VERSION}/root/usr/bin:${PATH}"
```
alexm118, line 116: This now lets the apache user run the django commands

wpears: I'm a little squeamish about this, mainly because I think the apache user should have a few privileges as possible to lower attack surface area. Could these commands be run as a different user as part of the docker build? (this would likely require reorganizing the dockerfile a bit)

alexm118: not in its current form, and the apache user has access to python 2.7 as is so I figured the overall surface area for this to be somewhat low

alexm118: The reason behind why I ended up here, was the django command we need to run to seed the index relies on python 3.6 and the dependencies we install as part of the `cfgov-dev` stages. These are all installed as root. We could look to refine how we currently populate our elasticsearch to not rely on the python container and docker networking to reach elasticsearch and instead move these out to either a separate jenkins job or a separate stage executed by another container. The limiting factor there is we would need to expose the ES container via DNS.

alexm118: Overall, I think expanding the PATH variable for the apache user isn't too risky, as anyone capable of getting into a running container would have the ability to expand it anyways. Now if we want to evaluate trimming down our docker container even more then I think that is a good discussion to have.

wpears: Yeah, I agree that this change itself is low risk and mirrors our conception of how the prod server should be set up and accessed. I suppose my point is that in a Docker regime, that conception is a bit flawed and we should try to move towards having running containers with minimal access rights/ability to run code other than what is their express purpose and to separate adjacent tasks into build stages/jobs as you suggest. Imo in docker, we should never expect to attach to the running prod container or use it to marshal tasks other than simply serving the website.

PR #5925, opened 2020-08-07
Pin unittest-future on Wagtail 2.10 release

cwdavies:
This task is to pin unittest-future tox environment on Wagtail 2.10 and ensure all tests pass.

There do not appear to be any upgrade considerations that we should be particularly concerned about with 2.10.

## Additions

- wagtail 2.10 in tox.ini

## Changes

- `INSTALLED_APP` has `wagtail.contrib.legacy.richtext` in `cfgov/cfgov/settings/base.py`
- Update djangorestframework to 3.11.1 in `requirements/libraries.txt'
- Used new releases of wagtail libraries 

## How to test this PR

1. tox

## Checklist

#### Browser testing

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Safari on iOS
- [ ] Chrome on Android

#### Accessibility

- [ ] Screen reader friendly

#### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


cfgov/ask_cfpb/tests/test_blocks.py
```diff
@@ -13,19 +15,28 @@ def setUp(self):
...
+                '</aside>'
+            )
+            self.expected_text_html = 'text'
         self.text_data = {
             'type': 'text',
             'value': {
                 'content': 'text'
             }
         }
-        self.expected_text_html = '<div class="rich-text">text</div>'
```
willbarton, line 28: This one in particular seems like it could affect rendering.

requirements/libraries.txt
```diff
@@ -39,12 +39,16 @@ sha3==0.2.1
...
+#wagtail-flags==4.2.1
+#wagtail-inventory==1.1
+#wagtail-sharing==2.2
+#wagtail-treemodeladmin==1.2.0
 
 # These packages are installed from GitHub.
+git+https://github.com/cfpb/wagtail-flags.git@wag210-test
+git+https://github.com/cfpb/wagtail-inventory.git@wag210-test
+git+https://github.com/cfpb/wagtail-sharing.git@wag210-test
+git+https://github.com/cfpb/wagtail-treemodeladmin.git@wag210-test
```
willbarton, line 51: As a methodology thing, if we put these in the tox.ini's `future-config` `deps` as well, we could merge this without having any impact on running code.

cwdavies: @willbarton That is a really great suggestion I had not thought of, so I have created a `wagtail-future.txt` requirements file to address the future-config wagtail dependencies, then updated `tox.ini` and `requirements/libraries.txt`.

cfgov/ask_cfpb/tests/test_blocks.py
cwdavies: @willbarton This change to ask_cfpb may impact the way block renders for users. This will require more investigation and browser testing as a result of the HTML `rich-text` changes that wagtail are making. I am very interested to see if the new cypress tests start to fail.

cwdavies:
> This looks great, and future tests pass.
> 
> One question: Does this mean we'll always be importing the CFPB Wagtail libraries via git+https?

@higs4281 No, I do not believe we will always import the CFPB Wagtail libraries using `git+https`.
There would be newer versions of Wagtail libraries used for `unittest-future` to those used by `unittest-current`

tox.ini
```diff
@@ -139,7 +139,7 @@ deps=
 basepython=python3.6
 deps=
     Django>=2.2,<3.0
-    wagtail>=2.3,<2.10
+    -r{toxinidir}/requirements/wagtail-future.txt
```
willbarton, line 142: Something like this is what I meant to suggest:

```suggestion
    wagtail==2.10rc2
    wagtail-autocomplete==0.6
    git+https://github.com/cfpb/wagtail-flags.git@wag210-test
    git+https://github.com/cfpb/wagtail-inventory.git@wag210-test
    git+https://github.com/cfpb/wagtail-sharing.git@wag210-test
    git+https://github.com/cfpb/wagtail-treemodeladmin.git@wag210-test
```

I'd really rather pin the temporary things we need for upgrading *here*, in the tox.ini, than in requirements files.

requirements/libraries.txt
willbarton: That's not quite what I meant. See below.

I'd really rather not expand our requirements files like this.

cfgov/ask_cfpb/tests/test_blocks.py
willbarton: The Cypress tests check functionality, not rendering. We don't have visual regression testing yet. Could you check?

You can also run Cypress locally if you want to see what it does.

tox.ini
willbarton: I can expand on this a little bit — I'm struggling with visualizing the "future" env a lot of the time — I think it might be worth talking about whether we go back to tox conventions for env naming for everything that's *not* `current` (`unittest-dj22-wag210`, for example). I don't want to make `-future` that little more concrete with a requirements file right now.

willbarton:
> One question: Does this mean we'll always be importing the CFPB Wagtail libraries via git+https?

No. They'll be released with 2.10 support once 2.10 is released.

tox.ini
cwdavies: Changes are made as requested to remove requirements file wagtail-future.txt
Added temporary wagtail repos for upgrading in tox.ini

cwdavies:
Adding Mattermost messages from @willbarton
I think I was thinking about wrong — deploying to UCP should not be in the same PR. So, if we can keep the requirements for 2.10 in tox.ini, that PR could be merged, RC or not. Testing all PRs against future changes is the point, so that’s why we want it merged.
And we could pin2.10 in another branch and deploy that branch to UCP, which would run the cypress tests. And we can run them locally too. And THAT branch is the upgrade branch
I just want to make sure we’re separating the concerns of “can we test against a future version” from “can we run a future version”.

cfgov/ask_cfpb/tests/test_blocks.py
cwdavies: What change to the blocks test are you requesting in this PR? How do you want me to check the rendering?

willbarton: I think the expected HTML should still be `<div class="rich-text">text</div>`. Instead of changing the test, did you try to figure out why the expected output changed?

cfgov/ask_cfpb/tests/test_blocks.py
```diff
@@ -72,25 +83,50 @@ def test_how_to_block_renders_schema(self):
...
-            '<li itemprop="step"'
-            '     itemscope'
-            '     itemtype="http://schema.org/HowToStep"'
-            '     class="schema-block_item">'
-            '<h3 itemprop="name" class="h4">Step one</h3>'
-            '<div itemprop="text">Step content</div>'
-            '</li>'
-            '</ol>'
-            '</div>'
-        )
```
willbarton, line 93: Ah I also see what's going on here — we're losing `<div class="rich-text">` around `test description` here too —I was thinking it's whitespace changes, but then [`self.assertHTMLEqual`](https://docs.djangoproject.com/en/2.2/topics/testing/tools/#django.test.SimpleTestCase.assertHTMLEqual) is supposed to ignore whitespace around HTML tags.

chosak: This is a [deliberate change in Wagtail 2.10](https://docs.wagtail.io/en/v2.10/releases/2.10.html#div-class-rich-text-wrappers-removed-from-rich-text). It's possible to at least partially restore the old behavior with the provided [legacy rich text tag](https://docs.wagtail.io/en/v2.10/reference/contrib/legacy_richtext.html#legacy-richtext), but I don't know if this will fix all occurrences (for example, if we manually import and run that method anywhere).

I do think it would be good if we didn't have this div wrapper in our output (see rationale and discussion at https://github.com/wagtail/wagtail/issues/1214), but removing it might have _visual_ implications based on CSS rules.

willbarton: Ok. So, I just think I'd like to confirm, @cwdavies, that there is no visual difference between Wagtail 2.9 and 2.10 that doesn't require any CSS changes. Can you confirm that?

cfgov/ask_cfpb/tests/test_blocks.py
cwdavies: I have not yet been able to figure out why the expected output changed in the expected HTML.

cfgov/ask_cfpb/tests/test_blocks.py
Scotchester: We have the following CSS and JavaScript that looks for that `.rich-text` class:

https://github.com/cfpb/cfgov-refresh/blob/18b75f5c37afe1495c02054c6b16a742a3a3c1ce/cfgov/unprocessed/css/organisms/info-unit-group.less#L49-L56

https://github.com/cfpb/cfgov-refresh/blob/18b75f5c37afe1495c02054c6b16a742a3a3c1ce/cfgov/unprocessed/css/pages/ask.less#L15-L26

https://github.com/cfpb/cfgov-refresh/blob/18b75f5c37afe1495c02054c6b16a742a3a3c1ce/cfgov/unprocessed/css/rich-text.less#L1-L22

https://github.com/cfpb/cfgov-refresh/blob/18b75f5c37afe1495c02054c6b16a742a3a3c1ce/cfgov/unprocessed/css/schema.less#L5-L7

https://github.com/cfpb/cfgov-refresh/blob/18b75f5c37afe1495c02054c6b16a742a3a3c1ce/cfgov/unprocessed/js/routes/on-demand/youth-employment-programs/buying-a-car/index.js#L30-L32

cwdavies: As rich-text is defined in css and javascript I am adding `wagtail.contrib.legacy.richtext` to `INSTALLED_APPS`

willbarton: @cwdavies that sounds fine. 

@Scotchester Should we add a task to remove the CSS dependency on the `rich-text` class?

Scotchester: Yes, I would very much like to rid ourselves of that wrapper and associated CSS/JS.

cwdavies: A separate task is required to update CSS and JavaScript to remove dependency on the `rich-text` class

cfgov/ask_cfpb/tests/test_blocks.py
cwdavies: I reverted the test back to using `rich-text` class

requirements/libraries.txt
```diff
@@ -38,11 +38,6 @@ requests_toolbelt==0.8.0
 sha3==0.2.1
 unipath>=1.1,<=2.0
 urllib3==1.25.2
-wagtail-autocomplete==0.6
-wagtail-flags==4.2.1
-wagtail-inventory==1.1
-wagtail-sharing==2.2
-wagtail-treemodeladmin==1.2.0
```
willbarton, line 45: @cwdavies Please don't move these — wagtail.txt should only contain wagtail, not Wagtail and all the Wagtail-related libraries.

cfgov/ask_cfpb/tests/test_blocks.py
```diff
@@ -19,13 +19,13 @@ def setUp(self):
             '<div class="rich-text">Tip content</div>'
             '</aside>'
         )
+        self.expected_text_html = '<div class="rich-text">text</div>'
```
willbarton, line 22: For a cleaner git history, could we move this back to line 28?

requirements/libraries.txt
cwdavies: Done

cfgov/ask_cfpb/tests/test_blocks.py
cwdavies: I will perform a squash and merge for a clean git history

requirements/libraries.txt
```diff
@@ -9,7 +9,7 @@ boto3==1.7.80
 # does until we can solve that problem.
 cryptography==2.9.2
 dj-database-url==0.5.0
-djangorestframework==3.9.1
+djangorestframework>=3.9.1,<4.0
```
willbarton, line 12: I'm sorry to keep doing this — but what's the reason for the version range here? Can we pin a definite version?

willbarton: Looks like Wagtail 2.10 requires DRF > 3.11.1, which looks like the latest release. Can we use that?

cfgov/cfgov/settings/base.py
```diff
@@ -37,8 +39,17 @@
...
-    "permissions_viewer",
+if wagtail.VERSION < (2, 10):
+    INSTALLED_APPS = (
+        "permissions_viewer",
+    )
+else:
+    INSTALLED_APPS = (
+        "permissions_viewer",
+        "wagtail.contrib.legacy.richtext",
+    )
```
willbarton, line 50: This could all be simplified:

```suggestion
INSTALLED_APPS = (
    "permissions_viewer",
)
if wagtail.VERSION >= (2, 10):
    INSTALLED_APPS += (
        "wagtail.contrib.legacy.richtext",
    )
```

requirements/libraries.txt
cwdavies: I will test with djangorestframework==3.11.1

cwdavies: I will test with djangorestframework==3.11.1

cfgov/cfgov/settings/base.py
cwdavies: I will use your recommended simplification

willbarton:
Thanks, @cwdavies!

PR #5926, opened 2020-08-10
Replace jsonfield's JSONField with djangos builtin postgres JSONField

alexm118:
Begins the process of removing the jsonfiled dependency. Creates a new migration that uses django's builting field as well as a squash migration. We need the squash migration so we can actively remove the jsonfield dependency.

---

## Additions

- Migrations around switching to Django's JSONField.

## Removals
 
- Usage of jsonfield library

## Changes

- Which JSONField we are using.

## How to test this PR

1. check the mortgage metadata constants such as FIPS
2. [PFC API](http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/api/school/131469/)

## Notes and todos

- This is the first half of the process, once this rolls into production we should remove the old migration files and stick with the sole squashed migration.

## Checklist


alexm118:
One of the results of moving our JSONField is that we can now pass actual JSON in via fixtures, sorry that it ends up bloating this PR considerably with those changes.

willbarton:
> we can now pass actual JSON in via fixtures

Makes it all worthwhile.

alexm118:
@willbarton Do we want to rename the migration if the intent is to remove the file entirely and use the squashed file when we remove the dependency in a follow up PR?

willbarton:
@alexm118 I think so, because it'll still be in the git history. Our convention is to make sure the migrations have a descriptive name, but there are also plenty of `auto` migrations in that history too.

I'll leave it to your discretion.

PR #5928, opened 2020-08-11
Remove broken feed from legacy base template

chosak:
The legacy base_update.html template currently includes a header link to a [/comments/feed/](https://www.consumerfinance.gov/comments/feed/) URL that no longer exists and returns 410 gone.

An example live page that uses this template is https://www.consumerfinance.gov/data-research/consumer-complaints/.

This commit removes the link to this feed.

## Checklist


willbarton:
> 410 gone

I do love this status.

PR #5931, opened 2020-08-12
Fix reports-as-webpages ToC sizes/weights

wpears:
Depends (and is based on) #5908 

Adjusts reports-as-webpages ToC styles to meet the spec


wpears:
Closing for PR hygiene

PR #5934, opened 2020-08-13
ParseLinksMiddleware shouldn't error if responses don't contain Content-Type

wpears:
Prevents 500s when content-type is mysteriously unset

### DIFF
--- a/cfgov/core/middleware.py
+++ b/cfgov/core/middleware.py
@@ -60,7 +60,10 @@ class ParseLinksMiddleware(object):
 
     def __call__(self, request):
         response = self.get_response(request)
-        if self.should_parse_links(request.path, response['content-type']):
+        if self.should_parse_links(
+            request.path,
+            response.get('Content-Type', '')
+        ):
             response.content = parse_links(
                 response.content,
                 request.path,
--- a/cfgov/core/tests/test_middleware.py
+++ b/cfgov/core/tests/test_middleware.py
@@ -47,6 +47,12 @@ class TestShouldParseLinks(TestCase):
             response_content_type='application/json'
         ))
 
+    def test_should_not_parse_links_if_empty(self):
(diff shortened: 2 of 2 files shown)


cfgov/core/middleware.py
```diff
@@ -60,7 +60,10 @@ def __init__(self, get_response):
 
     def __call__(self, request):
         response = self.get_response(request)
-        if self.should_parse_links(request.path, response['content-type']):
+        if (
```
chosak, line 63: Can we use the built-in [`HttpResponse.content_type`](https://docs.djangoproject.com/en/2.2/ref/request-response/#django.http.HttpRequest.content_type) attribute here instead? This should properly handle cases when the content type hasn't been set, [falling back](https://docs.djangoproject.com/en/2.2/ref/request-response/#django.http.HttpResponse.__init__) to an appopriate default.

higs4281: The flood of server errors are from document requests, which should have a response content-type of 'application/pdf' So I don't think we should fall back to 'text/html' – which would invoke unneeded link parsing.

I think if content-type is not available, using `get` and providing a blank default would be a safer way to stop the server errors and avoid unwanted link parsing while we dig further. Like this:

```
        response = self.get_response(request)
        available_content_type = response.get('content_type', '')
        if self.should_parse_links(request.path, available_content_type):
```

willbarton: @chosak Hmm, `HttpResponse` doesn't seem to offer a `content_type` attribute (you linked to `HttpRequest`). It looks like the `content_type` argument on `__init__` [is used to populate `'Content-Type'` in the dict-like headers](https://github.com/django/django/blob/master/django/http/response.py#L60-L62). It *should* get a default, but clearly somewhere along the way it's not.

I think I'm with @higs4281 and @wpears — we could just use a `response.get('content_type')` here without much downside.

PR #5950, opened 2020-08-21
Fix wellbeing server errors caused by translations

higs4281:
After the deployment of #5939, we started getting server errors on the
Spanish results page for the wellbeing questionnaire.

The server errors are caused by a KeyError originating in the
wellbeing view, which was translating keys in a data dict.
It turns out we don't need to translate those keys, because the template
can handle any needed translations.

## Testing
Locally, this page should raise a 500 error:

http://localhost:8000/consumer-tools/financial-well-being/results/es/

If you pull in this branch, it should render again.

## Note
There remains an odd translation failure for the "Age" comparison button, which should read
"Edad" on the Spanish results page, but instead is blank.

The translation in the cfgov/locale/es/LC_MESSAGES/django.po file is correct.
I tried retyping the entry and regenerating the django.mo file, to no effect.
This is not related to the current bug, and should not hold up its resolution.


willbarton:
@higs4281 those keys need to be wrapped in a gettext call somewhere if we want Django's automated `makemessages` to pick them up and not comment them out, unfortunately.

Scotchester:
@willbarton They are all in the templates.

higs4281:
The keys are generated values in the template: https://github.com/cfpb/consumerfinance.gov/blob/master/cfgov/wellbeing/jinja2/wellbeing/results.html#L637, so they are not wrapped there.

I can take another stab at resolving the KeyError, but in general I wish we could avoid translating keys in code.

PR #5951, opened 2020-08-24
Rural and underserved areas tool: IE11 fixes and code comments

anselmbradford:
IE11 has an error,  `Unable to get property 'remove' of undefined or null reference`, which is happening in IE11 when submitting an address in the rural and underserved areas tool. This error is due to the passed in SVG element not having a `classList` property. This PR adds a polyfill for the classList API on the SVG element.

## Changes

- Add SVG classList polyfill for IE11.
- Remove `return` in methods that return result of `_mutate` method, since that method doesn't return anything and will always be `undefined`.
- Add some missing jsdoc comments.
- Remove some obfuscation from `removeClass` method.

## How to test this PR

1. Visit https://www.consumerfinance.gov/rural-or-underserved-tool/ in IE11 in Sauce Labs. Submit an address and see the error that occurs in the dev console.
2. Pull this branch, build it, run Sauce Connect proxy and visit http://localhost:8000/rural-or-underserved-tool/ in IE11 in Sauce Labs. Submit an address and see that there is no error in the dev console.

## Screenshots

Screen Shot 2020-08-24 at 11 43 53 AM


contolini (approved):
Runs smooth through Sauce. 🎉

PR #5953, opened 2020-08-24
Handle localStorage logic in case it is blocked

anselmbradford:
If cookies are completely blocked, an error is thrown `Failed to read the 'localStorage' property from 'Window': Access is denied for this document.`

## Changes

- Utilize `web-storage-proxy` utility script where localStorage/sessionStorage is used.
- Try…catch to see if localStorage is available, and if not, gracefully fall back.

## How to test this PR

1. Go into Chrome settings via Settings > Privacy and Security > Cookies and other site data
2. Check "Block all cookies (not recommended)"
3. Visit https://www.consumerfinance.gov/owning-a-home/ and scroll down and see errors in dev console.
3. Pull branch and build and run locally. Visit http://localhost:8000/owning-a-home/ and scroll down and see that there are no errors in the dev console.

## Notes and todos

- The ["orange-explorer"](https://github.com/cfpb/consumerfinance.gov/blob/master/cfgov/hmda/templates/hmda/orange-explorer.html) has some direct references to localStorage too, but that page is a large inlined frankenstein monster I'm not going to touch in this PR. EDIT: This tool will be sunset [next month](https://github.com/cfpb/consumerfinance.gov/pull/5952)!


cfgov/unprocessed/js/modules/util/email-popup-helpers.js
```diff
@@ -1,8 +1,18 @@
...
 import throttle from 'lodash.throttle';
+import webStorageProxy from './web-storage-proxy';
+
+// If cookies are turned off, we set localStorage variables to an empty object.
+let _localStorage;
+
+try {
+  _localStorage = window.localStorage;
+} catch ( err ) {
+  _localStorage = {};
```
wpears, line 11: Can this be set to `null` for parity with how it's handled in analytics-gtm?

wpears (changes requested):
Just the one question

cfgov/unprocessed/js/modules/util/email-popup-helpers.js
anselmbradford: Unlike analytics-gtm, which bails out if `_localStorage` is `null`, here it's passed into `webStorageProxy` in multiple methods. So if we wanted email helpers and analytics-gtm to both be null or both be empty objects, we either would need logic to check if it's `null` in each method in the email popup helper, which doesn't seem great, or we make it an object in analytics-gtm and check if it's an empty object, which doesn't seem great either. So that's why it's different in each.

wpears: Then begrudgingly approved!

wpears (approved):

cfgov/unprocessed/js/modules/util/email-popup-helpers.js
anselmbradford: @wpears ! I thought about this further and added a null check inside the web-storage-proxy, now it can be null in email popup helpers! https://github.com/cfpb/consumerfinance.gov/pull/5953/commits/4d3627c2a53aae9dd5a5db2a7312edeae948d0b0

wpears (approved):
Giving the rare double approve on this fine piece of work
:+1: :+1:

PR #5954, opened 2020-08-24
Mega Menu: Change heading HTML elements to ARIA attributes

anselmbradford:
## Changes

- Change h2 and h3 mega menu headings to `role="heading"` and `aria-level`.
- Change `section` in mega menu to `div` as otherwise we get a warning that it doesn't have a heading.

## How to test this PR

1. This change removes the mega menu from the rotor headings list and from the document outline. It's unclear if the voiceover change would be reflected in an actual screenreader. Scott is going to try this out in JAWS.


anselmbradford:
@Scotchester Big thank you Scott!!

PR #5956, opened 2020-08-24
Properly validate translations

chosak:
The existing validate-translations tox environment (introduced in #5939) doesn't properly validate translations, because it tests the wrong path.

The tox config tries to test files under cfgov/locale, but it was using a changedir setting that changed the working directory to cfgov. So in practice it wasn't actually testing any files. Currently master does generate some changes in translation files, but these were being missed because of this misconfiguration.

This change modifies the tested path from cfgov/locale to just locale, so that the right files are tested.

## Notes and todos

There are some outstanding changes to the translation files that need to be fixed before tox will pass, mostly seeming to do with the wellbeing app. @higs4281 is this somehow related to #5950?

## Checklist


chosak:
The validate-translations failure is expected, due to the outstanding translations changes, but this demonstrates the necessity of this PR to test the right file:

```
validate-translations run-test: commands[0] | python manage.py makemessages
processing locale es
validate-translations run-test: commands[1] | python manage.py compilemessages
processing file django.po in /home/runner/work/consumerfinance.gov/consumerfinance.gov/cfgov/locale/es/LC_MESSAGES
validate-translations run-test: commands[2] | git diff-index --quiet HEAD -- 'locale/*/LC_MESSAGES/django.mo'
ERROR: InvocationError for command /usr/bin/git diff-index --quiet HEAD -- 'locale/*/LC_MESSAGES/django.mo' (exited with code 1)
```

higs4281:
> is this somehow related to #5950?

No doubt. I'll check it out.

higs4281:
#5939 Took translations out of the view to stop sever errors and allow the Spanish results page to render. 
The immediate problem was that some view-based translations were generating a KeyError.

I didn't realize the translation test was not running correctly. Unfortunately, if we fix the translation test, the wellbeing results page would break again, because the fixed translation test will overwrite the translations created for the generated values that reach the template.

willbarton:
Doh!

willbarton:
@chosak I think we're going to have a problem with the date stamp being updated in the .po file — even if there are no other changes, a `makemessages` run will update that.

chosak:
@willbarton happily, edits to the timestamp in the .po file doesn't affect the .mo file, which is what we diff in the check.

willbarton:
Ah! Yes, sorry. I had a moment.

chosak:
Can't merge this until #5965 is in.

chosak:
I think I've gotten this PR back into committable shape.

For the record, if developers make changes without regenerating the .mo file, doing these things will fail translation validation:

1. Manually editing translated text in a .po file
2. Removing all translations of a string from code (will cause entry in .po file to be commented out and moved to bottom of file)

These will NOT fail translation validation, as they don't affect the .mo file:

1. Making unrelated edits to source code in such a way that affects line numbers in the .po file
2. Adding new uses of translation (`_('something new')`) in code (default is translation works as a no-op)

The fact that adding a new use of `_()` doesn't require manually [dealing with translations](https://cfpb.github.io/consumerfinance.gov/translation/) while removing a use of `_()` potentially does feels unbalanced and potentially annoying to developers, but hopefully the benefit of keeping our .po files tidy is worth it.

PR #5959, opened 2020-08-25
Clean up Snyk cruft and unnecessary child app cruft

Scotchester:
## Removals

- Snyk workaround requirements files that are no longer needed, since they now support files not named exactly `requirements.txt`
- npm/Yarn-related files for child apps that have no Node package dependencies

## Checklist


cfgov/unprocessed/apps/hmda/.yarnrc
```diff
@@ -1 +0,0 @@
-yarn-offline-mirror "./npm-packages-offline-cache"
```
anselmbradford, line 1: Say a `package.json` is added in the future to this app and yarn is run, would the developer be expected to add this `.yarnrc` file back at that point? And what happens if they don't?

cfgov/unprocessed/apps/form-explainer/yarn.lock
```diff
@@ -1,4 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
anselmbradford, line 1: I suspect this file will pop up again, as these autogenerated yarn.lock files came up for me through some process of flushing the yarn cache and re-installing everything.

cfgov/unprocessed/apps/hmda/.yarnrc
Scotchester: I think it's fair to expect the developer to know to re-add the .yarnrc file, yes.

cfgov/unprocessed/apps/form-explainer/yarn.lock
Scotchester: I was assuming that nothing would appear, given that there is no package.json file anymore. But I could be wrong. Are you saying that _on this branch_, the `yarn.lock` files reappeared for you?

cfgov/unprocessed/apps/hmda/.yarnrc
anselmbradford: And if they don't re-add it, what happens? Does it use the npm cache at the repo root? Maybe we don't need these .yarnrc files in the apps at all? Perhaps a question for @wpears

cfgov/unprocessed/apps/form-explainer/yarn.lock
anselmbradford: Yes. If you throw out all the cache files and lock files and re-`yarn install`, we not only get things in the cache that should be updated (which seems like we should do periodically), but we also get these auto-generated empty lock files:

Screen Shot 2020-08-25 at 2 13 00 PM

Scotchester: Strange. I don't understand why Yarn would do anything in that folder in the absence of a `package.json` file. Maybe this is added incentive for us to try out Yarn 2, which I believe overhauled the lock system.

cfgov/unprocessed/apps/hmda/.yarnrc
wpears: IIRC because of the way we treat these apps as kinda separate sub-apps with their own `yarn.lock`, the `.yarnrc` file is needed (when js is present) in order to trigger use of the offline cache. Without it, it won't default to the parent cache, and even trying to reference the parent cache is somewhat fraught/buggy from my testing.

Ideally, imo, these apps shouldn't really be treated as their own sub-apps from a build standpoint so that they can be better bundled/share code with the main bundle more completely. Such a change would also obviate the need for these sub-app config files.

In sum, I support the removal of this specific `.yarnrc` file, the maybe the hope that in the future we will have a simpler build system and can then remove all of them except the top-level one!

anselmbradford: > Ideally, imo, these apps shouldn't really be treated as their own sub-apps from a build standpoint so that they can be better bundled/share code with the main bundle more completely. Such a change would also obviate the need for these sub-app config files.

I think the one situation we don't want to be in, whatever the setup, is where there's a shared dependency between say the owning-a-home rate checker (complex app) and somewhere else on the site that is far simpler. Someone goes in and updates the dependency on the far simpler thing, and either the rate checker is overlooked or it "looks fine" because there isn't a clear product owner and it's unclear how it should function. Months later it's discovered it's not fine and the team is faced with untangling what went wrong, while simultaneously keeping in mind what's changed in the site elsewhere. Or the other direction where the simpler area of the site is blocked from being updated because of the changes needed in the rate checker to allow the update to happen.

Scotchester: Yeah, I think that's a good argument for keeping them separate, Ans.

cfgov/unprocessed/apps/form-explainer/yarn.lock
Scotchester: Added ab34f192c8bb99979ec17d2d2370fdc9df73400b to prevent us from trying to run `yarn` when `package.json` exists but has no deps listed.

anselmbradford (approved):
Change looks good! 👍 

I'm still a bit apprehensive about new folders being copied from the old ones and `.yarnrc` files not being added when needed, but that's easy enough to fix if it happens.

PR #5964, opened 2020-08-26
Carousel: Add text to image link

anselmbradford:
The carousel image is a link, but it only contained an image. This PR adds visually hidden text for screen readers.

## Additions

-  Add text to image link in carousel.

## How to test this PR

1. Pull branch and run site and generate a lighthouse report for the homepage. It should not flag the carousel image link.


Scotchester:
Is this better than having the two adjacent links with redundant text? How do screen readers currently read the link around the image?

Scotchester:
> Looks good, @anselmbradford.
> 
> @Scotchester currently the desktop Lighthouse tests generate an error:
> 
> > Links do not have a discernible name: Link text (and alternate text for images, when used as links) that is discernible, unique, and focusable improves the navigation experience for screen reader users. [Learn more](https://dequeuniversity.com/rules/axe/3.1/link-name?application=lighthouse).
> 
> This PR fixes that error.

I understand that, but my question is whether fixing that error actually improves the user experience, or in fact makes it worse :)

anselmbradford:
@Scotchester Looks like I have the goldfish memory this time 😛 https://github.com/cfpb/consumerfinance.gov/pull/5433#issuecomment-571799667

So what we have is essentially "Failure Example 4" on https://www.w3.org/TR/2013/NOTE-WCAG20-TECHS-20130905/H2, except we don't have two directly adjacent links, as one is in a div. I still think having the duplicate links labelled is the best we can do since the image link is focusable, so having it labeled describes what's there, which is two links next to each other. I suppose we could make the label "Image: Access our resources" or something to have it differentiated from the text link next to it? What do you think about that?

Scotchester:
Heh, yeah. Again, I think we should do the work to combine these into a single `<a>` element, but as I said in the prior discussions, this is a small improvement over the current situation.

In this case, at least in VoiceOver, when you tab over to the link, it reads the final segment of the `href`, so in this example:
carousel on 2nd item with image focused
it reads "link: bureau taking action build more inclusive financial system", and I agree it's better to use the same text as the link instead, to make it more clear that the two links go to the same place.

anselmbradford:
@Scotchester Yeah let me know if you see a way to combine them, but since they aren't directly adjacent links in the markup I don't see an easy way without much hackery. 

Just to confirm, do you think having at a link label on the image, but prefixing it with `"Image: `, is the best path forward for this PR and improves on what's there?

Scotchester:
> @Scotchester Yeah let me know if you see a way to combine them, but since they aren't directly adjacent links in the markup I don't see an easy way without much hackery.

I'll try to find time to take a look at it.

> Just to confirm, do you think having at a link label on the image, but prefixing it with `"Image: `, is the best path forward for this PR and improves on what's there?

No, I think the PR is fine as is. Calling out the fact that it's an image doesn't seem any more helpful.

PR #5965, opened 2020-08-27
Fix Financial Wellbeing translations again

willbarton:
A followup to This change restores gettext around the FWB groups and changes those groups from `OrderedDict`s to a list of tuples. This should avoid any key-errors with translated strings when looping over `group_means` and still allow those strings to be picked up by gettext.

It also replaces `gettext_lazy` in the FWB views with [`gettext_noop`](https://docs.djangoproject.com/en/3.1/topics/i18n/translation/#marking-strings-as-no-op), since we want to mark the strings for translation but do not want them actually translated when the view is evaluated (that happens in the template).

I've also added gettext to the Dockerfile so that it's available in local development without having to yum install it.

`ResultsView` also contained a `SCORING` attribute duplicated from `wellbeing.forms.ResultsForm` that wasn't being used. I've removed it.

Finally, I've documented that we can use gettext_noop to mark strings for translation but not translate them.

This is a follow-on to #5939, #5950, and related to #5956. Hopefully I've gotten this right this time.

## Checklist


docs/translation.md
```diff
@@ -98,25 +98,25 @@ To ensure that strings in templates are picked up in message extraction (`django
...
 
-**Do not** include the `_()` call in a larger Jinja2 template data structure:
+#### Do not include the `_()` call in a larger Jinja2 template data structure:
 
 ```diff
 + {% set link_text = _("Visit our website") %}
-  {% set link_data = {
-      "url": "https://consumerfinance.gov",
+{% set link_data = {
+"url": "https://consumerfinance.gov",
```
chosak, line 106:
```suggestion
    "url": "https://consumerfinance.gov",
```

cfgov/wellbeing/views.py
```diff
@@ -1,7 +1,7 @@
 from collections import OrderedDict
 
 from django.shortcuts import render
-from django.utils.translation import gettext_lazy as _
+from django.utils.translation import gettext_noop as _
```
chosak, line 4: This is super clever, really nice find.

PR #5967, opened 2020-08-27
Fix Wagtail Redirect Issue

alexm118:
Wagtail Redirects are case-sensitive, but we introduced new 404 handling to allow for URLs to be case-insensitive. By increasing the priority in which we run the Wagtail RedirectMiddleware we can allow for wagtail to handle its redirect prior to django applying our custom 404 handler, this will allow us to have Wagtail Redirects run as well as urls match on case-insensitive resolution.

---

## Changes

- Move the 404 handler to a SelfHealingMiddleware and prioritize it above Wagtail's RedirectMiddleware

## How to test this PR

Locally with DEBUG=False

1. curl -I http://localhost:8000/MyNewMoneyGoal/
2. curl -I http://localhost:8000/data-research/HMDA/

Both URL's should return a 301

## Screenshots

image

## Notes and todos

- The ordering of these two middleware matter, as the SelfHealingMiddleware will always run if a 404 is thrown, and likely return a new path so running the Wagtail RedirectMiddleware after should add its result above that of the SelfHealingMiddleware.

## Checklist


cfgov/core/middleware.py
```diff
@@ -99,3 +100,29 @@ def __call__(self, request):
         response = self.get_response(request)
         translation.deactivate()
         return response
+
+
+class SelfHealingMiddleware:
```
Scotchester, line 105: Let's copy over the docstring from the previous 404 handling method to retain the explanation of what exactly the intention is here.

```suggestion
class SelfHealingMiddleware:
    """Attempt to self-heal 404-ing URLs.	
    Takes a 404ing request and tries to transform it to a successful request	
    by lowercasing the path and stripping extraneous characters from the end.	
    If those result in a modified path, redirect to the modified path.	
    If the path did not change, this is a legitimate 404, so continue handling	
    that as normal.	
    """
```

cfgov/core/middleware.py
```diff
@@ -99,3 +100,29 @@ def __call__(self, request):
...
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+
+        if response.status_code != 404:
```
Scotchester, line 112: Just a little extra clarity:

```suggestion
        # If this request isn't 404ing, just return the existing response.
        if response.status_code != 404:
```

PR #5969, opened 2020-08-27
Support the option to perform a delete purge of Akamai cache

cwdavies:
This task is to add the option to call Akamai's [delete purge API](https://developer.akamai.com/api/core_features/fast_purge/v3.html#apisummary) to our [Akamai frontend cache backend](https://github.com/cfpb/cfgov-refresh/blob/master/cfgov/v1/models/caching.py), add a corresponding management command that can perform it

## Additions

- Added `delete(url)` in `caching.py` to support the delete of Akamai cache
- `cfgov/v1/management/commands/delete_page_cache.py`
- `cfgov/v1/management/commands/delete_all_pages_cache.py`

## Changes

- Renamed `invalidate_page.py` to `invalidate_page_cache.py`
- Renamed `invalidate_all_pages.py` to `invalidate_all_pages_cache.py`

## How to test this PR

1. tox

## Notes and todos

-

## Checklist

#### Browser testing

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Safari on iOS
- [ ] Chrome on Android

#### Accessibility

- [ ] Screen reader friendly

#### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


cfgov/v1/models/caching.py
```diff
@@ -52,39 +52,41 @@ def get_auth(self):
...
-    def get_payload(self, obj):
+    def get_payload(self, obj, action):
         return {
-            'action': 'invalidate',
+            'action': action,
             'objects': [obj]
         }
 
-    def purge(self, url):
+    def purge(self, url, action):
```
willbarton, line 61: I'm not sure if this is going to work. Frustratingly, the Wagtail Frontend Cache backend API [is not really documented](https://docs.wagtail.io/en/latest/reference/contrib/frontendcache.html#), but the `purge()` method signature [doesn't include an `action` argument](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/frontend_cache/backends.py#L21). The [signal handlers](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/frontend_cache/signal_handlers.py#L7-L12) ends up [calling `purge_batch(urls)`](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/frontend_cache/utils.py#L97) on the backend, which in turn calls `purge(url)`. 

I think for now it might be worth moving the actual `post` to the Akamai API to another private method, leaving the `purge(url)` method signature alone, and maybe implementing a separate `delete(url)` method that constructs the "delete" payload. As it is I think this will cause a 500 when we publish or unpublish a page.

cwdavies: Left `purge(url)` method alone, and implemented a new `delete(url)` method in `caching.py`

cfgov/v1/models/caching.py
```diff
@@ -52,39 +52,80 @@ def get_auth(self):
...
+        )
+        logger.info(
+            u'Attempted to {action} cache for page {url}, '
+            'got back response {message}'.format(
+                action=action,
+                url=url,
+                message=resp.text
+            )
+        )
+        resp.raise_for_status()
```
willbarton, line 77: I don't think all of this needs to be repeated two times for single-urls and two times for "all".

You could still combine these into one, method, and then just have `purge(url)` call that method with `action="invalidate"`.

You could still combine the two single-url methods and the two "all" methods into one that takes "action" as an argument, but just retain the existing method signature for `purge()`, and have `purge()` call that other method.

cwdavies: I have simplified Akamai cache code to post a request that will invalidate or delete purge

cfgov/v1/models/caching.py
```diff
@@ -52,45 +52,59 @@ def get_auth(self):
...
-                obj=obj,
+                action=action,
+                url=url,
                 message=resp.text
             )
         )
         resp.raise_for_status()
 
+    def purge(self, url):
+        self.post(self, url, 'invalidate')
```
willbarton, line 103:
```suggestion
        self.post(url, 'invalidate')
```

cfgov/v1/models/caching.py
```diff
@@ -52,45 +52,59 @@ def get_auth(self):
...
+
+    def purge_all(self):
+        self.post_all(self, 'invalidate')
```
willbarton, line 106:
```suggestion
        self.post_all('invalidate')
```

cfgov/v1/models/caching.py
```diff
@@ -52,45 +52,59 @@ def get_auth(self):
...
+    def get_payload(self, obj, action):
         return {
-            'action': 'invalidate',
+            'action': action,
             'objects': [obj]
         }
 
-    def purge(self, url):
+    def delete(self, url):
+        self.post(self, url, 'delete')
```
willbarton, line 62:
```suggestion
        self.post(url, 'delete')
```

cfgov/v1/models/caching.py
```diff
@@ -52,45 +52,59 @@ def get_auth(self):
...
+
+    def delete_all(self):
+        self.post_all(self, 'delete')
```
willbarton, line 65:
```suggestion
        self.post_all('delete')
```

cwdavies: Removed self from post_all

cfgov/v1/models/caching.py
cwdavies: Removed self from post

cfgov/v1/models/caching.py
cwdavies: Removed self from post_all

cfgov/v1/models/caching.py
cwdavies: Removed self from post

PR #5970, opened 2020-08-27
Global Search: innerText fix for FF44

anselmbradford:
`innerText` is not supported in Firefox 44 and earlier. New Relic reported some users running into this. This PR works in a fallback to `textContent` which together should have almost universal coverage.

## Changes

- Adds fallback to `textContent` for `innerText` usage in the Global Search JS.

## How to test this PR

1. Pull branch, run locally and run Sauce Connect and visit the site in Sauce Labs with Firefox 44 or earlier and there should be no error in the dev console, compared to the live site, which has an error and the search doesn't open.


anselmbradford:
Superseded by https://github.com/cfpb/consumerfinance.gov/pull/5974

PR #5972, opened 2020-08-27
Create new CSP allowlist for media files

Scotchester:
This is needed to be able to use our new audio players that serve files uploaded to the Wagtail Media library, landing on files.consumerfinance.gov.

## Additions

- `CSP_MEDIA_SRC` setting for django-csp

## Removals

- `files.consumerfinance.gov` entries in other CSP lists, when the `*.consumerfinance.gov` wildcard should handle it just fine (we had added the explicit reference in a fit of haphazard troubleshooting, but the problem was not that)

## How to test this PR

Don't think we really can until it gets to production, unless the buildfiles bucket actually works somewhere that this branch can be deployed to.

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -148,7 +148,7 @@ MIDDLEWARE = (
     "core.middleware.DownstreamCacheControlMiddleware",
     "flags.middleware.FlagConditionsMiddleware",
     "core.middleware.SelfHealingMiddleware",
-    "wagtail.contrib.redirects.middleware.RedirectMiddleware",    
+    "wagtail.contrib.redirects.middleware.RedirectMiddleware",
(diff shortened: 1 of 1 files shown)


cfgov/cfgov/settings/base.py
```diff
@@ -631,6 +627,12 @@
     "*.qualtrics.com",
 )
 
+# These specify valid media sources (e.g., MP3 files)
+CSP_IMG_SRC = (
```
wpears, line 631: Isn't this duplicative of the `CSP_IMG_SRC` above? From the comment it seems like this should be `CSP_MEDIA_SRC`

Scotchester: 🤦 And this is why we have code review even for simple changes! :)

Scotchester: Fixed. Thanks for the catch!

PR #5983, opened 2020-09-01
Update primary branch name in GitHub Actions and docs

Scotchester:
## Changes

- Replaces a bunch of instances of the prior primary branch name with `main`

## Checklist


README.md
```diff
@@ -1,8 +1,10 @@
 # consumerfinance.gov
 
-[![Build Status](https://github.com/cfpb/consumerfinance.gov/workflows/test/badge.svg?branch=master)](https://github.com/cfpb/consumerfinance.gov/actions)
+<!--
+[![Build Status](https://github.com/cfpb/consumerfinance.gov/workflows/test/badge.svg?branch=main)](https://github.com/cfpb/consumerfinance.gov/actions)
+-->
```
Scotchester, line 5: I realized that our badge wasn't loading because we no longer have a single, unified workflow that tests the primary branch on merges to it. We should decide if we want to create something like that just for the build badge, or let the badge go. I'm not sure how much value there is in it when we prevent merges with failing builds?

willbarton: Well, theoretically the build badge should reflect like, regular builds of `main`. 

But I'm okay losing it if it doesn't make sense the way CI is set up right now.

Scotchester: Yeah, my point is that we have regular builds of what ends up being `main` on each PR, but (a) as far as I know, there's no way to get a badge for "most recent merged PR's build status"; and (b) even if there was, we don't have a single unified check, we have a handful; and (c) even if we did, wouldn't the badge theoretically always be green, since we require passing checks to merge PRs?

Scotchester:
Coverage checks are failing because we don't have a coverage report on `main` yet. Unsure if this is a situation that will resolve itself, or if there's something else we need to do to get one generated there.

willbarton:
@Scotchester Oh, I don't think that's right — I think we need to pass `main` to `diff-cover` explicitly: https://diff-cover.readthedocs.io/en/latest/README.html#compare-branch

PR #5991, opened 2020-09-03
Remove jsonp functions from webpack configs

anselmbradford:
## Changes

- ~~Fix jsonp functions that had incorrect names.~~
- Remove jsonp functions from webpack configs.

## How to test this PR

1. All should be as before. ~~These names just should be unique.~~


Scotchester:
I don't generally understand Webpack, but if we're not using these identifiers elsewhere in the codebase, do we even need the `jsonpFunction` config lines?

anselmbradford:
> I don't generally understand Webpack, but if we're not using these identifiers elsewhere in the codebase, do we even need the jsonpFunction config lines?

I'd need to investigate if we actually need them, but https://webpack.js.org/configuration/output/#outputjsonpfunction says the default is `webpackJsonp` and they are used when "target is set to 'web'", which is the default. They should be unique per bundle at least, which is what I'm doing here, but maybe could also be removed, but from reading that I think they'd all get the name `webpackJsonp`.

contolini:
I'm 96% sure we don't need this config option because we use our own [JS loader](https://github.com/cfpb/consumerfinance.gov/blob/146a9af377e85617210cf2ba57d7bfac1a317361/cfgov/jinja2/v1/_layouts/base.html#L174-L177) and don't dynamically generate or load chunks (with the exception of TDP's curriculum review tool that inherits create-react-apps' [webpack config](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/config/webpack.config.js#L219)).

So even though they're harmless I'd vote to delete these lines to avoid confusion.

anselmbradford:
I removed the jsonp settings. Ready for review again!

Scotchester (approved):

PR #5997, opened 2020-09-10
Only check live EventPages for archiving; append date to duplicate slugs

Scotchester:
Drafts that got archived were causing errors when, later, pages with the same slug were attempted to be archived again. This update only checks for live pages to archive (and only EventPages now), and if it finds a duplicate slug, it will append the event's start date to the slug before moving it.

## How to test this PR

1. Refresh your DB with last night's dump
1. Pull branch
1. Run `./cfgov/manage.py runscript archive_events`
1. Verify that it returns `No past events to be archived were found.`
1. Publish the draft current events with dates in the past
1. Run `./cfgov/manage.py runscript archive_events` again
1. Verify that it moves them to the archive, with their start date appended to the slug

## Notes and todos

- [x] Convert to management command
- [x] Add unit tests
- [ ] Update Jenkins job to use the new management command

## Checklist


cfgov/scripts/archive_events.py
```diff
@@ -23,21 +24,26 @@ def run():
...
+            for event in live_event_pages:
+                event = event.specific
+                if event.end_dt:
+                    if event.end_dt < timezone.now():
+                        if event.can_move_to(archived_events):
+                            try:
+                                event.move(archived_events, pos='last-child')
+                            except ValidationError:
+                                iso_date = event.start_dt.date().isoformat()
+                                event.slug = event.slug + '-' + iso_date
```
higs4281, line 39: Just wondering if it would be more useful to know when the event was archived, as opposed to when its original start date was. If so, you could convey that with `event.slug + '-archived-' + datetime.date.today().isoformat()`

higs4281 (commented):
I had one minor question about the slug suffix.

I think management commands are best as wrappers for tested scripts. So rather than repeat the script code here, could we call the script's run() function?

cfgov/scripts/archive_events.py
Scotchester: I think that for the purposes of differentiating these URL slugs, it's more natural to indicate the date that the event occurred (or started).

Scotchester:
> I think management commands are best as wrappers for tested scripts. So rather than repeat the script code here, could we call the script's run() function?

Yeah, I'm gonna end up with just the management command. I just left the existing script in place for the time being to make it easier to see the changes I was making to the logic of it. I'll pull it out before merging.

cfgov/scripts/archive_events.py
higs4281: I can see the usefulness of having the start date in the slug. Maybe no one cares when it was archived.

higs4281:
Otherwise, the script works as advertised.

cfgov/v1/management/commands/archive_events.py
```diff
@@ -0,0 +1,62 @@
...
+    BrowseFilterablePage, EventArchivePage
+)
+from v1.models.learn_page import EventPage
+
+
+class Command(BaseCommand):
+    help = 'Move event pages whose end date has passed into the archive'
+
+    def handle(self, *args, **options):
+        self.stdout.write('Searching for events to archive…')
```
higs4281, line 15: Is unicode ellipsis a preference? It was [explicitly removed](https://github.com/cfpb/consumerfinance.gov/commit/7e0721d) previously.

Scotchester: Yeah, personal preference. I see that the PR for the linked commit notes that it shouldn't be a problem now that we're on Python 3, though sticking with the dots may be worth doing to be extra cautious. @chosak Would you still take that position?

higs4281 (approved):
I have no objection to unicode ellipsis if previous troubles were related to Python2.

The management job works, appending a start date if a duplicate slug is found.

cfgov/v1/management/commands/archive_events.py
chosak: Unicode away! In Python 2 I believe we needed lines [like this](https://github.com/cfpb/consumerfinance.gov/blob/146a9af377e85617210cf2ba57d7bfac1a317361/cfgov/paying_for_college/apps.py#L1) but in Python 3 [the default encoding is already UTF-8](https://docs.python.org/3/howto/unicode.html#the-string-type).

PR #6000, opened 2020-09-14
Add missing `key-terms` slug to DefinedTerm URL tag

Scotchester:
Our contacts at Google alerted us to an oversight of mine when I implemented our DefinedTerm markup: the URL for each term that goes in the `<link itemprop="url" href="{URL}">` tag was missing the `key-terms/` part of the URL. Turns out, our use of RoutablePageMixin to generate the Key Terms pages meant that the `page.full_url` tag was not including that `key-terms/` part of the URL.

~~I tried using the `routablepageurl` template tag, but that only returns a root-relative URL, whereas we need an absolute URL, and it required hardcoding a `category='key-terms'` kwarg, anyway, so we might as well just hardcode it this way.~~

/cc @kelleyholden 

## How to test this PR

1. Visit http://localhost:8000/consumer-tools/auto-loans/answers/key-terms/, inspect the markup for a term, and see that the `<link itemprop="url" …>` tag's `href` attribute does not include the `key-terms/` path segment
1. Pull branch
1. Reload the page and see that the `key-terms/` is now included

## Checklist

### DIFF
--- a/cfgov/ask_cfpb/models/pages.py
+++ b/cfgov/ask_cfpb/models/pages.py
@@ -213,6 +213,7 @@ class PortalSearchPage(
     portal_category = None
     query_base = None
     glossary_terms = None
+    category_slug = None
     overview = models.TextField(blank=True)
     content_panels = CFGOVPage.content_panels + [
         FieldPanel('portal_topic'),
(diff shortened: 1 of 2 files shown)


Scotchester:
Thanks for the assist, @higs4281! Updated it to insert the slug in the term URL.

higs4281 (approved):

PR #6005, opened 2020-09-14
Reduce padding on top level mega menu items

wpears:
Addresses https://github.local/CFPB/super-regular/issues/145

Reduces padding by 2px on top level mega menu items in order to accommodate slightly longer titles without introducing line breaks.

Before:
Screen Shot 2020-09-14 at 2 37 58 PM

After:
Screen Shot 2020-09-14 at 2 38 15 PM


cfgov/unprocessed/css/organisms/mega-menu.less
```diff
@@ -902,10 +902,11 @@ body {
                     }
                 }
 
+                //Adjusts position by 2px to accommodate longer titles
                 &-icon-closed .cf-icon-svg,
                 &-icon-open .cf-icon-svg {
                     position: absolute;
-                    right: unit( 24px / @base-font-size-px, em );
+                    right: unit( ( 24px - 3.2 ) / @base-font-size-px, em );
```
anselmbradford, line 909: Should the 3.2 include units? Also, why does the code comment say 2px and this is a fractional value?

cfgov/unprocessed/css/organisms/mega-menu.less
```diff
@@ -902,10 +902,11 @@ body {
                     }
                 }
 
+                //Adjusts position by 2px to accommodate longer titles
```
anselmbradford, line 905: TRIVIAL: Add space after `//` for consistency with other comment.

cfgov/unprocessed/css/organisms/mega-menu.less
wpears: Arithmetic in less will default to the first units used, I could include them for consistency though here and below. The reason this is 3.2 is kinda convoluted. The computed font-size is 10px, meaning `em` keys off 10px and therefore 0.1 is equivalent to 1px. 3.2/16 (base-font-size-px) = 0.2em or 2px.

anselmbradford: Would using `rem` make more sense perhaps?

wpears:
@anselmbradford Do these changes work for you?

anselmbradford (approved):
Looks good. I think this makes the left/right side padding look more equal. Not sure if you want a team sign-off, but here's mine.

PR #6009, opened 2020-09-15
Filterable list control: Convert date inputs into datepicker fields

anselmbradford:
Most modern browsers have built in datepickers. Let's use them!

## Changes

- Changes date input field type to `date` in the filterable list control.

## How to test this PR

1. Pull branch and build.
2. Open /blog in various browsers and the iOS simulator and select a date.

## Screenshots

Chrome 
Screen Shot 2020-09-01 at 3 45 26 PM

Firefox
Screen Shot 2020-09-01 at 3 45 49 PM

Probably Opera
Screen Shot 2020-09-01 at 3 46 27 PM

Safari 😞 
Screen Shot 2020-09-01 at 3 46 56 PM

Edge
Screen Shot 2020-09-01 at 3 56 11 PM

iOS
ios-datepicker


cfgov/unprocessed/js/modules/util/validators.js
```diff
@@ -20,6 +20,14 @@ import ERROR_MESSAGES from '../../config/error-messages-config';
  */
 function date( field, currentStatus ) {
   const status = currentStatus || {};
+  let valueToEval = field.value;
+
+  /* TODO: Merge this into the regex checks.
+           This converts numbers from any other format to MM/DD/YYYY */
+  if ( !Number.isNaN( field.valueAsNumber ) ) {
```
contolini, line 27: IE11 doesn't support `Number.isNaN`. You can use `isNaN` though.

cfgov/v1/util/date_filter.py
```diff
@@ -2,20 +2,25 @@
...
+# text inputs, which occur in browsers that don't support the date input field
+# https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date
 
 # For more information on date formatting, see the python documentation here:
 # https://docs.python.org/2/library/datetime.html#strftime-and-strptime-behavior
 
 full_date_patterns = (
     '%m/%d/%Y',     # 10/25/2016, 9/1/2016
     '%m-%d-%Y',     # 10-25-2016, 9-1-2016
+    '%Y-%m-%d',     # 2016-10-25, 2016-9-1
```
contolini, line 15: `%m` and `%d` are zero-padded so it'd be cool if you could update all these comments accordingly (`2016-09-01`, `09-01-2016`, etc.) See https://docs.python.org/2/library/datetime.html#strftime-and-strptime-behavior.

contolini (changes requested):
Very graceful solution! Couple minor tweaks.

cfgov/v1/util/date_filter.py
anselmbradford: Good catch!

alexm118:
@anselmbradford Hey, just wanted to let you know there is an associated [cypress test](https://github.com/cfpb/consumerfinance.gov/blob/main/test/cypress/components/filterable-lists.js) for filterable lists. Looks like this change breaks one of the tests (unsurprisingly around the date)

anselmbradford:
Thanks @alexm118 ! I updated the associated test. This brings up the question of how does Cyprus fit into our code checks? Is the idea it will run in GitHub actions in PRs like this one (if it's possible to run headless), or will it run with `gulp test` like our other frontend tests, or something else? Unless I'm doing something wrong I needed to run `yarn run cypress open` and know to click into the filterable-list.js test, which seems like it would be quite easy to miss in the future.

alexm118:
@anselmbradford Currently, the cypress tests run in headless mode inside our jenkins server. Potentially once we have github hooks setup it may be possible for us to do something with Github Actions (we could possibly do this without jenkins, I just haven't dug into the potential work it would take to spin up the docker stack via Github Actions). 

I put some [documentation](https://cfpb.github.io/consumerfinance.gov/functional-testing/#running-cypress) in place for how Cypress is currently setup to run, there are a couple options basically one that leverages docker and another that allows you to run it via yarn. I find that running via yarn is helpful when modifying/creating tests and docker is easier to run when you just want to validate the tests will pass but YMMV.

contolini (approved):

PR #6010, opened 2020-09-15
Filters: Remove colons from form date input labels

anselmbradford:
## Removals

- Remove colons from form date input labels for consistency with other filter labels.

## How to test this PR

1. /blog date inputs labels should not have colons any longer.

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
+++ b/cfgov/jinja2/v1/_includes/organisms/filterable-list-controls.html
@@ -109,7 +109,7 @@
                                         <div class="m-form-field">
                                             <label class="a-label a-label__heading"
                                                    for="{{ form.from_date.id_for_label }}">
-                                                From:
+                                                From
                                             </label>
                                             {{ form.from_date }}
                                         </div>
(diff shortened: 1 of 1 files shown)


Scotchester:
Has this got a thumbs up from UX? I feel like the colons were probably an intentional decision to distinguish from the parent "Date range" label, but a better solution would probably be to drop the colons _and_ give the "From" and "To" a different style, like H5 style or something.

anselmbradford:
This came up in our team's sprint review. @stephanieosan have thoughts on Scott's comment?

sonnakim:
I'm not really a fan of removing the colon, because "Date", "From" and "To" are all the same font size right now, and to me having the colon there makes the hierarchical relationship among those labels clearer. Without the colons, "Date", "From" and "To" look like they're floating. I understand the reason you'd want to remove it, for consistency, but I think that would be better served if "From" and "To" could be a smaller font size, which would better show their subordinate relationship to the "Date" label. 

With colon 
image

Without colon
 image

anselmbradford:
Team would like to hold off this change. Closing.

PR #6015, opened 2020-09-17
Archiving: basic functionality

virginiacc:
Adds the ability to mark a page as archived and displays banner on archived pages.

## Additions

- `is_archived` and `archived_at` fields on CFGovPage
- Archived banner template & less
- Conditional to display archived banner in `banner_top` block on `base.html` template

## How to test this PR

1. Check out branch
2. Run migration and gulp styles
3. Open a page in admin, mark it as archived on Configuration tab,  and check that it displays archived banner

### Archived admin interface

Screen Shot 2020-09-17 at 12 29 08 PM

### Archived banner

Screen Shot 2020-09-17 at 12 29 40 PM

## Checklist

#### Browser testing

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Safari on iOS
- [ ] Chrome on Android

#### Accessibility

- [ ] Screen reader friendly

#### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


cfgov/v1/models/base.py
```diff
@@ -110,6 +110,17 @@ class CFGOVPage(Page):
         ),
     )
 
+    is_archived = models.BooleanField(
```
chosak, line 113: Are we supporting cases where there is a displayed archive date but also cases where we don't show the date?

To simplify the changes here, could we have a single field, maybe `archived_at` (to be consistent with Wagtail Page field `_at` [convention](https://github.com/wagtail/wagtail/blob/738003726900295328f263bfe91686026f70f596/wagtail/core/models.py#L694-L702), for example `locked_at`, `first_published_at`, etc)? This could still be checked like `if page.archived_at` if it defaults to null.

chosak (commented):
It'd be nice to add to [our custom page manager](https://github.com/cfpb/consumerfinance.gov/blob/d7aad3c2ee22d3706da96728fdddfb095d0e2185/cfgov/v1/models/base.py#L53) to add a new queryset filter like  `pages.archived()` using a technique like [this](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/jobmanager/models/pages.py#L23-L34), to be able to easily query for archived/non-archived pages.

cfgov/v1/models/base.py
virginiacc: @chosak That is a great question. We've recommended displaying the date a page was archived, but I don't think we've gotten feedback on that specifically so I'm not sure whether that could be subject to change. I can try to find out. In any case, changing this field to `archived_at` to follow Wagtail convention makes sense.

We may also need to add more fields here as there will likely be a need for different archiving text on different pages.

virginiacc: Updated field to `archived_at` in https://github.com/cfpb/consumerfinance.gov/pull/6015/commits/18c4814820c56409838615d646942e2c8afecb56 -- thanks for the suggestion, @chosak!

mistergone (approved):
I was able to pull this down, archive a page, and see the archived banner! 👍

willbarton (approved):
Model changes look good to me!

cfgov/v1/models/base.py
virginiacc: Following up -- my team discussed and we're not sure whether we can rely on the presence of an archived_at date, so we think it would make sense at this point to keep the is_archived field.

PR #6016, opened 2020-09-18
Prevent duplicate highlights in mega menu

chosak:
This change modifies the existing mega menu logic to prevent duplicate highlights in the mega menu, for cases where a menu link is duplicated between submenus.

The code now ensures that only at most one submenu and at most one link gets highlighted.

## How to test this PR

Edit the mega menu to duplicate a link between two submenus. Internally, see UCP PR deployment for this running live with an example.

## Checklist

### DIFF
--- a/cfgov/mega_menu/tests/test_frontend_conversion.py
+++ b/cfgov/mega_menu/tests/test_frontend_conversion.py
@@ -13,8 +13,6 @@ class FrontendConverterTests(TestCase):
     @classmethod
     def setUpTestData(cls):
         cls.root_page = Site.objects.get(is_default_site=True).root_page
-        cls.request = RequestFactory().get('/foo/bar/')
-        Site.find_for_request(cls.request)
 
         consumer_tools_page = cls.make_test_page('Consumer Tools')
         about_us_page = cls.make_test_page('About Us')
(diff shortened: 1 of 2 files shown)


cfgov/mega_menu/tests/test_frontend_conversion.py
```diff
@@ -95,30 +93,113 @@ def make_test_page(cls, title):
...
+                    {
+                        'url': '/other/2/',
+                        'text': 'Second other link',
+                        'icon': 'mail',
+                    },
+                ],
+            },
+        ])
+
+    def test_conversion_output_selection_ignores_query_string(self):
```
schbetsy, line 196: What are your thoughts about adding a test for the specific behavior that's added in this PR? I might be wrong, but this test seems more like #6012

schbetsy (commented):
Works for me 👍 . Had a question about the test.

cfgov/mega_menu/tests/test_frontend_conversion.py
chosak: Thanks, you're totally right. I'll add one.

chosak: @schbetsy I've added a test for this behavior; would you mind re-reviewing when you get a chance?

schbetsy (approved):

PR #6018, opened 2020-09-21
Refactor filterable list pages to be routable and bring the feed/ route in

willbarton:
This PR makes three non-trivial changes to filterable list pages:

1. Refactors the way that `FilterableListMixin` handles querysets to allow them to be overriden. This enables the creation of a `CategoryFilterableMixin` in place of `if`/`else` conditionals that enables searching categories of pages.

2. Makes `FilterableListMixin` inherit from `RoutablePageMixin`. This enables the creation of additional routes within a filterable list page (like `archive/`) that can share some or all of the form, template, queryset, etc, of the filterable list. This should allow us to provide additional views into the same basic filterable content within the same URL context on the site.

3. Refactors our RSS feeds for filterable list pages by making our `FilterableFeedPageMixin` use a route on `FilterableListMixin` instead of its own custom mixin class and route override.

One other note, the `render()` method I've added to `FilterableListMixin` for use in the routes is based on [Wagtail PR #6370](https://github.com/wagtail/wagtail/pull/6370/files#diff-7e25e39c579a36bba2a45eaa2a94154cR122-R164), which adds this method with support for overriding the context. That functionality will be in Wagtail 2.11, and once we upgrade to 2.11, we can remove the overlapping code and replace it with a `super().render()`.

Previously this PR included an added `archive/` route to filterable lists. I've removed that from this PR, and the archive functionality will follow in another PR.

## Checklist


chosak:
This is an interesting approach! I like how it allows the simplification of the RSS feed code. A couple of questions:

- In order for the new FilterableFeedRoutablePageMixin to work, the page class also has to inherit from RoutablePage. With these changes, FilterableListMixin inherits from RoutablePage, but FilterableFeedRoutablePageMixin does not.

   Would there ever be a case where someone might try to use FilterableFeedRoutablePageMixin and not realize they also need to inherit from RoutablePage? Is there a downside to having both mixins inherit?

   Looking at [the code of FilterableFeed](https://github.com/cfpb/consumerfinance.gov/blob/146a9af377e85617210cf2ba57d7bfac1a317361/cfgov/v1/feeds.py#L27) used by FilterableFeedRoutablePageMixin, it seems like it will only work if the page also inherits from FilterableListMixin anyway (plus has a FilterableList block). Is there any advantage maintaining the feed behavior as a separate mixin versus including it in FilterableListMixin? It'd be a nice side effect to automatically have RSS feeds for every filterable list.
- More significantly, does this approach require that the /archive/ page will have the exact same content as its sibling filtered page, with the exception of the items in the list? Is that practical? Won't we need to have different content (potentially entered by Wagtail editors), even fields like the page title and introductory text?

willbarton:
@chosak 

>  Is there any advantage maintaining the feed behavior as a separate mixin versus including it in FilterableListMixin? It'd be a nice side effect to automatically have RSS feeds for every filterable list.

Probably not? I'm certainly open to merging the feed route into `FilterableListMixin` and calling it a day.

> More significantly, does this approach require that the /archive/ page will have the exact same content as its sibling filtered page, with the exception of the items in the list? Is that practical? Won't we need to have different content (potentially entered by Wagtail editors), even fields like the page title and introductory text?

This is an excellent question.

willbarton:
In order to test this PR, you'll need to make some content that a filterable list filters as archived. This field is in the Wagtail admin on the "Configuration" tab under "Archive": 
image

cfgov/v1/models/filterable_list_mixins.py
```diff
@@ -52,29 +48,24 @@ def filterable_pages(self):
...
-            category_names = get_category_children(self.filterable_categories)
-            pages = pages.filter(categories__name__in=category_names)
-
+        queryset = self.get_model_class().objects.in_site(site).live()
         if self.filterable_children_only:
-            pages = pages.child_of(self)
+            queryset = queryset.child_of(self)
 
-        return pages
+        return self.get_model_class().objects.in_site(site).live()
```
chosak, line 55: Is this a mistaken duplication from above? Eliminating the `.child_of(self)` means that, for example, [the Events page](http://localhost:8000/about-us/events/) pulls in every filterable page on the whole site.

```suggestion
        return queryset
```

chosak (dismissed):

cfgov/v1/models/filterable_list_mixins.py
willbarton: Indeed!

willbarton:
I've updated this PR to remove the archive route for now. I'll be in a follow-up that adds it and navigation/linkability.

cfgov/v1/feeds.py
```diff
@@ -75,13 +56,15 @@ def get_appropriate_rss_feed_url_for_page(page, request=None):
     under the index page).
 
     Pages are considered to provide a feed if they inherit from
-    FilterableFeedPageMixin.
+    FilterableFeedRoutablePageMixin.
```
Scotchester, line 59: Should this say `FilterableListMixin`?
```suggestion
    FilterableListMixin.
```

cfgov/v1/models/filterable_list_mixins.py
```diff
@@ -130,12 +125,50 @@ def get_form_data(self, request_dict):
                 self.set_do_not_index(field, value)
         return form_data, has_active_filters
 
-    def serve(self, request, *args, **kwargs):
-        """Modify response headers."""
-        response = super(FilterableListMixin, self).serve(request)
+    def render(self, request, *args, context_overrides=None, **kwargs):
+        """Render with optional for context overrides."""
```
Scotchester, line 129:
```suggestion
        """Render with optional context overrides."""
```

cfgov/v1/models/filterable_list_mixins.py
```diff
@@ -130,12 +125,50 @@ def get_form_data(self, request_dict):
...
+        # Note: the context-overriding and template rendering can be replaced
+        # with super().render() in Wagtail 2.11, where RoutablePageMixin gains
+        # the context_overrides functionality built-in.
```
Scotchester, line 132:
```suggestion
        # TODO: the context-overriding and template rendering can be replaced
        # with super().render() in Wagtail 2.11, where RoutablePageMixin gains
        # the context_overrides functionality built-in.
```

alexm118 (approved):

Scotchester (approved):
A couple trivial comment suggestions, but everything works for me locally as in production.

PR #6019, opened 2020-09-21
Add script for mega menu link checking

chosak:
This new command line script allows for checking of mega menu links on a running cf.gov server. To run against www.consumerfinance.gov:

```
cfgov/scripts/validate_mega_menu.py
```

or to run against an alternate server:

```
cfgov/scripts/validate_mega_menu.py http://localhost:8000
```

The script checks each link in the mega menu (as it appears on the homepage) and validates that the link can be fetched with a 200.

## How to test this PR

To test, run a local server, edit the English menu (http://localhost:8000/admin/mega_menu/menu/edit/en/) and add a broken link. Run this script against localhost and it will fail.

## Checklist

### DIFF
--- a/cfgov/scripts/validate_mega_menu.py
+++ b/cfgov/scripts/validate_mega_menu.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python
+
+import argparse
+import re
+import sys
+
+import requests
+from bs4 import BeautifulSoup
+
+
+def validate_mega_menu(base):
+    failures = []
+
+    print(f'Retrieving mega menu from {base}')
+    response = requests.get(base, allow_redirects=True)
+    response.raise_for_status()
+
+    html = response.content
+    soup = BeautifulSoup(html, 'html.parser')
+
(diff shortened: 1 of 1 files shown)


wpears (approved):
Works as advertised!

One interesting site note is that external links that get fed through the `/external-site/` logic will never trigger a failure, but I don't that's really a concern of ours here.

chosak:
> One interesting site note is that external links that get fed through the /external-site/ logic will never trigger a failure, but I don't that's really a concern of ours here.

Ah, interesting. I didn't think of that. But, yes, I agree. As it happens we don't have any links like that, because the one external link is .gov, which doesn't go through the interstitial. It's probably a safe bet that we wouldn't have any non-.gov links in our site mega menu?

wpears:
>It's probably a safe bet that we wouldn't have any non-.gov links in our site mega menu
Yep, even that one non-cfpb link is kinda surprising to me (rather than a cfpb wagtail page with more easily editable content that itself links to the gpo page). So yeah, :ship:

PR #6021, opened 2020-09-21
Add Elasticsearch DSL to support Elasticsearch 7 for ask_cfpb

cwdavies:
consumerfinance.gov supports elasticsearch 2.4.1 for ccdb5_api and legacy haystack apps

## Additions

- consumerfinance.gov supports elasticsearch7 for ask_cfpb application

## Changes

- elasticsearch versions 2.4 and 7.9 are supported

## Modifications

- boolean feature flag added to switch between elasticsearch2 and elasticsearch7 code 

## How to test this PR

1. Confirm versions of `elasticsearch` are supported using
```
$ pip list|grep elastic
django-elasticsearch-dsl 7.1.4
elasticsearch            2.4.1
elasticsearch-dsl        7.2.2
elasticsearch7           7.9.1
```

2. Enable `ELASTICSEARCH_DSL_ASK` flag

3. Verify the ask-cfpb index runs, and elasticsearch7 is working
`manage.py update_index ask_cfpb`

4. Confirm the haystack indexing runs as before using elasticsearch2:
`cfgov/manage.py  search_index --rebuild -f --parallel`

5. Search Ask

## Checklist

#### Browser testing

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Safari on iOS
- [ ] Chrome on Android

#### Accessibility

- [ ] Screen reader friendly

#### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


cfgov/ask_cfpb/documents.py
```diff
@@ -0,0 +1,508 @@
...
+
+    # Overrides the default of page.id for comparing against split testing
+    # clusters. See: core.feature_flags.in_split_testing_cluster
+    @property
+    def split_test_id(self):
+        return self.answer_base.id
+
+
+@registry.register_document
+class AnswerPageDocument(Document):
```
willbarton, line 396: I'm still a bit wary of introducing a new meaning to an already overloaded term, "Document". What do folks think about being super-explicit about this "Document" belonging to Elasticsearch by naming this class (and doing the same all the similar classes we need for other searches) `AnswerPageElasticsearchDocument`? 

Maybe even `AnswerPageSearchDocument`, but to make sure the class name is clear about its purpose.

cfgov/ask_cfpb/documents.py
```diff
@@ -0,0 +1,508 @@
...
+        'text_suggestion', search_term, term={'field': 'text'})
+    response = s.execute()
+    try:
+        suggested_term = response.suggest.text_suggestion[0].options[0].text
+        return suggested_term
+    except IndexError:
+        return search_term
+
+
+class AnswerPageSearch:
```
willbarton, line 449: I also wonder whether this class could be placed in the `cfgov/search` package and take a model in its constructor. That'd give us a single place where we implement autocomplete, search, suggestions, etc, that would be reusable across all our apps.

@alexm118 @higs4281 do you think that would make sense? I'm basing the idea on a reading of how this works, and I'm not totally confident that reading is accurate.

willbarton (changes requested):
One absolutely major thing: we cannot remove Haystack from Ask CFPB yet. We need to use a feature flag or something like to switch the codepath. We do not have Elasticsearch 7 available yet in production!! Presumably this could involve just feature flagging `ask_search_es7`, et al, in `urls.py` with a fallback on `ask_search`.

The other thing is, we had talked about the `documents.py` file and how it would be nice to ensure that the Ask CFPB models all still live in `ask_cfpb/models`, and that the solution to the circular imports we had was to simply break up `ask_cfpb/models/pages.py` so that relevant models are in their own files.

willbarton:
I also think, before we get to the point of merging, that it'd be worth squashing down some of the commits on this branch.

cfgov/ask_cfpb/documents.py
cwdavies: Agree with renaming `AnswerPageDocument` class. This class is now called `AnswerPageSearchDocument`.

cfgov/ask_cfpb/views.py
```diff
@@ -93,12 +95,12 @@ def ask_search(request, language='en', as_json=False):
         results_page.result_query = ''
         return results_page.serve(request)
 
-    search = AskSearch(search_term=search_term, language=language)
-    if search.queryset.count() == 0:
+    search = AnswerPageSearch(search_term, language=language)
```
alexm118, line 98: We shouldn't make this change yet, as this is where we interact with ES 2.4

We should try and leave this code as is and implement an ES7 version that can be invoked within a flag check

cfgov/ask_cfpb/views.py
```diff
@@ -109,37 +111,100 @@ def ask_search(request, language='en', as_json=False):
...
+    results_page.result_query = response.get('search_term')
+    results_page.suggestion = response.get('suggestion')
+    results_page.answers = [
+        (result.url, result.autocomplete, result.preview)
+        for result in response['results']
+    ]
+    return results_page.serve(request)
+
 
+# def ask_autocomplete(request, language='en'):
```
alexm118, line 181: This is the ES 2.4 implementation, should remain functional and trigger the es7 version when a flag is checked.

cfgov/ask_cfpb/documents.py
```diff
@@ -0,0 +1,508 @@
+from django import forms
```
alexm118, line 1: we can move this file into `cfgov/ask_cfpb/models/search/documents.py`

cfgov/ask_cfpb/documents.py
```diff
@@ -0,0 +1,508 @@
...
+
+    # Overrides the default of page.id for comparing against split testing
+    # clusters. See: core.feature_flags.in_split_testing_cluster
+    @property
+    def split_test_id(self):
+        return self.answer_base.id
+
+
+@registry.register_document
+class AnswerPageSearchDocument(Document):
```
alexm118, line 396: Per the discussion in standup - we may want to avoid putting Search into Indexed classes, as it could muddy the waters between the `AnswerPageSearch` class which actually does the searching, where as this class is representing the ElasticSearch Document.

cfgov/ask_cfpb/documents.py
alexm118: I think we may want to generalize some of this code - like the analyzers into a common file to be referenced across multiple search areas.

cwdavies:
> One absolutely major thing: we cannot remove Haystack from Ask CFPB yet. We need to use a feature flag or something like to switch the codepath. We do not have Elasticsearch 7 available yet in production!! Presumably this could involve just feature flagging `ask_search_es7`, et al, in `urls.py` with a fallback on `ask_search`.
> 

Created a boolean flag called `ELASTIC_SEARCH_DSL` set to False by default in `cfgov/settings/base.py'
Added support for `ELASTIC_SEARCH_DSL` flag in `cfgov/urls.py` using `flagged_re_path`
for `ask_search_es7` and `ask_autocomplete_es7`

cfgov/ask_cfpb/documents.py
cwdavies: Done

cfgov/ask_cfpb/views.py
cwdavies: Done

cfgov/ask_cfpb/views.py
cwdavies: Done

cfgov/ask_cfpb/models/search/documents.py
```diff
@@ -0,0 +1,508 @@
...
+        pass
+
+
+def get_standard_text(language, text_type):
+    return get_reusable_text_snippet(
+        REUSABLE_TEXT_TITLES[text_type][language]
+    )
+
+
+class AnswerPage(CFGOVPage):
```
willbarton, line 153: @alexm118 does the `AnswerPage` model need to live with the search document, or does it just need to live in a separate model file in `ask_cfpb.models`?

cfgov/ask_cfpb/views.py
```diff
@@ -93,12 +97,17 @@ def ask_search(request, language='en', as_json=False):
         results_page.result_query = ''
         return results_page.serve(request)
 
-    search = AskSearch(search_term=search_term, language=language)
-    if search.queryset.count() == 0:
-        search.suggest(request=request)
+    if flag_enabled('ELASTIC_SEARCH_DSL'):
```
willbarton, line 100: It looks like you're using the flag in `urls.py` to switch view functions based on it. I don't believe this is doing anything, and it would probably be cleaner to use `urls.py` to switch whole views instead of checking the flag state multiple times within this view.

cfgov/cfgov/urls.py
```diff
@@ -455,23 +446,41 @@ def empty_200_response(request, *args, **kwargs):
...
         name='ask-spanish-answer'
     ),
-    re_path(
+
+    # If 'ELASTIC_SEARCH_DSL' is True, use elasticsearch7.
+    flagged_re_path(
+        'ELASTIC_SEARCH_DSL',
+        r'^(?P<language>es)/obtener-respuestas/buscar/$',
+        ask_search_es7,
+        name='ask-search-es'
```
willbarton, line 455: You'll need to provide the original view as a [`fallback` for these flagged urls](https://cfpb.github.io/django-flags/api/urls/#flagged_re_pathflag_name-route-view-kwargsnone-namenone-statetrue-fallbacknone). Something like:

```python
    flagged_re_path(
        'ELASTIC_SEARCH_DSL',
        r'^(?P<language>es)/obtener-respuestas/buscar/$',
        ask_search_es7,
        fallback=ask_search,
        name='ask-search-es'
    ),
```

cfgov/cfgov/urls.py
```diff
@@ -542,6 +551,39 @@ def empty_200_response(request, *args, **kwargs):
...
+        r'^ask-cfpb/api/autocomplete/$',
+        ask_autocomplete, name='ask-autocomplete-en'
+    ),
+    re_path(
+        r'^(?P<language>es)/obtener-respuestas/api/autocomplete/$',
+        ask_autocomplete, name='ask-autocomplete-es'
+    ),
+]
+urlpatterns = urlpatterns + ask_urlpatterns
+
```
willbarton, line 586: I don't think I understand the point of adding these URLs here. It overrides the flagged URLs you've got above.

cfgov/search/backends.py
```diff
@@ -1,10 +1,16 @@
...
+import haystack.backends.elasticsearch2_backend
 from django.conf import settings
 from haystack.backends.elasticsearch2_backend import (
     Elasticsearch2SearchBackend, Elasticsearch2SearchEngine
 )
 
+import elasticsearch
+
+
+haystack.backends.elasticsearch2_backend.elasticsearch = elasticsearch
```
willbarton, line 12: What is this attempting to do?

requirements/libraries.txt
```diff
@@ -21,6 +22,7 @@ django-storages==1.7.1
 django-treebeard==4.2.0
 django-watchman==0.15.0
 edgegrid-python==1.0.10
+elasticsearch7
```
willbarton, line 25: We should pin a specific version here.

cfgov/ask_cfpb/models/search/documents.py
```diff
@@ -0,0 +1,508 @@
...
+
+    # Overrides the default of page.id for comparing against split testing
+    # clusters. See: core.feature_flags.in_split_testing_cluster
+    @property
+    def split_test_id(self):
+        return self.answer_base.id
+
+
+@registry.register_document
+class AnswerPageSearchDocument(Document):
```
willbarton, line 396: I think @alexm118 had made the case that if this class is in `ask_cfpb.models.search`, we should be fine with calling it `AnswerPageDocument`, because the context is clearer.

cfgov/ask_cfpb/models/search/documents.py
```diff
@@ -0,0 +1,508 @@
...
+        'text_suggestion', search_term, term={'field': 'text'})
+    response = s.execute()
+    try:
+        suggested_term = response.suggest.text_suggestion[0].options[0].text
+        return suggested_term
+    except IndexError:
+        return search_term
+
+
+class AnswerPageSearch:
```
willbarton, line 449: I'm not sure that the `models` package is the right place for this to live — it's not a data model. 

I also agree with @alexm118, it would be nice to generalize this class and put it in `cfgov.search`. It looks like the only thing we'd have to do is maybe add a `document_class` argument to `__init__` (or use `base_query` and require it?) and use that instead of using `AnswerPageSearchDocument` directly. That way any of our searches could pass a `Document` subclass and get autocomplete, search, and suggestions without reimplementing any of this.

cfgov/ask_cfpb/models/search/documents.py
alexm118: Yeah, I doesn't need to live within documents.py it just cannot be in `pages.py` or `views.py`. 

I didn't think about it at the time, but potentially we may want to actually move the `PortalSearchPage` out of pages.py into something that more aligns with what its doing? Just a thought, as it seeems to be the offending class that actually has more logic than a standard wagtail page embedded within it.

cfgov/ask_cfpb/models/search/documents.py
alexm118: Agreed, I think we can make this a more generalized search class, as I believe the logic for implementing the autocomplete, search, and suggest features will be largely the same with the only outlier being the field within the Document we act upon. I'm not sure fully how that will look but I think giving it a shot and evaluating its utility over time could prove worthwhile.

requirements/libraries.txt
cwdavies: Done

cfgov/cfgov/urls.py
cwdavies: Done

cfgov/cfgov/urls.py
cwdavies: Removed after adding fallback to flagged_re_path

cfgov/ask_cfpb/views.py
cwdavies: Removed flag_enabled from ask_cfpb views.py

cfgov/ask_cfpb/models/search/documents.py
cwdavies: Renamed AnswerPageSearchDocument to AnswerPageDocument

cfgov/ask_cfpb/models/search/documents.py
willbarton: That'd be fine too — there's nothing that says all the Wagtail pages have to live in a `models/pages.py`. I think that's a convention we accidentally fell into.

I actually think the fact that we *don't* split up our page models more is part of how we end up with circular dependencies.

cfgov/search/backends.py
cwdavies: Not sure what was done here and why so I removed them.

cfgov/ask_cfpb/documents.py
cwdavies: Removed Search from Indexed classes. Renamed AnswerPageSearchDocument to AnswerPageDocument.

cfgov/ask_cfpb/views.py
cwdavies: Removed flag_enabled from ask_cfpb views.py as flag is already being used in urly.py.

cfgov/cfgov/urls.py
cwdavies: Added fallback to flagged_re_path.

cfgov/cfgov/urls.py
cwdavies: Removed as I added fallback to flagged_re_path

cfgov/ask_cfpb/models/search/documents.py
cwdavies: Renamed class to AnswerPageDocument

requirements/libraries.txt
cwdavies: A specific version of elasticsearch7 is now pinned.

cfgov/search/backends.py
cwdavies: I don't know why this was done so I removed the duplicate import of elasticsearch.

cfgov/ask_cfpb/models/search/documents.py
cwdavies: Moved AnswerPageSearch class from ask_cfpb models into cfgov.search

cwdavies: @willbarton Please expand on how to implement your suggestion as to which specific class you would like to add a `document_class` argument to `__init__`. Currently the search only supports ask_cfpb. I like the idea of making the search generic and support passing a `Document` subclass.

cfgov/ask_cfpb/models/search/documents.py
cwdavies: Moved AnswerPage class from search documents into ask_cfpb.models.

cwdavies: @alexm118 Perhaps moving `PortalSearchPage` out of pages.py could be done in a different PR.

willbarton (commented):
I realize we've moved a lot of files around, but I don't think there's a reason to ask_cfpb/models/search.py to ask_cfpb/models/ask_search.py anymore.

Another thing is the PR still seems to be removing Haystack functionality in places, instead of switching on the flag, unless I'm missing something. I'll pull it down and try it out — but there are a lot of places where we're *changing* unittests instead of adding parallel ones that test the flagged codepath. I'd expect the unittests to test both codepaths, so we'd just want to *add* new tests for our new code, that tests with `@override_settings(FLAGS={"ELASTIC_SEARCH_DSL": [("boolean", True)]})`.

cfgov/cfgov/settings/local.py
```diff
@@ -23,8 +23,16 @@
 INSTALLED_APPS += (
     "sslserver",
     "wagtail.contrib.styleguide",
+    "django_elasticsearch_dsl",
 )
 
+ELASTICSEARCH_DSL={
+    'default': {
+        'hosts': 'http://localhost:9200'
```
willbarton, line 31: Could we make this dependent on an environment variable and add that variable to `.env_sample`? I have to change it to get it to work locally in Docker.

cfgov/search/documents.py
```diff
@@ -0,0 +1,84 @@
...
+]
+
+
+def make_safe(term):
+    for char in UNSAFE_CHARACTERS:
+        term = term.replace(char, '')
+    return term
+
+
+class AnswerPageSearch:
```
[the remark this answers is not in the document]
cwdavies: Done

cfgov/ask_cfpb/search_indexes.py
```diff
@@ -1,77 +0,0 @@
...
-        self.prepared_data['suggestions'] = self.prepared_data['text']
-        return self.prepared_data
-
-    def get_model(self):
-        from ask_cfpb.models import AnswerPage
-        return AnswerPage
-
-    def index_queryset(self, using=None):
-        return self.get_model().objects.filter(
-            live=True, redirect_to_page=None)
```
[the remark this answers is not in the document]
cwdavies: Restored file

cfgov/ask_cfpb/tests/models/test_pages.py
```diff
@@ -170,9 +183,9 @@ def create_page(model, title, slug, parent, language="en", **kwargs):
             return new_page
 
         self.test_user = User.objects.last()
-        self.ROOT_PAGE = HomePage.objects.get(slug="cfgov")
+        self.root_page = HomePage.objects.get(slug="cfgov")
```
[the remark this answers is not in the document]
cwdavies: I do not know why these changes were made.

cfgov/ask_cfpb/models/pages.py
```diff
@@ -93,7 +98,7 @@ def get_portal_or_portal_search_page(portal_topic, language='en'):
 
 
 def get_ask_breadcrumbs(language='en', portal_topic=None):
-    DEFAULT_CRUMBS = {
+    default_crumbs = {
```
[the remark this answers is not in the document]
cwdavies: I do not know why this change was made.

cfgov/ask_cfpb/models/pages.py
```diff
@@ -393,11 +423,10 @@ def portal_category_page(self, request, **kwargs):
...
                 context)
-        self.query_base = SearchQuerySet().filter(
-            portal_topics=self.portal_topic.heading,
-            language=self.language,
-            portal_categories=self.portal_category.heading)
-        return self.get_results(request)
+        self.query_base = AnswerPageDocument.search().filter(
+            'match', portal_topics=self.portal_topic.heading).filter(
+            'match', portal_categories=self.portal_category.heading)
+        return self.get_results_es7(request)
```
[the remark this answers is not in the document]
cwdavies: Agreed and Done

cfgov/ask_cfpb/models/pages.py
```diff
@@ -369,18 +400,17 @@ def get_glossary_terms(self):
...
     @route(r'^$')
     def portal_topic_page(self, request):
-        self.query_base = SearchQuerySet().filter(
-            portal_topics=self.portal_topic.heading,
-            language=self.language)
         self.portal_category = None
-        return self.get_results(request)
+        self.query_base = AnswerPageDocument.search().filter(
+            "match", portal_topics=self.portal_topic.heading)
+        return self.get_results_es7(request)
```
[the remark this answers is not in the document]
cwdavies: Agreed and Done.

cfgov/ask_cfpb/tests/test_views.py
```diff
@@ -38,14 +71,14 @@ def setUp(self):
         from ask_cfpb.models import Answer
         from v1.models import HomePage
 
-        self.ROOT_PAGE = HomePage.objects.get(slug="cfgov")
+        self.root_page = HomePage.objects.get(slug="cfgov")
```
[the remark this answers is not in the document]
cwdavies: I do not know why these changes were done.

cfgov/ask_cfpb/documents.py
cwdavies: Moved AnswerPageSearch into cfgov/search package. Code generalization could be done in another PR when other applications start to use ES7.

cfgov/cfgov/settings/local.py
cwdavies: Added ES7_HOST variable to `.env_sample` and used that ES7 variable in `ELASTICSEARCH_DSL`

cfgov/ask_cfpb/migrations/0043_auto_20200928_1510.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('ask_cfpb', '0042_share_and_print_help'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='answerpage',
+            name='share_and_print',
+            field=models.BooleanField(default=False, help_text='Include share and print buttons above page content.'),
```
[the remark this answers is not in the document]
cwdavies: Resolved migration issue by updating share_and_print help_text in AnswerPage class

cfgov/ask_cfpb/models/pages.py
```diff
@@ -64,6 +57,18 @@ def get_standard_text(language, text_type):
 )
 
 
+def strip_html(markup):
+    """Make sure stripping doesn't mash headings into text."""
+    markup = re.sub("</h[1-6]>", " ", markup)
+    return strip_tags(markup)
+
+
+def get_standard_text(language, text_type):
```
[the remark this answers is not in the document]
cwdavies: I do not know why this function was moved.

cfgov/ask_cfpb/tests/models/test_pages.py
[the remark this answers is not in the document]
willbarton: It seems to have returned.

willbarton (commented):
Excellent work getting this cleaned up. Running locally, searches seem to behave as expected with the feature flag both turned off (using Haystack) and turned on (using dj-es-dsl).

I think if we can clean up some of the superfluous changes, and add the missing test coverage (glancing at the uncovered line numbers and the lines themselves, these all look like things that should be covered), I think we'll be good!

willbarton:
One thing that occurs to me is that it would probably be good to add some functional tests around search [that are a little more robust](https://github.com/cfpb/consumerfinance.gov/blob/main/test/cypress/integration/pages/ask-cfpb.js) so that we can run them against both Haystack-backed search and dj-es-dsl-backed search, instead of doing this manually.

It'll also make our work a lot easier when it comes to figuring our what we can abstract off for other apps, because we'll have tests that will let us know if we're breaking Ask search.

willbarton:
I've tried to create a PR that does just that: https://github.com/cfpb/consumerfinance.gov/pull/6055

willbarton (approved):
I think this all looks good! I have one trivial comment suggestion for the feature flag.

When we get to the point of merging this, **please** remember to do a squash merge for this PR, because of the number of WIP commits that it brings in.

cfgov/cfgov/settings/base.py
```diff
@@ -767,8 +769,8 @@
     # Q42020 Ask CFPB customer satisfaction survey.
     "ASK_SURVEY_INTERCEPT": [],
     # Used to enable use of django-elasticsearch-dsl and disable use of Haystack
-    # This will be used in the ask_cfpb and regulations applications
-    "ELASTIC_SEARCH_DSL": [("boolean", False)],
+    # These will be used in the ask_cfpb and regulations applications
```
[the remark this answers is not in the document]
cwdavies: Done. I like this suggested change.

PR #6022, opened 2020-09-22
Removes node_modules check in gulp subapp script

wpears:
Fixes broken js in prod for youth-employment-success and form-explainer

This check was failing because some subapps have
no dependencies (this makes sense!) There are
other workarounds/different checks that could be
written, but I think this check is mostly
unnecessary as subapp dependencies are automatically
installed in postinstall.


Scotchester:
This is interesting. Some stream of consciousness observations as I review/investigate:

- The `scriptsApps` script being edited in this PR hasn't been modified in 3 years. Why break now? (I assume that the youth transportation tool hasn't _always_ been broken and someone tested it successfully in prod at least once, but…)
- If I look locally, I have `node_modules` folders in those two apps with a `.yarn-integrity` file in them.
- I think maybe that leads to answering the question of "Why break now?" – my changes in #5959 make it so that `yarn` is not even invoked if there's a `package.json` with no dependencies.
- Wiping out those two `node_modules` folders and re-running `frontend.sh` locally, they do not come back.
- So the changes in this PR make sense!

Scotchester (approved):

anselmbradford:
This comment should be removed too https://github.com/cfpb/consumerfinance.gov/blame/main/gulp/tasks/scripts.js#L174-L175

anselmbradford:
This comment should be removed too https://github.com/cfpb/consumerfinance.gov/blob/main/gulp/tasks/scripts.js#L174-L175

PR #6023, opened 2020-09-22
Add script for validating redirects

chosak:
This new script allows for validation of a set of redirects as defined in a CSV or XLSX file. The first line of the file will be ignored as headers. The "from" and "to" paths default to being read from the 6th and 7th column of the input file, respectively.

To run against www.consumerfinance.gov, using an input XLSX:

```
cfgov/scripts/validate_redirects.py redirects.xlsx
```

To run against localhost, using an input CSV:

```
cfgov/scripts/validate_redirects.py redirects.csv http://localhost:8000
```

To specify alternate CSV columns for the "from" and "to" paths:

```
cfgov/scripts/validate_redirects.py redirects.xlsx --from-index=1 --to-index=2
```

## How to test this PR

To test, run without one of the redirects and observe that the check fails. Now add the redirect and note that the check passes as long as it redirects to the correct path.

## Checklist

### DIFF
--- a/cfgov/scripts/validate_redirects.py
+++ b/cfgov/scripts/validate_redirects.py
@@ -7,20 +7,23 @@ import requests
 from tablib import Dataset
 
 
-def load_redirects(redirects_filename):
+def load_redirects(redirects_filename, from_index, to_index):
     csv = redirects_filename.endswith('.csv')
     mode = 'rt' if csv else 'rb'
 
     with open(redirects_filename, mode) as f:
         dataset = Dataset().load(f, format='csv' if csv else None)
 
-    return [tuple(row[i].rstrip('*') for i in (5, 6)) for row in dataset]
+    return [
+        tuple(row[i].rstrip('*') for i in (from_index, to_index))
+        for row in dataset
+    ]
 
 
-def validate_redirects(redirects_filename, baseurl):
+def validate_redirects(redirects, baseurl):
     failures = []
 
-    for from_path, to_path in load_redirects(redirects_filename):
+    for from_path, to_path in redirects:
         print(f'Checking {from_path} -> {to_path}...', end='')
         response = requests.get(baseurl + from_path, allow_redirects=False)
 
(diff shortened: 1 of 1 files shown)


cfgov/scripts/validate_redirects.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def load_redirects(redirects_filename):
+    csv = redirects_filename.endswith('.csv')
+    mode = 'rt' if csv else 'rb'
+
+    with open(redirects_filename, mode) as f:
+        dataset = Dataset().load(f, format='csv' if csv else None)
+
+    return [tuple(row[i].rstrip('*') for i in (5, 6)) for row in dataset]
```
wpears, line 17: Could we maybe make 5 and 6 default values to arguments that are overwrite-able from the command line, ie `from_index=5, to_index=6` or some such? This seems to be keyed for a specifically formatted excel file but, if we need to rerun this in the future, it would be nice to not have to make this (albeit small) code change if the excel/csv format changes.

cfgov/scripts/validate_redirects.py
```diff
@@ -0,0 +1,66 @@
...
+        'redirects_filename',
+        help='Redirects filename (XLSX format)'
+    )
+    parser.add_argument(
+        'baseurl',
+        help='Base URL, defaults to %(default)s',
+        default='https://www.consumerfinance.gov',
+        nargs='?'
+    )
+
```
wpears, line 64: With the comment above, let's add arguments for the from and to indexes

wpears (changes requested):
Looks/works great overall, just a note about making this script more reusable.

cfgov/scripts/validate_redirects.py
chosak: Excellent suggestion as always. Will do.

chosak: @wpears just added a commit that adds optional `--from-index` and `--to-index` parameters, for review at your convenience.

wpears (approved):
🥇 love it

PR #6031, opened 2020-09-24
Remove jquery from base_update template

anselmbradford:
## Removals

- Removes jquery from base_update template. 

## How to test this PR

1. This template appears to only be used on:
http://localhost:8000/data-research/consumer-complaints/
https://www.consumerfinance.gov/paying-for-college2/understanding-your-financial-aid-offer/offer/test/


wpears (changes requested):
The `nemo_js` appears to also be used on:
`cfgov/paying_for_college/templates/paying_for_college/disclosure_technote.html`
(though doesn't look like any jquery is actually invoked there)
and there's a reference to the `nemo_js` block on:
`cfgov/legacy/templates/students/helping-borrowers-find-ways-to-stay-afloat/index.html`
but the parent/child relationship is less clear to me on that page.

anselmbradford:
Thanks @wpears !

> The nemo_js appears to also be used on:
cfgov/paying_for_college/templates/paying_for_college/disclosure_technote.html
(though doesn't look like any jquery is actually invoked there)

Removed in https://github.com/cfpb/consumerfinance.gov/pull/6031/commits/38f9beeea5ce0a91b9482b3fa624db692fa47ae8

> and there's a reference to the nemo_js block on:
cfgov/legacy/templates/students/helping-borrowers-find-ways-to-stay-afloat/index.html
but the parent/child relationship is less clear to me on that page.

This uses a different template and is removed in https://github.com/cfpb/consumerfinance.gov/pull/6030

wpears (approved):
:+1:

cfgov/legacy/templates/front/base_update.html
```diff
@@ -211,16 +204,6 @@
...
-<script type="text/javascript">
-    // If there are no breadcrumbs on this page, remove the ul tag so that screen
-    // readers don't read it (since the ul has nothing in it).
-    $(function(){
-        var numBreadCrumbs = $( 'ul.bread' ).children( 'li' ).length;
-        if( numBreadCrumbs === 0 ) {
-           $( 'ul.bread' ).remove();
-        }
-    });
-</script>
```
Scotchester, line 223: You can also remove the `<ul class="bread meta">` from this template, as nothing that extends this template puts anything in `{% block breadcrumbs %}`.

Scotchester (approved):
One small change that could be thrown in, but otherwise looks good.

PR #6039, opened 2020-09-25
Add child, sibling, and archive controls to filterable lists

willbarton:
This change moves the "filter_children" property off the `FilterableListMixin` and places it as a boolean control in the `FilterableListBlock` organism. It also adds corresponding `filter_siblings` and `filter_archive` boolean controls.

This will enable the content editor in the CMS to configure filterable list page behavior to filter either children or siblings, and/or whether to filter archived content or not.

This will allow for archive filterable list pages to live inside parent filterable list pages and filter sibling pages. For example, the "Blog" filterable list page, set to filter child pages, can now have a corresponding "Blog Archive" beneath it, set to filter siblings and archived pages, and one can create any number of content combinations across either page to link back and forth between them.

This PR also includes a migration to add the controls and a data migrations to ensure that the existing `filter_children` behavior of `NewsroomLandingPage` and `ActivityLogPage` is maintained.

The field labels and help text in the admin can be iterated on as needed, but this PR will get the desired functionality in for the 9/30 deadline.

## Screenshots

image

## Checklist


alexm118 (approved):

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -792,6 +792,29 @@ class FilterableList(BaseExpandable):
...
+        required=False,
+        help_text=(
+            "If checked this list will only filter its sibling pages."
+        ),
+    )
+    filter_archive = blocks.BooleanBlock(
+        default=False,
+        required=False,
+        help_text=(
+            "If checked this list will only filter archived pages."
```
chosak, line 813:
```suggestion
            "If checked this list will only filter archived pages. "
```

virginiacc (approved):
This works perfectly -- thank you, @willbarton!

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -792,6 +792,29 @@ class FilterableList(BaseExpandable):
...
+            "If checked this list will only filter its child pages."
+        ),
+    )
+    filter_siblings = blocks.BooleanBlock(
+        default=False,
+        required=False,
+        help_text=(
+            "If checked this list will only filter its sibling pages."
+        ),
+    )
```
Scotchester, line 808: These both being booleans with the help text as-is opens the possibility for confusion about what happens if both are checked. The logic in the model checks `filter_children` and then only if that is `False` does it check `filter_siblings`, but that's not evident in the UI created by these block definitions.
1. Should we allow for the possibility of both being true, i.e. a list showing both siblings and children? Grandchildren or further descendants? Nieces and nephews? :trollface: 
2. If not, could we make this a ChoiceBlock defaulting to children?

cfgov/v1/migrations/0232_set_filter_children_on_newsroom_activitylog.py
```diff
@@ -0,0 +1,38 @@
...
+def forward_mapper(page_or_revision, data):
+    # Manipulate the stream block data forwards
+    data['filter_children'] = False
+    return data
+
+
+def forwards(apps, schema_editor):
+    page_types_and_fields = [
+        ('v1', 'ActivityLogPage', 'content', 'filter_controls'),
+        ('v1', 'NewsroomLandingPage', 'content', 'filter_controls'),
```
Scotchester, line 17: Newsroom no longer includes Blog entries, so I think this is incorrect? We do want it to filter its children.
```suggestion
```

cfgov/v1/models/browse_filterable_page.py
```diff
@@ -107,6 +107,5 @@ def get_form_class():
 class NewsroomLandingPage(CategoryFilterableMixin, BrowseFilterablePage):
     template = 'newsroom/index.html'
     filterable_categories = ['Newsroom']
```
Scotchester, line 109: And this should also be removed, right?
```suggestion
```

Scotchester (commented):
I don't totally understand the alchemy behind why the Activity Log works, but it does. Not approving quite yet because I have a couple significant questions.

cfgov/v1/migrations/0232_set_filter_children_on_newsroom_activitylog.py
chosak: @Scotchester In the current code, `NewsroomLandingPage.filterable_children_only = False`, meaning that Newsroom filters any pages that meet its (category) criteria, not just its children. Because this PR defaults `FilterableList.filter_children = True` (filter only children by default), I think we do need to set this line to maintain the current behavior.

I believe Newsroom includes any pages in [one of the newsroom categories](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/v1/util/ref.py#L100-L105), using the logic [you ask about down below](https://github.com/cfpb/consumerfinance.gov/pull/6039/files#r495289839).

Are you suggesting that these two things happen to be equivalent in the current site:

- Filterable children of Newsroom
- Filterable pages in the newsroom categories, anywhere on the site

I did a query and there appear to be 6 live filterable pages marked with newsroom categories that don't live under the Newsroom:

- [Truth in Savings Act (TISA) examination procedures](https://www.consumerfinance.gov/policy-compliance/guidance/supervision-examinations/truth-in-savings-act-tisa-examination-procedures/)
- [Fair Debt Collection Practices Act (FDCPA) examination procedures](https://www.consumerfinance.gov/policy-compliance/guidance/supervision-examinations/fair-debt-collection-practices-act-fdcpa-examination-procedures/ )
- [Fair Credit Reporting Act (FCRA) examination procedures](https://www.consumerfinance.gov/policy-compliance/guidance/supervision-examinations/fair-credit-reporting-act-fcra-examination-procedures/)
- [Consumer Leasing Act (CLA) examination procedures](https://www.consumerfinance.gov/policy-compliance/guidance/supervision-examinations/consumer-leasing-act-cla-examination-procedures/)
- [Student Debt Domino Effect?](http://www.consumerfinance.gov:8000/about-us/blog/student-debt-domino-effect/)
- [Help for Military Homeowners](https://www.consumerfinance.gov/about-us/blog/help-for-military-homeowners/)

You can confirm this by e.g. searching the newsroom for ["domino"](https://www.consumerfinance.gov/about-us/newsroom/?title=domino&from_date=&to_date=) and seeing that a blog page is returned. Whether or not this is intended behavior I don't know, but it is current behavior.

cfgov/v1/models/browse_filterable_page.py
chosak: I don't think so, see comment above.

cfgov/v1/migrations/0232_set_filter_children_on_newsroom_activitylog.py
Scotchester: Hmm… I see. I think it is probably _not_ intended behavior, but I guess better safe than sorry, so let's leave it that way for now.

cfgov/v1/atomic_elements/organisms.py
willbarton: You're right in that the behavior of the booleans is non-obvious in the way they interact. I don't think there's much to do to help that right now – a `ChoiceBlock` might work, but I think it'd add some additional complexity in the implementation that might not be worth it (more on that in a moment). As a general thing, the way archival appears in the admin is already going to be confusing and non-obvious until we actually spend time designing that experience. 

For filterable lists themselves, I think a better way than any of this would be to expose the parent page to filter as a page selector, or, if no page is selected, allow category selection in the admin. But that has a lot more implications that we'd need to handle, and potentially more pages that could break.

willbarton:
I will also note, that the reason the filter boolean fields are on the filterable list block itself, rather than the page models, is the result of consulting with our UXers. @niqjohnson rightly points out that the filterable list block is where we already include some configuration for how the list of filtered pages renders, and so it makes sense to include these fields in that context.

PR #6041, opened 2020-09-25
Add temporary script to archive content older than 2 years

willbarton:
This change adds a script that will take a filterable list page URL path as its argument and then archive the pages within that filterable list queryset if the page's `last_published_at` date is more than two years from when the script runs.

This is based on the scripts for the IA Refresh by @schbetsy in #6026 

The idea is that this script can be run on the Newsroom, Blog, and Research and Reports, like so:

```
./cfgov/manage.py runscript bulk_archive_content --script-args about-us/blog
./cfgov/manage.py runscript bulk_archive_content --script-args about-us/newsroom
./cfgov/manage.py runscript bulk_archive_content --script-args data-research/research-reports
```

## Checklist


willbarton:
There's a minor issue with blogs at the moment, there's a blog page that throws the following error on save:

```
django.core.exceptions.ValidationError: {'__all__': ['Social sharing image must be less than 4096w x 4096h']}
```

I'm looking into this.

alexm118 (approved):
lgtm

Scotchester (commented):
I understand we're in a bit of a crunch here and this was an expedient solution, but presumably we'll need to do bulk archiving again in the future, unless we set it up to be done on a rolling basis. Could we consider making this a management command with a parameter for how far back in time you want to go?

chosak:
This error happens on [this page](https://www.consumerfinance.gov/about-us/blog/guard-your-new-medicare-id-card-avoid-fraud/), which uses [this](https://files.consumerfinance.gov/f/images/bcfp_medicare-id-fraud_6250x2958.original.jpg) overly large 6250x2958 image as its social sharing image.

We can bypass these checks by manually updating only the model fields that we're trying to change, and do so in a significantly more efficient manner by using [`update()`](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#update) directly on the queryset, instead of instance by instance. We can also use [`select_for_update()`](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#select-for-update) to avoid any potential conflicting transactions.

I've pushed a commit that makes this change.

chosak (approved):
> Could we consider making this a management command with a parameter for how far back in time you want to go?

@Scotchester I agree this would be a nice thing to do (plus the addition of tests) but propose moving forward with this as-is due to launch schedule and then revisiting when possible.

Scotchester:
Thanks, @chosak! I'll confirm that your fix works for the bad image and then merge

Scotchester (approved):
Works as expected with Andy's fix. (I also reduced the size of the image on the live site, for good measure.)

PR #6042, opened 2020-09-25
Fix validate_redirects.py to handle absolute paths

chosak:
The current redirects validation script (validate_redirects.py) assumes  that the server responds with redirects pointing to relative paths, like Django returns. Apache instead returns absolute paths, which fails the script. This change makes it so that the script properly handles either case.

## How to test this PR

You can test this using an input spreadsheet that includes an Apache-served redirect; internally this can be demonstrated now on UCP stack 6024.

## Checklist


wpears (approved):
This code is very sneaky

PR #6043, opened 2020-09-26
Make mega-menu content-2 flyout width content-independent

wpears:
Closes https://github.local/CFPB/super-regular/issues/162

Previously, between 901px and 1260px, the width of the mega menu flyout could change based on the content. This was somewhat hard to reproduce  and resolved around `display: table-cell` weirdness of the child columns.

This now sets an explicit `width: 100%;` between those dimensions, which keeps the flyout a constant width, not matter the number of columns/presence of an overview link, etc.

Before:
Screen Shot 2020-09-25 at 5 04 29 PM

After:
Screen Shot 2020-09-25 at 5 05 07 PM

### DIFF
--- a/cfgov/unprocessed/css/organisms/mega-menu.less
+++ b/cfgov/unprocessed/css/organisms/mega-menu.less
@@ -1131,6 +1131,15 @@ body {
         }
 
     } );
+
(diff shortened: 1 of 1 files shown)


anselmbradford (approved):
LGTM 👍

PR #6045, opened 2020-09-28
Improve Docker stack notifications further and fix ES indexing when DB is not refreshed

marcesher:
This simplifies navigating to the Stack URL, which is helpful for accessing logs for both failed and successful deploys

Also removes unhelpful variation in main vs PR notifications, leading to a more consistent message for both failed and successful builds

This will result in messages similar to these:

Screen Shot 2020-09-27 at 11 35 18 PM
Screen Shot 2020-09-27 at 11 35 25 PM
Screen Shot 2020-09-27 at 11 35 53 PM


docker-stack.yml
```diff
@@ -87,8 +87,9 @@ services:
 
                 if [ $$django_tables_exist -gt 0 ] || [ $$REFRESH_DB == 'true' ]; then
                     ./refresh-data.sh 
-                    ./cfgov/manage.py update_index
                 fi
+
+                ./cfgov/manage.py update_index
```
chosak, line 92: @marcesher is this a deliberate change?

marcesher: @chosak Yes. I added an explanation to the commit but haven't added it to the PR yet. I discovered the bug while testing the notification changes. I will update the PR description with the explanation momentarily

marcesher: @chosak If you'd prefer I pull that into a separate PR, I'm happy to do so. Whatever the team norms are for a case like this, I'll gladly abide

chosak: I'll defer to others more familiar with the Elasticsearch changes -- I don't know enough about those tests to understand why this would be necessary again. Is it because the ES content gets wiped out with each redeploy, unlike the Postgres content (if `REFRESH_DB` is unset)?

Scotchester: The good news is that updating the index will not take long if there are no changes to it (if I remember correctly).

Scotchester: Wait 🤦  that's not true at all if the image is rebuilt and there is no index. I do think we need to take a look at that alternate approach sooner, rather than later.

Scotchester (approved):
Some wonderful improvements here; thank you, Marc!

docker-stack.yml
marcesher: Roger that, Scott. Just this morning, we prioritized solving the Jenkins job-dsl rebuild problem that we discussed elsewhere and which I anticipated would be a tough nut to crack, and fortunately we solved that unexpectedly quickly. So that serendipitously leaves an opening for us to prioritize this in our current sprint, as well. I'll talk to the team about it. 🙌 

Would you be OK merging this PR now, knowing that we'll prioritize the alternate approach this sprint?

Scotchester: Yeah, for sure! Merge when ready :)

PR #6046, opened 2020-09-28
Add better fallbacks for Wagtail pages with no search description #3828

alexhoffman617:
Wagtail pages without a provided search description use existing page content as a fallback description and only use the default CFPB description as a last resort:

1. The search description provided in Wagtail
2. Hero content
3. Text intro content
4. {Any other short pieces of content that might make a good description}
5. An empty string (because empty is better than repeating a generic description)

---

## Additions

-

## Removals

-

## Changes

-

## How to test this PR

1.

## Notes and todos

-

## Checklist

#### Browser testing

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Safari on iOS
- [ ] Chrome on Android

#### Accessibility

- [ ] Screen reader friendly

#### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/cfgov/jinja2/v1/_layouts/base.html
+++ b/cfgov/jinja2/v1/_layouts/base.html
@@ -60,9 +60,7 @@
     <meta name="description"
           content="
             {%- block desc -%}
-                {{- page.search_description if page and page.search_description
-                    else 'Our vision is a consumer finance marketplace that works for American consumers, responsible providers, and the economy as a whole.'
-                -}}
+                {{- meta_description -}}
             {%- endblock -%}
           ">
     <link rel="canonical" href="{{ request.build_absolute_uri() | lower }}">
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -255,6 +255,15 @@ class CFGOVPage(Page):
         if self.schema_json:
             context['schema_json'] = self.schema_json
 
+        if self.search_description:
+            context['meta_description'] = self.search_description
+        else:
+            context['meta_description'] = ''
+
+        #hero only in header
+        #text_introduction can be in header OR content
+        #test example base.py TestCFGOVPageContext
+
         return context
 
     def serve(self, request, *args, **kwargs):


cfgov/v1/models/base.py
```diff
@@ -255,6 +267,37 @@ def get_context(self, request, *args, **kwargs):
...
+            context['meta_description'] = self.get_streamfield_content(
+                self.content,
+                'text_introduction',
+                'intro')
+        if hasattr(self, 'header') and not context['meta_description']:
+            context['meta_description'] = self.get_streamfield_content(
+                self.header,
+                'item_introduction',
+                'paragraph')
+        if not context['meta_description']:
```
higs4281, line 298: Already set at 270

cfgov/v1/models/base.py
```diff
@@ -255,6 +267,37 @@ def get_context(self, request, *args, **kwargs):
...
+        if hasattr(self, 'content') and not context['meta_description']:
+            context['meta_description'] = self.get_streamfield_content(
+                self.content,
+                'text_introduction',
+                'intro')
+        if hasattr(self, 'header') and not context['meta_description']:
+            context['meta_description'] = self.get_streamfield_content(
+                self.header,
+                'item_introduction',
+                'paragraph')
```
higs4281, line 297: This may be in the reverse hierarchy intended. Each "if" section overwrites the previous one, if the sought content exists. Do we prefer search description if it exists, and then try to find other bits of content as fallbacks?

higs4281 (changes requested):

cfgov/v1/models/base.py
higs4281: But I see your test for blank meta_description fails without it.

higs4281: Ah -- the tests for existence of `header` or `content` then dive into each field's streamfield, which can return None, thus overriding line 270.

Rather than repeating tests for existence, It might be nicer to have a separate function that checks the various fields for possible meta_descriptions and returns either a winner or ''.

cfgov/v1/models/base.py
Scotchester: It would be a little cleaner to make every `if` after the first one an `elif` and get rid of all the `and not context['meta_description']`s.

cfgov/v1/models/base.py
```diff
@@ -255,6 +267,37 @@ def get_context(self, request, *args, **kwargs):
...
+        context['meta_description'] = ''
+        if self.search_description:
+            context['meta_description'] = self.search_description
+        if hasattr(self, 'header') and not context['meta_description']:
+            context['meta_description'] = self.get_streamfield_content(
+                self.header,
+                'hero',
+                'body')
+        if (
+            hasattr(self, 'preview_description')
```
Scotchester, line 279: I don't think this needs the `hasattr` constructions, since `preview_description` is a model field, and not part of a StreamField.

cfgov/v1/models/base.py
Scotchester: Well, maybe not. Because that would require some gnarly constructions for the things we're looking for inside StreamField blocks.

cfgov/v1/models/base.py
Scotchester: @higs4281 I'm not sure what this would look like that would be any more elegant than this brute-force approach. Can you sketch it out in pseudocode?

Scotchester: The `context['meta_description'] = ''` on L270 _is_ unnecessary, though, given this check at the end.

higs4281:
@Scotchester 
#6066 has function alternative.

cfgov/v1/models/base.py
```diff
@@ -235,6 +237,16 @@ def get_appropriate_descendants(self, inclusive=True):
     def get_appropriate_siblings(self, inclusive=True):
         return CFGOVPage.objects.live().sibling_of(self, inclusive)
 
+    def remove_html_tags(self, text):
+        clean = re.compile('<.*?>')
+        return re.sub(clean, ' ', text)
```
higs4281, line 242: This is outside scope of the PR, but a `strip_with_spaces` utility could be useful in other contexts, like building search indexes.

cfgov/jinja2/v1/_layouts/base.html
```diff
@@ -60,9 +60,7 @@
     <meta name="description"
           content="
             {%- block desc -%}
-                {{- page.search_description if page and page.search_description
-                    else 'Our vision is a consumer finance marketplace that works for American consumers, responsible providers, and the economy as a whole.'
-                -}}
+                {{- meta_description -}}
```
Scotchester, line 63: The new `meta_description` will only be available to Wagtail pages, and this template can serve plain Django pages, as well, so we'll need to maintain a check for `page` and offer a fallback.

```suggestion
                {{- meta_description if page else '' -}}
```

cfgov/v1/models/base.py
```diff
@@ -255,6 +297,10 @@ def get_context(self, request, *args, **kwargs):
         if self.schema_json:
             context['schema_json'] = self.schema_json
 
+        if self.search_description:
+            context['meta_description'] = self.search_description
+        else:
+            context['meta_description'] = self.get_meta_description()
         return context
```
higs4281, line 304: With search_description added to the function, we no longer need to check it here.

```suggestion
        context['meta_description'] = self.get_meta_description()
        return context
```

cfgov/v1/models/base.py
higs4281: `preview_description` is a defined model field on the 5 descendants of AbstractFilterPage, but if you try to call it from  another page type (e.g. AnswerPage) it will raise `AttributeError`. If we want this to work globally, the `hasattr` test is needed.

`search_description`, which is defined on wagtail.core.Page, is universal for us, so that doesn't need a test.

Scotchester: D'oh, right, sorry.

higs4281 (approved):

Scotchester (dismissed):
My final suggestion in `base.html` is definitely needed before merging.

PR #6049, opened 2020-09-28
Update curriculum-review-tool to version 2.1.3

nbucknor:
Bump Curriculum Review Tool to version [2.1.2](https://github.com/cfpb/curriculum-review-tool/compare/2.0.3...2.1.2)

Please create a feature branch environment for this, so we can get Product Owner approval before it is merged into the main branch.


requirements/libraries.txt
```diff
@@ -30,7 +30,7 @@ lxml==4.2.5
 Markdown==3.2.1
 ntplib==0.3.4
 openpyxl==3.0.3
-psycopg2==2.7.3.2
+psycopg2-binary==2.8.6
```
willbarton, line 33: @nbucknor Is there a reason why we're pinning this?

willbarton: It looks like y'all were seeing an error building psycopg2 somewhere, per https://github.com/cfpb/curriculum-review-tool/pull/413

I don't think this is necessary to pin here in consumerfinance.gov because we already successfully build psycopg2 for deployment to our servers and in our Docker environment.

Scotchester (approved):

PR #6052, opened 2020-09-30
Add ElasticSearch 7.4 Alpine to Docker Stack

alexm118:
Swaps over to using alpine based images for ElasticSearch and Kibana both locally and in our deployed docker environment. 

---

## Additions

- ES7 Alpine based image that incorporates our synonyms files and uses an elasticsearch user.

## Changes

- the ElasticSearch 7.4 image
- the container names for our elasticsearch containers, renaming ElasticSearch 2.3.5 (formerly elasticsearch) to elasticsearch2. This was done due to restraints connecting the new ES 7.4 image to kibana.

## How to test this PR

1. The docker deployed stack should be running a new ES7 service.
2. Locally you should see ES 2.3.5 and ES 7.4 running as well as kibana connected to ES 7.4

## Notes and todos

- **Please Note the updates to .env_SAMPLE** due to the renaming of our elasticsearch containers you will need to update your .env file if you use docker for elasticsearch!

## Checklist


docker-compose.yml
```diff
@@ -4,28 +4,32 @@ volumes:
...
     kibana:
-        image: kibana:7.4.2
+        image: blacktop/kibana:7.4
         depends_on:
-            - elasticsearch7
+            - elasticsearch
         ports:
             - "5601:5601"
-        environment:
-            ELASTICSEARCH_HOSTS: http://elasticsearch7:9200
```
Scotchester, line 28: Does the Kibana container no longer need this variable?

docker-compose.yml
```diff
@@ -4,28 +4,32 @@ volumes:
...
-        ports: 
             - "9000:9200"
             - "9100:9300"
+    elasticsearch:
+        build:
+            context: ./
+            dockerfile: ./docker/elasticsearch/7.4/Dockerfile
+        ports: 
+            - "9200:9200"
+            - "9300:9300"
```
Scotchester, line 20: You've swapped the way these ports are configured between the ES2 and ES7 services. Is that intentional?

docker-compose.yml
```diff
@@ -46,10 +50,13 @@ services:
...
         depends_on:
             - elasticsearch
-            - elasticsearch7
+            - elasticsearch2
             - postgres
         stdin_open: true
         tty: true
+        environment:
+            ES7_HOST: http://elasticsearch:9200
+            ES_HOST: http://elasticsearch2
```
Scotchester, line 59: Why are these needed now but weren't before? Why doesn't `ES_HOST` need a port?

docker-stack.yml
```diff
@@ -49,6 +63,7 @@ services:
       DATABASE_URL: postgres://cfpb@postgres/cfgov
       DJANGO_ADMIN_USERNAME: admin
       ES_HOST: elasticsearch
+      ES7_HOST: http://elasticsearch7:9200
```
Scotchester, line 66: This needs to be like this now, doesn't it?
```suggestion
      ES_HOST: elasticsearch2
      ES7_HOST: http://elasticsearch:9200
```

docker-stack.yml
```diff
@@ -87,10 +102,9 @@ services:
 
                 if [ $$django_tables_exist -gt 0 ] || [ $$REFRESH_DB == 'true' ]; then
                     ./refresh-data.sh 
+                    ./cfgov/manage.py update_index
                 fi
 
-                ./cfgov/manage.py update_index
-
```
Scotchester, line 93: Are you intentionally reverting the fix that Marc made in #6045? As noted there, the current situation is not ideal, but better than it was, I think.

Scotchester (commented):
A few questions from looking over the code that might be Docker/ES n00b questions, but here they are :)

A more overarching question: Are we ready to actually use this with application code? If not, I'd prefer not to be adding two new images every PR. (And I'd like to hear more about the use case for pgAdmin.)

docker-compose.yml
alexm118: I've noticed some flaws in the alpine Kibana image that aren't able to be overcome, one is that it basically only works connecting to `http://elasticsearch:9200` and that is its default setting. This is what lead to the renaming of our ElasticSearch containers and swapping the ports they expose locally.

docker-compose.yml
alexm118: Yes, due to the alpine kibana image assuming configuration and not being able to overwrite it (looks like a bug as their documentation indicates you should be able to overwrite the ES endpoint but haven't seen it work yet)

docker-compose.yml
alexm118: The configuration in our settings assumes the ES Port for Elasticsearch2 is 9200, for now I'm passing the port along with ES7 as that is how the django-elasticsearch-plugin by default expects its host to be set.

docker-stack.yml
alexm118: It doesn't but it might make more sense to do this swap. 

The docker stack is still deploying elasticsearch 2.3.5 as a container named elasticsearch, and I've just added elasticsearch7. This is not consistent with how we run them locally via docker-compose and we can make that change.

docker-stack.yml
alexm118: Nope - will add it back into the mix.

willbarton (approved):

docker-stack.yml
Scotchester: Yeah, I'd like to see this so it's consistent with the conventions in the `docker-compose.yml` file.

alexm118: Done!

Scotchester (approved):

PR #6058, opened 2020-10-01
Add Document class for Regulations3k

alexm118:
Adds the model for our regulations3k document that we search on. This will allow us to create and index all regulations in ES7.

---

## Additions

- `SectionParagraphDocument` - Represents the data we load into elasticsearch for regulations.

## How to test this PR

1. Inside the python container of our local docker stack run `cfgov/manage.py search_index --rebuild -f` and all regulations are loaded into ES7 index `regulations3k`. Can be confirmed via the kibana dashboard.

## Checklist


cfgov/cfgov/settings/local.py
```diff
@@ -17,7 +17,7 @@
         "for how to reinstall."
     )
 
-DEBUG = True
+DEBUG = False
```
willbarton, line 20: Why are we changing the local `DEBUG` setting?

willbarton (commented):
One quick comment, but I love how simple and readable this is.

cfgov/cfgov/settings/local.py
alexm118: Oops - I think I meant to disable Django Debug Toolbar locally and flipped the wrong value. I'll correct

cfgov/regulations3k/tests/test_documents.py
```diff
@@ -0,0 +1,42 @@
+from unittest import TestCase
```
higs4281, line 1: Would we want Django's TestCase here instead of unittest?

alexm118: Good Question, I hadn't considered the differences but upon some research I don't think we need Django's TestCase:

```
Using unittest.TestCase avoids the cost of running each test in a transaction and flushing the database, but if your tests interact with the database their behavior will vary based on the order that the test runner executes them. This can lead to unit tests that pass when run in isolation but fail when run in a suite.
```

This doesn't rely on a database, we are just defining the class and preparing the document in memory, so I think unittest.TestCase makes more sense here

cfgov/cfgov/settings/test.py
```diff
@@ -92,3 +92,6 @@
 HAYSTACK_CONNECTIONS["default"]["INDEX_NAME"] = (
     "test_" + HAYSTACK_CONNECTIONS["default"]["INDEX_NAME"]
 )
+
+ELASTICSEARCH_DSL_AUTO_REFRESH = False
+ELASTICSEARCH_DSL_AUTOSYNC = False
```
higs4281, line 97:
```suggestion
ELASTICSEARCH_DSL_AUTOSYNC = False

```

We exclude settings from linting, but I'd spend a line-ender.

higs4281 (approved):
One nit suggestion.

higs4281 (changes requested):
I'd like to do some more testing of Haystack operations.

higs4281 (approved):

PR #6060, opened 2020-10-01
Remove option to export page metadata as an .xlsx file

higs4281:
The XLSX download option was error-prone for page metadata reports,
possibly because of their size. The XLSX option required that the data be
held in memory before output, whereas the CSV option allows
for a more efficient streaming process.

In any case, not much is lost by removing the option, because the CSV can be
opened in Excel, which then saves the result as an .xlsx file by default.

## Update
The XLSX option was also removed from the site-history report.

## Testing
There should be no change in the CSV download option, which takes a bit of time to download.  
But the icon at the top right of the report page at `Reports -> Page Metadata` should change to a single choice:

Before:

csv_and_xlsx_download

Using this branch:

download_csv

### DIFF
--- a/cfgov/v1/templates/v1/page_metadata_report.html
+++ b/cfgov/v1/templates/v1/page_metadata_report.html
@@ -1,4 +1,13 @@
 {% extends 'wagtailadmin/reports/base_page_report.html' %}
+{% load i18n wagtailadmin_tags %}
+
+{% block actions %}
+    {% if view.list_export %}
+        <div class="button match-width">
+            <a href="{{ view.csv_export_url }}" class="button bicolor button--icon">{% icon name="download" wrapped=1 %}{% trans 'Download CSV' %}</a>
+        </div>
+    {% endif %}
+{% endblock %}
 
 {% block listing %}
     {% include 'v1/_list_page_metadata.html' %}
--- a/cfgov/v1/templates/v1/page_metadata_report.html
+++ b/cfgov/v1/templates/v1/page_metadata_report.html
@@ -3,7 +3,7 @@
 
 {% block actions %}
     {% if view.list_export %}
-        <div class="button match-width">
+        <div class="actionbutton">
             <a href="{{ view.csv_export_url }}" class="button bicolor button--icon">{% icon name="download" wrapped=1 %}{% trans 'Download CSV' %}</a>
         </div>
     {% endif %}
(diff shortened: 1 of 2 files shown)


Scotchester (approved):
This works. It makes the button slightly ugly, but I don't think I care enough to try to find a fix for that.

I wish there were a built-in way to add/remove options in a report's view code, without overriding the `actions` block. Maybe that's a future contribution we can make upstream.

cfgov/v1/templates/v1/page_metadata_report.html
```diff
@@ -1,4 +1,13 @@
 {% extends 'wagtailadmin/reports/base_page_report.html' %}
+{% load i18n wagtailadmin_tags %}
+
+{% block actions %}
+    {% if view.list_export %}
+        <div class="button match-width">
```
niqjohnson, line 6: Looks like the `button` class comes with some extra padding that's not needed if there's no dropdown. I think changing that to `actionbutton` should get it looking right (that's the class that's used for buttons like "add an image" in the image manager).

```suggestion
        <div class="actionbutton">
```

niqjohnson (changes requested):
One requested style change to get rid of some extra padding in the button, otherwise it's looking good.

This might be outside the scope of this PR, but is there a not-too-hard way to get rid of the XSLX option globally (even for the standard Wagtail reports)? Our site history report is also huge and gives a 504 error if you try to download the XSLX, though I don't imagine that report will get as many downloads as our custom ones.

willbarton (approved):
It's a shame that the Excel export action is [implemented entirely in the template](https://github.com/wagtail/wagtail/blob/master/wagtail/admin/templates/wagtailadmin/reports/base_report.html#L18-L28) like this.

higs4281:
@niqjohnson It wasn't too hard to remove the XLSX option for site history, but going beyond that does seem out of scope for this PR and runs into the new workflow reports, which I have not explored.

higs4281:
> It makes the button slightly ugly ...

@Scotchester Fixed, thanks to @niqjohnson.

niqjohnson (approved):
The button looks great now, @higs4281, thanks. I'm fine leaving the XSLX option available for the other reports since it's working fine for now.

Scotchester:
Thank you for being less lazy than me, @niqjohnson 😁

PR #6061, opened 2020-10-01
Keep duplicated filterable list params when paginating

### LINKED ISSUE #2199: Page number form in filterable list pagination molecule drops parameters

wpears:
Closes #2199 from over 4 years ago!

The problem here stemmed from the use of `request.GET.items()`, which returns only the last value when multiple params have the same key, by design.

See https://docs.djangoproject.com/en/3.1/ref/request-response/#django.http.QueryDict.items

Changing to `.lists()` and updating the logic a bit to handle the resulting lists stops params from being dropped.
Also, filtering out the 'page' param prevents the other bug here which was repeated page params in the querystring.

### DIFF
--- a/cfgov/jinja2/v1/_includes/molecules/pagination.html
+++ b/cfgov/jinja2/v1/_includes/molecules/pagination.html
@@ -72,12 +72,14 @@
                     {{ current_page }} out of {{ total_pages }} total pages
                 </span>
             </label>
-            {% for (key, value) in request.GET.items() %}
-                {% if value != '' %}
-                    <input type="hidden"
-                           name="{{ key }}"
-                           value="{{ value }}">
-                {% endif %}
+            {% for (key, value_as_list) in request.GET.lists() %}
+                {% for list_item in value_as_list %}
+                    {% if list_item != '' and key != 'page' %}
+                        <input type="hidden"
+                               name="{{ key }}"
+                               value="{{ list_item }}">
+                    {% endif %}
+                {% endfor %}
             {% endfor %}
             <input id="m-pagination_current-page-{{ index | string }}"
                    class="m-pagination_current-page"


willbarton (approved):
This is fine... I'm kind of surprised that we're diving into `request.GET` in a template, though. That should really be done in a Django form that hands off the appropriate values. But so should a lot of other places where we're using request params.

anselmbradford:
@wpears 

Do me a favor… 
 - pull this branch and run it
 - go to http://localhost:8000/blog 
 - expand the filter and check "At the CFPB" and "Director's notebook"
 - click "Apply filters'
 - scroll to the bottom of the results and enter "3" in the pagination component.
 - click "Go"

Do you get this error?

Screen Shot 2020-10-02 at 11 56 31 AM

wpears:
@anselmbradford I did encounter this (once) but I don't think it's related to this PR. I'm looking into the source of the error.

wpears:
Almost certainly related to https://github.com/cfpb/consumerfinance.gov/pull/5984

wpears:
Yeah confirmed, will merge this and PR a different fix (the error is local only)

PR #6066, opened 2020-10-05
Break meta-description ranking into function

higs4281:
This may not be much more efficient or elegant, but the order of preference seems clearer.

We don't lint v1 tests, but I also cleaned up some trailing-space issues in the tests.

### DIFF
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -264,7 +264,7 @@ class CFGOVPage(Page):
                 self.header, 'text_introduction', 'intro')
             candidates['header_item_intro'] = self.get_streamfield_content(
                 self.header, 'item_introduction', 'paragraph')
-        if hasattr(self, 'preview_description'):
+        if self.preview_description:
             candidates['preview_description'] = self.preview_description
         if hasattr(self, 'content'):
             candidates['content_text_intro'] = self.get_streamfield_content(
--- a/cfgov/v1/models/base.py
+++ b/cfgov/v1/models/base.py
@@ -247,6 +247,33 @@ class CFGOVPage(Page):
                 return self.remove_html_tags(item.value[value].source)
         return
 
+    def get_meta_description(self):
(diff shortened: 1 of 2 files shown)


cfgov/v1/models/base.py
```diff
@@ -247,6 +247,33 @@ def get_streamfield_content(self, section, blockType, value):
...
+    def get_meta_description(self):
+        """Deliver a meta_description, following a preference order."""
+        preference_order = [
+            'header_hero_body',
+            'preview_description',
+            'header_text_intro',
+            'content_text_intro',
+            'header_item_intro',
+        ]
+        candidates = {item: None for item in preference_order}
```
Scotchester, line 259: There's no particular need to set all of the keys, is there?

```suggestion
        candidates = {}
```

cfgov/v1/models/base.py
```diff
@@ -247,6 +247,33 @@ def get_streamfield_content(self, section, blockType, value):
...
+        if hasattr(self, 'header'):
+            candidates['header_hero_body'] = self.get_streamfield_content(
+                self.header, 'hero', 'body')
+            candidates['header_text_intro'] = self.get_streamfield_content(
+                self.header, 'text_introduction', 'intro')
+            candidates['header_item_intro'] = self.get_streamfield_content(
+                self.header, 'item_introduction', 'paragraph')
+        if hasattr(self, 'preview_description'):
+            candidates['preview_description'] = self.preview_description
```
Scotchester, line 268:
```suggestion
        if self.preview_description:
            candidates['preview_description'] = self.preview_description
```

Scotchester (dismissed):
I like this approach! A couple minor suggestions, but also: Any particular reason not to include `search_description` in the `get_meta_description()` method and skip the if/else in `get_context()`?

alexhoffman617 (approved):
Looking at this I do think its easier to understand the order of preference. I think thats worth the minimal extra loops to get all of the candidates. Merging this in and going with this method

PR #6072, opened 2020-10-08
Use newer mapbox styles in find-a-housing-counselor and r-o-u-t

wpears:
(see internal chat)


cfgov/unprocessed/apps/rural-or-underserved-tool/js/show-map.js
```diff
@@ -1,10 +1,10 @@
 import DT from './dom-tools';
 
 // See API docs at https://docs.mapbox.com/mapbox.js/api/v3.2.0/l-mapbox-map
-const MAPBOX_JS_URL = 'https://api.mapbox.com/mapbox.js/v3.2.0/mapbox.js';
-const MAPBOX_CSS_URL = 'https://api.mapbox.com/mapbox.js/v3.2.0/mapbox.css';
+const MAPBOX_JS_URL = 'https://api.mapbox.com/mapbox.js/v3.3.1/mapbox.js';
+const MAPBOX_CSS_URL = 'https://api.mapbox.com/mapbox.js/v3.3.1/mapbox.css';
 const mapboxAccessToken = 'pk.eyJ1IjoiY2ZwYiIsImEiOiJodmtiSk5zIn0.VkCynzmVYcLBxbyHzlvaQw';
-const mapIdString = 'cfpb.k55b27gd';
+const mapIdString = 'mapbox://styles/mapbox/streets-v11';
```
sephcoster, line 7: If an app uses multiple styles, it might be worth separating the prefix URI into a separate variable so we can declare additional layers with their IDs instead of the full string. Doesn't seem required for this particular use case though if it's just the one layer.

wpears: Yeah, I think that type of prefix + id code makes sense in the use case you describe, just seemed unnecessary here to me.

anselmbradford (approved):

cfgov/unprocessed/apps/rural-or-underserved-tool/js/show-map.js
```diff
@@ -92,11 +91,9 @@ resultsMapDom.addEventListener( 'click', function( evt ) {
 
         DT.nextFrame( function() {
           const latlng = window.L.latLng( lon, lat );
-          const map = window.L.mapbox.map(
-            id,
-            mapIdString,
-            { center: latlng }
-          ).setView( latlng, 12 );
+          const map = window.L.mapbox.map(id)
```
anselmbradford, line 94: Run `gulp lint --fix`

PR #6073, opened 2020-10-13
Isolate ES2.3.5 image creation to be singular vs every PR/Build/etc.

jduquette:
Isolate the regular creation of ElasticSearch images, to prevent unnecessary re-indexing.


higs4281:
CCDB is the only remaining user of Elasticsearch2, so as soon as they flip the switch to ES7, we can remove the ES2 container entirely.

Scotchester:
> CCDB is the only remaining user of Elasticsearch2, so as soon as they flip the switch to ES7, we can remove the ES2 container entirely.

Should we just close this, to ensure we don't accidentally disrupt them as they finish their transition?

Scotchester:
No longer worth doing.

PR #6074, opened 2020-10-14
Add SchoolDocument class for paying_for_college

cwdavies:
Adds the model for the paying_for_college document that we search on.
This will allow the creation and indexing of all schools in ES7.

---

## Additions

- `SchoolDocument` - Represents the data we load into elasticsearch for paying_for_college.

## How to test this PR

1. Inside the python container of our local docker stack run `cfgov/manage.py search_index --rebuild -f` and all schools are loaded into ES7 index `paying_for_college`. Can be confirmed via the kibana dashboard.

## Checklist

### DIFF
--- a/cfgov/ask_cfpb/documents.py
+++ b/cfgov/ask_cfpb/documents.py
@@ -1,36 +1,8 @@
 from django_elasticsearch_dsl import Document, fields
 from django_elasticsearch_dsl.registries import registry
-from elasticsearch_dsl import analyzer, token_filter, tokenizer
 
 from ask_cfpb.models.answer_page import AnswerPage
-
-
-label_autocomplete = analyzer(
-    'label_autocomplete',
-    tokenizer=tokenizer(
-        'trigram',
-        'edge_ngram',
-        min_gram=2,
-        max_gram=25,
-        token_chars=["letter", "digit"]
-    ),
-    filter=['lowercase', token_filter('ascii_fold', 'asciifolding')]
-)
-
-synonynm_filter = token_filter(
-    'synonym_filter_en',
-    'synonym',
-    synonyms_path='/usr/share/elasticsearch/config/synonyms/synonyms_en.txt'
-)
-
-synonym_analyzer = analyzer(
-    'synonym_analyzer_en',
-    type='custom',
-    tokenizer='standard',
-    filter=[
-        synonynm_filter,
-        'lowercase'
-    ])
+from search.elasticsearch_helpers import label_autocomplete, synonym_analyzer
 
 
 @registry.register_document
--- a/cfgov/ask_cfpb/documents.py
+++ b/cfgov/ask_cfpb/documents.py
@@ -1,8 +1,36 @@
 from django_elasticsearch_dsl import Document, fields
 from django_elasticsearch_dsl.registries import registry
+from elasticsearch_dsl import analyzer, token_filter, tokenizer
 
 from ask_cfpb.models.answer_page import AnswerPage
-from search.elasticsearch_helpers import label_autocomplete, synonym_analyzer
+
+
+label_autocomplete = analyzer(
+    'label_autocomplete',
+    tokenizer=tokenizer(
+        'trigram',
+        'edge_ngram',
+        min_gram=2,
+        max_gram=25,
+        token_chars=["letter", "digit"]
+    ),
+    filter=['lowercase', token_filter('ascii_fold', 'asciifolding')]
+)
+
+synonynm_filter = token_filter(
+    'synonym_filter_en',
+    'synonym',
+    synonyms_path='/usr/share/elasticsearch/config/synonyms/synonyms_en.txt'
+)
+
+synonym_analyzer = analyzer(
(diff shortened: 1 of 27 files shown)


cfgov/paying_for_college/documents.py
```diff
@@ -0,0 +1,60 @@
...
+    city = fields.TextField()
+    state = fields.TextField()
+
+    def prepare_autocomplete(self, instance):
+        alias_strings = [a.alias for a in instance.alias_set.all()]
+        nickname_strings = [n.nickname for n in instance.nickname_set.all()]
+        return alias_strings + nickname_strings
+
+    def prepare_city(self, instance):
+        return instance.city
```
alexm118, line 42: We can use a fields list instead of a prepare function if we aren't making any changes to a field.

https://django-elasticsearch-dsl.readthedocs.io/en/latest/quickstart.html

cfgov/paying_for_college/documents.py
```diff
@@ -0,0 +1,60 @@
+from django_elasticsearch_dsl import Document, fields
+from django_elasticsearch_dsl.registries import registry
+from elasticsearch_dsl import analyzer, token_filter, tokenizer
+
+from paying_for_college.models import School
+
+
+label_autocomplete = analyzer(
```
alexm118, line 8: We shouldn't duplicate our analyzers and filters, maybe define a generic `elasticsearch_helpers.py` file in the search app.

cfgov/paying_for_college/tests/test_documents.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def test_prepare(self):
+        school = School(school_id="999999",
+                        settlement_school="Example University",
+                        city="Example City", state="VA")
+        doc = SchoolDocument()
+
+        prepared_data = doc.prepare(school)
+        self.assertEqual(prepared_data, {
+            'autocomplete': doc.prepare_autocomplete(school),
```
alexm118, line 17: I don't think we should be calling `doc.prepare_*` in our test assertions for the expected results, we can hard code them, or reference them directly from the school object. This test would be more testing that the `prepare` function calls each defined fields `prepare` function, rather than what we want to test which is that the prepare function properly maps all fields from our django model to the document class.

alexm118 (commented):
Just a couple suggestions and a fix or two.

cfgov/paying_for_college/tests/test_documents.py
cwdavies: Removed doc.prepare_city and doc.prepare_state from test assertion.
The city and state are now referenced from the school object.

cfgov/ask_cfpb/documents.py
```diff
@@ -1,36 +1,8 @@
 from django_elasticsearch_dsl import Document, fields
 from django_elasticsearch_dsl.registries import registry
```
higs4281, line 2: Should these changes be i a separate PR for ask_cfpb? (same issue in PR 6075)

cwdavies: Agree. I created a separate ask_cfpb PR: https://github.com/cfpb/consumerfinance.gov/pull/6076

cfgov/paying_for_college/documents.py
```diff
@@ -0,0 +1,35 @@
...
+
+from paying_for_college.models import School
+from search.elasticsearch_helpers import label_autocomplete
+
+
+@registry.register_document
+class SchoolDocument(Document):
+
+    autocomplete = fields.TextField(analyzer=label_autocomplete)
+    text = fields.TextField(attr='settlement_school')
```
higs4281, line 12: We should not be indexing the "settlement_school" field. This is for internal use and would server no search purpose.

The things that would be useful for search are:
- primary_alias (a property that needs to called in via "prepare" on an instance, since it's not a db column
- pk
- city
- state
- zip5

cwdavies: Added zip5 and changed settlement_school to primary_alias

cfgov/paying_for_college/documents.py
```diff
@@ -0,0 +1,39 @@
...
+    autocomplete = fields.TextField(analyzer=label_autocomplete)
+    text = fields.TextField(attr='primary_alias', boost=10)
+    nicknames = fields.TextField()
+
+    def prepare_autocomplete(self, instance):
+        alias_strings = [a.alias for a in instance.alias_set.all()]
+        nickname_strings = [n.nickname for n in instance.nickname_set.all()]
+        return alias_strings + nickname_strings
+
+    def prepare_primary_alias(self, instance):
```
alexm118, line 20: This can exist just within the fields list below.

alexm118: Also - it looks like this prepare function isn't needed as you haven't defined a `primary_alias` field on this document. You are accessing the `primary_alias` data by indexing it within the `text` field.

cwdavies: Removed prepare_primary_alias

alexm118 (approved):
LGTM

PR #6075, opened 2020-10-15
Add Elasticsearch 7 support to paying_for_college search behind a flag

cwdavies:
Places ES7 paying_for_college search functionality behind a flag

---

## Additions

- Modified paying_for_college search code to work with ES7 behind a flag
- Created `test_search.py` with `SchoolSearchTest` class

## Updates

- Moved `test_school_search` from `test_views.py` into `test_search.py`

## How to test this PR

1. Enable `ELASTICSEARCH_DSL_PFC` flag
2. Load up ES7 index `cfgov/manage.py search_index --rebuild -f --parallel`
3. Search paying_for_college
https://cfgov-pr-6075.demo.cfpb.gov/paying-for-college2/understanding-your-financial-aid-offer/api/search-schools.json?q=Kansas

## Checklist


cfgov/ask_cfpb/documents.py
```diff
@@ -1,36 +1,8 @@
 from django_elasticsearch_dsl import Document, fields
 from django_elasticsearch_dsl.registries import registry
```
higs4281, line 2: Should these changes be in a separate PR for ask_cfpb?

higs4281 (commented):
A couple minor comments.
Still want to run more code.

cfgov/ask_cfpb/documents.py
cwdavies: Agree. I created a separate ask_cfpb PR: https://github.com/cfpb/consumerfinance.gov/pull/6076

cfgov/paying_for_college/views.py
```diff
@@ -292,6 +293,19 @@ def get(self, request):
...
+    if not safe_term:
+        return HttpResponse(json.dumps([]), content_type='application/json')
+    try:
+        results = SchoolSearch(search_term=safe_term).autocomplete()
+        response = json.dumps(results)
+        return HttpResponse(response, content_type='application/json')
+    except IndexError:
+        return HttpResponse(json.dumps([]), content_type='application/json')
+
+
```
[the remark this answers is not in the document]
cwdavies: Agreed and Done

cfgov/paying_for_college/disclosures/urls.py
```diff
@@ -25,8 +27,12 @@
                 template_name="paying_for_college/disclosure_technote.html"),
             name="pfc-technote"),
 
-    re_path(r"^api/search-schools.json", school_search_api,
-            name="school_search"),
+    flagged_re_path(
+        'ELASTICSEARCH_DSL_PFC',
+        r"^api/search-schools.json",
```
alexm118, line 32: Unrelated but for my understanding - why do we have a file type ending in a route? Does this return a file?

cfgov/paying_for_college/models/search.py
```diff
@@ -0,0 +1,44 @@
+from paying_for_college.documents import SchoolDocument
+
+
+UNSAFE_CHARACTERS = [
+    '#', '%', ';', '^', '~', '`', '|',
+    '<', '>', '[', ']', '{', '}', '\\'
+]
+
+
+def make_safe(term):
```
alexm118, line 10: Don't we have a similar function for `ask-cfpb`? We should try to reuse them.

cfgov/paying_for_college/models/search.py
```diff
@@ -0,0 +1,44 @@
...
+    for char in UNSAFE_CHARACTERS:
+        term = term.replace(char, '')
+    return term
+
+
+class SchoolSearch:
+    def __init__(self, search_term, base_query=None):
```
alexm118, line 17: It looks like theres never a time when `base_query` could be passed into this class, we should remove it and clean up the search logic.

cfgov/paying_for_college/views.py
```diff
@@ -309,6 +311,27 @@ def school_search_api(request):
...
+        return HttpResponse(json.dumps([]), content_type='application/json')
+    response = SchoolSearch(search_term).search()
+
+    document = [{'schoolname': school.text,
+                 'id': school.school_id,
+                 'city': school.city,
+                 'nicknames': school.nicknames,
+                 'state': school.state,
+                 'zip5': school.zip5,
+                 'url': reverse("paying_for_college:disclosures:school-json",
```
alexm118, line 327: Depending on time, it may be easier to just index the url field and map it here instead of calling reverse on each result everytime it gets returned.

alexm118 (commented):
Couple of cleanup recommendations.

cfgov/paying_for_college/models/search.py
cwdavies: Agree and done. This is a great suggestion. I removed `base_query` and updated search logic.

cfgov/paying_for_college/models/search.py
cwdavies: Yes. I moved `UNSAFE_CHARACTERS` and `make_safe` to `elasticsearch_helpers.py` to be reused by `ask-cfpb`

cfgov/paying_for_college/disclosures/urls.py
cwdavies: @higs4281 Do you recall if there is a file type ending in a route that returns a file?

cfgov/paying_for_college/views.py
cwdavies: I will look into the changes needed to add the `url` field into the `SchoolDocument` class.

cwdavies: Also removed `reverse` from `url` in `school_search`

cfgov/paying_for_college/models/search.py
```diff
@@ -0,0 +1,32 @@
...
+        self.results = []
+
+    def autocomplete(self):
+        if self.search_term != '':
+            search = SchoolDocument.search().query(
+                'match', autocomplete=self.search_term)
+            total_results = search.count()
+            self.results = [
+                {'question': str(result.autocomplete), 'url': result.url}
+                for result in search[:total_results]
```
higs4281, line 17: This appears to be copied from ask-cfpb. The return values need to match what the page is expecting:
id, schoolname, city, state.

cfgov/paying_for_college/documents.py
```diff
@@ -10,6 +12,7 @@ class SchoolDocument(Document):
 
     autocomplete = fields.TextField(analyzer=label_autocomplete)
     text = fields.TextField(attr='primary_alias', boost=10)
+    url = fields.TextField()
     nicknames = fields.TextField()
 
     def prepare_autocomplete(self, instance):
```
higs4281, line 18: Autocomplete is not catching nicknames or school IDs. I think we need to send all those values to the index as a string. Something like this:

```python
    def prepare_autocomplete(self, instance):
        alias_strings = [a.alias for a in instance.alias_set.all()]
        nickname_strings = [n.nickname for n in instance.nickname_set.all()]
        auto_strings = alias_strings + nickname_strings + [str(instance.pk)]
        return " ".join(auto_strings)
```

Note that I can't make this a code suggestion because I can't select the full method.

cfgov/paying_for_college/documents.py
```diff
@@ -20,6 +23,10 @@ def prepare_autocomplete(self, instance):
     def prepare_nicknames(self, instance):
         return [n.nickname for n in instance.nickname_set.all()]
 
+    def prepare_url(self, instance):
+        return reverse("paying_for_college:disclosures:school-json",
+                       args=[instance.school_id])
+
```
higs4281, line 29: We need to exclude schools that have closed from being indexed. The ES7 index is finding 8,693 schools, but it should only find 6,910, because the others have closed.

We can override the document's `get_queryset` method to accomplish this:

```suggestion

    def get_queryset(self):
        """Prevent schools that have closed from being indexed."""
        return super(
            SchoolDocument, self).get_queryset().filter(operating=True)

```

higs4281 (changes requested):
This is looking good, but I think we need to make three changes to exclude schools that have closed, and to fix autocomplete to enable searches by nickname and school ID.

cfgov/paying_for_college/disclosures/urls.py
```diff
@@ -25,8 +27,12 @@
...
+        school_search,
```
higs4281, line 33: The app only uses an autocomplete search box, so we need to point the url at the `school_autocomplete` view:

```suggestion
        school_autocomplete,
```

cfgov/paying_for_college/disclosures/urls.py
```diff
@@ -1,9 +1,11 @@
 from django.urls import re_path
 
+from flags.urls import flagged_re_path
+
 from paying_for_college.views import (
     BaseTemplateView, ConstantsRepresentation, ExpenseRepresentation,
     FeedbackView, OfferView, ProgramRepresentation, SchoolRepresentation,
-    StatsRepresentation, VerifyView, school_search_api
+    StatsRepresentation, VerifyView, school_search, school_search_api
```
higs4281, line 8:
```suggestion
    StatsRepresentation, VerifyView, school_autocomplete, school_search_api
```

cfgov/paying_for_college/models/search.py
```diff
@@ -0,0 +1,32 @@
...
+            ]
+        return self.results
```
higs4281, line 19: The autocomplete method is missing an `execute` step. We also don't need to return all results during the autocomplete result sequence. 10 max, I'd say.

```suggestion
            response = search.execute()
            self.results = [{
                'schoolname': result.text,
                'id': result.school_id,
                'city': result.city,
                'state': result.state}
                for result in response[:10]
            ]
        return self.results
```

cfgov/paying_for_college/documents.py
cwdavies: I updated autocomplete to include school IDs.

cfgov/paying_for_college/documents.py
cwdavies: That explains why the number of schools found in the index was incorrect.
I made the change necessary to filter on schools that are operating which should exclude schools that are closed.

cfgov/paying_for_college/disclosures/urls.py
cwdavies: Done. I changed `school_search` to `school_autocomplete` in `urls.py`.

cfgov/paying_for_college/disclosures/urls.py
cwdavies: Done. I changed `school_search` to `school_autocomplete` in `urls.py`.

cfgov/paying_for_college/views.py
```diff
@@ -309,6 +311,33 @@ def school_search_api(request):
...
+    document = [{'schoolname': school.text,
+                 'id': school.school_id,
+                 'city': school.city,
+                 'nicknames': str(school.nicknames),
+                 'state': school.state,
+                 'zip5': school.zip5,
+                 'url': school.url}
+                for school in response.get('results')]
+
+    return HttpResponse(json.dumps(document), content_type='application/json')
```
higs4281, line 338: I think school_autocomplete is the only es7 view we need. There is no other use of PFC search.

```suggestion
```

cfgov/paying_for_college/views.py
```diff
@@ -309,6 +311,33 @@ def school_search_api(request):
     return HttpResponse(json_doc, content_type='application/json')
 
 
+def school_autocomplete(request):
+    search_term = request.GET.get('q', '').strip()
+    if not search_term:
+        return HttpResponse(json.dumps([]), content_type='application/json')
```
chosak, line 317: Could you remove some repeated use of `HttpResponse(json.dumps(...), content_type='application/json')` by using [`JsonResponse`](https://docs.djangoproject.com/en/3.1/ref/request-response/#jsonresponse-objects) instead?

```suggestion
        return JsonResponse([])
```

cfgov/paying_for_college/models/search.py
cwdavies: I fixed the results definition and added the missing `execute` step. I also checked the `autocomplete` method in `search.py` for `ask_cfpb` and that is also missing the `execute` step.

cfgov/paying_for_college/views.py
higs4281: I think we should switch to JsonResponse in most cases here, but when the data source is a list, I think we'll raise security hackles with the required safe=False.

higs4281: We could make sure all data payloads are dicts, but then some of the simplicity goes out the window, and we have to handle the extra unpacking in a template.

cfgov/paying_for_college/views.py
cwdavies: Done. I replaced `school_search` with `school_autocomplete` as the only ES7 view.

cfgov/paying_for_college/views.py
cwdavies: Changed school_autocomplete to return a JsonResponse

cwdavies:
> This is looking good, but I think we need to make changes to exclude schools that have closed, and to fix autocomplete to enable searches by nickname and school ID.

I recommend having a test to confirm a school that has closed is excluded.
Do you know of a specific school that is closed and is in test_fixture.json?

I hope that I have fixed autocomplete to enable searches by school_id and nickname.
Please let me know if those type of searches are working and I will add new test cases.

higs4281:
> Do you know of a specific school that is closed and is in test_fixture.json?

All the schools in test_fixture.json are `operating=true`, but you could add one that is closed, or add one within a test to avoid changing result counts in other tests.

cfgov/paying_for_college/models/search.py
```diff
@@ -0,0 +1,21 @@
...
+        self.results = []
+
+    def autocomplete(self):
+        if self.search_term != '':
+            search = SchoolDocument.search().query(
+                'match', autocomplete=self.search_term)
+            total_results = search.count()
+            search = search[0:total_results]
+            response = search.execute()
+            self.results = response[0:total_results]
```
higs4281, line 17: Why are we still capturing a count and delivering the entire response set? For autocomplete, this delivers tons of unhelpful partial matches. The default top 10 objects is a better starting point.

```suggestion
            response = search.execute()
            self.results = response
```

higs4281 (changes requested):
I think we're close. Just two suggested tweaks, for documents.py and models.search.py

cfgov/paying_for_college/views.py
```diff
@@ -309,6 +311,24 @@ def school_search_api(request):
...
+def school_autocomplete(request):
+    document = []
+    search_term = request.GET.get('q', '').strip()
+    if search_term:
+        response = SchoolSearch(search_term).autocomplete()
+
+        document = [{'schoolname': school.text,
+                     'id': school.school_id,
+                     'city': school.city,
+                     'nicknames': str(school.nicknames),
```
higs4281, line 323: No longer need to force the list to a string:

```suggestion
                     'nicknames': school.nicknames,
```

cfgov/paying_for_college/models/search.py
cwdavies: Modified autocomplete to return the default top 10 hits instead of the entire response set

cfgov/paying_for_college/documents.py
```diff
@@ -10,16 +12,27 @@ class SchoolDocument(Document):
...
     def prepare_autocomplete(self, instance):
         alias_strings = [a.alias for a in instance.alias_set.all()]
         nickname_strings = [n.nickname for n in instance.nickname_set.all()]
-        return alias_strings + nickname_strings
+        auto_strings = alias_strings + nickname_strings + [str(instance.pk)]
+        return " ".join(auto_strings)
 
     def prepare_nicknames(self, instance):
         return [n.nickname for n in instance.nickname_set.all()]
```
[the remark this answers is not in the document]
cwdavies: Updated prepare_nicknames to return a string instead of a list in SchoolDocument

cfgov/paying_for_college/views.py
cwdavies: Removed str from nicknames in school_autocomplete because list-as-string is not friendly in the API response

higs4281 (approved):
App's working well now. Great work!

cfgov/paying_for_college/documents.py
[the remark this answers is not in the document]
alexm118: I believe a `KeywordField` is the ideal type if you want to provide a list of strings.

cfgov/paying_for_college/models/search.py
alexm118: Agreed, the only reason to do a count is to get all results, something like autocomplete is better off returning the highest 10 matches.

PR #6080, opened 2020-10-19
Add archive options to Filterable List Controls

willbarton:
This PR adds the ability to filter based on the archive status of pages to our filterable list controls.

image

If the filterable pages in the filterable list contains archived pages, the option to either exclude, include, or only filter archived pages will be presented to users (defaulting to exclude).

## How to test this PR

1. Visit a filterable list page, i.e. http://localhost:8000/data-research/research-reports
2. Observe that the option to filter based on archived items does not appear
3. Archive a page that is within that filterable list page and publish the change
4. Observe that the option to filter based on archived items does appear
5. Observe that the newly archived page is excluded from the results by default

## Notes and todos

This PR intentionally preserves for now the ability to set filterable pages to include archived pages *only* that was introduced in #6039, although the functionality this PR introduces makes that unnecessary. Depending on how archiving ends up implemented in practice, one of these approaches will need some cleanup.

## Checklist


alexm118 (approved):

higs4281 (approved):
Works as advertised, and "Exclude archived items" shows up as the default.

PR #6083, opened 2020-10-22
Use Django forms to sanitize and validate prepaid search parameters

willbarton:
This change uses Django forms to validate and sanitize the prepaid agreements search parameters.

It adds two forms. `SearchForm` and `FilterForm` that handle the search fields and the filter fields, respectively. These are two separate forms because the filter choices depend on the results from the search.

At present this does not present any information to users on the validity of the form submission — default results will be returned for search if the search query, field, or results page number are not valid; the default filter selection (i.e. no filters) will be returned if any filter option is invalid.

The motivation for this change is to prevent a 500 error on invalid filters.

This change also removes the `PREPAID_AGREEMENTS_SEARCH` feature flag.

## How to test this PR

Compare the results from a search that includes a null character as part of a filter option between beta and localhost (or our dev Docker cluster):

- https://beta.consumerfinance.gov/data-research/prepaid-accounts/search-agreements/?prepaid_type=Travel%00
- http://localhost:8000/data-research/prepaid-accounts/search-agreements/?prepaid_type=Travel%00

Note that the request to beta produces a 500 error, where localhost does not.

Run the Cypress prepaid agreements tests (`yarn run cypress open`) to confirm that general behavior is as expected, and/or compare general search and filter option selection between beta and localhost (or our dev Docker cluster):

- https://beta.consumerfinance.gov/data-research/prepaid-accounts/search-agreements/
- http://localhost:8000/data-research/prepaid-accounts/search-agreements/

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -735,8 +735,6 @@ FLAGS = {
     "BETA_EXTERNAL_TESTING": [],
     # Used to hide new youth employment success pages prior to public launch
     "YOUTH_EMPLOYMENT_SUCCESS": [],
-    # Release of prepaid agreements database search
-    "PREPAID_AGREEMENTS_SEARCH": [],
     # Used to hide CCDB landing page updates prior to public launch
     "CCDB_CONTENT_UPDATES": [],
     # During a Salesforce system outage, the following flag should be enabled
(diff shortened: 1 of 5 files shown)


cwdavies (approved):
I verified that the following link causes a 500: Server error:
https://beta.consumerfinance.gov/data-research/prepaid-accounts/search-agreements/?prepaid_type=Travel%00

The Prepaid product agreements page loaded correctly using the following PR links:
https://cfgov-pr-6083.demo.cfpb.gov/data-research/prepaid-accounts/search-agreements/?prepaid_type=Travel%00
https://cfgov-pr-6083.demo.cfpb.gov/data-research/prepaid-accounts/search-agreements/?status=Active%00

alexm118 (approved):

PR #6086, opened 2020-10-26
Prepaid: Ensure that the paginator is always getting a valid page

willbarton:
This changes moves where we get the page number from the search form up into the same block where we ensure the form is valid.

If the form was not valid, we were still trying to get the `page` from the form's cleaned_data, and that could fail. This fixes that issue.

## How to test this PR

Try to give the prepaid search an invalid `page`, i.e.: 

http://localhost:8000/data-research/prepaid-accounts/search-agreements/?page=3?page=3

And notice this 500s on `main`, but does not on this branch.

## Checklist


Scotchester (approved):
Confirmed, does not 500.

PR #6088, opened 2020-10-28
Update design-system to 0.4.0

anselmbradford:
## Changes

- Update design-system to 0.4.0 to bring in https://github.com/cfpb/design-system/pull/1080

## How to test this PR

1. Pull branch and build and visit and see that the two digit bulleted list on http://localhost:8000/administrative-adjudication-proceedings/ is not cropped. 

## Screenshots

Before:
Screen Shot 2020-10-28 at 2 12 22 PM

After:
Screen Shot 2020-10-28 at 2 14 07 PM


contolini (approved):
IT'S FIXED. so long as we avoid going over 99 list items which from a UX standpoint we can only hope never happens.

PR #6092, opened 2020-10-30
Refactor TDP activities search for Elasticsearch 7

higs4281:
In switching TDP's search from haystack to django-elasticsearch-dsl,
we also wanted to improve the search page's efficiency by avoiding
hundreds of database and elasticsearch queries per search.

This PR adds a db table to store a cache of card and facet setups to avoid 
most of the database roundtrips, as well as a good chunk of Elasticsearch calls.

The indexing is fairly quick, which bodes well for setting up auto-refresh 
of the index and of the card/facet map. The cache may not need to be updated often; 
the models haven't changed in months, and that pattern could continue.

The PR is a work in progress, with these items under construction:
- [x] Refining the faceting to meet TDP requirements
- [x] Deciding on an update scheme for the cache.  **I think this should be another PR; this one's too big already.**
- [x] Tests! ~~[a few more to add]~~ **New code is now tested.**

## Testing
This app is best tested in docker development mode, so that Elasticsearch 2 or 7 
can be run.
- Pull in the branch and either refresh your database or run `migrate`

```
docker-compose exec python bash

# AND

./refresh-data.sh

# OR

./cfgov/manage.py migrate`

# then build the index:

./cfgov/manage.py search_index --rebuild -f --models teachers_digital_platform
```

You could  also run the Elasticsearch2 index to confirm that that process still works:

```
./cfgov/manage.py update_index teachers_digital_platform
```

- Confirm the haystack-driven page still works (but faster, because of the cached objects):   
<http://localhost:8000/practitioner-resources/youth-financial-education/teach/activities/>
- Open your local Wagtail admin and enable the `ELASTICSEARCH_DSL_TDP` flag at http://localhost:8000/admin/flags/ELASTICSEARCH_DSL_TDP/  

That will switch the TDP search from using haystack to using django-elasticsearch-dsl and Elasticsearch7.  

Then you can compare the dsl-driven local page with results in prod at  
<https://www.consumerfinance.gov/practitioner-resources/youth-financial-education/teach/activities/>

## Notes
Search-term queries should work. ~~but faceting doesn't work right yet in the nested-facet area ("Topic"), and when multiple facets are chosen within a single heading~~  

The current tool is built to use AND logic between search terms and between search and faceting choice groups.  
Subsequent facet group choices use OR logic, which *increases* hits.

One beef I have with the faceting behavior is that it uses the term "filter" for these facet changes, which I think is confusing.
Filters reduce things, but multiple filter choices in the live tool *increase* the hit count. 
I think "refine" might be a better word, and something like this might be less confusing:

> Refine results  
Fine-tune your search by selecting activity features to include.

And, in the filter status area above results: 

> Choices applied

But I've beaten this drum before.


cwdavies (changes requested):
There is a lint error to fix:
./cfgov/teachers_digital_platform/models/activity_index_page.py:12:1: F401 'elasticsearch_dsl.A' imported but unused

higs4281:
> There is a lint error to fix

Thanks. Fixed.

cwdavies (changes requested):
6 teachers_digital_platform tests have errors when running `tox -e unittest-current teachers_digital_platform.tests`
```
ERROR: test_activity_index_page_renders (teachers_digital_platform.tests.models.test_activity_index_page.TestActivityIndexPageSearch)

  File "/Users/DaviesC/Projects/consumerfinance.gov/cfgov/teachers_digital_platform/tests/models/test_activity_index_page.py", line 90, in test_activity_index_page_renders
    response.render()
AttributeError: 'HttpResponse' object has no attribute 'render'


ERROR: test_activity_index_page_renders_with_query_parameters (teachers_digital_platform.tests.models.test_activity_index_page.TestActivityIndexPageSearch)

  File "/Users/DaviesC/Projects/consumerfinance.gov/cfgov/teachers_digital_platform/tests/models/test_activity_index_page.py", line 98, in test_activity_index_page_renders_with_query_parameters
    response.render()
AttributeError: 'HttpResponse' object has no attribute 'render'


ERROR: test_search_get_all_facets_with_building_block_filter (teachers_digital_platform.tests.models.test_activity_index_page.TestActivityIndexPageSearch)

  File "/Users/DaviesC/Projects/consumerfinance.gov/.tox/unittest-current/lib/python3.6/site-packages/mock/mock.py", line 1297, in patched
    arg = patching.__enter__()
  File "/Users/DaviesC/Projects/consumerfinance.gov/.tox/unittest-current/lib/python3.6/site-packages/mock/mock.py", line 1353, in __enter__
    self.target = self.getter()
  File "/Users/DaviesC/Projects/consumerfinance.gov/.tox/unittest-current/lib/python3.6/site-packages/mock/mock.py", line 1523, in <lambda>
    getter = lambda: _importer(target)
  File "/Users/DaviesC/Projects/consumerfinance.gov/.tox/unittest-current/lib/python3.6/site-packages/mock/mock.py", line 1210, in _importer
    thing = _dot_lookup(thing, comp, import_path)
  File "/Users/DaviesC/Projects/consumerfinance.gov/.tox/unittest-current/lib/python3.6/site-packages/mock/mock.py", line 1199, in _dot_lookup
    __import__(import_path)
ModuleNotFoundError: No module named 'teachers_digital_platform.models.pages.SearchQuerySet'; 'teachers_digital_platform.models.pages' is not a package


ERROR: test_search_get_all_facets_with_topic_block_is_nested_filter (teachers_digital_platform.tests.models.test_activity_index_page.TestActivityIndexPageSearch)

  File "/Users/DaviesC/Projects/consumerfinance.gov/.tox/unittest-current/lib/python3.6/site-packages/mock/mock.py", line 1297, in patched
    arg = patching.__enter__()
  File "/Users/DaviesC/Projects/consumerfinance.gov/.tox/unittest-current/lib/python3.6/site-packages/mock/mock.py", line 1353, in __enter__
    self.target = self.getter()
  File "/Users/DaviesC/Projects/consumerfinance.gov/.tox/unittest-current/lib/python3.6/site-packages/mock/mock.py", line 1523, in <lambda>
    getter = lambda: _importer(target)
  File "/Users/DaviesC/Projects/consumerfinance.gov/.tox/unittest-current/lib/python3.6/site-packages/mock/mock.py", line 1210, in _importer
    thing = _dot_lookup(thing, comp, import_path)
  File "/Users/DaviesC/Projects/consumerfinance.gov/.tox/unittest-current/lib/python3.6/site-packages/mock/mock.py", line 1199, in _dot_lookup
    __import__(import_path)
ModuleNotFoundError: No module named 'teachers_digital_platform.models.pages.SearchQuerySet'; 'teachers_digital_platform.models.pages' is not a package


ERROR: test_search_index_page_handles_bad_query (teachers_digital_platform.tests.models.test_activity_index_page.TestActivityIndexPageSearch)

  File "/Users/DaviesC/Projects/consumerfinance.gov/cfgov/teachers_digital_platform/tests/models/test_activity_index_page.py", line 127, in test_search_index_page_handles_bad_query
    response.render()
AttributeError: 'HttpResponse' object has no attribute 'render'


ERROR: teachers_digital_platform.tests.models.test_pages_utility_definitions (unittest.loader._FailedTest)

ImportError: Failed to import test module: teachers_digital_platform.tests.models.test_pages_utility_definitions
  File "/Users/DaviesC/.pyenv/versions/3.6.9/lib/python3.6/unittest/loader.py", line 428, in _find_test_path
    module = self._get_module_from_name(name)
  File "/Users/DaviesC/.pyenv/versions/3.6.9/lib/python3.6/unittest/loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "/Users/DaviesC/Projects/consumerfinance.gov/cfgov/teachers_digital_platform/tests/models/test_pages_utility_definitions.py", line 8, in <module>
    from teachers_digital_platform.models.pages import validate_results_per_page
ImportError: cannot import name 'validate_results_per_page'

Ran 23 tests in 4.788s

FAILED (errors=6)
```

higs4281:
> 6 teachers_digital_platform tests have errors when running `tox -e unittest-current teachers_digital_platform.tests`

Still adding tests, but these all run now.

nbucknor (commented):
@higs4281 A bit of functionality thats been lost with the new elasticsearch solution is, determining the available facets for a group, based on the selected facets in other groups. That would explain why I've been seeing more facets that return no results when testing this branch locally. 

https://github.com/cfpb/consumerfinance.gov/pull/6092/files/b74478adfde2c63e3902a3648a5aead5973182e0#diff-299c968a685ea3297d6e24f5f875bb75d75a6d045f92beb7c53ce28f358feaf2R304-R310

cfgov/teachers_digital_platform/models/activity_index_page.py
```diff
@@ -0,0 +1,528 @@
...
+        return context
+
+    # TODO: Remove this function when we switch to DSL
+    def get_all_facets(self, facet_map, sqs, facet_counts, facet_queries, selected_facets):  # noqa: E501
+        all_facets = {}
+        if 'fields' in facet_counts:
+            for facet, facet_config in facet_map:
+                class_object, is_nested = facet_config
+                all_facets_sqs = sqs
+                other_facet_queries = [
```
nbucknor, line 304: @higs4281 A bit of functionality thats been lost with the new elasticsearch solution is, determining the available facets for a group, based on the selected facets in other groups. That would explain why I've been seeing more facets that return no results when testing this branch locally.

cfgov/teachers_digital_platform/models/activity_index_page.py
```diff
@@ -0,0 +1,528 @@
...
+                facet_ids = [
+                    int(value) for value in request.GET.getlist(facet)
+                    if value.isdigit()
+                ]
+                selected_facets[facet] = facet_ids
+                for entry in all_facets[facet]:
+                    if entry["id"] in facet_ids:
+                        entry["selected"] = True
+                        if entry.get("children"):  # a nested facet
+                            for child in entry["children"]:
```
nbucknor, line 142: @higs4281 Wouldn't you need to check if the entry has a parent too? If the entry is selected and has a parent, the parent should either have an "indeterminate" or "checked" status too.

cfgov/teachers_digital_platform/models/activity_index_page.py
```diff
@@ -0,0 +1,528 @@
...
+        verbose_name = "TDP Activity search page"
+
+
+def default_nested_facets(class_object):
+    """Build a nested facet tree, initially only for ActivityTopic objects."""
+    setup = []
+    parents = class_object.objects.filter(parent=None)
+    default_attrs = {
+        "selected": False,
+        "child_selected": False,
```
nbucknor, line 390: @higs4281 It looks like "child_selected" will always be set to False?

higs4281: > child_selected" will always be set to False?

The approach here is to start with a cached facet tree that matches the initial state of the activity index page: No search terms, no facet selections, no queries required.

Then, rather than building a new facet structure every time with Elasticsearch queries, we can adjust the cached default facet tree to match a request.

higs4281:
@nbucknor I think the faceting is very close now to current behavior.

nbucknor:
Thanks @higs4281 

The behavior of the Topics facets is much better now. However, I still see some inconsistencies in the facet behavior.

### Existing behavior on [consumerfinance.gov Activity Search](https://www.consumerfinance.gov/practitioner-resources/youth-financial-education/teach/activities/?q=borrowing&topic=22&grade_level=2&age_range=2&student_characteristics=5)

**Note the available facets that are highlighted in blue:**
Search-for-activities-Consumer-Financial-Protection-Bureau

### New behavior on [current branch](http://localhost:8000/practitioner-resources/youth-financial-education/teach/activities/?q=borrowing&topic=22&grade_level=2&age_range=2&student_characteristics=5) allows for many more facets that yield no results or do nothing

Note that all facets highlighted in **orange** don't actually change to checked when you click them, nor do they affect the available results. The exception being top level topics, which although they do remain checked, they do not refine the result in any way. 

All facets highlighted in **red** causes the page to return no results.

Search-for-activities-Consumer-Financial-Protection-Bureau (1)

higs4281:
Thanks @nbucknor 
I'll keep debugging.

higs4281 (commented):
comment moved

cfgov/teachers_digital_platform/tests/models/test_activity_index_page.py
```diff
@@ -60,51 +66,282 @@ def test_can_create_activity_index_page(self):
...
+        request = HttpRequest()
+        self.assertEqual(validate_page_number(request, paginator), 1)
+        request.GET.update({"page": "2"})
+        self.assertEqual(validate_page_number(request, paginator), 2)
+        request = HttpRequest()
+        request.GET.update({"page": "1000"})
+        self.assertEqual(validate_page_number(request, paginator), 1)
+        request = HttpRequest()
+        request.GET.update({"page": "<script>Boo</script>"})
+        self.assertEqual(validate_page_number(request, paginator), 1)
```
higs4281, line 279: This test (and the function tested) exist in both ask_cfpb and teachers_digital_platform.
We should (in another PR) move the function and test to the search app for sharing.

cfgov/teachers_digital_platform/documents.py
```diff
@@ -0,0 +1,136 @@
...
+    council_for_economic_education = fields.KeywordField()
+    grade_level = fields.KeywordField()
+    jump_start_coalition = fields.KeywordField()
+    school_subject = fields.KeywordField()
+    student_characteristics = fields.KeywordField()
+    teaching_strategy = fields.KeywordField()
+    topic = fields.KeywordField()
+
+    class Index:
+        name = 'teachers-digital-platform'
```
alexm118, line 51: You will need to incorporate the [new helper function](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/search/elasticsearch_helpers.py#L56) to prefix the environment to this index name.

cfgov/teachers_digital_platform/tests/models/test_activity_index_page.py
cwdavies: There might also be a search function or a test in paying_for_college that exist in an other elasticsearch application.
To reduce code any common search functions or tests should be moved to the search application for sharing in that PR.

cwdavies (commented):
Code coverage is reporting 99% which is excellent.
models/activity_index_page.py (98.4%): Missing lines 214,218,237,261

There always seems to be a small number of lines that are reported as missing code coverage in models which are possibly edge cases.

higs4281:
> Code coverage is reporting 99% which is excellent.
> models/activity_index_page.py (98.4%): Missing lines 214,218,237,261
> 
> There always seems to be a small number of lines that are reported as missing code coverage in models which are possibly edge cases.

The coverage for new code is 100%. I don't think the remaining uncovered Haystack lines are worth pursuing, since we will be abandoning that code, and it has never been covered.

For the same reason, I excluded the untested Haystack `search_indexes.py` file from coverage

cfgov/teachers_digital_platform/documents.py
higs4281: > You will need to incorporate the [new helper function](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/search/elasticsearch_helpers.py#L56) to prefix the environment to this index name.

Thanks! [Fixed](873f8d15fea247d78508c39b8c9e28da60d1894b)

cwdavies (approved):
It make sense to exclude old haystack search index code from code coverage

willbarton:
@higs4281 It doesn't have to happen in this PR, but do you think you could write Cypress tests for TDP search based on your learning here?

higs4281:
> @higs4281 It doesn't have to happen in this PR, but do you think you could expand our Cypress tests for TDP search based on your learning here? Or does [what we have](https://github.com/cfpb/consumerfinance.gov/blob/main/test/cypress/integration/pages/tdp-activity-search.js) provide a sufficient amount of coverage of the search?

Our current Cypress tests are pretty minimal. I'll pursue expanding.

cfgov/teachers_digital_platform/models/activity_index_page.py
higs4281: This has been resolved.

Scotchester (approved):
This is working well for me locally, but I have a couple minor questions.

cfgov/teachers_digital_platform/models/activity_index_page.py
```diff
@@ -0,0 +1,592 @@
...
+            sqs = sqs.facet(str(facet), size=facet_config[2])
+            if facet in request.GET and request.GET.get(facet):
+                selected_facets[facet] = [
+                    int(value) for value in request.GET.getlist(facet)
+                    if value.isdigit()
+                ]
+                facet_queries[facet] = facet + '_exact:' + (
+                    " OR " + facet + "_exact:").join(
+                    [str(value) for value in selected_facets[facet]]
+                )
```
Scotchester, line 233: Looks like there isn't a test that hits these lines, if the conditional on 225 is `False`.  Should probably cover these, they seem important.

cfgov/teachers_digital_platform/models/activity_index_page.py
```diff
@@ -0,0 +1,592 @@
...
+            'facet_queries': facet_queries,
+            'all_facets': {},
+        }
+
+        # Apply search query if it exists, but don't apply facets
+        if search_query:
+            sqs = sqs.filter(
+                content=search_query).order_by('-_score', '-date')
+        else:
+            sqs = sqs.order_by('-date')
```
Scotchester, line 249: This line isn't covered by a test.

cfgov/teachers_digital_platform/models/activity_index_page.py
```diff
@@ -0,0 +1,592 @@
...
+        expanded_facets = ALWAYS_EXPANDED.union(set(conditionally_expanded))
+
+        payload.update({
+            'all_facets': all_facets,
+            'expanded_facets': expanded_facets,
+        })
+
+        # Apply all the active facet values to our search results
+        for facet_narrow_query in facet_queries.values():
+            sqs = sqs.narrow(facet_narrow_query)
```
Scotchester, line 273: This line isn't covered by a test.

cfgov/search/elasticsearch_helpers.py
```diff
@@ -1,10 +1,22 @@
...
 
 
+def strip_html(markup):
+    """
+    Make sure markup stripping doesn't mash content elements together.
+
+    Also remove no-break space characters.
+    """
+    clean = re.compile("<.*?>")
+    return re.sub(clean, " ", markup).strip().replace("\xa0", "")
```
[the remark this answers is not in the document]
higs4281: I was hoping we could share this utility where appropriate, with Ask for starters. We were using django strip_tags in ask, and it was mashing together headings and following text (mortgageA being just one example; try searching for mortgageA in prod Ask). Both Ask and TDP also have lots of random NBSP pollution.

cfgov/teachers_digital_platform/jinja2/teachers_digital_platform/activity_search_result.html
```diff
@@ -4,8 +4,8 @@ <h3><a class="activity_title-link" href="{{ result.url }}">{{ result.title }}</a
         <span class="a-date">
             Updated
             <span class="datetime">
-                <time class="datetime_date" datetime="{{ result.date.strftime("%Y-%m-%d") }}">
-                    {{ result.date.strftime("%b %d, %Y") }}
+                <time class="datetime_date" datetime="{{ result.date_attr }}">
+                    {{ result.date }}
```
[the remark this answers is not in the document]
higs4281: They are getting formatted in the [ActivitySetUp](https://github.com/cfpb/consumerfinance.gov/blob/54b7f89f878c088dfda6ba725e1fad9d886a06dc/cfgov/teachers_digital_platform/models/activity_index_page.py#L518-L519). The idea is to prep all these things that don't change, and cache it.

cfgov/teachers_digital_platform/models/activity_index_page.py
higs4281: This is legacy Haystack code that we're not changing or adding tests for. It will be gone soon.

cfgov/teachers_digital_platform/models/activity_index_page.py
higs4281: Ditto from above. This has never been covered by tests, and now is not the time to add them.

cfgov/teachers_digital_platform/models/activity_index_page.py
higs4281: Ditto.

Scotchester:
Sorry about the noise on the missing coverage. I remember the earlier discussion, but had assumed that the addition to the `.coveragerc` file took care of it.

PR #6098, opened 2020-11-04
Configures Synonyms to load directly into indexes

alexm118:
Reads synonyms for english and spanish into our synonym analyzer

---

## Additions

- Adds support for spanish synonyms
- Configures local, test, and deployed synonym paths.

## Changes

- How we load synonyms.

## How to test this PR

1. Rebuild your ES7 indexes, confirm the settings for indexes that use synonyms have all synonyms loaded.
  * This can be done via `manage.py search_index --rebuild --parallel -f`

## Checklist


cfgov/search/models/__init__.py
```diff
@@ -0,0 +1 @@
+from .models import Synonym
```
higs4281, line 1: Views may differ here, but for the tiny `search` app, it seems overkill to have models.py file in a models folder with an __init__ import.

Unless we foresee a big future for search models, I'd vote for simplicity and a single models.py file at /cfgov/search/

cfgov/search/models/models.py
```diff
@@ -0,0 +1,4 @@
+from django.db import models
+
+class Synonym(models.Model):
+    synonym = models.CharField(max_length=500)
```
higs4281, line 4: We may want to consult with content editors, but a `language` field might be handy for editors who want to see synonyms for a particular language.

cfgov/search/elasticsearch_helpers.py
```diff
@@ -25,14 +29,20 @@ def make_safe(term):
     filter=['lowercase', token_filter('ascii_fold', 'asciifolding')]
 )
 
+def get_synonyms():
+    try:
+        return list(Synonym.objects.values_list('synonym', flat=True))
+    except Exception:
```
higs4281, line 35: Do we need a blanket exception here? If we fear there may be no synonym files in the db, we can catch that. But unforeseen errors in a simple database request might be worth raising.

higs4281 (commented):
I like the db approach, and the idea of exposing synonyms to content editors.

I left some comments, but the big question for me is how best to enable editing, which would probably involve adding synonyms to the Wagtail admin via a modeladmin setup.

But I see this as outside of scope for this PR.

cfgov/search/models/__init__.py
alexm118: I can move it

cfgov/search/elasticsearch_helpers.py
alexm118: The problem is this snippet gets run when attempting to apply migrations on the database, and since there isn't a `search_synonym` table, I felt it logical to just safely return an empty list.

higs4281: For that narrow pre-migration case, how about: 

```python
from django.db.utils import ProgrammingError
# ...

def get_synonyms():
    try:
        return list(Synonym.objects.values_list('synonym', flat=True))
    except ProgrammingError:  # if migration has yet to run, no table available
        return []
```

higs4281 (approved):
This is working well for me when I populate the Synonym table and run Ask queries (searches for "basket" and "handbasket" return the same results). I think we'll benefit from having synonyms in the database. 

I would prefer having a language field in the Synonym table, so that the table can become the source of truth for synonyms and we can leave the `search/resources` files behind. 

But we'll probably have to rebuild a lot more than a language field if we decide to enable editing, so we can ride with two sources of truth for a while.

willbarton (approved):
I do like this. 

I agree with @higs4281, I think enabling editing — and thinking through what the data model needs to look like if we do — is out of scope. But I love that this gets us to a place where it's something we can enable.

PR #6102, opened 2020-11-05
Remove unused function references

anselmbradford:
## Removals

- Removes imported functions that don't have a reference in the files they are imported into.

## How to test this PR

1. PR checks should pass. `./frontend.sh` should pass.


cfgov/unprocessed/apps/teachers-digital-platform/js/index.js
```diff
@@ -6,7 +6,6 @@ const cfExpandables = require( '@cfpb/cfpb-expandables/src/Expandable' );
 const scroll = require( './scroll' );
 const search = require( './search' );
```
schbetsy, line 7: (Maybe I don't know how JS works, but) It looks like `search` is also imported but not referenced in this file.

schbetsy (approved):
Looks good to me. Added an optional comment about another import that might be okay to remove.

cfgov/unprocessed/apps/teachers-digital-platform/js/index.js
anselmbradford: I noticed this one as well, but it does appear to be used. In JS it is possible to require a module and then handle initialization within that module itself. It appears to initialize itself here https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/apps/teachers-digital-platform/js/search.js#L266
We should maybe discourage this pattern since it's ambiguous.

PR #6104, opened 2020-11-09
Remove unused template: textarea.html

schbetsy:
Looking at the code in this repo, it appears the `textarea` template is never used. This PR removes it.

Double check me here. Is there a way it could somehow be needed even though this repo never uses it?

## Removals

- `atoms/textarea.html`

## Checklist


Scotchester (approved):
I think that a lack of results searching for `textarea.html` is sufficient proof that it's not in use.

It used to be used on the conference registration form, but (1) it no longer is, and (2) we're using Eventbrite for that these days, anyway.

schbetsy:
Thanks for reviewing!

PR #6106, opened 2020-11-13
Consistent IDs for form inputs

schbetsy:
The new `unique_id_in_context` method will replace the use of `get_unique_id` method to generate unique identifiers for form inputs. `get_unique_id` would generate a random sequence of characters on each page load, meaning the input id would always change. This caused an issue for our web archive, since it caused unnecessary diffs on pages that didn't have substantive changes.

With `unique_id_in_context`, the markup for input forms should remain consistent between page loads, which will allow the web archive to be more useful and not contain superfluous diffs.

## Additions

- `unique_id_in_context()` method and jinja tag

## Changes

- Update several templates to use `unique_id_in_context()` instead of `get_unique_id()`

## How to test this PR

1. Run this branch
2. Visit a page that contains an email sign-up form, such as http://localhost:8000/owning-a-home/
3. Use the inspector to verify that the `label` and `input` in the HTML of the email sign-up form use single-digit identifiers, like in the screenshot below. This should also be true in the email pop-up form.

Screen Shot 2020-11-13 at 12 01 13 PM

## Notes and todos

- Some templates still use the `get_unique_id()` method. We will replace those with `unique_id_for_context()` in an upcoming PR, then remove `get_unique_id` as a jinja tag.

## Checklist

### Front-end testing
No visual changes.

### DIFF
--- a/cfgov/v1/jinja2tags/__init__.py
+++ b/cfgov/v1/jinja2tags/__init__.py
@@ -98,6 +98,36 @@ def render_stream_child(context, stream_child):
     return Markup(unescaped)
 
 
+def unique_id_in_context(context):
+    """Return an ID that is unique within the given context
+
+    For a given request, return a unique ID each time this method is
+    called. The goal is to generate IDs to uniquely identify elements
+    in a template that are consistent between page loads.
+
+    If the context has a request object, the generated id will increment:
+
+    >>> context = {'request': request}
+    >>> unique_id_in_context(context)  # returns 1
+    >>> unique_id_in_context(context)  # returns 2
+    >>> unique_id_in_context(context)  # returns 3
+
+    If the context lacks a request, this function will return a 14-character
+    unique alphanumeric string.
+    """
+    request = context.get('request')
+    if request:
+        attribute_name = '__last_unique_id'
+        if not hasattr(request, attribute_name):
+            setattr(request, attribute_name, 0)
+        id = getattr(request, attribute_name) + 1
+        setattr(request, attribute_name, id)
+        return id
+
+    else:
+        return get_unique_id()
+
+
 class V1Extension(Extension):
     def __init__(self, environment):
         super(V1Extension, self).__init__(environment)
@@ -118,6 +148,7 @@ class V1Extension(Extension):
             'is_report': ref.is_report,
             'is_filter_selected': contextfunction(is_filter_selected),
             'render_stream_child': contextfunction(render_stream_child),
+            'unique_id_in_context': contextfunction(unique_id_in_context),
             'app_url': app_url,
             'app_page_url': app_page_url,
         })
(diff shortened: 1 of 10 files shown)


cfgov/v1/jinja2tags/__init__.py
```diff
@@ -98,6 +98,34 @@ def render_stream_child(context, stream_child):
...
+    If the context lacks a request, this function will return a 14-character
+    unique alphanumeric string.
+    """
+    request = context.get('request')
+    if request:
+        attribute_name = '__unique_input_ids'
+        if not hasattr(request, attribute_name):
+            setattr(request, attribute_name, 0)
+        id_count = getattr(request, attribute_name)
+        setattr(request, attribute_name, id_count + 1)
```
chosak, line 122: I would expect that the request would contain the last number used (for example, 0 if one call, 1 if two calls, etc). This might be helpful to debug any issues because at any moment the request object would represent the last ID used, although in that case I might expect the attribute name to be something like `__last_unique_id`.

If you'd prefer to leave the logic as is, could we change the attribute name to something like `__next_unique_id`?

It seems like we don't necessarily need to include the word `input` in the attribute name (or mention "input fields" in the docstring above) since in theory one might want to use this template tag on some other kind of element needing a unique ID (like a `<div>`).

cfgov/v1/tests/jinja2tags/test_jinja2tags.py
```diff
@@ -95,3 +97,58 @@ def test_query_parameter_archived_selected(self):
...
+            '0 and 1'
+        )
+
+    def test_multiple_renders_multiple_unique_ids(self):
+        request = HttpRequest()
+        rendered = [
+            self.render(self.template, {'request': request})
+            for _ in range(5)
+        ]
+        self.assertEqual(rendered, [ '0', '1', '2', '3', '4'])
```
chosak, line 147:
```suggestion
        self.assertEqual(rendered, ['0', '1', '2', '3', '4'])
```

I'm surprised that the linter didn't fail on this. Looks like it [ignores the v1 tests directory](https://github.com/cfpb/consumerfinance.gov/blob/e34cc78b3de4627a23c6a981901a2da2da65c785/.flake8#L40). 😳

cfgov/v1/jinja2tags/__init__.py
```diff
@@ -98,6 +98,34 @@ def render_stream_child(context, stream_child):
     return Markup(unescaped)
 
 
+def unique_id_in_context(context):
+    """For a given request, return a unique ID each time this method is
```
chosak, line 102: Per [the docstrings PEP](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings), would you mind changing this so that it consists

> of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description.

chosak (approved):
💯

PR #6109, opened 2020-11-16
Setup configuring index name via env variable

alexm118:
In order to avoid using the same index across all DEV or EXT servers, we can use an environment variable to configure the name of our index.

---

## Additions

- ES_INDEX_PREFIX environment variable to add a prefix to our ES index.

## How to test this PR

1. Rebuild your elasticsearch indexes after setting up the .env `ES_INDEX_PREFIX`. 
`./manage.py search_index --rebuild --parallel -f`

## Checklist


.env_SAMPLE
```diff
@@ -91,6 +91,7 @@ export ES_PORT=9200
 export ES_HOST=elasticsearch2
 export ES7_HOST=elasticsearch
 export SHEER_ELASTICSEARCH_INDEX=content
+#export ES_INDEX_PREFIX=local-
```
Scotchester, line 94: Any reason not to just make this on by default?

Scotchester (approved):
One trivial question, but otherwise looks good.

.env_SAMPLE
alexm118: I recalled us talking about using no prefix for local and production, so I excluded it by default but wanted to leave the comment to indicate how it could be set. We can use it locally, it doesn't change the interaction at all codewise. Personally I like local matching prod so I can copy the curl commands from kibana without adjusting the index but thats negligible gain.

Scotchester: Makes sense :+1: Leave as-is.

higs4281 (commented):
I would prefer "SEARCH_INDEX_PREFIX" for a variable name, to avoid confusion with all the 'es' and 'ES' Spanish name-spacing in our code base.

Scotchester:
That's a great point, @higs4281.

cfgov/cfgov/settings/base.py
```diff
@@ -468,6 +468,9 @@
     DEFAULT_FILE_STORAGE = "storages.backends.s3boto3.S3Boto3Storage"
     MEDIA_URL = os.path.join(os.environ.get("AWS_S3_URL"), AWS_LOCATION, "")
 
+# Prefix for our elasticsearch indexes based on the environment
+ES_INDEX_PREFIX = os.environ.get('ES_INDEX_PREFIX', '')
```
chosak, line 472: Could we use the existing `settings.DEPLOY_ENVIRONMENT` for this instead of making a new setting/environment variable?

cfgov/ask_cfpb/documents.py
```diff
@@ -35,7 +37,7 @@ def prepare_url(self, instance):
         return instance.url
 
     class Index:
-        name = 'ask-cfpb'
+        name = name = f'{settings.ES_INDEX_PREFIX}ask-cfpb'
```
chosak, line 40: It'd be nice not to have to duplicate this each time we define a new index. What would you think about doing something like (completely untested):

```py
# somewhere central
class EnvironmentSpecificIndex:
    @property
    def name(self):
        return settings.ES_INDEX_PREFIX + self.name_no_prefix

# each time we define an index
@registry.register_document
class MyDocument(Document):
    ...

    class Index(EnvironmentSpecificIndex):
        name_no_prefix = 'ask-cfpb'
        ...
```

This way we'd still have to remember to use EnvironmentSpecificIndex but maybe that's more easily documented and less error-prone than copying the prefix each time?

alexm118: Tried the class approach, but looks like the way the dsl library access the name attribute wasn't working with the property. Opted instead to move it into a central function `environment_specific_index`. The name is a working title, but I think this addresses reusability concerns. We also wanted both local and production indexes to have no prefix, so I added the little snippet of logic around handling that.

cfgov/search/elasticsearch_helpers.py
```diff
@@ -49,3 +51,10 @@ def get_synonyms():
         synonynm_filter,
         'lowercase'
     ])
+
+
+def environment_specific_index(base_name):
+    if settings.DEPLOY_ENVIRONMENT == 'production' or settings.DEPLOY_ENVIRONMENT == 'local':  # noqa: E501
+        return base_name
+    else:
+        return f'{settings.DEPLOY_ENVIRONMENT}-{base_name}'
```
chosak, line 60: If `settings.DEPLOY_ENVIRONMENT` is unset, should this return the base name without the leading dash?

alexm118: I would think so

cfgov/search/elasticsearch_helpers.py
```diff
@@ -49,3 +51,10 @@ def get_synonyms():
         synonynm_filter,
         'lowercase'
     ])
+
+
+def environment_specific_index(base_name):
+    if settings.DEPLOY_ENVIRONMENT == 'production' or settings.DEPLOY_ENVIRONMENT == 'local' or settings.DEPLOY_ENVIRONMENT is None:  # noqa: E501
```
chosak, line 57:
```suggestion
    if settings.DEPLOY_ENVIRONMENT in (None, '', 'local', 'production'):
```

PR #6112, opened 2020-11-18
Bump mapbox version to 3.3.1 and update CSP

wpears:
Moves fahc off deprecated (and now non-functional) mapbox styles


cfgov/cfgov/settings/base.py
```diff
@@ -649,7 +649,7 @@
     "'self'",
     "*.consumerfinance.gov",
     "*.google-analytics.com",
-    "*.tiles.mapbox.com",
```
ascott1, line 653: Are you certain that it's ok to remove this across the site?

ascott1 (approved):
One minor comment, but otherwise :+1:

cfgov/cfgov/settings/base.py
wpears: Pretty sure, but as I'd prefer to get this in quickly before running a full audit, I added it back.

PR #6113, opened 2020-11-18
Include all answer_content text content in search index

Scotchester:
Standardizing on the use of `get_searchable_content` so we don't have to name individual blocks, which isn't very future-proof. We were already not indexing HowTo and FAQ schema blocks, which was the main driver of this work.

## Changes

- Updates the `AnswerPage.text()` method that populates the ES7 index to call a single `answer_text()` method and prepend that with the `question` text
- `answer_text()` rolls up all of the functionality previously found in `short_answer_data()` and `answer_content_text()` (as well as `extract_raw_text()`, which `answer_content_text()` also called)
- Renamed `answer_content_data()` to `answer_content_preview()` to make its purpose more clear

## How to test this PR

1. Pull branch
1. `docker-compose up`
1. Load up an AnswerPage in the Wagtail admin and add a bunch of junk to its content StreamField, including something with a unique searchable term that won't already be in the DB, and including either a HowTo or FAQ schema block
1. `docker-compose exec python bash`
1. `./cfgov/manage.py search_index --rebuild -f --models ask_cfpb`
1. Open the Kibana console: http://localhost:5601/app/kibana#/dev_tools/console
1. Run this query, replacing `banana` with whatever unique term you added above:
   ```json
   GET /ask-cfpb/_search
   {
     "query": {
       "match": {
         "text": "narrator"
       }
     }
   }
   ```
1. Observe that the results return the page you edited, and that the `text` field includes the content you added in all blocks

## Notes and todos

- ~~I think the remaining uses of `extract_raw_text` can be removed, but I wanted to confirm with folks more familiar with the initial implementation. Will add a comment about this shortly.~~

## Checklist


cwdavies (approved):
These changes to ask_cfpb AnswerPageDocument and AnswerPage models look good to me.
Can the remaining uses of extract_raw_text be removed now in this same PR?

Scotchester:
> Can the remaining uses of `extract_raw_text` be removed now in this same PR?

Oops, realizing now that I didn't follow up with the promised comment about `extract_raw_text`.

OK, so there's one use of it here in `pages.py`:

https://github.com/cfpb/consumerfinance.gov/blob/a3095e43b408109f59d5ad7fd93b9e11d9d3a83c/cfgov/ask_cfpb/models/pages.py#L82-L86

Comment says it's used for the search result previews, but that can't be true because the changes I made already change the search result previews.

The above `get_answer_preview` function is called down below within the `TagResultsPage`'s `tag_search` method:

https://github.com/cfpb/consumerfinance.gov/blob/a3095e43b408109f59d5ad7fd93b9e11d9d3a83c/cfgov/ask_cfpb/models/pages.py#L480-L529

We have but one `TagResultsPage` on the site, which is displays search results when given a "tag" in its URL: https://www.consumerfinance.gov/ask-cfpb/search-by-tag/mortgages/

The search results on that page differ from what can be seen by using the regular Ask search and searching for `mortgages`: https://www.consumerfinance.gov/ask-cfpb/search/?q=mortgages

@higs4281 confirms for me that `search-by-tag` is "for old (and some not-so-old) links, but we don't have other navigation to them".

So, I think the conclusion to this long stream-of-consciousness exploration is that the code in `TagResultsPage.tag_search` also needs to be updated to use the new methods introduced in this PR, and then `extract_raw_text` can go away.

---

P.S.: There is a _duplicate_ `extract_raw_text` in the old ES2 implementation in [`ask_cfpb/search_indexes.py`](https://github.com/cfpb/consumerfinance.gov/blob/a3095e43b408109f59d5ad7fd93b9e11d9d3a83c/cfgov/ask_cfpb/search_indexes.py#L19-L32), which we can leave as is and remove with the rest of the ES2 cleanup.

higs4281:
> I think the conclusion to this long stream-of-consciousness exploration is that the code in `TagResultsPage.tag_search` also needs to be updated to use the new methods introduced in this PR, and then `extract_raw_text` can go away.

👍

higs4281 (approved):
This is a nice improvement, and search picked up all the "junk" I could throw at it. Removing `get_answer_preview` will be a bonus.

PR #6114, opened 2020-11-19
Exclude Haystack search_indexes.py file from coverage

cwdavies:
I don't think the remaining uncovered Haystack lines are worth pursuing, since we will be abandoning that code, and they may have never been covered. For this, I have excluded the untested Haystack search_indexes.py file from code coverage.


higs4281 (commented):
I think this is unnecessary. The search_indexes.py file is already fully tested, so we can leave this file and its tests alone, and delete them once we move to Elasticsearch7.

cwdavies:
This change is not necessary, but the search index file will be deleted once the move to elasticsearch 7 has been deployed to production.

PR #6127, opened 2020-11-24
IA Migration: Script to update outdated Wagtail redirects

schbetsy:
The IA migration adds new redirects to Apache's redirects.conf. These make some existing Wagtail redirects outdated. This script fixes the ones we know about immediately. After the migration, we can do a more complete accounting of which redirects are still legitimate.

## Additions

- `update_existing_redirects` script to correct the set of outdated redirects that causes smoke test failures

## Notes and todos

- This is a hotfix to get the IA Migration out the door. Once that deadline has passed, we will take more time to create a more effective system around managing redirects and ensuring that all of them are effective.

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [ ] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [ ] Future todos are captured in comments and/or tickets
- [ ] Project documentation has been updated

PR #6128, opened 2020-11-25
Fix incorrect webpack exclude regex

anselmbradford:
Webpack was not transpiling `@cfpb` node modules coming from the design system.

## Changes

- Fix incorrect webpack exclude regex to properly include `@cfpb` node modules in webpack transpiling.

## How to test this PR

1. Pull branch and `gulp clean && gulp build`.
2. Built files in `cfgov/static_built/js/` should not contain `const`.


Scotchester (approved):
Approved, but please hold for merging after #6127

PR #6134, opened 2020-12-01
Only install gettext as needed for GitHub Actions

chosak:
Currently we install gettext on all of our backend GitHub Actions runs, even though it is only needed for the 'validate-translations' matrix run. This change modifies the configuration to only install gettext as needed, which should speed up the other matrix runs.

## Notes and todos

This logic was added in https://github.com/cfpb/consumerfinance.gov/pull/5939.

## Checklist

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -34,9 +34,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      # Make sure we have gettext installed for the translation validation
-      - run: |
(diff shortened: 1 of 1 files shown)


Scotchester:
I think you might have a syntax issue, given that all Actions are stalled?

chosak:
@Scotchester GHA is having issues globally: https://www.githubstatus.com/

image

higs4281 (approved):

PR #6136, opened 2020-12-02
Update asynchronous loading of fonts CSS

anselmbradford:
Before:
Screen Shot 2020-12-02 at 9 54 41 AM

See solution in https://stackoverflow.com/questions/32759272/how-to-load-css-asynchronously/46750893#46750893

**Note: I didn't include the optional preloading. I think we'd move this to the head if we decided to use that.**

## Changes

- Update font CSS loading behavior to avoid incorrect `media="none"` value.

## How to test this PR

1. Checking the network panel, the fonts.net CSS should not take longer to load than on production.


Scotchester (approved):
Awesome! Thanks for finding this updated method.

PR #6139, opened 2020-12-02
Add notice and consent warning, server indicator banner

Scotchester:
The primary purpose of this PR is to add the CFPB's "notice and consent" warning to comply with [NIST Special Publication 800-53 (Rev. 4) control AC-8](https://nvd.nist.gov/800-53/Rev4/control/AC-8).

More interestingly, I took the opportunity to reduce the customization of our custom login page template and bring it up to date with the latest in the current release of Wagtail. This includes removing our customized heading logic, which I suspect never actually clues anyone in to the fact that they're using Build. To compensate, though, I have implemented a little banner at the top of the left-hand menu (or all the way across the top on narrower viewports) that highlights what server you're on. 

---

## Additions

- Notice and consent warning added to login page
- Server indicator banner added to all pages

## Removals

- Some of the customization in our custom login template

## How to test this PR

1. Pull branch
1. `gulp scripts:admin`
1. Load up http://localhost:8000/admin/ and look at the login page. Log in to see the banner in its idea placement at the top of the sidebar.

## Screenshots

Login screen with notice and consent warning

Server banner on a wide viewport

Server banner on a narrow viewport

## Notes and todos

- [x] Confirm with Gerald that this text, which I copied from our phone lock screens, is acceptable.
- At the moment, the banner is always the same color, but it might be nice to have it be color-coded for our common environments.

## Checklist

#### Browser testing

- [ ] Firefox
- [x] Chrome
- [ ] Safari
- [ ] Safari on iOS
- [ ] Chrome on Android

#### Accessibility

- [ ] Screen reader friendly

#### Other


willbarton:
>  I have implemented a little banner at the top of the left-hand menu (or all the way across the top on narrower viewports) that highlights what server you're on.

raw

niqjohnson (commented):
@Scotchester, I had a couple of questions and thoughts, none of which are particularly critical.

First, for the disclaimer text, do you know if it needs to appear before the login form to satisfy AC-8? Clearly no one is going to read this, so I'd rather have it as out of the way as possible. Given that we're trying to keep modifications to the base login template to a minimum, I think stuffing it in the `below_login` block so it shows up after the login form would be preferable (maybe with a 1 em top margin so it doesn't sit too tight with the button). But I'm not sure if there are rules around where that message has to appear to satisfy the control.

Second, I noticed a couple of minor things on the login page:

1. The heading has changed from "Log in to consumerfinance.gov" to "Sign in to Wagtail"; was that intentional to keep the changes from the base template to a minimum? I actually prefer the "Sign in to Wagtail" heading (especially coupled with the new environment banner), but I wasn't sure if there was some sort of fraught history of what that heading has to say that we need to preserve.
2. It looks like the "Forgotten it" password reset link (specifically the `.help` element) is absolutely positioned from the top of `.content-wrapper`, which puts it in a weird spot with the addition of the disclaimer text (and totally out of the form if we move the disclaimer text to `below_login`). Is that worth reworking so it's positioned to the password form field instead? Or is that too much customization?

Third, I *love* the environment banner! I often have the same page open to compare in different environments and constantly get confused which tab is which. A couple of things that jumped out at me for the banner:

1. Should we make that a color in the Wagtail color scheme at `/admin/styleguide/#palette`? Maybe something that stands out but is not alarming, like salmon or orange? Or is that styleguide outdated?
2. I wonder if we should have that banner appear on every environment but production? I think we have a contingent of editors who only work in prod and who might not need a banner like that to help them know where they're working since they only work in one place. I can ask around if that's a question we want to investigate.
3. This is minor, but I find the combination of bold + 1 px letter spacing makes the environment name a little too airy and hard to read. Maybe 0.5 px letter spacing instead? Does that feel too tight?
4. This is even more minor, but I happened to be testing this out on 127.0.0.0 instead of localhost, and "127" in the banner threw me for a second. I have to imagine that anyone savvy enough to run cf.gov locally can figure out what "127" means there, though, so probably not something we need to change. If we do want to change it, though, maybe we can map all the possible local environment names to display "localhost"?

Scotchester:
> First, for the disclaimer text, do you know if it needs to appear before the login form to satisfy AC-8?

Good news! I asked Gerald about this and he confirmed it would be fine. Did find out that we need to change the button text to "Accept and sign in", though.

> 1. The heading has changed from "Log in to consumerfinance.gov" to "Sign in to Wagtail"; was that intentional to keep the changes from the base template to a minimum? I actually prefer the "Sign in to Wagtail" heading (especially coupled with the new environment banner), but I wasn't sure if there was some sort of fraught history of what that heading has to say that we need to preserve.

Yep, that's why I did it. The reason we customized it in the first place was to more prominently indicate to people if they were signing into Build vs. any other Wagtail location, but in practice I really doubt anyone notices.

> 2. It looks like the "Forgotten it" password reset link (specifically the `.help` element) is absolutely positioned from the top of `.content-wrapper`, which puts it in a weird spot with the addition of the disclaimer text (and totally out of the form if we move the disclaimer text to `below_login`). Is that worth reworking so it's positioned to the password form field instead? Or is that too much customization?

Good catch! Thank you, I hadn't noticed this. Pushed a fix, which unfortunately _added_ to the customization of the template, but I should PR this change (making the link be positioned relative to the password field) back to Wagtail for everyone's benefit.



> 1. Should we make that a color in the Wagtail color scheme at `/admin/styleguide/#palette`? Maybe something that stands out but is not alarming, like salmon or orange? Or is that styleguide outdated?

Yeah, I think using color(s) from the Wagtail style guide is a good idea. Some thoughts on differentiating the colors:
- Content: Salmon
- Beta: Orange
- Build: Teal Dark
- dev2–5/Docker stacks: Blue
- localhost: Hmm… Grey 5? Or venture out of the Wagtail style guide and leave it CFPB purple as already specified here

> 2. I wonder if we should have that banner appear on every environment but production? I think we have a contingent of editors who only work in prod and who might not need a banner like that to help them know where they're working since they only work in one place. I can ask around if that's a question we want to investigate.

If it doesn't _bother_ them, I'd prefer to leave it for the benefit of those of us that work across all these environments.

> 3. This is minor, but I find the combination of bold + 1 px letter spacing makes the environment name a little too airy and hard to read. Maybe 0.5 px letter spacing instead? Does that feel too tight?

Can do!

> 4. This is even more minor, but I happened to be testing this out on 127.0.0.0 instead of localhost, and "127" in the banner threw me for a second. I have to imagine that anyone savvy enough to run cf.gov locally can figure out what "127" means there, though, so probably not something we need to change. If we do want to change it, though, maybe we can map all the possible local environment names to display "localhost"?

If it's not a big deal, I'd prefer not to add the extra JS that would be needed to do this.

Scotchester:
Haven't actually pushed the link positioning fix I referred to because I've also got color changes that I haven't separated out yet.

Scotchester:
OK, @niqjohnson, pushed a commit that fixes the positioning and adds the individual colors we discussed via chat! Ready for final review.

cfgov/templates/wagtailadmin/css/global.css
```diff
@@ -0,0 +1,52 @@
...
+[data-env="build"]:before {
+    background-color: #71b2d4; /* Blue */
+    color: #000;
+}
+
+[data-env="dev2"]:before,
+[data-env="dev3"]:before,
+[data-env="dev4"]:before,
+[data-env="dev5"]:before,
+[data-env^="cfgov-"] /* Dev Docker stacks */ {
```
niqjohnson, line 45:
```suggestion
[data-env^="cfgov-"]:before /* Dev Docker stacks */ {
```

Just a missing `:before`, otherwise all the colors look great.

niqjohnson (changes requested):
Just one small CSS change; everything else looks amazing. I'm super excited for the environment banner.

cfgov/templates/wagtailadmin/css/global.css
Scotchester: Good catch; thanks!

niqjohnson (approved):
Looks great, thanks!

PR #6142, opened 2020-12-03
Mega Menu: Convert unordered list label from aria-label to title

anselmbradford:
HTML validator says the aria-label on `ul` element in the mega menu may be an error:

Screen Shot 2020-12-02 at 10 03 15 AM

https://fae.disability.illinois.edu/rulesets/LIST_2/ says 

> The title attribute can also be used to add a label to a list container element to provide an explicit text description of its contents.

So this PR turns the `aria-label` into `title` attributes.

## Changes

- Change mega menu `ul` `aria-label` into `title` attributes.

## How to test this PR

1. Run voiceover on the menu and see that the list has a title

Screen Shot 2020-12-02 at 11 56 53 AM

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
+++ b/cfgov/jinja2/v1/_includes/organisms/mega-menu.html
@@ -67,7 +67,7 @@
                 {% if aria_group_title %}
                 aria-labelledby="{{ aria_group_title | slugify ~ '-menu' }}"
                 {% elif nav_title %}
-                aria-label="{{ nav_title | safe }}"
+                title="{{ nav_title | safe }}"
                 {% endif %}>
                 {% if nav_depth == 1 -%}
                     {% set complaint_url = '/es/enviar-una-queja/' if language == 'es' else '/complaint/' %}


Scotchester:
The reference you linked seems to _prefer_ `aria-label` to `title`. I think I'd rather leave it that way and let the validator catch up. Maybe we should file an issue report, as suggested.

cfgov/jinja2/v1/_includes/organisms/mega-menu.html
```diff
@@ -67,7 +67,7 @@
                 {% if aria_group_title %}
                 aria-labelledby="{{ aria_group_title | slugify ~ '-menu' }}"
                 {% elif nav_title %}
-                aria-label="{{ nav_title | safe }}"
+                title="{{ nav_title | safe }}"
```
Scotchester, line 70: Would also be nice to update this to indicate that this is one of multiple lists under the same section.

```suggestion
                title="{{ nav_title | safe }} group {{ loop.index }} of {{ loop.length }}"
```

anselmbradford:
> The reference you linked seems to prefer aria-label to title. 

Are you saying that because `aria-label` appears before `title` in the bullets or are you reading something else?

From https://www.w3.org/TR/wai-aria/#aria-label

> Most host languages provide an attribute that could be used to name the element (e.g., the title attribute in HTML), yet this could present a browser tooltip. In the cases where a visible label or visible tooltip is undesirable, authors MAY set the accessible name of the element using aria-label.

So the question may be whether we want the tooltip or not.

Scotchester:
> > The reference you linked seems to prefer aria-label to title.
> 
> Are you saying that because `aria-label` appears before `title` in the bullets or are you reading something else?

Yep. The `title` is presented as an alternative option.
 
> From https://www.w3.org/TR/wai-aria/#aria-label
> 
> > Most host languages provide an attribute that could be used to name the element (e.g., the title attribute in HTML), yet this could present a browser tooltip. In the cases where a visible label or visible tooltip is undesirable, authors MAY set the accessible name of the element using aria-label.
> 
> So the question may be whether we want the tooltip or not.

Interesting! I think we'd certainly prefer no tooltip, since the visual structure is already clear if one can see it.

anselmbradford:
👍 Thanks Scott! I opened an issue here https://github.com/validator/validator/issues/1054 and am closing this issue.

PR #6145, opened 2020-12-04
Qualifies the Apache ServerName directive with https

wpears:
When resolving redirects, Apache checks the scheme of the current request and returns that in the `Location` header. However, because we terminate SSL before Apache, this scheme is `http`, resulting in an unnecessary redirect before Apache redirects users to the right place, see https://github.local/CFPB/super-regular/issues/225

However, with a scheme set in the ServerName directive, Apache will instead always use that scheme.

I initially thought we'd have to do some parameterization around different servernames, (dev servers, content, etc) but that doesn't seem to be the case. Indeed, the ServerName was previously just set to `consumerfinance.gov` for all environments, so anything that was relying on this matching the actual server domain name wouldn't have worked anyhow except in prod (and even then, it wouldn't match the www subdomain).

This is live on dev4 to test redirect behavior or other possible fallout from this change that I'd missed.


chosak:
TIL that `ServerName` [supports scheme](https://httpd.apache.org/docs/2.4/mod/core.html#servername).

I don't quite understand how this works. I see you've deployed this to internal DEV4 server, and this happens:

```sh
$ curl -I https://internal-dev4-servername/owningahome
HTTP/1.1 301 Moved Permanently
Location: https://internal-dev4-servernamev/owning-a-home
```

But because we don't currently set the Apache [`UseCanonicalName`](https://httpd.apache.org/docs/2.4/mod/core.html#usecanonicalname) parameter, it defaults to off, which the docs suggest

> Apache httpd will form self-referential URLs using the hostname and port supplied by the client if any are supplied (otherwise it will use the canonical name, as defined above)

If you provide another valid host header ("hostname and port supplied by the client"), it gets used instead:

```sh
curl -I -H 'Host: a-different-internal-dev-servername' https://internal-dev4-servername/owningahome
HTTP/1.1 301 Moved Permanently
Location: http://a-different-internal-dev-servername/owning-a-home
```

Without a host header, Apache should use the "canonical name" which is the name in `ServerName`. But if this code is deployed to DEV4, why isn't it returning www.consumerfinance.gov?

wpears:
>Without a host header, Apache should use the "canonical name" which is the name in ServerName. But if this code is deployed to DEV4, why isn't it returning www.consumerfinance.gov?

I think you're being tricked by default behavior. [The spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23) says HTTP/1.1 requests MUST carry a Host header and, indeed, browsers and curl will set one by default, matching the domain name of the request target (you can see this in curl by setting the `-v` option and noting the `> Host:` header getting set.

Indeed, removing this header manually by running `curl -I -H 'Host:' https://internal-dev4-servername/owningahome` will result in a `400`.

Under this behavior, this means with `UseCanonicalName` off and no manual `Host` setting shenanigans we'll get URLs resolved as we desire (and with the proper scheme once this PR is merged).

For this bit:
>If you provide another valid host header ("hostname and port supplied by the client"), it gets used instead:

>```
>curl -I -H 'Host: a-different-internal-dev-servername' https://internal-dev4-servername/owningahome>
>HTTP/1.1 301 Moved Permanently
>Location: http://a-different-internal-dev-servername/owning-a-home
>```

I believe this is only possible because of the wildcard cert we have set for our internal dev server names(`*.otherstuff.gov`), indeed, rerunning the above command with the host set to www.consumerfinance.gov you'll get a 404:

```
curl -I -H 'Host: www.consumerfinance.gov' https://internal-dev4-servername/enforcement/
HTTP/1.1 404 Not Found
Date: Wed, 09 Dec 2020 18:03:46 GMT
Server: Apache/2.4.18 (Red Hat)
Connection: close
Content-Type: text/html; charset=iso-8859-1
```

And more importantly, curling cf.gov with an alternate Host proceeds as follows:

```
curl -I -H 'Host: internal-dev4-servername'  https://www.consumerfinance.gov
HTTP/2 400
server: AkamaiGHost
mime-version: 1.0
content-type: text/html
content-length: 208
expires: Wed, 09 Dec 2020 18:45:38 GMT
date: Wed, 09 Dec 2020 18:45:38 GMT
```

So I think that covers everything?

chosak (approved):
Thanks for the detailed response @wpears. Let's give it a shot! Test on beta first, and then on prod?

schbetsy (approved):
This is working for me on DEV4. I agree we should try it.

PR #6150, opened 2020-12-07
Remove Haystack code, flag references, index file, and tests from PFC

cwdavies:
Elasticsearch7 is live, so we can remove the Haystack code, files, tests and Flag from PFC.

## How to test this PR
- Run unittest-current and check 100% coverage
```
tox -e unittest-current paying_for_college
coverage report -m | egrep "Stmts|paying_for_college"
```

Check out the search page (Docker setup) and make sure search works as expected.
<http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/api/search-schools.json?q=Harvard>


higs4281:
The coverage report indicates that `class ElasticSchool` (lines 21-27 in test_search.py) isn't being used, so I think that declaration  can be dropped.

cfgov/cfgov/settings/base.py
```diff
@@ -767,7 +767,7 @@
     # Enable django-elasticsearch-dsl and disable haystack in regulations3k.
     "ELASTICSEARCH_DSL_REGULATIONS": [("boolean", False)],
     # Enable django-elasticsearch-dsl and disable haystack in paying_for_college.
-    "ELASTICSEARCH_DSL_PFC": [("boolean", False)],
+    "ELASTICSEARCH_DSL_PFC": [("boolean", True)],
```
higs4281, line 770: I think we should remove the flag, now that it's not being used.

cwdavies: Agreed and I removed the DSL_PFC flag.

higs4281 (approved):

PR #6157, opened 2020-12-11
Remove Haystack (and ES2) from regulations3k code

cwdavies:
Removes all code and tests around regulations3k integrating with haystack.

---

## Removals

- Code and tests around django-haystack for regulations3k

## Changes

- replaced the original regulations3k functions with their ES7 implementation.

## How to test this PR

1. Load ES 7 index (`./manage.py search_index --rebuild --parallel -f`)
2. Run Cypress Tests

## Checklist


higs4281 (changes requested):
We can't get rid of update_regulation_index.py, because that's what updates the paragraph corpus that we search. 

We just need to convert it to run the DSL index command instead of Haystack.

cwdavies:
> We can't get rid of update_regulation_index.py, because that's what updates the paragraph corpus that we search.
> 
> We just need to convert it to run the DSL index command instead of Haystack.

I restored the file and changed the Haystack index `update` command to the DSL index `rebuild` command.
I am not sure if this is the correct way to update regulation indexes.
Were there Jenkins jobs that have been modified to update regulation indexes?

cfgov/regulations3k/management/commands/update_regulation_index.py
```diff
@@ -9,15 +9,15 @@
 logger = logging.getLogger(__name__)
 
 
-def _run_haystack_update():
-    """Update the Haystack index after prepping section paragraphs."""
-    call_command('update_index', 'regulations3k', '--remove')
+def _run_elasticsearch_rebuild():
+    """Rebuild the Elasticsearch index after prepping section paragraphs."""
+    call_command('search_index', 'regulations3k', '--rebuild -f --parallel')
```
higs4281, line 14: Indexing one model in DSL requires a `--models` arg, and since it's one app, we probably don't need parallel.

```suggestion
    call_command('search_index', '--rebuild', '-f', '--models', 'regulations3k')
```

cwdavies: Added `--models` argument and removed `parallel` argument

higs4281:
Could we restore the bits of test_search_indexes.py that test the management command? 
It's still useful to check paragraph generation and the calls the command makes.

higs4281:
Not related to this PR, but since we're adjusting tests, could we delete the `settings.py` file at `regulations3k/tests/`?
It isn't used.

cwdavies:
> Not related to this PR, but since we're adjusting tests, could we delete the `settings.py` file at `regulations3k/tests/`?
> It isn't used.

Deleted settings.py and verified regulations3k tests still run without it.

cwdavies:
> Could we restore the bits of test_search_indexes.py that test the management command?
> It's still useful to check paragraph generation and the calls the command makes.

Restored `test_search_indexes.py` to test management command.

higs4281:
> Were there Jenkins jobs that have been modified to update regulation indexes?

The management command is preserved with the same name, so any jobs using it would be unaffected.
The Jenkins job that updates all Django ES7 indexes doesn't use the regs3k management command.

higs4281 (approved):
Looking good to me. Great job!

Scotchester:
Please hold merging. Have a high-priority production bug fix to get out this afternoon.

Scotchester:
Good to go now.

PR #6159, opened 2020-12-11
Have Jenkins report PR build/functional test status to GitHub

willbarton:
This changes adds GitHub status API calls for our internal Jenkins deploy and functional test runs. This should allow us to see in GitHub PRs when our internal deploys fail and when our automated functional test runs that depend on those deploys fail.

This is imperfect right now because it depends on our Jenkins polling GitHub, but it gets the information into the PR where it's most relevant.

This will require editing our `/etc/hosts` files to ensure the "Details" link works.

## Checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,6 +4,14 @@
 // This pipeline uses Jenkins Shared Libraries for several
 // pipeline steps. For details on how those work, see
 // GHE repo: app-ops/app-ops-jenkins-shared-libraries
+
+def postGitHubStatus(String context, String status, String message, String url) {
+    masked_url = url.replace(env.JENKINS_URL, "http://dev-jenkins/")
+    withCredentials([string(credentialsId: 'cfpbot-github-api-token', variable: 'GITHUB_TOKEN')]) {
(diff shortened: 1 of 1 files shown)


Scotchester (commented):
I think these changes will enable linking to the Jenkins jobs via a dummy `dev-jenkins` hostname that we can direct to the right place with an entry in our hosts files.

Jenkinsfile
```diff
@@ -4,6 +4,14 @@
 // This pipeline uses Jenkins Shared Libraries for several
 // pipeline steps. For details on how those work, see
 // GHE repo: app-ops/app-ops-jenkins-shared-libraries
+
+def postGitHubStatus(String context, String status, String message, String url) {
+    masked_url = masked_job_url = url.replace(env.JENKINS_URL, "https://dev-jenkins/")
```
Scotchester, line 9:
```suggestion
    masked_url = url.replace(env.JENKINS_URL, "https://dev-jenkins/")
```

Scotchester (approved):
Woohoo!

PR #6160, opened 2020-12-11
Use Python pathlib instead of third-party Unipath

chosak:
This commit deprecates use of the third-party Unipath package and instead uses the built-in Python pathlib package.

Unipath [is now in maintenance mode](https://github.com/mikeorr/Unipath). We used it previously with Python 2. Python 3.4 introduced [pathlib](https://docs.python.org/3/library/pathlib.html) which has the same functionality for our purposes.

We currently use Unipath for path-handling logic, but `pathlib.Path` can do the same thing. The default Django project settings file [uses pathlib in this way](https://github.com/django/django/blob/45f4282149e13a2c1548a579f60d098e397a33d7/django/conf/project_template/project_name/settings.py-tpl#L13-L16).

There are [no other uses](https://github.com/search?q=user%3Acfpb+unipath&type=code) of pathlib in any of the cf.gov satellite apps.

## Notes and todos

Because this removes a dependency on a third-party package, this will (very incrementally) reduce the size of cf.gov builds and speed up the deploy process.

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -1,4 +1,5 @@
 import os
+from pathlib import Path
 
 from django.conf import global_settings
 from django.utils.translation import ugettext_lazy as _
@@ -6,17 +7,16 @@ from django.utils.translation import ugettext_lazy as _
 import dj_database_url
 from elasticsearch7 import RequestsHttpConnection
 from requests_aws4auth import AWS4Auth
-from unipath import DIRS, Path
 
 from cfgov.util import admin_emails
 
 
 # Repository root is 4 levels above this file
-REPOSITORY_ROOT = Path(__file__).ancestor(4)
+REPOSITORY_ROOT = Path(__file__).resolve().parents[3]
 
 # This is the root of the Django project, 'cfgov'
-PROJECT_ROOT = REPOSITORY_ROOT.child("cfgov")
-V1_TEMPLATE_ROOT = PROJECT_ROOT.child("jinja2", "v1")
+PROJECT_ROOT = REPOSITORY_ROOT.joinpath("cfgov")
+V1_TEMPLATE_ROOT = PROJECT_ROOT.joinpath("jinja2", "v1")
 
 SECRET_KEY = os.environ.get("SECRET_KEY", os.urandom(32))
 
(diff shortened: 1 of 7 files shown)


Scotchester (approved):
This appears to have no ill effects locally or on the dev Docker stack 👍

cfgov/cfgov/settings/base.py
```diff
@@ -275,12 +275,14 @@
...
-    PROJECT_ROOT.child("templates", "wagtailadmin"),
+    PROJECT_ROOT.joinpath("static_built"),
+    PROJECT_ROOT.joinpath("templates", "wagtailadmin"),
 ]
 
 # Also include any directories under static.in
-STATICFILES_DIRS += REPOSITORY_ROOT.child("static.in").listdir(filter=DIRS)
+STATICFILES_DIRS += [
+    d for d in REPOSITORY_ROOT.joinpath("static.in").iterdir() if d.is_dir()
+]
```
higs4281, line 285: This pattern reaches /static.in/fake-fonts/, but not /static.in/fake-fonts/fonts/, where the files live. 
using glob('**') instead of iterdir() will get there.

I wonder if this could be why we see sporadic missing-font errors?

```suggestion
STATICFILES_DIRS += [
    d for d in REPOSITORY_ROOT.joinpath("static.in").glob('**') if d.is_dir()
]
```

chosak: @higs4281 I believe this is deliberate and consistent with current behavior on the main branch. Font files are always referred to in code using a `fonts/` prefix, for example [here](https://github.com/cfpb/consumerfinance.gov/blob/701354b2b11675cc6c1c1e0616922d1c07d97ae5/cfgov/jinja2/v1/_layouts/base.html#L134) in our base template or [here](https://github.com/cfpb/design-system/blob/cefdfc3a6402a8551d3ed40effe962553f2d7805/packages/cfpb-typography/src/cfpb-typography.less#L18) in the design system. This means that when we invoke `static('fonts/something')`, that gets added on to the `/static.in/fake-fonts/` path defined here, producing the correct filename.

> we see sporadic missing-font errors

I'm not familiar with these. The custom fonts (Avenir Next) work locally for me. Is there a related issue?

higs4281 (approved):

cfgov/cfgov/settings/base.py
higs4281: No issue.

PR #6162, opened 2020-12-11
IA migration cleanup: Script to update links within Regs3k

schbetsy:
Links within the Interactive Bureau Regulations tool are stored in the reg markdown as literal `a` tags, like this: `<a href="/policy-compliance/rulemaking/regulations/1002/2/#l" data-linktag="0">1002.2(l),</a>`. All of those pages moved as part of the IA migration on Nov 25 from `/policy-compliance/rulemaking/regulations/` to `/rules-policy/regulations/`, so we need to update the links to their new destinations.

This PR updates those links to use relative paths, to prevent this issue in the future. Example:

`<a href="/policy-compliance/rulemaking/regulations/1002/2/#l" data-linktag="0">1002.2(l),</a>`
Within reg 1002, changes to: `<a href="../2/#l" data-linktag="0">1002.2(l),</a>`
Within another reg, changes to: `<a href="../../1002/2/#l" data-linktag="0">1002.2(l),</a>`

See https://github.local/CFPB/super-regular/issues/215 for more information.

## Additions

- One-time script to update links within regulations

## How to test this PR

1. `cfgov/manage.py runscript update_internal_regs3k_links`
2. Visit a reg page such as http://localhost:8000/rules-policy/regulations/1002/5/
3. Check that all links lead to the intended destinations and use the new URL pattern, `/rules-policy/regulations/`

I'm also planning to run this change on DEV4. I can let y'all know when I do that.

## Checklist

### DIFF
--- a/cfgov/scripts/update_internal_regs3k_links.py
+++ b/cfgov/scripts/update_internal_regs3k_links.py
@@ -1,0 +1,35 @@
+import logging
+
+from regulations3k.models.django import Section
+
+
+logger = logging.getLogger(__name__)
+
+
+# This is a one-time script to update internal links within the Regs3k tool
+# that were outdated when the regulations tool moved to a new URL,
+# /policy-compliance/rulemaking/regulations/ to /rules-policy/regulations/.
+# Delete this script after running it in Production, December 2020.
+# Run it with the following command:
+# cfgov/manage.py runscript update_internal_regs3k_links
+def run():
+    count = 0
+    url_base = "/policy-compliance/rulemaking/regulations/"
+    for section in Section.objects.all():
+        # First, update any links to this reg section
+        # /policy-compliance/rulemaking/regulations/[part]/[section]/#[pg]
+        # to ../[section]/#[pg]
+        part = section.subpart.label
+        this_section = url_base + part + "/"
+        update1 = section.contents.replace(this_section, "../")
+
+        # Next, update any links to other reg sections
+        # /policy-compliance/rulemaking/regulations/[part]/[section]/#[pg]
+        # to ../../[part]/[section]/#[pg]
+        update2 = update1.replace(url_base, "../../")
+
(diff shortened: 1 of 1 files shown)


niqjohnson:
Just a few notes, @schbetsy, none of which should change anything, but I wanted to call them out just in case:

1. Links in regulation text can also be added using Markdown's `[text](url)` format. Markdown link URLs should be using the same format as HTML links, though, so the script looks like it will catch any of those.
2. We've definitely added some relative links when making updates before—I used all relative links in the Reg F update I worked on last week. But those URLs should already be in the `../{section}/#{paragraph}` format, so it looks like the script will ignore those just fine.
3. There is technically one other URL pattern that could exist: previous and future effective versions have URLs like `/rules-policy/regulations/1026/2020-12-28/1/#a` (note the effective date in there between the part number and section number). I would be *very* surprised if we had any links in any of the regs that point directly to a specific effective version, though, so I think we're safe to continue to ignore URLs like that in the script. And, like we talked about before, changing all the reg links to the relative `../` versions should fix issues we've seen in the past where previous and future versions inadvertently link to the current version, so that's a huge bonus!

Also, not needed for this PR, but @higs4281, do we need to update the script we use to parse imported regulations for links to use relative URLs instead?

higs4281:
> Also, not needed for this PR, but @higs4281, do we need to update the script we use to parse imported regulations for links to use relative URLs instead?

Yes, indeed. Thanks for the heads-up.

schbetsy:
That's excellent information, thank you @niqjohnson! I just ran the change on DEV4. Would you (and @higs4281, perhaps) be willing to take a look and make sure it looks okay?

higs4281 (approved):
This looks great on DEV4. Thanks for cleaning these up!   
I'll follow up with the link-generating script.

schbetsy:
Thanks for reveiewing!

PR #6166, opened 2020-12-14
CCDB: Update highcharts to 7.2.2 to fix security warning

anselmbradford:
Screen Shot 2020-12-14 at 12 18 46 PM

## Changes

- Updates highcharts from 7.2.1 to 7.2.2 to fix security warning.

## How to test this PR

1. Pull branch, run `./frontend.sh`, and visit http://localhost:8000/data-research/consumer-complaints/ and all should be as it was.


flacoman91 (approved):
Works beautifully.  IE 11 works too. 👍

PR #6170, opened 2020-12-14
Only expose django on localhost

wpears:
Follow-up to https://github.com/cfpb/consumerfinance.gov/pull/6121

This prevents django from being opened across the network on 8000, which was missing in the previous PR.


contolini (approved):
End the exposition. 👍

PR #6175, opened 2020-12-16
Use the Jenkins Blue Ocean URL for "Details"

willbarton:
This small change uses the Jenkins Blue Ocean URL for the "Details" link in our Jenkins GitHub status checks. Previous we were only using it for failures. But it's prettier than standard Jenkins, and is always specific to the run rather than the parent job.
## Checklist

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -82,8 +82,8 @@ pipeline {
                 DOCKER_BUILDKIT = '1'
             }
             steps {
-                postGitHubStatus("jenkins/deploy", "pending", "Building", env.JOB_URL)
-                postGitHubStatus("jenkins/functional-tests", "pending", "Waiting", env.JOB_URL)
+                postGitHubStatus("jenkins/deploy", "pending", "Building", env.RUN_DISPLAY_URL)
+                postGitHubStatus("jenkins/functional-tests", "pending", "Waiting", env.RUN_DISPLAY_URL)
 
                 script {
                     LAST_STAGE = env.STAGE_NAME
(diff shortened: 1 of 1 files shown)


higs4281 (approved):
Great find. I didn't know RUN_DISPLAY_URL was available.

willbarton:
> Great find. I didn't know RUN_DISPLAY_URL was available.

I didn't either until I went to try to figure out why the MatterMost "Details" URL went to Blue Ocean (especially nice in the case of a failure).

Scotchester:
My choices here were intentional, because I think most people only want to look at the log if something went wrong (in which case I did use `RUN_DISPLAY_URL`). If it's just running normally, I prefer to look at the pipeline stage view. Not a hill I'm gonna die on, but that's what I was thinking.

Jenkinsfile
```diff
@@ -206,7 +206,7 @@ pipeline {
 
                 if (env.DEPLOY_SUCCESS == false) {
                     postGitHubStatus("jenkins/deploy", "failure", "Failed", env.RUN_DISPLAY_URL)
-                    postGitHubStatus("jenkins/functional-tests", "error", "Cancelled", env.JOB_URL)
```
Scotchester, line 209: And there's no point to look at a log for a check that didn't run.

willbarton: Well, we still need the URL and we need to report an "error" status back to GitHub.

Scotchester: `env.JOB_URL` is still a URL ;)

willbarton: Oh, I misunderstood from the diff — Yes, it is, but you can still see the failure logs directly for the prerequisite steps with `RUN_DISPLAY_URL`.

Scotchester: Sorry, yeah, should have commented on the green line.

Fair point, they probably want those logs if the functional tests didn't run.

PR #6176, opened 2020-12-16
Move language macro from main container to blocks

contolini:
We have a [macro](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/jinja2/v1/_includes/macros/accessible-languages.html) that swaps the language direction for Arabic. We currently use it on a page's container element which causes all children inline elements to flow from right to left, breaking our inline-block layout and making the site unusable for Arabic speakers. Adding it to individual children blocks preserves the layout while still reversing text flow.

## How to test this PR

1. `./refresh-data.sh`
1. Compare the [new Arabic content](http://content.localhost:8000/language/ar/coronavirus/mortgage-and-housing-assistance/) before and after this PR.

## Screenshots

| before | after |
|--------|-------|
| before | after  |

## Notes

- In the above screenshot the English content is also RTL'ed -- this is bad. We're deciding if we should either remove English content from Arabic pages or find another solution.

## Checklist

#### Browser testing

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Safari on iOS
- [ ] Chrome on Android

#### Accessibility

- [ ] Screen reader friendly

#### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS

### DIFF
--- a/cfgov/jinja2/v1/_includes/templates/render_block.html
+++ b/cfgov/jinja2/v1/_includes/templates/render_block.html
@@ -1,4 +1,5 @@
 {% macro render(block, index) %}
+    {%- import 'macros/accessible-languages.html' as accessible_languages with context -%}
     <div class="block
                 {{ 'block__flush-top' if index == 1
                    and not (block.value.has_top_border or block.value.has_top_rule_line) else '' }}
@@ -9,7 +10,8 @@
                 {{ 'block__padded-bottom block__border-bottom'
                    if block.value.has_rule or block.value.has_bottom_border else '' }}
 
-                {{ block.block.meta.classname if block.block.meta.classname else '' -}}">
+                {{ block.block.meta.classname if block.block.meta.classname else '' -}}"
+         {{ accessible_languages.render() }}>
         {{ render_stream_child(block) | safe }}
     </div>
 {% endmacro %}
--- a/cfgov/jinja2/v1/_includes/templates/render_block.html
+++ b/cfgov/jinja2/v1/_includes/templates/render_block.html
@@ -11,7 +11,7 @@
                    if block.value.has_rule or block.value.has_bottom_border else '' }}
 
                 {{ block.block.meta.classname if block.block.meta.classname else '' -}}"
-         {{ accessible_languages.render() }}>
+        {{ accessible_languages.render() }}>
         {{ render_stream_child(block) | safe }}
     </div>
 {% endmacro %}
--- a/cfgov/jinja2/v1/_includes/macros/accessible-languages.html
+++ b/cfgov/jinja2/v1/_includes/macros/accessible-languages.html
@@ -11,10 +11,8 @@
    ========================================================================== #}
 
 {% macro render() %}
-    {%- if page -%}
-        lang="{{ page.language if page.language else 'en' }}"
-              {%- if page.language in ['ar'] -%}
-                  dir="rtl"
-              {%- endif -%}
+    {%- set rtl_languages = ['ar'] -%}
+    {%- if page and page.language in rtl_languages -%}
+        lang="{{ page.language }}" dir="rtl"
     {%- endif -%}
 {% endmacro %}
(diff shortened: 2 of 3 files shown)


chosak:
Although this might work for this specific case, we don't use render_block.html in every place that we render blocks.

For example, in [the BrowsePage template](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/jinja2/v1/browse-basic/index.html), we only use render_block.html to render header blocks that aren't featured content -- all other blocks on the page use some other form of rendering, for example [`{% include_block %}`](https://docs.wagtail.io/en/stable/advanced_topics/jinja2.html?highlight=include_block#include-block).

The specific case you cite is a [SublandingPage](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/jinja2/v1/sublanding-page/index.html), which uses the [streamfield-sidefoot.html](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/jinja2/v1/_includes/templates/streamfield-sidefoot.html) template for its sidebar, and that template only uses render_block.html for a subset of its child blocks.

So while this might work here, I worry that it would give us only a partial solution that might be hard to remember/debug down the road.

If we do decide to move ahead with this, one option for solving the English content flow might be to add a language attribute to the ReusableText snippet, that is used to set the direction when the snippet is rendered.

contolini:
> If we do decide to move ahead with this, one option for solving the English content flow might be to add a language attribute to the ReusableText snippet, that is used to set the direction when the snippet is rendered.

I dig this. In the interim should we apply this patch to `streamfield-sidefoot.html` instead of `render_block.html`?

contolini:
@Scotchester How do you feel about this [this change](https://github.com/cfpb/consumerfinance.gov/pull/6176/files#diff-41c66d0da3b48db758da858a74b076da092c726055d7bc57b2a27d5f342bd21b) to prevent it from spamming `lang="en"` on english pages?

Scotchester:
> @Scotchester How do you feel about this [this change](https://github.com/cfpb/consumerfinance.gov/pull/6176/files#diff-41c66d0da3b48db758da858a74b076da092c726055d7bc57b2a27d5f342bd21b) to prevent it from spamming `lang="en"` on english pages?

Don't we still need that macro to output the `lang` attribute on non-RTL-language pages?

contolini:
> > @Scotchester How do you feel about this [this change](https://github.com/cfpb/consumerfinance.gov/pull/6176/files#diff-41c66d0da3b48db758da858a74b076da092c726055d7bc57b2a27d5f342bd21b) to prevent it from spamming `lang="en"` on english pages?
> 
> Don't we still need that macro to output the `lang` attribute on non-RTL-language pages?

There's a `lang` attribute on both the `html` element and the main wrapper of every page. Note: Our non-english pages are actually mixed language so this isn't ideal but fixing it involves a much larger strategy discussion.

Scotchester:
> > > @Scotchester How do you feel about this [this change](https://github.com/cfpb/consumerfinance.gov/pull/6176/files#diff-41c66d0da3b48db758da858a74b076da092c726055d7bc57b2a27d5f342bd21b) to prevent it from spamming `lang="en"` on english pages?
> > 
> > 
> > Don't we still need that macro to output the `lang` attribute on non-RTL-language pages?
> 
> There's a `lang` attribute on the `html` element which covers all its children (i.e. the entire page). Note: Our non-english pages are actually mixed language so this isn't ideal but fixing it involves a much larger strategy discussion.

Ah, right. How quickly I forget 😆

cfgov/jinja2/v1/_includes/templates/render_block.html
```diff
@@ -9,7 +10,8 @@
                 {{ 'block__padded-bottom block__border-bottom'
                    if block.value.has_rule or block.value.has_bottom_border else '' }}
 
-                {{ block.block.meta.classname if block.block.meta.classname else '' -}}">
+                {{ block.block.meta.classname if block.block.meta.classname else '' -}}"
+         {{ accessible_languages.render() }}>
```
Scotchester, line 14:
```suggestion
        {{ accessible_languages.render() }}>
```

Scotchester (approved):
I guess this is fine as an immediate fix to the hidden content, but we really need to follow-up on this quickly and clean up this mess. Maybe something for the Dev Services backlog early next year?

PR #5703, opened 2020-05-05
Require labels on subparts and sections

willbarton:
This change requires the label on subparts and sections to not be blank.

It was previously possible to save a section or subpart in the Wagtail admin without a label. This would then cause 500 errors when the label, which was expected to exist, did not.

This fixes some issues we've seen in production.

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [x] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

PR #6177, opened 2020-12-16
Override the TreeModelAdmin URL Helper for Regs breadcrumbs

willbarton:
This change overrides the TreeModelAdmin URL Helper for Regulations breadcrumbs. in #5703 we ensured that the regs models' `__str__` methods included their parents, so that one could see what regulation and version a particular section object belongs to in the model drop-downs and anywhere else Django renders the model to string.

Unfortunately the `__str__` is the same mechanism TreeModelAdmin uses to generate the breadcrumbs at the top. This seems reasonable, which is why I don't want to change it there.

So to eliminate the duplicity in the breadcrumbs, this change overrides the TreeModelAdmin URL helper class to remove the parent's string if the URL helper is generating a breadcrumb for a specific model and a parent instance is given.

Before:

image

After:

image

## Checklist

### DIFF
--- a/cfgov/regulations3k/tests/test_hooks.py
+++ b/cfgov/regulations3k/tests/test_hooks.py
@@ -10,6 +10,7 @@ from regulations3k.models.django import (
     EffectiveVersion, Part, Section, Subpart
 )
 from regulations3k.models.pages import RegulationLandingPage, RegulationPage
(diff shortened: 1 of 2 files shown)


Scotchester (approved):
Works a treat!

higs4281 (approved):
Looks great!

PR #6181, opened 2020-12-17
Add rule to remove trailing . when visiting the site

alexm118:
Currently we don't redirect back to the consumerfinance.gov when a user submits consumerfinance.gov. 

This rule should strip the trailing . from the request per this [stack overflow post](https://stackoverflow.com/questions/31972945/how-to-remove-the-trailing-dot-after-tld-by-using-mod-rewrite)

---

## Additions

- Remove trailing . from requests.

## How to test this PR

1. Visit the deployed docker stack with a trailing . on the url and confirm you are redirected.

## Notes and todos

-

## Checklist


cfgov/apache/conf.d/redirects.conf
```diff
@@ -1026,3 +1026,9 @@ RewriteRule ^(?!.*(csv|html|json|jsonp|xml)$).*$ https://ffiec.cfpb.gov/data-pub
 ## Except for requests to specific file endpoints, which we mark as gone
 RewriteCond %{HTTP_HOST} api.consumerfinance.gov
 RewriteRule ^.+(csv|html|json|jsonp|xml)$ - [G,NC]
+ 
+ 
+# Strip trailing period for domain name
+# https://stackoverflow.com/questions/31972945/how-to-remove-the-trailing-dot-after-tld-by-using-mod-rewrite
+RewriteCond %{HTTP_HOST} ^(.+?)\.$
+RewriteRule ^ http://%1%{REQUEST_URI} [L,R=302,NE]
```
Scotchester, line 1034:
```suggestion
RewriteRule ^ http://%1%{REQUEST_URI} [L,R=301]
```

I think we want these to be permanent redirects, and I'm pretty sure we don't need the No Escape flag.

Scotchester (commented):
This is working for me, but I have a couple suggested changes to the flags, and Bill expressed some reservations, so I definitely would like someone besides just me to approve.

cfgov/apache/conf.d/redirects.conf
```diff
@@ -1026,3 +1026,9 @@ RewriteRule ^(?!.*(csv|html|json|jsonp|xml)$).*$ https://ffiec.cfpb.gov/data-pub
 ## Except for requests to specific file endpoints, which we mark as gone
 RewriteCond %{HTTP_HOST} api.consumerfinance.gov
 RewriteRule ^.+(csv|html|json|jsonp|xml)$ - [G,NC]
+ 
+ 
+# Strip trailing period for domain name
+# https://stackoverflow.com/questions/31972945/how-to-remove-the-trailing-dot-after-tld-by-using-mod-rewrite
+RewriteCond %{HTTP_HOST} ^(.+?)\.$
+RewriteRule ^ http://%1%{REQUEST_URI} [L,R=301]
```
higs4281, line 1034: The replace seems to work correctly, but I'm not following why we'd want a space after the `^` in the RewriteRule.

It's not a hill worth dying on, but this seems to be the wrong place to address this. Indexing bots should know that a domain followed by a period is not duplicate content, and analytics systems should likewise not be thrown for a loss when encountering a fully qualified domain in a request.

higs4281 (commented):
Just one comment.

cfgov/apache/conf.d/redirects.conf
alexm118: I couldn't tell you the specifics on why the rule contains the space other than its what was posted on the stackoverflow link and it worked :joy:

alexm118: My gut makes me think that the space is what allows it to match the host specifically and not a path? But I could be misinterpreting

higs4281: Can't argue with success.

Scotchester: > It's not a hill worth dying on, but this seems to be the wrong place to address this. Indexing bots should know that a domain followed by a period is not duplicate content, and analytics systems should likewise not be thrown for a loss when encountering a fully qualified domain in a request.

Google may not be thrown, but it does seem like the analytics tool that brought this issue to our attention isn't smart enough to handle it :\

I also think it's just preferable from a UX POV. We don't want there to be two identical URLs that access the same page.

wpears: I'm curious about how this redirect with the hardcoded `http` redirect will work. Is it returning `http://cf.gov` for `https://cf.gov.`? We might need to flip this to hardcoded https

Scotchester: D'oh, right. Almost certianly want it to be `https`

higs4281: > preferable from a UX POV

I don't get the UX angle. I think we'd be hard-pressed to find a user who could spot a difference in response behavior after this change. 

I don't think we should fight DNS protocol, but since I can't speak for UX or analytics, I'll stop now.

alexm118: swapped to https

Scotchester (approved):
Approved, but hold for the new year.

PR #6182, opened 2020-12-18
CCDB: Add consumer-complaints functional tests

cwdavies:
Add Cypress tests to CCDB landing page (/data-research/consumer-complaints/) to test dynamic parts

## How to test this PR

Run `yarn run cypress open` and select the cypress integration test `consumer-complaints.js`

## Checklist

### DIFF
--- a/test/cypress/integration/pages/consumer-complaints.js
+++ b/test/cypress/integration/pages/consumer-complaints.js
@@ -1,0 +1,75 @@
+import { ConsumerComplaints } from '../../pages/data-research/consumer-complaints';
+
+const page = new ConsumerComplaints();
+const states = ['AK', 'AL', 'AZ', 'AR', 'CA', 'CO', 'CT', 'DE', 'FL', 'GA',
+  'HI', 'ID', 'IL', 'IN', 'IA', 'KS', 'KY', 'LA', 'ME', 'MD', 'MA', 'MI', 'MN',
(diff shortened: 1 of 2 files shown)


test/cypress/integration/pages/consumer-complaints.js
```diff
@@ -0,0 +1,75 @@
...
+
+describe( 'Consumer Complaint Database', () => {
+  beforeEach( () => {
+    cy.visit( '/data-research/consumer-complaints/' );
+  } );
+
+  it( 'should display chart with complaints for all 50 states', () => {
+    page.clickButton( 'Complaints' );
+    page.checkLegend( 'description' ).should( 'contain', 'Complaints' );
+    page.chartSize( 580, 940 );
```
sephcoster, line 17: Because the chart will update at different viewport sizes it might be sufficient to test that the element exists and contains tiles, rather than looking for specific pixles (this fails locally with a height of 925, for example) but otherwise these look great!

sephcoster (approved):

test/cypress/integration/pages/consumer-complaints.js
cwdavies: Thanks for your review comments. I will remove the test for the map size and just verify that the map is visible. Currently only the static chart is being tested. More effort would be required to get the elasticsearch for CCDB functional in Docker where the functional tests are being run.

PR #6184, opened 2020-12-18
Prefer date_filed from final_dispostion for enforcement actions

wpears:
EnforcementActionPages are currently structured to allow the date_filed to be updated in the final_disposition section in Wagtail. This updates the template to match that expected behavior.


schbetsy (approved):
Works for me.

PR #6185, opened 2020-12-18
Bugfix: Update Table module import to fix initialization

anselmbradford:
## Changes

- Update Table module import to fix initialization.

## How to test this PR

1. https://www.consumerfinance.gov/ask-cfpb/what-do-i-do-if-i-think-i-have-been-a-victim-of-identity-theft-en-31/ should have an init error in the dev console.
2. Pull branch and `gulp clean && gulp build` and http://localhost:8000/ask-cfpb/what-do-i-do-if-i-think-i-have-been-a-victim-of-identity-theft-en-31/ should not have an error.

### DIFF
--- a/cfgov/unprocessed/js/routes/on-demand/table.js
+++ b/cfgov/unprocessed/js/routes/on-demand/table.js
@@ -3,4 +3,5 @@
    ========================================================================== */
 
 // See https://github.com/cfpb/design-system/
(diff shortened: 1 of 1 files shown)


Scotchester (approved):
Works :+1: Also confirmed that enabling sorting on that table didn't work before, and this makes that work again.

PR #6186, opened 2020-12-21
Fix the Docker attach line in docs

willbarton:
This change fixes the line in our docs for the `docker attach` command when you need to interact with a `pdb` session. I think when we renamed this repo from `cfgov-refresh` to `consumerfinance.gov`, we mistakenly broke this command. It should not have the dot in the container name; it should be `consumerfinancegov_python_1`.

## Checklist

### DIFF
--- a/docs/running-docker.md
+++ b/docs/running-docker.md
@@ -80,10 +80,16 @@ and need to interact with the running Django process when the breakpoint is reac
 you can run [`docker attach`](https://docs.docker.com/engine/reference/commandline/attach/):
 
 ```bash
-docker attach consumerfinance.gov_python_1
+docker attach consumerfinancegov_python_1
 ```
 
 When you're done, you can detach with `Ctrl+P Ctrl+Q`.
+
+!!! note
+    `docker attach` takes the specific container name or ID. 
+    Yours may or may not be `consumerfinancegov_python_1`. 
+    To verify, use `docker container ls` 
+    to get the Python container's full name or ID.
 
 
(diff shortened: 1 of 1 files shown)


cwdavies (approved):

higs4281 (commented):
This container name not work for everyone, depending on how they've set up their local clone. 

For instance, my docker python container is `cfgov-refresh_python_1` because I didn't change my directory name when I changed the remote name,  and `docker-compose.yml` picks uses the [current directory](https://github.com/cfpb/consumerfinance.gov/blob/main/docker-compose.yml#L46) for its `context` value.

We could perhaps advise checking `docker container list` to see what the container names are, or for consistency maybe we could specify a [container-name](https://docs.docker.com/compose/compose-file/compose-file-v3/#container_name) in our compose file. That does not scale, but might be fine for local development.

Scotchester:
> This hard-coded container name may not work for everyone, depending on how they've set up their local environment.

Can you not just refer to it as `python` as you can with other commands? Or is that something specific to `docker-compose`?

higs4281:
> just refer to it as `python`

`docker attach` isn't a docker-compose command; it requires the specific container name. 

If our docs are intended for someone cloning the repo for the first time, maybe this is overkill, but it would be friendly to note that if you clone the repo into a named directory, your python container's name may be different, and you can check `docker container ls` to see the container's full name.

willbarton:
@higs4281 added a note to that effect in 6daaa52.

PR #6193, opened 2020-12-23
Update sorting for enforcement actions filter page to match publishing date

wpears:
Fixes an issue where the currently inconsistent `date_filed` field was causing enforcement actions to appear out of order

### DIFF
--- a/cfgov/v1/forms.py
+++ b/cfgov/v1/forms.py
@@ -279,7 +279,7 @@ class EnforcementActionsFilterForm(FilterableListForm):
     def get_page_set(self):
         query = self.generate_query()
         return self.filterable_pages.filter(query).distinct().order_by(
-            '-date_filed'
+            '-date_published'
         )
 
     def get_query_strings(self):


schbetsy (approved):

chosak:
Is this a temporary fix, or permanent change? If the latter, could we remove the definition of [`EnforcementActionsFilterForm.get_page_set`](https://github.com/cfpb/consumerfinance.gov/blob/32a734152cd16db54faee4879b91a661c4b550bd/cfgov/v1/forms.py#L279-L283) since it now duplicates the method inherited from its parent [`FilterableListForm.get_page_set`](https://github.com/cfpb/consumerfinance.gov/blob/32a734152cd16db54faee4879b91a661c4b550bd/cfgov/v1/forms.py#L127-L131)?

wpears:
Temporary! `date_filed` will return in the new year, this is just to keep the sorting accurate until I am available to build then final enforcement data model and transfer all the existing actions to it (but `date_filed` will be consistent once again in that model, so this will be reverted then)

PR #6194, opened 2020-12-28
Remove turbolinks cruft

anselmbradford:
Turbolinks was an experiment that was added in https://github.com/cfpb/consumerfinance.gov/pull/4020/files and https://github.com/cfpb/consumerfinance.gov/commit/e60341c4507cec437ae2a6fba80f356134d3bfdc#diff-99fb3cd201a06b0dd48b150e747494aa3e120954d0885ef6677687a8735040f6 

However, it is hidden behind a feature flag and that flag is turned off in production. According to @contolini it can be removed.

## Removals

- Removes turbolinks dependency and associated files and flag.

## How to test this PR

1. PR checks should pass.


contolini (approved):
turbogone

PR #6195, opened 2020-12-28
Use transition JavaScript modules from Design System

anselmbradford:
This repo and the design system both have transition modules that are nearly identical. This PR removes the JS in this repo and points the codebase at the modules brought in through the design system.

https://github.com/cfpb/design-system/tree/main/packages/cfpb-atomic-component/src/utilities/transition

## Changes

- Use `BaseTransition.js`, `MoveTransition.js`, and `AlphaTransition.js` from design system and remove redundant modules and tests from this repo.

## How to test this PR

1. Pull branch, run local server. Components that use move and alpha transitions should work as before. This is the mega menu, global search, blog filters, and feedback form, such as on http://localhost:8000/ask-cfpb/what-is-a-debt-to-income-ratio-why-is-the-43-debt-to-income-ratio-important-en-1791/

## Note

Transitions add an event listener that listens for the `transitionEnd` event, however, the element may have transitions that are performed outside of the transition class, making their `transitionEnd` event fire early and multiple times. To get around this, I've added a `CSS_PROPERTY` value that gets passed into the `BaseTransition` constructor, that is then checked in the `transitionEnd` handler when it fires. Not super happy with this solution, as the `CSS_PROPERTY` need to match the transitioned property in the CSS, which is a shame to have to duplicate that value. Also, this would break if an outside transition was working on the same CSS property, but that seems like a wrong implementation if that's happening.


anselmbradford:
Transitions needs updating in DS

anselmbradford:
https://github.com/cfpb/design-system/pull/1134 needs to go in and be released first.

contolini (approved):
And look there's a whole page with implementation details https://cfpb.github.io/design-system/patterns/transition-patterns 🆒

PR #6197, opened 2020-12-31
Admin Drafttables: Overwrite null save with acceptable blank format

alexhoffman617:
Fixes the bug where you cannot save a blank cell on the admin table. Fixes GHE/CFGOV/platform#3966

## Changes

- On a null save for a table cell, overwrite null with the appropriate format for a table cell and set the text to an empty string

## How to test this PR

1. Create new blog page
2. Add a table
3. Open a cell and save it with some text in it
4. Open the cell again, delete the text, and save

## Notes and todos

-

## Checklist


Scotchester (changes requested):
Unfortunately, this doesn't quite do it. After emptying out the cell that previously had content, it will indeed save with no visible content, but:

1. It leaves an undesirable `<br>` tag in the cell
2. I can then no longer double-click to reopen the cell for editing; that throws the following error: "Uncaught TypeError: Cannot read property 'contentBlocks' of null"

Scotchester (approved):
It's working! Nice job, @alexhoffman617 😄

PR #6208, opened 2021-01-12
Improve keyboard navigation of VideoPlayer

chosak:
This commit improves keyboard navigation of the VideoPlayer module by changing which elements are assigned tabstops.

This fixes https://github.com/cfpb/design-system/issues/1124 which reports that tabbing through the video player pauses on invisible elements. This is because currently both the close button and video iframe can be tabbed to even when they are invisible because the video is not playing. An example live page where this behavior can be seen is https://www.consumerfinance.gov/consumer-tools/managing-someone-elses-money/.

This commit adds `tabindex="-1"` to both the iframe and close button so that, in the initial page load state, neither of these invisible elements can be tabbed to.

The JS code is altered here so that, when the video is played, the `tabindex` attribute is removed, thus making these elements tabbable. The play button also then gets `tabindex="-1"`, as it gets hidden when the video is played. Additionally, the "Close video" now gets the focus when this happens (and also a new `aria-label`).

When the video is stopped, the JS code now reverses the process, and the focus is set to the play button.

## How to test this PR

To test, use VoiceOver on the Mac (for example as documented [here](https://www.imore.com/how-enable-voiceover-mac)) and tab through the page, noting that only appropriate elements can be selected. Use the space bar to open/close the video. Alternatively or additionally, you can also add a watch expression for `document.activeElement` in Chrome developer tools, and observe how it changes as you tab.

## Notes and todos

One potential followup to this work would be to consider the placement of the close button relative to the video controls. Currently the module HTML puts both the play and close buttons AFTER the video iframe. This means that someone using keyboard navigation would have to tab "in reverse" to get to the video controls, where they might expect these to come after play/close. Fixing this would require moving the buttons before the iframe, and I'm not sure what else this might impact.

## Checklist

#### Accessibility

- [x] Screen reader friendly

#### Other

- [ ] Is useable without CSS
- [x] Is useable without JS

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/video-player.html
+++ b/cfgov/jinja2/v1/_includes/organisms/video-player.html
@@ -54,17 +54,6 @@
         <div class="o-video-player_video-container
                     {{- ' o-video-player_video-container__flexible' if not is_fcm else '' }}
                     u-show-on-video-playing">
-            <div class="o-video-player_iframe-container">
-                <iframe class="o-video-player_iframe"
-                    width="100%"
-                    src="{{ video_url }}"
-                    frameborder="0"
-                    allow="autoplay"
-                    allowfullscreen="true"
-                    scrolling="no"
-                    tabindex="-1">
-                </iframe>
-            </div>
             <div class="o-video-player_video-actions-container">
                 <div class="o-video-player_video-actions">
                     <button class="a-btn o-video-player_close-btn"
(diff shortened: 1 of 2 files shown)


anselmbradford (approved):
Works great! 

One question: so you can play and close the video with the keyboard, but should the user be able to use the keyboard to interact with the native controls when the video is open?

chosak:
> One question: so you can play and close the video with the keyboard, but should the user be able to use the keyboard to interact with the native controls when the video is open?

@anselmbradford yes; that is possible, although as mentioned above in "Notes and todos", the order is a little unexpected. Because the open/close button are "after" the video element in the DOM, the user would have to first use the keyboard to press the play button, then tab in reverse to get to those native controls within the YouTube player.

What do you think about switching the order of that button and the player element? Is that feasible? Maybe I should leave that for a subsequent issue/PR.

contolini:
> What do you think about switching the order of that button and the player element? Is that feasible? Maybe I should leave that for a subsequent issue/PR.

Good call, I just swapped the actions container with the player container and there's no visual difference but the elements are read in a much more logical order. I'll open a PR with the change.

chosak:
@contolini should I merge this and then you'll open a subsequent one? Or did you want to PR to this PR? Feel free to push a commit here if you like.

contolini:
@chosak Pushed to this PR just now, give it a quick test.

chosak:
@contolini that works great! Nice improvement, thank you.

PR #6210, opened 2021-01-13
Paying for College: Remove bindEvent and move private methods from export

anselmbradford:
1. `bindEvent` iterates over a set of events and applies `addEventListener` to the target element. However, this only makes sense if there are a large number of events being applied. For the majority of uses in PfC it's added for a single event, in which case `elem.addEventListener( … )` should be used directly. 

2. The modules within the app export private methods. This PR moves these outside of the exported object since they shouldn't appear outside the module's API.

## Changes

- Paying for College: Replace `bindEvent` with native `addEventListener`.
- Paying for College: Move event handler methods that are private and being exported from a module to outside the export.

## How to test this PR

1. Pull branch, run local server, and app functionality should be unchanged.


mistergone (approved):
Looks good - thanks!

PR #6216, opened 2021-01-19
Update httpd.conf to pass environment variables

willbarton:
This PR updates our Apache config to ensure that our environment variables are available within the running httpd processes. This is necessary for our RHEL7 server upgrade, and I've confirmed there are no detrimental effects on RHEL6

## Checklist


alexm118 (approved):

Scotchester (approved):
Should we add something to our documentation calling out that this will be necessary for any new env vars we create?

willbarton:
Is creating env vars that need to be passed via Apache something we do often? This doesn't count things that get put in `environment.json` (i.e. the vars in this repo's `.env` file), and I wouldn't want to create confusion around that.

PR #6217, opened 2021-01-19
Change Featured Content body to plain text

chosak:
The Featured Content Module (FCM) is documented [in the design system](https://cfpb.github.io/design-system/patterns/featured-content-module). Its Wagtail block implementation is [`v1.atomic_elements.organisms.FeaturedContent`](https://github.com/cfpb/consumerfinance.gov/blob/904c49d2db1492ea47b62771b318700754393ffc/cfgov/v1/atomic_elements/organisms.py#L1051).

The design documentation mentions "descriptive body copy text" and that

> Descriptions should be action-oriented and let the user know what they will get out of the featured content and how it’s related to the main page content, not just what it is. Descriptions should be 160 characters or less (including spaces).

This description and the documented examples suggest that the body text should be plain text and thus should not contain links, images, or text formatting). But in code this field is implemented as a Wagtail [`RichTextBlock`](https://docs.wagtail.io/en/stable/topics/streamfield.html#richtextblock), which allows for all of these things.

This commit modifies the FCM body field to instead be a [`CharBlock`](https://docs.wagtail.io/en/stable/topics/streamfield.html#charblock). This allows for only a single line of plain text.

## How to test this PR

Run a local server and create or edit a page with a Featured Content module. Note that the body is now plain text. Visit a live page with an FCM ([example](http://localhost:8000/data-research/research-reports/)) and notice that the body still renders correctly.

## Screenshots

Before -- FCM body could contain rich text elements including links and images:

image

After -- FCM body must be plain text:

image

## Notes and todos

See internal https://github.local/CFGOV/platform/issues/3808#issuecomment-274183 for some related discussion including a link to a detailed analysis of our existing production pages and how this won't affect them.

One design consideration: the Wagtail editing experience is now slightly annoying if you have a long sentence, as it cuts off a requires scrolling at ~100 characters. Our design guidance suggests 160 characters or less (and we exceed this often), so this is something users might run into often. On the other hand, maybe this is good because it encourages brevity?

Another option would be to use a [`TextBlock`](https://docs.wagtail.io/en/stable/topics/streamfield.html#textblock) here instead, which allows for multiple lines of input, and automatically wraps. That would look like this:

image

The one downside is that, because this supports newlines, it might mislead editors to think they can create multiple paragraphs, where in practice, they won't be able to. For example this input:

image

will still get rendered like this in the FCM:

image

Any strong preferences on `CharBlock` vs `TextBlock`? @csebianlander or others?

## Checklist


anselmbradford:
😍

anselmbradford (approved):
Works as described. I defer the UX choices to others based on much text these may generally contain.

willbarton (approved):

csebianlander:
My thought is that multiline is definitely better from a usability point of view, but maybe we should also include some helper text along the lines of "Line breaks will be ignored" so content managers are aware of that.

chosak:
Thanks @csebianlander; I've changed to a multi-line `TextBlock` and added that help text:
image

PR #6222, opened 2021-01-19
Remove bindEvent utility

anselmbradford:
`bindEvent` iterates over a set of events and applies `addEventListener` to the target element. However, this only makes sense if there are a large number of events being applied, and even then it saves barely any code. We should just use `elem.addEventListener( … )` directly. 

## Removals

- Remove `bindEvent` utility and replace it with the native `addEventListener`.

## How to test this PR

1. The modules touched by this should work as before:
  - eregs
  - Ask search autocomplete, such as on http://localhost:8000/ask-cfpb/what-is-a-debt-to-income-ratio-why-is-the-43-debt-to-income-ratio-important-en-1791/
  - TDP analytics events (hard to test without putting in a console log, so maybe take my word) http://localhost:8000/consumer-tools/educator-tools/youth-financial-education/teach/activities/


contolini (approved):
iregs is looking good 👍

PR #6224, opened 2021-01-21
Update regdown and remove dependency on sha3

chosak:
**Hold until a new version of regdown has been released including https://github.com/cfpb/regdown/pull/4.**

This commit bumps the version of our external regdown package and also removes an unnecessary dependency on the sha3 package. This dependency is no longer needed on Python 3.6+.

## Notes and todos

Our inclusion of the external [sha3](https://pypi.org/project/sha3/) package can cause issues when trying to `pip install -e` on Macs using Python installed via pyenv. If sha3 is installed into a Python 3.6 virtual environment, in certain cases trying to `pip install -e` can result in errors like

> ...
> AttributeError: module '_hashlib' has no attribute 'openssl_sha3_224'
>
>  During handling of the above exception, another exception occurred:
> ...
> AttributeError: module '_sha3' has no attribute 'sha3_224'

## Checklist

### DIFF
--- a/requirements/libraries.txt
+++ b/requirements/libraries.txt
@@ -26,11 +26,10 @@ ntplib==0.3.4
 openpyxl==3.0.3
 psycopg2==2.7.3.2
 python-dateutil==2.7.3
-regdown==1.0.2
+regdown==1.0.3
 requests==2.22.0
 requests_aws4auth==0.8.0
 requests_toolbelt==0.8.0
-sha3==0.2.1
 urllib3==1.25.2
 # wagtail-autocomplete==0.6 TODO: Restore when wagtail-autocomplete #77 is merged
 wagtail-flags==5.0.0


chosak:
@Scotchester @willbarton I've released a new version of [regdown](https://pypi.org/project/regdown/#history) that removes the dependency of sha3. All tests pass there and here. I'm not quite sure how to verify that sha3-related code is working properly on regdown/iRegs pages -- can you confirm?

The Snyk failures above are puzzling and don't seem related to this change. The report (behind Snyk login) lists two high vulnerabilities, both for Pillow==6.2.2, but we resolve to Pillow==8.1.0.

willbarton:
Hmm, it looks like it's because of 

> Introduced through: wagtail-placeholder-images@0.1.1, wagtailmedia@0.6.0

Weird that it doesn't consider our explicit pin of Pillow==8.1.0.

willbarton:
>  I'm not quite sure how to verify that sha3-related code is working properly on regdown/iRegs pages -- can you confirm?

Sure, I'll verify when the Docker deploy is done. 

Basically, it generates a unique id for a paragraph from a hash of its contents, if the paragraph doesn't have an explicit label. 

[This test in regdown should cover it](https://github.com/cfpb/regdown/blob/main/regdown/tests.py#L23-L31), and I'm not worried, especially [since we were already using `hashlib.sha3_224` in Python 3](https://github.com/cfpb/regdown/pull/4/files#diff-e05fdfa206380e78e23b753fad28c0f5fd364d2550b7a3bbe55c99294a26d89dL10-L17)

willbarton:
Should we explicitly pin Pillow?

chosak:
> Introduced through: wagtail-placeholder-images@0.1.1, wagtailmedia@0.6.0
> Should we explicitly pin Pillow?

Neither [w-p-i](https://github.com/rinti/wagtail-placeholder-images/blob/04f82b07ad9f0ab85cc1016b5ed9f5a5faa4a488/setup.py#L21) nor [wm](https://github.com/torchbox/wagtailmedia/blob/b88b94b4edf9ebfbe30f290eb63826ceab38f525/setup.py#L67) pin Pillow; they both require `wagtail>=2.7`. Wagtail 2.7 itself pins [`Pillow>=4.0.0,<7.0.0`](https://github.com/wagtail/wagtail/blob/72dcbb674cac38b488de984719cb20a08542eb5f/setup.py#L30), which must be how Snyk finds the version 6.2.2 is complaining about. For cf.gov we currently pin [wagtail==2.11.3](https://github.com/cfpb/consumerfinance.gov/blob/d0006df58018c17178c9d44b8289ea444d8af71f/requirements/wagtail.txt#L1); that version specifies [`Pillow>=4.0.0,<9.0.0`](https://github.com/wagtail/wagtail/blob/69f3f5c0d200785a1151127af4f2b0a4e3ac418d/setup.py#L30).

Given that combination of things, pip should resolve the latest Pillow `<9.0.0`, which is our current [8.1.0](https://pypi.org/project/Pillow/#history). So I don't think there's a need to pin explicitly -- that doesn't explain why Snyk is complaining on this PR. Is it possible that it's missing our Wagtail version somehow?

When I look in the Snyk [dashboard](https://app.snyk.io/org/cfpb) (behind login), I see a few requirements files listed there, but neither django.txt or wagtail.txt are listed. Do these need to be added?

willbarton:
> When I look in the Snyk dashboard (behind login), I see a few requirements files listed there, but neither django.txt or wagtail.txt are listed. Do these need to be added?

They probably should be. But it might not resolve the problem with `libraries.txt`. We might be best just marking it as not vulnerable in Snyk.

chosak:
> We might be best just marking it as not vulnerable in Snyk.

@willbarton done.

willbarton (approved):

PR #6227, opened 2021-01-22
Enforcement chart data

wpears:
Hold for review

Adds the ability to filter charts on the enforcement database page by keys present in the data. The select options are automatically generated from the chosen data values (selected in Wagtail) so that the organism can stay general.

More to come..!


wpears:
Superceded by https://github.com/cfpb/consumerfinance.gov/tree/simple-chart-csv

PR #6229, opened 2021-01-22
Restrict EnforcementActionPages to one part of the page tree

schbetsy:
Ensure all `EnforcementActionPage`s are children of `EnforcementActionFilterPage`. This helps ensure that when we use `EnforcementActionPage.objects.all()` (or similar), we're only getting Enforcement Actions that should be included in the list.

## Changes

- `EnforcementActionPage` can only be a child of `EnforcementActionFilterPage`.
- `EnforcementActionFilterPage` can only have two page types as children: `EnforcementActionPage` and `LearnPage` (the Enforcement Action Definitions page).
- `EnforcementActionPage` can't have child pages.

## How to test this PR

1. Check that you can create a new `EnforcementActionPage` as a child of the Enforcement Actions Filter page.
2. Check that you can't create one in other parts of the page tree.

## Checklist

### DIFF
--- a/cfgov/v1/models/browse_filterable_page.py
+++ b/cfgov/v1/models/browse_filterable_page.py
@@ -77,6 +77,8 @@ class BrowseFilterablePage(FilterableListMixin, CFGOVPage):
 
 
 class EnforcementActionsFilterPage(BrowseFilterablePage):
+    subpage_types = ['v1.EnforcementActionPage', 'v1.LearnPage']
+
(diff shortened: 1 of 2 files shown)


wpears:
I think this prevents moving an enforcement action page to the Trash, might need to expand the parent page types to include that

schbetsy:
Gooooood call. Updated to allow the `HomePage` type, which Trash is an instance of.

PR #6231, opened 2021-01-22
Stop server errors when copying pages

higs4281:
Currently, an attempt to copy a page in Wagtail copies the page and its tags,
but then the admin API raises a server error when it tries to handle tags.
If we exclude tags from admin operations, the errors cease, and page copying
can complete as expected.

This sidesteps a broader question about whether author tags should get copied
to a new page. For now, tags and authors will still get copied to the new page.

## Testing

- In the local admin, pick a page with tags and authors. This Newsroom Page works:
    <http://localhost:8000/admin/pages/3395/>
- From the "MORE" button, choose "Copy," change the title and slug, and click on
COPY THIS PAGE.

You should see no errors, and if you search for your new title, you should find
the new page, and the previous tag and author should be applied.

## Checklist


alexm118 (approved):

cwdavies (approved):
I like the simplicity of excluding tags and authors fields when copying pages in base model.
Thanks for adding a test for the CFGOV page copy

willbarton (approved):

csebianlander (approved):

chosak:
@higs4281 for this lurker's understanding, when you say

> This sidesteps a broader question about whether author tags should get copied to a new page. For now, tags and authors will still get copied to the new page.

Does this mean, after this PR, tags and authors are still copied? If so, what behavior is changed by this PR?

higs4281:
End behavior is unchanged. The 500 error that interrupts editing workflow goes away.

PR #6232, opened 2021-01-22
Add product and at risk group filters to enforcement action filter page

wpears:
Testing:
 - Go to the enforcement filterable list page in wagtail and check the "filter enforcement products" and "filter enforcement at risk groups" check boxes
 - Filter by various combinations of these on the enforcement filterable page
 - Check that visiting the page by URL with selected params works
 - Check that other filterable list pages (blog, etc.) aren't affected

Wagtail checks:
Screen Shot 2021-01-22 at 2 52 53 PM

Filterable List controls:
Screen Shot 2021-01-22 at 2 52 07 PM

### DIFF
--- a/cfgov/v1/forms.py
+++ b/cfgov/v1/forms.py
@@ -277,6 +277,28 @@ class EnforcementActionsFilterForm(FilterableListForm):
         widget=widgets.CheckboxSelectMultiple()
     )
 
+    products = forms.MultipleChoiceField(
+        required=False,
+        choices=enforcement_action_page.enforcement_products,
+        widget=widgets.SelectMultiple(attrs={
+            'id': 'o-filterable-list-controls_products',
+            'class': 'o-multiselect',
+            'data-placeholder': 'Search for products',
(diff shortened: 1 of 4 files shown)


cfgov/v1/forms.py
```diff
@@ -277,6 +277,28 @@ class EnforcementActionsFilterForm(FilterableListForm):
...
+            'id': 'o-filterable-list-controls_products',
+            'class': 'o-multiselect',
+            'data-placeholder': 'Search for products',
+            'multiple': 'multiple',
+        })
+    )
+
+    at_risk_groups = forms.MultipleChoiceField(
+        required=False,
+        choices=enforcement_action_page.enforcement_at_risk_groups,
```
schbetsy, line 293: Should the choices in the dropdown be in a certain order? They appear random currently.

wpears: Fixed in https://github.com/cfpb/consumerfinance.gov/pull/6241

PR #6188, opened 2020-12-22
Get rid of old browser tests in the cucumber directory

cwdavies:
Browser tests: figure out what we're missing that we're still interested in
Moved new feature files and step_definitions from `cucumber-new` into `cucumber` directory

## Removals

- Deleted old feature files and step_definitions from `cucumber` directory

## Notes and todos

- Create cypress tests from feature files that were in `cucumber-new` directory

## Checklist

- [x] PR has an informative and human-readable title
  - PR titles are used to generate the change log in [releases](../../releases); good ones make that easier to scan.
  - Consider prefixing, e.g., "Mega Menu: fix layout bug", or "Docs: Update Docker installation instructions".
- [x] Changes are limited to a single goal (no scope creep)
- [ ] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [ ] Future todos are captured in comments and/or tickets
- [ ] Project documentation has been updated, potentially one or more of:
  - [This repo’s docs](https://cfpb.github.io/consumerfinance.gov/) (edit the files in the `/docs` folder) – for basic, close-to-the-code docs on working with this repo
  - CFGOV/platform wiki on GHE – for internal CFPB developer guidance
  - CFPB/hubcap wiki on GHE – for internal CFPB design and content guidance

#### Browser testing

Visually tested in the following supported browsers:
- [ ] Firefox
- [x] Chrome
- [ ] Safari
- [ ] Edge 18 (the last Edge prior to it switching to Chromium)
- [ ] Internet Explorer 11 and 8 (via emulation in 11's dev tools)
- [ ] Safari on iOS
- [ ] Chrome on Android

PR #6236, opened 2021-01-26
Deprecate tox-based browser tests

chosak:
#6188 broke `tox -e acceptance`. Instead of fixing it, this PR proposes deprecating this browser testing path in favor of our two other existing alternatives:

- [Cypress tests](https://github.com/cfpb/consumerfinance.gov/blob/eefa4b19152b78ebbe2a233c3b788c970ce9c812/Jenkinsfile#L164-L185) (Jenkins automation triggered by GitHub actions, requires an external running server)
- [yarn run gulp test:acceptance](https://cfpb.github.io/consumerfinance.gov/browser-acceptance-tests/#running-the-current-browser-tests-locally) (no automation exists that ensures this runs properly; also assumes you have a running server already)

The idea behind the `tox -e acceptance` workflow was to avoid having to run a separate server, and have Django do it for us. I propose that the complexity of maintaining this separate path is not worth the upkeep.

Even if the breakage from #6188 is mended (a missed change of invoking [`gulp test:acceptance:protractor`](https://github.com/cfpb/consumerfinance.gov/blob/eefa4b19152b78ebbe2a233c3b788c970ce9c812/cfgov/core/testutils/runners.py#L79) to `gulp:test:acceptance`), the tests still wouldn't pass. The `tox -e acceptance` flow uses a new, empty test
database for testing. This database doesn't have mega menu content, and thus any tests that check the menu won't pass. We could continue to maintain our Python test data to keep these tests running, but at this point we seem to have made the decision to do browser testing against a production-like site.

Benefits of this change are simplification of our tox file and removal of some rarely-used code in Python to trigger gulp.

## How to test this PR

All existing tests and GitHub Actions checks should continue to pass; this only removes code.

## Checklist


Scotchester (approved):
I think it would be good to come back around to doing browser tests with test data, but I'm not sure when or if that will be prioritized, so I think this is good to do at this time.

PR #6239, opened 2021-01-27
Invalidate Akamai cache for EnforcementAPI endpoint

schbetsy:
Invalidate Akamai cache for EnforcementAPI endpoint when the collection of EnforcementActionPages changes.

## Additions

- A listener to the "publish" and "move" signals for the EnforcementActionPage now triggers a `purge_url_from_cache` action (the same action that the `cf.gov-flush-akamai-cache` job uses)

## Notes and todos

- To do: Add tests.

## Checklist

### DIFF
--- a/cfgov/v1/apps.py
+++ b/cfgov/v1/apps.py
@@ -3,7 +3,11 @@ from django.contrib.auth import get_user_model
 from django.contrib.staticfiles import storage
 from django.db.models.signals import post_save
 
-from .signals import user_save_callback
+from wagtail.core.signals import (
+    page_published, page_unpublished, post_page_move
+)
+
+from .signals import break_enforcement_cache, user_save_callback
 
 
 class V1AppConfig(AppConfig):
@@ -25,3 +29,11 @@ class V1AppConfig(AppConfig):
                 (r"""(@import\s*["']\s*(.*?)["'])""", """@import url("%s")"""),
             )),
         )
+
+        from v1.models.enforcement_action_page import EnforcementActionPage
+        page_published.connect(break_enforcement_cache,
+                               sender=EnforcementActionPage)
+        page_unpublished.connect(break_enforcement_cache,
+                                 sender=EnforcementActionPage)
+        post_page_move.connect(break_enforcement_cache,
+                               sender=EnforcementActionPage)
--- a/cfgov/v1/signals.py
+++ b/cfgov/v1/signals.py
@@ -1,8 +1,10 @@
 from datetime import timedelta
 
 from django.core.cache import caches
+from django.urls import reverse
 from django.utils import timezone
 
+from wagtail.contrib.frontend_cache.utils import PurgeBatch
 from wagtail.core.signals import page_published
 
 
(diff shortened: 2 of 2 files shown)


cfgov/v1/views/enforcement_api.py
```diff
@@ -86,6 +92,12 @@ class Meta:
         ]
 
 
+@receiver(page_published, sender=EnforcementActionPage)
```
chosak, line 95: Could I suggest possibly locating these signals handlers in the existing [v1/signals.py](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/v1/signals.py) file? We don't have a uniform policy on this, but there's at least some consistency; for example we also have a [jobmanager/signals.py](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/jobmanager/signals.py) file.

Per [the Django docs](https://docs.djangoproject.com/en/3.1/topics/signals/#connecting-receiver-functions):

> Where should this code live?
>
> Strictly speaking, signal handling and registration code can live anywhere you like, although it’s recommended to avoid the application’s root module and its models module to minimize side-effects of importing code.
>
> In practice, signal handlers are usually defined in a signals submodule of the application they relate to.

schbetsy: When I try to import the `EnforcementActionPage` model into `signals.py`, I get this error: `django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.`. I don't quite understand that one. How can I get around it?

chosak: Ah. I think that's what's being (unclearly) referenced in the rest of the [Django docs link](https://docs.djangoproject.com/en/3.1/topics/signals/#connecting-receiver-functions) that I didn't include:

> Signal receivers are connected in the ready() method of your application configuration class. If you’re using the receiver() decorator, import the signals submodule inside ready().

Re-reading this, it seems like the signal _handlers_ should be defined in signals.py, but the _connection_ of the signals has to happen in apps.py `ready()` because models aren't available until that time. We do that like this in the v1 app:

https://github.com/cfpb/consumerfinance.gov/blob/f9f3d26b9d054b0f819bd8379dc9278112e98d19/cfgov/v1/apps.py#L15

and slightly differently in the jobmanager app:

https://github.com/cfpb/consumerfinance.gov/blob/f9f3d26b9d054b0f819bd8379dc9278112e98d19/cfgov/jobmanager/apps.py#L9-L11

schbetsy: Thanks for your help. Updated in the latest push.

schbetsy:
Question (maybe for @chosak): when I add URLs to be purged from the cache, do they need to be full URLs, e.g. `https://www.consumerfinance.gov/enforcement/actions/`, or can it be a relative path, e.g. `/enforcement/actions/`?

chosak:
> do they need to be full URLs

I believe but am not 100% sure that they need to be full URLs. The Wagtail cache purge code refers to [`page.full_url`](https://github.com/wagtail/wagtail/blob/bf33842f450ef9270a3cf258dafc2b8303b4913b/wagtail/contrib/frontend_cache/utils.py#L108) and our own invalidation command [asks for the full URL](https://github.com/cfpb/consumerfinance.gov/blob/24ae6aca2a8fbd7e8afa124d2a8b91a38c5d17a9/cfgov/v1/management/commands/invalidate_page_cache.py#L7). It's possible this will work with the relative URL, but I'm not sure.

wpears (approved):
Let's try it out!

PR #6242, opened 2021-01-28
Update so that margins auto by default and 170 or 270 classes apply regardless of screen size

alexhoffman617:
Fixes inset images to be correct size and centered on smaller screens

## Changes

- Updates to inset.less

## How to test this PR

1. Visit the following link while running locally: 
http://localhost:8000/start-small-save-up/start-saving/how-to-use-your-tax-refund-to-build-your-emergency-funds/
2. Using Chrome tools change the size dynamically to make sure the page images still look good.

## Checklist


Scotchester (changes requested):
New lines 93–95 need to be removed. They're now overriding the `width` set by `&__170` above.

alexhoffman617:
Ok @Scotchester , those lines are removed.

cfgov/unprocessed/css/molecules/inset.less
```diff
@@ -74,25 +76,25 @@
...
+        width: unit( 170px / @base-font-size-px, em );
+    }
+
+    &__270 {
+        width: unit( 270px / @base-font-size-px, em );
+    }
+
     .respond-to-min( @bp-sm-min, {
         float: right;
         width: unit( 270px / @base-font-size-px, em );
```
Scotchester, line 89: Oops, this one needs to go, too.

```suggestion
```

Scotchester: And then I would also move the `float: right;` into a `&__right` block in the mixin below. We don't need two `.respond-to-min( @bp-sm-min, {…} )` mixins right never to each other.

Scotchester (changes requested):

cfgov/unprocessed/css/molecules/inset.less
Scotchester: Wait, argh. Let me think about this.

Those defaults may be being relied upon by those other components that share the `.m-inset` base class. I'll play around a bit.

alexhoffman617: Is there anything I can do to help you with this. Looking at something, pairing, etc.

Scotchester: Paired with Alex to confirm that we do in fact need a default for some blocks that just use `.m-inset` and rely on the default width and floating of that. Pushed a commit to make the 170px inset image selector more specific.

Scotchester (approved):

PR #6251, opened 2021-02-03
Optionally skip Django migrations during tests

chosak:
This commit restores the ability to optionally skip (most) Django migrations during Python tests. This repository used to have the ability to do this using `tox -e fast`, but this feature was removed some time ago.

With this commit, by setting the `SKIP_DJANGO_MIGRATIONS` environment variable, the Django [`settings.MIGRATION_MODULES`](https://docs.djangoproject.com/en/2.2/ref/settings/#std:setting-MIGRATION_MODULES) will be set to a special object that skips all migrations. Our custom test runner will then run a few specific data migrations which create required test data, for example a single Wagtail Site object.

Running tests this way significant speeds up testing because you avoid the time required to run migrations. This is noticeable during the testing step:

"Creating test database for alias 'default'..."

An anecdotal timing comparison:

```
$ SKIP_DJANGO_MIGRATIONS= time tox -e unittest-current
233.62 real       187.74 user         5.94 sys

$ SKIP_DJANGO_MIGRATIONS=1 time tox -e unittest-current
183.44 real       141.04 user         5.41 sys
```

So we save about 40-50 seconds by not running migrations.

## Screenshots

Docs changes:

image

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/test.py
+++ b/cfgov/cfgov/settings/test.py
@@ -95,3 +95,14 @@ HAYSTACK_CONNECTIONS["default"]["INDEX_NAME"] = (
 
 ELASTICSEARCH_DSL_AUTO_REFRESH = False
 ELASTICSEARCH_DSL_AUTOSYNC = False
+
+if os.getenv('SKIP_DJANGO_MIGRATIONS'):
+    class _NoMigrations:
+        def __contains__(self, item):
+            return True
+
+        def __getitem__(self, item):
+            return None
+
+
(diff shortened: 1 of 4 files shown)


higs4281 (approved):
Thanks for restoring this option.
When I ran `unittest-current` with `SKIP_DJANGO_MIGRATIONS` set, it shaved 36 seconds off the full test run.

Scotchester:
It would be helpful for the documentation to discuss when it is safe/unsafe to use this option.

PR #6256, opened 2021-02-05
ESLint: ignore sort-imports linter rule

anselmbradford:
`sort-imports` can often break as module names and formats change, and under some circumstances can't be fixed through opposing rules colliding. It's mostly noise in our linter I think, and may obfuscate things more, as related packages can't be arranged together if they aren't alphabetical. Let's turn it off.

## Changes

- ESLint: ignore sort-imports linter rule

## How to test this PR

1. `gulp lint` should go from 613 warnings on main, to 574 in this PR.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1267,7 +1267,7 @@ rules:
 
   # Sort import declarations within module.
   sort-imports:
-    - 1
+    - 0
     -
       ignoreCase: false
       ignoreMemberSort: false


.eslintrc
```diff
@@ -1267,7 +1267,7 @@ rules:
 
   # Sort import declarations within module.
   sort-imports:
-    - 1
+    - 0
     -
       ignoreCase: false
       ignoreMemberSort: false
```
Scotchester, line 1273: Can we get rid of the config we no longer need here, as well? (+5 lines below that couldn't be selected for commenting)

Scotchester (approved):
Makes sense to me.

.eslintrc
anselmbradford: We have several others that are turned off but have the options. I originally did this so if the defaults changed in an ESLint upgrade, the rules wouldn't automatically change their defaults. Maaaaaybe something better for its own PR to  remove them for all turned off rules, so that this is just sort-imports.

anselmbradford: Hmm looking at some of the other ones I'm inclined to just leave it. If we turn these on again we don't need to figure out the format of the options.

PR #6257, opened 2021-02-05
Enforcement database/sidebar rollbacks

wpears:
HOLD


cfgov/jinja2/v1/_includes/molecules/enforcement-action-metadata.html
```diff
@@ -91,7 +93,7 @@ <h3 class="h4">
       </div>
     {% endif %}
     {{_dispositions(page.enforcement_dispositions.all())}}
-    {% if defendant_types|length > 100 %}
+    {% if show_enf_data and defendant_types|length > 0 %}
```
schbetsy, line 96: Why guard these fields with an `if` statement instead of removing them entirely?

wpears: My thinking was that the data is still present in Wagtail and Enforcement may conceivably want to pursue some graphs/visualization on their terms in the future. Keeping them guarded makes that potential future work faster and also  shows that there's lots of data available.

schbetsy: I see. I was under the impression we were going to remove the model fields for the metadata they don't want to show, since they won't be including them in the content template in the future. But I guess we could leave them on the model. 

But if they won't even be providing us the metadata for future enforcement actions, it seems unlikely they'll want to show it or use it for analysis in the future, right?

wpears: I'm unsure if they won't be providing that metadata, I assumed they still would be.

schbetsy (approved):
We'll leave the fields on the model for now

PR #6259, opened 2021-02-09
Paying for College: Don't double-import design system packages

anselmbradford:
Paying for college CSS is importing `cfpb-` less files, which is leading to strange behavior in the header and layout. Instead this PR changes those imports to references so that only the parts required elsewhere will be included.

## Changes

- Paying for College: change less imports to reference.

## How to test this PR

2. Visit http://localhost:8000/paying-for-college/your-financial-path-to-graduation/ and compare to https://www.consumerfinance.gov/paying-for-college/your-financial-path-to-graduation/

## Screenshots

Before:
Screen Shot 2021-02-09 at 9 27 45 AM

After:
Screen Shot 2021-02-09 at 9 27 55 AM

Before:
Screen Shot 2021-02-09 at 9 28 11 AM

After:
Screen Shot 2021-02-09 at 9 28 19 AM

### DIFF
--- a/cfgov/unprocessed/apps/paying-for-college/css/main.less
+++ b/cfgov/unprocessed/apps/paying-for-college/css/main.less
@@ -6,13 +6,13 @@
 // Project-specific Less rules go after the imports.
 
 // Import Capital Framework components.
-@import (less) "cfpb-core.less";
-@import (less) "cfpb-buttons.less";
-@import (less) "cfpb-forms.less";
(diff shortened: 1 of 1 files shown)


Scotchester (approved):
This does what it sets out to do, but @mistergone, it would be nice to move from ye olde `cf-notifications` patterne that PFC has in its `cf-enhancements.less` to the cfpb-notifications package. That would enable getting rid of PFC's `cf-theme-overrides.less`.

PR #6262, opened 2021-02-10
Add support for Elasticsearch-based filterable lists

alexm118:
Swaps our functionality for searching with FilterableLists to use Elasticsearch instead of postgres. 

---

## Additions

- Elasticsearch Index, Document, and Search classes. 
- Concept of "filterable_root" which defines the parent to which the filterable lists searches under
- "filterable_categories" added to the base FilterableMixin class to enable searching through only specific categories sitewide. 

## Changes

- No longer search using django's ORM. 

## How to test this PR

1. Visit any of our pages that leverage the FilterableList and search
2.  Compare results with production to see if there are any notable gaps

## Notes and todos

- Setup Flag to enable toggling functionality
- Unit Tests
- Determine how we want to search the title - Do we want to use a wildcard to match or only match exact terms?

## Checklist

#### Browser testing

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Safari on iOS
- [ ] Chrome on Android

#### Accessibility

- [ ] Screen reader friendly

#### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


cfgov/v1/documents.py
```diff
@@ -0,0 +1,201 @@
...
+        return AbstractFilterPage.objects.live().public()
+
+    def prepare_content(self, instance):
+        try:
+            content_field = instance._meta.get_field('content')
+            value = content_field.value_from_object(instance)
+            content = content_field.get_searchable_content(value)
+            content = content.pop()
+            return content
+        except Exception:
```
chosak, line 43: [PEP8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations) recommends catching "specific exceptions whenever possible". What is being caught here? Can `Exception` be more specific?

cfgov/v1/documents.py
```diff
@@ -0,0 +1,201 @@
...
+    content = fields.TextField()
+    date_published = fields.DateField(attr='date_published')
+    url = fields.KeywordField()
+    start_dt = fields.DateField()
+    end_dt = fields.DateField()
+    statuses = fields.KeywordField()
+    initial_filing_date = fields.DateField()
+
+    def get_queryset(self):
+        return AbstractFilterPage.objects.live().public()
```
chosak, line 34: If you add [`.specific()`](https://docs.wagtail.io/en/stable/reference/pages/queryset_reference.html?#wagtail.core.query.PageQuerySet.specific) here, it should pull back all specific instances, avoiding the need to do `instance.specific` several times down below.

cfgov/v1/documents.py
```diff
@@ -0,0 +1,201 @@
...
+            return None
+
+    def prepare_url(self, instance):
+        return instance.url
+
+    def prepare_start_dt(self, instance):
+        try:
+            return instance.specific.start_dt
+        except Exception:
+            return None
```
chosak, line 53: Instead of repeating this pattern, could you instead do

```py
return getattr(instance.specific, 'start_dt', None)
```

Using an `except Exception` here means that any errors will be caught, including those that are unrelated to whether the instance has a particular attribute. For example, if there's some database connectivity issue, that would be caught here, where we'd prefer that it get raised.

cfgov/v1/documents.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def __init__(self,
+                 prefix='/', topics=[], categories=[],
+                 authors=[], to_date=None, from_date=None,
+                 title='', archived=None, statuses=[]):
+        super(EnforcementActionFilterablePagesDocumentSearch, self).__init__(
+            prefix, topics, categories, authors,
+            to_date, from_date, title, archived
+        )
+        self.statuses = statuses
```
chosak, line 185: There's a lot of duplication here. Could this method be made more generic, e.g.

```py
def __init__(self, **kwargs):
    self.statuses = kwargs.pop('statuses')
    super().__init__(**kwargs)
```

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -880,6 +880,7 @@ def get_filterable_topics(filterable_page_ids, value):
             return []
 
     def get_context(self, value, parent_context=None):
+
```
chosak, line 883:
```suggestion
```

cfgov/cfgov/settings/local.py
```diff
@@ -106,4 +106,4 @@
     "placekitten.com",
 )
 
-ELASTICSEARCH_SYNONYMS_HOME = './search/resources'
+ELASTICSEARCH_SYNONYMS_HOME = './search/resources'
```
chosak, line 109: Could you please remove this change?

cfgov/v1/forms.py
```diff
@@ -278,10 +302,20 @@ class EnforcementActionsFilterForm(FilterableListForm):
...
+            authors=self.cleaned_data.get('authors'),
+            to_date=self.cleaned_data.get('to_date'),
+            from_date=self.cleaned_data.get('from_date'),
+            title=self.cleaned_data.get('title'),
+            statuses=self.cleaned_data.get('statuses')).search()
+
+        # query = self.generate_query()
+        # return self.filterable_pages.filter(query).distinct().order_by(
+        #     '-initial_filing_date'
+        # )
```
chosak, line 318:
```suggestion
```

cfgov/v1/forms.py
```diff
@@ -297,6 +331,17 @@ def get_query_strings(self):
...
 class EventArchiveFilterForm(FilterableListForm):
+
+    def get_page_set(self):
+        return EventFilterablePagesDocumentSearch(
+            prefix=self.filterable_root,
+            topics=self.cleaned_data.get('topics'),
+            categories=self.cleaned_data.get('categories'),
+            authors=self.cleaned_data.get('authors'),
+            to_date=self.cleaned_data.get('to_date'),
+            from_date=self.cleaned_data.get('from_date'),
```
chosak, line 342: Shouldn't this use `start_dt` and `end_dt`?

chosak (commented):
I took a quick first pass through this. A high-level question: are you intentionally keeping the existing behavior (search only searches title field) and envisioning a second iteration that expands search to content fields? I see that you've added a `content` field and a `prepare_content` method but those don't seem to be used yet.

cfgov/v1/documents.py
alexm118: Yeah I meant to force this exception to catch the specific type - it occurs on the case the specific Page Model doesn't have the field being referenced.

cfgov/v1/documents.py
alexm118: Thats way better! I also dislike repeating try blocks and prefer to implement a more functional accessor. Thanks!

cfgov/cfgov/settings/local.py
alexm118: Will do!

cfgov/v1/forms.py
alexm118: I think I'll reenable this section and use a flag to indicate which behavior needs to occur

cfgov/v1/forms.py
alexm118: Oddly enough, the form itself passed the data in as `to_date` and `from_date`. I opted to maintain the values here. We specify start_dt and end_dt as part of the search object within the `filter_dates` function on the Event specific search class.

cfgov/v1/documents.py
alexm118: Done!

cfgov/cfgov/settings/local.py
alexm118: Should be fixed now.

chosak (approved):
Let's merge! Can we rename this PR to something a little more accurate, something like "Add support for Elasticsearch-based filterable lists"? Since we're not actually ready to turn this on (yet).

PR #6270, opened 2021-02-12
Block on Elasticsearch updates during tests

chosak:
This commit adds a new TestCase mixin for use during tests of Elasticsearch using django-elasticsearch-dsl. By default, updates to Elasticsearch are not blocking. This can lead to some unpredictability if we do tests like this:

1. Put some data in the database
2. Index this data into Elasticsearch (e.g. by calling `manage.py search_index`)
3. Do some tests that query Elasticsearch

This is because, even though the code in step 2 may return, there's no guarantee (by default) that the Elasticsearch index has finished and will be reflected in queries.

ES does provide the ability to force this, by using [`refresh='wait_for'`](https://www.elastic.co/guide/en/elasticsearch/reference/master/docs-refresh.html), but unfortunately currently dj-es-dsl does not provide a way to specify this option when updating.

This command adds a TestCase mixin that patches the dj-es-dsl update command to add this parameter, forcing updates to work as if they are blocking.

It also does a bit of cleanup/reorganization of the FilterableListForm tests, using the standard Django convention of calling `form.is_valid` to do cleaning, instead of manually calling `form.full_clean`.

## Checklist

### DIFF
--- a/cfgov/search/elasticsearch_helpers.py
+++ b/cfgov/search/elasticsearch_helpers.py
@@ -1,4 +1,5 @@
 import re
+from mock import patch
 
 from django.conf import settings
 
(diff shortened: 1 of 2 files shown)


alexm118 (approved):
LGTM - Thanks for tackling this!

PR #6271, opened 2021-02-12
Use unittest.mock instead of mock package

chosak:
Currently we install the [mock](https://pypi.org/project/mock/) package as part of our test requirements. Starting in Python 3.3, this package is part of the standard library as [`unittest.mock`](https://docs.python.org/3/library/unittest.mock.html).

This commit replaces all uses of `mock` with `unittest.mock` and removes the test dependency on the external package.

## How to test this PR

All tests pass!

## Checklist

### DIFF
--- a/cfgov/agreements/tests/test_legacy.py
+++ b/cfgov/agreements/tests/test_legacy.py
@@ -1,11 +1,10 @@
 import random
+from unittest.mock import patch
 
 from django.core.paginator import Page
 from django.http import HttpResponse
 from django.test import TestCase
 from django.urls import reverse
-
-from mock import patch
 
 from agreements import models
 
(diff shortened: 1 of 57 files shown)


cwdavies (approved):
This change is necessary when we upgrade from Python 2 to Python 3.
I verified all tests pass using `tox -e lint` and `tox -e unittest-current`
Ran 1710 tests in 379.510s
OK
  unittest-current: commands succeeded
  congratulations :)

cwdavies:
Should we create a PR to start testing Python 3.3 in Docker by adding a new python3 container?

chosak:
@cwdavies our default Python version is already 3.6.

cwdavies:
> @cwdavies our default Python version is already 3.6.

@chosak Yes, that's right we are already on a new version of Python! It's Django that is on 2.2.

PR #6272, opened 2021-02-17
Elasticsearch-based filters suggestions

chosak:
Suggestions for #6262:

- Simplifying tests and removing some duplication.
- Removing indexing of content field since we're not using it yet and it'll require a more complicated implementation to handle different fields (the header) on different pages.
- Use [`PageQuerySet.specific()`](https://docs.wagtail.io/en/stable/reference/pages/queryset_reference.html#wagtail.core.query.PageQuerySet.specific) to optimize indexing.

### DIFF
--- a/cfgov/search/elasticsearch_helpers.py
+++ b/cfgov/search/elasticsearch_helpers.py
@@ -1,7 +1,9 @@
 import re
+import sys
 from unittest.mock import patch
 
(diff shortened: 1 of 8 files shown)


alexm118 (approved):
LGTM - Like the function approach to rebuilding the index, Makes sense to remove content for now.

PR #6274, opened 2021-02-17
Restore two Ask CFPB Spanish translations

Scotchester:
This PR corrects the following two translation regressions:

**1. "Read answer" on the landing page**

This translation was automatically commented out in the translation file for unknown reasons. I manually restored it.

**2. "Was this answer helpful to you?" on the feedback module**

It appears that this translation was automatically commented out due to the text being un-hard-coded in our templates and moved to a field in Wagtail. I manually restored it.

I'm not sure if either will get commented again when someone else runs `makemessages`. We really need to figure out a way to prevent that from happening in the future.

## How to test this PR

1. Pull branch
2. Open http://localhost:8000/es/obtener-respuestas/ and observe that "Read answer" links are now translated (compare to the untranslated string in production)
3. Go into an answer and observe that "Was this answer helpful to you?" is now translated (compare to the untranslated string in production)

Or wait for the dev Docker stack deployment and check it out there.

## Checklist


willbarton:
> I'm not sure if either will get commented again when someone else runs makemessages. We really need to figure out a way to prevent that from happening in the future.

The way to prevent that from happening is to ensure that they're discoverable by Django. In this case, I suspect [it's this issue](https://cfpb.github.io/consumerfinance.gov/translation/#do-not-include-the-_-call-in-a-larger-jinja2-template-data-structure). Indeed, `makemessages` finds it if I change those lines to:

```jinja2
{% set read_answer_text = _("Read answer") %}
{% for page in portal.featured_answers %}
    <li class="m-list_item">
    {{ partially_styled_link.render({
        'plain_text': page.question,
        'underlined_text': read_answer_text,
        'url': page.url })
    }}
    </li>
{% endfor %}
```

For the content entered in Wagtail... I'm not sure how to make that work with makemessages. Where does it get entered? I suspect this is a problem where we're trying to user gettext in a way that we can't quite do with Django to translate a user-entered string.

willbarton (changes requested):
I would prefer to fix the underlying issues with the translatable strings that makes them not discoverable by `makemessages` — otherwise, our translations will become unmaintainable.

Also, the `validate-translations` test will pass 😁

willbarton (approved):

PR #6279, opened 2021-02-19
Elasticsearch test improvements

chosak:
#6262 added some new Python tests that rely on a running Elasticsearch server. This commit changes test behavior so that, when running locally, these tests are skipped if ES is not running. These tests are still required to pass on GitHub Actions.

This commit also does a bit of refactoring to remove a `time.sleep` in one of the tests.

## How to test this PR

To test, first confirm that all tests pass when running with a local Elasticsearch server (exposed at host `ES7_HOST` and port `ES_PORT`).

Next, stop your ES server and confirm that all tests pass, although the ones that depend on ES are skipped. You should see something like "OK (skipped=9)" in your test output.

Next, with ES still stopped, define the `GITHUB_ACTIONS` environment variable, and re-run. You'll see that the tests now (correctly) fail. (The `GITHUB_ACTIONS` environment variable is [always set when running on GitHub Actions](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables).)

## Screenshots

New documentation:

image

## Checklist


alexm118 (approved):
Guess you can't sleep anymore 😂

PR #6281, opened 2021-02-22
Hide archive filter options in the filterable UI

cwdavies:
CE wants the archive filter options in filterable lists to be hidden.
They are placed behind a feature flag so that it's possible to enable and disable if we need to.

## Additions

- `HIDE_ARCHIVE_FILTER_OPTIONS` feature flag

## Changes

- Modified `filterable_list_mixins` model to use `HIDE_ARCHIVE_FILTER_OPTIONS` flag

## How to test this PR

1. Enable `HIDE_ARCHIVE_FILTER_OPTIONS` flag in Wagtail admin
2. Navigate to `/about-us/blog/` page and look at filter posts
3. Click on Show filters on `/data-research/research-reports/` page and look at filter reports
4. Navigate to `/about-us/newsroom/` and look at filter posts
5. Verify archive options are removed from filterable UI

## Notes and todos

- demo this feature to CE before it goes live

## Checklist

#### Browser testing

- [ ] Firefox
- [x] Chrome
- [ ] Safari
- [ ] Safari on iOS
- [ ] Chrome on Android

#### Accessibility

- [ ] Screen reader friendly

#### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


cfgov/cfgov/settings/base.py
```diff
@@ -764,6 +764,8 @@
     # Controls whether or not to include Qualtrics Web Intercept code for the
     # Q42020 Ask CFPB customer satisfaction survey.
     "ASK_SURVEY_INTERCEPT": [],
+    # Hide archive filter options in the filterable UI
+    "ARCHIVE_FILTER_OPTIONS": [("boolean", False)],
```
willbarton, line 768: We don't want to force it off in settings. This should be empty. We'll need to manually add the flag in production *before* this code gets there so that it doesn't get turned off unintentionally.
```suggestion
    "ARCHIVE_FILTER_OPTIONS": [],
```

cfgov/v1/jinja2tags/__init__.py
```diff
@@ -68,7 +69,8 @@ def is_filter_selected(context, fieldname, value):
 
     # Dirty hack to check the default option for the `archived` filter
     if fieldname == 'archived' and value == 'include':
-        return True
+        if flag_enabled('ARCHIVE_FILTER_OPTIONS'):
+            return True
```
willbarton, line 73: I don't think this is necessary. It's sufficient to ensure that the field doesn't appear.

cfgov/v1/models/filterable_list_mixins.py
```diff
@@ -95,7 +97,10 @@ def get_context(self, request, *args, **kwargs):
 
         form_data, has_active_filters = self.get_form_data(request.GET)
         queryset = self.get_filterable_queryset()
-        has_archived_posts = queryset.filter(is_archived='yes').count() > 0
+        if flag_enabled('ARCHIVE_FILTER_OPTIONS'):
```
willbarton, line 100: Can we add a comment above here that this flag check is so that we can enable or disable the the archive filter options?

Also, if you pass `flag_enabled` the request, it'll possible to condition it on request-dependent properties in the future.
```suggestion
        if flag_enabled('ARCHIVE_FILTER_OPTIONS', request=request):
```

willbarton (commented):
This looks good — I have a couple comments, mostly

cfgov/cfgov/settings/base.py
cwdavies: Updated so archive filter options is not turned off by default

cfgov/v1/jinja2tags/__init__.py
cwdavies: Good catch. This is not needed.

cfgov/v1/models/filterable_list_mixins.py
cwdavies: Added comment and request to flag check

cfgov/cfgov/settings/base.py
```diff
@@ -764,6 +764,8 @@
     # Controls whether or not to include Qualtrics Web Intercept code for the
     # Q42020 Ask CFPB customer satisfaction survey.
     "ASK_SURVEY_INTERCEPT": [],
+    # Hide archive filter options in the filterable UI
+    "ARCHIVE_FILTER_OPTIONS": [],
```
willbarton, line 768: One other minor thing: I think the flag name should be clearer that it disables the archive filter options, so maybe `DISABLE_ARCHIVE_FILTER_OPTIONS`.

cfgov/v1/models/filterable_list_mixins.py
```diff
@@ -95,7 +97,11 @@ def get_context(self, request, *args, **kwargs):
 
         form_data, has_active_filters = self.get_form_data(request.GET)
         queryset = self.get_filterable_queryset()
-        has_archived_posts = queryset.filter(is_archived='yes').count() > 0
+        # flag check to enable or disable archive filter options
+        if flag_enabled('ARCHIVE_FILTER_OPTIONS', request=request):
+            has_archived_posts = False
+        else:
+            has_archived_posts = queryset.filter(is_archived='yes').count() > 0
```
willbarton, line 104: I suspect to pass the coverage test, you'll need to copy the test that covers this method with `@override_settings` to enable the flag, as well as with the flag disabled.

cfgov/cfgov/settings/base.py
cwdavies: Done

willbarton:
@cwdavies I'm sorry! I just realized the Wagtail Flags UI uses "Enable" and "Disable" so the button would be "Disable `DISABLE_ARCHIVE_FILTER_OPTIONS`", which is really awkward. Could we name the flag `HIDE_ARCHIVE_FILTER_OPTION`?

cwdavies:
> @cwdavies I'm sorry! I just realized the Wagtail Flags UI uses "Enable" and "Disable" so the button would be "Disable `DISABLE_ARCHIVE_FILTER_OPTIONS`", which is really awkward. Could we name the flag `HIDE_ARCHIVE_FILTER_OPTION`?

I renamed the archive filter flag to be prefixed with `HIDE_`. We should squash commits before merge to remove rename flag.

cfgov/v1/models/filterable_list_mixins.py
cwdavies: Added `HIDE_ARCHIVE_FILTER_OPTIONS` flag to `v1/tests/models/test_filterable_list_mixins.py`

cfgov/v1/tests/models/test_filterable_list_mixins.py
```diff
@@ -46,6 +46,16 @@ def test_get_context_raises_exception_for_super_obj_has_no_get_context(
...
 
+    @override_settings(FLAGS={"HIDE_ARCHIVE_FILTER_OPTIONS": [("boolean", True)]})
+    def test_get_context_raises_exception_for_super_obj_has_no_archive_context(
+        self
+    ):
+        self.assertRaises(
+            AttributeError,
+            self.mixin.get_context,
+            request=self.factory.get("/")
+        )
```
willbarton, line 57: I don't think this test is going to touch the `flag_enabled` conditional. 

At a glance, it actually doesn't look like we have a test that directly tests the `get_context` method, but you could add two that check that `has_archived_posts` is `False` even when the queryset has objects that meet the condition, and then another test that sets has_archived_posts as expected.

Scotchester (approved):

cfgov/v1/tests/models/test_filterable_list_mixins.py
cwdavies: Paired with @Scotchester to directly test the `get_context` method.

PR #6284, opened 2021-02-23
Update default video thumbnail to better fit dimensions

anselmbradford:
## Changes

- Adjust default no-js video thumbnail image dimensions to show the CFPB logo at all screensizes.

## How to test this PR

1. Pull branch and run gulp clean && gulp build
2. Visit a video page, such as http://localhost:8000/consumer-tools/everyone-has-a-story/leah-repaying-student-loans/, http://localhost:8000/consumer-tools/reverse-mortgages/, http://localhost:8000/coronavirus/avoiding-scams/
3. Open the dev console settings and disable JavaScript.
4. Reload the page to show the default video image.
5. Resize the window and note that the image is cropped at certain sizes.

## Screenshots

Before:
Screen Shot 2021-02-23 at 2 25 38 PM

After:
Screen Shot 2021-02-23 at 1 50 11 PM

Before:
video-default-old

After:
video-default


cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -26,6 +26,11 @@
 .o-video-player {
     overflow: hidden;
     background: @white;
+
+    // When JavaScript is disabled add a border to stand-alone videos.
+    .no-js &:not( .o-featured-content-module_visual ) {
+        border: 1px solid @black;
```
Scotchester, line 32: We don't ever use black as a border color, do we? How about one of the grays?

Scotchester (changes requested):
Minor question about the border color, but one other thing (maybe two) has to happen.

1. Can we crop the image down to just the logo and center it on the white background?
2. Whether the answer to (1) is yes or no, the image needs to be compressed. I'm happy to assist with that once we determine if the image can be cropped or not.

Scotchester:
Also realizing that it seems kinda problematic that the preview image on a standalone video player is so much taller now. We want to ensure that stays at a 16:9 aspect ratio.

anselmbradford:
> Can we crop the image down to just the logo and center it on the white background?

The way it's set up currently is that the bounding box adjusts to the image:

Screen Shot 2021-02-23 at 6 43 47 PM

Screen Shot 2021-02-23 at 6 43 56 PM

So if we wanted to do what you're describing, we'd need to convert the `img` to a `div` and set the image as a background.

Scotchester:
> So if we wanted to do what you're describing, we'd need to convert the `img` to a `div` and set the image as a background.

Well, it's already in a `div`, but this gives me two ideas:

1. Reduce the `img` down to just the logo with no padding, as suggested in my previous comment, then change the background color of `div.o-video-player_image-container` to white and center the image vertically (it's already being centered horizontally when it's cropped on the left and right). I don't think the current black background is ever seen, except for perhaps milliseconds during loading?
2. Just make it the background image of `div.o-video-player_image-container` and have the preview image from YouTube load in on top of it.

Which do you think would be preferable?

anselmbradford:
@Scotchester Check https://github.com/cfpb/consumerfinance.gov/pull/6284/commits/30041d08982545581f01bb48a1406c1f86b77d10 for the latest. 

Few general notes:
 - We really shouldn't mix the video player and featured content module classes together. Aside from all the encapsulation difficulties, we also contend with having the video-player.less in this repo and the featured-content-module.less in the design-system repo. We should probably at least move the video player to the Design System, and eventually aim to untangle those two (or combine them further).
  - Events use the same default image, so I couldn't delete the old one without doing CSS work there, which I think is out of scope here. See, for example, https://www.consumerfinance.gov/about-us/events/archive-past-events/2021-release-of-the-taskforce-on-federal-consumer-financial-law-report/

contolini (approved):
Some 🧙‍♂️ ✨ . The only issue I see is that IE10 gets the no-JS treatment and doesn't support `:before` so the thumbnails aren't vertically-centered. Fortunately, it's IE10 and already a hot mess so users probably won't notice.

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -83,39 +82,29 @@
...
-        background: #fff;
-        display: flex;
-
-        img {
-            padding: 10px;
-            box-sizing: border-box;
-            align-self: center;
-        }
+    // Using the default CFPB cover card image.
+    &[src="/static/img/cfpb_video_cover_card_954x200.png"] {
```
Scotchester, line 89: This can't be exact because filenames have cachebusting random strings appended in production.

anselmbradford: Fixed in https://github.com/cfpb/consumerfinance.gov/pull/6284/commits/ed57f474b42f7c1cffa8b126931b185542baf46a

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -48,18 +53,52 @@
...
+
+    // Create an aspect ratio box, see https://css-tricks.com/aspect-ratio-boxes/#the-psuedo-element-tactic
+    &:before {
+        content: '';
+        width: 1px;
+        margin-left: -1px;
+        float: left;
+        height: 0;
+        // 1280px / 720px (16/9) aspect ratio.
+        padding-top: 720px / 1280px * 100%;
```
Scotchester, line 69: No need for the pixel values, we can just make this abstract.

```suggestion
        // Calculate padding to set the height of the container proportionally
        // so that we end up with a 16:9 aspect ratio.
        padding-top: 9 / 16 * 100%;
```

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -48,18 +53,52 @@
...
+        align-self: center;
+
+        // Overrides values set in featured-content-module.less file in the DS.
+        &.o-featured-content-module_img {
+            max-width: 100%;
+            height: auto;
+            left: 0;
+            transform: none;
+        }
+    }
```
Scotchester, line 101: Suggesting increasing the padding when used alone (as on http://localhost:8000/consumer-tools/everyone-has-a-story/leah-repaying-student-loans/) and then overriding to bring it down to something more reasonable when in an FCM (suggesting bumping that to 15px since we use that value more often).

Also, should these values be calculated from `@grid_gutter-width`?

```suggestion
    &[src^="/static/img/cfpb_video_cover_card_954x200"] {
        padding: 60px;
        box-sizing: border-box;
        align-self: center;

        // Overrides values set in featured-content-module.less file in the DS.
        &.o-featured-content-module_img {
            max-width: 100%;
            height: auto;
            padding: 15px;
            left: 0;
            transform: none;
        }
    }
```

Scotchester (changes requested):
A couple minor suggestions, but something else happened when I tried to test this locally: I couldn't run `gulp build` successfully because some JS in PFC was looking for jQuery, which is presumably because it was removed in #6314. I'm not sure why builds aren't failing everywhere. Ah, the mysteries of Node module caching!

anselmbradford:
> A couple minor suggestions, but something else happened when I tried to test this locally: I couldn't run gulp build successfully because some JS in PFC was looking for jQuery, which is presumably because it was removed in #6314. I'm not sure why builds aren't failing everywhere. Ah, the mysteries of Node module caching!

Yikes! Sorry for the inconvenience. This is fixed in main now though, thanks to https://github.com/cfpb/consumerfinance.gov/pull/6285. Updating to bring that change in here.

cfgov/unprocessed/css/organisms/video-player.less
anselmbradford: Updated in https://github.com/cfpb/consumerfinance.gov/pull/6284/commits/14b49c067cee32f11f3bc5a4b5e87bc495a8415e to use ems. 

> Also, should these values be calculated from `@grid_gutter-width`?

Hmmm these seem kind of arbitrary values that fit with the image but happen to be divisible by `@grid_gutter-width`, right? I've left them as hardcoded value (but converted to ems).

Scotchester: Yeah, fair point.

Scotchester (approved):

PR #6244, opened 2021-01-28
Send searches on Spanish-language pages to the Spanish search endpoint, update retirement to 0.15.2

Scotchester:
Currently, all searches using the global search box go to the English search. With these updates, Spanish pages will now send searches to our Spanish search index.

It also fixes the width of the button on the Spanish global search form to not cut off the label.

---

## Additions

- `language` parameters in template macros where they didn't always exist before
- Conditionals to change the Search.gov `affiliate`/`siteHandle` properties to `cfpb_es` if the `language` context variable is set to `'es'`
- `show_mega_menu` and `language` params on the Django page header template helper (see cfpb/retirement#268)
- New styles to fix Spanish search button width

## Removals

- Duplicative Search.gov script on Django template `base_update.html`

## Checklist

- [x] PR has an informative and human-readable title
  - PR titles are used to generate the change log in [releases](../../releases); good ones make that easier to scan.
  - Consider prefixing, e.g., "Mega Menu: fix layout bug", or "Docs: Update Docker installation instructions".
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [ ] Project documentation has been updated, potentially one or more of:
  - [This repo’s docs](https://cfpb.github.io/consumerfinance.gov/) (edit the files in the `/docs` folder) – for basic, close-to-the-code docs on working with this repo
  - CFGOV/platform wiki on GHE – for internal CFPB developer guidance
  - CFPB/hubcap wiki on GHE – for internal CFPB design and content guidance

PR #6286, opened 2021-02-23
Restore mega menu and banners on Django pages

Scotchester:
It disappeared due to the global header change in #6244.

This also restores the Beta banner for this page when on Beta.

## Additions

- Sets default values for `show_mega_menu`, `show_banner`, and `language` when Django pages call the `header.render()` macro

## How to test this PR

1. Load up http://localhost:8000/data-research/consumer-complaints/ and see no mega menu
1. Pull branch
1. Reload and see mega menu
1. Go to https://beta.consumerfinance.gov/data-research/consumer-complaints/ and see no Beta banner
1. Enable `BETA_NOTICE` flag for all requests at http://localhost:8000/admin/flags/BETA_NOTICE/
1. Reload http://localhost:8000/data-research/consumer-complaints/ and see the Beta banner

## Checklist

### DIFF
--- a/cfgov/legacy/views/complaint.py
+++ b/cfgov/legacy/views/complaint.py
@@ -33,7 +33,9 @@ class ComplaintLandingView(TemplateView):
 
         context.update({
             'technical_issues': flag_enabled('CCDB_TECHNICAL_ISSUES'),
-            'ccdb_content_updates': flag_enabled('CCDB_CONTENT_UPDATES')
+            'ccdb_content_updates': flag_enabled('CCDB_CONTENT_UPDATES'),
+            'show_banner': True,
+            'show_mega_menu': True,
         })
 
         return context
--- a/cfgov/legacy/views/complaint.py
+++ b/cfgov/legacy/views/complaint.py
@@ -33,9 +33,7 @@ class ComplaintLandingView(TemplateView):
 
         context.update({
             'technical_issues': flag_enabled('CCDB_TECHNICAL_ISSUES'),
-            'ccdb_content_updates': flag_enabled('CCDB_CONTENT_UPDATES'),
-            'show_banner': True,
-            'show_mega_menu': True,
+            'ccdb_content_updates': flag_enabled('CCDB_CONTENT_UPDATES')
         })
 
         return context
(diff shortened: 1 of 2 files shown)


chosak:
Fixing this way leaves missing mega menus on any other pages that use the base templates, for example:

- https://www.consumerfinance.gov/paying-for-college2/understanding-your-financial-aid-offer/feedback/
- https://www.consumerfinance.gov/paying-for-college2/understanding-your-financial-aid-offer/about-this-tool/
- https://www.consumerfinance.gov/fair-lending/
- https://www.consumerfinance.gov/jobs/supervision/

Ideally we want the header to work properly when included using [`{% include_header %}`](https://github.com/cfpb/consumerfinance.gov/blob/da94cbd7659cd32853c7b046a1c6ff3caa03faf1/cfgov/legacy/templates/front/base_update.html#L160) even if the parent template doesn't explicitly specify those variables. As an alternative, could we provide defaults for the header template rendered [here](https://github.com/cfpb/consumerfinance.gov/blob/121f49cce588f811380054a9296a3c10d0aeabe7/cfgov/v1/jinja2/v1/header.html#L22-L24), something like

```jinja
{{ header.render(
    show_banner=show_banner | default( true ),
    show_mega_menu=show_mega_menu | default( true ),
    language=language | default( 'en' )
) }}
```

willbarton (dismissed):

willbarton:
This seems like a question of using a test to paper over something that should be more robust, but should we have a functional test for the mega menu's existence on pages that don't inherit from our Jinja2 templates? This crops up about once a year, it seems.

Scotchester:
> Ideally we want the header to work properly when included using [`{% include_header %}`](https://github.com/cfpb/consumerfinance.gov/blob/da94cbd7659cd32853c7b046a1c6ff3caa03faf1/cfgov/legacy/templates/front/base_update.html#L160) even if the parent template doesn't explicitly specify those variables. As an alternative, could we provide defaults for the header template

I feel like we thought about this and had some reason why it would be undesirable to default it to true, but I can't for the life of me come up with it now. Happy to switch to your suggested alternative.

chosak (approved):

PR #6294, opened 2021-02-26
Handle uppercase env vars when naming indexes

higs4281:
Elasticsearch requires lowercase names for indexes, and we add prefixes, 
based on the DEPLOY_ENVIRONMENT var, to index names to keep them separate.

The DEPLOY_ENVIRONMENT vars on servers are lowercase, but we're not indexing
from servers anymore. We index from Jenkins, where the vars are configured as uppercase.

Our `elasticsearch_helpers` shouldn't assume the case of the variable.

### DIFF
--- a/cfgov/search/elasticsearch_helpers.py
+++ b/cfgov/search/elasticsearch_helpers.py
@@ -72,10 +72,11 @@ synonym_analyzer = analyzer(
 
 
 def environment_specific_index(base_name):
-    if settings.DEPLOY_ENVIRONMENT in (None, '', 'local', 'production'):
+    env = settings.DEPLOY_ENVIRONMENT
+    if not env or env.lower() in ('local', 'production'):
         return base_name
     else:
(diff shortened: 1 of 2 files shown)


cfgov/search/elasticsearch_helpers.py
```diff
@@ -72,10 +72,11 @@ def get_synonyms():
 
 
 def environment_specific_index(base_name):
-    if settings.DEPLOY_ENVIRONMENT in (None, '', 'local', 'production'):
+    env = settings.DEPLOY_ENVIRONMENT
```
willbarton, line 75: Why don't we call `lower` once here?
```suggestion
    env = settings.DEPLOY_ENVIRONMENT.lower()
```

higs4281: can't call .lower() on None, if that happens.

willbarton: Ah, I see.

willbarton (approved):

PR #6295, opened 2021-03-02
Remove AChecker a11y checker and gulp audit task

anselmbradford:
Per a recent email:

> The AChecker website will shut down on April 30, 2021.

This PR removes references to the AChecker website and removes the `gulp audit` task, as the a11y task was the only audit present (we may want to bring that task back if we bring other audits into the build in the future).

## Removals

- Removes code dependent on the AChecker website.
- Remove `gulp audit` task.

## How to test this PR

1. `gulp` shouldn't error.


contolini (approved):
RIP

PR #6298, opened 2021-03-02
Add memory limits to dev Docker stack containers

Scotchester:
Per Lamin's suggestion at https://GHE/CFGOV/platform/issues/3979

## Checklist


docker-stack.yml
```diff
@@ -29,8 +35,14 @@ services:
         delay: 10s
         max_attempts: 3
         window: 120s
+      resources:
```
alexm118, line 38: Your elasticsearch service failed to start in the docker cluster - my guess is this is insufficient allocation for the container. [Guidelines](https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html) don't seem to specify a requirement, maybe we can look at usage stats for other stacks and see what they are.

Scotchester: Trying 2 GB limits with 1 GB reservations. We'll see how the next deployment goes.

higs4281 (approved):
We could probably whack the Elasticsearch2 services way back, since they're unused in almost all our stacks. Can't wait to drop them entirely. But I think these limits are worth a try.

cwdavies (changes requested):
There might be issues adding the --user option to the docker run command in the Jenkinsfile.
This seems out of scope in this PR to run a container as a specific user and should not impact memory limitations.

Scotchester:
> There might be issues adding the --user option to the docker run command in the Jenkinsfile.
> This seems out of scope in this PR to run a container as a specific user and should not impact memory limitations.

Yeah, I'm gonna revert that. I thought it might have been the solution to this job's workspace being undeletable, but then I realized that it has to be set to begin with so that the created screenshots are created with that user.

cwdavies (approved):
These memory limits and reservations look good to me in docker-stack.yml

PR #6300, opened 2021-03-03
Remove supervision template and assets

anselmbradford:
Supervision page is moving to Wagtail!

## Removals

- Remove supervision template and assets.

## How to test this PR

1. Once this merged the redirect from https://www.consumerfinance.gov/jobs/supervision/ to the new page will be live.


csebianlander (approved):

chosak (approved):
👏 

In order to pass tests, you'll need to remove a test that is now deprecated -- [this whole file](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/v1/tests/test_legacy_pages.py) can be removed.

PR #6305, opened 2021-03-05
Update fetch method to provide expected response type

wpears:
The `axios` `get` call in this instance was (1) not returning the data structure the code was expecting and (2) causing CORS problems on Chrome. Replaced with `cross-fetch`, already included in our root `package.json`

The reason this was working at all was that there's a fallback check from the object lookup where the Census API is queried. The underlying data structure issue meant that *NO* county was being pulled from the FIPS array, but instead was always falling back to what was designated rural by the census, which is sometimes different than our designations.

Testing:
 - Pull branch and run the frontend build
 - Search for `300 N Main Kingfisher OK` in the tool at `http://localhost:8000/rural-or-underserved-tool/`
 - Should be rural
 - Search for `1700 G St NW Washington DC`
 - Should not be rural

### DIFF
--- a/cfgov/unprocessed/apps/rural-or-underserved-tool/js/get-rural-counties.js
+++ b/cfgov/unprocessed/apps/rural-or-underserved-tool/js/get-rural-counties.js
@@ -1,15 +1,15 @@
 require( 'es6-promise' ).polyfill();
-import { get } from 'axios';
+import { fetch } from 'cross-fetch';
 
 /**
  * @param {string} year - A year.
  * @returns {Object} Result of the call to a JSON file.
  */
 function getRuralCounties( year ) {
-  return get(
+  return fetch(
     'https://files.consumerfinance.gov/rural-or-underserved-tool/data/' +
     year + '.json'
-  );
+  ).then( v => v.json() );
 }
 
 export default getRuralCounties;


anselmbradford:
Are we going to need to handle IE11 lack of fetch support, or does axios handle that?

higs4281 (approved):
This is great. I couldn't puzzle out why the counties check was failing.

anselmbradford:
Should `axios` be removed [here](https://github.com/cfpb/consumerfinance.gov/blob/c829d70f9c899ceec7f55c63d242ac50ebeed0a6/cfgov/unprocessed/apps/rural-or-underserved-tool/package.json#L19)?

higs4281:
Well, I'm still getting a CORS error on the PR stack's requests: 
https://cfgov-pr-6305.demo.cfpb.gov/rural-or-underserved-tool/#rural-or-underserved

searching for 
300 Main St, Kingfisher, OK 73750

ascott1:
Hmm, it is working as expected for me @higs4281 (and I don't see a CORs error in the console). Could you have a browser extension that is triggering the error?

higs4281:
Ah, I can't test this from subnet. From Citrix, success:

PROD (failing):
kingfisher_fail_prod

PR6305 (via citrix)
kingfisher_success_PR6305

PR #6307, opened 2021-03-05
Paying for College: Move legacy django-college-costs-comparison tool pages into cf.gov repo

anselmbradford:
There are three legacy pages that use the `front/base_nonresponsive` template:

- https://www.consumerfinance.gov/paying-for-college/manage-your-college-money/
- https://www.consumerfinance.gov/paying-for-college/choose-a-student-loan/
- https://www.consumerfinance.gov/paying-for-college/repay-student-debt/

This PR moves these pages into this repo more or less wholesale. The background image is gone, and the share buttons at the bottom of the page are gone.

## Additions

- Paying for College: Move legacy django-college-costs-comparison tool pages into cf.gov repo.

## How to test this PR

1. Compare local versions of the following to the production pages:
 - http://localhost:8000/paying-for-college/manage-your-college-money/
 - http://localhost:8000/paying-for-college/choose-a-student-loan/
 - http://localhost:8000/paying-for-college/repay-student-debt/
2. Unlike production, the mega menu should be present on those pages in this PR at mobile size.
3. Existing PfC pages should not be broken (https://www.consumerfinance.gov/paying-for-college/your-financial-path-to-graduation) 

## Notes
- There will be padding issues, but this is aiming for slightly better than the current state of those pages.
- Where the Georgia typeface was used, it was converted to Avenir.
- I think https://github.com/cfpb/django-college-costs-comparison can be archived after this is merged. Yes? @higs4281


contolini (approved):
Looks good, the pages definitely need some responsive help but that's outside the scope of this PR.

higs4281 (commented):
This looks great. Thank you for taking this on.
I wonder if we could let product owner Kate Mullan give the pages a peek in the PR stack so there are no surprises?

higs4281:
> can be archived after this is merged.

@ans Yes, indeed, we can finally archive. And we can remove this import: 
https://github.com/cfpb/consumerfinance.gov/blob/main/requirements/libraries.txt#L49

higs4281 (approved):
Kate Mullan says the changes look great.

Could you add the deletion of this import in requirements/libraries.txt?

https://github.com/cfpb/consumerfinance.gov/blob/main/requirements/libraries.txt#L49

higs4281:
Oh, and removing this line in settings should happen along with the requirements deletion:

https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/cfgov/settings/base.py#L104

anselmbradford:
@higs4281 references removed in https://github.com/cfpb/consumerfinance.gov/pull/6307/commits/855bbfd8e5806a488bc7f1c7684d58a751736b38 thanks!

PR #6310, opened 2021-03-06
Remove non-responsive template and assets

anselmbradford:
PR for when all templates are off the `front/base_nonresponsive` template.

## Removals

- Remove non-responsive `front/base_nonresponsive` template.
- Remove gulp tasks used to support non-responsive template.
- Remove `gulp-replace` dependency.
- Refresh yarn cache.

## How to test this PR

1. PR checks should pass.


gulp/tasks/scripts.js
```diff
@@ -208,24 +187,22 @@ function scriptsApps() {
...
+    'scripts:ondemand:footer'
   )
 );
 
+gulp.task( 'scripts:apps', scriptsApps );
+gulp.task( 'scripts:external', scriptsExternal );
+gulp.task( 'scripts:modern', scriptsModern );
+gulp.task( 'scripts:polyfill', scriptsPolyfill );
+gulp.task( 'scripts:admin', scriptsAdmin );
+
```
Scotchester, line 205: Why move these down here?

Scotchester (approved):
One trivial question, but otherwise this looks good to go.

gulp/tasks/scripts.js
anselmbradford: Just a consequence of re-adding the on-demand scripts in a different place. I moved these back, though it shouldn't matter.

PR #6312, opened 2021-03-09
Setup custom signal processing for Elasticsearch

alexm118:
Since we want our Elasticsearch indices to follow the behavior of wagtail pages, and not django models it makes sense for us to create a custom signal handler and hook into wagtail signals instead of relying on the django signals for our events. 

---

## Additions

- `WagtailSignalProcessor` - hooks into the wagtail publish/unpublish and move events to allow us to handle index updating based on page events that mirror what actually occurs within our site. 

## Changes

- Now use the new signal processor for handling index events.

## How to test this PR

1. Create a new page - Like a blog post but save it as a draft
2. Ensure the page does not exist in the index
3. Publish page and confirm it now appears in the index
4. Unpublish page and confirm it no longer appears in the index
5. Move a published page to the Trash site and confirm it is deleted from the index

## Notes and todos

- 

## Checklist


cfgov/search/elasticsearch_helpers.py
```diff
@@ -148,3 +155,32 @@ def rebuild_elasticsearch_index(*models, stdout=sys.stdout):
...
+            registry.delete(instance_to_delete, raise_on_error=False)
+        else:
+            registry.delete(instance, raise_on_error=False)
+
+    def setup(self):
+        # Wagtail Specific Events
+        page_published.connect(self.handle_save)
+        page_unpublished.connect(self.handle_delete)
+        pre_page_move.connect(self.handle_delete)
+        post_page_move.connect(self.handle_save)
```
chosak, line 179: If I understand how this works, after a page is moved, it will get reindexed. If it gets moved to the Trash, but not unpublished, it will get reindexed with [`url = None`](https://github.com/cfpb/consumerfinance.gov/blob/261197d0429513b22711e39c0ae71e0c88cd6f78/cfgov/v1/documents.py#L41-L42). Are there any potential side effects or downsides of doing it this way?

Wouldn't it be better to remove this pages from the index as opposed to letting them hang around? Do we run the risk of some other logic down the road (for example [here](https://github.com/cfpb/consumerfinance.gov/blob/261197d0429513b22711e39c0ae71e0c88cd6f78/cfgov/v1/documents.py#L135-L136) without a prefix) returning those pages?

alexm118: So I opted for this strategy on the off chance we are moving pages somewhere that isn't trash and ensuring they're indexed correctly. You're correct that if a published page is moved to trash and never unpublished it will get a new url of None, but that should be healthy. We set prefix based on the forms `get_filterable_root` [function](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/v1/forms.py#L150). Now - if moving to the trash resulted in an actual url (something like /trash/pages) that would be an actual problem as they could potentially show up in sitewide searches such as activity log.

We could write a custom handler for if the page ends up in the trash section, don't reindex it. I would need to look more into what information we have from the post_page_move hook.

willbarton: I also wonder about this from the other side: if we're moving a page to the Trash, should it not be unpublished? And maybe that's something we should do as part of that move?

chosak: > maybe that's something we should do as part of that move?

I believe that is current editor "best practice" but I suspect not always followed, nor should we rely on it being so.

> will get a new url of None, but that should be healthy. We set prefix based on the forms get_filterable_root function

Even if the current filterable list code filters by root path, it's still the case that these pages remain in the index which feels dangerous. I think our goal should be to keep the index in sync with what we think should be the searchable models in the DB.

alexm118: My local DB shows 5 unpublished pages in the Trash section - which seemed to surprise content editors. So I'm guessing the expectation on their end is that it should be unpublished - maybe automating that might be helpful for editors.

> Even if the current filterable list code filters by root path, it's still the case that these pages remain in the index which feels dangerous. I think our goal should be to keep the index in sync with what we think should be the searchable models in the DB.

This is true - and a solvable problem. Do we want to tack that onto this PR or follow up with it later?

willbarton: > I believe that is current editor "best practice" but I suspect not always followed, nor should we rely on it being so.

Yeah. I was thinking, should we have a `pre` or `post_page_move` signal handler that explicitly unpublishes a page if it's moved to the Trash?

PR #6320, opened 2021-03-11
Filterable Search Enhancements

alexm118:
Enhance filterable search by expanding the ngram tokenizer we use for matching terms, as well as introducing text based searching against the content, preview description, and topics of the results. 

---

## Additions

- New tokenizer that isn't shared with autocomplete - expands the size of which we can match substrings on search results.
- New searchable fields - content, preview description, and tag names.

## Changes

- Now uses a multimatch search to query across several fields instead of just the title field.

## How to test this PR

1. Reindex your elasticsearch cluster
2. Search a term not found in the title but found in the description (I found blogs on pandemic end up resulting in more results)
3. Confirm you see additional results based on new matching logic.

## Notes and todos

-

## Checklist


cfgov/v1/documents.py
```diff
@@ -12,11 +18,25 @@
...
+    'filterable_ngram_tokenizer',
+    tokenizer=tokenizer(
+        'filterable_tokenizer',
+        'ngram',
+        min_gram=1,
+        max_gram=16,
+        token_chars=["letter", "digit", "symbol"]
+    ),
+    filter=['lowercase', token_filter('ascii_fold', 'asciifolding')]
+)
```
willbarton, line 31: Could or should this live in `search.elasticsearch_helpers` along side the existing `ngram_tokenizer` we have there, with naming/commentary on what's different between them?

willbarton (approved):
I think this looks good on a read-through.

I have one minor question on code organization/naming.

cfgov/v1/documents.py
alexm118: I debated that - I initially left it here as I didn't have a good plan for explaining the difference between the two so I felt grouping this analyzer with its usage made more sense. I guess the main difference would be the size of the tokens, as well as the addition of symbols to this analyzer.

cfgov/v1/documents.py
```diff
@@ -116,9 +154,16 @@ def filter_archived(self, search):
         return search.filter("terms", is_archived=self.archived)
 
     def search_title(self, search):
-        return search.query(
-            "match", title={"query": self.title, "operator": "AND"}
-        )
+        if flag_enabled('EXPAND_FILTERABLE_LIST_SEARCH'):
+            query = MultiMatch(
+                query=self.title,
```
higs4281, line 159: I'm not getting expected results from the title search with the flag enabled.

Example: I would expect, with the 'EXPAND_FILTERABLE_LIST_SEARCH' flag enabled, that content from the `preview_description` field would be indexed, but a search for "pagkalat," which is in the search description field of live Blog page 13848, doesn't show up in search. 

I wonder if this line should be 


```suggestion
                query=search,
```

higs4281 (commented):
One question about the title search.

cfgov/v1/documents.py
higs4281: Ah, I see self.title is the search title param. Still, I'm not getting why the search description term isn't found.

Also, is it intended behavior that the "item name" search only matches an exact phrase in a title?

cfgov/v1/documents.py
```diff
@@ -34,6 +54,8 @@ class FilterablePagesDocument(Document):
     statuses = fields.KeywordField()
     initial_filing_date = fields.DateField()
     model_class = fields.KeywordField()
+    content = fields.TextField()
+    preview_description = fields.TextField()
```
chosak, line 58: Is it deliberate that only these fields are included in the search? Some of the filterable pages include other fields; for example, DocumentDetailPage, EventPage, and LearnPage include a [`header`](https://github.com/cfpb/consumerfinance.gov/blob/12bcc7de85ab15ca9c9961ff7724990e745c3680/cfgov/v1/models/learn_page.py#L30-L34) field, EventPage includes several other [body fields](https://github.com/cfpb/consumerfinance.gov/blob/12bcc7de85ab15ca9c9961ff7724990e745c3680/cfgov/v1/models/learn_page.py#L178-L182), etc.

cfgov/v1/documents.py
higs4281: Do I have this right? SublandingFilterablePage (/about-us/blog/) doesn't hit Elasticsearch? The FilterableListMixin just gets querysets?

alexm118: So you will need two flags on - `EXPAND_FILTERABLE_LIST_SEARCH` and `ELASTICSEARCH_FILTERABLE_LISTS` both enabled. 

We also don't explicitly index the search_description but rather the preview_description (I think they are identical values in most cases though)

cfgov/v1/documents.py
alexm118: These we're the fields specified by Sonna as the initial ask

cfgov/v1/documents.py
higs4281: All questions cleared up after enabling the 'ELASTICSEARCH_FILTERABLE_LISTS' flag as well as 'EXPAND_FILTERABLE_LIST_SEARCH'.

I'm now seeing expected results from the search_description field, including my test case, "pagkalat," and multi-term searches.

higs4281 (approved):

cfgov/v1/documents.py
alexm118: Ended up not using the analyzer so deleting it - using a match_phrase_prefix query now.

cfgov/cfgov/settings/base.py
```diff
@@ -767,7 +767,9 @@
     # Hide archive filter options in the filterable UI
     "HIDE_ARCHIVE_FILTER_OPTIONS": [],
     # Enable ES as the backend for FilterableLists
-    "ELASTICSEARCH_FILTERABLE_LISTS": []
+    "ELASTICSEARCH_FILTERABLE_LISTS": [],
+    # Expand ES Filterable List Search
+    "EXPAND_FILTERABLE_LIST_SEARCH": []
```
willbarton, line 772: Incredibly minor thing, can we add a trailing `,` here?
```suggestion
    "EXPAND_FILTERABLE_LIST_SEARCH": [],
```

willbarton (approved):
One minor comment, but I think this looks good to me on a read-through.

cfgov/v1/documents.py
```diff
@@ -135,6 +170,7 @@ def order_results(self, search):
         if not self.title:
             return search.sort('-date_published')[0:total_results]
         else:
+            print(search.to_dict())
```
Scotchester, line 173: Debugging statement left in? It's failing the unit tests for writing to stdout.

Scotchester: Other than that, was this waiting on anything else?

PR #6322, opened 2021-03-12
Notification: add type option and make available on Browse & Learn pages

virginiacc:
Adds the notification module to Browse & Learn page types so it can be used on coronavirus pages. On these page types, we think notifications will be used towards the top of the page (like the FCM), so the module is included in the Header rather than Content section. 

This PR also reintroduces the notification type dropdown that was removed in #4977 since there's interest in using the default (gray) style on coronavirus pages to highlight information in a way that's more attention-grabbing than a well but less alarming than the warning style. The type dropdown only includes the default (gray) and warning styles -- there isn't a clear need for the success or error style -- and the warning style continues to be the default since the fact that it was the only option for 1.5+ years suggests it's likely to be the most common use case. 

**Note:** The default gray notification style is called `Information` in the dropdown to avoid confusion since it's the default style for our notifications but not the default option of the notification Wagtail block.

## Additions

- Notification module on Browse and Learn page types
- Type dropdown in notification module

## How to test this PR

2. Create a new Browse or Learn page, add a notification, and verify that default `Information` style can be applied to it

### Past notification on coronavirus page:
Screen Shot 2021-03-12 at 12 18 41 PM

### Future approach:
Screen Shot 2021-03-12 at 12 25 46 PM

## Notes and todos
-  Should add a test for the notification type

## Checklist


anselmbradford:
> Note: The default gray notification style (m-notification__default) is called Gray in the dropdown to avoid confusion since it's not the default style for the module.

Hmmmmm it sure seems like the default to me. If you preview the page and remove `m-notification__default` it looks the same. Am I missing something? It seems like [this line](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/jinja2/v1/_includes/molecules/notification.html#L49) should be modified to not produce `m-notification__default`.

virginiacc:
@anselmbradford Maybe I just confused things with that note. :D The "Gray" option should result in the default style for the notification molecule, but it's not the default option for the type dropdown, so I thought calling it "Default" in the dropdown might be confusing. Or is there something else not working correctly?

anselmbradford:
> @anselmbradford Maybe I just confused things with that note. :D The "Gray" option should result in the default style for the notification molecule, but it's not the default option for the type dropdown, so I thought calling it "Default" in the dropdown might be confusing. Or is there something else not working correctly?

I see what you're saying, but I think we should follow how this is named in the design system, or change the name there, https://cfpb.github.io/design-system/components/notifications#default-base-notification-1. I know the `default='warning'` code structure looks a little confusing, but a user in Wagtail won't see this:

```
type = blocks.ChoiceBlock(choices=[
        ('default', 'Default'),
        ('warning', 'Warning'),
    ], required=True, default='warning')
```

A problem I see with "Gray" is it presumes the presentation of the default notification. Maybe in the future we discover the contrast isn't right for accessibility reasons or whatever and make the background white. Then we have the appearance baked into the code and it becomes a pain to change if it is already applied to published pages.

virginiacc:
Yeah, I definitely agree about the presentational downside to the term `Gray`. There's discussion about potential more descriptive terms than `Default` for that style in Hubcap #114, but it didn't seem like we'd reached a conclusion about how to convey its purpose / recommended use there so I went with `Gray` since that's what we seem to call it in conversation. Happy to change it if there's a better term or set it back to `Default` if that's the best option!

anselmbradford:
> Happy to change it if there's a better term or set it back to Default if that's the best option!

I'd vote for `Default` to keep the design system as the canonical reference. The other option I can think of is something like `Info`.

virginiacc:
`Information` is an interesting option -- I'll float that vs `Default` with my team on Monday. Thanks for the feedback, @anselmbradford!

virginiacc:
@anselmbradford We decided to go with `Information` for the default type in the dropdown & that's updated in 595397e378ded6aaaaf5823663e646c707a57fe9.

Scotchester (approved):
This looks good to me. You can ignore the Code QL check failure on unrelated existing code.

Scotchester:
P.S.: Please be sure to squash-and-merge it.

virginiacc:
Thanks for reviewing, @Scotchester!

PR #6326, opened 2021-03-16
Fix CodeQA-flagged errors in Python

willbarton:
This PR is intended to fix the errors that CodeQA has flagged. These were predominantly around where we pass parts of URLs off from parameters to redirects without performing any sanitization/validation on them. Most of these changes are in redirects for legacy Ask CFPB and eRegulations URLs to their replacements. 

For the most part, I've fixed this by using Django forms for handling the cleaning/validating of the fields. In Ask CFPB, this means I've created an `AutocompleteForm` and a `SearchForm`, and moved some of the sanitizing that we were doing multiple times in multiple places to the `clean_q` method on that form. 

I've taken a similar approach to the eRegs redirects in creating a `SearchForm`. 

The other place where that CodeQL flagged is in our `check_permissions` method, which handles admin redirects. I've wrapped the destination in `is_safe_url()` to ensure that the the URL is matches the request's allowed hosts.

Where test coverage wasn't, I've added it *before* making the changes to ensure the behavior as it existed remains.

## Checklist

### DIFF
--- a/cfgov/ask_cfpb/forms.py
+++ b/cfgov/ask_cfpb/forms.py
@@ -1,0 +1,32 @@
+from django import forms
+from django.core.validators import RegexValidator
+
+from ask_cfpb.models.search import make_safe
+
+
+legacy_facet_validator = RegexValidator(
+    regex=r'(?:category|audience|tag)_exact:\S+',
+    message='Not a valid legacy facet',
+)
+
+
+class AutocompleteForm(forms.Form):
+    term = forms.CharField(strip=True)
+
+    def clean_term(self):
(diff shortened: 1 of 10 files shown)


alexm118 (approved):
Seems good to me - I'm not sure I love the pattern of introducing a form when looking to handle requests parameters, seems like we avoided it because it bloats the codebase when in use.

willbarton:
I agree that the Django forms API can be a little heavy-handed for processing a querystring, but it avoids us introducing custom cleaning/validation in different places across the codebase. If everything is a Django form, we all know where the cleaning and validating live, and there's a consistent API for doing it.

PR #6327, opened 2021-03-16
Fix error caused by missing PFC image; fix legacy breadcrumb styling

Scotchester:
The moving of old PFC pages to this repo in #6307 did not bring with it one image, `Page_icon.png` that is referenced in a commented HTML `img` tag. Django still looks for that image and outputs its path inside the HTML comment, causing a 500 error if someone tries to load that page in a production environment.

The fix proposed is to use a Jinja comment around the `static` tag, leaving the reference there for consideration in future work on this template.

This PR also fixes the breadcrumb styling on these older pages migrated in #6307, and updates the Gulp config so that gulp-newer correctly triggers a rebuild when `.css` files are changed.

---

## Changes

- Puts `static` tag referencing missing image in Jinja comments so that it doesn't cause a 500 error in production environments
- Fix legacy PFC pages' hero breadcrumb styling
- Tell gulp-newer to look for unprocessed `.css` files (in addition to `.less` files) when determining if the styles should be rebuilt

## How to test this PR

1. Run local server with `DEBUG=False`
1. Open http://localhost:8000/paying-for-college/repay-student-debt/ and see a 500 error
1. Pull branch
1. Reload http://localhost:8000/paying-for-college/repay-student-debt/ and see that the page loads, but has janky breadcrumb styling
1. Run `gulp styles:apps`
1. Reload http://localhost:8000/paying-for-college/repay-student-debt/ and see the correct breadcrumb styling

### Before

Before: bad breadcrumb styling

### After

After: good breadcrumb styling

## Notes

- I am not sure why `tox -e validate-assets` does not return an error (prior to the change in this PR) for the missing image.

## Checklist

#### Browser testing

- [x] Firefox
- [x] Chrome
- [x] Safari
- [x] Safari on iOS
- [x] Chrome on Android


willbarton:
> I am not sure why tox -e validate-assets does not return an error (prior to the change in this PR) for the missing image.

Well, it solves a *slightly* different problem than our 500s when looking up images. 

It [runs `collectstatic` with the production settings](https://github.com/cfpb/consumerfinance.gov/blob/main/tox.ini#L160-L175) to ensure that `ManifestStaticFileStorage` can resolve all references *between* static files. Because `ManifestStaticFileStorage` adds a cache-busting hash to the end of each static filename, it needs to ensure that `.css` files that reference other `.css` files (for example) reference the unique-to-this-build cache-busting filename. It doesn't do this with references that use the API to get static file paths, because it can write those in real-time (so, the use of `static()` in the template).

There's a *little* bit more in [this comment](https://GHE/CFGOV/platform/issues/3021#issuecomment-182305) on a related issue.

`validate-assets` ensures that all static assets that are referenced by other static assets exist so that we don't break **at build time**, because if `ManifestStaticFileStorage` doesn't find referenced files during the `collectstatic` stage, it will error, and that will break the build. It doesn't solve the problem of us returning a 500 when trying to look up non-existent static files in templates.

Scotchester:
Ah! Got it now. Thanks for the explanation, @willbarton.

cfgov/unprocessed/apps/paying-for-college/css/guides.css
```diff
@@ -503,7 +496,7 @@ This just the hero for the Comparison Tool.
 .breadcrumbs a {
     background-color: transparent;
     border-bottom: none;
-    color: #328ed8;
+    /* color: #328ed8; */
```
anselmbradford, line 499: Maybe just delete this?

anselmbradford (approved):
Thanks for cleaning up my mess @Scotchester !!

cfgov/unprocessed/apps/paying-for-college/css/guides.css
Scotchester: D'oh, meant to after testing it. Thanks.

PR #6331, opened 2021-03-17
Notifications: Consolidate type settings | remove unused default class | change 'default' to 'information' behind the scenes

anselmbradford:
`m-notification__default` added in https://github.com/cfpb/consumerfinance.gov/pull/5742 is an unused modifier and can be removed.

## Removals

- `m-notification__default` modifier.

## Changes

- Consolidates notification type and icon settings together to make it clearer what the combonations are.
- Changes `default` notification to `information`, since that's how it is used.
- Minor whitespace cleanup.

## How to test this PR

2. Create a new Browse or Learn page, add a notification, and verify that default Information style can be applied to it
3. Clicking through http://localhost:8000/paying-for-college/your-financial-path-to-graduation should show unchanged notification appearances.

## Notes and todos

- Design system update for default notification is here https://github.com/cfpb/design-system/pull/1212


cfgov/jinja2/v1/_includes/molecules/notification.html
```diff
@@ -30,26 +31,18 @@
...
-    {% set type = ' m-notification__success' %}
-  {% elif type == 'caution' %}
-    {% set type = ' m-notification__error' %}
-  {% elif type %}
-    {% set type = ' m-notification__' + type | string %}
-  {% else %}
-    {% set type = '' %}
-  {% endif %}
+  {% set icon = type_lookup[type | default( 'information' )][1] %}
+  {% set type = type_lookup[type | default( 'information' )][0] %}
```
virginiacc, line 45: I wonder if we'd want to handle bad type values (like `default` would be now) by replacing them with `information` to prevent an error when accessing the lookup?

`set type = type if type in type_lookup else 'information'`

The downside is that passing in an erroneous value would essentially fail silently, but the upside is you'd get usable output instead of an error / broken page.

virginiacc:
@anselmbradford This looks good to me in general!

I assume if notifications had been added with the `default` type we'd need to create a data migration -- but [our first use case](https://www.consumerfinance.gov/coronavirus/managing-your-finances/guide-economic-impact-payments/) for the default style is still using a well, so maybe that's not necessary?

anselmbradford:
> I assume if notifications had been added with the default type we'd need to create a data migration -- but our first use case for the default style is still using a well, so maybe that's not necessary?

Yeah hoping I can get this in before it's in use.

cfgov/jinja2/v1/_includes/molecules/notification.html
anselmbradford: I think we'd want the error. If someone mistyped `errorr` or whatever and ended up with an `information` notification, that would fail silently as you said.

virginiacc (approved):

PR #6337, opened 2021-03-18
Paying for College: Migrate disclosures feedback form to jinja templates

anselmbradford:
Migrates disclosure feedback templates off legacy base_update template and into jinja templates. Drops the social media molecule and uses our standard layout classes.

## Changes

- Paying for College: Migrate disclosures feedback form to jinja templates

## How to test this PR

1.  Visit http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/feedback/ and test entering text and submitting the form. 
2. Compare to https://www.consumerfinance.gov/paying-for-college2/understanding-your-financial-aid-offer/feedback/ 

## Screenshots

Before:
Screen Shot 2021-03-18 at 2 39 38 PM

After:
Screen Shot 2021-03-18 at 2 39 16 PM

Before:
Screen Shot 2021-03-18 at 2 38 33 PM

After:
Screen Shot 2021-03-18 at 2 38 43 PM


Scotchester:
Could we just move this form to a Wagtail page instead?

anselmbradford:
> Could we just move this form to a Wagtail page instead?

Maybe both? Move it off the legacy template here and then move it to Wagtail down the line? 

This page is based on the activity log template. Do we have another page type in Wagtail that doesn't have a sidebar? If not, we'd probably want to rename the `Activity log page` page type and use that, or create a new full-width page type. I'm not sure what the best practice would be. Then maybe it could use our existing feedback module from the ask pages.

Scotchester:
I'd just put it on a Learn page with an empty or nearly-empty sidebar.

higs4281 (approved):
Great improvement. I confirmed that feedback is stored properly to the db.
Thanks for chasing this.

cfgov/paying_for_college/jinja2/paying-for-college/disclosure_feedback.html
```diff
@@ -0,0 +1,61 @@
+{% extends 'layout-full.html' %}
+
+{% block title %}
+    Paying for College - Compare financial aid and college cost | Consumer Financial Protection Bureau
+{% endblock %}
```
Scotchester, line 5: This title should be more specific.

```suggestion
{% block title -%}
    Paying for College feedback | Consumer Financial Protection Bureau
{%- endblock %}
```

cfgov/paying_for_college/jinja2/paying-for-college/disclosure_feedback.html
```diff
@@ -0,0 +1,61 @@
...
+
+{% block content_main %}
+
+<header class="block block__flush-top">
+    <h1>Paying for College</h1>
+    <p class="lead-paragraph">Understanding your financial aid offer</p>
```
Scotchester, line 11: Similarly, the H1 should be more specific, and this lead paragraph becomes superfluous at that point,

```suggestion
    <h1>Give us feedback on Paying for College</h1>
```

Scotchester (commented):
A couple minor suggestions, which should also be applied to the Thanks page.

Scotchester (approved):

PR #6340, opened 2021-03-19
Add reusable text to legacy blog and newsroom pages

niqjohnson:
Occasionally the CFPB mission statement in the blurb at the end of press releases changes. To make it easier to manage those changes, we want to pull that blurb out of the raw HTML content of legacy blog and newsroom pages and put it in a reusable text snippet. That way the blurb just needs to be updated in one place any time it changes.

That blurb is already a reusable text snippet on non-legacy press release pages. This PR adds the ability to insert a reusable text snippet to legacy blog and newsroom pages (they share a template) to match the non-legacy pages. Technically we don't need a reusable text block on legacy blog pages, only legacy newsroom pages, but I don't think it hurts to have that option available (plus it was way easier to do it this way).

One thing to note: on legacy newsroom pages, the reusable text is in a separate `block`, meaning it has a 60 px top margin. On non-legacy newsroom pages, the reusable text is part of the full-width content block, so it has a regular paragraph top margin. I couldn't figure out an easy way to trim down the top margin on legacy newsroom pages, and I also don't think the extra margin is a huge problem, so I'm fine leaving the 60 px margin unless anyone can think of a good way to get rid of it.

---

## Additions

- Reusable text block on legacy newsroom and blog pages

## How to test this PR

1. Go to any legacy newsroom page
2. Add a reusable text block
3. Make sure the migration runs OK (I'm a migration novice)

## Screenshots

legacy-newsroom

## Notes and todos

- In the spirit of keeping our PRs small, I'm only adding the reusable text block to legacy blog and newsroom pages here. I'll do a separate PR for a script to automate removing the press release blurb from the raw HTML content and adding it as a reusable text snippet.

## Checklist


willbarton (approved):
In code, this looks fine to me.

My only question is: is migrating the existing text in legacy blog pages going to be a manual job? Could we do it with a data migration? Would we want to try? How many pages are we talking about?

niqjohnson:
@willbarton: There are about 500 pages that will need the blurb removed from the raw HTML and added as a reusable text snippet. It's a little tricky because the blurb's markup varies a little right now. I was planning to try to script that migration in a separate PR, but do you think it's better to include it in this one?

willbarton:
> I was planning to try to script that migration in a separate PR, but do you think it's better to include it in this one? 

A separate PR would make sense.

PR #6345, opened 2021-03-22
Simple-chart fixed selects

wpears:
Bugfix left out of https://github.com/cfpb/consumerfinance.gov/pull/6334


schbetsy:
This fixes the issue of the start date not showing up in the dropdown when a user presses a time-select button. 👍 

There's a bug (that might be pre-existing; sorry I didn't catch it earlier) where the date in the "start date" dropdown doesn't match the start date of data shown in the chart. The data in the chart start 1-3 months before the date in the dropdown. This happens regardless of whether I use the dropdown or a time selector button.

The "end date" dropdown seems to work as expected.

Screen Shot 2021-03-23 at 11 24 08 AM

wpears:
@schbetsy For testing, I'm gonna call this a won't fix. This is definitely a bug and I don't like the behavior but the solution is not forthcoming unfortunately. I say we allow testing to be kicked off with this code as is and backlog this bug. If users much prefer the time slider, we may not even be using this, so I don't want to delay testing any more to try to get this sorted (after wrestling with it today).

schbetsy (approved):

PR #6349, opened 2021-03-23
Remove the POST from external site interstitial view

willbarton:
This change removes POST support from the external site interstitial view.

The POST was being used to POST URLs that didn't necessarily match the URL in the GET. This was causing issues in production.

The POST also seems unnecessary, as it adds another step through our servers before the redirect, when we can simply link directly to the external site from the button.

This change does require adding `/external-site/` to the parse links exception list, because without it, the user ends up in an endless loop of being sent to /external-site/ when `ParseLinksMiddleware` parses the rendered interstitial page.

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -837,6 +837,9 @@ PARSE_LINKS_EXCLUSION_LIST = [
     r"^/policy-compliance/rulemaking/regulations/\d+/",
     # DjangoRestFramework API pages where link icons are intrusive
     r"^/oah-api/",
+    # External site interstitial (if we're here, the links have already been
+    # parsed)
+    r"^/external-site/",
 ]
 
 # Required by django-extensions to determine the execution directory used by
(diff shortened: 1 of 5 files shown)


cwdavies (approved):
I fully support the removal of POST as there were issues with this in production.

Since the definition of post has been removed from `cfgov/core/views.py`, is there still a need to have a definition of `raise_404` that is now only being referenced in `def get` ?

I am wondering if you can just return `raise Http404` instead of `return self.raise_404()` after the line: `if not form.is_valid():`

Scotchester:
> I fully support the removal of POST as there were issues with this in production.
> 
> Since the definition of post has been removed from `cfgov/core/views.py`, is there still a need to have a definition of `raise_404` that is now only being referenced in `def get` ?
> 
> I am wondering if you can just return `raise Http404` instead of `return self.raise_404()` after the line: `if not form.is_valid():`

Good call, Craig! I'll make that update while Will is away.

wpears (changes requested):
The auto-submitting JS needs to be updated to update the document.location instead of submitting the (now non-existent) form, I'll look into where this code is.

Scotchester:
> The auto-submitting JS needs to be updated to update the document.location instead of submitting the (now non-existent) form, I'll look into where this code is.

Another good catch! Thanks, Wyatt.

Scotchester (approved):
Confirming that Will's changes and Wyatt's addition work for me locally!

wpears (approved):

PR #6350, opened 2021-03-23
Update robots.txt to remove certain parameter pages from crawling

alexandersirris:
Robots.txt helps control search engine crawlers from seeing, accessing, and following certain pages and directories on our site (also controls crawl bloat and helps to index pages faster). Our tools rely on following the same directive to minimize inflated numbers and to give us an accurate look at what SEO errors exist.

---

This will add the following parameter pages to be disallowed from crawling by search engine bots. This pages are deindexed and removing from crawl will reduce false positive SEO errors. 

## How to test this PR

1. Use a crawler like screaming frog or SEMRush and adjust setting to respect robots.txt
2. Use tooling to run a crawl on pages where these parameter changes are present
3. See if pages change from crawled to blocked. 

## Checklist

### Front-end testing
1. Use a crawler like screaming frog, SEMRush, or the [robots.txt testing tool](https://support.google.com/webmasters/answer/6062598?hl=en) to test URLs that have the parameter pages to be blocked.
2. Use tooling to run a crawl on pages where these parameter changes are present
3. See if pages change from crawled to blocked.

### DIFF
--- a/cfgov/unprocessed/root/robots.txt
+++ b/cfgov/unprocessed/root/robots.txt
@@ -12,5 +12,13 @@ Disallow: /ask-cfpb/search/
 Disallow: /ask-cfpb/search-by-tag/
 Disallow: /es/obtener-respuestas/buscar/
 Disallow: /es/obtener-respuestas/buscar-por-etiqueta/
+Disallow: *&authors
+Disallow: *?authors
+Disallow: *&categories
+Disallow: *?categories
+Disallow: *&from_date
+Disallow: *&to_date
+Disallow: *?from_date
+Disallow: *?to_date
 
 sitemap: https://www.consumerfinance.gov/sitemap.xml


cfgov/unprocessed/root/robots.txt
```diff
@@ -12,5 +12,13 @@ Disallow: /ask-cfpb/search/
...
 Disallow: /es/obtener-respuestas/buscar/
 Disallow: /es/obtener-respuestas/buscar-por-etiqueta/
+Disallow: *&authors
+Disallow: *?authors
+Disallow: *&categories
+Disallow: *?categories
+Disallow: *&from_date
+Disallow: *&to_date
+Disallow: *?from_date
+Disallow: *?to_date
```
niqjohnson, line 22: I can never totally wrap my head around how robots.txt directives are parsed, but I *think* we can condense these and get the same result of disallowing URLs with the filter parameters in them:

```suggestion
Disallow: /*authors=
Disallow: /*categories=
Disallow: /*from_date=
Disallow: /*to_date=
Disallow: /*page=
```

I also added `Disallow: /*page=` in there. We have `Disallow: /?page=` on line 8 right now, but that won't catch URLs where the `page` parameter happens to not be the first parameter (like /?foo=bar&page=3), right?

I guess we can confirm that works in [Google's robots.txt tester](https://support.google.com/webmasters/answer/6062598), but that requires a Google account (I can log in on my personal machine and try it if that's easiest).

niqjohnson, line 22: It looks like `from_date` and `to_date` get appended to filter query strings even if you don't use those fields. For example, when I select only the "mortgages" topic on the [blog landing page](https://www.consumerfinance.gov/about-us/blog/), the resulting URL is `https://www.consumerfinance.gov/about-us/blog/?title=&topics=mortgages&from_date=&to_date=`. If there are some filter queries we still want to index—I used a topic filter as an example here intentionally since I notice that it wasn't included in the disallow directives—disallowing `from_date` and `to_date` would prevent those URLs from being indexed.

alexandersirris: Hmm good catch here that is something we'd want to avoid disallowing. If we're keeping certain parameters like ?titles indexed, this would remove those from being crawled and eventually Google would drop them from the index. Any idea why the `from_date `and `to_date ` parameters are added before someone has queried for a date range? Is that something we could potentially fix to only add those once a user has engaged with that filter?

Scotchester: It's definitely not ideal, but I'm not sure anyone can remember the particulars of why we had to do it at the time. We should take a quick look at whether or not that can be changed.

Scotchester: I'm gonna tag @chosak on this, if he can find the time, as the BEWD with the most recent filterable list experience, to see how hard it would be to not include unused query string parameters.

chosak: I agree, we should try to drop those if we can. Is this a backend thing, though? When you visit [/about-us/blog/](https://www.consumerfinance.gov/about-us/blog/) and hit "Apply filters" without selecting anything, that's just a form GET - isn't the GET URL determined by how the HTML form inputs are defined, before even hitting Django does?

In theory Django could redirect to a cleaner version of the URL without those empty query string params, but is there some way of excluding them based solely on how the HTML form is rendered? + @anselmbradford

niqjohnson: I'll defer to actual FEWDs here, but I *think* the only way to get rid of empty query string parameters is to manipulate the `GET` request with some Javascript on form submission. That seems like it would add an extra level of complexity to the already complex filterable list code, so I'm not super keen on that. If there's an easier way to not include empty fields as blank parameters in the query string though, I'm all for it.

Now that I'm thinking about this more, let me read up a little on how crawlers prioritize `allow` and `disallow` statements in `robots.txt`. It just occurred to me that the ideal way to structure these rules would be to disallow crawling any filterable list URLs with a query string *except* if the query string has a parameter we actually care about. Like:

```
Disallow: /*title=
Allow: /*topics=
```

If `allow` overrides the `disallow`—which is what I want to read up on—then that would block crawling on any filterable list URL (I'm pretty sure they all have a `title` parameter in the URL, but we could always include all possible parameters in separate `disallow` lines) except if it happens to contain a `topics` parameter, which is what we want.

niqjohnson: OK, I banged my head against `robots.txt` for a while and didn't really get anywhere. Using Googlebot's [robots.txt tester](https://support.google.com/webmasters/answer/6062598?hl=en&ref_topic=10267931) from my personal Google account, I tested this configuration for our `robots.txt`:

```
User-agent: *
Disallow: /save-hud-counselors-list
Disallow: /about-us/the-bureau/leadership-calendar/pdf
Disallow: /es/obtener-respuestas/*.imprimir.html
Disallow: /external-site
Disallow: /*?success
Disallow: *form-id=*
Disallow: /hud-api-replace
Disallow: /yourstory
Disallow: /ask-cfpb/search/
Disallow: /ask-cfpb/search-by-tag/
Disallow: /askcfpb/search/
Disallow: /es/obtener-respuestas/buscar/
Disallow: /es/obtener-respuestas/buscar-por-etiqueta/
Disallow: /*authors=
Disallow: /*from_date=
Disallow: /*to_date=
Disallow: /*page=
Disallow: /*title=
Allow:    /*categories=
Allow:    /****topics=

sitemap: https://www.consumerfinance.gov/sitemap.xml
```

(Note that in that configuration, I changed `/*categories=` from the `Disallow` in the original PR to `Allow` since `categories` is what we use to segment types of newsroom items like press releases, but that's not super relevant for what I found.)

The main thing I found is, that, indeed, Googlebot uses a priority order to decide how to handle conflicting directives. From [Google's documentation](https://developers.google.com/search/docs/advanced/robots/robots_txt#order-of-precedence-for-group-member-lines):

>At a group-member level, in particular for `allow` and `disallow` directives, the most specific rule based on the length of the `[path]` entry trumps the less specific (shorter) rule. In case of conflicting rules, including those with wildcards, the least restrictive rule is used.

That means that, for the configuration above, URLs like `/about-us/newsroom/?categories=press-release` (which is a link in the mega-menu) and `/about-us/newsroom/?title=&categories=press-release&from_date=&to_date=` (what you get if you hit the apply filter button) are allowed according to the Googlebot testing tool. Awesome, just what we want so far!

Topic-filtered URLs, like `/about-us/newsroom/?title=&topics=mortgages&from_date=&to_date=` also work because of the absurd number of `*` I added to the `Allow:    /****topics=` directive—that's not a typo. Google isn't kidding about the longest directive winning; I had to add all of those asterisks to get that directive longer than `/*from_date=`. So that's weird and kind of hacky, but it still gets us what we want.

Things get complex quickly, though. It looks like most crawlers don't support much regex beyond a simple wildcard, `*`, in directives, which means that there's no good, generalized way to *allow* a URL with blank parameters like `/about-us/newsroom/?title=&categories=press-release&from_date=&to_date=` but *disallow* a URL with some of those same parameters not blank, like `/about-us/newsroom/?title=&categories=press-release&from_date=2020&to_date=2021`. That means we'd still allow indexing on lists filtered by, say, date if that list also happened to include a `categories` parameter. That doesn't seem much better than what we have right now.

But! In looking into all of this, I stumbled across what seems like a very helpful tool, [Google's URL Parameters tool](https://support.google.com/webmasters/answer/6080548?hl=en&ref_topic=10267931) that lets you block crawling of parameters outside of `robots.txt` with much more granularity. It only applies to Googlebot, but if Googlebot is mostly what we care about, maybe that's OK?

Given that, I feel like this has now grown bigger than a PR review, so maybe we should close this PR and take this discussion elsewhere—we can always open another PR once we decide what to do. What do you think, @alexandersirris?

alexandersirris: @niqjohnson excuse the delayed response here, also mostly notes dumping. I reviewed your configuration and it is a step in the right direction. I noticed that the allow for categories and topics allows crawling, but we still have the date parameters that are still not blocked.

Agree on the Google URL Parameter tool, I feel that helps us isolate the specific parameters combination we're looking to block. (i.e we want categories/topics to be crawled, but not any URL with from_date= and to_date= as we already crawl the "parent" index and all links that would be provided in a date range)

I do believe that removing date parameters that are automatically added without engaging with them in the filter would help clear things up a bit, but if we can use the Google URL parameter tool that could help remove specific parameters. There's also the combination of search parameters, ie. the last URL in the list below. In that case I don't see an option to be able to block crawling as that is a duplicate index of everything that already lives under each of those categories.

e.g Let me know if you have questions here. I don't believe we need to crawl any page with a date parameter. 

- https://www.consumerfinance.gov/about-us/blog/ | Crawl
- https://www.consumerfinance.gov/about-us/blog/?page=2#o-filterable-list-controls | Crawl
- https://www.consumerfinance.gov/about-us/blog/?title=&categories=at-the-cfpb | Crawl
     - https://www.consumerfinance.gov/about-us/blog/?title=&categories=at-the-cfpb&from_date=&to_date= | No Crawl
- https://www.consumerfinance.gov/about-us/blog/?title=&topics=mortgages | Crawl
     - https://www.consumerfinance.gov/about-us/blog/?page=2&topics=mortgages#o-filterable-list-controls | Crawl
     - https://www.consumerfinance.gov/about-us/blog/?title=&topics=mortgages&from_date=&to_date= | No Crawl
     - https://www.consumerfinance.gov/about-us/blog/?title=&categories=policy_compliance&topics=mortgages&from_date=&to_date= | No Crawl

This brings me back to my next point, which is what is the overall benefit of crawling search parameters? Each piece of content that gets posted to blog lives in the main index, and all pagination. i.e /about-us/blog/?page=2#o-filterable-list-controls...?page=3....?page=4 and all filtered indexes are just duplicates of what exists (if that makes sense). So, by using robots.txt or Google's URL Parameters tooling we're only going to get so far before we run into the issue of allowing crawl on combinations of parameters. I didn't have knowledge of Google's URL parameters tooling, so I'd need to dig further into this if those combinations would be blocked or if they'd be excluded given the URL doesn't pass 1 blocked parameter. 

Happy to take this discussion elsewhere @niqjohnson, should I file an intake?

niqjohnson: @alexandersirris, yeah, I like making this an intake so we can talk more about what we want crawled. I'll close this PR, and we can think more about what we want to do in robots.txt when you submit the intake request.

niqjohnson:
After some discussion (see https://github.com/cfpb/consumerfinance.gov/pull/6350#discussion_r619445041), we're going to give more thought to how we want to handle crawling of parameterized filterable list URLs. @alexandersirris is going to submit an intake request for the work, so I'm going to close this PR.

PR #6351, opened 2021-03-23
Allow limiting of request body sizes

willbarton:
This PR adds Apache [`LimitRequestBody`](https://httpd.apache.org/docs/2.4/mod/core.html#limitrequestbody) if `LIMIT_REQUEST_BODY` is provided in the environment.

There is a corresponding PR to our automations to add this value.

## Checklist

### DIFF
--- a/cfgov/apache/conf.d/headers.conf
+++ b/cfgov/apache/conf.d/headers.conf
@@ -9,8 +9,7 @@ Header always set X-Frame-Options SAMEORIGIN
 Header always set X-XSS-Protection: "1; mode=block"
 Header always set X-Content-Type-Options: nosniff
 
-# Add LimitRequestBody if the value is provided in the LIMIT_REQUEST_BODY 
-# variable.
+# Add LimitRequestBody when LIMIT_REQUEST_BODY env var is set
 <If "-n osenv('LIMIT_REQUEST_BODY')">
     LimitRequestBody "${LIMIT_REQUEST_BODY}" 
 </If>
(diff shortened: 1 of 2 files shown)


cfgov/apache/conf.d/headers.conf
```diff
@@ -5,11 +5,16 @@ LimitRequestFieldSize 8190
 KeepAlive On
 MaxKeepAliveRequests 500
 
-
 Header always set X-Frame-Options SAMEORIGIN
 Header always set X-XSS-Protection: "1; mode=block"
 Header always set X-Content-Type-Options: nosniff
 
+# Add LimitRequestBody if the value is provided in the LIMIT_REQUEST_BODY 
+# variable.
```
Scotchester, line 13: Trivial suggestion to get the comment on one line and be consistent with the one below.

```suggestion
# Add LimitRequestBody when LIMIT_REQUEST_BODY env var is set
```

cfgov/apache/conf.d/headers.conf
```diff
@@ -5,11 +5,16 @@ LimitRequestFieldSize 8190
...
+<If "-n osenv('LIMIT_REQUEST_BODY')">
+    LimitRequestBody "${LIMIT_REQUEST_BODY}"
```
Scotchester, line 15: I don't know _why_ we use gigantic 8-space indentation in similar blocks below, but we do. Should we do the same here for consistency?

willbarton: Probably, yes.

ascott1 (approved):

PR #5289, opened 2019-10-10
Remove September 2019 CCDB feature flag

niqjohnson:
The September 2019 CCDB feature flag was only needed to prevent content from going live before the go-live date. That date has now passed, so the flag is no longer needed.

## Removals

- Removed the feature-flagged parts of `complaint-landing.html` 

## Changes

- Renamed the CCDB landing page feature flag to be more generic 

## Testing

1. Visit http://localhost:8000/data-research/consumer-complaints/ and make sure it matches exactly what's live at https://www.consumerfinance.gov/data-research/consumer-complaints/
2. Put some content behind the newly named `CCDB_CONTENT_UPDATES` feature flag in the landing page template
3. With the feature flag disabled, visit http://localhost:8000/data-research/consumer-complaints/ and make sure the content doesn't appear
4. Enable the `CCDB_CONTENT_UPDATES` flag, visit http://localhost:8000/data-research/consumer-complaints/ again, and make sure the feature flagged content now appears

## Screenshots

ccdb-not-flagged

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code can be automatically merged (no conflicts)
- [x] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [x] Passes all existing automated tests
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future todos are captured in comments
- [x] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [x] Project documentation has been updated
- [x] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Browsers

- [x] Chrome on desktop
- [x] Firefox
- [x] Safari on macOS
- [x] Edge
- [x] Internet Explorer 9, 10, and 11
- [x] Safari on iOS
- [x] Chrome on Android

### Accessibility

- [x] Keyboard friendly
- [x] Screen reader friendly

### Other

- [x] Is useable without CSS
- [x] Is useable without JS
- [x] Flexible from small to large screens
- [x] No linting errors or warnings
- [x] JavaScript tests are passing

PR #6359, opened 2021-03-29
Remove unused ccdb_content_updates flag

anselmbradford:
This flag appears unused as it doesn't appear in the complaint-landing template.

## Removals

- Remove unused `ccdb_content_updates` flag.

### DIFF
--- a/cfgov/legacy/views/complaint.py
+++ b/cfgov/legacy/views/complaint.py
@@ -32,8 +32,7 @@ class ComplaintLandingView(TemplateView):
         context.update(self.is_ccdb_out_of_date(ccdb_status_json))
 
         context.update({
-            'technical_issues': flag_enabled('CCDB_TECHNICAL_ISSUES'),
-            'ccdb_content_updates': flag_enabled('CCDB_CONTENT_UPDATES')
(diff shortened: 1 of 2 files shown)


cfgov/legacy/views/complaint.py
```diff
@@ -32,8 +32,7 @@ def get_context_data(self, **kwargs):
         context.update(self.is_ccdb_out_of_date(ccdb_status_json))
 
         context.update({
-            'technical_issues': flag_enabled('CCDB_TECHNICAL_ISSUES'),
-            'ccdb_content_updates': flag_enabled('CCDB_CONTENT_UPDATES')
+            'technical_issues': flag_enabled('CCDB_TECHNICAL_ISSUES')
```
Scotchester, line 35: Let's leave the comma here so this line doesn't have to change now, or when another flag is added :)

```suggestion
            'technical_issues': flag_enabled('CCDB_TECHNICAL_ISSUES'),
```

Scotchester (approved):
I had suggested leaving this around for the future in #5289, but seeing as that hasn't happened in 18 months, I'm fine to let it go and re-add it if it's ever needed again.

PR #6362, opened 2021-04-02
Update filter height on resize of window

anselmbradford:
Currently when you load https://www.consumerfinance.gov/about-us/blog/ at desktop size and resize down to mobile. The filter crops to the point of un-usability:

This PR re-calculates the filter's expandable height on resize of the window.

## Changes

- Make the expandable resize on changes to the window size.

## How to test this PR

1. Visit https://www.consumerfinance.gov/about-us/blog/ and resize the window to see the issue where the filter gets cropped.
2. Pull this branch, run locally, and visit /about-us/blog/ and resize the window to compare to production.

### DIFF
--- a/cfgov/unprocessed/js/organisms/FilterableListControls.js
+++ b/cfgov/unprocessed/js/organisms/FilterableListControls.js
@@ -53,28 +53,29 @@ function FilterableListControls( element ) {
     const _expandables = Expandable.init( _dom );
     _expandable = _expandables[0];
 
-    /**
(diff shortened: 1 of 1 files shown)


contolini (approved):
I dig the debounce. 👍 If you want to cover the new lines you might be able to add a unit test that rapid fires `window.dispatchEvent(new Event('resize'));` a few times and verifies that `_expandable.transition.expand()` only gets called once.

PR #6365, opened 2021-04-05
Enable Apache server status endpoint

willbarton:
This change enables Apache's mod_status and adds a server status endpoint that sits behind a check of a new environment variable, `APACHE_STATUS_ALLOW_FROM`. If that variable is defined, the endpoint is added, and the `Allow from` for access to it is populated from its contents. By default the endpoint will `Deny from all`.

In the dev prod-like Docker compose file I've set this to "localhost", which enables us to debugging locally.

The intention behind this change is to enable us better insight into and to instrument Apache with New Relic.

## Checklist


Scotchester (approved):
THis works a treat, at least in the local prod-like environment, but I don't love adding the double underscore to the plain `server-status` suggested in the mod_status docs. Is there something I'm not considering about why just using `server-status` would be a bad idea?

willbarton:
> Is there something I'm not considering about why just using server-status would be a bad idea?

Eh, I'm just trying to avoid using a default status endpoint that would return an Apache deny (not our custom Django 404, and thus make it obvious that its the Apache server status page).

willbarton:
@Scotchester renamed it `/apache-status`, so it's unambiguous.

PR #6366, opened 2021-04-05
Remove public csrf

willbarton:
[CSRF is a problem specific to authenticated pages and actions](https://owasp.org/www-community/attacks/csrf), and our use of it on feedback forms and the Financial Well-Being score prevent us from being able to cache those URLs. 

This PR is an attempt to exempt all our non-authenticated URLs from CSRF. This is somewhat complicated by the fact that Django's `CsrfViewMiddleware` operates specifically on views, and we want to keep it enabled for authenticated URLs.  Because all our Wagtail URLs are served by a single view, Wagtail Sharing's `ServeView`, the Django `csrf_exempt` decorator is not going to be useful to us. 

This is a known issue for the way in which Wagtail pages share a single serve view (by default, `Page.serve`), per https://github.com/wagtail/wagtail/issues/3066. The recommended approach in that issue is a custom middleware that subclasses `CsrfViewMiddleware` and exempts as necessary within its `process_view` method.

This PR adds just such middleware, `PathBasedCsrfViewMiddleware`. This middleware will check for the existence of a `CSRF_REQUIRED_PATHS` setting, and only enforce CSRF for requests that match the paths listed in that setting. If the setting does not exist, it will enforce CSRF everywhere, just as `CsrfViewMiddleware` does.

I've also set `CSRF_REQUIRED_PATHS` to our known-authenticated paths. 

With this middleware in place, all public-facing uses of CSRF, and public-facing exemptions from CSRF, can be removed. The subsequent commits on the PR do that.

## Checklist


higs4281 (approved):
This is a great solution!

Scotchester (changes requested):
Missed a few spots that can be removed:

https://github.com/cfpb/consumerfinance.gov/blob/1bb8f50f7d398c8ffe5e32cb15c2f1cd30be61e5/cfgov/diversity_inclusion/jinja2/diversity_inclusion/voluntary-assessment-form.html#L185

https://github.com/cfpb/consumerfinance.gov/blob/1bb8f50f7d398c8ffe5e32cb15c2f1cd30be61e5/cfgov/paying_for_college/jinja2/paying-for-college/disclosure_feedback.html#L31

I think everything in `cfgov/legacy/templates/common/` in can be deleted now, as well. I can't find evidence of any of these templates still being used.

Finally, unsure about this one… is CSRF still useful in this context? (Probably shouldn't be cached by Akamai, regardless.) Maybe @higs4281 should comment. https://github.com/cfpb/consumerfinance.gov/blob/1bb8f50f7d398c8ffe5e32cb15c2f1cd30be61e5/cfgov/paying_for_college/templates/paying_for_college/disclosure.html#L105

higs4281:
@Scotchester Good point! Yes, disclosures don't need CSFR -- there is no authentication.

I wondered about satellite apps too, but I think we're clear there.
- retirement, oah-api and ccdb5-api and curriculum-tool have references in little-used stand-alone test settings, but those up can be a followup task.
- Only curriculum tool has a live reference that we should get: https://github.com/cfpb/curriculum-review-tool/blob/c74bed94ae9b55a3011136a89efab0fa854a7671/crtool/views.py#L13

willbarton:
Sounds good.

In my testing locally, the references shouldn't be blockers — the CSRF token simply won't be checked. I think cleaning up the satellites can happen asynchronously with this PR. 

I'll open new PRs there and then a new PR here with updates once those are in. But this one can stand alone.

willbarton:
@Scotchester thanks for catching those! They're removed in 569a736.

lfatty:
This is great but I am wondering why not  including `CsrfResponseMiddleware` middleware.

willbarton:
@laminfaty we're subclassing Django's `CsrfViewMiddleware`, which is the successor to `CsrfResponseMiddleware`, which [was deprecated in Django 1.4](https://docs.djangoproject.com/en/3.2/internals/deprecation/#deprecation-removed-in-1-4).

lfatty:
Got it!

Scotchester:
> I think everything in `cfgov/legacy/templates/common/` in can be deleted now, as well. I can't find evidence of any of these templates still being used.

@willbarton Did you miss this comment?

willbarton:
@Scotchester Oh! No, I didn't, I just forgot to follow-up. In the interest of small, tightly-scoped PRs, I am going to open a separate PR to do that.

Scotchester (approved):

willbarton:
Following up on this:

> I wondered about satellite apps too, but I think we're clear there, with one exception.
>
> -retirement, oah-api and ccdb5-api and curriculum-tool have references in little-used stand-alone test settings, but those can be a followup task.
> -Only curriculum tool has a live reference that we should get: https://github.com/cfpb/curriculum-review-tool/blob/c74bed94ae9b55a3011136a89efab0fa854a7671/crtool/views.py#L13

PRs to each of these to remove these references have been opened and referenced above.

PR #6368, opened 2021-04-06
Paying for College: Migrate remaining disclosure tools to jinja from django templates

anselmbradford:
## Changes

- Move disclosure and disclosure_technote django templates to jinja.
- Move source JavaScript from https://github.com/cfpb/college-costs/tree/master/src/disclosures/js to the cf.gov repo to replace minified https://github.com/cfpb/consumerfinance.gov/blob/fbcdc6001aa4cf58533f2e5c59a04054b29dc131/cfgov/unprocessed/apps/paying-for-college/js/disclosure-main.js
- Remove csrf_token from form.

## How to test this PR

1. http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/offer/test/
http://localhost:8000/paying-for-college2/understanding-your-financial-aid-offer/about-this-tool/ should still work like production. Note that the about-this-tool page injects $ amounts into the text, so it's not just static text.


higs4281 (approved):
This was a big lift, and it's looking great. Thanks, @anselmbradford

PR #6375, opened 2021-04-07
Paying for College - Fix for $0 debt bug

mistergone:
This is a small fix which will address the nagging bug which occasionally causes debt at graduation to appear to be $0.

## Additions
- Add handler to ensure the programLength is updated in the financialModel when
the user changes programLength radio button

## How to test this PR
1. On production, to reproduce the bug, fill out the Financial Path to Graduation tool as normal, adding a school and other details. Importantly, when the "Costs" page appears with its "Yes/No" question, answer "Yes." Then add a tuition cost, then add some federal loans, then check "Affording your loans -> Is this too much debt?" page. (Make sure you navigate directly there from the loans page.) This should show $0 debt (incorrectly).
2. Reproduce the same steps on a local version with this patch. It should show the correct debt!

## Checklist


virginiacc (approved):
This works as described for me! 👍

PR #6378, opened 2021-04-08
Allow CCPI chart x axes to be quarter-based

wpears:
As requested by OR.

To test:

Make a chart based on CCPI data. View the x axis both at max range and at 3 & 5 year ranges.

Then, in the style overrides in Wagtail, add the following:

```
{
"tooltip.formatter": "fn__ccpi_quarterLabels",
"xAxis.labels.formatter": "fn__ccpi_xAxisLabels",
"xAxis.tickPositioner": "fn__ccpi_tickPositioner"
}
```

Re-examine the xAxis at various time scales. Note also that at short time scales the tooltip is properly aligned specific tick marks.

### DIFF
--- a/cfgov/unprocessed/js/routes/on-demand/simple-chart/chart-hooks.js
+++ b/cfgov/unprocessed/js/routes/on-demand/simple-chart/chart-hooks.js
@@ -1,10 +1,20 @@
 /* eslint camelcase: [0] */
-const ccpi_quarterMap = {
-  'Mar 31': 'Q1: Jan-Mar',
-  'Jun 30': 'Q2: Apr-Jun',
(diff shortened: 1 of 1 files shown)


schbetsy (approved):
So, when the chart is zoomed out enough that the x-axis only shows years, not quarters, they line up like this, where Q4 2007 matches up with the 2008 label. Is that the desired behavior?
Screen Shot 2021-04-08 at 3 14 57 PM

If so, this looks great to me.

wpears:
I think that's the most accurate representation of what's actually being represented by the data and was indeed my intent. If there's designer/user feedback about this later we can revisit it (but we've had a small discussion where the implemented behavior was more or less agreed upon).

PR #6380, opened 2021-04-08
fix simple chart bottom margin

wpears:
Removed the simple chart bottom margin at some point by mistake, here it is, in all its glory.

Testing:

Check out this branch, two charts on a page should have some breathing room (whereas they don't on build)


anselmbradford:
Should this use a `unit( … )` calculation, to make it clearer how you came to `2.8125em`?

schbetsy (approved):
LGTM. @anselmbradford, do you want to review again?

PR #6386, opened 2021-04-12
Financial Well-Being: Move js from apps to routes js

anselmbradford:
Financial Well-Being app has its initialization code inside the routes js directory, which calls js in the apps directory. The apps directory is really for applications that have dedicated templates that call the scripts from apps directly. This doesn't happen with Financial Well-Being.

## Changes

- Move js from apps to routes js

## How to test this PR

1. Pull branch and `gulp clean && gulp build` and http://localhost:8000/consumer-tools/financial-well-being/ and http://localhost:8000/consumer-tools/financial-well-being/results/ should work like production.


contolini (dismissed):
I'm getting the same error as the PR check, looks like the test files need to be updated with the new paths.

anselmbradford:
After talking with @Scotchester we're going to move everything to apps and get rid of the URL-dependent route scripts.

anselmbradford:
@Scotchester moved the scripts to apps from routes.

cwdavies (approved):

Scotchester (approved):
This works as expected, but a question: Any reason not to just load fwb-questions.js and fwb-results.js in the templates? I.e., is it necessary to create home.js and results.js?

anselmbradford:
> This works as expected, but a question: Any reason not to just load fwb-questions.js and fwb-results.js in the templates? I.e., is it necessary to create home.js and results.js?

We could, but as a general best practice I think we should avoid modules initializing themselves. Since this goes through the bundler the two files shouldn't matter.

PR #6388, opened 2021-04-13
Fix the PFC view and tests for jinja2 use

higs4281:
This PR fixes three bits of the technote rendering.

### Template function calls 
Jinja2 templates run Python code in a more Pythonic way than do Django templates.
Django allows templates to call functions without the standard `()` evocation,
whereas Jinja executes functions with `()` and arguments.

### Tests
The Django test client also does not populate the `context` variable in
Jinja2 responses, so tests can't check `context` for expected values.

In this case, we can just as easily check the `content` of the response
for the values we're expecting.

### Constants

The PFC technote template performed a bunch of span replacement actions to pull in and format 
constants that are stored in the database. These swaps were not working in the jinja2 template. 
Since the constants are available to the technote's view, it seems simpler to format them in the view 
and deliver them at render time without any swapping.   

This PR delivers formatted constants in the text where needed, using the slug names 
that editors see in the database. This makes it easier to see where a constant is being used in the text.

### A semi-related nit

The offer view originally allowed a bare response -- rendering the tool without
any school or program specified, and with no warning dialog. Since the tool is useless and confusing 
in this state, I think it makes sense to return 404 when there is no GET data.

### DIFF
--- a/cfgov/paying_for_college/jinja2/paying-for-college/disclosure_technote.html
+++ b/cfgov/paying_for_college/jinja2/paying-for-college/disclosure_technote.html
@@ -404,7 +404,7 @@
                             period
                         </li>
                         <li class="list_item">
-                            Loan fee of {{ DLOriginationFee }}   of the
+                            Loan fee of {{ DLOriginationFee }} of the
                             total loan amount
                         </li>
                         <li class="list_item">
--- a/cfgov/paying_for_college/jinja2/paying-for-college/disclosure_technote.html
+++ b/cfgov/paying_for_college/jinja2/paying-for-college/disclosure_technote.html
@@ -229,7 +229,7 @@
                             Awarded based on financial need
                         </li>
                         <li class="list_item">
-                            Up to <span data-cap="pell">$X</span> per academic
+                            Up to {{ pellCap }} per academic
                             year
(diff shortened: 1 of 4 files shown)


cfgov/paying_for_college/jinja2/paying-for-college/disclosure_technote.html
```diff
@@ -409,9 +404,7 @@ <h4>
                             period
                         </li>
                         <li class="list_item">
-                            Loan fee of <span data-financial="DLOriginationFee"
-                            data-fee="origination"
-                            data-percentage_value="true">[XX]</span>% of the
+                            Loan fee of {{ DLOriginationFee }}   of the
```
anselmbradford, line 407: Could get rid of the extra space after the variable, or I can get it in my PR.

anselmbradford (approved):

cfgov/paying_for_college/jinja2/paying-for-college/disclosure_technote.html
higs4281:
```suggestion
                            Loan fee of {{ DLOriginationFee }} of the
```

PR #6391, opened 2021-04-14
Future dates cause errors in filterable lists

cwdavies:
This change should fix an internal server error seen recently in issue [5365](https://github.local/CFGOV/cfgov-alerts/issues/5365), that has something to do with how filterable lists handle future dates.
Putting future dates in the `from_date` and `to_date` filterable list fields should not cause an internal server error.

## Additions

- Add default_input_formats to `FilterableDateField` class in `cfgov/v1/forms.py`
- Add [freezegun](https://pypi.org/project/freezegun/) and test for future dates

## Removals

- `cfgov/v1/util/date_filter.py` module

## Changes

- `cfgov/v1/forms.py`
- `cfgov/v1/tests/test_forms.py`
- `cfgov/v1/tests/util/test_datetimes.py`
- `cfgov/v1/util/datetimes.py`

## How to test this PR

Attempt to add future date into the FilterableList form using a Safari or IE11 browser
e.g. from_date and to_date: 05/04/2021 and 11/04/2021
Try flipping the month and day, European style, which looks like a time range in the future
Filter a filterable list on a date range entirely in the future

1. Visit https://www.consumerfinance.gov/about-us/newsroom/ in Safari or IE11 in Sauce Labs, or any other browser that doesn't support the native date input control
2. Enter some date filter and click Apply Filters
3. Verify there is no internal server error

## Notes and todos

- This must be testing in IE11, Safari, and Opera Mini browsers

## Checklist

#### Browser testing

- [ ] Firefox
- [x] Chrome
- [x] Safari
- [ ] Safari on iOS
- [ ] Chrome on Android

#### Accessibility

- [ ] Screen reader friendly

#### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


cfgov/v1/forms.py
```diff
@@ -241,11 +244,8 @@ def set_interpreted_date_values(self, cleaned_data):
                     earliest_results, self.preferred_datetime_format)
 
             if to_date:
-                transformed_to_date = end_of_time_period(
-                    self.data['to_date'], cleaned_data['to_date'])
-                cleaned_data['to_date'] = transformed_to_date
                 self.data['to_date'] = date.strftime(
-                    transformed_to_date, self.preferred_datetime_format)
+                    cleaned_data['to_date'], self.preferred_datetime_format)
```
willbarton, line 248: It looks like [`end_of_time_period` was added to handle ](https://github.com/cfpb/consumerfinance.gov/commit/624c5ae4ad2dddf0b30bb671a3467cc08d017181#diff-40757824786a8c38156d12b1db8227629193ded9e4279342230a81d3e4844250) ambiguous dates. Are those still handled?

cwdavies: @willbarton I can add more tests to check for ambiguous dates to verify that they are handled.
Do you have any specific examples in mind?

cwdavies: Good catch. I added `end_of_time_period` to handle ambiguous dates.

willbarton (changes requested):
I can't speak for anyone else, but I often use tests to understand intention behind code. Could you add some tests that specifically test `date_from_pattern` and `end_of_time_period` directly? I suspect re-using the tests for the old functions would be a good way to ensure they continue to work as intended.

Scotchester:
> I can't speak for anyone else, but I often use tests to understand intention behind code. Could you add some tests that specifically test `date_from_pattern` and `end_of_time_period` directly? I suspect re-using the tests for the old functions would be a good way to ensure they continue to work as intended.

+1 to that, and also Craig, would you mind un-renaming the module so that we can clearly see in the diff what has changed from before?

cwdavies:
The GHE issue 4035 stated:
```
From that investigation, we thought the best way to remedy this is to replace the whole `date_filter.py` module
```
That is the reason why I removed both date_filter.py and test_date_filter.py.

I put both `date_from_pattern` and `end_of_time_period` into the `datetimes.py` module.
The above two are still needed to handle ambiguous dates.
My understanding of ambiguous dates are those that contain just a year or just a month and a year.

The original behavior is left intact to use the end of month if only a month and year are entered.
e.g. if a user entered 3/2021 in a filter form, the date used will be March 31st 2021

Also the original behavior is left intact to use the end of year if only a year is entered.
e.g. if a user entered 2020 in a filter form, the date used will be December 31st 2020

All date and time utilities used are now in a single module.

willbarton (dismissed):

willbarton:
@cwdavies The issue was specifically referencing seeing if we could replace it with the use of [`dateutil.parser`](https://dateutil.readthedocs.io/en/stable/parser.html). This was discussed when triaging the issue, but I think in retrospect is probably not possible. Either way, I think this works — the core issue was the returning of a `None` from those functions that we didn't handle in the form itself, and that is definitely fixed.

willbarton (approved):

PR #6397, opened 2021-04-19
Switch from file-based to DB cache

willbarton:
This change removes our production file-based default cache in favor of a default database cache.

There will be a corresponding PR to remove permission-setting on `eregs_cache` from our automations, but we already run `createcachetable` for our database post preview cache.

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/production.py
+++ b/cfgov/cfgov/settings/production.py
@@ -88,8 +88,9 @@ STATIC_ROOT = os.environ['DJANGO_STATIC_ROOT']
 
 CACHES = {
     'default': {
-        'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
-        'LOCATION': '/tmp/eregs_cache',
+        'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
+        'LOCATION': 'cfgov_default_cache',
+        'TIMEOUT': None,
     },
     'post_preview': {
         'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
--- a/Dockerfile
+++ b/Dockerfile
@@ -126,11 +126,10 @@ USER apache
 
 # Build frontend, cleanup excess file, and setup filesystem
 # - cfgov/f/ - Wagtail file uploads
-# - /tmp/eregs_cache/ - Django file-based cache
 RUN ln -s ${SCL_HTTPD_ROOT}/etc/httpd/modules ${APACHE_SERVER_ROOT}/modules && \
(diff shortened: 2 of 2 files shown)


chosak:
Can [these references](https://github.com/cfpb/consumerfinance.gov/blob/1bc235666611aea3a1702b0b7989e5812ba55adf/Dockerfile#L129-L133) to eregs_cache go as well?

Scotchester:
Is there any simple way to test this?

cwdavies:
Should there be a `TIMEOUT` defined for the Database cache? It is currently set as `None`.

In `cfgov/cfgov/settings/test.py` it looks like cache is disabled for testing with a TIMEOUT of 0.
Also in `cfgov/cfgov/settings/local.py` cache is disabled when working locally.

When we get an export of production data, will there be cache defined in a database table?

willbarton:
[Per the Django docs](https://docs.djangoproject.com/en/3.1/topics/cache/#cache-arguments), the `None` will mean cache keys never expire by default.

cfgov/cfgov/settings/production.py
```diff
@@ -88,8 +88,9 @@
 
 CACHES = {
     'default': {
-        'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
-        'LOCATION': '/tmp/eregs_cache',
+        'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
+        'LOCATION': 'cfgov_default_cache',
+        'TIMEOUT': None,
```
willbarton, line 93: To test this, you can copy this config:

```python
CACHES = {
    'default': {
        'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
        'LOCATION': '/tmp/eregs_cache',
        'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
        'LOCATION': 'cfgov_default_cache',
        'TIMEOUT': None,
    }
}
```

Into `cfgov/cfgov/settings/local.py`, and then run

```
./cfgov/manage.py createcachetable
```

Then you should be able to cache in the shell:

```python

```
from django.core.cache import cache
cache.set('my_key', 'hello, world!', 10)
cache.get('my_key')
```

You should get `hello, world!` back.

Then wait 10 seconds and try again:

```
cache.get('my_key')
```

And you should get nothing.

cwdavies: Thanks for the steps to test.

cwdavies (approved):
I verified caching works locally using the steps provided.

PR #6400, opened 2021-04-19
Add functional tests workflow for GitHub Actions

Scotchester:
**Status as of 5/24/21:** This is running correctly against the site with the initial data load! Most of the tests fail without production-like data. What's needed now is to figure out how to dump only the data we need for these tests and load it into the database used for this workflow.

---

## Additions

- New GitHub Actions workflow that sets up the site and runs our Cypress tests against it

## How to test this PR

1. Check out the **functional / cypress** check results.

## Notes and todos

- [ ] Need to get appropriate production-like data loaded in before running Cypress

## Checklist


Scotchester:
I've updated the OP with the current state of this PR.

willbarton:
> What's needed now is to figure out how to dump only the data we need for these tests and load it into the database used for this workflow.

@Scotchester and @chosak had the idea of assembling a SQLite test database and committing it to the repo. That should enable us to run the functional tests against that development/test data in GitHub Actions with the configuration. The only blocker to that is that we currently require PostgreSQL because of use of `JSONField`, but with Django 3.2, `JSONField` will not require PostgreSQL.

willbarton:
@baruva this is awesome! I do think there are a couple more things I'd like to see on this PR before we review it:

- It does look like some of the failures result from errors in connecting to Elasticsearch. When Scott originally opened this PR, our tests didn't depend on it (AFAIR). So, it looks like we will need to add an elasticsearch service along side Postgres. You could probably copy what we do [in the backend workflow](https://github.com/cfpb/consumerfinance.gov/blob/main/.github/workflows/backend.yml#L32-L36).
- I'd like to explore a way to exclude the tests that fail from this action, so that the tests that pass now are the ones that continue to pass, and then we can iterate on adding back in the ones that fail. But I don't think we should exclude them globally — we still want them running internally and the ability to run them locally. This could mean passing `--spec` to Cypress manually in this PR, but I'm not sure what different capability Cypress has for achieving this right off.
- We should also capture a list of the spec files that fail, and post them to the internal GHE issue, so we can task out the process of adding data and making them pass.

But it's sooooo great to see the tests running in GHA and some of them passing!

willbarton:
> This could mean passing --spec to Cypress manually in this PR, but I'm not sure what different capability Cypress has for achieving this right off, or what's exposed via the Cypress GitHub action.

I wrote that and *then* went to check, of course, and it looks like we can specify spec files for the action to run: https://github.com/cypress-io/github-action#specs

willbarton (approved):
When we merge I'd like to make sure we announce that (1) we have a partial functional test run in GitHub Actions, and (2) it's very partial — it's important for people to continue to check both this run and the Jenkins run for now.

willbarton:
I also want to ping @Scotchester, because he might be interested in us actually moving forward with this 😀 .

Scotchester:
So this is currently just running the tests that don't need specific database data? Well, looking forward to seeing the SQLite approach when you get to Django 3.2+ :)

Still exciting to see it get merged, though!

PR #6403, opened 2021-04-21
Ensure regulation section labels are unique within their version

willbarton:
This change validates that a section's label is unique to the version of the regulation it belongs to.

This means that instead of creating a section that then results in 500 errors, content editors will see an error like this:

image

## Checklist


cfgov/regulations3k/models/django.py
```diff
@@ -292,6 +292,21 @@ def extract_graphs(self):
...
+        super().validate_unique(*args, **kwargs)
+
+        # Ensure that the section's label is unique within this version.
+        qs = self.__class__.objects.filter(
+            subpart__version=self.subpart.version,
+            label=self.label
+        ).exclude(
+            pk=self.pk
+        )
+        if qs.exists():
```
Scotchester, line 305: Something slightly descriptive for this variable would be nice, but I'm not sure I can come up with something more concise than `sections_with_duplicate_labels`. What do you think?

```suggestion
        sections_with_duplicate_labels = self.__class__.objects.filter(
            subpart__version=self.subpart.version,
            label=self.label
        ).exclude(
            pk=self.pk
        )
        if sections_with_duplicate_labels.exists():
```

Scotchester (approved):
Works on my local! One minor comment on variable naming.

niqjohnson:
@willbarton, is it worth mentioning the uniqueness constraint in the label field's help text ([cfgov/regulations3k/models/django.py L224-L226](https://github.com/cfpb/consumerfinance.gov/blob/1840e522cf61673e6dfd4045fd3cb1db0b601a86/cfgov/regulations3k/models/django.py#L224-L226))? Maybe something like:

```python
help_text='Labels must be unique and always require at least 1 '
          'alphanumeric character, then any number of alphanumeric '
          'characters and hyphens, with no spaces.'
```

willbarton:
@niqjohnson excellent idea! Included in eec8619.

PR #6408, opened 2021-04-22
Ensure Ask CFPB answer id and language are always unique

willbarton:
This change enforces uniqueness between Ask CFPB answer pages and languages. This means that a single answer id (represented by the `Answer` object can only have one `AnswerPage` with a given `language`. So, for example, there can only be one `en` `AnswerPage` for an id `5678`.

Note: this will raise a `ValidationError` on page copy that Wagtail currently *does not catch*. The copy will fail, but the content editor will see a 500. We've decided this is a better situation to be in than currently, where a copy will succeed but attempting to *view* an answer on the website will 500. That is a problem we'll fix in another PR.

## Checklist


higs4281 (approved):
This is much better than the current state. The error isn't friendly, but that's for another PR.

cfgov/ask_cfpb/models/answer_page.py
```diff
@@ -349,6 +350,30 @@ def __str__(self):
...
+            language=self.language,
+            answer_base=self.answer_base
+        ).exclude(
+            pk=self.pk
+        )
+        if pages_with_same_answer_and_language.exists():
+            raise ValidationError(
+                {
+                    NON_FIELD_ERRORS: (
+                        f"An answer page in {self.language} already exists "
```
higs4281, line 371: I briefly considered something like this to avoid the odd "For answer page in en" error messaging, but subsequent error catching could make this moot.

```python
'...'
language_map = {tup[0]: f"{tup[1]}" for tup in settings.LANGUAGES}
'...'
f"An answer page in {language_map[self.language]} already exists"
```

cwdavies (approved):
Yes, this is much better than the current state now the answer id and language are unique

PR #6410, opened 2021-04-22
Enable previewing of unpublished Ask answers on a Wagtail-Sharing sharing site

willbarton:
This change enables our Ask CFPB answer pages to be previewed on the sharing site if they're new and have never been published.

Previously, such pages would 404, because of the `answer_page.live` check. This change reorganizes the code to check for the existence of a sharing site for the request first.

## Checklist


higs4281 (approved):
Reorganization works great. Thanks for fixing this.

PR #6417, opened 2021-04-28
Enable default cache locally based on env var

willbarton:
This change enables our default cache locally based on an environment variable. This will enable us to use the default cache in local development more easily.

## How to test this PR

Set `ENABLE_DEFAULT_CACHE` in `.env` (see `.env_SAMPLE` in this PR).

Run `./cfgov/manage.py createcachetable`

Run `./cfgov/manage.py shell` and:

```python
from django.core.cache import cache
cache.set('my_key', 'hello, world!', 10)
cache.get('my_key')
```

You should get hello, world! back.

Then wait 10 seconds and try again:

```python
cache.get('my_key')
```

And you should get nothing.

## Notes and todos

- We should settle on use cases of the Django cache and document them, along with this env var to enable the cache locally.

## Checklist

### DIFF
--- a/.env_SAMPLE
+++ b/.env_SAMPLE
@@ -33,6 +33,7 @@ export WAGTAIL_SHARING_HOSTNAME=content.localhost
 # Location used by refresh_data.sh for retrieving latest database dump.
 #export CFGOV_PROD_DB_LOCATION=<some_database_dump_url>
 
+#export ENABLE_DEFAULT_CACHE=1
 #export ENABLE_POST_PREVIEW_CACHE=1
 #export EMAIL_HOST=<email_server_hostname>
 #export ADMIN_EMAILS=<comma_delimited_list_of_emails>
--- a/cfgov/cfgov/settings/local.py
+++ b/cfgov/cfgov/settings/local.py
@@ -85,6 +85,14 @@ CACHES = {
     for k in ("default", "post_preview")
 }
 
+# Optionally enable default cache
+if os.environ.get("ENABLE_DEFAULT_CACHE"):
+    CACHES["default"] = {
+        "BACKEND": "django.core.cache.backends.db.DatabaseCache",
+        "LOCATION": "default_cache",
+        "TIMEOUT": None,
+    }
+
 # Optionally enable cache for post_preview
 if os.environ.get("ENABLE_POST_PREVIEW_CACHE"):
     CACHES["post_preview"] = {


Scotchester (approved):
Works as advertised! Couple questions:

1. Any risk in having that env var exported by default?
2. Is "default cache" the clearest name for this? What do you think about `ENABLE_DJANGO_DEFAULT_CACHE`?

willbarton:
> 1. Any risk in having that env var exported by default?

You mean always enabled? I think it could result in some unexpected frustration if, for example, I was to use the Django cache to store authors, topics, and categories that we use in the filterable lists, and someone expects that to change locally when they add or edit a page and it doesn't. 

For local development, I think making turning on the cache intentional makes more sense than having it on by default.

> 2. Is "default cache" the clearest name for this? What do you think about ENABLE_DJANGO_DEFAULT_CACHE?

It's consistent with the `ENABLE_POST_PREVIEW_CACHE` variable. Both of those could certainly have `ENABLE_DJANGO_` prefixed, but I think I'd want to maintain that consistency if we're going to be explicit about it. I don't have strong feelings beyond that though.

Scotchester:
> > 1. Any risk in having that env var exported by default?
> 
> You mean always enabled? I think it could result in some unexpected frustration if, for example, I was to use the Django cache to store authors, topics, and categories that we use in the filterable lists, and someone expects that to change locally when they add or edit a page and it doesn't.
> 
> For local development, I think making turning on the cache intentional makes more sense than having it on by default.

Makes sense :+1:

> > 2. Is "default cache" the clearest name for this? What do you think about ENABLE_DJANGO_DEFAULT_CACHE?
> 
> It's consistent with the `ENABLE_POST_PREVIEW_CACHE` variable. Both of those could certainly have `ENABLE_DJANGO_` prefixed, but I think I'd want to maintain that consistency if we're going to be explicit about it. I don't have strong feelings beyond that though.

I considered that but wasn't sure if changing an established env var was going to cause too many headaches, and `POST_PREVIEW_CACHE` is considerably more specific and descriptive than `DEFAULT_CACHE`, so one is not likely to not know (or need to know) that it's also a Django cache, but I don't feel strongly about it. Still approved, either way.

willbarton:
I would like to document how to use the Django cache, once we have some use cases. I think that might be a good time to revisit the naming question? I'm going to merge this one for now.

higs4281 (approved):

PR #6418, opened 2021-04-28
Disallow the curriculum review tool completion pages in robots.txt

higs4281:
We don't need or want the CRT completion pages indexed for search.
Users should always start upstream.

### DIFF
--- a/cfgov/unprocessed/root/robots.txt
+++ b/cfgov/unprocessed/root/robots.txt
@@ -13,5 +13,6 @@ Disallow: /ask-cfpb/search-by-tag/
 Disallow: /askcfpb/search/
 Disallow: /es/obtener-respuestas/buscar/
 Disallow: /es/obtener-respuestas/buscar-por-etiqueta/
+Disallow: /consumer-tools/educator-tools/youth-financial-education/curriculum-review/tool/*
 
 sitemap: https://www.consumerfinance.gov/sitemap.xml


cfgov/unprocessed/root/robots.txt
```diff
@@ -13,5 +13,6 @@ Disallow: /ask-cfpb/search-by-tag/
 Disallow: /askcfpb/search/
 Disallow: /es/obtener-respuestas/buscar/
 Disallow: /es/obtener-respuestas/buscar-por-etiqueta/
+Disallow: /consumer-tools/educator-tools/youth-financial-education/curriculum-review/tool/*
```
niqjohnson, line 16: Not sure if it matters, but this will also disallow crawling of the top-level tool page, `/consumer-tools/educator-tools/youth-financial-education/curriculum-review/tool/`, which looks like it redirects to `/consumer-tools/educator-tools/youth-financial-education/curriculum-review/before-you-begin/` if you haven't started a review yet. I *think* `/consumer-tools/educator-tools/youth-financial-education/curriculum-review/before-you-begin/` will still get crawled even if we disallow a redirect to it, so it might not matter.

mrclay (approved):

cfgov/unprocessed/root/robots.txt
higs4281: I was intending to disallow `/tool/` and anything that follows, including hash strings and query strings, to exclude any URL that exposes data entry.

PR #6421, opened 2021-04-29
Improve filterable list performance

willbarton:
This PR makes a number of changes to our filterable lists.

- The `FilterablePagesDocumentSearch` class has been refactored to more easily share an instance.
- The `FilterablePagesDocumentSearch` object is now shared between `FilterableListMixin` and `FilterableListForm` (instead of two separate instances being used).
- `FilterableListForm` now uses the Django cache to cache the list of authors, topics, and all filterable pages to avoid repeatedly fetching and querying the information when it has not changed.
- We have a signal handler that will purge the Django cache and the frontend Akamai cache when a filterable page for has been published or unpublished.

All of these changes should improve the overall performance of our filterable lists. This reduces duplicate queries that we previous had that constructed Django querysets from Elasticsearch results two separate times during a single request cycle, and the addition of cached values on the `FilterableListForm` should provide further improvement. Finally, implementing a signal handler that will only purge filterable lists from Akamai when their filterable pages are changed means we can eliminate our 10-minute maxage on those pages.

Note: This is not a full refactor of filterable lists. We still have quite a lot of improvements we can make. Separation of concerns between `FilterableListMixin`, `FilterableListForm`, and `FilterablePagesDocumentSearch` is still vague, and there's a lot cross reference between `FilterableListMixin` and `FilterableListForm` that could be eliminated. The goal here is performance improvement, which I believe this achieves, and the refactoring that is done has been done to enable that without making the diff so complex that the performance-specific changes can't be seen.

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/organisms/filterable-list.html
+++ b/cfgov/jinja2/v1/_includes/organisms/filterable-list.html
@@ -35,15 +35,14 @@
 
    show_post_tags: Whether to render the tags in post previews.
 
+   has_unfiltered_results: Whether the filterable list has anything to filter.
+
    ========================================================================== #}
 
 {% import 'organisms/filterable-list-controls.html' as filterable_list_controls with context %}
 {% import 'molecules/notification.html' as notification with context %}
 
 {# SHOW A NOTIFICATION IF THERE'S NOTHING TO FILTER. #}
-
-{% set has_unfiltered_results = filter_data.form.filterable_pages.first() %}
-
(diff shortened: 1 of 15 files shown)


Scotchester (approved):
This is all working locally as expected! I haven't done a thorough reading of the code changes, though Will walked us through most of them earlier today.

higs4281 (approved):
I'd like to better understand why some local page rendering is slow, but we should give this a try for the sake of bot relief.

PR #6423, opened 2021-04-30
Remove sender from filterable list signal handler

willbarton:
I made an assumption that the `sender` of a Django signal could be a superclass. It cannot be. This change removes the specific sender model for publish/unpublish signals. Now the signal handler will process any publish/unpublished signal, and test if the instance is an instance of `AbstractFilterPage` (which does work).

I've also fixed that `if` so that it `return`s instead of `pass`ing.

## Checklist


Scotchester (approved):
The updated logic makes sense :+1:

higs4281 (approved):

PR #6426, opened 2021-05-04
Update Spanish content for FWB about page

niqjohnson:
Adds a paragraph and link to an additional report to the end of the financial well-being survey's about page.

---

## Additions

- Paragraph and link to an additional report

## Changes

- The single jump link that was at the end of the page is now a link list to get the margin between that and the new paragraph that follows it correct (`.a-link__jump + p` has no margin in between them).

## How to test this PR

1. Fire up /consumer-tools/financial-well-being/about/es/ and make sure it looks like the screenshot below
2. Make sure the new link points to /data-research/research-reports/measuring-financial-well-being-hispanics-2018-benchmarks/
3. Make sure the English about page, /consumer-tools/financial-well-being/about/, is unchanged from what's in production

## Screenshots

fwb-about-es

## Notes and todos

- There is no change needed for the English about page; this content should only appear on the Spanish page
- I kept the link text as the translated string of `{{ _('See the report') }}` even though it only appears on the Spanish page just in case we ever want to change that translation globally

## Checklist

### DIFF
--- a/cfgov/wellbeing/jinja2/wellbeing/about.html
+++ b/cfgov/wellbeing/jinja2/wellbeing/about.html
@@ -92,10 +92,25 @@
         {{ _('The tool allows you to compare your financial well-being score to other adults in the United States.') }}
         {{ _('The comparison of your score to others’ scores is made possible by our national survey. In late 2016, we collected responses to our financial well-being scale in a nationwide survey of U.S. adults for the first time.') }}
     </p>
-    <a class="a-link__jump"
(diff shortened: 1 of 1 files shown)


Scotchester (approved):
Approved. Not a blocker by any means, but this is a situation where I'd just toss in one of our margin utility classes, rather than building out the whole `m-list__links` structure for two single-item lists. Food for thought, if you want to switch it up :)

PR #6428, opened 2021-05-05
Formalize logic into feature of the template organism

mrclay:
* Moves logic to own module
* Adds option to the "expandable" template
* Makes the search module a better team player by not initializing itself and instead allowing the index module to do this.

- [x] Add test


mrclay:
These are what we changed :+1: 
```
PASS test/unit_tests/apps/teachers-digital-platform/js/search-spec.js
...
PASS test/unit_tests/apps/teachers-digital-platform/js/expandable-mobile-spec.js
```

PR #6434, opened 2021-05-11
Sc geo design

wpears:
Further design improvements to the geographic charts in simple-chart


schbetsy (approved):
Can't reproduce the leaky tooltip issue, so we'll wait and address that if it comes up again. Otherwise, this looks good.

PR #6438, opened 2021-05-11
Paying for College - Non-first-year fixes

mistergone:
This Pull Request fixes problems with the calculations of federal Direct loans for non-first-year students. It also fixes the limits enforced on Direct subsidized loans based on "progress" of the student (years they have spent in a program).

## Changes
- Change calculations to fix non-first-year debt totals
- Change Direct loan limits to match non-first-year limits
- Change CSS to fix some errors in non-first-year content

## How to test this PR
* Pull it down and get it running.
* The following should be true:
  * The limit for first-year students Federal Direct subsidized loans should be $3500
  * Second-year should be $4500
  * Third year should be $5500
  * Debt totals should reflect that we are assuming the user increases their subsidized loan amount each year (so that the percentage of total subsidized loan limit is taken each year)
  * Feel free to ask questions if anything doesn't make sense!

### DIFF
--- a/cfgov/unprocessed/apps/paying-for-college/css/college-costs/state-based.less
+++ b/cfgov/unprocessed/apps/paying-for-college/css/college-costs/state-based.less
@@ -358,7 +358,7 @@ main.college-costs {
 	}
 
 	&[data-state_programprogress="0"],
-	&[data-state_programprogress="n"] {
+	&:not( [data-state_programprogress="n"] ) {
 		[data-state-based-visibility="is_first_year-true"] {
 			display: block;
 		}
--- a/cfgov/unprocessed/apps/paying-for-college/css/college-costs/state-based.less
+++ b/cfgov/unprocessed/apps/paying-for-college/css/college-costs/state-based.less
@@ -358,7 +358,7 @@ main.college-costs {
 	}
 
 	&[data-state_programprogress="0"],
-	&:not( [data-state_programprogress="n"] ) {
+	&[data-state_programprogress="n"] {
 		[data-state-based-visibility="is_first_year-true"] {
 			display: block;
 		}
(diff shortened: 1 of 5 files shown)


cfgov/unprocessed/apps/paying-for-college/js/models/financial-model.js
```diff
@@ -167,13 +167,21 @@ const financialModel = {
   _enforceLimits: () => {
     let unsubCap = 0;
     const errors = {};
+    const yearMap = {
```
anselmbradford, line 170: Could this be exported from `debt-calculator.js` and then imported and used in this file so there isn't a redundant data map between the two files?

cfgov/unprocessed/apps/paying-for-college/css/college-costs/state-based.less
```diff
@@ -358,7 +358,7 @@ main.college-costs {
 	}
 
 	&[data-state_programprogress="0"],
-	&[data-state_programprogress="n"] {
+	&:not( [data-state_programprogress="n"] ) {
```
virginiacc, line 361: Is the ":not" needed here, since a progress value of "n" equates to being a first year?

mistergone: This is probably an oversight from resolving a merge conflict! You're right, and I will push a fix!

virginiacc (approved):
The updated loan calculations seem to work as described for me. 👍 

I agree with @anselmbradford that it would be nice to manage the progress data map in one place.

cfgov/unprocessed/apps/paying-for-college/js/models/financial-model.js
mistergone: Yes! I'd like to actually refactor this all to be simpler, but that'll have to wait for a future update. Thanks!

PR #6443, opened 2021-05-13
Remove fixtooltip chart callback

wpears:
Fixes tooltip text breaking out of its container


schbetsy (approved):
This fixes the issue for me!

PR #6451, opened 2021-05-19
Simplify/clarify simple-chart wagtail interface

wpears:
Addresses https://github.local/CFPB/super-regular/issues/250#issuecomment-284850


cfgov/v1/atomic_elements/organisms.py
```diff
@@ -552,8 +552,12 @@ def get_limit(self, value):
...
 class SimpleChart(blocks.StructBlock):
     title = blocks.CharBlock(required=True)
-    subtitle = blocks.CharBlock(required=False)
-    figure = blocks.CharBlock(required=False)
+    subtitle = blocks.TextBlock(required=False)
+    description = blocks.TextBlock(
+        required=True,
+        help_text='Accessible description of the chart content'
+    )
+    figure_number = blocks.CharBlock(required=False)
```
schbetsy, line 560: For both fields that were renamed (figure and x-axis data), those values will be missing on any existing charts after this change (meaning the charts will be broken). We could prevent that by creating a data migration, but that seems like overkill, since there aren't any charts in production yet. We'll just need to make sure to update all existing charts on staging environments as soon as this change gets published.

schbetsy (changes requested):

cfgov/v1/atomic_elements/organisms.py
schbetsy: Alternatively, we could prevent that by not renaming the fields, instead using the `label` to just change the display names of the fields.

schbetsy (approved):

PR #6457, opened 2021-05-24
Support the option to perform a delete purge of Akamai cache

cwdavies:
This task is to add the option to call Akamai's [delete purge API](https://developer.akamai.com/api/core_features/fast_purge/v3.html#apisummary) to our [Akamai frontend cache backend](https://github.com/cfpb/consumerfinance.gov/blob/master/cfgov/v1/models/caching.py), add a corresponding management command that can perform it

## Additions

- Added `delete(url)` in `caching.py` to support the delete of Akamai cache
- `cfgov/v1/management/commands/delete_page_cache.py`
- `cfgov/v1/management/commands/delete_all_pages_cache.py`

## How to test this PR

1. tox

## Checklist


cfgov/v1/management/commands/delete_page_cache.py
```diff
@@ -0,0 +1,21 @@
...
+            required=True,
+            nargs="+",
+            help=(
+                "The full URL for the page cache to delete "
+                "(can specify multiple)"
+            )
+        )
+
+    def handle(self, *args, **options):
+        purge_urls_from_cache(options["url"])
```
willbarton, line 21: This is the same as it previously was, but I don't think this will actually work.

`purge_urls_from_cache` [just issues a `purge_batch` to the backend](https://github.com/wagtail/wagtail/blob/main/wagtail/contrib/frontend_cache/utils.py#L104), which will end up with an invalidation. 

I think I'm going to do a bit of refactoring here.

cfgov/v1/management/commands/delete_page_cache.py
```diff
@@ -1,4 +1,7 @@
-from django.core.management.base import BaseCommand
+from unittest import mock
```
higs4281, line 1:
```suggestion
```

cfgov/v1/models/caching.py
```diff
@@ -106,6 +102,21 @@ def purge_all(self):
...
+    """ Akamai backend that performs a 'delete' purge
+
+    This is a special-case backend, and should not be globally configured. """
+
+    def purge(self, url):
+        self.post(url, 'delete')
+
+    def purge_all(self):
+        raise NotImplementedError(
+            "Purging all by deletion is intentionall not supported through "
```
higs4281, line 115:
```suggestion
            "Purging all by deletion is intentionally not supported through "
```

requirements/test.txt
```diff
@@ -1,4 +1,5 @@
 coverage==4.5.1
+diff_cover==2.6.0
```
higs4281, line 2: From another PR?

higs4281 (commented):
Just some nits.

cfgov/v1/models/caching.py
```diff
@@ -100,6 +102,21 @@ def purge_all(self):
         self.post_all('invalidate')
 
 
+class AkamaiDeletingBackend(AkamaiBackend):
```
willbarton, line 105: The Wagtail frontend cache invalidator API doesn't really support multiple types of purges, where we want to be able to issue both "invalidate" and "delete" purges. The cleanest way to do this and retain use of all the frontend cache invalidator API seems to be introduce a second backend specific to deletion. Because deletion is not something we want to do often, we'll leave it unconfigured. For when we do want to delete, we have a management command that will assemble a configuration for it from the normal Akamai backend.

requirements/test.txt
willbarton: How'd you guess 😆

willbarton (approved):

PR #6459, opened 2021-05-25
Add google optimize anti-flicker script

anselmbradford:
When testing the A/B optimize script, the change in the hero flicker was apparent, so we should add the anti-flicker script detailed in https://support.google.com/optimize/answer/7100284

Also, remove the inline optimize script as we're delivering it via GTM.

## Removals

- Remove inline optimize script.

## Changes

- Adds optimize anti-flicker script.

## How to test this PR

1. Will be tested on production :-\


contolini (approved):
If the flicker remains, try moving it to the very [top of `<head>`](https://support.google.com/optimize/answer/7100284#snippet-placement-best-practices).

PR #6460, opened 2021-05-25
Add new story page type

contolini:
Adds a new "Story Page" content type with a sidebar-less centered column and images that bleed into the left and right gutters. It's use case is prose-heavy landing pages designed to tell a story.

To let editors choose when they want either RichTextBlock images or FullWidthText images to bleed into the gutters, a 'bleed' format option has been added to both image fields. Selecting this option for an image that's not on a story page will have no affect (it'll be treated as a standard full width image).

See context at https://github.local/CFPB/el-camino/issues/406

## Notes

- It'd be ideal to only show the "bleed" image format when editing a story page but I don't think that's possible given how the [formats are defined](https://docs.wagtail.io/en/stable/advanced_topics/customisation/page_editing_interface.html#rich-text-image-formats).
- Open to any styling recommendations. I'm using a [content modifier class](https://github.com/cfpb/consumerfinance.gov/compare/story-page?expand=1#diff-388c7690c0d544528764b456f7f7b7b76aa7758049ff55d062338dcf11c32083R22) to scope the CSS. Would a template level class (e.g. `.t-story`) be more appropriate? Or something else altogether? The `.bleed-image` class is vaguely named because it has to apply to both content-image molecules and rich text field images.

## Additions

- Story page type

## Changes

- "Bleed" image format option to both rich text images and full width text images.

## How to test this PR

1. I made a demo page here: https://cfgov-pr-6460.DEMO-PROXY/lorem-ipsum-dolor-sit-amet/
1. Log in with admin/admin to view the editing experience

or

1. Create a Story page.
1. Add a few full width text blocks.
1. Add images and rich text fields with images and they should have a new "bleed" image format.
1. Preview the page and the appropriate images should hang into the left/right gutters.

## Screenshots

| full-width text image selector | rich text image selector | preview |
|--------|--------|---------|
| editor   | editor   | preview    |

## Checklist

#### Browser testing

- [x] Firefox
- [x] Chrome
- [x] Safari
- [x] Safari on iOS
- [x] Chrome on Android

#### Accessibility

- [x] Screen reader friendly

#### Other

- [x] Is useable without CSS
- [x] Is useable without JS


cfgov/v1/jinja2/v1/story_page.html
```diff
@@ -0,0 +1,36 @@
...
+{% block hero %}
+    {% for block in page.header %}
+        {% if block.block_type == 'features' %}
+            <section class="wrapper
+                            wrapper__match-content
+                            block
+                            block__bg
+                            block__border
+                            block__flush-top
+                            u-mb0">
```
anselmbradford, line 12: Could this be `block__flush-bottom` instead?

contolini: Let me try it. I directly copied the HTML from [campaign-page.html](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/v1/jinja2/v1/campaign_page.html) which is the only other page (aside from the homepage) that uses a jumbo hero.

contolini: Hmmm yeah the utility class is used because `block__flush-bottom` removes the bottom border which we don't want.

screenshot

cfgov/unprocessed/css/enhancements/layout.less
```diff
@@ -321,6 +321,49 @@
...
+
+    & > section {
+        margin: 0 auto;
+        max-width: 41.875rem;
+
+        .o-full-width-text-group {
+            overflow: inherit;
+        }
+
+        img.bleed-image {
```
anselmbradford, line 336: We should probably settle on `bleed-image` or `image-bleed` (see `content__center-image-bleed` above)

anselmbradford: Do we need `img` here? Maybe just have it `.bleed-image` (if we switched to `<picture>` or something down the road.

cfgov/v1/jinja2/v1/story_page.html
```diff
@@ -0,0 +1,36 @@
...
+                {% include_block block %}
+            </section>
+        {% else %}
+            {% include_block block %}
+        {% endif %}
+    {% endfor %}
+{% endblock %}
+
+{% block content_main_modifiers -%}
```
anselmbradford, line 21: Should be `{% block content_modifiers -%}`

cfgov/v1/jinja2/v1/story_page.html
```diff
@@ -0,0 +1,36 @@
...
+  {{ super() }} content__center-image-bleed
```
anselmbradford, line 22: Technically this is not correct BEM. This is creating:
```
main.content
  div.content_wrapper
    div.content_main.content__center-image-bleed
    …
```

Buuuuut the modifier `content__center-image-bleed` should be modifying `content`, not `content_main`. So the structure should be:

```
main.content.content__center-image-bleed
  div.content_wrapper
    div.content_main
    …
```

See earlier comment.

cfgov/unprocessed/css/enhancements/layout.less
```diff
@@ -321,6 +321,49 @@
 }
 
 
+// Center content and bleed full-width images into gutters
+// Used on our story pages
+.content__center-image-bleed {
+
+    & > section {
```
anselmbradford, line 328: This would be `& .content_main > section {` if `content__center-image-bleed` is moved up to `content`.

anselmbradford: We should probably give `section` a name [in the template](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/jinja2/v1/_layouts/content-base-main-only.html#L7), like `content_main-inner` or something. Maaaaaybe something for another PR. Your call.

cfgov/unprocessed/css/enhancements/layout.less
```diff
@@ -321,6 +321,49 @@
...
+        margin: 0 auto;
+        max-width: 41.875rem;
+
+        .o-full-width-text-group {
+            overflow: inherit;
+        }
+
+        img.bleed-image {
+            @half-gutter-width: @grid_gutter-width / 2;
```
anselmbradford, line 337: Is this used?

cfgov/v1/jinja2/v1/story_page.html
```diff
@@ -0,0 +1,36 @@
+{% extends "layout-full.html" %}
+
+{% block hero %}
+    {% for block in page.header %}
+        {% if block.block_type == 'features' %}
```
anselmbradford, line 5: There's a use-case for `features` being added to this template?

Screen Shot 2021-05-27 at 1 49 58 PM

cfgov/unprocessed/css/enhancements/layout.less
contolini: Yeah L342 and L343. The constant multiplication and division was visually overwhelming so I moved the basic gutter calcs to these variables.

cfgov/v1/jinja2/v1/story_page.html
contolini: The jumbo hero is designed to have optional features tucked underneath it. See [SSSU](https://www.consumerfinance.gov/start-small-save-up/) or the homepage. We won't be using it for our new inequity page but future story pages might need it.

cfgov/unprocessed/css/enhancements/layout.less
anselmbradford: Oh I see. I had done a find search and didn't find it in the escaped string below. All good!

schbetsy (approved):

cfgov/unprocessed/css/enhancements/layout.less
contolini: > We should probably give `section` a name [in the template](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/jinja2/v1/_layouts/content-base-main-only.html#L7), like `content_main-inner` or something. Maaaaaybe something for another PR. Your call.

Sounds good to me. Seems harmless and helpful.

cfgov/unprocessed/css/enhancements/layout.less
contolini: Changed everything to `.image-bleed` in https://github.com/cfpb/consumerfinance.gov/commit/e9f1d87096db0a1e546dadc7b3f2dc17c0bd8d2f 👍

cfgov/v1/jinja2/v1/story_page.html
contolini: Fixed in https://github.com/cfpb/consumerfinance.gov/commit/e9f1d87096db0a1e546dadc7b3f2dc17c0bd8d2f

cfgov/v1/jinja2/v1/story_page.html
contolini: Fixed in https://github.com/cfpb/consumerfinance.gov/commit/e9f1d87096db0a1e546dadc7b3f2dc17c0bd8d2f

cfgov/unprocessed/css/enhancements/layout.less
contolini: Added the new class in  https://github.com/cfpb/consumerfinance.gov/commit/e9f1d87096db0a1e546dadc7b3f2dc17c0bd8d2f

cfgov/v1/image_formats.py
```diff
@@ -0,0 +1,11 @@
+from wagtail.images.formats import Format, register_image_format
+
+
+register_image_format(
+    Format(
+        'bleed',
+        'Bleed into left/right margins',
+        'richtext-image image-bleed',
+        'original'
```
Scotchester, line 9: Let's set this to the max width of a bleeding image so that it doesn't request originally-uploaded images that are unnecessarily large.

```suggestion
        'width-1170'
```

cfgov/v1/models/story_page.py
```diff
@@ -0,0 +1,47 @@
...
+)
+from wagtail.core.blocks import StreamBlock
+from wagtail.core.fields import StreamField
+from wagtail.core.models import PageManager
+
+from v1.atomic_elements import molecules, organisms
+from v1.models.base import CFGOVPage
+
+
+class StoryHeader(StreamBlock):
```
Scotchester, line 12: Did you consider just importing `CampaignHeader` instead of redefining it here as `StoryHeader`? I may be missing something, but seems like that would be a little more DRY.

cfgov/v1/models/story_page.py
```diff
@@ -0,0 +1,47 @@
...
+    full_width_text = organisms.FullWidthText()
+    image = molecules.ContentImage()
+    info_unit_group = organisms.InfoUnitGroup()
+    text_introduction = molecules.TextIntroduction()
+    video_player = organisms.VideoPlayer()
+
+
+class StoryPage(CFGOVPage):
+    header = StreamField(StoryHeader, blank=True)
+    content = StreamField(StoryContent, blank=True)
```
Scotchester, line 35: This is not important and not wrong as-is, but just as something you might be interested to know, `StoryHeader` (`CampaignHeader`, originally) was defined as its own class so that its `Meta` `block_counts` could be set. When that `Meta` is not needed for a StreamField, as in the case of the `content` StreamField here, we usually just define the StreamField items within `content = StreamField(…, blank=True)`.

cfgov/v1/image_formats.py
```diff
@@ -0,0 +1,11 @@
+from wagtail.images.formats import Format, register_image_format
+
+
+register_image_format(
```
Scotchester, line 4: I have a small concern about adding the ability to add images in this format inside rich text fields. Most notably, it requires that alt text be added for each inserted image, which is a minor accessibility concern if the image is decorative (i.e., a screen reader reading text about the image wouldn't add value for the user). Aside from that, we have more control over the markup output when using our custom ContentImage block.

All that said, I know it can be annoying to have to break up long stretches of text with separate image blocks (when editing in the Wagtail interface), so it's probably not the end of the world. Just try to make the alt text as useful or unobtrusive as possible.

cfgov/v1/jinja2/v1/story_page.html
```diff
@@ -0,0 +1,36 @@
...
+        {% if 'featured_content' in block.block_type %}
+            <div class="block{{ ' block__flush-top' if loop.first else '' }}">
+                {% include_block block %}
+            </div>
+        {% else %}
+            {% import 'templates/render_block.html' as render_block with context %}
+            {{ render_block.render(block, loop.index) }}
+        {% endif %}
+    {%- endfor %}
+{% endblock %}
```
Scotchester, line 36: Did some testing and I am pretty confident this can be simplified to just this:

```suggestion
{% block content_main %}
    {% for block in page.content %}
        <div class="block{{ ' block__flush-top' if loop.first else '' }}">
            {% include_block block %}
        </div>
    {% endfor %}
{% endblock %}
```

(`render_block` is an older custom method that should eventually be completely replaced by the Wagtail native `include_block`)

Scotchester (changes requested):
A few little bits and bobs to consider before merging. I would say that the `width-1170` one is "blocking", so that's why I chose "request changes". The others are up to you!

cfgov/v1/image_formats.py
contolini: Makes sense 🎉

cfgov/v1/jinja2/v1/story_page.html
contolini: TY TIL. I'll pull it down and verify nothing's borked.

cfgov/v1/image_formats.py
contolini: I agree with you and if we continue doing more "story" like content on cf.gov we'll want to explore more standardized methods of embedding images/diagrams/videos.

cfgov/v1/models/story_page.py
contolini: I hear what you're saying but that feels risky because I can see story pages diverging from campaign pages in the future and some cross-contamination occurring. I suppose we could just redefine `StoryHeader` when that happens but it seems cleaner to keep them separate from the get-go.

Scotchester (approved):

cfgov/v1/image_formats.py
niqjohnson: I'm late on this, but I just wanted to note that, as of Wagtail 2.13 (which we'll be on soon), alt text is no longer required for rich text images. We'll be able to insert decorative images with a correct, blank alt attribute once the update to 2.13 is done.

PR #6470, opened 2021-06-04
Delete test/browser_tests/.eslintrc

anselmbradford:
Cucumber browser tests were deleted in https://github.com/cfpb/consumerfinance.gov/pull/6429/, but the hidden eslint config for the tests appears to have been overlooked. We have `test/.eslintrc`, so `test/browser_tests/.eslintrc` appears to be unnecessary.

## Removals

- Unused `test/browser_tests/.eslintrc` config file.

## How to test this PR

1. `gulp lint` should pass.


contolini (approved):
✂️

PR #6472, opened 2021-06-08
Filterable-list controls: Check if expandable is already open before expanding it to refresh size

anselmbradford:
https://github.com/cfpb/consumerfinance.gov/pull/6362 introduced this bug, which can be seen at https://[GHE]/CFGOV/platform/issues/4074

## Changes

- Check if expandable is already open before open it to refresh size.

## How to test this PR

1. Check something like https://www.consumerfinance.gov/activity-log/ and resize the screen and see that the control opens with a reversed show/hide button. Compare to behavior on http://localhost:8000/activity-log/ where it doesn't open when it is closed and resizing.
 

## Notes and todos

- The Expandable really should have an API to check if it is opened or not, and to programmatically open/close it without needing to delve into the transition directly.


contolini (approved):
Works as expected! Thanks for the fix. I agree that exposing a property/method to determine if it's open would be ideal.

PR #6475, opened 2021-06-09
Wagtail: Add Simplified Chinese to language dropdown

anselmbradford:
## Additions

-  Add Simplified Chinese to language dropdown to Wagtail.
## How to test this PR

1. Pull branch, run migrations.
2. Edit or create a page in wagtail and go to the configuration tab and see that the language drop-down now has Chinese (Siimplified) as an option.

## Screenshots

Screen Shot 2021-06-09 at 3 35 56 PM

## Notes and todos

- This is part one of a multi-part update. After this goes in all the chinese pages will be changed to "Chinese (Simplified)" and then "Chinese" will be updated to "Chinese (Traditional)". Then the chinese pages will again be updated to be either traditional or simplified.


cfgov/v1/models/base.py
```diff
@@ -67,7 +67,7 @@ class CFGOVPage(Page):
     tags = ClusterTaggableManager(through=CFGOVTaggedPages, blank=True,
                                   related_name='tagged_pages')
     language = models.CharField(
-        choices=ref.supported_languagues, default='en', max_length=2
+        choices=ref.supported_languagues, default='en', max_length=7
```
willbarton, line 70: Given the [possible options for language subtags](https://en.wikipedia.org/wiki/IETF_language_tag#Syntax_of_language_tags), and the fact that this field is populated by the set list of choices, I'd propose making this large enough that we don't have to add a migration when we might add a particular language with possible other subtags later.

anselmbradford: Have we settled on 100?

willbarton: Yup. As @chosak suggests, [Wagtail's built-in `Locale` model uses 100](https://github.com/wagtail/wagtail/blob/b8a86ac15de331654024f3b4590088f0b88626fc/wagtail/core/models.py#L341-L402).

anselmbradford: Done! https://github.com/cfpb/consumerfinance.gov/pull/6475/commits/b2b97bfd6a5cdf4975a156cc48b5f959a69344fd

willbarton (approved):

PR #6476, opened 2021-06-10
Add a tox env for backend test coverage

willbarton:
This small change adds a tox env that will run `diff-cover` on the coverage generated by `unittest-current` or `unittest-future`. This is intended as a convenience means of reporting out the change in test coverage vs `main` that doesn't require running `diff-cover` directly.

I have not changed the GitHub actions for now, as I think it's important to maintain consistency between the backend and frontend actions.

## How to test this PR

1. Run `tox -e unittest-current` on a branch with changes from `main`, and then run `tox -e coverage`. Alternatively, run `tox -e unittest-current,coverage`. Observe that coverage is reported out like it is in our GitHub Actions.

## Checklist

### DIFF
--- a/docs/python-unit-tests.md
+++ b/docs/python-unit-tests.md
@@ -138,7 +138,14 @@ From the root of `consumerfinance.gov`.
 
 ### Coverage
 
-To see Python code coverage information, run
+To see Python code coverage information immediately following a test run, 
+you can add the `coverage` env to the list of envs for tox to run:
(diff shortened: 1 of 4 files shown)


higs4281 (approved):
I think this will be useful.

Would be nice to add a mention in docs at https://cfpb.github.io/consumerfinance.gov/python-unit-tests/#coverage/

PR #6484, opened 2021-06-15
Update enforcement export script

wpears:
Closes https://github.local/CFPB/super-regular/issues/327


cfgov/scripts/export_enforcement_actions.py
```diff
@@ -38,33 +30,31 @@ def assemble_output():
             continue
         page_categories = ','.join(
             c.get_name_display() for c in page.categories.all())
+        content = ''
+        soup = BeautifulSoup(str(page.content), 'html.parser')
+        para = soup.findAll(['p', 'h5'])
+        for p in para:
+            content += p.get_text()
+            link = p.find('a', href=True)
```
schbetsy, line 38: This code assumes that there's no more than one anchor tag per `p`. That's probably a safe assumption, but it might be helpful to put a comment here in case someone is troubleshooting later. Something like `<p> tags won't have multiple links, per style guide`.

cfgov/scripts/export_enforcement_actions.py
```diff
@@ -92,7 +82,7 @@ def export_actions(path='/tmp', http_response=False):
         write_questions_to_csv(response)
         return response
     file_path = '{}/{}'.format(path, slug).replace('//', '/')
-    with open(file_path, 'w', encoding='windows-1252') as f:
+    with open(file_path, 'w', encoding='utf-8') as f:
```
schbetsy, line 85: When I look at the csv output in vim, it looks great. When I open it in Excel on a Mac, I see encoding weirdness on apostrophes and quotation marks, e.g. on [this page](http://localhost:8000/enforcement/actions/capital-one-bank/) (see the screenshot below). Not sure how it will appear to those opening in Excel on a Windows computer. Probably not an issue, but FYI, I guess.

Screen Shot 2021-06-16 at 5 24 02 PM

schbetsy (approved):
Overall, this looks great. Major improvement over what was here before. I left some minor comments that you can take or leave.

schbetsy:
OH, also, when I ran `./frontend.sh`, I saw a bunch of diffs of deleted files in the `npm-packages-offline-cache` folder. Make sure there's no git weirdness there before you merge.

cfgov/scripts/export_enforcement_actions.py
wpears: yeah I see the encoding weirdness as well. Spent a bit of time trying to get it fixed in excel for mac but eventually punted as I don't think it's widespread or important (and may not even be present on windows)

PR #6496, opened 2021-06-22
Add search synonyms to Django Admin

willbarton:
This change adds a really simple `ModelAdmin` to the Django admin (so, it's limited to folks with access to it) that lets us manage our in-database search synonyms.

This is an intermediate step toward content-managable synonyms, because changing a synonym doesn't automatically rebuild the Elasticsearch index to use it (for that, a Jenkins job will still have to be run).

But this should replace the need to manage these in-database on the command line.

Once it's approved and ready to merge, we have internal documentation that needs to be updated.

## Screenshots

image

image

## Checklist


higs4281 (approved):
Good first step.

PR #4096, opened 2018-05-07
Add ARIA label field to Hyperlink atom

Scotchester:
One of the things we learned in the course of the accessibility mini-project is that we can use the `aria-label` attribute on `<a>` elements to provide additional context to screen reader users in cases where the cleared link text was something like "Learn more" or when including the context would significantly harm the experience for sighted users (e.g., [repetitive links in a long table](https://www.consumerfinance.gov/policy-compliance/guidance/supervision-examinations/#anchor_manual)).

This PR adds an ARIA label field to the Hyperlink atom that's found in many modules so that Wagtail editors can add a value for the `aria-label` attribute when appropriate.

## Additions

- New ARIA label field on Hyperlinks and the template update to output it

## Testing

1. Pull branch
1. Migrate
1. Add/edit any module that includes a Hyperlink
1. Preview it without adding anything to the ARIA label field and see that it has no `aria-label` attribute
1. Enter something in the ARIA label field and preview it
1. Observe that it now has an `aria-label` attribute

## Screenshots

screen shot 2018-05-07 at 11 59 15

## Notes

- Unfortunately, this doesn't do anything for links added within a rich text field. After the move to Wagtail 2.0, we should consider a Draftail plugin that would enable that.

## Checklist

- [ ] PR has an informative and human-readable title
- [ ] Changes are limited to a single goal (no scope creep)
- [ ] Code can be automatically merged (no conflicts)
- [ ] Code follows the standards laid out in the [development playbook](https://github.com/cfpb/development)
- [ ] Passes all existing automated tests
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future todos are captured in comments
- [ ] Visually tested in supported browsers and devices (see checklist below :point_down:)
- [ ] Project documentation has been updated
- [ ] Reviewers requested with the [Reviewers tool](https://help.github.com/articles/requesting-a-pull-request-review/) :arrow_right:

### Browsers

- [ ] Chrome
- [ ] Firefox
- [ ] Safari
- [ ] Internet Explorer 8, 9, 10, and 11
- [ ] Edge
- [ ] iOS Safari
- [ ] Chrome for Android

### Accessibility

- [ ] Keyboard friendly
- [ ] Screen reader friendly

### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS
- [ ] Flexible from small to large screens
- [ ] No linting errors or warnings
- [ ] JavaScript tests are passing

PR #6497, opened 2021-06-22
Wagtail: Add aria-label option to links

anselmbradford:
Pages like https://www.consumerfinance.gov/owning-a-home/ have links like "Learn more" that are not descriptive enough. This PR adds the ability to add an aria-label to links in Wagtail that aren't descriptive enough for assistive devices.

## Additions

- Adds aria-label input field to Wagtail hyperlinks.

## How to test this PR

1. Visit a page with a hyperlink like Buying a House (http://localhost:8000/admin/pages/3701/edit/) and see that you can add an aria label to link.

## Screenshots

Screen Shot 2021-06-22 at 1 05 49 PM


cfgov/v1/atomic_elements/atoms.py
```diff
@@ -26,6 +26,14 @@ class Meta:
...
 class Hyperlink(blocks.StructBlock):
     text = blocks.CharBlock(required=False)
+    aria_label = blocks.CharBlock(required=False,
+                                  help_text='Add an ARIA label if the link '
+                                            'text does not describe the '
+                                            'destination of the link '
+                                            '(e.g. has ambiguous text like '
+                                            '"Learn more" that is not '
+                                            'descriptive on its own).'
+    )
```
willbarton, line 36: Minor formatting suggestion:

```suggestion
    aria_label = blocks.CharBlock(
        required=False,
        help_text=(
            'Add an ARIA label if the link text does not describe the '
            'destination of the link (e.g. has ambiguous text like '
            '"Learn more" that is not descriptive on its own).'
        )
    )
```

cfgov/v1/atomic_elements/organisms.py
```diff
@@ -1061,9 +1062,10 @@ def links(self):
         for hyperlink in self.get('links') or []:
             url = hyperlink.get('url')
             text = hyperlink.get('text')
+            aria_label = hyperlink.get('aria_label')
 
             if url and text:
-                links.append({'url': url, 'text': text})
+                links.append({'url': url, 'text': text, 'aria_label': aria_label})
```
willbarton, line 1068: This should satisfy the linter as well:

```suggestion
                links.append({
                    'url': url, 
                    'text': text, 
                    'aria_label': aria_label
                })
```

willbarton (approved):
Python looks good to me.

cfgov/v1/tests/atomic_elements/organisms/test_organisms.py
```diff
@@ -445,7 +445,8 @@ def test_render(self):
             ('/bar/', 'Another link'),
         ):
             self.assertIn(
-                f'<a class="m-list_link" href="{url}">{text}</a>',
+                f'<a class="m-list_link"\n                       '
+                f'href="{url}"\n                       >{text}</a>',
```
anselmbradford, line 449: Is there a better way to do this @willbarton ?

Scotchester:
Looks like this accomplished most of what I was intending to eventually get around to on #4096, so that can probably be closed.

anselmbradford:
> Looks like this accomplished most of what I was intending to eventually get around to on #4096, so that can probably be closed.

Oh, look at that! I forgot about the overlap there and it indeed looks very similar… which makes me question why we didn't merge yours and if I'm missing an issue in mine?

Scotchester:
Mine had only added it to the Hyperlink atom template, and I realized there places where the Hyperlink atom was used on the back end, but when it was rendered, the Hyperlink template wasn't used. One of those was the FCM, as you updated in this PR. I have a suspicion there are others, as well.

One place that's a little more complicated but definitely needs `aria-label`s is ResourceLists, where there usually ends up being a whole column of `Download` or `Order` links. Those should probably just be handled in the template automatically (i.e., don't create a place to manually enter an `aria-label` for each of those, but rather automatically add on to the specific link text, like `arial-label="{{ link text }} {{ resource name }}"` (pseudocode).

PR #6499, opened 2021-06-24
Remove the functional tests run from our Jenkinsfile

willbarton:
Because the functional tests have been so temperamental running against the Docker stacks, we're going to temporarily remove them from the PR cycle and run them in our continuous deploy to staging.

This change removes everything related to them from the Jenkinsfile.

## Checklist


Jenkinsfile
```diff
@@ -33,9 +30,7 @@ pipeline {
         DEPLOY_SUCCESS = false
         // Determines if Elasticsearch image should be updated
         IS_ES_IMAGE_UPDATED = 'false'
-        // Determines if Cypress image should be updated
-        IS_CYPRESS_IMAGE_UPDATED = 'false'
-        DOCKER_HUB_REGISTRY = 'https://hub.docker.com'
+        DOCKER_HUB_REGISTRY = 'https://registry.hub.docker.com'
```
chosak, line 33: I think perhaps we can remove this definition and use the default `docker login` below without a registry parameter.

Jenkinsfile
```diff
@@ -129,7 +114,7 @@ pipeline {
                             usernameVariable: 'DOCKER_HUB_USER'
                         )
                     ]) {
-                        sh 'docker login -u $DOCKER_HUB_USER -p $DOCKER_HUB_PASSWORD'
+                        sh 'docker login $DOCKER_HUB_REGISTRY -u $DOCKER_HUB_USER -p $DOCKER_HUB_PASSWORD'
```
chosak, line 117: See above; will this work if we don't provide the registry URL?

chosak (approved):
✂️ ✂️ ✂️ ✂️ ✂️ ✂️ ✂️ ✂️ ✂️ ✂️ ✂️

higs4281 (approved):
Thank you for your service, Cypress

Jenkinsfile
willbarton: That doesn't work — see the build history of https://github.com/cfpb/consumerfinance.gov/pull/6498#partial-pull-merging

Jenkinsfile
willbarton: See comment above.

PR #6503, opened 2021-06-28
Simple chart tilemap styling updates

wpears:
From https://github.local/CFPB/super-regular/issues/280#issuecomment-284788


schbetsy:
Have you seen this bug where some state abbreviations on the tile map disappear when you resize the window? It exists on this PR, and I'm also seeing it on Build.


https://user-images.githubusercontent.com/4295388/123837199-90466480-d8d8-11eb-9fe4-94449df43e04.mov

schbetsy (approved):
The map and legend look good to me, so I'm going to approve this, but check out the bug I mentioned above in case it's in scope for this PR.

wpears:
I've noticed that bug on and off but haven't been able to track down the source of the problem. I think I'm okay with it for now since resizing the window is fairly uncommon and it's fixed on refresh. I'll ticket it for later.

PR #6505, opened 2021-07-01
Have New Relic ignore exceptions that generate HTTP 429

willbarton:
This change adds 429: Too Many Requests to our list of ignored status codes for New Relic. We occasionally have [`rest_framework.exceptions.Throttled`](https://www.django-rest-framework.org/api-guide/exceptions/#throttled) show up in our error violations, when it shouldn't because that's a valid exception to throw when we intend to throttle API requests.

Per the [New Relic documentation](https://docs.newrelic.com/docs/agents/python-agent/configuration/python-agent-configuration/#error-ignore-status-codes), it sounds like this should work, because Django does use exceptions to return HTTP responses. If this doesn't work, we can add the exception to [`error_collector.expected_classes`](https://docs.newrelic.com/docs/agents/python-agent/configuration/python-agent-configuration/#error-ignore).

## Checklist

### DIFF
--- a/cfgov/newrelic.ini
+++ b/cfgov/newrelic.ini
@@ -236,7 +236,7 @@ error_collected.enabled = true
 #
 # This setting is only compatible with some web frameworks, as some
 # frameworks do not use exceptions to return HTTP responses.
-error_collector.ignore_status_codes = 100-102 200-208 226 300-308 429
+error_collector.ignore_status_codes = 100-102 200-208 226 300-308 404 429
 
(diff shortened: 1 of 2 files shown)


willbarton:
@chosak I've bumped the New Relic agent and made one minor modification to the config file as a result — the 6.4.0.157 release deprecated `error_collector.ignore_errors` in favor of `error_collector.ignore_classes`. It's not a setting we currently use, but if ignoring the 429 status code doesn't work, that's what I'll try next.

chosak (approved):

PR #6506, opened 2021-07-02
Filters: Swap authors filter with a languages filter

anselmbradford:
Get rid of the authors filter and replace it with a languages filter.

## Removals

- Remove link from author byline on posts.

## Changes

- Get rid of the authors filter and replace it with a languages filter.

## How to test this PR

1. Pull branch and run a local server.
2. Language filter should appear and be functional on any filter with a Topics filter:

http://localhost:8000/administrative-adjudication-proceedings/administrative-adjudication-docket/
http://localhost:8000/compliance/supervisory-guidance/
http://localhost:8000/data-research/research-reports/
http://localhost:8000/rules-policy/final-rules/
http://localhost:8000/consumer-tools/everyone-has-a-story/
http://localhost:8000/rules-policy/rules-under-development/
http://localhost:8000/data-research/cfpb-researchers/
http://localhost:8000/consumer-tools/credit-reports-and-scores/consumer-reporting-companies/companies-list/
http://localhost:8000/about-us/events/
http://localhost:8000/rules-policy/notice-opportunities-comment/open-notices/
http://localhost:8000/rules-policy/notice-opportunities-comment/archive-closed/
http://localhost:8000/compliance/amicus/briefs/
http://localhost:8000/about-us/events/archive-past-events/

## Screenshots

Screen Shot 2021-07-02 at 3 00 45 PM


anselmbradford:
Holding as we need to restrict this to certain filters and which is yet to be determined.

anselmbradford:
This filter should appear only on these pages:

https://www.consumerfinance.gov/activity-log/
https://www.consumerfinance.gov/about-us/newsroom/
https://www.consumerfinance.gov/about-us/blog/

However, we'll turn it off the others after deploy.

cfgov/v1/tests/test_forms.py
```diff
@@ -10,17 +10,6 @@
 
 class TestFilterableListForm(TestCase):
 
-    @mock.patch('v1.forms.FilterableListForm.__init__')
-    @mock.patch('taggit.models.Tag.objects')
-    def test_set_authors_filters_tags_on_pageids(self, mock_tag_objects, mock_init):
```
anselmbradford, line 15: @willbarton Should I be replacing this test with something? The languages are a hardcoded list as opposed to the generated list of the authors filter, so I wasn't sure if they needed a dedicated test.

willbarton: I don't think so? I think removing this is fine.

cfgov/v1/tests/test_filterable_list_form.py
```diff
@@ -31,7 +31,7 @@ def setUpTestData(cls):
         blog1.categories.add(CFGOVPageCategory(name='bar'))
         blog1.tags.add('foo')
         blog1.authors.add('richa-agarwal')
-        blog1.authors.add('sarah-simpson')
+        blog1.language.add('es')
```
willbarton, line 34: For model fields like `authors`, the `add()` method adds a related object to them. Because `language` is a field on the model, you could actually pass it in the constructor (`BlogPage(title='test page', language='es')`), or just use `blog1.language = 'es'`.

cfgov/v1/forms.py
```diff
@@ -141,13 +139,13 @@ def __init__(self, *args, **kwargs):
         self.all_filterable_results = self.get_all_filterable_results()
         page_ids = self.get_all_page_ids()
         self.set_topics(page_ids)
-        self.set_authors(page_ids)
+        self.set_languages(page_ids)
```
willbarton, line 142: It's actually unnecessary to have a separate method here for this, nor one that takes `page_ids` since `languages` isn't dependent on them. You could replace this will the line

```python
self.fields['language'].choices = ref.supported_languages
```

Below.

anselmbradford: Removed in https://github.com/cfpb/consumerfinance.gov/pull/6506/commits/1d5aec108fc0b10390d89419e9d90038e04e3ea9

willbarton (approved):

PR #6512, opened 2021-07-09
Fix Cypress filter tests | Update to Cypress@7.7.0

anselmbradford:
https://github.com/cfpb/consumerfinance.gov/pull/6506 neglected to update the Cypress tests to check for language instead of author.

Additionally, the typeahead tests were not correctly testing anything as they forced clicking the multiselect items, regardless if they were visible or not. This PR fixes that.

## Changes

- Update tests to test the language multiselect in place of the former author multiselect.
- Add type-ahead utility methods for testing the type-ahead in topics in languages.
- Fix erroneous type-ahead test logic to have it actually type and click a visible result from the multiselect.
- Update to Cypress@7.7.0

## How to test this PR

1. Pull branch, run `yarn run cypress run`. It should pass.


willbarton:
Weirdly, the tests pass with `yarn run cypress run --browser chrome`. But the default Electron browser renders our multi-select controls wrong, allowing the `ul` to overflow its `fieldset` and thus be unscrollable, but `li`s that are outside of the `fieldset` are not detected as "visible":

Image Pasted at 2021-7-19 13-18

This doesn't happen in Electron with `yarn run cypress open`, and it doesn't happen when using `--browser chrome`.

willbarton:
I added `--browser chrome` to the Jenkins job and the functional tests now pass as-is on this PR. @anselmbradford I think I'm satisfied with that, and @wpears signed off on it.

willbarton (approved):

PR #6516, opened 2021-07-13
Run our internal functional test job against PR deployments

willbarton:
This change has our dev Docker deploys run our internally defined functional test job against the newly deployed stack and report back to PRs.

This should give us the benefit of running the functional test suite against PRs again.

## Checklist


higs4281 (approved):
Looks good from here, but I'm not sure how to test.

willbarton:
@higs4281 I think really the only thing here is that the functional test run showed back up in the GitHub checks. It failed, but @anselmbradford is working on that in #6512.

PR #6517, opened 2021-07-14
Catch an ElasticSearch RequestError in autocomplete

willbarton:
Occasionally we hit an `elasticsearch7.exceptions.RequestError` in the Ask CFPB autocomplete search method because the `search_phase_execution_exception` character limit is reached. This change catches that error so that it doesn’t propogate, and returns an empty of list of autocomplete results instead.

I've also added explicit tests for `AnswerPageSearch.autocomplete`.

## Checklist


higs4281 (approved):
Looks good.

PR #6452, opened 2021-05-19
Update tox to use Wagtail 2.13 for unittest-future

cwdavies:
I have been using this documentation to help in the upgrade to wagtail 2.13:
https://docs.wagtail.io/en/stable/releases/2.13.html
https://github.com/wagtail/wagtail/releases/tag/v2.13
https://docs.wagtail.io/en/latest/advanced_topics/customisation/streamfield_blocks.html
New in version 2.13: The data-contentpath attribute is now required on a containing element around the render_form output.
Adding support for form_classname meta in FieldBlock in https://github.com/wagtail/wagtail/pull/6470

This commit has been made to Remove now-unused render_form methods but is not mentioned in release notes?
https://github.com/wagtail/wagtail/commit/d0f0c30cbb343df8617fd384b7c7d23c56656132#diff-738f6d73ebe8cc8440bfde3968c3a2c04b88c0f2a627406bb9c8fe78eb587f70

## Additions

- Add `Meta` class to `PlaceholderCharBlock` class in `cfgov/v1/blocks.py` for wagtail 2.13

## Changes

- Update TestInfoUnitGroup in `cfgov/v1/tests/atomic_elements/organisms/test_organisms.py`

## How to test this PR

1. tox -e lint
2. tox -e unittest-current v1.tests.test_blocks
3. tox -e unittest-future v1.tests.test_blocks
4. tox -e unittest-current v1.tests.atomic_elements.organisms.test_organisms
5. tox -e unittest-future v1.tests.atomic_elements.organisms.test_organisms.TestInfoUnitGroup

## Notes and todos

Improve code coverage and create tests for the following organisms:
- FilterableListControls
- Sidebar Breakout
- PostPreviewSnapshot
- RelatedPosts

## Checklist

- [x] PR has an informative and human-readable title
  - PR titles are used to generate the change log in [releases](../../releases); good ones make that easier to scan.
  - Consider prefixing, e.g., "Mega Menu: fix layout bug", or "Docs: Update Docker installation instructions".
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)
- [x] Future todos are captured in comments and/or tickets
- [ ] Project documentation has been updated, potentially one or more of:
  - [This repo’s docs](https://cfpb.github.io/consumerfinance.gov/) (edit the files in the `/docs` folder) – for basic, close-to-the-code docs on working with this repo
[...]

PR #6520, opened 2021-07-15
Remove heading requirement for InfoUnitGroup

willbarton:
This change removes the attempt to require a heading if `InfoUnitGroup` blocks have intro text.

The attempt to require headings apparently did not work previously, and with our changes to that test in #6452, is now causing content issues.

## Checklist


niqjohnson:
@willbarton: We should also remove the help text in [`/cfgov/v1/atomic_elements/organisms.py#L80-L81`](https://github.com/cfpb/consumerfinance.gov/blob/fix/infounitgroup/cfgov/v1/atomic_elements/organisms.py#L80-L81) (I think).

willbarton:
@niqjohnson indeed! That's done in effd51d.

niqjohnson:
@willbarton, the help text changes needs migrations too, right?

niqjohnson (approved):
Everything was working as expected for me in the PR preview environment. I tested all the heading and intro text combinations I could think of—heading only, intro only, heading and intro, no heading or intro—and all of them previewed and published fine.

PR #6521, opened 2021-07-16
Fix Ask CFPB table block rendering

willbarton:
Prior to [Wagtail 2.12.5](https://docs.wagtail.io/en/stable/releases/2.12.5.html) `include_block` was improperly escaping HTML, resulting in CVE-2021-32681 (🎩  to @niqjohnson for catching this). Unfortunately we were relying on that behavior for rendering of table blocks in Ask CFPB content.

This fix explicitly turns off autoescape for table blocks in Ask CFPB content blocks.

## Checklist

### DIFF
--- a/cfgov/ask_cfpb/tests/test_blocks.py
+++ b/cfgov/ask_cfpb/tests/test_blocks.py
@@ -1,6 +1,8 @@
 from django.test import TestCase, override_settings
 
-from ask_cfpb.models.blocks import FAQ, AskAnswerContent, HowTo, Tip
+from ask_cfpb.models.blocks import (
+    FAQ, AskAnswerContent, AskContent, HowTo, Tip
+)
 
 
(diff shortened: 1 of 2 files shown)


chosak (approved):
Thanks for the quick fix. Would be nice to have a regression test for this if it's not too much trouble. Maybe something along [these lines](https://github.com/cfpb/consumerfinance.gov/blob/35d71158c6f1921804b56649717f8059cca63b52/cfgov/ask_cfpb/tests/models/test_pages.py#L847), making an `AnswerPage` with some data that includes a table, and then rendering it. But I realize that may be hard given the way the tests are written.

niqjohnson (approved):
I checked all that Ask pages that had tables, and the tables all rendered for me locally. The non-Ask pages with tables I checked continued to render fine.

PR #6526, opened 2021-07-21
CCDB: Remove non-responsive header and footer CSS

anselmbradford:
Our legacy base template that ccdb-ui uses has special non-responsive styles for the header and footer that appear to do nothing. The page is unusable in IE9 anyway. 

## Removals

- non-responsive header and footer CSS and associated gulp tasks.

## How to test this PR

1. Pull branch, build and run a local server.
2. Boot up Sauce Connect.
3. In Sauce Labs IE9 compare http://localhost:8000/data-research/consumer-complaints/search/ to https://www.consumerfinance.gov/data-research/consumer-complaints/search/
4. There should be a menu (albeit difficult to see with all the reloading that happens)

### DIFF
--- a/cfgov/legacy/templates/front/base_update.html
+++ b/cfgov/legacy/templates/front/base_update.html
@@ -91,11 +91,6 @@
     {% block page_css %}
     {% endblock %}
 
-    <!--[if lt IE 9]>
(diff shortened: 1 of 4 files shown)


flacoman91 (approved):
Awesome. glad to have this removed.

PR #6528, opened 2021-07-22
Info-unit: Add aria-label from Wagtail to links in template

anselmbradford:
https://github.com/cfpb/consumerfinance.gov/pull/6497 added an aria-label to links edited in Wagtail, but neglected to add the aria-label to the info-unit template, so the entered aria-labels don't show up in that case.

## Changes

- Add aria-label to info-unit template

## How to test this PR

1. Edit a page with an info-unit and add an aria-label to a link and see it shows up in the published page's info-unit link.


cfgov/jinja2/v1/_includes/molecules/info-unit-2.html
```diff
@@ -108,8 +108,14 @@
...
-                    {%- set link = '<a class="m-list_link" href="' ~
-                        link.url ~ '">' ~ link_text ~ '</a>' %}
+                    {%- if link.aria_label -%}
+                        {%- set link = '<a class="m-list_link" href="' ~
+                            link.url ~ '" aria-label="' ~
+                            link.aria_label ~ '">' ~ link_text ~ '</a>' %}
+                    {%- else -%}
+                        {%- set link = '<a class="m-list_link" href="' ~
+                            link.url ~ '">' ~ link_text ~ '</a>' %}
+                    {%- endif %}
```
niqjohnson, line 118: Looks like as of Jinja 2.8 we can [now use block assignments to set variables](https://jinja.palletsprojects.com/en/3.0.x/templates/#block-assignments). Can we do something like this to make it a little less repetitive?

```suggestion
{%- set link -%}
    <a class="m-list_link"
        href="{{link.url}}"
        {%- if link.aria_label -%}
            aria-label="{{link.aria_label}}"
        {%- endif %}>{{link_text}}</a>
{%- endset -%}
```

Seems to work in my quick testing, but I'm not sure if there would be any side effects doing it this way.

anselmbradford: 💯 Great update! Done here https://github.com/cfpb/consumerfinance.gov/pull/6528/commits/bed95b797b35527feb3039ded846db963f807273

niqjohnson: 👍

niqjohnson (approved):
I tested this on a couple of pages locally. All the links in info unit groups I added ARIA labels to rendered the `aria-label` attribute as expected, and all the info unit links with blank ARIA labels rendered as before (with no `aria-label` attribute). Looking good to me.

PR #6529, opened 2021-07-23
Expand fake-school options

higs4281:
We've had an obviously fictional school – Example University – in our paying_for_college database for several years, for demo purposes. Now we want to have three more options, in order to show schools typical of different sizes.

This change adds three demo-school primary keys to be excluded from updates
and most other live operations. So the former `FAKE_SCHOOL_PK` constant becomes a list.

This should have no effect on the legacy EFIP tool or the financial-path tool.

### DIFF
--- a/cfgov/paying_for_college/models/disclosures.py
+++ b/cfgov/paying_for_college/models/disclosures.py
@@ -16,7 +16,13 @@ import requests
 # Our database has a fake school for demo purposes
 # It should be discoverable via search and API calls, but should be excluded
 # from cohort calculations and from College Scorecard API updates.
-FAKE_SCHOOL_PK = 999999
+FAKE_SCHOOL_PKS = [
+    999999,
+    999998,
+    999997,
+    999996,
+]
+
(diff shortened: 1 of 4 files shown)


cfgov/paying_for_college/models/disclosures.py
```diff
@@ -16,7 +16,13 @@
 # Our database has a fake school for demo purposes
 # It should be discoverable via search and API calls, but should be excluded
 # from cohort calculations and from College Scorecard API updates.
```
willbarton, line 18:
```suggestion
# Our database has fake schools for demo purposes
# They should be discoverable via search and API calls, but should be excluded
# from cohort calculations and from College Scorecard API updates.
```

willbarton (approved):
Seems okay to me!

cfgov/paying_for_college/models/disclosures.py
higs4281: Thanks for catching that.

PR #6531, opened 2021-07-26
Allow HomePages to be created anywhere

chosak:
Currently HomePages can only be created at the site root. We want themto be able to exist elsewhere for better multi-language support. This commit removes this restriction to allow for creation of HomePages at any location in the page tree.

See platform#4103 for additional context.

## Checklist

### DIFF
--- a/cfgov/v1/models/home_page.py
+++ b/cfgov/v1/models/home_page.py
@@ -55,9 +55,6 @@ class HomePage(CFGOVPage):
         ),
     ])
 
-    # Sets page to only be createable at the root
-    parent_page_types = ['wagtailcore.Page']
-
     objects = PageManager()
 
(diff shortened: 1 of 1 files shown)


anselmbradford (approved):
Probably not worth it, but would we possibly want to change the page type name since it's not a homepage if it appears other places than the root?

Screen Shot 2021-07-26 at 2 29 49 PM

chosak:
@anselmbradford thanks. I'm thinking we may run into this as we iterate on homepage implementation, for example if we want to keep the existing design on some pages and move to a new one on others, we may need to split out the Wagtail page types. I'm inclined to punt on any renaming until that path is a little clearer - plus such a change would necessitate additional code changes / potential migrations.

PR #6532, opened 2021-07-26
Filters: Set languages drop-down based on languages in the result set

anselmbradford:
Follow on to https://github.com/cfpb/consumerfinance.gov/pull/6506

## Changes

- Filters: Set languages drop-down based on languages in the result set

## How to test this PR

1. The following pages should have language filters with only the languages in the results set. For instance, newsroom is only english and spanish.

## Notes and todos

- Unit tests to come…


chosak:
@anselmbradford I've added a unit test as we discussed; I also modified how the language choices are determined, leveraging an Elasticsearch aggregation instead of doing another pass over all of the results. I think we might be able to do a similar optimization for the topic choices, as well.

To help with this work I added the ability to log Elasticsearch queries to console, including a new section of the docs:

image

Could you confirm that the language choices work as expected for you? If so I am :+1: for this PR.

chosak (approved):

PR #6534, opened 2021-07-26
Regs3k: update dependencies to fix security warning

anselmbradford:
The `browserslist` and `lodash` dependencies in regs3k app have security warnings 
https://github.com/cfpb/consumerfinance.gov/security/dependabot

This PR refreshes the yarn lock and npm cache to pull in the latest. It also updates workbox while we're in here and adds a url for the workbox script.

## Changes

- Regs3k: refresh npm cache.
- Update to "workbox-build": "6.1.5" from "6.0.2"
- Adds url for workbox script to urls.py

## How to test this PR

1. Delete `cfgov/regulations3k/jinja2/regulations3k/regulations3k-service-worker.js` and any `workbox-` scripts and sourcemap files.
1. Pull branch and run `gulp clean && gulp build`.
1. Visit a regs3k page like http://localhost:8000/rules-policy/regulations/1002/3/ and see that there are no errors in the console and it works as expected.


cfgov/cfgov/urls.py
```diff
@@ -464,6 +464,15 @@ def empty_200_response(request, *args, **kwargs):
         name='regulations3k-service-worker.js'
     ),
 
+    re_path(
```
anselmbradford, line 467: bewds, does this make sense?

cfgov/cfgov/urls.py
```diff
@@ -464,6 +464,15 @@ def empty_200_response(request, *args, **kwargs):
...
+        r'^workbox-99cd3a5e.js',
```
contolini, line 468: Is `99cd3a5e` a cache-busting hash that changes every time we deploy?

anselmbradford: Mm this crossed my mind too. It's given that name in my regulations3k-service-worker.js. What name do you have in yours (after throwing that file out and running gulp build)?

anselmbradford: If I throw it out and rebuild it gets the same name, which made me think maybe it was a versioning thing versus a cache busting thing, but maybe I'm confused.

willbarton: Hmm, do we know when/if this might change?

Django's [`ManifestStaticFilesStorage`](https://docs.djangoproject.com/en/3.2/ref/contrib/staticfiles/#django.contrib.staticfiles.storage.ManifestStaticFilesStorage) does create cache-busting names for all our static files, but that happens automatically and won't make us use hard-coded paths.

anselmbradford: It's created by the JavaScript on `gulp scripts:apps` at https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/apps/regulations3k/webpack-config.js#L139

The service-worker file (`cfgov/regulations3k/jinja2/regulations3k/regulations3k-service-worker.js`) gets the name of the workbox script so it can load it. So it has `define("./regulations3k-service-worker.js",['./workbox-99cd3a5e'], function (workbox) { …`, (or `define("./regulations3k-service-worker.js",['./workbox-54723c33'], function (workbox) { …` before this PR).

Safest would be if we could make the URL `r'^workbox-*.js'` where `*` is a wildcard. Is that possible?

willbarton: That would be possible, because it is a regular expression. I'm wary about the `template_name` below being tied directly to that file, and thus having to be edited if the file changes (same with `name`, but we could probably leave `name` out — it's only useful if we're going to try to do a reverse match for this URL in Python, which I doubt we'll need to).

anselmbradford: So I was just following the format of the url scheme for `regulations3k-service-worker.js` above. How should this one be done? The script `regulations3k-service-worker.js` needs a URL it can load `workbox-99cd3a5e.js` from and I think it will look at the same path where it itself lives.

willbarton: The other thought I have, mostly because I keep trying to wrap my head around how all of this might play with `ManifestStaticFilesStorage` (my gut suggests, it won't because these are defined as templates?) and keep getting pulled away before I can: can we deploy this branch to a dev server to ensure it works with our production collectstatic and static file serving configurations?

willbarton: @anselmbradford I'm not terribly familiar with service worker requirements, but if `regulations3k-service-worker.js` needs it to live side-by-side, then this seems like the right approach. I'm mostly worried about the hashed name and the overhead that might add to maintain, and whether `ManifestStaticFilesStorage` would break that when deployed (since it does some light editing of file paths in static files to use cache-busting hashed filenames). 

I think that second concern is probably moot because, to Django, these appear as templates, not static files, but I'd like to be sure.

anselmbradford: @contolini after chatting with Will it does appear this is a cache hash. To workaround these configuration issues, I turned on the `inlineWorkboxRuntime` option. The file is about 15k. Does that seem okay to you?

Screen Shot 2021-08-12 at 11 35 03 AM

From https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-build#.generateSW

willbarton (approved):

PR #6536, opened 2021-07-27
Add rental assistance finder tool

virginiacc:
Imports external rental assistance finder tool and makes it available in Wagtail.

## Additions

- RAF tool Wagtail block and template
- RAF tool JS
- Gulp task to copy RAF tool JS to on-demand directory  

## How to test this PR

2. Create browse page and add RAF Tool block
3. Verify that tool shows on the page and works as expected

### Rental assistance finder tool in a Wagtail browse page

Screen Shot 2021-07-26 at 9 51 52 PM

## Notes and todos

- A future iteration will import the JS in an npm module, add the CSS as a file rather than a style block in the tool block's template, and pull the JSON data from s3 rather than bundling it in the tool's main js file.

#### Browser testing

- [x] Firefox
- [x] Chrome
- [x] Safari
- [x] Safari on iOS
- [x] Chrome on Android

#### Accessibility

- [ ] Screen reader friendly

#### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


willbarton:
I'll take care of the migration test failure.

gulp/tasks/copy.js
```diff
@@ -75,6 +83,7 @@ gulp.task( 'copy',
   gulp.parallel(
     'copy:icons',
     'copy:lightbox2',
+    'copy:erap',
```
anselmbradford, line 86: Why's the copy task necessary? Can we put `cfgov/unprocessed/erap/js/main.js` in `cfgov/unprocessed/js/routes/on-demand/erap/main.js` to start with?

Where is the source for this main.js file?

anselmbradford (changes requested):

gulp/tasks/copy.js
virginiacc: @anselmbradford The tool is maintained as an [external standalone React app](https://github.com/cfpb/rental-assistance-finder). At the moment we're compiling and importing the JS directly because the app hasn't yet been released as an npm module. Once it is, the copy task will move the built JS file from `node_modules` to its `static_built` destination, like the existing lightbox2 task, and the current `unprocessed` source directory will be removed.

virginiacc: At the moment we're planning to have the tool on two pages, the English page that's launching today and a Spanish page (with translated tool content) that will be added in the near future.

anselmbradford (commented):
Discussed in chat. Big picture we want the JS/CSS assets placed in https://github.com/cfpb/consumerfinance.gov/tree/main/cfgov/unprocessed/apps and then it would have it's own package.json and when it's a npm module it would install in there. It also doesn't seem like we want it added to browse pages if it's only for two pages and instead would want dedicated templates like https://github.com/cfpb/consumerfinance.gov/tree/main/cfgov/jinja2/v1/owning-a-home or similar.

anselmbradford (approved):
Newline in copy.js looks like it was left by mistake, otherwise looks fine.

willbarton:
> It also doesn't seem like we want it added to browse pages if it's only for two pages and instead would want dedicated templates like https://github.com/cfpb/consumerfinance.gov/tree/main/cfgov/jinja2/v1/owning-a-home or similar.

This is a bigger-picture item, but I disagree. There's no harm in creating these components as blocks that can be added to standard page types.

anselmbradford:
> This is a bigger-picture item, but I disagree. There's no harm in creating these components as blocks that can be added to standard page types, even if they may currently be limited to one or two pages. It leaves their placement and surrounding copy entirely up to content editors.

This is not entirely harmless. Unless we have completely standardized spacing and a testable matrix of all possible page block combinations there very well could be two blocks that end up next to each other with too little or too much margin/padding or other layout issues. That puts the onus on the content editors to report page combinations that are breaking, if they notice, since the big picture of how layouts should be arranged may only be known by design/UX. Then the fewds need to fix a theoretical set of page combinations that may not even ever be used, or only fix the ones in use and risk breaking a future combination.

willbarton:
@anselmbradford I get that concern, and that's actually one side of a larger discussion in the Wagtail community. The two possible approaches are a small number of page types that can have a large number of combinations of blocks (which is the route we went down very early on), or a large number of page types with a small number of block options. The Wagtail community seems fairly evenly divided in terms of which approach is chosen.

I have a strong preference for things that live in the page tree, rather than templates and hard-coded urls. Without re-architecting the CMS, that means we either create new page types that can have a small combination of blocks (i.e. in this case, a `RAFToolPage` that has the appropriate content blocks for the tool pages, and nothing else), and document them as such, or we add these blocks to our existing page types (which is what CFPT chose to do here). I personally don't have a preferred approach between those.

One thing I think we don't have the personnel to maintain is additional pages that are templates with content only editable by developers. That's what I would personally like to avoid.

anselmbradford:
Thanks for the context! FWIW I think distinct page types, at least in regard to a "tool," seem to make more sense to me. If the tool is a block that can be added to a whole bunch of page types, that implies that there is the intention to use it on a variety of page types (which should be the way, if that's the case), but if it's only ever going to live in one place (or two, with a translated content page), it seems like a page type is the way.

I guess my thought with a template is if you have a block that gets added to a page with no settings and nothing else, what's the point? A developer still needs to go in and edit it to make changes, but now has the further complication of figuring out how it lives in the Wagtail database. But I suppose with the Block Inventory that's not as big a deal.

PR #6537, opened 2021-07-27
Add a max length for Ask search fields

niqjohnson:
**This is a work in progress**. It still needs unit and functional tests, but before that, I'm hoping a front-end developer can give everything a look to see if the code is on the right track and suggest any improvements.

Adds a max length to Ask search fields to prevent visitors from accidentally trying to enter complaints there. See GHE/CFGOV/platform/issues/4091 for more details.

---

## Additions

- Max length to the autocomplete molecule and Ask search organism

## How to test this PR

1. Go to the complaint landing page, `/complaint/`. Edit that page in Wagtail and check the "Direct long searches to submit a complaint" checkbox. Publish that change.
2. Try pasting or typing more than 50 characters into the Ask search field. You should see a field-level error and be prevented from entering any more text, like in the screenshot below.
3. Once you have 50 characters in the box, try deleting one or two. You should get autocomplete results as normal again.
4. Try deleting all but one character. Autocomplete results should disappear until you type a second character (as they do now).
5. Go to another page with an autocomplete-enabled Ask search field, like `/ask-cfpb/`. Try pasting or typing more than 50 characters into that field. You should get the same error, but the error message should only show the first sentence ("Searches are limited to 50 characters.").

## Screenshots

search-error

## Notes and todos

- [x] Get answers to the slew of questions I put in the code as TODO comments
- [x] Check all the browser testing steps down below
- [x] ~~Write unit tests~~ (there aren't any existing unit tests for any of the files I changed, so I'm going to keep that task as an effort separate from this PR
- [x] Write functional tests
- [x] Set up a custom analytics event
- [x] Replicate a similar max length validation on the back end to make this work without Javascript. I'll put that in a separate PR, though.

## Checklist

#### Browser testing

- [x] Firefox
- [x] Chrome
- [x] Safari
- [x] Safari on iOS
- [x] Chrome on Android

#### Accessibility

- [x] Screen reader friendly

#### Other

- [x] Is useable without CSS
- [x] Is useable without JS


cfgov/jinja2/v1/_includes/organisms/ask-search.html
```diff
@@ -54,7 +57,21 @@
                                name="q"
                                id="o-search-bar_query"
                                value="{{ ask_query }}"
-                               placeholder="{{ placeholder }}">
+                               placeholder="{{ placeholder }}"
+                               maxlength="{{ max_length }}"
+                               aria-describedby="o-search-bar-error_message">
+                    </div>
+                    {# TODO: Is there a better way to render this hidden initially? #}
```
anselmbradford, line 64: I think this is fine. This is a patten we use elsewhere. Have `u-hidden` in markup and remove using JS.

cfgov/unprocessed/js/molecules/Autocomplete.js
```diff
@@ -45,9 +52,21 @@ function Autocomplete( element, opts ) {
...
+       https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute#non-existing_attributes */
+    // 1024 is our upper limit for Elasticsearch queries
+    maxChars: _input.getAttribute( 'maxlength' ) ?
+      _input.getAttribute( 'maxlength' ) :
+      1024,
     delay: 300,
     url: '',
     list: [],
+
+    /* TODO: Should toggling the error message on and off be handled here in the
```
anselmbradford, line 66: I'd move it to ask-autocomplete.js since that's where the dom reference is and you now are broadcasting an event that can be listened to.

cfgov/unprocessed/js/molecules/Autocomplete.js
```diff
@@ -121,20 +140,56 @@ function Autocomplete( element, opts ) {
...
+      maxLengthExceeded: _maxLengthExceeded
+    } );
+  }
+
   /**
    * Binds input, blur, and keydown events on _input element and
    * resize event on window.
    */
   function _bindEvents() {
     _input.addEventListener( 'input', function() {
```
anselmbradford, line 175: I'd probably move the event handler function to it's own named function, like is done with `handleMaxCharacters`. It should get an `_` prefix on the name as it isn't exported from `Autocomplete`.

cfgov/unprocessed/js/molecules/Autocomplete.js
```diff
@@ -37,6 +40,10 @@ function Autocomplete( element, opts ) {
   let _suggestions;
   let _isVisible;
   let _selection;
+  let _maxLengthExceeded = false;
+
+  /* TODO: Is there a better way to get this to use later to dispatch the event? */
```
anselmbradford, line 45: Have you confirmed changing `_this.dispatchEvent` to `this.dispatchEvent` on line 165 doesn't work? It probably will be referencing `_input` huh.

cfgov/jinja2/v1/_includes/organisms/ask-search.html
niqjohnson: Works for me. I removed the TODO in 8b4c07f.

cfgov/unprocessed/js/molecules/Autocomplete.js
niqjohnson: Yeah, the more I think about this, the more I think it makes sense to move the error message stuff into the Ask organism. That keeps the underlying autocomplete molecule focused only on the input field and the results list. I made that change in 8b4c07f.

cfgov/unprocessed/js/molecules/Autocomplete.js
niqjohnson: Done in 8b4c07f—the `input` event listener now calls `_handleInput` instead of an anonymous function.

cfgov/unprocessed/js/molecules/Autocomplete.js
niqjohnson: Yeah, `this` in that context is the `_input` HTML node. Given that, is there a more preferred way to do this? Looks like we do `const _this = this` in a couple of places (like [cfgov/unprocessed/js/modules/TabTrigger.js#L31](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/js/modules/TabTrigger.js#L31), but I'm happy to change it to whatever if you have a preference, @anselmbradford.

anselmbradford: That seems fine. You can also do `_instance` if that's clearer https://github.com/cfpb/design-system/blob/main/packages/cfpb-forms/src/organisms/Multiselect.js#L66 

Alternative, you can use bind to set the value of `this` by created a binded alias of the `_toggleMaxLengthError` function, like [this](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/js/modules/behavior/FlyoutMenu.js#L55).

cfgov/unprocessed/js/molecules/Autocomplete.js
```diff
@@ -121,23 +135,58 @@ function Autocomplete( element, opts ) {
...
+    _this.dispatchEvent( 'maxCharacterChange', {
+      maxLengthExceeded: _maxLengthExceeded
+    } );
+  }
+
+  /**
+   * Event handler for input into the _input element
+   */
+  function _handleInput() {
+    _searchTerm = this.value;
```
anselmbradford, line 163: I'd set this through the event parameter. 
Without testing I think it would be:
```
function _handleInput( evt ) {
  _searchTerm = evt.target.value;
```

cfgov/unprocessed/js/molecules/Autocomplete.js
anselmbradford: Hmm relevant to the `_this` question, you could try 

```
const _handleInputBinded = _handleInput.bind( this );
_input.addEventListener( 'input', _handleInputBinded );
```
… and the value of `this` should be correct inside `_toggleMaxLengthError `. You'd need to change `_searchTerm = this.value;` to get the value through the event parameter though.

cfgov/unprocessed/js/molecules/Autocomplete.js
```diff
@@ -153,6 +153,7 @@ function Autocomplete( element, opts ) {
 
   /**
    * Event handler for input into the _input element
+   * @param {Object} event The event object
```
anselmbradford, line 156: You could log this to confirm, but pretty sure the type here would be [`InputEvent`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event) not `Object`.

cfgov/unprocessed/js/molecules/Autocomplete.js
niqjohnson: I went with `_instance` 1e506c1—that felt like the most readable option.

cfgov/unprocessed/js/molecules/Autocomplete.js
niqjohnson: See https://github.com/cfpb/consumerfinance.gov/pull/6537#discussion_r680220429. Since we're keeping `const _instance = this`, I left this as is

jslay-excella (approved):
Looks very well done and complete.

cfgov/jinja2/v1/_includes/organisms/ask-search.html
```diff
@@ -54,7 +57,20 @@
...
+                               placeholder="{{ placeholder }}"
+                               maxlength="{{ max_length }}"
+                               aria-describedby="o-search-bar-error_message">
+                    </div>
+                    <div class="a-form-alert a-form-alert__error u-hidden"
+                         id="o-search-bar-error_message"
+                         role="alert">
+                        {{ svg_icon('error-round') }}
+                        <p class="a-form-alert_text">
+                            Searches are limited to 50 characters.
```
higs4281, line 69: We could avoid a hard-coded value here.

```suggestion
                            Searches are limited to {{ autocomplete_max_chars }} characters.
```

cfgov/unprocessed/js/molecules/Autocomplete.js
```diff
@@ -121,23 +130,60 @@ function Autocomplete( element, opts ) {
       inputCoords.left ) + 'px';
   }
 
+  /**
+   * Toggles the error state and message that's shown when the
+   * max search term length is hit
+   */
+  function _toggleMaxLengthError() {
+    _maxLengthExceeded = _searchTerm.length >= _settings.maxChars ? true :
```
higs4281, line 138: Should this test just be greater-than? Currently, the warning pops when a user *hits* 50 chars, effectively limiting search terms to 49 characters.

```suggestion
    _maxLengthExceeded = _searchTerm.length > _settings.maxChars ? true :
```

higs4281 (approved):
This is looking great. I have one minor suggestion:

- The limit is hard-coded in one place.

cfgov/jinja2/v1/_includes/organisms/ask-search.html
higs4281: Nit to be ignored at will:   
I had noticed that freezing the input at 50 characters in practice limits a search value to 49 characters. I thought it might be easy to change the [test](https://github.com/cfpb/consumerfinance.gov/pull/6537/files#diff-80d79644e1dbe5feb3a4ca46b46b6815c7a1e6f522aace7b712d67de99587019R138) to `>` instead of `>=`, but that messes up the interactions, and it doesn't seem worth it to chase the semantic difference between whether the limit is the barrier or the last allowed value. It does the job well either way.

niqjohnson: Good call on using the new variable in the error message—I made that change in 8fb3d3c.

I went back and forth on whether the 50th or 51st character should trigger the error when I was setting this up. From the search data I've seen I don't think an extra character will make much of a difference, so I'm inclined to leave it as is for now.

cfgov/unprocessed/js/molecules/Autocomplete.js
```diff
@@ -121,23 +130,60 @@ function Autocomplete( element, opts ) {
...
+        _toggleMaxLengthError();
+      }
+      if ( _settings.url ) {
+        _throttleFetch();
+      } else {
+        _settings.filterList( _settings.list );
+      }
+    } else if ( _searchTerm.length >= _settings.maxChars ) {
+      _toggleMaxLengthError();
+    } else {
```
anselmbradford, line 173: Might be better organized to be:

```
} else if ( _maxLengthExceeded === true ) {
  _toggleMaxLengthError();
} else {
  _hide();
}
```

cfgov/unprocessed/js/molecules/Autocomplete.js
```diff
@@ -121,23 +130,60 @@ function Autocomplete( element, opts ) {
...
+
+  /**
+   * Event handler for input into the _input element
+   * @param {InputEvent} event The input event object
+   */
+  function _handleInput( event ) {
+    _searchTerm = event.target.value;
+    if ( _searchTerm.length >= _settings.minChars &&
+         _searchTerm.length < _settings.maxChars ) {
+      if ( _maxLengthExceeded === true ) {
```
anselmbradford, line 163: Could this be moved to the top of the function like this? 

```
/**
 * Event handler for input into the _input element.
 * @param {InputEvent} event - The input event object.
 * @returns {boolean} False if maximum character limit was reached, true otherwise.
 */
function _handleInput( event ) {
  if ( _maxLengthExceeded === true ) {
    _toggleMaxLengthError();
    return false;
  }

  if ( _searchTerm.length >= _settings.minChars &&
       _searchTerm.length < _settings.maxChars ) {
    if ( _settings.url ) {
      _throttleFetch();
    } else {
      _settings.filterList( _settings.list );
    }
  } else if ( _searchTerm.length >= _settings.maxChars ) {
    _toggleMaxLengthError();
  } else {
    _hide();
  }
    
  return true;
}
```

cfgov/unprocessed/js/molecules/Autocomplete.js
anselmbradford: Superseded by https://github.com/cfpb/consumerfinance.gov/pull/6537/files#r701994064

cfgov/jinja2/v1/_includes/organisms/ask-search.html
```diff
@@ -54,7 +63,23 @@
                                name="q"
                                id="o-search-bar_query"
                                value="{{ ask_query }}"
-                               placeholder="{{ placeholder }}">
+                               placeholder="{{ placeholder }}"
+                               maxlength="{{ max_length }}"
+                               aria-describedby="o-search-bar-error_message">
+                    </div>
+                    <div class="a-form-alert a-form-alert__error u-hidden"
+                         id="o-search-bar-error_message"
```
anselmbradford, line 71: Technically in BEM you'd have a block with an element inside it. In this case we have `o-search-bar-error` as the block and `_message` as the element. We don't have a `o-search-bar-error` ID as a container. We do have `o-search-bar` as a class. We may want to make this `o-search-bar_error-message` to say even though this is an ID it is the `_error-message` element of the `o-search-bar` block.

cfgov/unprocessed/js/routes/on-demand/ask-autocomplete.js
```diff
@@ -3,13 +3,33 @@
...
   es: '/es/obtener-respuestas/api/autocomplete/?term='
 };
 
 const autocompleteContainer = document.querySelector( '.m-autocomplete' );
+const errorMessage = document.querySelector( '#o-search-bar-error_message' );
+const submitButton = document.querySelector( '.o-search-bar button[type="submit"]' );
+
+/**
+ * Disable the submit button if the query character limit is reached
+ * @param {Object} event The maxCharacterChange event object dispatched from the
```
anselmbradford, line 19: I usually put a dash after the param name and add a period to the end, like:

```
* @param {Object} event - The maxCharacterChange event object dispatched from the
* autocomplete.
```

cfgov/unprocessed/js/molecules/Autocomplete.js
niqjohnson: Fixed in 608db2c59260146ac4f6e5630f84471ed0d60758, although a little differently than what's suggested here. This bit…

```javascript
if ( _maxLengthExceeded === true ) {
    _toggleMaxLengthError();
}
```

…is used to handle the case where you've entered 50 characters, gotten the error, but then deleted one or more characters, so we need to remove the error state and message and start giving you autocomplete results again. Moving that `if` up to the top of `_handleInput` without `return false` does just that in a much simpler way than what I had before.

cfgov/unprocessed/js/molecules/Autocomplete.js
niqjohnson: Fixed in 608db2c (see https://github.com/cfpb/consumerfinance.gov/pull/6537#discussion_r702029297).

cfgov/jinja2/v1/_includes/organisms/ask-search.html
niqjohnson: Fixed in 608db2c.

cfgov/unprocessed/js/routes/on-demand/ask-autocomplete.js
niqjohnson: Fixed in 608db2c

anselmbradford (approved):

PR #6548, opened 2021-07-29
Handle filterable list results without aggregation

chosak:
#6532 added a new Elasticsearch aggregation to filterable list search results, but neglected to account for cached ES results that may not have that aggregation.

This commit adds a check so that old cached data won't cause search pages to 500. The cache will still need to be manually cleared for now to get the language choices to properly update.

As discussed with @willbarton, this is a temporary fix that can hopefully be improved with a smarter strategy overall around these caches (including perhaps getting rid of them altogether).

I've also included a setting change here to make it easier to enable Django caches locally.

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -882,3 +882,16 @@ CSRF_REQUIRED_PATHS = (
     "/admin",
     "/django-admin",
 )
+
+CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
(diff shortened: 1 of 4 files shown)


anselmbradford (approved):
The parts I understand look fine to me. Perhaps @willbarton should approve though too.

willbarton (approved):
Talked about this in chat, and @chosak had a fantastic idea for tying the cache keys to the queries that should prevent this from happening again. 

But this looks good as a solution for the language aggregations!

jslay-excella (approved):

PR #6550, opened 2021-07-29
Mega Menu: Translate button label text and move it inside the button icon spans

anselmbradford:
The mega menu has an aria-label "Main menu" on the the `nav` item, which got a Spanish translation in https://github.com/cfpb/consumerfinance.gov/commit/d8a9063920370b3eac1edd2ac8b89564334379e6

However, the button inside the mega menu to open and close the hamburger menu on mobile has content that is also named "Main menu" and it is not translated. This PR creates two button labels and places them into the icon spans so they are selectively shown and hidden (but always visually hidden) when the menu is opened and closed.

## Changes

- Mega Menu: Translate button label text and move it inside the button icon spans
 

## How to test this PR

1. Resize to mobile. Turn on VoiceOver and visit /es/ or another Spanish language page and tab into the mega menu. It should be read the Spanish language "Open the main menu". Hit return to open the menu and it should read the Spanish language "Close the main menu"


cfgov/locale/es/LC_MESSAGES/django.po
```diff
@@ -284,6 +284,16 @@ msgstr "Destacado"
 msgid "Main menu"
 msgstr "Menú principal"
 
+#: jinja2/v1/_includes/organisms/mega-menu.html:313
+#| msgid "Main menu"
+msgid "Open the main menu"
+msgstr "Abre el menú principal"
```
anselmbradford, line 290: @doguin @OrlandoSoto Is this translation correct?

cfgov/locale/es/LC_MESSAGES/django.po
```diff
@@ -284,6 +284,16 @@ msgstr "Destacado"
...
+
+#: jinja2/v1/_includes/organisms/mega-menu.html:319
+#| msgid "Main menu"
+msgid "Close the main menu"
+msgstr "Cerrar el menú principal"
```
anselmbradford, line 295: @doguin @OrlandoSoto Is this translation correct?

cfgov/locale/es/LC_MESSAGES/django.po
anselmbradford: @doguin reports this as correct.

cfgov/locale/es/LC_MESSAGES/django.po
anselmbradford: @doguin reports this as correct.

contolini (approved):
🇪🇸

PR #6556, opened 2021-08-03
Related metadata: Remove category sidebar

anselmbradford:
According to @wpears 

> that module supported an initial version of the enforcement work which has now been replaced. It can safely be removed.

Originally added in https://github.com/cfpb/consumerfinance.gov/pull/5344/files

## Removals

- Category block option from related metadata block.

## How to test this PR

1. Pull branch, run migrations, edit a browse filterable page and view the related metadata sidefoot. Topics should be an option, but Categories should no longer be.

Screen Shot 2021-08-03 at 10 00 33 AM


wpears (approved):

chosak (approved):
At @anselmbradford request I checked the DB to see if we have any pages using RelatedMetadata categories (using [this gist](https://gist.github.com/chosak/10b33897161ed8e3521e21b756d5f1b9)), and we do not.

PR #6560, opened 2021-08-04
Support testing of alternate home page designs

chosak:
This change adds the ability to use a new ~~`NEW_HOME_PAGE`~~`HOME_PAGE_2021` feature flag to toggle between the existing home page designs (both English and Spanish) to a new (placeholder) design.

This follows the same pattern used previously in #5335.

## How to test this PR

The flag will automatically be enabled on all systems except for production when `?home_page_2021=True` is appended to the URL. So, for example, to test locally, visit http://localhost:8000/?home_page_2021=True.

## Notes and todos

The new home page template included here is a copy of the current home page template, intended to be used as a starting point for development.

## Checklist


anselmbradford:
Do we care about the bewd coverage error?

chosak:
@anselmbradford no; it's complaining about the line of code inside the feature flag.

anselmbradford (approved):

PR #6564, opened 2021-08-06
Add ability to preview the 2021 home page version

chosak:
#6560 added a new stub template for a new 2021 version of the website home page. This commit adds the ability to preview this new version directly from the Wagtail editor, without having to publish it, and in the production environment where the new template is not yet enabled.

On the page edit view, the Preview button now becomes a dropdown with the ability to preview the new template:

image

This commit also includes a minor edit to the stub template to make it easier to distinguish from the current one.

## Checklist

### DIFF
--- a/cfgov/v1/jinja2/v1/home_page_2021.html
+++ b/cfgov/v1/jinja2/v1/home_page_2021.html
@@ -3,7 +3,7 @@
 {% import "organisms/card-group.html" as card_group with context %}
 
 {% block title -%}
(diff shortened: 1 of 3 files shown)


anselmbradford (approved):
This will be easy enough to strip out when the time comes?

chosak:
@anselmbradford thanks for the review; yes, just by removing this code, which will be part of any eventual feature flag cleanup once we switch.

PR #6566, opened 2021-08-06
Rural or underserved tool: Update tigerweb API version mappings

anselmbradford:
We built this tool for some prior year version of the tigerweb API, but pegged it to current. In the years since it has updated and the layer mappings don't work anymore.

## Changes

- Change API version and layer mappings from current to 2019.

## How to test this PR

1. Pull branch, `gulp clean && gulp build`, and visit http://localhost:8000/rural-or-underserved-tool/ and enter an address and see that it resolves. Compare to production where it hangs and has a console.log error.


wpears (approved):
I think a full solution would be to get this up to date with current data, but approving this as it's much better than a failed API call.

PR #6567, opened 2021-08-09
Expand try/except around Ask CFPB autocomplete

willbarton:
It seems that Elasticsearch can throw a `RequestError` due to `maxClauseCount ` at any point during the autocomplete query. We had put this `try`/`except` in place as a stop-gap until we have some better user-facing limitations and errors for the Ask CFPB autocomplete (#6537). This expands the `try` to cover both the query initialization and when we access its data.

## Checklist


jslay-excella (approved):
lgtm

PR #6569, opened 2021-08-10
Wagtail fields for 2021 home page design

chosak:
This commit adds some new Wagtail-editable fields to the HomePage model in support of a potential new home page design:

- answer_page_links: A list of Ask CFPB page links.
- highlight_cards: A set of cards with a heading, body, and link.

The existing home_page_2021.html template has been modified to display these fields when populated. This can be previewed using the existing custom preview functionality in Wagtail added in #6564.

The existing card and card-group templates have been modified to support the highlight card designs, but they need additional frontend styling.

A new URLOrRelativeURLBlock has been added to atomic elements atoms, leveraging the third-party [django-url-or-relative-url-field](https://pypi.org/project/django-url-or-relative-url-field/) package.

## Screenshots

|Edit view|Rendered view|
|-|-|
|image|image|

## Notes and todos

- We have a few other Wagtail URL blocks that are currently CharBlocks but would be better using this URLOrRelativeURLBlock, for example [`Hyperlink.url`](https://github.com/cfpb/consumerfinance.gov/blob/4edc71be614dd219c81abe7086c92773b54cb901/cfgov/v1/atomic_elements/atoms.py#L35).

## Checklist


cfgov/jinja2/v1/_includes/molecules/card.html
```diff
@@ -6,24 +6,34 @@
...
+   Creates markup for a Card molecule, which displays some text and a link,
+   with an optional heading and icon.
 
    value.text:                  Card body text.
 
    value.link_text:             Card link text.
 
    value.link_url:              Card link URL.
 
+   value.heading:               Card heading (optional).
```
anselmbradford, line 18: Other places we put `(optional)` first, but we're not consistent. See https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/jinja2/v1/_includes/molecules/text-introduction.html for example.

cfgov/jinja2/v1/_includes/molecules/card.html
```diff
@@ -6,24 +6,34 @@
...
 {# We want to make the whole card clickable. #}
 <article class="m-card m-card__featured">
     <a href="{{ value.link_url }}">
         <h2 class="m-card_heading">
-            <span class="m-card_icon">{{ svg_icon( value.icon ) }}</span>
-            <span class="u-visually-hidden">{{ value.link_text }}</span>
+            {% if value.icon -%}
+                <span class="m-card_icon">{{ svg_icon( value.icon ) }}</span>
+            {%- endif %}
+            {% if value.heading -%}
```
anselmbradford, line 32: Hmm this makes heading optional, but in the Wagtail admin it's requiring me to put a heading in for the cards?

cfgov/v1/tests/models/test_home_page.py
```diff
@@ -17,3 +17,12 @@ def test_preview_default_template(self):
 
     def test_preview_2021_template(self):
         self.check_preview_template('home_page_2021', 'v1/home_page_2021.html')
+
+    def test_highlight_card_block_rendering(self):
+        block = HighlightCardBlock()
+        value = block.to_python({
+            'heading': 'Highlight',
```
anselmbradford, line 24: Should you include a test without a heading?

cfgov/v1/models/home_page.py
```diff
@@ -33,11 +33,37 @@ class Meta:
         }
 
 
+class HighlightCardValue(blocks.StructValue):
+    @property
+    def link_text(self):
+        return 'Read more'
```
anselmbradford, line 39: How will we handle translations with this?

cfgov/jinja2/v1/_includes/molecules/card.html
chosak: The card _template_ supports cards with and without headings (the current home page vs. the new), whereas the card _block_ I've prototyped here specifically for the home page does require one.

cfgov/v1/tests/models/test_home_page.py
chosak: See above, the block requires a heading, even though the card template makes it optional.

cfgov/jinja2/v1/_includes/molecules/card.html
chosak: Thanks, I like that better. Will change.

cfgov/v1/models/home_page.py
chosak: Good call. Happily [we already translate](https://github.com/cfpb/consumerfinance.gov/blob/169420b38960b5377d546220dc9de9be34e23f30/cfgov/locale/es/LC_MESSAGES/django.po#L397-L399) this phrase, so we can use that.

chosak:
Thanks for the review @anselmbradford.

> Beyond the comments above, another additional one: Why does it say "True" below the card entry fields in Wagtail? (See your screenshot)

Oops, that was due to some Wagtail debugging  I was doing on another issue that inadvertently got left behind on my local environment. I've updated the image above.

PR #6571, opened 2021-08-10
Update Cypress@8.2.0

anselmbradford:
## Changes

- Update to Cypress@8.2.0
- Refresh npm cache

## How to test this PR

1. PR checks should pass.


docker/cypress/Dockerfile
```diff
@@ -34,7 +34,7 @@ RUN yum update -y && \
         nss* \
         alsa-lib* && \
     yum clean all && rm -rf /var/cache/yum && \
-    npm install -D cypress@8.1.0 && \
+    npm install -D cypress@8.2.0 && \
```
contolini, line 37: If we [globally install yarn](https://github.com/cfpb/consumerfinance.gov/blob/c4a515d472af59f755d7b494af5363f26709b2a3/docker/cypress/Dockerfile#L22) should we be using it to install and run cypress?

anselmbradford: Works for me https://github.com/cfpb/consumerfinance.gov/pull/6571/commits/8e64f1d15896ca51e82ea38a0cf489e1159aae10

anselmbradford: Does `$(npm bin)/cypress verify` need to be updated? I not so sure what that is doing

willbarton: In fact, we should remove the Cypress Dockerfile entirely at this point.

wpears (approved):

PR #6572, opened 2021-08-10
VideoPlayer: Add default `00:00` time

anselmbradford:
When the `<time>` element is empty [HTML Validation](https://validator.w3.org/nu/) returns `The text content of element time was not in the required format: The literal did not satisfy the time-datetime format.`

## Changes

- VideoPlayer: Add default `00:00` time

## How to test this PR

1. Pages wish videos, like http://localhost:8000/data-research/credit-card-data/terms-credit-card-plans-survey/ should be unchanged with how their timestamp appears when compared to production.


contolini (approved):
⏳

PR #6600, opened 2021-08-18
Buying a House: Remove pdf icon cruft that should be handled by middleware

anselmbradford:
## Removals

- Remove logic to present two different link markup based on whether there was an external pdf icon or not.

## How to test this PR

1. Check out the "Check out other tools and resources from Buying a House" links on https://www.consumerfinance.gov/owning-a-home/closing-disclosure/ and compare to http://localhost:8000/owning-a-home/closing-disclosure/ There should be no change beyond that they should wrap at very small screen sizes.
2. Also check https://www.consumerfinance.gov/owning-a-home/loan-estimate/ versus http://localhost:8000/owning-a-home/loan-estimate/

## Notes and todos

- https://github.com/cfpb/design-system/issues/1273 opened to address absolutely positioned icons.


chosak (approved):
🧗

PR #6601, opened 2021-08-18
Add a back-end 50-character limit to autocomplete submissions

higs4281:
Most of long autocomplete entries will be blocked by front-end code,
but if bots or mischief-makers send long autocomplete requests to our back end,
this change will happily ignore anything over 50 chars.

The constant, `AUTOCOMPLETE_MAX_CHARS`, will live in cfgov/search/models.py.  
It is used to limit the autocomplete queries sent to Elasticsearch to 50 characters,     
and gets delivered to the template and JS via jinja2.environment.globals.

## Testing

Turn on readable Elasticsearch logging by setting this env variable locally (courtesy of @chosak):
```
export ENABLE_ES_LOGGING=1
```
This lets you easily see the request that the back end sends to Elasticsearch in the runserver console.

Then send way more than 50 characters to the ask_cfpb autocomplete API:

```
http://localhost:8000/ask-cfpb/api/autocomplete/?term=You saw the masterpiece, she looks a lot like you! wrapping her left arm around your right ready to walk you through You saw the masterpiece she looks a lot like you wrapping her left arm around your right ready to walk you through You saw the masterpiece she looks a lot like you wrapping her left arm around your right ready to walk you through You saw the masterpiece she looks a lot like you wrapping her left arm around your right ready to walk you through You saw the masterpiece she looks a lot like you wrapping her left arm around your right ready to walk you through You saw the masterpiece she looks a lot like you wrapping her left arm around your right ready to walk you through
```

You should see that the query goes to Elasticsearch as

```json
{
    "autocomplete": "You saw the masterpiece, she looks a lot like you!"
}
```

### DIFF
--- a/cfgov/ask_cfpb/models/search.py
+++ b/cfgov/ask_cfpb/models/search.py
@@ -1,14 +1,13 @@
 from elasticsearch7.exceptions import RequestError
 
 from ask_cfpb.documents import AnswerPageDocument
+from search.models import AUTOCOMPLETE_MAX_CHARS
 
 
 UNSAFE_CHARACTERS = [
     '#', '%', ';', '^', '~', '`', '|',
     '<', '>', '[', ']', '{', '}', '\\'
 ]
-# maximum length of autocomplete form entries before we call a halt
-ASK_AUTOCOMPLETE_MAX = 50
 
 
 def make_safe(term):
@@ -30,7 +29,7 @@ class AnswerPageSearch:
             s = AnswerPageDocument.search().filter(
                 "term", language=self.language
             ).query(
-                'match', autocomplete=self.search_term
+                'match', autocomplete=self.search_term[:AUTOCOMPLETE_MAX_CHARS]
             )
         except RequestError:
             results = []
--- a/cfgov/ask_cfpb/models/search.py
+++ b/cfgov/ask_cfpb/models/search.py
@@ -7,6 +7,8 @@ UNSAFE_CHARACTERS = [
     '#', '%', ';', '^', '~', '`', '|',
     '<', '>', '[', ']', '{', '}', '\\'
 ]
+# maximum length of autocomplete form entries before we call a halt
+ASK_AUTOCOMPLETE_MAX = 50
 
 
 def make_safe(term):
--- a/cfgov/ask_cfpb/wagtail_hooks.py
+++ b/cfgov/ask_cfpb/wagtail_hooks.py
@@ -1,6 +1,6 @@
 from django.conf import settings
 from django.shortcuts import render
-from django.urls import reverse
+from django.urls import re_path, reverse
 from django.utils.html import format_html
 
 from wagtail.admin.menu import MenuItem
@@ -9,12 +9,6 @@ from wagtail.core.models import Page
 
 from ask_cfpb.models import Answer, AnswerPage
 from ask_cfpb.scripts import export_ask_data
-
-
-try:
-    from django.urls import re_path
-except ImportError:
-    from django.conf.urls import url as re_path
 
 
 def export_data(request):
--- a/cfgov/ask_cfpb/forms.py
+++ b/cfgov/ask_cfpb/forms.py
@@ -1,7 +1,7 @@
 from django import forms
 from django.core.validators import RegexValidator
 
-from ask_cfpb.models.search import make_safe
+from ask_cfpb.models.search import ASK_AUTOCOMPLETE_MAX, make_safe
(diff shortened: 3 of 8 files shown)


cfgov/ask_cfpb/models/search.py
```diff
@@ -7,6 +7,8 @@
     '#', '%', ';', '^', '~', '`', '|',
     '<', '>', '[', ']', '{', '}', '\\'
 ]
+# maximum length of autocomplete form entries before we call a halt
+ASK_AUTOCOMPLETE_MAX = 50
```
niqjohnson, line 11: Is there any way we can set the max length in one place and have it available everywhere we need it? For the front-end, I'm setting the max length in [cfgov/jinja2/v1/_includes/organisms/ask-search.html](https://github.com/cfpb/consumerfinance.gov/pull/6537/files#diff-b908e3e8b32fee11c18f10fa28fc449ad1b3523eafadbbf6a39258e20eb4a3ebR35). Is there one place both the model and template can pull the same max length from?

anselmbradford: Here's how we do that in find a housing counselor:
- Backend https://github.com/cfpb/consumerfinance.gov/blob/2ace1532d9b24595db6b436f008f02cc7e03ae19/cfgov/housing_counselor/jinja2/housing_counselor/index.html#L395-L403
- Frontend https://github.com/cfpb/consumerfinance.gov/blob/2ace1532d9b24595db6b436f008f02cc7e03ae19/cfgov/unprocessed/apps/find-a-housing-counselor/js/common.js#L6-L8

Other ways to do this might be to:
- Similar to above create a `<script>` tag, but give it an `id` attribute. Include a json string with the data. On the frontend query the page for the script id and then parse the internal json within the tag. (Avoids storing things in the `window`.)
- Creating a JSON API endpoint that the frontend hits via `fetch` to retrieve the settings before it initializes. (Probably the cleanest, but most difficult solution.)

higs4281: @anselmbradford 
Could we expose the limit in jinja2's environment [globals](https://jinja.palletsprojects.com/en/2.11.x/api/#jinja2.Environment.globals), so that at render time any autocomplete box on the site could use our agreed-on character limit?

anselmbradford: Sure! It would still need to get passed to the JavaScript through one of the means above though.

higs4281: I pushed an implementation of a global limit for back and front ends to this branch.

cfgov/ask_cfpb/wagtail_hooks.py
```diff
@@ -11,12 +11,6 @@
 from ask_cfpb.scripts import export_ask_data
 
 
-try:
-    from django.urls import re_path
-except ImportError:
-    from django.conf.urls import url as re_path
-
-
 def export_data(request):
```
higs4281, line 14: This deleted block is a legacy python 2 shim.

cfgov/ask_cfpb/wagtail_hooks.py
```diff
@@ -11,12 +11,6 @@
 from ask_cfpb.scripts import export_ask_data
 
 
-try:
-    from django.urls import re_path
-except ImportError:
-    from django.conf.urls import url as re_path
-
-
```
higs4281, line 19: This deleted block is a legacy python 2 shim.

cfgov/ask_cfpb/models/search.py
higs4281: > Sure! It would still need to get passed to the JavaScript through one of the means above though.

I don't so. The [render method](https://github.com/cfpb/consumerfinance.gov/blob/acb41d945c084c3988a99e8892b4cf1b081cec9f/cfgov/jinja2/v1/_includes/organisms/ask-search.html#L35) picks up the global value, which is used to create a [maxlength](https://github.com/cfpb/consumerfinance.gov/blob/acb41d945c084c3988a99e8892b4cf1b081cec9f/cfgov/jinja2/v1/_includes/organisms/ask-search.html#L61) input attribute, which is all the JavaScript needs.

anselmbradford: Oh sweet! Yea that looks great! Maybe we should do something similar with the mapbox token in HUD 🤔

niqjohnson (approved):
Amazing! I like the additional back-end protection, and the front end continues to work as before.

PR #6638, opened 2021-08-30
[TDP survey] Fixes intermittent bug in initials encoding scheme

mrclay:
The old one used vanilla Base64 encoding (with problematic chars) and a broken xor() function. We support the legacy hashes (some won't work, nothing we can do about that) but only create newer ones with a simpler encoder.

Motivation: The only requirement for initials encoding was casual obfuscation. The only reason the scores encoding is tamper-proof is that we get string signing for free.

---

## Additions

- New initials encoding scheme in TDP survey

## Changes

- Simplifies legacy encoding code and prepares it to be easily removed in the future.

## Checklist


higs4281:
FYI, I pushed this branch to staging, and the survey tests fail at the same spot: expecting encoded initials. 

You can see the errors by running cypress locally against staging. Note that when the next staging build comes along, this branch will be wiped out there, so this testing path is temporary.

mrclay:
@higs4281 I think I see the problem in the cypress test.

mrclay:
@higs4281  This will still require a re-deploy of the branch but these new tests should work on it. Old tests had encoded a full URL (which due to server key differences was never going to work across environments).

higs4281:
I didn't need to deploy to staging again, because the cypress code runs locally. The tests pass now on staging with the new encoding routine there.

higs4281:
The form for entering initials limits input to four characters. Was it a consideration that allowing four-character initials exposes the app to a universe of 4-letter words that can be rendered by students at the top of the tool's results page?

mrclay:
@higs4281 Like "DARN"!? Yes, we raised this point but were told a student would be unlikely to send profanity directly to their teacher.

higs4281:
So we're assuming students will not send a URL to friends or share them on social media?

mrclay:
@higs4281 From Hannah:
> Leslie and the content team believe that it is not an issue because students will know it's for a grade - they don't see it as different from students write on a worksheet. It WAS a consideration and we were told we were being overly cautious, but if they want us to bring that concern again, we definitely can.

higs4281:
A live federal web page that can be visited and manipulated any number of times is hardly the same as a paper worksheet in a classroom. Is anything preventing a student from sending the teacher a result, then going back and sending a fun and untraceable version to social media? 

I think something as simple as limiting initials to three characters could cut the fun factor dramatically.

mrclay:
We're meeting with Leslie next week and will raise these concerns again.

cfgov/unprocessed/apps/teachers-digital-platform/js/encode-name.js
```diff
@@ -48,54 +44,61 @@ const x = 'x';
 const y = 'y';
 const btoa = xtoy.replace( x, b ).replace( y, a );
 const atob = xtoy.replace( x, a ).replace( y, b );
+const base64Encode = wwindow[btoa];
+const base64Decode = wwindow[atob];
```
anselmbradford, line 48: Why `wwindow` instead of `window`? Is it to avoid two lookups on `window`?

mrclay: The overall original goal was to obfuscate the _minified_ source, but I'm going to clean this up if I can today.

higs4281 (approved):

PR #6643, opened 2021-09-01
FlyoutMenu: Add missing `event` parameter

anselmbradford:
https://github.com/cfpb/consumerfinance.gov/commit/621aeb1c1b2730639551de4812c2277b5f7e882b#diff-ac718465bb29fe87ef80afe5a19aa3eab8e55bb3280ceb809ee721c3b200a43cR177 added a code comment for the `event` parameter, but neglected to add it to the function itself. This leads to `ReferenceError: event is not defined.` in Firefox 56.

## Changes

- Add missing `event` parameter.

## How to test this PR

1. PR checks should pass.


chosak:
How can this be reproduced?

anselmbradford:
> How can this be reproduced?

@chosak It's this New Relic error https://onenr.io/037jb5ElBRy
So, e.g. visit the site in Firefox 60 via Sauce Labs and hover over the menu and note that there is an error in the dev console. 

We use the `event.trigger` parameter for when the trigger is clicked [here](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/js/organisms/MegaMenuMobile.js#L137).

We don't use it for the [triggerOver](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/js/modules/behavior/FlyoutMenu.js#L177) or [triggerOut](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/js/modules/behavior/FlyoutMenu.js#L195). However, I made `triggerOver`, `triggerOut`, and `triggerClick` all have the same event object signature for consistency of the API. So this `event` added here is added to the `triggerOver` event just for consistency, but it is currently functionally unused.

Alternatively, we could have the event object for `triggerClick` be:

```
{
  target: this,
  trigger: event.target,
  type: 'triggerClick'
}
```

And have the `triggerOver` and `triggerOut` be different as:

```
{
  target: this,
  type: 'triggerOver'
}
```

…and…

```
{
  target: this,
  type: 'triggerOut'
}
```

wpears (approved):

PR #6644, opened 2021-09-01
[TDP survey] Check GET for 'r' before accessing it

mrclay:
Prevents logged errors if a /view/ URL is accessed is requested without `?r=`. Also DRYs up URIs for unexpected errors.

## Changes

In TDP survey view page, check 'r' in GET before accessing.

## Checklist

### DIFF
--- a/cfgov/teachers_digital_platform/views.py
+++ b/cfgov/teachers_digital_platform/views.py
@@ -20,6 +20,7 @@ from .UrlEncoder import UrlEncoder
 
 _tdp = 'teachers_digital_platform'
 _signer = signing.Signer()
+_unexpected_error_uri = '../../../assess/survey/'
 
 
 def akamai_never_cache(view_func):
@@ -133,7 +134,7 @@ def _handle_result_url(request: HttpRequest, raw: str, code: str,
     url_encoder = UrlEncoder(AVAILABLE_SURVEYS)
     res = url_encoder.loads(code)
     if res is None:
-        return HttpResponseRedirect('../../../assess/survey/')
+        return HttpResponseRedirect(_unexpected_error_uri)
 
     survey = get_survey(res['key'])
     total = sum(res['subtotals'])
@@ -167,13 +168,13 @@ def student_results(request: HttpRequest):
         return HttpResponse(status=404)
 
     if 'resultUrl' not in request.COOKIES:
-        return HttpResponseRedirect('../../../assess/survey/')
+        return HttpResponseRedirect(_unexpected_error_uri)
 
     raw = request.COOKIES['resultUrl']
     try:
         result_url = _signer.unsign(raw)
     except signing.BadSignature:
-        return HttpResponseRedirect('../../../assess/survey/')
+        return HttpResponseRedirect(_unexpected_error_uri)
 
     return _handle_result_url(request, raw, result_url, True)
 
@@ -188,14 +189,17 @@ def view_results(request: HttpRequest):
     if request.method != 'GET':
         return HttpResponse(status=404)
 
+    if 'r' not in request.GET:
+        return HttpResponseRedirect(_unexpected_error_uri)
+
     raw = request.GET['r']
     if not isinstance(raw, str):
-        return HttpResponseRedirect('../../../assess/survey/')
+        return HttpResponseRedirect(_unexpected_error_uri)
 
     try:
         result_url = _signer.unsign(raw)
     except signing.BadSignature:
-        return HttpResponseRedirect('../../../assess/survey/')
+        return HttpResponseRedirect(_unexpected_error_uri)
 
     return _handle_result_url(request, raw, result_url, False)
 


cfgov/teachers_digital_platform/views.py
```diff
@@ -20,6 +20,7 @@
 
 _tdp = 'teachers_digital_platform'
 _signer = signing.Signer()
+_unexpected_error_uri = '../../../assess/survey/'
```
willbarton, line 23: Hard-coding URL paths in Django apps is an anti-pattern. Instead, you should be [naming your URLs and using `reverse()` to look them up by name](https://docs.djangoproject.com/en/2.2/topics/http/urls/#reverse-resolution-of-urls).

cfgov/teachers_digital_platform/views.py
```diff
@@ -188,14 +189,17 @@ def view_results(request: HttpRequest):
     if request.method != 'GET':
         return HttpResponse(status=404)
 
+    if 'r' not in request.GET:
+        return HttpResponseRedirect(_unexpected_error_uri)
```
willbarton, line 193: Any time we access a request parameter it should really be using [a Django form](https://docs.djangoproject.com/en/2.2/topics/forms/). We get this kind of required field validation for free in a paradigmatic way, along with any particular built-in sanitization we might need for a particular parameter type.

willbarton: We'll definitely need a unit test around expected inputs for `r` here, regardless.

cfgov/teachers_digital_platform/views.py
mrclay: @willbarton The path is a Wagtail page. Any suggestion for how to get a URL to that?

cfgov/teachers_digital_platform/views.py
mrclay: @willbarton We have cypress tests for it, but I can look into the form route too. Using a form would require more custom JS to build the URL because the user that completes the survey never visits the URL but needs to have it.

willbarton: A Django form doesn't have to imply anything on the front-end. It can simply handle the input validation and sanitization, and is the preferred approach to do so. An example is [where we use a form class to handle the autocomplete field in Ask CFPB](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/ask_cfpb/forms.py#L13-L17) and [check its validity in the view](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/ask_cfpb/views.py#L147-L151) but do not use it for rendering.

The other thing is that Cypress is used for functional testing. It's not how we unit test our Python, nor does it indicate coverage of our Python code.

cfgov/teachers_digital_platform/views.py
willbarton: I would certainly not recommend hard-coding Wagtail page URLs in code.

mrclay: @willbarton What I'm saying is the only reasonable option for error page is an existing wagtail page, but I don't how I'd use reverse() to get a link to that.

cfgov/teachers_digital_platform/views.py
mrclay: Yeah, definitely more python units are coming. I'll give the form a shot.

cfgov/teachers_digital_platform/views.py
willbarton: Yeah, I think we should have some prior art for discovering the URL of a particular page using its model. I think that's the correct approach — I just need some time to work up some suggested pseudocode.

mrclay: @willbarton Could you find that code? Can we instead (somehow?) hardcode the URL of the wagtail page so it can't be moved AND create a cypress test to ensure the survey links end up on that page?

willbarton: @mrclay I'm sorry, the past few days have been tremendously busy.

As to what you suggest, no, I don't think that would be any better. Late last year we did a massive reorganization of a large amount of content in the CMS, which is what makes me wary about assuming any particular URL structure as opposed to looking up the page we want here.

However, since this requires that that particular page exists, I think the best we can do without converting these views into routable pages and using their Wagtail page tree, is to use its page id to look it up. It looks like the destination page is `4301`, so here's so pseudocode to do that lookup and get the URL:

```
from v1.models import SublandingPage
destination_page = SublandingPage.objects.get(pk=4301)
destination_url = destination_page.get_url(request)
```

mrclay: @willbarton Are you sure the pk for https://www.consumerfinance.gov/consumer-tools/educator-tools/youth-financial-education/assess/survey/ is 4301?

mrclay: I've been told 15596, going with that...

mrclay:
@willbarton All redirects/links to page now use the wagtail query.

cfgov/teachers_digital_platform/views.py
```diff
@@ -12,15 +12,34 @@
...
+
+
+def _findSurveyGradeSelectionUrl(request: Optional[HttpRequest],
+                                 default='../../../assess/survey/'):
+    """
+    Get URL of the survey grade selection page from Wagtail
+    """
+    try:
+        destination_page = SublandingPage.objects.get(pk=_gradeSelectionPagePk)
+    except Page.DoesNotExist:
```
willbarton, line 38: `DoesNotExist` is actually available on all Django models, so you could catch `SublandingPage.DoesNotExist` here and you won't need the `Page` import.

cfgov/teachers_digital_platform/tests/test_forms.py
```diff
@@ -1,7 +1,33 @@
 from django import forms
+from django.core import signing
 from django.test import TestCase
 
-from ..forms import SurveyForm, markup
+from ..forms import SharedUrlForm, SurveyForm, markup
+from ..UrlEncoder import UrlEncoder
```
willbarton, line 6: We prefer absolute imports in Python:

```suggestion
from teachers_digital_platform.forms import SharedUrlForm, SurveyForm, markup
from teachers_digital_platform.UrlEncoder import UrlEncoder
```

cfgov/teachers_digital_platform/views.py
```diff
@@ -12,15 +12,34 @@
...
+from wagtail.core.models import Page
+
 from core.decorators import akamai_no_store
 
+from v1.models import SublandingPage
+
+from .forms import SharedUrlForm
 from .resultsContent import ResultsContent
 from .surveys import AVAILABLE_SURVEYS, ChoiceList, Question, get_survey
 from .UrlEncoder import UrlEncoder
```
willbarton, line 24: Same here with the imports.
```suggestion
from teachers_digital_platform.forms import SharedUrlForm
from teachers_digital_platform.resultsContent import ResultsContent
from teachers_digital_platform.surveys import AVAILABLE_SURVEYS, ChoiceList, Question, get_survey
from teachers_digital_platform.UrlEncoder import UrlEncoder
```

willbarton (commented):
A couple quick things for consistency with the rest of the codebase.

mrclay:
With initials addressed in #6668, when can this move forward?

willbarton:
Could we address the missing test coverage?

mrclay:
@willbarton We are committed to 100% python coverage and are doing that work in https://github.com/cfpb/consumerfinance.gov/pull/6618. The goal was keep this PR narrowly focused on ending the logged errors. If you'd prefer us to merge the python tests into this, we can, but then it will have to contain more small refinements to improve testability.

willbarton:
@mrclay well, our coverage check only checks to make sure the lines changed on a PR branch are fully covered when compared to `main`. The goal is to ensure that the coverage check can only fail on the changes relevant to a particular PR, and to ensure each PR is 100% covered. That ensures that none of our PRs decrease our overall test coverage.

So, for example, in this PR, it flags these lines as missing in the backend coverage:

```
cfgov/teachers_digital_platform/views.py (30.0%): Missing lines 34-37,39-40,73,104,142,178-179,181-184,186,199-202,204
```

Those are all lines that this PR changes, so it's behaving as expected.

Unfortunately it looks like there's no test coverage at all for `result-page.js` and `survey-page.js`, but what the coverage *check* is interested in are lines that are changed in this PR:

```
cfgov/unprocessed/apps/teachers-digital-platform/js/survey/result-page.js (0.0%): Missing lines 153
cfgov/unprocessed/apps/teachers-digital-platform/js/survey/survey-page.js (0.0%): Missing lines 152
```

I don't know what the right answer is for you all given the disconnect between code and tests right now, but I'm inclined to ask if you all can ensure that this PR (and beyond) pass the coverage check that ensures the PR is 100% covered by tests.

Ensuring that each PR is 100% tested is how we reduce risk of changes and make sure that any of us can troubleshoot an issue if needed even if we don't necessarily have the background context (because reading a unit test can be a good way to understand how the particular unit of code is expected to behave).

willbarton:
I will add, our functional tests are experiencing unrelated failures across multiple PRs. That's something we're working on.

PR #6646, opened 2021-09-01
Cypress folder reorganization

mistergone:
The current folder structure of cypress has confusing double-entries for folders like `page` and `components` where files had the same name, and there was no clear reason why. These "double" files represented integration tests as well as "helper" files which contained "models" and functions for the pages being tested.

While these "helpers" are no doubt useful, the naming and folder structure was confusing.

In this PR, I move the "helper" files to live alongside the integration tests they "help". This gives a lot more clarity about what each file is for, and makes it easy to find the "helper" file that goes with each integration test file.

Documentation has also been updated to explain this.

## Changes
- Remove redundantly-named page and component directories
- Move "helper" files to the integration folder
- Ignore *-helper.js files when testing
- Update docs on functional testing

## How to test this PR
1. Pull it down!
2. Run cypress tests! For instance, `yarn run cypress open`.

## Checklist


mistergone:
The functional tests that fail are not failing consistently. Also, locally, Find a Housing Counselor's functional test seems to fail because of a XHR issue (perhaps a timing issue?). Any advice on this is welcome.

flacoman91 (approved):
this looks great. and i like the restructuring.  seems pretty straightforward!

docs/functional-testing.md
```diff
@@ -1,6 +1,6 @@
 # Functional Testing with Cypress
 
-In order to ensure that as we upgrade our wagtail/django dependencies our backend python code is still functional from a frontend perspective we are integrating a new browser testing tool [Cypress](https://www.cypress.io). The goal is to implement and maintain a testing suite that enables confidence in dependency upgrades. 
+In order to ensure that as we upgrade our wagtail/django dependencies our backend python code is still functional from a frontend perspective we are integrating a new browser testing tool [Cypress](https://www.cypress.io). The goal is to implement and maintain a testing suite that enables confidence in dependency upgrades.
```
anselmbradford, line 3: I'd shorten the wordiness here to 

```
In order to ensure confidence that our frontend and backend code is still functional when we upgrade our third-party dependencies, and/or make other code changes, we have integrated a browser testing tool [Cypress](https://www.cypress.io). 
```

test/cypress/integration/consumer-tools/obtener-respuestas/obtener-respuestas-helpers.js
```diff
@@ -0,0 +1,26 @@
+export class ObtenerRespuestasBuscar {
+
+  open() {
```
anselmbradford, line 3: We could consider ditching this `open` helper completely and just calling `cy.visit(…)` where it's needed. Being one line, the helper doesn't seem to be adding much. It would only make sense for test suites that open the same URL multiple times in several places (e.g. outside of a `beforeEach`).

anselmbradford: Just a thought. Doesn't need to be removed here.

anselmbradford (approved):
Filter Blog Posts fails, but does that on `main` too and is probably an electron versus chrome version issue.
Screen Shot 2021-09-08 at 12 44 42 PM

PR #6647, opened 2021-09-02
Limit Akamai caching of sitemap.xml

chosak:
This commit adds an Edge-Control: no-cache header to the sitemap.xml view, which changes how Akamai will cache the sitemap.

Currently, the sitemap is cached like most every other page on the site: served from cache for 7 days without revalidation with the origin, then revalidated after that and re-sent from origin if necessary. This is problematic for the sitemap because it should be updated whenever a new page is published.

With this change, the Akamai-specific Edge-Control header is used, described thusly [in the Akamai docs](https://techdocs.akamai.com/start/docs/caching):

> The Edge-Control header is an Akamai-specific HTTP response header header that provides controls and parameters on content served through the edge network. On the Akamai network, Edge-Control settings settings override Cache-Control and Expires headers.

It works similarly to the standard Cache-Control header, so setting it to no-cache [should work like this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control):

> The response may be stored by any cache, even if the response is normally non-cacheable. However, the stored response MUST always go through validation with the origin server first before using it

If I understand this correctly, this means that requests to sitemap.xml will always trigger revalidation with the origin, but Akamai will return 304 Not Modified if the content is the same, which should properly use the Last-Modified header.

## How to test this PR

To verify this behavior, you can run a local server and do:

```sh
$ curl -Is http://localhost:8000/sitemap.xml | grep Edge-Control
Edge-Control: no-cache
```

## Checklist

### DIFF
--- a/cfgov/cfgov/urls.py
+++ b/cfgov/cfgov/urls.py
@@ -21,6 +21,7 @@ from wagtailautocomplete.urls.admin import (
 from ask_cfpb.views import (
     ask_autocomplete, ask_search, redirect_ask_search, view_answer
 )
+from core.decorators import add_headers
 from core.views import (
     ExternalURLNoticeView, govdelivery_subscribe, regsgov_comment
 )
@@ -443,7 +444,10 @@ urlpatterns = [
             'diversity_inclusion'),
             namespace='diversity_inclusion')),
 
-    re_path(r'^sitemap\.xml$', sitemap),
+    re_path(
+        r'^sitemap\.xml$',
+        add_headers(sitemap, {'Edge-Control': 'no-cache'})
+    ),
 
     re_path(
         r'^consumer-tools/educator-tools/youth-financial-education/',
--- a/cfgov/core/decorators.py
+++ b/cfgov/core/decorators.py
@@ -1,0 +1,22 @@
+from functools import wraps
+
+
+def add_headers(view, headers):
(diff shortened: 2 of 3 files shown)


wpears (approved):
Works as advertised!

higs4281:
Too late: I think this is worth trying to preserve some caching efficiencies, but if we really don't want sitemap requests cached, the better directive might be "no-store"

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#directives

chosak:
@higs4281 I deliberately chose no-cache over no-store; it should be okay for the requests to be cached as long as stale content doesn't get served. From further down that link, no-cache behavior means:

> Clients can cache a resource but must revalidate each time before using it. This means HTTP request occurs each time though, it can skip downloading HTTP body if the content is valid.

Here "revalidate" should mean that in practice a new sitemap.xml will be generated by the origin server, but will only be returned to the client (or the Akamai edge server) if it is different than the cached version. I believe there are two ways of determining this, one based on date (the client sends If-Modified-Since and the origin sends Last-Modified) and one based on content (the client sends If-None-Match and the origin sends ETag).

If the file hasn't changed, Akamai should send 304 Not Modified without re-sending the body of the response. This unfortunately still incurs the server load to generate the sitemap, but saves on bandwidth and is faster for the client because it can serve the file from its cache.

PR #6648, opened 2021-09-02
Consolidate decorator to prevent Akamai caching

chosak:
Followup to #6647; use no-store instead of no-cache for the cf.gov sitemap.xml file. Consolidate logic that does this with existing TDP view logic that does the same thing.

## How to test this PR

```
$ curl -Is http://localhost:8000/sitemap.xml | grep Edge-Control
Edge-Control: no-store
```

## Checklist


higs4281 (approved):
This is a great way to centralize our use of no-cache.

PR #6649, opened 2021-09-02
Removed incorrect href

brandon-rusk:
Removed href from <button> tag due to not being used and incorrect implementation. According to HTML specification an href can not exist within a `<button>`.

Related [ghe]/CFGOV/platform/issues/4156


anselmbradford:
This appears to have been added in https://github.com/cfpb/consumerfinance.gov/pull/5069/files when the link was converted to a button and it looks like @Scotchester and I overlooked removing the href. I don't see another reason this would be used.

anselmbradford (approved):

PR #6653, opened 2021-09-03
Add missing Ask search label

niqjohnson:
Adds a visually hidden text label for the Ask search input when that input has no other label. Note that the input will have a label if `show_label` is true—the `h2` or `h3` that appears above the input—hence all the ugly Jinja conditionals to prevent two labels in that case. (Is a `span` the best way to mark up the search icon when another label is there?)

Fixes GHE/CFGOV/platform/issues/4149.

---

## Additions

- Visually hidden label for the Ask search input if that input has no other text label

## Changes

- For some reason we were using the raw SVG instead of the `svg_icon` in the template tag, so I swapped that out—does that have any impacts I overlooked?

## How to test this PR

1. Run a WAVE check on https://www.consumerfinance.gov/complaint/ and note the two errors related to the Ask search input: "Missing form label" and "Empty form label". Then run a WAVE check on https://www.consumerfinance.gov/ask-cfpb/ and note the "Empty form label" error.
2. Run the same WAVE checks on this branch and note the errors related to the Ask search input are gone.
3. Check to make sure that the magnifying glass icon is marked up as a `label` and has a visually hidden text description when `show_label` is false (as on the complaint landing page) but marked up as a `span` with no text description when `show_label` is true (as on the Ask landing page or any Ask question page).

## Checklist


cfgov/jinja2/v1/_includes/organisms/ask-search.html
```diff
@@ -57,7 +57,22 @@
                                 input-contains-label
                                 {% if autocomplete %}m-autocomplete{% endif %}"
                          data-language="{{ language }}">
-                        <label for="query" class="input-contains-label_before input-contains-label_before__search"><svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 880.6 1200" class="cf-icon-svg"><path d="M860.1 879.2L645.7 664.8c90.8-136.8 76-323-44.6-443.5-137.5-137.5-360.5-137.5-498 0s-137.5 360.5 0 498c118.5 118.4 303.9 137 443.5 44.6L761 978.3c27.3 27.3 71.7 27.3 99 0s27.4-71.8.1-99.1zm-508-116.9C191.4 762 61.3 631.5 61.6 470.8c.3-160.7 130.8-290.8 291.5-290.5s290.8 130.8 290.5 291.5c-.2 118.2-71.9 224.6-181.5 269.1-34.9 14.2-72.3 21.5-110 21.4z"></path></svg></label>
+                        {% if show_label %}
+                        <span
```
anselmbradford, line 61: Hmm I wonder if we want to have open tags like this that visually look like a mistake of omitting the `>` on first glance. Maybe we should have something like this:

```
{% if show_label %}
  {% set html_tag = 'span' %}
{% else %}
  {% set html_tag = 'label for="o-search-bar_query"` %}
{% endif %}
<{{ html_tag }} class="input-contains-label_before input-contains-label_before__search">
```

niqjohnson: @anselmbradford, good call. I took a similar approach in dc6d8f1 but used a couple of inline if expressions to (a) try and keep it a little more readable and (b) be able to reuse the same variable in the closing tag. How's that looking?

cfgov/jinja2/v1/_includes/organisms/ask-search.html
```diff
@@ -57,22 +57,16 @@
...
                             {{ svg_icon('search') }}
                             {% if not show_label %}
                             <span class="u-visually-hidden">{{ _('Search') }}</span>
                             {% endif %}
-                        {% if show_label %}
-                        </span>
-                        {% else %}
-                        </label>
-                        {% endif %}
+                        </{{ label_element}}>
```
anselmbradford, line 69: 🥇

anselmbradford (approved):
This seems great! One thing I notice is that when hovering over the magnifying glass on the production pages the input loses its hover. Your span fix here fixes it for the input that uses a span, but the one that uses the label still has the issue of the hover disappearing. Do you mind opening an issue for this issue or fixing it here if it makes sense?

Screen Shot 2021-09-08 at 11 36 21 AM

Hover 
Screen Shot 2021-09-08 at 11 36 26 AM

PR #6656, opened 2021-09-07
Remove underlined space in related content links

chosak:
If a related content link gets an icon (external link or download), the icon will be incorrectly connected to the link text with an underline. This commit makes a slight change to the template markup to fix this.

|Before|After|
|-|-|
|image|image|

## How to test this PR

Run a local server and visit http://localhost:8000/data-research/, then compare the sidebar links to production at https://www.consumerfinance.gov/data-research/.

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/molecules/related-content.html
+++ b/cfgov/jinja2/v1/_includes/molecules/related-content.html
@@ -42,10 +42,11 @@
         {% for link in value.links %}
             <li class="m-list_item">
                 <a href="{{ link.url }}"
-                   class="m-list_link"
-                   {% if link.aria_label %}aria-label="{{ link.aria_label }}"{% endif %}>
-                    {{ link.text }}
-                </a>
+                  class="m-list_link"
+                  {%- if link.aria_label -%}
+                    aria-label="{{ link.aria_label }}"
+                  {% endif -%}
+                >{{ link.text }}</a>
             </li>
         {% endfor %}
         </ul>


cfgov/jinja2/v1/_includes/molecules/related-content.html
```diff
@@ -42,10 +42,11 @@ <h2 class="a-heading">
...
                 <a href="{{ link.url }}"
-                   class="m-list_link"
-                   {% if link.aria_label %}aria-label="{{ link.aria_label }}"{% endif %}>
-                    {{ link.text }}
-                </a>
+                  class="m-list_link"
+                  {%- if link.aria_label -%}
+                    aria-label="{{ link.aria_label }}"
+                  {% endif -%}
+                >{{ link.text }}</a>
```
anselmbradford, line 49: Not super keen on the floating `>` but I guess that's just as weird as putting it on the prior line
```
{% if link.aria_label %}aria-label="{{ link.aria_label }}"{% endif %}>
   {{ link.text }}</a>
```

anselmbradford (approved):

cfgov/jinja2/v1/_includes/molecules/related-content.html
brandon-rusk: Prior line or allowing it to have its own line would be more readable for me. If I ever looked back on this code, I would probably miss `>` in its current position.

PR #6658, opened 2021-09-08
Properly translate mega menu previews

chosak:
Currently, the mega menu preview view ([for example](http://localhost:8000/admin/mega_menu/menu/preview/es/)) doesn't properly render using non-English translations.

This is because even though translations are activated during the view render call itself, TemplateViews return a TemplateResponse that isn't actually rendered until after the view logic has ended. Wagtail has [some custom logic](https://github.com/wagtail/wagtail/blob/cf064f2b9990e90f5c5ca5d4405fe75a16604e0a/wagtail/admin/auth.py#L174-L191) that ensures that admin views render in the user's default language, which overrides custom languages as in the menu preview.

This commit manually renders the template view response so that it gets rendered while the proper translation is activated.

Fixes internal platform#4190.

## How to test this PR

[Create a Spanish-language mega menu](http://localhost:8000/admin/mega_menu/menu/create/) with some content, and then visit [its preview view](http://localhost:8000/admin/mega_menu/menu/preview/es/) and verify that all content is in Spanish.

## Checklist

### DIFF
--- a/cfgov/mega_menu/tests/test_model_admin.py
+++ b/cfgov/mega_menu/tests/test_model_admin.py
@@ -13,24 +13,27 @@ class ModelAdminTests(TestCase, WagtailTestUtils):
 
     @classmethod
     def setUpTestData(cls):
-        Menu.objects.create(language='en', submenus=json.dumps([
-            {
-                'type': 'submenu',
-                'value': {
-                    'columns': [
-                        {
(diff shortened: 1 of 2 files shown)


anselmbradford (approved):
Bugcrusher!! 💯 💯 💯

PR #6660, opened 2021-09-08
TDP Search: refactor custom expandable CSS

nbucknor:
On initial load, Facet block containers that are expanded by default have their text cutoff at the bottom. If you collapse and re-expand the component, the text is no longer cutoff.

This is happening because there is custom CSS overriding the padding and borders in Expandables used in the TDP search app. I've refactored the CSS so Expandables work as expected in the TDP search app.

---

## How to test this PR

1. Go to /consumer-tools/educator-tools/youth-financial-education/teach/activities/
2. Look at the "Building Block", "School Subject", and "Topics" facet blocks

### Before:
Screen Shot 2021-09-08 at 1 07 08 AM

Screen Shot 2021-09-08 at 1 07 58 AM

### After:

Screen Shot 2021-09-08 at 1 27 47 AM

Screen Shot 2021-09-08 at 1 28 10 AM

## Notes and todos

- I was able to recreate this bug in Google Chrome, but not in Safari.

## Checklist

#### Browser testing

- [ ] Firefox
- [x] Chrome
- [x] Safari
- [ ] Safari on iOS
- [ ] Chrome on Android


anselmbradford (dismissed):
If this is a bug in the Expandable, we need to fix it in https://github.com/cfpb/design-system, if TDP is not using a standard expandable, but instead is using a copy copied awhile ago, it should be updated with the latest from the design system.

nbucknor:
> If this is a bug in the Expandable, we need to fix it in https://github.com/cfpb/design-system, if TDP is not using a standard expandable, but instead is using a copy copied awhile ago, it should be updated with the latest from the design system.

@anselmbradford I looked into this further and discovered that it's only an issue in the TDP search  app because there is CSS overriding the spacing of Expandables. I've refactored the CSS to resolve the spacing issues without having to mess with the `u-is-animating` class.

anselmbradford (approved):
I see the fixes in the screenshots. 

We should revisit this at some point. Components like `o-expandable` should only have their margin set by parent components, since changes to the `o-expandable` component could easily break things here. However, the current expandables don't support the use-case here so we should look into migrating these patterns in the design system at some point.

PR #6664, opened 2021-09-08
Fixed accessibility issues with close and play buttons

brandon-rusk:
### Removals

- Removed extra `div` preventing "Close" button accessibility

### Additions

- Added tabbing accessibility outlines to both the "Close" and "Play" buttons in video player
- Overrode user agent default on "Play" button to align with CFPB standards

### How to test this PR

- Visit https://www.consumerfinance.gov/consumer-tools/reverse-mortgages/
- Tab to play button
- Press enter to play the video
- Tab to close button
- Press enter to close the video

Related [ghe]/CFGOV/platform/issues/4171


cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -148,6 +148,11 @@
         transform: translate( -50%, -50% );
     }
 
+    &:focus-visible {
+        outline: 1px dotted #0072ce !important;
+        outline-offset: 1px !important;
```
contolini, line 153: Are the `!important`s necessary? I could be wrong but I removed them locally and it seems to work fine without them.

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -308,7 +313,12 @@ button.o-video-player_play-btn,
         display: block;
         // Pushes button out of masked window when not playing.
         margin-top: @standard-btn-height__px;
+
+        &:focus {
+        outline: 1px dotted #0072ce !important;
```
contolini, line 318: For clarity swap out `#0072ce` for `@pacific`. We don't advertise it well but all our official CFPB colors have [Less variables](https://github.com/cfpb/design-system/blob/main/packages/cfpb-core/src/brand-colors.less#L69).

contolini (changes requested):
Thanks for fixing this! Just a couple minor comments.

cfgov/unprocessed/css/organisms/video-player.less
brandon-rusk: You are correct, those were leftover artifacts from the panic of nothing wanting to work haha. Fixed!

cfgov/unprocessed/css/organisms/video-player.less
brandon-rusk: Fixed!

cfgov/jinja2/v1/_includes/organisms/video-player.html
```diff
@@ -55,14 +55,12 @@
                     {{- ' o-video-player_video-container__flexible' if not is_fcm else '' }}
                     u-show-on-video-playing">
             <div class="o-video-player_video-actions-container">
-                <div class="o-video-player_video-actions">
```
anselmbradford, line 58: Looks like `o-video-player_video-actions` has styles [here](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/css/organisms/video-player.less#L205), [here](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/css/organisms/video-player.less#L234), [here](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/css/organisms/video-player.less#L272), and [here](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/css/organisms/video-player.less#L297). 

We probably could strip this div container out, but those styles would need to be migrated to `o-video-player_video-actions-container` and that class would need to be renamed because it's not a container anymore. 

I think this removal is messing with the video on non-featured content module videos. See  https://www.consumerfinance.gov/about-us/blog/new-mortgage-servicing-rule-aims-to-prevent-avoidable-foreclosures-as-protections-expire/

Looks like something is getting messed up there:

This PR
Screen Shot 2021-09-09 at 9 55 53 AM

Prod
Screen Shot 2021-09-09 at 9 56 06 AM

chosak: FYI, one hopefully convenient way to easily test this without relying on a production page is the video player template debug view, available locally when you are logged into Wagtail at http://localhost:8000/admin/template_debug/v1/video_player/.

anselmbradford: "video player template debug view" = Rick Astley royalty machine

brandon-rusk: Thanks for catching this, Ans. This is what it looks like with the PR now:

image

anselmbradford:
@brandon-rusk Check the video debug view across screen sizes. I'm seeing this one `main`:

Screen Shot 2021-09-13 at 11 41 38 AM

And this on this branch:

Screen Shot 2021-09-13 at 11 41 34 AM

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -146,6 +146,11 @@
         left: 50%;
         top: 50%;
         transform: translate( -50%, -50% );
+        
+        &:focus-visible {
```
anselmbradford, line 150: This will only apply to the centered play button. We have the `bottom-right` version also (line 156).

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -200,7 +205,7 @@ button.o-video-player_play-btn,
         position: absolute;
         top: 0;
     }
-} );
+});
```
anselmbradford, line 208: This should be reverted. The opening line for this `.respond-to-min( @bp-sm-min, {` has a space in the parentheses (and we generally pad around parens).

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -335,4 +344,4 @@ button.o-video-player_play-btn,
             opacity: 1;
         }
     }
-} );
+} );
```
anselmbradford, line 347: This should be reverted. We generally keep ending newlines.

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -295,8 +299,13 @@ button.o-video-player_play-btn,
         display: block;
         // Pushes button out of masked window when not playing.
         margin-top: @standard-btn-height__px;
+
+        &:focus {
```
anselmbradford, line 303: I don't think we need this block as the close button will pick up the standard focus rectangle of buttons now that it is unmasked.

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -146,11 +146,21 @@
...
+            outline: 1px dotted @pacific;
+            outline-offset: 1px;
+        }
     }
 
     &__bottom-right {
         right: unit( @grid_gutter-width / 2 / @play-btn-icon-size, em );
         bottom: unit( @grid_gutter-width / 2 / @play-btn-icon-size, em );
+
+        &:focus-visible {
```
anselmbradford, line 160: To ensure the focus state is kept in-sync for both button positions, it's probably best to define this in one place. To do that you'd make a block below:

```
 &__center,
 &__bottom-right {
   &:focus-visible {
     outline: 1px dotted @pacific;
     outline-offset: 1px;
   }
 }
```

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -153,6 +153,38 @@
...
+                display: block;
+            }
+
+            &:before {
+                width: 60px;
+                height: 60px;
+                left: -2px;
+                top: -2px;
+                border: 1px solid @white;
+                outline: 1px solid white;
```
anselmbradford, line 175: So this is my mistake… we should never use color keywords. Either our less variables, hex, or rgba values. So this should be `@white` instead of `white`. I sometimes use `white` in the dev console for testing and this was the result of an erroneous copy/paste.

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -153,6 +153,38 @@
         bottom: unit( @grid_gutter-width / 2 / @play-btn-icon-size, em );
     }
 
+    &__center,
+    &__bottom-right {
+        &:focus-visible {
+            outline: none;
+
+            &:before,
+            &::after {
```
anselmbradford, line 162: You have an after here with a double colon `::after`. Technically double colons is to spec, but it is not supported in older browsers. See https://css-tricks.com/to-double-colon-or-not-do-double-colon/

cfgov/unprocessed/css/organisms/video-player.less
```diff
@@ -153,6 +153,38 @@
         bottom: unit( @grid_gutter-width / 2 / @play-btn-icon-size, em );
     }
 
+    &__center,
+    &__bottom-right {
+        &:focus-visible {
```
anselmbradford, line 158: Both Safari and IE don't support `focus-visible`. Would you see any issue with putting…

```
&:focus-visible,
&:focus {
```

…here to catch IE?

(I can't figure out Safari).

brandon-rusk: I don't see there being any issue with it, let me test though

anselmbradford (approved):
🎉

contolini (approved):
:shipit:

PR #6665, opened 2021-09-10
Remove dead flag from regulations3k tests

higs4281:
This should have no effect on the tests or any other code.

### DIFF
--- a/cfgov/regulations3k/tests/test_blocks.py
+++ b/cfgov/regulations3k/tests/test_blocks.py
@@ -1,6 +1,6 @@
 import datetime
 
-from django.test import TestCase, override_settings
+from django.test import TestCase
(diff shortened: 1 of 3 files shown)


willbarton (approved):

anselmbradford:
Unrelated to this one, but just in general do we have a failsafe way to verify a flag is unused, aside from institutional knowledge?

higs4281:
> Do we have a failsafe way to verify a flag is unused, aside from institutional knowledge?

Not failsafe, but we can search the repo for the flag, and it's not used in code, and it's several years old and has been removed from settings. Often there's a PR that removed the setting that confirms it's not in use. That's the case for all the ones I've marked for deletion so far.

PR #6666, opened 2021-09-10
Remove secondary navigation cruft

anselmbradford:
The secondary navigation has a lot of years-old cruft to support non-wagtail pages that I don't think we have a use-case for anymore.

## Removals

- Removes non-wagtail secondary navigation logic from templates.

## Changes

- Move paying for college secondary navigation import to the template in which it is used.

## How to test this PR

1. Check pages that have a secondary navigation and compare to production and see that they are unchanged:
 - http://localhost:8000/paying-for-college/your-financial-path-to-graduation/
 - http://localhost:8000/consumer-tools/educator-tools/resources-youth-employment-programs/transportation-tool/
 - http://localhost:8000/start-small-save-up/start-saving/
 - http://localhost:8000/about-us/events/
 - http://localhost:8000/about-us/careers/current-openings/
 - http://localhost:8000/about-us/doing-business-with-us/
 - http://localhost:8000/about-us/the-bureau/
 - http://localhost:8000/about-us/newsroom/?categories=press-release
 - http://localhost:8000/rules-policy/regulations/1002/c/

Are there any others???


wpears (approved):
Looks right to me, dunno about other pages though

contolini (approved):
🚢

PR #6668, opened 2021-09-13
[TDP survey] Display user initials only in print media

mrclay:
Hide user-given initials except for print media.

---

## Checklist


cfgov/unprocessed/apps/teachers-digital-platform/css/pages/survey-results.less
```diff
@@ -1,3 +1,14 @@
+@media not print {
```
anselmbradford, line 1: Just a thought but should you maybe make it `display: none` by default and display it with `@media print`? I haven't looked into the support for `@media` versus `@media not`, but negating the hiding seems safer for any browser that may not support negated media queries.

mrclay: Updated.

higs4281 (commented):
The "share results" button pops a dialog asking for initials that won't be used. Shouldn't the initials entry box only appear  when you click on "Print results"?

mrclay:
@higs4281 The initials are still encoded in the shared URL to be visible to the teacher (when printed).

mrclay:
To be clear, this change still allows setting 4 arbitrary characters, but the "fun share factor" is considerably lower since the receiver won't see them unless they also print.

higs4281:
It's great that the initials aren't displayed, but why ask for initials at all for the shareable link if neither student nor teacher will see them in the linked page?

mrclay:
Presumably some teachers who want to print will appreciate having the initials on there (or to preview just in case they forget who sent it). Is this functionality still a problem?

higs4281:
Oh, i see. The teacher gets a URL that shows no initial, but if printed the initials show up.  

Sorry, I was getting confused testing locally because some of the steps have www hard-coded, and I was being taken to the live site instead of localhost:8000.

higs4281 (approved):

PR #6672, opened 2021-09-14
Update search inputs to make them consistent

anselmbradford:
After discussing with @niqjohnson we're going to make an addendum to https://github.com/cfpb/consumerfinance.gov/pull/6653/ to make the magnifying glass in the search input a label across all search inputs and not make the heading above the input a label.

## Changes

- Make magnifying glass a label in the ask search.
- Make the heading label on the ask search just a heading.
- In regs search, remove unused ending icon in search input that was once used for clearable input.
- In regs search, make the heading label on the ask search just a heading.

## How to test this PR

1. Pull branch and compare search to production on following pages:

http://localhost:8000/rules-policy/regulations/search-regulations/
http://localhost:8000/ask-cfpb/what-is-a-money-market-account-en-915/
http://localhost:8000/ask-cfpb/search/
http://localhost:8000/ask-cfpb/

## Notes and todos
 
- The change to a heading element on the regs search moves it slightly up versus using a label, but that seems fine.


cfgov/jinja2/v1/_includes/organisms/ask-search.html
```diff
@@ -43,11 +43,9 @@
...
     <form method="get" action="{{ _('/ask-cfpb/search/') }}">
         {% if show_label %}
-        <label for="o-search-bar_query">
-            <{{ 'h3' if is_subsection else 'h2' }}>
-                {{ _('Search for your question') }}
-            </{{ 'h3' if is_subsection else 'h2' }}>
-        </label>
+        <h2 {{ 'class=h3' if is_subsection else '' }}>
+            {{ _('Search for your question') }}
+        </h2>
```
niqjohnson, line 48: I think this needs to stay switching the element itself instead of the style (so `{{ 'h3' if is_subsection else 'h2' }}`). On Ask answer pages, the outline right now looks like:

```
h1: {Main heading}
    h2: Don't see what you're looking for?
        h3: Browse related questions
        h3: Search for your question

I assume we want to keep that hierarchy on answer pages.

File: cfgov/jinja2/v1/_includes/organisms/ask-search.html [hunk 2]
```diff
@@ -57,16 +55,12 @@
...
                          data-language="{{ language }}">
-                        {% set label_element = 'span' if show_label else 'label' %}
-                        {% set label_for = '' if show_label else 'for="o-search-bar_query"' %}
-                        <{{ label_element }} {{ label_for }}
+                        <label for="o-search-bar_query"
                                class="input-contains-label_before
                                       input-contains-label_before__search">
                             {{ svg_icon('search') }}
-                            {% if not show_label %}
                             <span class="u-visually-hidden">{{ _('Search') }}</span>
```
niqjohnson, line 62: Since we're getting rid of the old label, should we change this label to "Search for your question"? I'm not sure why I didn't use that as the label back in #6653, honestly. I guess we'd have to make sure the Spanish translation came through on Spanish pages, but I don't see why it wouldn't.

niqjohnson (changes requested):
I think we should make that one change to the heading levels to keep the hierarchy the same on Ask answer pages, but the content of the label is up to you, @anselmbradford (it might be nice to be "Search for your question" instead of just "Search" to differentiate it from the global search, which I wish I would have thought of when I changed it initially).

Otherwise, everything looks good to me. I still don't know the full history of how that heading-label came to be, but I think it makes sense getting rid of it and keeping all search inputs the same. Maybe @virginiacc knows more of the history?

cfgov/jinja2/v1/_includes/organisms/ask-search.html
niqjohnson: Fixed in e3e9bd9

cfgov/jinja2/v1/_includes/organisms/ask-search.html
niqjohnson: Fixed in e3e9bd9

niqjohnson (approved):
Looks great to me. WAVE shows no more missing label errors and a good outline. I like how much this simplifies the `ask-search` template, too. Nice work!

PR #6675, opened 2021-09-15
Remove babel-polyfill

anselmbradford:
Remove deprecated babel-polyfill dependency from find a housing counselor and analytics-gtm apps. This allows us to remove two package.json files as there were no other dependencies.

## Removals

- Remove deprecated babel-polyfill dependency and package.json from find a housing counselor and analytics-gtm apps. 

## How to test this PR

1. I don't actually see how this package was used, but http://localhost:8000/find-a-housing-counselor/ should be unchanged.


contolini (approved):
Cruft be gone 👍

willbarton:
This is broken on staging, where we're seeing a 500 on find-a-housing-counselor.

It looks like [we still need the package.json to exist in order for the apps to be discovered and built](https://github.com/cfpb/consumerfinance.gov/blob/main/scripts/yarn/apps-install/index.js#L22-L23).

PR #6680, opened 2021-09-16
Remove ASK_CFPB_H1 cluster-testing flag

higs4281:
This flag was used in 2018 to facilitate turning a feature on or off
in various "clusters" of pages. The cluster-testing method and the code that enables it
are documented [here](https://cfpb.github.io/consumerfinance.gov/split-testing/).

This change deletes the 2018-specific flag, ~~but leaves open the question of whether
we want to maintain a rather involved custom method for testing clusters of pages.~~ The nays have it. Thank you for your service, cluster-testing rig.


willbarton:
I wonder if it would be worth breaking off the split testing cluster flag condition into a separate library for django-flags?

higs4281:
> ... worth breaking off the split testing cluster flag condition

Maybe, if that kind of testing is used a lot.

chosak:
I would vote to remove that code and its documentation in the interest of simplifying our codebase, not to mention [AHA over DRY](https://kentcdodds.com/blog/aha-programming). It's there in history should we need to revive it (or we can bug @Scotchester).

higs4281:
I vote to remove.

willbarton:
I'm fine with removing.

higs4281:
Cluster testing code, tests and docs are gone.

willbarton (approved):

PR #6681, opened 2021-09-17
Update ccdb5-ui release for bug fix

higs4281:
Release 2.4.4 resolves a UI bug in two filter choices, ZIP codes and companies, when pagination is in play.


flacoman91 (approved):
🚢  
:shipit:

PR #6684, opened 2021-09-17
Simplify the CCDB landing page view and template

higs4281:
The CCDB legacy landing view contained a routine for ensuring that
sample complaint narratives and data were available and up-to-date,
and it handled conditions when different banner messages would be displayed. 
The process made a request to Elasticsearch just to get metadata.

This can now be simplified.

Complaint narrative samples are no longer used on the landing page and
don't need to be checked. In addition, the check for stale data never worked,
because of a bug. The view code was checking for a nonexistent [response.status](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/legacy/views/complaint.py#L48) value
instead of `response.status_code`, and it always returned blank json.

Data for the page's high-charts map and metadata for staleness were moved to S3,
where they are updated every morning. Those resources are called from the page.

All this allows us to stop making view calls to Elasticsearch entirely.
The view just needs to deliver the news when the CCDB_TECHNICAL_ISSUES flag is
activated, and that's all we need to test now.

## PR testing
- Pull in the branch and check that the CCDB landing page renders locally
at http://localhost:8000/data-research/consumer-complaints/
- in the local Wagtail admin, go to /admin/flags/ and enable the flag
` CCDB_TECHNICAL_ISSUES`
- The landing page should now show a warning dialog below the hero:

ccdb_warning


flacoman91:
how does CCDB_TECHNICAL_ISSUES get changed? is it automatic, or is someone supposed to manually set it in wagtail?

higs4281:
@flacoman91 `CCDB_TECHNICAL_ISSUES` is a flag that can be set in Wagtail if a problem is discovered.

cfgov/jinja2/v1/complaint/complaint-landing.html
```diff
@@ -68,22 +68,16 @@ <h1 class="m-hero_heading">Consumer Complaint Database</h1>
 
 {% block content_main %}
 
-<div class="m-notification m-notification__warning {% if data_down %} m-notification__visible show-data-notification {% elif narratives_down %} m-notification__visible show-narratives-notification {% endif %}">
+<div class="m-notification m-notification__warning {% if technical_issues %} m-notification__visible show-data-notification {% endif %}">
```
anselmbradford, line 71: If this is a straight-up static notification that is shown/hidden it looks like we can get rid of the `show-data-notification` class here, `data-text` class on line 76, and this CSS https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/apps/ccdb-landing-map/css/landing-page.less#L63-L80

higs4281: √

flacoman91:
Looks good. I was able to get this to work.

flacoman91 (approved):
👍

cfgov/jinja2/v1/complaint/complaint-landing.html
```diff
@@ -68,12 +68,13 @@ <h1 class="m-hero_heading">Consumer Complaint Database</h1>
 
 {% block content_main %}
 
-<div class="m-notification m-notification__warning {% if technical_issues %} m-notification__visible show-data-notification {% endif %}">
+{% if technical_issues %}
+<div class="m-notification m-notification__warning m-notification__visible ">
```
anselmbradford, line 72: TRIVIAL: could be remove the ending whitespace after the end of the classes `m-notification__visible">`

higs4281: √

Scotchester:
<obligatory drive-by this-should-be-a-Wagtail-page comment>

higs4281:
@Scotchester `<obligatory true-that>`

PR #6686, opened 2021-09-20
Remove Elasticsearch2 from Docker configs

higs4281:
Now that consumer complaints are being indexed in Elasticsearch-7, we no longer
need to support operations in the unsupported version 2.

This change removes references to Elasticsearch-2 from configurations
and stops building ES2 containers.

## Testing
To test, you'll need to take down your local cfgov Docker environment and bring it up again.

I tested by removing all containers, networks, etc with `docker system prune -af`, and starting fresh with this configuration, and it delivered a working set of services minus ES2.

For testing, you should be able to just take your local stack down and up, but be aware that you'll probably have to restore postgres and elasticsearch data after this:

```bash
docker-compose down
docker-compose up
```

Then check `docker-compse ps` to see that Elasticsearch2 is no longer running:

```bash
docker-compose ps

            Name                           Command               State                         Ports
---------------------------------------------------------------------------------------------------------------------------
cfgov-refresh_docs_1            sh /src/consumerfinance.go ...   Up      127.0.0.1:8888->8888/tcp
cfgov-refresh_elasticsearch_1   /bin/tini -- /usr/local/bi ...   Up      127.0.0.1:9200->9200/tcp, 127.0.0.1:9300->9300/tcp
cfgov-refresh_kibana_1          /bin/tini -- /usr/local/bi ...   Up      127.0.0.1:5601->5601/tcp
cfgov-refresh_postgres_1        docker-entrypoint.sh postgres    Up      127.0.0.1:5432->5432/tcp
cfgov-refresh_python_1          ./docker-entrypoint.sh pyt ...   Up      127.0.0.1:8000->8000/tcp
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,24 +4,15 @@ volumes:
     postgres_data:
 
 services:
-    elasticsearch2:
-        build:
-            context: ./
-            dockerfile: ./docker/elasticsearch/Dockerfile
-        ports:
-            - 127.0.0.1:9000:9200
-            - 127.0.0.1:9100:9300
     elasticsearch:
         build:
             context: ./
-            dockerfile: ./docker/elasticsearch/7/Dockerfile
+            dockerfile: ./docker/elasticsearch/Dockerfile
         ports:
             - 127.0.0.1:9200:9200
             - 127.0.0.1:9300:9300
         environment:
             discovery.type: single-node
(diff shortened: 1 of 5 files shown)


Scotchester:
:tada: 😌

willbarton (approved):

PR #6698, opened 2021-09-24
Wagtail 213 tableblock

wpears:
Contains integration code for draftail + handsontable to work properly.

Tables should feel fairly usable and user-friendly and be error free, check in wagtail via the block inventory.


jslay-excella:
Working through this now, we will need to patch migrations.

niqjohnson (changes requested):
I've run into a couple of issues while testing tables:

1. You can't add links or document links in a cell. It looks like what's happening is that a click in the link picker modal is registering as a click outside of the `handsontable` editor, so the editor is closing and no link is being added.
2. The edit/remove link popover is cut off when you click on an existing link.

The second issues seems solvable with a little CSS tweak to not hide overflow on the Draftail container in tables, so I'm not too worried about that. The first issue seems a lot trickier since that's core `handsontable` functionality we'll have to muck with—closing the editor when you click on things outside of the editor *except* for link picker things.

jslay-excella:
### Migrations

After much consideration, I think we will be fine with how the code currently is with migrations. This is because we aren't technically removing any models/blocks, just moving where they are located in code. This leaves a dependency on the [importing of `AtomicTableBlock` into `organisms.py`](https://github.com/cfpb/consumerfinance.gov/blob/wagtail-213-tableblock/cfgov/v1/atomic_elements/organisms.py#L22) for the migrations to continue to work, so we would just need to keep an eye out for this if [`FullWidthText`](https://github.com/cfpb/consumerfinance.gov/blob/wagtail-213-tableblock/cfgov/v1/atomic_elements/organisms.py#L561) block is ever removed.

There are also a few other places (such as [`browse_page.py`](https://github.com/cfpb/consumerfinance.gov/blob/wagtail-213-tableblock/cfgov/v1/models/browse_page.py#L38), [`enforcement_action_page.py`](https://github.com/cfpb/consumerfinance.gov/blob/wagtail-213-tableblock/cfgov/v1/models/enforcement_action_page.py#L219), few more) that import and use `AtomicTableBlock` from `organisms.py`. Not sure if this is something we want to change, or leave it as is. We will have to make flake8 ignore [this import statement](https://github.com/cfpb/consumerfinance.gov/blob/wagtail-213-tableblock/cfgov/v1/atomic_elements/organisms.py#L22) if [`FullWidthText`](https://github.com/cfpb/consumerfinance.gov/blob/wagtail-213-tableblock/cfgov/v1/atomic_elements/organisms.py#L561) block is ever removed.

Our options in terms of migrations and importing of `AtomicTableBlock`:
* Leave everything as is, with this understanding of import dependency.
* Move `AtomicTableBlock` back in to `organisms.py`
* Update migration files to point to new `AtomicTableBlock` in `tables.py` instead of `organisms.py`
AND/OR
* Update other models using `AtomicTableBlock` to import from `tables.py`
* Delete all migrations from `v1`, resetting with correct `n` for `v1`
* Delete all migrations from the project, resetting with correct `n` for each "app". 

I was also considering updating the documentation around migrations with some of this information, but I think this is too "use-case specific", and the existing documentation is satisfactory for "universal use-case".  I also do think these tables are a unique enough thing to be isolated into `tables.py`, so not entirely sure moving it back into `organisms.py` is the right move here either.

### `RichTextTableInput` Tests
I also found that there are a few issues left behind with backend unit testing. There are some issues on [`test_tables.py`](https://github.com/cfpb/consumerfinance.gov/blob/wagtail-213-tableblock/cfgov/v1/tests/atomic_elements/organisms/test_tables.py#L11), specifically around `RichTextTableInput`, that will need to be addressed. Not sure how the old json methods were being used, or if they are still needed.

jslay-excella:
Outstanding Items
* Frontend unit tests/coverage for `cfgov/unprocessed/js/admin/rich-text-table.js`
* Tests and coverage for [these lines](https://github.com/cfpb/consumerfinance.gov/blob/wagtail-213-tableblock/cfgov/v1/atomic_elements/tables.py#L32-L33), if they are still needed (I assume so). 
* Fix hyperlink input issues mentioned by @niqjohnson 
* Fix functional tests for tables.

willbarton:
One last thing we also need to do is ensure we add the fields to the Wagtail block that are only implemented client-side in the old table block implementation.

niqjohnson:
[Handsontable's `outsideClickDeselects` option](https://handsontable.com/docs/api/options/#outsideclickdeselects) seems promising for keeping the editor active when you're interacting with a link picker modal:

>If `true`, mouse click outside the grid will deselect the current selection. Can be a function that takes the click event target and returns a boolean.

cfgov/templates/wagtailadmin/css/table-block.css
```diff
@@ -1,3 +1,23 @@
...
+ max-width: 800px !important;
+}
+
+.Draftail-Editor .DraftEditor-root {
+    padding: 0.75em;
+}
+
+.handsontableInputHolder {
+  left: 0 !important;
+}
```
niqjohnson, line 13:
```suggestion
.handsontableInputHolder {
  left: 0 !important;
  margin-top: -67px;
}
```

It's a little magic number-y, but adding a negative top margin here brings the editor up so that the editor text is right over the cell text and your cursor isn't right on top of a Draftail toolbar button when the editor appears. It's a minor thing but kind of nice to not have to move the cursor too much to start typing or have an inadvertent button click. I don' feel strongly about this if we'd rather not add the complexity, though.

cfgov/templates/wagtailadmin/css/table-block.css
```diff
@@ -1,3 +1,23 @@
...
+
```
niqjohnson, line 14:
```suggestion

.handsontableInputHolder .Draftail-Editor__wrapper {
  overflow-y: visible !important;
}

```

That should fix the link editor popover from getting clipped. I'm not totally sure why `overflow-y: hidden` is set on the `.Draftail-Editor__wrapper` element in the first place, but overriding it doesn't seem to break anything as far as I could tell.

cfgov/unprocessed/js/admin/rich-text-table.js
```diff
@@ -0,0 +1,174 @@
+import { stateToHTML } from 'draft-js-export-html';
+
+// https://handsontable.com/docs/7.2.2/tutorial-cell-editor.html
+( function( Handsontable ) {
+  console.log( 'INIT 2' );
```
niqjohnson, line 5: Can we remove this `console.log`?

niqjohnson:
Do we still need [`cfgov/unprocessed/js/admin/table-block.js`](https://github.com/cfpb/consumerfinance.gov/blob/wagtail-213-tableblock/cfgov/unprocessed/js/admin/table-block.js)?

cfgov/unprocessed/js/admin/rich-text-table.js
wpears: yep!

wpears:
> Do we still need [`cfgov/unprocessed/js/admin/table-block.js`](https://github.com/cfpb/consumerfinance.gov/blob/wagtail-213-tableblock/cfgov/unprocessed/js/admin/table-block.js)?

Doesn't look like it. Removed in b509d14

willbarton:
The backend coverage failure right now is around `StructBlockValidationError`s that we can address in the `wagtail-213` branch generally, and I think they're out of scope for this PR.

PR #6699, opened 2021-09-24
Fix Spanish translation for "In this section"

chosak:
The Spanish translation for "In this section" in the secondary navigation wasn't working properly, for example on mobile on
https://www.consumerfinance.gov/es/coronavirus/asistencia-hipotecas-y-viviendas/asistencia-propietarios-de-vivienda/.

This has to do with the way the template invoked the translation code, which violated an (admittedly tricky) best practice documented at https://cfpb.github.io/consumerfinance.gov/translation/#do-not-include-the-_-call-in-a-larger-jinja2-template-data-structure.

With this change, this phrase is now properly translated.

Closes internal platform#3709 (again).

## Screenshots

|Before|After|
|-|-|
|image|image|

## Checklist


cfgov/locale/es/LC_MESSAGES/django.po
```diff
@@ -1695,9 +1699,6 @@ msgstr "Jubilado"
 #~ msgid " of "
 #~ msgstr " de "
 
-#~ msgid "In this section"
-#~ msgstr "En esta sección"
```
anselmbradford, line 1699: Weird. We had this in prior, but disabled it??

anselmbradford: Oh wait, I see your post comment. It was in there, but not working right.

cfgov/jinja2/v1/_includes/organisms/secondary-navigation.html
```diff
@@ -15,8 +15,9 @@
 <nav class="o-secondary-navigation
             {{ '' if has_children else 'o-secondary-navigation__no-children' }}"
      aria-label="{{ _('Section navigation') }}">
+    {% set sec_nav_settings_label = _("In this section") %}
```
anselmbradford, line 18: If it matters, we had single quotes before and this is double quotes. Not sure what we usually use for these.

anselmbradford (approved):

cfgov/jinja2/v1/_includes/organisms/secondary-navigation.html
chosak: I don't know that it matters; I copied [the example we have documented](https://cfpb.github.io/consumerfinance.gov/translation/#do-not-include-the-_-call-in-a-larger-jinja2-template-data-structure).

cfgov/locale/es/LC_MESSAGES/django.po
chosak: Right. According to [this comment](https://github.com/cfpb/consumerfinance.gov/pull/5956#issuecomment-682835663) back when `validation-translations` went in, "[r]emoving all translations of a string from code" should cause that job to fail, but unfortunately this change happened before we had that check.

PR #6703, opened 2021-09-27
Fix hidden elements being tab-able

brandon-rusk:
## Additions

-Added `aria-hidden="true"` option to multiple elements to prevent screen readers from tabbing to them when they were hidden

## Test this PR

-Go to /about-us/racial-equity/
-Turn on Voice Over (CMD + FN + F5)
-Use Voice Over tabbing (CTRL + OPTION + RIGHT/LEFT ARROWS) to tab through video elements

Related [ghe]/CFGOV/platform/issues/4181


anselmbradford:
Remind me how this was working before? Like if I voiceover through https://www.consumerfinance.gov/about-us/racial-equity/ it doesn't seem like it's going into the hidden elements. Is this in iOS only thing or something?

brandon-rusk:
> Remind me how this was working before? Like if I voiceover through https://www.consumerfinance.gov/about-us/racial-equity/ it doesn't seem like it's going into the hidden elements. Is this in iOS only thing or something?

So you actually have to do a weird voice over specific tabbing with these hotkeys: CTRL + OPTION + Right/Left arrows.
When you do that you will see it tabs to the hidden elements. I just tested on your link to confirm and it does.

anselmbradford (approved):
Works as described. The play video button is read when the video is playing, but that's a separate issue.

PR #6706, opened 2021-09-28
 Update ES docker image to  tag 7.14.1 

lfatty:
Update ES docker image tag to `7.14.1` which has fewer high CVEs than the current `7.10.1`.


willbarton (approved):

willbarton:
I think this version *should* match https://github.com/cfpb/consumerfinance.gov/blob/main/docker/elasticsearch/Dockerfile#L1 as well (it doesn't currently, though).

lfatty:
This PR can be removed because we are going to use the latest `ES:7.16.1`

PR #6712, opened 2021-09-29
Check for webpack-config when building, not package.json

anselmbradford:
## Changes

- Check for webpack-config when building, not package.json
- Remove app package.json files that specify no dependencies.
- Remove unused erap webpack-config file.

## How to test this PR

1. Hmm PR checks should pass I think would do it?


willbarton:
> Hmm PR checks should pass I think would do it?

It's interesting... I *think* vallidate-assets should've failed on #6675, but it didn't. I'm looking into why. But I can confirm that on `main` after building `cfgov/static_built/apps/find-a-housing-counselor/js/common.js` does not exist, but on this branch it does.

willbarton (approved):

PR #6714, opened 2021-09-30
Add dimensions to logo

niqjohnson:
We don't set an explicit width and height on the logo image in the global header, which Lighthouse flags as possibly contributing to cumulative layout shift (CLS). Even though our CLS scores are really good, we wanted to add `width` and `height` attributes just in case.

---

## Additions

- `width` and `height` attributes to the logo image

## How to test this PR

1. Fire up any page in Chrome and run a Lighthouse report in dev tools
2. Confirm in the report that the logo image does not appear in the "image elements do not have explicit `width` and `height` diagnostic (if you run the report on a page with no other images, as in the screenshots below, that diagnostic shouldn't appear at all)
3. Confirm that the logo still looks OK at all screen sizes

## Screenshots

Lighthouse report on `main` | Lighthouse report on this branch
---- | ----
lighthouse-before | lighthouse-after
 
## Checklist


anselmbradford:
Any idea why @Scotchester removed these values in https://github.com/cfpb/consumerfinance.gov/pull/3702/files#diff-5919c52d002f7bba7b4ae5f72f9b8bb9544ebe682501a07ddd80b9dc785ce9d4L77-L78 ?

niqjohnson:
> Any idea why @Scotchester removed these values in https://github.com/cfpb/consumerfinance.gov/pull/3702/files#diff-5919c52d002f7bba7b4ae5f72f9b8bb9544ebe682501a07ddd80b9dc785ce9d4L77-L78 ?

@anselmbradford, that's a good question. I assume it's because we also moved to `srcset` in that commit, and it's not obvious how `width` and `height` attributes work with `srcset`. According to the [documentation linked to in Lighthouse reports](https://web.dev/optimize-cls/#images-without-dimensions), though, it sounds like they play fine together as long as all the images specified in `srcset` have the same aspect ratio:

>When working with responsive images, `srcset` defines the images you allow the browser to select between and what size each image is. To ensure `` width and height attributes can be set, each image should use the same aspect ratio.

Scotchester:
I think mainly I just didn't think they were necessary, and I hate defining things in HTML that are redundant with/overridden by CSS :)

Definitely wasn't considering layout shift at the time, though.

wpears (approved):

PR #6718, opened 2021-10-05
Add django-cors-headers and env var for origins

willbarton:
This PR adds the [django-cors-headers](https://pypi.org/project/django-cors-headers/) package to automatically add Cross-Origin Resource Sharing (CORS) headers via settings and middleware to allow origins specified in the `CORS_ALLOWED_ORIGINS` environment variable.

The intended use of this is to permit GovDelivery subscriptions from the [HMDA pages](https://ffiec.cfpb.gov/).

## How to test this PR

To test the CORS Access-Control-Allow-Origin header, set the `CORS_ALLOWED_ORIGINS` environment variable to a valid JSON array of origins:

```bash
export CORS_ALLOWED_ORIGINS=[\"https://example.com\"]
./runserver.sh
```

Then send the `Origin` header in a test request with one of those origins:

```bash
curl -I -H "Origin: https://example.com" http://localhost:8000
```

And observe the `Access-Control-Allow-Origin` header in the response with that origin:

```http
Access-Control-Allow-Origin: https://example.com
```

## Notes and todos

- To use this in deployments, this will require adding the `CORS_ALLOWED_ORIGINS` variable to our Ansible config.

## Checklist

### DIFF
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -1,8 +1,10 @@
+import json
 import os
 import secrets
 from pathlib import Path
 
 from django.conf import global_settings
+from django.core.exceptions import ImproperlyConfigured
 from django.utils.translation import ugettext_lazy as _
 
 import dj_database_url
@@ -99,6 +101,7 @@ INSTALLED_APPS = (
     "teachers_digital_platform",
     "wagtailmedia",
     "django_elasticsearch_dsl",
+    "corsheaders",
 
     # Satellites
     "ccdb5_ui",
@@ -139,6 +142,7 @@ WAGTAILSEARCH_BACKENDS = {
 MIDDLEWARE = (
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.http.ConditionalGetMiddleware",
+    "corsheaders.middleware.CorsMiddleware",
     "django.middleware.common.CommonMiddleware",
     "core.middleware.PathBasedCsrfViewMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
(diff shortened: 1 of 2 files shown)


cfgov/cfgov/settings/base.py
```diff
@@ -794,3 +798,15 @@
...
+
+# Set our CORS allowed origins based on a JSON list in the
+# CORS_ALLOWED_ORIGINS environment variable.
+try:
+    CORS_ALLOWED_ORIGINS = json.loads(
+        os.environ.get("CORS_ALLOWED_ORIGINS", "[]")
+    )
+except (TypeError, ValueError):
+    raise ImproperlyConfigured(
+        "Environment variable CORS_ALLOWED_ORIGINS is either not defined or "
```
chosak, line 810: > either not defined

If the variable isn't defined, this exception won't happen, will it, since the value will default to `"[]"`?

willbarton: Correct. The `ImproperlyConfigured` error should only be raised if the variable is defined but not valid JSON.

willbarton: Right. I see. I'll fix the message.

cfgov/cfgov/settings/base.py
```diff
@@ -794,3 +798,15 @@
...
+        "is not valid JSON. Expected a JSON array of allowed origins."
```
chosak, line 811:
```suggestion
        "Environment variable CORS_ALLOWED_ORIGINS is not valid JSON. "
        "Expected a JSON array of allowed origins."
```

jslay-excella (approved):
lgtm, and yes, i looked at it ;)

cfgov/cfgov/settings/base.py
```diff
@@ -794,3 +798,15 @@
         'TIMEOUT': None,
     }
 }
+
+# Set our CORS allowed origins based on a JSON list in the
+# CORS_ALLOWED_ORIGINS environment variable.
+try:
+    CORS_ALLOWED_ORIGINS = json.loads(
+        os.environ.get("CORS_ALLOWED_ORIGINS", "[]")
```
jslay-excella, line 806: could also use `os.getenv` but i think it essentially does the exact same thing.

willbarton: Yeah. We use `os.environ.get()` for most all of this.

PR #6719, opened 2021-10-05
Mega Menu and Global search: Handle tabbing out with focusout event

anselmbradford:
A longstanding issue is we create a hidden button after the mega menu and global search that when tabbed to would close either component. This hidden button is flagged as an accessibility issue.

## Changes

- Uses `focusout` event instead of introducing a hidden button for closing the mega menu and global search on tabbing out of each component.

## How to test this PR

2. Run a local server and visit any page.
3. Try the menu and global search at different screen sizes and in different browsers. You should be able to expand either and then tab through them and then when tabbed out of the open search or menu they should close.

### DIFF
--- a/cfgov/unprocessed/js/organisms/MegaMenuDesktop.js
+++ b/cfgov/unprocessed/js/organisms/MegaMenuDesktop.js
@@ -151,6 +151,17 @@ function MegaMenuDesktop( baseClass, menus ) {
   }
 
   /**
+   * Close the mega menu.
+   * @returns {MegaMenuDesktop} A instance.
+   */
+   function collapse() {
+     // Close the menu.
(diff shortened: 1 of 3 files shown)


cfgov/unprocessed/js/organisms/MegaMenuDesktop.js
```diff
@@ -150,6 +150,17 @@ function MegaMenuDesktop( baseClass, menus ) {
     }
   }
 
+  /**
+   * Close the mega menu.
+   * @returns {MegaMenuDesktop} A instance.
```
chosak, line 155:
```suggestion
   * @returns {MegaMenuDesktop} An instance.
```

PR #6730, opened 2021-10-18
Adding caching max age to our java script files 

baruva:
Using apache configuration to add a max age cache rule to all js, CSS, and image files served by the app itself. This will help improve our Webapp test scores, as the current lack of rules  are causing failures in multiple pages. For instance event-listener-scroll-tracking.js is prevalent in a lot of low cache scores sites, and with this new rule it will no longer be a failure.

## Additions

- There is one addition to the httpd.conf file in the apache configurations, its a rule that will be applied to all files that end in js, css, jpg, png, gif , ico, or sw. This should cover just about any extra catchable file we missed with the previous apache  rule we use to establish cache via Edge Control header [found here](https://github.com/cfpb/consumerfinance.gov/blob/7af2b34a38545bad1ba7e7d686ed7d01e65f932e/cfgov/apache/conf.d/headers.conf#L27-L31)

## How to test this PR

1.Open up DEV5 in google chrome
2. Open up Dev tools and go to the Network tab. From there you can filter to JS files and search for js files that have failed webpage tests before like tab-navigation-tracking.js and you will seee in its Response Headers max-age=31536000, public.

## Checklist

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -37,22 +37,35 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2.3.4
+      - uses: dorny/paths-filter@v2.10.2
+        id: filter
+        with:
+          filters: |
+            backend:
+              - '**.py'
+              - 'requirements/**.txt'
+              - '**.html'
+              - '**.mo'
+              - '**.po'
 
       - name: Install gettext for translation validation
-        if: matrix.toxenv == 'validate-translations'
+        if: steps.filter.outputs.backend == 'true' && matrix.toxenv == 'validate-translations'
         run: sudo apt-get install gettext
 
       - name: Set up Python
         uses: actions/setup-python@v2.2.2
+        if: steps.filter.outputs.backend == 'true'
         with:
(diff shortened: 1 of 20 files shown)


anselmbradford:
> @anselmbradford, it looks like you worked on most of those tag manager scripts—do you remember if not adding cache-busting to those was a deliberate decision?

No I don't see any reason those should not be cached.

willbarton:
> I think @willbarton mentioned something about that being because those files aren't in Django's static files manifest

Correct... as I mentioned in the issue, these do not appear to be caught by [ManifestStaticFilesStorage](https://docs.djangoproject.com/en/2.2/ref/contrib/staticfiles/#django.contrib.staticfiles.storage.ManifestStaticFilesStorage), which is what adds that unique cache key. I'm very curious about why that might be, but I think your suggestion below might be better — don't try to match the cache key at all.

> I guess that wouldn't help for images served from S3 via Cloudfront, which @baruva was hypothesizing might pick up the max-age if we set it in Apache. 

This won't help, because Cloudfront and S3 have no interaction with Apache.

PR #6747, opened 2021-11-04
Grab the node rpm package directly, prevently mirror mismatch trouble

wpears:
Fixes the weird `[Errno -1] Package does not match intended download.` issue on build by grabbing the package directly beforehand.


higs4281 (approved):
Would be nice to know what's going on with mirror sources, but this works!

PR #6748, opened 2021-11-05
Simplify Elasticsearch environment variables

chosak:
This commit tries to simplify how we configure Elasticsearch via environment variables, for virtualenv/local Docker/deployed Docker.

Currently, as of #6695, Django will use the combination of ES7_HOST and ES_PORT to configure the Elasticsearch client. ES_HOST is no longer used anywhere, since our migration to Elasticsearch 7.x. In Django settings, these default to localhost/9200 if the environment variables are not defined.

This PR changes this so we standardize everywhere to ES_HOST and ES_PORT.

With this change, the following configurations are still supported:

1. Local virtualenv - these settings are now no longer set in .env, but default to localhost/9200 for a locally-running Elasticsearch. The definitions have been left in .env_SAMPLE commented out for documentation purposes.

2. Local Docker - port defaults to 9200, but ES_HOST is set in docker-compose.yml to "elasticsearch" instead.

3. Deployed Docker stack - port defaults to 9200, but ES_HOST is set in docker-stack.yml to "elasticsearch" instead.

## Notes and todos

- If/when this is merged, there is some cleanup to be done in our internal Ansible deployment scripts to remove reference to the now-deprecated ES7_HOST environment variable.
- (See below) We also need to modify [the ccdb5-api repo](https://github.com/cfpb/ccdb5-api/blob/0310e5245bc28750b892cea096f881f9f3fb38e8/complaint_search/es_interface.py#L39) to update usage of ES7_HOST to ES_HOST.

## Checklist

### DIFF
--- a/.env_SAMPLE
+++ b/.env_SAMPLE
@@ -88,9 +88,8 @@ export NODE_ENV=development
 # Elasticsearch - for indexing and storing searchable content.
 ##############################################################
 
-export ES_PORT=9200
-export ES_HOST=elasticsearch
-export ES7_HOST=elasticsearch
+#export ES7_HOST=localhost
+#export ES_PORT=9200
 
 ##########################################################
 # Complaint Search Elasticsearch (ES) Indexes Information.
@@ -190,6 +189,4 @@ export WATCHMAN_TOKENS=''
 if [ $(uname -s) = "Darwin" ]; then
   export CFGOV_HOSTNAME=localhost:8000
   export DATABASE_URL=postgres://cfpb@localhost/cfgov
-  export ES_HOST=localhost
-  export ES7_HOST=localhost
 fi
--- a/.env_SAMPLE
+++ b/.env_SAMPLE
@@ -88,7 +88,7 @@ export NODE_ENV=development
 # Elasticsearch - for indexing and storing searchable content.
 ##############################################################
 
-#export ES7_HOST=localhost
+#export ES_HOST=localhost
 #export ES_PORT=9200
 
 ##########################################################
--- a/cfgov/cfgov/settings/base.py
+++ b/cfgov/cfgov/settings/base.py
@@ -315,7 +315,7 @@ HOUSING_COUNSELOR_S3_PATH_TEMPLATE = (
 )
 
 # ElasticSearch 7 Configuration
-ES7_HOST = os.getenv('ES7_HOST', 'localhost')
+ES_HOST = os.getenv('ES_HOST', 'localhost')
 ES_PORT = os.getenv("ES_PORT", "9200")
 ELASTICSEARCH_BIGINT = 50000
 ELASTICSEARCH_DEFAULT_ANALYZER = "snowball"
@@ -329,7 +329,7 @@ if os.environ.get('USE_AWS_ES', False):
     )
     ELASTICSEARCH_DSL = {
         'default': {
-            'hosts': [{'host': ES7_HOST, 'port': 443}],
+            'hosts': [{'host': ES_HOST, 'port': 443}],
             'http_auth': awsauth,
             'use_ssl': True,
             'connection_class': RequestsHttpConnection,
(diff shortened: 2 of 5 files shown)


higs4281 (commented):
The CCDB repos still use `ES_HOST` for local operations
- [ccdb-data-pipeline](https://github.com/cfpb/ccdb-data-pipeline/blob/main/INSTALL.md#option-2-a-docker-only-setup), which I can adjust.
- [ccdb5-api](https://github.com/cfpb/ccdb5-api/blob/0310e5245bc28750b892cea096f881f9f3fb38e8/complaint_search/es_interface.py#L29-L30), which will default properly

It might have made sense to use `ES_HOST` everywhere now that 7 is our only option, but maybe it's simplest to stick with ES7 than track them all down.

chosak:
@higs4281 thanks, nice catch.

> ccdb-data-pipeline, which I can adjust.

Technically this doesn't have to be consistent with this repo, does it? Although...

> ccdb5-api, which will default properly

Hm, but this won't default properly if using the Docker setup, after my changes, if we stop setting ES_HOST.

> It might have made sense to use ES_HOST everywhere now that 7 is our only option, but maybe it's simplest to stick with ES7 than track them all down.

That was my thinking as well, but maybe the better thing to do is instead deprecate our use of ES7_HOST in cf.gov and instead use ES_HOST. This will avoid having to modify CCDB code, although it would require some Ansible changes. What do you think of that approach vs. this one?

higs4281:
@chosak 
The CCDB changes to use ES7_HOST are trivial and only affect local operations, which are somewhat rarely used. This doesn't affect anything in prod ccdb indexing, and prod ccdb searching already [uses ES7_HOST](https://github.com/cfpb/ccdb5-api/blob/0310e5245bc28750b892cea096f881f9f3fb38e8/complaint_search/es_interface.py#L39).

Given how fond of breaking changes ES seems to be, I don't think it's a big sin to have a version-specific var. The next time we upgrade, we'll probably need two settings again to get Django and CCDB converted in tandem, and having the old tagged to ES7 might help with clarity.

I'd vote to proceed with ES7_HOST.

higs4281:
If we want to go with the pristine `ES_HOST`, we'd probably need this sequence:
- Make sure `ES_HOST` is set in all environments
- Change the CCDB repos to use `ES_HOST` only.
- After those changes clear, remove `ES7_HOST` from cfgov code, docker configs, jenkins-automation jobs and ansible.

chosak:
I ran into another issue with the [prod-like Docker image](https://cfpb.github.io/consumerfinance.gov/running-docker/#production-like-docker-image): the suggested command doesn't load docker-compose.override.yml and thus _doesn't read in the .env file_ (normally done [this way](https://github.com/cfpb/consumerfinance.gov/blob/7aec0c9751f0f621696b631555fbad15955bfb06/docker-compose.override.yml#L12)), which means that the only environment variables passed to the Apache container are the ones [explicitly defined here](https://github.com/cfpb/consumerfinance.gov/blob/7aec0c9751f0f621696b631555fbad15955bfb06/docker-compose.prod.yml#L12-L22). So, the container doesn't get a value for ES7_HOST right now, which means that any Elasticsearch interactions fail.

Currently we define [ES_HOST in docker-compose.yml](https://github.com/cfpb/consumerfinance.gov/blob/412e17a2a86c9fc1077e4ce88650f65f3f69fa54/docker-compose.yml#L46), but rely on the .env behavior to get ES7_HOST in there, inconsistently, and unnecessarily, it seems to me.

> If we want to go with the pristine ES_HOST
> Make sure ES_HOST is set in all environments

It looks to me from inspecting the Ansible code that we're already passing the same URL for both ES_HOST and ES7_HOST, so I think we're good there.

> Change the CCDB repos to use ES_HOST only.

I guess that's [this line](https://github.com/cfpb/ccdb5-api/blob/0310e5245bc28750b892cea096f881f9f3fb38e8/complaint_search/es_interface.py#L39)? Is there any other place or documentation that would need to be updated if we made this change?

higs4281:
> I guess that's [this line](https://github.com/cfpb/ccdb5-api/blob/0310e5245bc28750b892cea096f881f9f3fb38e8/complaint_search/es_interface.py#L39)? Is there any other place or documentation that would need to be updated if we made this change?

"this line" is fixed in [ccdb5-api](https://github.com/cfpb/ccdb5-api/pull/182)

That version of ccdb5-api is added to requirements in [PR 6753](https://github.com/cfpb/consumerfinance.gov/pull/6753)

higs4281 (approved):
I think we're good.

PR #6751, opened 2021-11-08
Add New Relic Browser FedRAMP endpoint to CSP

chosak:
Per the New Relic Browser documentation for FedRAMP endpoints, we need to include gov-bam.nr-data.net in our list of allowed CSP hostnames:

https://docs.newrelic.com/docs/browser/new-relic-browser/getting-started/compatibility-requirements-browser-monitoring/

Currently, every page on the site (for example the homepage at https://www.consumerfinance.gov/) generates an error in the browser console like:

> nr-1211.min.js:1 Refused to load the script 'https://gov-bam.nr-data.net/...' because it violates the following Content Security Policy directive: "script-src ..."

This commit fixes that error.

## Notes and todos

I notice that our NR settings currently exclude AJAX request tracking for this domain:

image

Is this something we've manually added at some point to try to eliminate these errors? @anselmbradford 

## Checklist


anselmbradford:
> Is this something we've manually added at some point to try to eliminate these errors?

Hmmm, this was not a setting I set. Any harm in changing it, at least to experiment?

chosak:
@anselmbradford thanks, I wouldn't think so.

contolini (approved):

PR #6756, opened 2021-11-09
Simplify use of DATABASE_URL

chosak:
This commit removes DATABASE_URL from our default .env file in favor of more explicit definition in Django settings and docker-compose files.

This change is intended to remove a piece of complexity from the .env, specifically the "if" clause that causes so much trouble with certain versions of docker-compose. (After this commit there will still be one piece of this clause remaining, but I'd like to remove that in a subsequent change.)

After this change, developers will no longer need to set DATABASE_URL in their local environments, and cf.gov should work properly regardless of how it is being run.

- When running locally in virtualenv: by default, database is set via settings.base to our recommended Postgres setup (local, DB name cfgov, user cfpb, password cfpb). This can be overridden using the DATABASE_URL environment variable, as documented in the settings file and in .env_SAMPLE.

- When running locally with Docker: the docker-compose.yml file now explicitly sets DATABASE_URL to use our Postgres container settings (postgres://cfpb:cfpb@postgres/cfgov).

- When running locally with the prod-like Docker container: we no longer need to pass in database credentials in docker-compose.prod.yml since we're now including them in docker-compose.yml.

- When running tox locally in a virtualenv: tox no longer needs to set a fallback DATABASE_URL because it can use the same logic as the regular Django server. I've simplified the docs around this.

- When running tox inside the Python Docker container: tox now always works when run within the container because DATABASE_URL is now set explictly, and won't potentially get overwritten from your local .env.

- When we run tox on GitHub Actions: we explicitly set DATABASE_URL to the correct configuration for how Postgres is run on GH.

I believe that this covers all cases. I removed a few explicit references to the PGPASSWORD variable, as these would only be needed if the password wasn't set in the DATABASE_URL, which is now no longer the case after this change.

(This also reverts #6755 -- sorry @schbetsy! -- as I believe/hope it should no longer be needed after this set of changes.)

## How to test this PR

Try running cf.gov and tox in whatever configuration you prefer - hopefully it will continue to function.

## Checklist


cfgov/cfgov/settings/base.py
```diff
@@ -230,12 +230,15 @@
 if ALLOW_ADMIN_URL:
     DATA_UPLOAD_MAX_NUMBER_FIELDS = 2000  # For heavy Wagtail pages
 
-# Databases
-DATABASES = {}
-
-# If DATABASE_URL is defined in the environment, use it to set the Django DB
-if os.getenv("DATABASE_URL"):
-    DATABASES["default"] = dj_database_url.config()
```
chosak, line 238: Prior to this change, it was at least theoretically possible to pass this code without having any databases defined. I can't think of any cases where we actually depend on this logic, though.

docs/python-unit-tests.md
```diff
@@ -94,8 +94,7 @@ These default environments are:
   versions of Python, Django, and Wagtail. We require this environment to
   pass in CI.
 
-Tests will run against the default Django database as configured in settings
-using the `DATABASE_URL` environment variable.
+Tests will run against the default Django database.
```
willbarton, line 97: Does this mean one needs to define `DATABASE_URL` in the environment to run `tox` against PostgreSQL running in Docker?

chosak: Hm, good catch! To clarify, this means a mix of virtualenv for Python code and Postgres as a Docker container? In that case, wouldn't you already have to have manually set DATABASE_URL to something in your .env? If so, tox should already use that.

willbarton: > To clarify, this means a mix of virtualenv for Python code and Postgres as a Docker container? 

Correct.

> In that case, wouldn't you already have to have manually set DATABASE_URL to something in your .env? If so, tox should already use that.

Well, part of this PR's change gets rid of tox configuration that defaults to the containerized Postgres:
https://github.com/cfpb/consumerfinance.gov/pull/6756/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449L84

I wonder though... it feels weird to remove that line from tox and make it rely on the env var when we're switching to explicitly defining the URL in a bunch of other places.

willbarton: The other thing I notice is I said "defaults to containerized Postgres", but really, it just defaults to a Postgres that's running locally with a password.

Should we standardize on password/nopassword for both local Postgres and Docker Postgres? Then it won't matter which it is for tox, the same URL will work for either from localhost and it can use the Django settings default.

chosak: > Should we standardize on password/nopassword for both local Postgres and Docker Postgres?

I may be misremembering, but I seem to recall that we had to set a password for the Docker Postgres. I don't feel strongly about this, but it's convenient not to use a password locally.

> the same URL will work for either from localhost and it can use the Django settings default

I don't exactly follow. By "either" here do you mean, Postgres running locally on your Mac vs. Postgres running in a container? if so, I don't see how the same URL could work.

willbarton: Right now the only difference between the URL settings default to in this PR and the URL I use to make tox use Docker Postgres is the password.

To use Docker Postgres and Elasticsearch from tox, here's the bottom of my .env:

```
if [ $(uname -s) = "Darwin" ]; then
  export DATABASE_URL=postgres://cfpb:${PGPASSWORD}@localhost/cfgov
  export ES_HOST=localhost
  export ES7_HOST=localhost
fi
```

willbarton: I've confirmed that if I remove the `POSTGRES_PASSWORD` variable from the `postgres` container in `docker-compose.yml`, and add `POSTGRES_HOST_AUTH_METHOD: trust`, I can use the default `postgres://cfpb@localhost/cfgov` with tox without setting a `DATABASE_URL` in the environment at all.

willbarton: So, this is the change that works for me:

```diff
    postgres:
        …
        environment:
            POSTGRES_USER: cfpb
            POSTGRES_PASSWORD: cfpb
            POSTGRES_DB: cfgov
+           POSTGRES_HOST_AUTH_METHOD: trust
        …
```

Then I can run tox from outside of Docker and it will use the default database URL you have from settings and use the `postgres` Docker container, all without a password. This also means you could run cf.gov locally in a venv and use the `postgres` Docker container too!

willbarton: I do wonder though if this would mean we'll need to define the `postgres` container in `docker-compose.prod.yml` without that env var.

chosak: That's a good point @willbarton. After some further thought, I think it might be better to standardize on using the `cfpb` password everywhere. I will add a commit that proposes that, and also updates the docs.

chosak:
@willbarton et al: per the discussion above, I've updated the default configuration to use the 'cfpb' password everywhere when working locally, which should support all combinations of tox/virtualenv/Docker without ever needing to set DATABASE_URL, unless you have some other reason to do so. I've updated [the relevant docs](https://cfpb.github.io/consumerfinance.gov/installation/#install-postgres):

image

FWIW, the default Homebrew installation of Postgres enables "trust" authentication for local connections, which means that you can login to any account on localhost with or without its password (or with a wrong password). This means that developers that have an existing `cfpb` user without a password shouldn't run into any issues locally even if the code starts trying to pass one.

willbarton (approved):

PR #6763, opened 2021-11-17
Pin Postgres image to avoid unexpected data loss

chosak:
This repo's docker-stack.yml file specifies how we deploy cf.gov to internal PR preview environments. As part of that stack, we deploy a PostgreSQL container, and currently that image [is pinned to `postgres:10-alpine`](https://github.com/cfpb/consumerfinance.gov/blob/f4ee503637f2427097102f63da2809989f993d18/docker-stack.yml#L28).

When new commits are pushed to an open PR, the stack gets redeployed. Ideally, this should only replace any running containers that have changed, typically only the Python container running the cf.gov code. The Postgres container shouldn't be replaced; if it is, any data edits made (e.g. in Wagtail) will be lost unexpectedly and have to be redone.

Unfortunately, the pin `postgres:10-alpine` is insufficiently stable to prevent this from happening at unexpected times. The image with this tag can be updated:

1. When Postgres 10 has a new minor release, as just happened on November 11th with [Postgres 10.19](https://www.postgresql.org/docs/release/10.19/).
2. When Alpine OS has a new minor release, as happened last on June 15th with [Alpine 3.14](https://alpinelinux.org/releases/).
3. When the official Postgres Docker image is updated with a change independent of either Postgres or Alpine, as seems to happen very often, as frequently as [once per week](https://github.com/docker-library/postgres/commits/master/10).

In practice this causes a problem this way:

1. A developer opens a new PR. A new Docker stack is spun up. A new Postgres container is started. When the Python container starts up, it runs [code that checks if the DB is empty and, if so, initializes it](https://github.com/cfpb/consumerfinance.gov/blob/1f5988d17eec10a8e832c5e87d29b3c6d2b7ef4f/docker-stack.yml#L102-L109).
2. Subsequent commits to the PR cause a Docker stack redeploy, but these don't cause the Postgres container to get recreated. The DB state isn't reinitialized when the Python container restarts, because it isn't empty.
3. A user makes data edits to the running PR Docker stack, for example by creating content in Wagtail.
4. Some time passes, perhaps as little as a few days. A new version of the `postgres:10-alpine` image is released.
5. A developer pushes another commit to the open PR, perhaps something completely unrelated to the database. This triggers a redeploy of the Docker stack. Because the Postgres Docker image has been updated, the existing container is shut down and a brand new one is created, losing any data edits that were made. The restarting Python container then
reinitializes the now-incorrectly-empty database.

Thanks to @hkeeler for assistance with debugging this and brainstorming potential solutions.

One thought was to somehow alter the Docker stack deploy process, to prevent the Postgres container from being recreated after the first deployment. But there doesn't seem to be an easy way to do this in a clean way, especially the way we implement the stack deployment.

A simpler solution is to just pin the Postgres image version on a slightly older version that won't get additional updates. This
commit changes the pin to `postgres:10.18-alpine`, which had its last release [22 days ago](https://github.com/docker-library/official-images/commit/9a84dc2d308e75abf1a692a373f71960b0c68454#diff-fa6dc4e4c26179b634d7e1f1a74fd6bab4e442e444d5625ffc293e4a385a1738) (and shouldn't get any subsequent updates, now that Postgres 10.19 has been released).

As it happens, PostgreSQL 10.18 is the [latest version available on AWS RDS](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts.General.version10), and the one we are using in production.

This seems like a reasonable tradeoff: the risk is that we might miss out on some minor improvements to either Postgres or Alpine or the Postgres image, but in exchange we will have more reliable Docker stacks.

## How to test this PR

The Docker stack deploy should work as expected and deploy the right version of the Postgres image.

## Checklist


chosak:
> The Docker stack deploy should work as expected and deploy the right version of the Postgres image.

I confirmed that the expected image got deployed. Inspecting in MKE I see

`postgres@sha256:f96b0bf2a84590b528d84519a1b20ef865c2e5f2687925c943d6132ed529d81e`

which is the correct SHA for the [last release of 10.18](https://github.com/docker-library/repo-info/commit/1c23be0c862973ce9f1efff0c445666e36e8dd7d?short_path=cf24437#diff-cf244372247ddfd0c2638e4752773e38f7fe0b8805763861ab7ab5b8c3680239).

wpears (approved):

PR #6766, opened 2021-11-18
Prepaid agreements: sort issuer names in filter

virginiacc:
A stakeholder noticed that the values are no longer showing in alphabetical order in the issuer name filter on the prepaid agreements database page. I think the ordering may have originally occurred in the Django view and was affected by unrelated changes to that code. This PR moves the alphabetization to the front end, adding a sort filter when the multiselect is generated in the Jinja template.

---

## Additions

- Sort filter for issuer names in Jinja2 template

## How to test this PR

1. Check out branch
2. Navigate to [prepaid agreements database](http://localhost:8000/data-research/prepaid-accounts/search-agreements/) and verify that values appear in alphabetical order in issuer name multiselect

### Unsorted issuer names on production
Screen Shot 2021-11-18 at 12 02 15 PM

### Sorted issuer names in this branch

Screen Shot 2021-11-18 at 12 01 44 PM


chosak (approved):
This works, although it might be preferable to do this sorting on the backend so that there's less logic in the template. I suspect this bug was introduced in #6083, which changed how the backend worked. You can see in the diff that [this `order_by` clause](https://github.com/cfpb/consumerfinance.gov/pull/6083/files#diff-029c2648d818d241d5458b30e8f1d3f8da3aec1b5bb381a356a02847b4b69766L124-L125) was removed and not replaced elsewhere.

An alternate fix on the backend that seems to work:

```diff
diff --git a/cfgov/prepaid_agreements/views.py b/cfgov/prepaid_agreements/views.py
index 2812a8098..0fa4df0c2 100644
--- a/cfgov/prepaid_agreements/views.py
+++ b/cfgov/prepaid_agreements/views.py
@@ -15,7 +15,7 @@ from v1.models.snippets import ReusableText
 def get_available_filters(products):
     available_filters = {'prepaid_type': [], 'status': [], 'issuer_name': []}
 
-    for product in products.all():
+    for product in products.order_by('issuer_name'):
         prepaid_type = product.prepaid_type
         if prepaid_type and prepaid_type != '':
             if prepaid_type not in available_filters['prepaid_type']:
```

This orders the dropdown by issuer name but preserves the ordering of search results by product name.

virginiacc:
Thanks for that suggestion, @chosak! I updated in 771cc05ebe840b892c4f240a7bce72b0fb9f3f95 to sort the issuer names on the backend.

chosak:
@virginiacc I should have tested more -- my suggestion broke the tests. I've pushed a commit that fixes them, and also makes sure that the other filters are alphabetized. You didn't mention above, but the product types and statuses were also in random order. After my commit:

|Before|After|
|-|-|
image|image|

Does this change seem okay to you?

chosak (approved):

virginiacc:
That looks great -- thanks so much for your help on this, @chosak!!

PR #6771, opened 2021-11-23
Fence the TDP survey UI test to only run headed

willbarton:
This change fences off the TDP survey UI test to run only when Cypress is headed.

This will remove the test from our automated run (where it is currently failing [because it does not wait after a `click()` for the page to change before testing the change](https://github.com/cfpb/consumerfinance.gov/blob/main/test/cypress/integration/consumer-tools/youth-financial-education/survey-progress.js#L64-L69)).

This should allow time for a correct fix to the test while also not innuring us to broader failures of our integration tests. We [do something similar](https://github.com/cfpb/consumerfinance.gov/blob/main/test/cypress/integration/consumer-tools/youth-financial-education/survey-errors.js#L13) when testing TDP survey errors.

## How to test this PR

1. `yarn run cypress run` will run the tests headless locally. They should all pass.
2. The functional test run on this PR should succeed.

## Checklist


niqjohnson (approved):
Works for me. Locally, the tests in the new conditional run and fail as expected when run using `yarn run cypress run --spec test/cypress/integration/consumer-tools/youth-financial-education/survey-progress.js --browser chrome --headed`. When run headless using `yarn run cypress run --spec test/cypress/integration/consumer-tools/youth-financial-education/survey-progress.js --browser chrome`, those tests are skipped and all tests pass.

PR #6772, opened 2021-11-30
Simplify our setup docs and quickstart

willbarton:
This is an attempt to simplify our setup documentation. It is opinionated, but hopefully maintains the multiple paths one can take to running consumerfinance.gov.

I've preferenced the Docker setup for running cf.gov out-of-the-box, due to the comparative simplicity of the steps involved and the (IMHO) lower likelihood of breakage should Homebrew decided to update PostgreSQL or Elasticsearch. I have preserved the documentation on running cf.gov entirely locally with Homebrew, as well as running locally against Dockerized PostgreSQL and Elasticsearch for those that might wish to. 

This is submitted for discussion and comment — particularly by those who've gone through the setup recently.

There is probably additional work to be one to our "running" documentation as well, but I decided to keep this scoped to getting consumerfinance.gov up and working.

The changes can be previewed with GitHub's markdown rendering:

- https://github.com/cfpb/consumerfinance.gov/blob/install-docs/README.md
- https://github.com/cfpb/consumerfinance.gov/blob/install-docs/docs/installation.md

Or locally with MkDocs's rendering:

- http://localhost:8888/installation/

## Checklist


higs4281 (commented):
Just some typos. 

I wonder if it's worth the trouble to add `pyenv local consumerfinance.gov` to activate the env. It's not as intrusive as autoenv was, but may not be worth the complexity.

willbarton:
> I wonder if it's worth the trouble to add `pyenv local consumerfinance.gov` to activate the env. It's not as intrusive as autoenv was, but may not be worth the complexity.

Oh! I didn't know it could do that!

I think it might be worth it. Anyone else have thoughts?

README.md
```diff
@@ -11,55 +11,84 @@ and several standalone Django apps for specific parts of the site.
 Full installation and usage instructions are available in
 [our documentation](https://cfpb.github.io/consumerfinance.gov).
 
-This project requires Python 3.6, Node 14, and Gulp 4.
-We recommend the use of [virtualenv](https://virtualenv.pypa.io/en/stable/) and
-[virtualenvwrapper](https://virtualenvwrapper.readthedocs.io/en/latest/)
-for keeping the project's Python dependencies contained.
+This quickstart requires a working Docker Desktop installation,
```
chosak, line 14: To promote simplicity and avoid duplication, what do you think about removing these instructions from the README in favor of pointing to the docs site? Having the same quickstart content in two places introduces potential for drift, especially because we have to link to the sub-steps in different ways (absolute URLs here, relative in the docs).

willbarton: I think that's reasonable. Duplicating it made sense in an [ARID sort of way](https://www.writethedocs.org/guide/writing/docs-principles/#arid), but I'm fine removing it and linking.

chosak:
Nice work, @willbarton.

Having a "quickstart" that requires "a working Docker Desktop installation, as well as a pyenv-installed Python 3.6 with pyenv-virtualenv and Node.js 14 with Yarn" feels somewhat less than quick. Is there something we can do to simplify this? Some ideas:

- Under typical/quick usage, is a .env really necessary? As far as I can tell, the main thing the average person uses this for is to specify a database download URL, which we treat as an optional step anyway. We could create a separate documentation page around environment customization that mentions the DB download URL as one of the possible variables.
   - Another option to consider would be automatically loading the DB dump as part of container startup, using `docker-entrypoint-initdb.d` - see "Initialization scripts" [here](https://hub.docker.com/_/postgres). We've used this approach in our internal Ansible repo. 
- "Set up a local Python environment for unit testing" is something that a backend developer might want to do but isn't necessary just to get the project up and running. It adds significant complexity (Pyenv, Python version, virtualenvs) to support testing on the localhost while still encouraging Docker as the standard way of running Django. Could we consider encouraging use of `cfgov/manage.py test` in the container itself as an alternate to tox?
- Similarly, can we run the frontend build in the container, avoiding the need to install Node and Yarn on the host?
- Is `docker network create cfgov` really necessary? What is this step for?

With the above simplification, the quickstart would become more like:

- Clone the repo
- `docker-compose up`
- Run setup commands inside the containers - either via `docker-compose exec python whatever`, or maybe we make a new shell script ([like this](https://github.com/wagtail/docker-wagtail-develop/blob/master/setup-db.sh)) that encapsulates those quickstart commands.

willbarton:
@chosak I love all of that — and I agree.

I think we can drop the Python env setup from the quickstart, and possibly the use of `.env` (I'll need to do some testing to make sure).

Beyond that... I'll look into moving the frontend build, but I don't think I want to modify the actual setup steps in this PR as much as just document the current reality (our docs do **not** reflect that current reality).

> Is docker network create cfgov really necessary? What is this step for?

IIRC it was added to enable Cypress to test the Python container from its own container. But we've since removed that container. It can *probably* go away, but I'd like to verify.

docs/installation.md
```diff
@@ -1,417 +1,386 @@
...
-
-**Continue following the [usage instructions](../running-virtualenv/).**
+Where `[GHE]` is our GitHub Enterprise URL.
 
-## Docker-based installation
+### Build the frontend
 
-### Tools we use for developing with Docker
+!!! note
+    Our frontend quires [Node.js 14](https://nodejs.org/en/)
```
anselmbradford, line 180: typo `quires`

anselmbradford: Also should be Node 16 at this point.

jslay-excella (approved):

PR #6780, opened 2021-12-03
[simple-chart] Remove leading quotes from comment strings when excel sneaks them in

wpears:
Excel will sneak in quotes when saving as csv (appropriate) but this can break our comment-detection logic if it makes a line start with a quote.

### DIFF
--- a/cfgov/unprocessed/js/routes/on-demand/simple-chart/simple-chart.js
+++ b/cfgov/unprocessed/js/routes/on-demand/simple-chart/simple-chart.js
@@ -61,7 +61,9 @@ function fetchData( url, isCSV ) {
 
     return prom.then( d => {
       if ( isCSV ) {
-        d = d.replace( /(^"|\n")(#|\/\/)/g, '$2' );
+        /* Excel can put quotes at the start of our # or // comments
+           This strips those quotes */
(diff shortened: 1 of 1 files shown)


cfgov/unprocessed/js/routes/on-demand/simple-chart/simple-chart.js
```diff
@@ -61,6 +61,7 @@ function fetchData( url, isCSV ) {
 
     return prom.then( d => {
       if ( isCSV ) {
+        d = d.replace( /\n"(#|\/\/)/g, '$1' );
```
schbetsy, line 64: This won't remove a leading quotation mark on the first line of the file, since the first line doesn't have `\n` at the start.

schbetsy (approved):
This solves the issue at hand, though it doesn't account for future data files that might have a leading quotation mark on the first line. We could merge this, since we want to get the charts working, and backlog the first-line issue.

cfgov/unprocessed/js/routes/on-demand/simple-chart/simple-chart.js
wpears: Good catch!

schbetsy (approved):

PR #4136, opened 2018-05-14
Bump wagtail-inventory version

virginiacc:
## Changes

- Bumps wagtail-inventory to latest version

PR #6781, opened 2021-12-04
Apply "Edge-Cache-Tag" to NewsroomLandingPage headers

higs4281:
The goal is to invalidate Akamai cache on all flavors of the newsroom landing page 
– including all variations of query-string faceting that have been cached – 
when a child page has been published.

The standard cache strategy of purging by URL doesn't work in this case,
because there are limitless query string combinations for categories,
topics, languages, date ranges and keywords.

Applying an Edge-Cache-Tag to all renderings of the NewsroomLandingPage
allows Akamai to invalidate all the variations it has cached at once.

Cache_tag invalidation is triggered when a NewsroomPage or a LegacyNewsroomPage
gets published.

### Testing

We plan to test the triggered cache invalidation on beta.consumerfinance.gov.
In the meantime, you can see that the Edge-Cache-Tag is getting applied to all
variations of /about-us/newsroom/:
- Open your browser dev tools to the Network panel
- Visit <localhost:8000/about-us/newsroom/> and check the response header
for the presence of "Edge-Cache-Tag: newsroom." You may need to hard-refresh
your browser.
- Apply a number of facets and/or search keywords and then see that the result
page also has the tag applied.

This change addresses platform issue `#4136`

### DIFF
--- a/cfgov/v1/models/browse_filterable_page.py
+++ b/cfgov/v1/models/browse_filterable_page.py
@@ -22,6 +22,9 @@ from v1.models.filterable_list_mixins import (
 from v1.models.learn_page import EventPage
 
 
+NEWSROOM_CACHE_TAG = "newsroom"
+
+
 class BrowseFilterableContent(StreamBlock):
     """Defines the StreamField blocks for BrowseFilterablePage content.
 
@@ -122,3 +125,8 @@ class NewsroomLandingPage(CategoryFilterableMixin, BrowseFilterablePage):
     filterable_categories = ['Newsroom']
 
     objects = PageManager()
+
+    def serve(self, request, *args, **kwargs):
+        response = super().serve(request, *args, **kwargs)
+        response['Edge-Cache-Tag'] = NEWSROOM_CACHE_TAG
+        return response
--- a/cfgov/v1/models/__init__.py
+++ b/cfgov/v1/models/__init__.py
@@ -8,7 +8,8 @@ from v1.models.base import (
 )
 from v1.models.blog_page import BlogPage, LegacyBlogPage
 from v1.models.browse_filterable_page import (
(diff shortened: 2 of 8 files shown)


cfgov/v1/models/browse_filterable_page.py
```diff
@@ -22,6 +22,9 @@
 from v1.models.learn_page import EventPage
 
 
+NEWSROOM_CACHE_TAG = "newsroom"
```
willbarton, line 25: If this were something like the newsroom page slug (or any filterable list's page slug), could this be implemented on the filterable list itself, instead of being specific to Newsroom? We already have logic [in `signals.py`](https://github.com/cfpb/consumerfinance.gov/pull/6781/files#diff-79d70b0e1eabb1410a6294959f01356ac3d11035c190a72b93f0e512e2032116R111-R121) to determine whether a page belongs to a filterable list.

higs4281: I think the options for cache-tag use could grow a lot once we get access to Akamai configs and to the API cache_tag invalidation request. Newsroom seems to be the most pressing case, and I'd be thrilled to get that far.

higs4281: Now that Akamai config has been ironed out, I'd propose merging this to fix newsroom caching, and then moving on to other targets, such as the landing page at `/about-us/blog/`, which could get a cache-tag purge whenever a BlogPage or LegacyBlogPage is published.

Since we control how cache-tags are applied, with no Akamai configs needed, the newsroom tag won't lock us in to any scheme if we want to use namespaces or concoct a wider filterable-list cache group.

Simple tags makes sense to me for targeted purges of special pages like `about-us/blog/` and `about-us/newsroom/`

jslay-excella (approved):
Works for me.

PR #6783, opened 2021-12-06
Support testing of alternate Elasticsearch hosts

chosak:
When running tox to run the Python tests, it isn't currently possible to pass the `ES_HOST` environment variable, meaning that the tests always try to run against localhost. There are some cases where it would be useful to be able to use an alternate Elasticsearch:

- in a Docker environment where the Elasticsearch container is not exposed on a local port
- when Elasticsearch is running locally, but with X-Pack authentication, requiring use of a username and password.

This commit adds `ES_HOST` to the list of environment variables that can be passed to tox.

## Checklist


higs4281 (approved):
Seems like a useful option.

PR #6784, opened 2021-12-06
add sourceType param to babel config

higs4281:
The patch is from a babel [issue](https://github.com/babel/babel/issues/12731#issuecomment-819158117)


anselmbradford:
Note that we should update `module.exports =` calls to `export default` in the future and then could probably remove this config setting.

anselmbradford (approved):

PR #6788, opened 2021-12-07
Introduce distinct search for beta.consumerfinance.gov

chosak:
This PR modifies the behavior of the website when deployed to beta.consumerfinance.gov, so that global site searches use beta-specific search indexes. Recall that our global site search currently uses Search.gov and has two indexes: `cfpb` for regular English-language pages and `cfpb_es` for Spanish-language pages.

This change modifies how the website works on beta such that it uses `cfpb_beta` and `cfpb_beta_es`. This is reflected two ways: first, the header search now queries Search.gov with those affiliate codes; and second, the Search.gov [JavaScript snippet](https://search.gov/manual/code.html) now uses these codes as well. Under the hood this behavior relies on our existing `DEPLOY_ENVIRONMENT` Django setting.

In order to support search on beta, this PR also modifies how we serve our robots.txt file. Currently, robots.txt is a hardcoded text file which allows all robots and disallows certain paths. We currently override robots.txt on beta as part of our (internal) deploy code. With this PR, robots.txt is now served by Django instead, and uses a new `ROBOTS_TXT_SEARCH_GOV_ONLY` feature flag in order to serve different robots directives on beta. Specifically, on beta we will now attempt to block all robots except the Search.gov one, based on its [`usasearch`](https://search.gov/manual/go-live.html) user agent.

Finally, a minor bit of cleanup included here is removal of the `show_banner` parameter to our global header; this is never used and has been superseded by alternate banner code.

## How to test this PR

To test, you can run a local server and verify that searches on http://localhost:8000/ and http://localhost:8000/es/ work as expected (look at the `affiliate=cfpb` part of the search query string when you submit the form), and that the included JS snippet has the right code (look in source for `usasearch_config`).

Then, try running a local server but set the `DEPLOY_ENVIRONMENT` environment variable to `beta`, and confirm that the Search.gov codes update as expected.

## Notes and todos

- This change requires removing the code in our internal deployment repo that overwrites robots.txt on beta; I will open a separate PR for that purpose. Edit: a-c-f#394
- We could use some documentation of how we use Search.gov.

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/molecules/global-search.html
+++ b/cfgov/jinja2/v1/_includes/molecules/global-search.html
@@ -14,10 +14,18 @@
 
 {% macro render( language='en' ) %}
 
+{%- set search_url_base = "https://search.consumerfinance.gov/search" -%}
+{%- set search_url = (
+    search_url_base
+    ~ "?utf8=✓"
+    ~ "&affiliate=" ~ search_gov_affiliate()
+    ~ "&query="
+) -%}
+
 <div class="m-global-search"
      data-js-hook="behavior_flyout-menu">
     <div class="m-global-search_fallback">
-        <a href="https://search.consumerfinance.gov/search?utf8=%E2%9C%93&affiliate=cfpb&query=">
+        <a href="{{ search_url }}">
             {{ _('Search') }}
         </a>
     </div>
@@ -39,7 +47,7 @@
          aria-expanded="false"
          role="search">
         <form class="m-global-search_content-form"
-              action="https://search.consumerfinance.gov/search"
+              action="{{ search_url_base }}"
               method="get">
             {#
                 The following two inputs are required by the search portal.
@@ -51,7 +59,7 @@
             <input type="hidden"
                    id="affiliate"
                    name="affiliate"
-                   value="{{ 'cfpb_es' if language == 'es' else 'cfpb' }}">
+                   value="{{ search_gov_affiliate() }}">
 
             <div class="o-form__input-w-btn">
                <div class="o-form__input-w-btn_input-container">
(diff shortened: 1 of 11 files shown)


cfgov/jinja2/v1/_includes/molecules/global-search.html
```diff
@@ -14,10 +14,16 @@
 
 {% macro render( language='en' ) %}
 
+{%- set search_url_base = "https://search.consumerfinance.gov/search" -%}
+{%- set search_url = (
+    search_url_base ~ "?utf8=%E2%9C%93&affiliate=" ~ search_gov_affiliate() ~ "&query="
+    "query="
```
higs4281, line 20: Not following the purpose of the extra "query=", but it doesn't show up in use.

higs4281 (approved):
This all looks good, and works as advertised. 
I had one question about setting the search URL in jinja2, but it's probably my lack of understanding.

cfgov/jinja2/v1/_includes/molecules/global-search.html
chosak: Nice catch! That looks like a rebase/editing error on my part, that actually does break things when JavaScript is disabled:

image

Will correct.

higs4281: Also was wondering about the stand-in "?utf8=✓" query-string starter. It's often used so all subsequent query bits can start with "?" – but I though specifying the unicode ✓ character forced some older browsers to use UTF8. Wouldn't we use ✓ instead of the url-quoted entity for such a purpose?

chosak: > It's often used so all subsequent query bits can start with "?"

Can you explain more? This code existed before this PR, and I wasn't really sure the merits of the different options. In the Search.gov docs they also use `value="&#x2713;"` in the form whereas we use `value="✓"`. 

It looks like @anselmbradford added the `utf8=%E2%9C%93` [a mere 16 months ago](https://github.com/cfpb/consumerfinance.gov/commit/e894384dc7bb64526f40417a500c2da4a1fce0e4) - @anselmbradford is there a particular reason why the link is formatted this way?

higs4281:
TIL: How to concatenate strings in jinja2 with `~`

cfgov/jinja2/v1/_includes/molecules/global-search.html
higs4281: I guess if we're using a dummy query starter, I'd use `?utf=✓` and have it do some good in the world.

https://stackoverflow.com/questions/27338154/why-do-some-websites-have-utf8-in-their-title#answers-header

higs4281: > It's often used so all subsequent query bits can start with ~~"?"~~ "&"

My typo. 

When query-building, you don't have to decide which query term begins with "?" if you always start with `?utf8=✓`

chosak: Thanks @higs4281; I've added that change and rebased it in to clean up the commits.

anselmbradford: Yeah I'm not recalling anything on the URL-encoding of the checkmark. This is for the link when the JS is turned off and the browser appears to parse the encoding before it has been clicked. This is hovering the link:
Screen Shot 2021-12-08 at 9 44 37 AM

PR #6789, opened 2021-12-08
Don't serve robots.txt from Apache

chosak:
#6788 tried to move robots.txt into Django, but neglected to remove the existing Apache rule for that alias. That was an oversight, which is corrected by this commit.

## How to test this PR

Run a local server and verify that http://localhost:8000/robots.txt still works as expected.

## Checklist


contolini (approved):
Ty for your service 👍

PR #6791, opened 2021-12-10
Add optional SSO support and configuration

willbarton:
This change implements support for SAML2 single sign-on based on [djangosaml2](https://djangosaml2.readthedocs.io/) and [pySAML2](https://pysaml2.readthedocs.io/).

It adds environment variables that, when populated, will enable SAML2 authentication, and the required URL patterns, against the SAML2 identity provider with the given metadata URL using the given entity ID for the Django site. ***Without those environment variables, SAML2 auth will not be enabled and all existing authentication will continue to work as it does now.***

It does require the `xmlsec1` binary to be in place at `/usr/bin/xmlsec1`. That requirement will be handled in our deployment playbooks. There is a corresponding PR for that change and with dev2 configuration in that internal repository.

## How to test this PR

This PR is deployed with SSO enabled to dev2 at present for testing. However, you will need test credentials (I believe everyone on the team should now have those).

This is not easily testable locally. Please reach out to me and I can walk you through it. 

In short:

- Change docker-compose.yml to expose the `python` container on port 443:

   ```yaml
   python:
       …
       ports:
         - 127.0.0.1:443:8000
   ```
- Change Dockerfile run `runsslserver` instead of `runserver`

   ```dockerfile
   CMD ["python", "./cfgov/manage.py", "runsslserver", "0.0.0.0:8000"]
   ```

- Rebuild your `python` Docker container:

   ```shell
   docker-compose build python
   ```

- Add the following variables to your `.env` file (reach out to me for their expected values) and restart the Docker container:

   ```shell
   export SAML_AUTH=True
   export SAML_ROOT_URL="https://<dev2 FQDN>"
   export SAML_ENTITY_ID="<dev2 entity ID>"
   export SAML_METADATA_URL="<dev2 metadata URL>"
   ```

- Edit your `/etc/hosts` to point our dev2 FQDN to 127.0.0.1
- Start or restart your `python` Docker container. Verify that you can access that dev2 FQDN over https locally.
- Try to access the admin at `/admin`. This should redirect you to our test SAML2 provided. As will using dev2 directly, you'll need test credentials to login.

## Checklist


cfgov/cfgov/urls.py
```diff
@@ -597,14 +597,48 @@ def empty_200_response(request, *args, **kwargs):
 urlpatterns = urlpatterns + category_redirects
 
 if settings.ALLOW_ADMIN_URL:
-    patterns = [
-        re_path(r'^login/$', login_with_lockout, name='cfpb_login'),
+    # If SAML2 auth is enabled, /login will redirect to /saml2/login,
+    # which in turn will redirect to the configured identity provider.
+    if settings.SAML_AUTH:
```
chosak, line 602: To avoid adding this complexity to this file, what do you think about creating some other auth-specific urls.py, perhaps in a separate auth-specific app? We could then move the various v1 auth-specific views in there as well.

cfgov/cfgov/urls.py
```diff
@@ -597,14 +597,48 @@ def empty_200_response(request, *args, **kwargs):
...
+                RedirectView.as_view(
+                    url='/saml2/logout/',
+                    query_string=True,
+                ),
+                name='logout'
+            ),
+
+            # For backup/admin/emergency auth, a pure Django login is still
+            # provided.
+            re_path(r'^django_login/$', login_with_lockout, name='cfpb_login'),
```
chosak, line 624: Another possible simplification, that avoids switching which URL goes to which view depending on environment: could we modify [`settings.LOGIN_URL`](https://github.com/cfpb/consumerfinance.gov/blob/bfaf14ad0accbb350538066a6f29b3c1524d28eb/cfgov/cfgov/settings/base.py#L407) if SAML is enabled?

Personally I typically login via /admin/, and when we share admin views around, they get shared as /admin/something. I believe Wagtail will redirect to `LOGIN_URL`, not /login/ specifically, as the way for users to login. If we point `LOGIN_URL` to the SAML login URL when enabled, we can leave urls.py a little simpler because we're just adding a new path, while leaving the old URLs alone.

I'm less sure if there's something similar that can be done with logout.

cfgov/cfgov/settings/production.py
```diff
@@ -95,3 +99,75 @@
...
+    # Map Django attributes to our SAML identity provider
+    SAML_DJANGO_USER_MAIN_ATTRIBUTE = 'email'
+    SAML_DJANGO_USER_MAIN_ATTRIBUTE_LOOKUP = '__iexact'
+    SAML_ATTRIBUTE_MAPPING = {
+        'emailAddress': ('email', ),
+    }
+    SAML_CREATE_UNKNOWN_USER = False
+
+    # URL lookups
+    SAML_ROOT_URL = os.environ.get("SAML_ROOT_URL")
```
chosak, line 137: Is it worth changing these various `os.environ.get()` to `os.environ[]`, to deliberately raise an exception if the SAML specific environment variables aren't set, if `SAML_AUTH` is? Otherwise, these settings will be set to None, which will presumably raise some unpredictable error downstream. By using `[]`, we'd know right away because the app won't even start up.

cfgov/cfgov/settings/production.py
```diff
@@ -95,3 +99,75 @@
         "Environment variable ALLOWED_HOSTS is either not defined or is "
         "not valid JSON. Expected a JSON array of allowed hostnames."
     )
+
+# SAML2 Authentication
+#
+# Requires the SAML_AUTH, SAML_ROOT_URL, SAML_ENTITY_ID, and SAML_METADATA_URL
+# environment variables to be defined.
+#
```
chosak, line 107: Worth including a link here [to the docs](https://djangosaml2.readthedocs.io/contents/setup.html#configuration)?

.env_SAMPLE
```diff
@@ -179,6 +179,18 @@ export WATCHMAN_TOKENS=''
 #export NEW_RELIC_LICENSE_KEY=<new relic license key>
 #export NEW_RELIC_APP_NAME="cf.gov <your username here> python"
 
+############################################################################
+# SAML2 Authentication for local testing
+#
+# If testing our SAML2 authentication support, these variables are required,
+# as is running with the production settings.
```
chosak, line 186: Worth including a link here [to the docs](https://djangosaml2.readthedocs.io/contents/setup.html#configuration)?

cfgov/cfgov/urls.py
willbarton: I like that idea a lot, acutally.

cfgov/cfgov/urls.py
willbarton: Hmm, I'll see how that works. I think I had an issue with just setting `LOGIN_URL` earlier in my testing, but this has been a very start/stop-y iterative process.

willbarton: Correction, [`WAGTAIL_FRONTEND_LOGIN_URL` appears to only apply to page privacy](https://github.com/wagtail/wagtail/blob/main/wagtail/core/wagtail_hooks.py#L17-L41), not to the general admin.

So, I think we're going to need to at least redirect `/admin/login` (where Wagtail lands you) to the SAML2 URL.

willbarton: Because Wagtail doesn't respect `LOGIN_URL`, and because we have a permanent redirect of `/admin/login` to `/login`, I'm inclined to keep `/login` the point that redirects to SSO.

However, maybe instead of continuing to expose our Wagtail login, we simply allow `/django-admin/login` to function, as our emergency admin login point?

.env_SAMPLE
willbarton: Done in e765fc6.

cfgov/cfgov/settings/production.py
willbarton: Done in e765fc6.

cfgov/cfgov/settings/production.py
willbarton: Done!

cfgov/cfgov/urls.py
willbarton: Moved in 33e1ae1!

cfgov/cfgov/urls.py
willbarton: This is how it's currently implemented in 5636600.

willbarton:
The missing line of coverage in `login/views.py` I'm inclined to leave as-is for this PR. I believe we'll be able to remove most or all of our custom login/password/permission handling code, or at the very least refactor it, once SSO is enabled. I'm not sure it's worth spending a lot of time refactoring it before that.

wpears (approved):

cfgov/cfgov/settings/production.py
```diff
@@ -95,3 +99,80 @@
...
+    ACS_URL = format_lazy(
+        '{root_url}{acs_path}',
+        root_url=SAML_ROOT_URL,
+        acs_path=reverse_lazy('saml2_acs')
+    )
+    ACS_DEFAULT_REDIRECT_URL = reverse_lazy('check_permissions')
+
+    # Configure PySAML2 for our identity provider
+    SAML_CONFIG = {
+        'debug': 1,
```
willbarton, line 152: This should be set to `settings.DEBUG`

willbarton: Fixed in ae06f21

PR #6792, opened 2021-12-13
Upgrade elasticsearch to 7.16.1

wpears:


willbarton (approved):
At some point soon I'd like us to switch to using the `opensearchproject/opensearch` images that more closely mirror our production setup.

PR #6793, opened 2021-12-13
Refactor RAFT to use JS/CSS assets from app assets

anselmbradford:
RAFT = https://github.com/cfpb/rental-assistance-finder/

## Changes

- Remove RAFT gulp copy task and serve JS via app assets JS.
- Move RAFT inline CSS to css file in app assets CSS.

## How to test this PR

1. `gulp clean && gulp build` and http://localhost:8000/coronavirus/mortgage-and-housing-assistance/renter-protections/find-help-with-rent-and-utilities/ should work like production.


cfgov/jinja2/v1/_includes/blocks/raf_tool.html
```diff
@@ -1,19 +1,6 @@
...
-    word-break: break-all;
-  }
-  #rental-assistance-finder .m-form-field + .m-form-field {
-    margin-top: 15px
-  }
-  .no-js .js-is-hidden {
-    display: block !important;
-  }
-</style>
+<link rel="stylesheet" href="{{ static('apps/erap/css/main.css') }}">
```
anselmbradford, line 1: I moved this to the app assets so the JS and CSS were in one place, but it could be inlined as it was if that is preferred.

cfgov/unprocessed/apps/erap/css/main.less
```diff
@@ -0,0 +1,12 @@
+#rental-assistance-finder {
```
virginiacc, line 1: It would be nice to pull this from the CSS distributed in the rental-assistance-finder npm module, but we probably want to update the code in the RAFT repo to generate a single `main.css` first.

virginiacc (approved):
👍 This looks good to me! I added a couple of aspirational comments but nothing that needs to happen now. :)

cfgov/jinja2/v1/_includes/blocks/raf_tool.html
virginiacc: Ideally we'd be able to load the tool's CSS in `base.html` from a `Media` entry on the RAFT block (the way the JS is currently added to the page), but that would require some global work so I think this update makes sense now.

PR #6803, opened 2021-12-20
Update project requirements from node 14 to 16

anselmbradford:
## Changes

- Update project requirements from node 14 to 16

## How to test this PR

1. PR checks should pass.


willbarton (approved):
I think this looks good! Bumping it in `docker/deployable-zipfile/Dockerfile` should handle artifact-building on Jenkins, so there should not be an internal changes needed!

PR #6806, opened 2021-12-21
Update our Elasticsearch image to 7.16.2

willbarton:
This bumps our Elasticsearch version to 7.16.2, which uses log4j 2.17.0.

It also adds some `echo`s to help debugging Elasticsearch pushes in future.

## Checklist


Jenkinsfile
```diff
@@ -128,6 +128,8 @@ pipeline {
                                 IS_ES_IMAGE_UPDATED = 'false'
```
willbarton, line 128: In future ES bump PRs, if we need to force an image build and push, temporarily overriding this variable to `true` will do that.

wpears (approved):

PR #6810, opened 2021-12-23
Move draftail dependencies to cfpb-admin assets app

anselmbradford:
~~These dependencies appear unused in the project. I may be missing how these are used, so please review draftail functionality closely.~~

They are used to satisfy peerDependencies of the draftail plugin.

## Removals

- ~~Remove unused draftail dependencies.~~

Edit:
## Changes

- Move draftail dependencies from global package.json to a new cfpb-admin assets app.
- Remove dedicated scripts:admin gulp task and use existing assets app infrastructure.

## How to test this PR

1. PR checks should pass.
2. Draftail should work as before. I'm unclear how to test… maybe 

You can checkout this branch locally and use ./runserver.sh
or use the Docker container that this Pull Request creates to:

1. Open or create a wagtail page
2. Modify the rich text
3. Click to save the changes
4. Verify the changes are correctly saved with no errors locked

From https://github.com/cfpb/consumerfinance.gov/pull/5899


wpears:
These are peer dependencies (or dependencies of peer dependencies) of draft-js-export-html and are therefore required

wpears:
It may be that we have a better method or can use things already included in collectstatic though

anselmbradford:
> These are peer dependencies (or dependencies of peer dependencies) of draft-js-export-html and are therefore required
> It may be that we have a better method or can use things already included in collectstatic though

Yea if I run `python cfgov/manage.py collectstatic` draftail.js is in `collectstatic/wagtailadmin/js/draftail.js`. I don't see how a copy in `node_modules`, but never referenced, would be used. The rich-text-table.js [expects `window.DraftJS`](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/js/admin/rich-text-table.js#L113) to be available, but without an import from `node_modules` for draftJS anywhere in our codebase, I'd say this has to be coming from the version inside the wagtail admin [settings](https://docs.wagtail.io/en/v2.12.1/reference/settings.html#rich-text).

We'd need a bewd to look though, as I'm not certain what to check for this.

anselmbradford:
Hrm though I do see 

```
 Module not found: Error: Can't resolve 'draft-js' in '/home/runner/work/consumerfinance.gov/consumerfinance.gov/node_modules/draft-js-utils/esm'
Module not found: Error: Can't resolve 'immutable' in '/home/runner/work/consumerfinance.gov/consumerfinance.gov/node_modules/draft-js-utils/esm'
```

on the build from GH actions.

wpears:
[rich-text-table requires draft-js-export-html](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/js/admin/rich-text-table.js#L3)
[draft-js-export-html requires draft-js-utils](https://github.com/sstur/draft-js-utils/blob/master/packages/draft-js-export-html/package.json#L20) (and also specifies draft-js and immutable as peer deps)
[draft-js-utils imports from draft-js](https://github.com/sstur/draft-js-utils/blob/master/packages/draft-js-utils/src/callModifierForSelectedBlocks.js#L1) (and also specifies draft-js and immutable as peer deps).

wpears:
Generally, the wagtail frontend code gets included by being put in the `static` directory of one of our [`INSTALLED_APPS`](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/cfgov/settings/base.py#L44-L115). `collectstatic` picks those files up by default and dumps them into the `STATIC_ROOT`.

The fact we have this weird little file (rich-text-table.js) that's doing a bunch of crud and also pulling in duplicate libraries to what's already built and included in wagtail is PRETTY GROSS.

So I'd say instead of trying to find exact what to remove here that won't technically break the site, we should re-architect how this file is being include/what it's doing (eg, can we achieve stateToHTML in a different way and just drop this whole chain).

anselmbradford:
Updated this to move the admin js to an asset app named cfpb-admin, which removes the bespoke scripts:admin gulp task and uses our existing app assets infrastructure. This doesn't remove the peer dependencies, but it moves them to their own package.json scope, which may make them clearer to deal with later (at least it makes it clear which dependencies are there for the admin JS).

wpears (approved):
Neat!

Scotchester:
Seems like a lot of hoop jumping for rich text in tables. I suppose the work is already done, but I just wanted to call out, in case anyone else hadn't, that [Wagtail 2.15 includes a new table block](https://docs.wagtail.io/en/stable/reference/contrib/typed_table_block.html) that supports rich text :)

PR #6811, opened 2021-12-29
Remove engines setting in app assets

anselmbradford:
The engines setting is optional and should pick up the engines set in the root directory package.json, which seems preferable for maintenance purposes.

## Removals

- Node `engines` settings in individual app assets package.json files.

## How to test this PR

1. PR checks should pass.


contolini (approved):
Works locally 👍

PR #6816, opened 2022-01-04
Add a managemant command for purging Akamai by cache tag

higs4281:
Currently we have signals that trigger a cache-tag purge
for newsroom searches when news items change, but we need
to be able to issue cache-tag purges from Jenkins, via
a management command.

This provides the command to handle automated purge requests.
The intial goal is to enable daily purges of complaint search
URLs in all their query-string varieties.

The actual application of cache tags to complaint response headers
will be in following PRs.


jslay-excella (approved):
lgtm, tests and all

PR #6821, opened 2022-01-11
Add page path and depth to Elasticsearch index

chosak:
This commit adds two new fields to the Elasticsearch index that we use for searching filterable pages: path and depth. These fields exist on the core Wagtail page model and come from the django-treebeard package used to organize the page hierarchy.

With these two fields, it is possible to construct a search to that filters by a wide variety of page relationships: parents, children, etc. The current Elasticsearch index only includes the page URL, which makes it harder to, for example, filter only direct children of a given page.

The core Wagtail page filtering based on Django querysets uses page path and depth, as does the core Wagtail page search.

We have an open internal issue platform#4147 regarding our Events page (https://www.consumerfinance.gov/about-us/events/) - this page currently incorrectly includes archived events. This bug was introduced [when filterable page search was migrated to Elasticsearch](https://github.com/cfpb/consumerfinance.gov/pull/6344/files#diff-2530b52656353c72735085eb64afb6580522ec0df5f31bd0e0c5d9986c0c8d70L86), because the current logic doesn't support filtering only to direct children.

## How to test this PR

Because this only introduces new fields, there's no real way to test it beyond inspecting the index manually. One way to do that is to first reindex pages:

```sh
cfgov/manage.py search_index --rebuild -f --models v1
```

and then inspect the indexed data, for example in Kibana, to confirm that the new fields exist.

## Screenshots

Screenshot from Kibana showing the new fields:

image

Note that `depth: 5` is expected here; the root Wagtail page is depth 1, our site root (/) is depth 2, and so pages like /about-us/newsroom/something/ are depth 5.

## Notes and todos

This commit only introduces these new fields, and doesn't modify the search logic to rely on them. That will need to be introduced in a subsequent PR; I'm splitting this up so that the new index configuration can get rolled out before any logic that relies on it.

## Checklist


higs4281 (approved):
These look like useful fields to index, and I can see them added to the V1 index.

PR #6823, opened 2022-01-11
Un-appify raft, since it's just included as an on-demand block in wagtail

wpears:
Alternate approach to https://github.com/cfpb/consumerfinance.gov/pull/6822


wpears:
Went with 6822!

PR #6826, opened 2022-01-12
Bump design-system deps to 0.14.0

wpears:
Most interesting change is dropping ie7 support, which we can no longer test with saucelabs and which has <0.1% usage.


anselmbradford:
Changes, for ref https://github.com/cfpb/design-system/pull/1343


Want to ditch https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/css/enhancements/base.less in here too?

anselmbradford (approved):

PR #6827, opened 2022-01-12
Update to Python 3.8

willbarton:
This PR updates our `tox.ini`, `Dockerfile`, and deployable-zipfile `Dockerfile` to Python 3.8 from Python 3.6. 

To support both our deployments to Docker without a virtualenv and our deployments to servers using a virtualenv, it supplies two WSGI configuration files, one that does set `python-home` as an argument to `WSGIDaemonProcess`, and one that does.

## How to test this PR

Use the production-like Docker containers to run with Apache, SCL Python 3.8, and the mod_wsgi configuration without a virtualenv:

```
docker-compose -f docker-compose.yml -f docker-compose.prod.yml up --build
```

To test with a virtualenv requires deploying this PR to a dev server using our internal Ansible playbooks on the `python38` branch.

## Notes and todos

- We need to update our documentation to require Python 3.8 rather than 3.6. #6772 might also need to be updated.

## Checklist


higs4281 (approved):
works great; apache running and 3.8 in use.

PR #6831, opened 2022-01-13
Remove inaccurate/invalid leading plus

wpears:
Printing should still be non-broken in firefox


cfgov/unprocessed/apps/youth-employment-success/css/main.less
```diff
@@ -365,7 +365,7 @@ input:-moz-ui-invalid {
 }
 
 /** FIREFOX PRINT FIXES **/
-+.o-mega-menu_content {
```
anselmbradford, line 368: I recall `&` being implied in less, but this doesn't appear to be nested. Do we want to add `&` to this one? https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/apps/youth-employment-success/css/main.less#L281

anselmbradford: I guess that's outta scope.

anselmbradford (approved):

cfgov/unprocessed/apps/youth-employment-success/css/main.less
wpears: Just trying to prevent bad rules from being generated. So I suppose I'll call it out of scope :)

PR #6832, opened 2022-01-13
CCA: Improve import_agreements command to prevent encoding error

schbetsy:
Update the `import_agreements` management command, which is used to import a zipfile of credit card agreements into the credit card agreements database.

Previously, the command would only process a zip file if it contained a directory structure like `/Issuer/agreement.pdf`. Now, this command can process a zip file with an additional directory layer, like `/Containing-folder/Issuer/agreement.pdf`. This is the directory structure that is produced by default when downloading the file of agreements from Office365.

Previously, the process to update the agreement database included a step of downloading a file from O365, modifying its directory structure, then zipping and uploading that file. When that step was done on a Mac, it introduced encoding errors, so we want to avoid that extra processing if possible. 

## Changes

- Modify the `import_agreements` management command to allow a slightly different directory structure in the target zipfile. This allows us to simplify our process for updating the agreements database, removing the need for manual file-munging steps.
- Re-order the steps in the management command so that existing agreements are only deleted if the input file is valid and ready to import.

## How to test this PR

1. Run this branch. Ask me for a sample zipfile of agreements downloaded from O365.
2. In your environment, `unset AGREEMENTS_S3_UPLOAD_ENABLED ` (in case you had it set previously, so you don't overwrite agreements on S3).
2. Run the management command on the zipfile: `cfgov/manage.py import_agreements -p agreements.zip`
3. See that the command runs without errors.
4. Visit http://localhost:8000/credit-cards/agreements/. See that the drop-down list leads to correctly-imported issuers and agreements.
4. Visit http://localhost:8000/credit-cards/agreements/issuer/chase-bank-usa/. See that the Spanish-language agreements now say "Tarjeta de Crédito" instead of "Tarjeta de Cr├çdito"

## Screenshots
Screen Shot 2022-01-13 at 12 54 34 PM

## Checklist

- [x] Project documentation has been updated, including:
  - Wiki page on updating the credit card agreements database


higs4281 (commented):
TIL about ZIP file encoding specs!

This works as advertised, encodings and all.

There are a couple lost lines of test coverage, which I have yet to explore.

higs4281:
Might be good to advise unsetting `AGREEMENTS_S3_UPLOAD_ENABLED` if one happens to have that set up locally.

cfgov/agreements/management/commands/_util.py
```diff
@@ -29,6 +30,16 @@ def upload_to_s3(pdf_obj, s3_key):
     s3_client.upload_fileobj(pdf_obj, AWS_STORAGE_BUCKET_NAME, s3_key)
 
 
+def filename_in_zip(file_info):
+    # Zip files default to IBM Code Page 437 encoding unless a specific bit
+    # is set. See Appendix D in the zip file spec:
+    # https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT
+    if (file_info.flag_bits & 0x800) == 0:
```
higs4281, line 37: I wish I were more conversant in bitwise operations!

higs4281 (approved):
A great solution. Thanks.

PR #6836, opened 2022-01-14
Bump ds to 0.14.1, fixing nav list style type

wpears:
We took out the [base.less](https://github.com/cfpb/consumerfinance.gov/pull/6826/files) enhancement, because we were tricked into thinking it was IE7 only. Looking again at the style, it's clearly a core style, so it was added in https://github.com/cfpb/design-system/pull/1344


anselmbradford:
The bullets in the mega menu items are no more!

anselmbradford (approved):

PR #6837, opened 2022-01-18
Bump ccdb5-api to version 1.6.4

higs4281:
This change enables cache tags for CCDB API responses, which should resolve
internal GHE 4306.

## Testing
The code has been pushed to DEV4, and an API response like the one below should show `Edge-Cache-Tag: complaints` in the response header.

DEV4/data-research/consumer-complaints/search/api/v1/?date_received_max=2022-01-18&date_received_min=2011-12-01&field=all&size=25&sort=created_date_desc&size=0


flacoman91 (approved):
👍

PR #6838, opened 2022-01-21
Alpine Docker Image

jslay-excella:
Switch to `python:3.8-alpine` Docker image in preparation for Kubernetes.


willbarton:
@jslay-excella will this change anything about our (poorly organized, at best) [Docker documentation](https://cfpb.github.io/consumerfinance.gov/running-docker/)?

jslay-excella:
Everything looks good and passes. Apache changes are in place. Infra is ready.

willbarton (approved):
Alright. We should communicate with devs who use Docker that the base image is changing before we merge.

Is there reason to preserve the individual commit history here, or can this branch be squashed entirely, or the history compressed somewhat?

This is fantastic work, @jslay-excella. I love the reduced complexity. It looks good to me, with the above two things!

jslay-excella:
> Alright. We should communicate with devs who use Docker that the base image is changing before we merge, what that means, and what they should do.
> 
> Is there reason to preserve the individual commit history here, or can this branch be squashed entirely, or the history compressed somewhat?
> 
> This is fantastic work, @jslay-excella. I love the reduced complexity. It looks good to me, with the above two things!

I've cleaned up the squash commit message to include all the pertinent changes.

PR #6842, opened 2022-01-24
Mock less in form_module_handlers tests

willbarton:
This change removes our mocking of a page, a request, and `hasattr` in the tests for the `form_module_handler` context hook. I've kept the mock of `get_streamfields` as it seems a decent way to inject only the field we want to test with into the handler.

This change also removes several superfluous tests that appear to have been built around the mocking.

This should fix some recursion errors we're seeing in backend tests since Python 3.8.

## Checklist


higs4281 (approved):
works for me without coverage loss.

I'm not following why we were mocking 'hasattr' at all. The hooks tests succeeded without them, except for the one assertion for what `hasattr` was called with.

PR #6844, opened 2022-01-25
Add script to dump the local database

willbarton:
This change adds a script to make it easy to dump the local database. This is intended to be part of the workflow for creating and updating test data to use in the functional tests now that #6400 is. merged.

A database dump for the functional tests, and the process of creating and updating its contents will be documented separately.

The roughly intended usage is:

```shell
./dump-data.sh
./refresh-data.sh test.sql.gz
```

## Checklist


higs4281 (approved):
handy!

PR #6845, opened 2022-01-25
Update the search index as part of refresh-data.sh

willbarton:
This change adds `./cfgov/manage.py search_index --rebuild -f` to our `refresh-data.sh` script so that we update the search indexes with the contents of the newly refreshed data when we refresh it.

This also moves the command out of the `docker-entrypoint.sh` script that gets run in our Python container. That script includes a call to `refresh-data.sh`.

## Checklist

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -23,8 +23,6 @@ if ! psql ${DATABASE_URL} -c 'SELECT COUNT(*) FROM auth_user' &> /dev/null; then
     fi
     echo "Create the cache table..."
     ./cfgov/manage.py createcachetable
-    echo "Update the search indexes..."
-    ./cfgov/manage.py search_index --rebuild -f --parallel
 fi
 
 # Do first-time build of the front-end if necessary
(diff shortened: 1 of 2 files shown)


higs4281 (commented):
This works great, but would be a pain for me. I reload the database a lot when changing contexts (e.g. moving from or to a PR that runs migrations) or when getting fresh a new dump, which I do most days. But I rarely need or want to rebuild all the indices. 

It makes sense as a default for our docker process, but an option to not reindex would be nice.

With the latest download available, refresh currently takes me 1:25. With indexing, it's 3:14.

willbarton:
@higs4281 give that a try?

higs4281:
Sorry, but it's not working for me.

If I use `--noindex` before the dump name, or with no dump name, `--noindex` is read as the dump name and fails

When I run `./refresh-data.sh test  DUMPNAME --noindex`, the dump is loaded and indexing runs.

higs4281:
STRIKE THAT! I hadn't refreshed.

higs4281 (approved):
Thanks for the adjustment. 

I've heard many complaints about having to remember the indexing command, and this helps.

PR #6851, opened 2022-02-02
Removes unneeded external site gulp task

wpears:
Also, renames & moves the external-site js file to clarify that it isn't auto-included at a given route by base.html

Testing:
 - Pull and build
 - Navigate to any blog page and click on one of the social share icons
 - The external site page should function as normal


cfgov/jinja2/v1/external-site/index.html
```diff
@@ -24,7 +24,7 @@
 
 {% block javascript %}
 <script type='text/javascript'>
-    {% include '/js/external-site.js' %}
+    {% include '/js/routes/external-site.js' %}
```
anselmbradford, line 27: This whole /routes/ directory was set up back in the Wordpress days to mimic Wordpress' single and index page hierarchy. At some point `/routes/on-demand/` was added, which doesn't fit the paradigm of the other directories in `/routes/`. To not add another directory that doesn't work like the others, how about instead of moving `external-site.js` to routes, move it to `/js/modules/external-site.js`. You could also put it in a parent directory if it makes more sense:

```
/js/modules/external-site/external-site.js
/js/modules/external-site/ExternalSite.js
```

Or those two files could even be combined at some point.

wpears: This doesn't add a new directory, it just moves the `routes/external-site/index.js` file to `routes/external-site.js`, which is already a place the build looks when running the `scripts` task.

I think it makes sense to combine the files though, I'll do that.

anselmbradford: Yea but the scripts in `/routes/` have special meaning inside the base.html template, (see https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/jinja2/v1/_layouts/base.html#L328 for `index.js`, for instance). If it's not named `single.js` or `index.js` (or is not in the `on-demand` directory), I would not include it in `/routes/` at all and move it to `/modules/`.

wpears: `scriptsModern` builds js files in `/js/routes/` and subdirs, which is why I think `external-site.js` should go there. Nothing from `/js/modules` is built in gulp except as dependencies, so anything in there needs something in routes anyway (this explains the initial division between the current external site module and `/route/external-site/index.js` files).

Also, the magic behavior of `base.html` doesn't matter in this case because [external-site/index.html overrides the javascript block](https://github.com/cfpb/consumerfinance.gov/pull/6851/files#diff-cd863e989c0360f0219650ffbebd2b166d47c04f8e1cefc96d10b0d710dc0d1eR27)

anselmbradford: Mmmmm I gotcha. I forgot `scriptsModern` doesn't reach into anything but routes. `/apps/` has scripts that are included in a custom template, like external site is doing, but I can see why you wouldn't want the extra infrastructure for this. Okay okay sounds good, thank you for humoring me! :D

anselmbradford (approved):

PR #6852, opened 2022-02-02
Update usa tagline text

wpears:
This addresses concerns from https://github.com/uswds/uswds/issues/4419 that the current copy can be confusing/inaccessible.

Also, our atom is extra confusing on screenreaders due to the text being split between the wrapping div and a span.

This updates our atom to the suggested text.

**Testing**:
 - Pull and run `yarn run gulp styles`
 - Tests should pass and the tagline in the global eyebrow should be updated everywhere.
 - Global eyebrow should be visible in the mobile view, with just the tagline displayed
 - Upon opening the mobile mega-menu, the languages and contact number should be visible at the bottom, but not the tagline. 

Screen Shot 2022-02-02 at 7 08 38 PM


brandon-rusk (approved):
LGTM -- Thanks for taking care of this, Wyatt!

chosak:
When run locally this now loses the multi-language links on mobile:

|Before|After|
|-|-|
|image|image|

wpears:
Rearranged some things so that the tagline is always visible on mobile, but not shown in the mega-menu.

brandon-rusk:
Most of this is looking good, however there is a bit of a gap between the tagline and wrapper holding the mega menu/cfpb icon.

image

cfgov/jinja2/v1/_includes/atoms/tagline.html
```diff
@@ -15,8 +15,7 @@
 <div class="a-tagline{{ ' a-tagline__large' if is_large else '' }}">
 	<span class="u-usa-flag"></span>
    <div class="a-tagline_text">
-      {{ _('An official website of the') }}
-      <span class="u-nowrap">{{ _('United States government') }}</span>
+      {{ _('This is a United States government website') }}
```
anselmbradford, line 18: What's the thinking behind abandoning preventing US gov from wrapping? I thought that was something we didn't want to happen. It's easier if we don't care how it wraps, but just checking on whether we want it that way.

brandon-rusk:
I did accessibility testing using Chrome mobile emulator and iOS on CFPB iPhone -- This resolves the accessibility bug in platform issue 4219.

cfgov/jinja2/v1/_includes/atoms/tagline.html
wpears: This is a good question! When a screenreader reads that bit, it inserts a pause between the div and the span, which kinda sounds funky.

Moreover, `This is a United States government website` at the `large` tagline size in the footer doesn't break at 375px which is on the lower end of common viewport widths. This is essentially a trade off between screenreader performance and how that `large` tagline looks in the footer on especially small phones.

chosak:
@wpears should this be closed pending future decision on changing the tagline text?

PR #6854, opened 2022-02-02
Update Python 2 super() calls to Python 3

chosak:
In Python 2, it was necessary to specify the super type. This is no longer necessary in Python 3, but we have numerous examples of the old approach that were never migrated. This commit does so.

## Checklist


higs4281 (approved):
The gardener's work is never done. Thank you!

willbarton (approved):

PR #6856, opened 2022-02-03
Add troubleshooting around gettext and dates

willbarton:
This adds a quick bit of troubleshooting help to our translation documentation around the version of gettext that Django invokes.

gettext below 0.20 [had an issue where a `POT-Creation-Date` field in the `.po` file would continue through into the `.mo` file](https://savannah.gnu.org/bugs/?59658#comment0). Django updates that date string every time `makemessages` is run. This will fail our validate-migrations check.

The fix is to ensure a version of gettext > 0.20.

## Checklist

### DIFF
--- a/docs/translation.md
+++ b/docs/translation.md
@@ -138,3 +138,13 @@ mystring = gettext_noop("Hello World!")
 ```jinja
 {{ _(mystring) }}
 ```
+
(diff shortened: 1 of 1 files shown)


wpears (approved):
Heroic work

PR #6858, opened 2022-02-03
Remove explicit class inheritance from object

chosak:
In Python 3, [it's no longer necessary](https://portingguide.readthedocs.io/en/latest/classes.html#new-style-classes) to define classes with an explicit inheritance from `object`.

This commit changes all examples of

```py
class Foo(object):
```

to the simpler and equivalent

```py
class Foo:
```

## Checklist


higs4281 (approved):
The gardener looks up at the coming clouds; moves on to the next row.

PR #6860, opened 2022-02-07
Highlight cards: make body optional

chosak:
This commit modifies the highlight cards on the new home page design so that the card body is optional.

FYI @anselmbradford.

## How to test this PR

Build frontend assets, run a local server, try adding highlight cards to the home page, publish, visit http://localhost:8000/?home_page_2021=True.

## Screenshots

|Card body is now optional|Highlights render without body|
|-|-|
|image|image|

## Notes and todos

Note that this violates the current documented highlight card pattern in the design system:

https://cfpb.github.io/design-system/patterns/cards#highlight-card

## Checklist


contolini (approved):
> Note that this violates the current documented highlight card pattern in the design system

If we move forward with this pattern we'll have to amend that.

PR #6864, opened 2022-02-08
Bump the ES_JAVA_OPTS config to 512m

higs4281:
The current config of 128m doesn't allow indexing of our
current consumer complaint data, which is needed
for data products development.

We could probably return to the smaller value once
we have more robust dev and staging environments available
that include kibana and complete indexes.


jslay-excella (approved):
lgtm, per mattermost discussion.

PR #6868, opened 2022-02-10
Remove unused page categories for Supervisory Guidance pages

schbetsy:
Since the [Compliance and Guidance section](https://www.consumerfinance.gov/policy-compliance/guidance/) was restructured in 2019, and the [Supervisory guidance filterable list ](https://www.consumerfinance.gov/policy-compliance/guidance/supervisory-guidance/) no longer uses the "Implementation resource" page type categories of "Compliance aid" and "Official guidance" for filtering, displaying, or organizing those doc detail pages, they can be removed from that list of document detail page types/categories. 

## Changes

- Remove the 'compliance aid' and 'official guidance' categories from filterable list code, as well as their parent page type ('implementation resource'), since it had no categories.
- Update a test to accommodate the change
- Data migration removes existing instances of the removed categories

## How to test this PR

1. Behavior of the [Supervisory guidance filterable list ](https://www.consumerfinance.gov/policy-compliance/guidance/supervisory-guidance/) should be completely unchanged.

## Notes and todos

- ~~Some pages still have the 'official guidance' category, even though it isn't displayed anywhere and isn't used for filtering. Before deploying this code to delete the categories, I'll run code to delete that category from the pages that have it (`CFGOVPageCategory.objects.filter(name='official-guidance').delete()`)~~
- Can I also remove `Implementation Resources` from the [`page_types` list](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/v1/util/ref.py#L31)? It seems safer to leave it than to remove it, since I wasn't sure whether removing it would affect "related posts" functionality in an unexpected way.

## Checklist


willbarton:
> Some pages still have the 'official guidance' category, even though it isn't displayed anywhere and isn't used for filtering. Before deploying this code to delete the categories, I'll run code to delete that category from the pages that have it (CFGOVPageCategory.objects.filter(name='official-guidance').delete())

Would it make sense to put this in a data migration? Or, it doesn't seem like there's any harm in that category hanging out in the database until we delete it, so it's okay if that deletion flows from prod in the opposite direction to the code?

schbetsy:
Ah, of course a data migration would be better. Thanks @willbarton I'll work on that. 

I forgot to mention: the reason it's important to remove the category instances is because if we don't, they show up as zombie categories in the Wagtail admin and prevent saving a page: 
Screen Shot 2022-02-11 at 10 57 23 AM

schbetsy:
@willbarton When you get a chance, could you review again? I added the data migration and the schema migration, and I think this is much closer to ready.

willbarton (approved):

PR #6874, opened 2022-02-11
Remove legacy webfont architecture from paying-for-college and retirement

anselmbradford:
The Paying for College disclosures was applying webfonts via the `font` CSS attribute. However, it needed to be `font-family` to actually apply:

Screen Shot 2022-02-11 at 1 35 44 PM

Screen Shot 2022-02-11 at 1 57 08 PM

So all the `font` rules are not applying, so we can just remove them as the app will be unchanged to how it is now.

## Removals

- Removes invalid `font` CSS rules from paying for college disclosures app.
- Removes legacy `@webfont-` variables and licensed-fonts font-face rules.

## How to test this PR

1. `./frontend.sh` and contact Bill H for paying for college disclosures test URL.
2. Compare /retirement/ to production and see that some things are bolder and some things are less bold.


anselmbradford:
I pushed up https://github.com/cfpb/consumerfinance.gov/pull/6874/commits/1566b9c4819f27c04fc36608b9e6c907b1d4b87a, which migrates the Avenir demi to regular avenir with a "bolder" setting. This is slightly less bold than the current implementation. I think it's worth it, maintenance-wise. 

Before:
Screen Shot 2022-02-11 at 2 08 25 PM

After:
Screen Shot 2022-02-11 at 2 08 19 PM

These changes are more pronounced on /retirement/, which sees some things getting bolder (also note the dollar amount numbers were not the same style as those in the chart before):

Before:
Screen Shot 2022-02-11 at 2 27 08 PM

After:
Screen Shot 2022-02-11 at 2 26 58 PM

wpears (approved):
I think these changes are reasonable. If future design review finds them lacking, we should fix the hierarchy throughout our fonts instead of having different font hierarchies defined in separate apps.

PR #6876, opened 2022-02-11
Added overlay when Global Search is open

brandon-rusk:
### Changes

Global search now has overlay when opened, and overlay properly works when switching between the mega menu and overlay.

image

Resolves 2820 in CFGOV platform


contolini (approved):
Works great 🎉

PR #6881, opened 2022-02-14
Remove old CCDB landing page template

niqjohnson:
Now that the [Consumer Complaint Database landing page](https://www.consumerfinance.gov/data-research/consumer-complaints/) is a standard Wagtail page with no special content, we can get rid of all the old templates and assets that powered the old, custom landing page. See GHE/CFGOV/platform/issues/4325 for more details.

@anselmbradford, @flacoman91: I deleted front-end assets for the old tile map kind of willy-nilly. Do you see anything I missed? Or anything I deleted that I shouldn't have?

@higs4281, @willbarton: I think I excised all the back-end bits we don't need any more, but let me know if you spot anything else.

---

## Removals

- `ComplaintLandingView`
- Old landing page images
- Old landing page template
- CSS that styled the old landing page
- JS that powered the tile map on the old landing page
- JS unit tests that tested the tile map
- `CCDB_TECHNICAL_ISSUES` feature flag

## Changes

- Re-enabled the CCDB functional tests and eliminated trivial tests of landing page links. ~~Right now they will pass with production data; I still need to add a CCDB landing page to the test database so they'll pass in GitHub actions (see todos below).~~

## How to test this PR

1. Run all the automated unit and functional tests. Everything should pass.
2. Fire up the site locally. Everything should still work, especially `/data-research/consumer-complaints/`.

## Notes and todos

~~One thing still left to do in this PR:~~

- ~~As part of this PR, I want to add a CCDB landing page to the [test database](https://github.com/cfpb/consumerfinance.gov/blob/main/test.sql.gz) so functional tests will pass in Github actions. A couple of other people are working on adding data to that test database now, though. I'm going to wait until their changes are in `main` so I don't create any conflicts.~~

Instead of doing that 👆 , I changed course and removed the parts of the functional tests that hit the landing page. Now that the landing page is a regular Wagtail page, those parts weren't actually testing anything interesting. Now the tests start on the CCDB search page.

And some post-PR todos:

- Once this PR is deployed, we can remove the `CCDB_TECHNICAL_ISSUES` feature flag from the database (just deleting the setting won't delete it from the database)
- Add the complaint data outage banner that was part of the old template to the Wagtail banner manager
- Move the CCDB search [view](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/legacy/views/complaint.py) and [tests](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/legacy/tests/views/test_complaint.py) out of the `legacy` app, since they aren't legacy any more now that they're no longer combined with the old CCDB landing page stuff.

## Checklist


higs4281 (changes requested):
I think we want to keep the `CCDB_TECHNICAL_ISSUES` flag. It's how we manually raise an alert on the complaint search page.

https://github.com/cfpb/ccdb5-api/blob/ea5c7f38e1b4aaea98b9c625b4b4c1941824c3c3/complaint_search/es_interface.py#L224

anselmbradford:
Search the codebase for `ccdb-landing-map` and there's a `transformIgnorePatterns` in `jest.config.js` that looks like it can go --> https://github.com/cfpb/consumerfinance.gov/blob/main/jest.config.js#L31-L33

niqjohnson:
> I think we want to keep the `CCDB_TECHNICAL_ISSUES` flag. It's how we raise an alert on the complaint search page.

Good call, @higs4281, I forgot the alert in the CCDB interface can be enabled manually by that flag in addition to being turned on automatically by the staleness checks.

The bummer here is that I'd like to make that banner on the CCDB landing page into one that's managed in our normal [Wagtail banner manager](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/v1/models/banners.py#L21), which doesn't use feature flags at all. So that means we'd have to turn on two things—the flag and the banner—if we ever run into data issues and want an alert to show up on all CCDB-related pages. That doesn't seem like too big of a pain to me. What do you think? Is there a better way to approach the bannering of CCDB pages (keeping in mind I'd like to keep the new CCDB landing page a standard sub-landing page with no custom code to maintain).

niqjohnson:
> Search the codebase for `ccdb-landing-map` and there's a `transformIgnorePatterns` in `jest.config.js` that looks like it can go --> https://github.com/cfpb/consumerfinance.gov/blob/main/jest.config.js#L31-L33

Thanks, @anselmbradford. Fixed in d253ce3.

niqjohnson:
> Thanks, @anselmbradford. Fixed in [d253ce3](https://github.com/cfpb/consumerfinance.gov/commit/d253ce362eb53bd66868c738a8e274f69e8ebc58).

Actually, @anselmbradford, I guess it's fixed in a55551e? Removing `transformIgnorePatterns ` wholesale caused unit tests to fail to run, so now it's just a blank array. Does that seem right? As is obvious by now, I don't actually understand what that Jest config is doing 😄 .

higs4281:
@niqjohnson 
The banner takes a regex argument, so couldn't we apply it to `/data-research/consumer-complaints/.+` to capture the landing page and any variation of the downstream search page?

niqjohnson:
> @niqjohnson The banner takes a regex argument, so couldn't we apply it to `/data-research/consumer-complaints/.+` to capture the landing page and any variation of the downstream search page? Or does the page have to be a Wagtail page?

@higs4281, yeah, right now banners in the Wagtail banner manager will [only show up on `CFGOVPage` pages](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/v1/models/base.py#L306). We have an old issue to expand  banner manager banners to any page (GHE/CFGOV/platform/issues/3742), but I don't see that getting picked up any time soon. In the interim I'm fine with having to toggle on two things if we want a stale data alert to show up on all CCDB related pages given how much code complexity that would eliminate—especially since I don't get the sense that this is an alert we need to show often (but I'm not sure if I'm wrong about that assumption). What do you think?

higs4281:
@niqjohnson 
I agree the two-banner solution seems acceptable until we get address expanding banners to any page.   
Banners have been relatively rare in my experience, and the warnings that the search page currently raises for either stale data or the flag appear right with the search results, which I think is more likely to be noticed than the thin page-topper.

niqjohnson:
@higs4281, sounds good. I reverted the removal of the `CCDB_TECHNICAL_ISSUES` flag in b81fc6b.

anselmbradford:
> Actually, @anselmbradford, I guess it's fixed in https://github.com/cfpb/consumerfinance.gov/commit/a55551eb5baf9e60797d1f086453996bb66a9123? Removing transformIgnorePatterns wholesale caused unit tests to fail to run, so now it's just a blank array. Does that seem right? As is obvious by now, I don't actually understand what that Jest config is doing 😄 .
 
Hmm very weird. I don't understand how the errors shown relate to removing `transformIgnorePatterns` but ok.

higs4281 (approved):

PR #6894, opened 2022-02-17
Navigate to rather than hard coding an Ask answer in our functional tests

willbarton:
This change removes the hard coded Ask answer URL in our funcitonal tests with navigation to the first answer on the Ask landing page.

This hard-coded URL is currently causing our Ask tests to fail in Jenkins because the answer at that URL was unpublished.

This will require portal topic tags to be added to our test data. This PR will fail the GHA functional test run until that's done.

## Checklist


jslay-excella (approved):

willbarton:
This fixes the Ask CFPB failure in Jenkins. We're seeing unrelated timeouts though. I'm going to merge this though since the problem it fixes is fixed.

PR #6896, opened 2022-02-18
Pin Jinja 2.11's MarkupSafe dependency

willbarton:
A new release of MarkupSafe is incompatible with how Jinja 2.11 uses it, however Jinja doesn't include an upper bound on its pin and pulls it in anyway. This change pins the last working version until we can either upgrade Jinja to > 3, or there's an upstream patch release.

## Checklist


Colin-Seifer (approved):
Change fixes pipeline issue required for tests to pass.

PR #6899, opened 2022-02-22
Apply cache tags to all filterable lists

willbarton:
This PR moves the cache tag implementation from the `Newsroom` to `FilterableListMixin`, and makes the tag the page’s slug. This enables cache tags for all filterable list pages. It also moves the cache purge for newsroom to the general filterable list page invalidation signal handler above to work for all filterable list pages.

While working on cache tagging, it also converts our `CCDBSearchView` into a more generic `CacheTaggedTemplateView` to add cache tags to what is otherwise a normal `TemplateView` defined in `urls.py`.

## How to test this PR

Pull the branch down locally (or use the dev Docker deploy) and run:

```
curl -sI http://localhost:8000/about-us/blog/
```

Observe the `Edge-Cache-Tag: blog` header. 

You can also check the RSS feed URL and other filterable lists and confirm they have `Edge-Cache-Tag` headers as well:

```
curl -sI http://localhost:8000/about-us/blog/feed/
curl -sI http://localhost:8000/about-us/newsroom/
```

You can also check that the CCDB gets the `complaints` cache tag:

```
curl -sI http://localhost:8000/data-research/consumer-complaints/search/
```

## Checklist


higs4281 (approved):
Brilliant. I feel like Gimli: I just couldn't see that far ahead.

Checking for filterable mixins on every page-publish or unpublish event seemed potentially burdensome, but unless we're doing some programmatic page updates, it's probably not so.

willbarton:
@higs4281 this feels like a natural evolution after you proved how well cache tags will work!

PR #6900, opened 2022-02-23
Add documents report

niqjohnson:
Adds a report with details of all the documents managed in Wagtail. See GHE/CFGOV/platform/issues/4339 for more info.

---

## Additions

- A documents report using [Wagtail's standard reports interface](https://docs.wagtail.org/en/stable/extending/adding_reports.html)
- Unit tests for the new documents report

## Changes

- Since this is our second custom report in Wagtail, I made the unit tests—including the name of the test file—a little more generic to cover all the reports

## How to test this PR

1. Load the latest dump of production data
2. Go to `/admin/reports/documents/`, where you should see a list of all the documents managed in Wagtail
3. Hit the download CSV button on that list to make sure the CSV download works and all the data looks good in the CSV

## Notes and todos

- I'm going to add line comments once I open this PR to a couple of things I wasn't quite sure about that I'm hoping you lovely reviewers can pay extra attention to.

## Checklist


cfgov/v1/views/reports.py
```diff
@@ -49,3 +60,49 @@ def get_filename(self):
...
+        "tags.names": "Tags",
+        "created_at": "Uploaded on",
+        "uploaded_by_user": "Uploaded by",
+    }
+
+    """TODO: Why must tags be processed here but not page metadata report?"""
+    custom_field_preprocess = {
+        "tags.names": {
+            "csv": process_tags
+        },
```
niqjohnson, line 94: There seems to be a difference with how `ReportView` and `PageReportView` handle `tags.names` in the CSV export. As far as I can tell, `PageReportView` spits out a nice comma-separated list for `tags.names`, while `ReportView` returns a queryset of tag names for it. I'm not 100% sure why that's the case—or if I'm misunderstanding something—but I do know using just `tags.names` for the CSV export like the page metadata report does on line 34 above doesn't work here—it puts a stringified queryset in that column of the CSV. To work around that, I add a simple `process_tags` function to preprocess the queryset of tag names into a comma-separated string. But if there's a better way to do that, I'll happily change (and if there's not, I'll remove the TODO comment).

cfgov/v1/tests/views/test_reports_view.py
```diff
@@ -0,0 +1,78 @@
...
+    def test_process_categories(self):
+        category_string = process_categories(self.blog_page.categories.all())
+        self.assertIn(
+            self.category.get_name_display(), category_string)
+        self.assertEqual(
+            self.category.get_name_display(), self.category_tuple[1])
+
+    def test_process_tags(self):
+        tag_name_queryset = self.document.tags.all().values_list(
+            "name", flat=True)
```
niqjohnson, line 51: This is related to the [question I had about what `tags.names` returns in `ReportView`](https://github.com/cfpb/consumerfinance.gov/pull/6900/files#r813255512). `tags.names` gives a queryset of the name of every tag applied to an object in  in `ReportView`, so I'm passing in the same kind of queryset in the test. Is that the best way to do this? Should I add a comment explaining what kind of queryset is expected?

cfgov/v1/views/reports.py
willbarton: This is a fascinating rabbit hole I've gone down — there' appears to be [nothing specific to the `PageReportView` around tags or/related models at all](https://github.com/wagtail/wagtail/blob/main/wagtail/admin/views/reports/base.py#L63-L75). So, I started looking at the mixins, since `ReportView` inherits the `custom_field_preprocess` from `SpreadsheetExportMixin`, but there's nothing there either.

However, then I realized, tags in our pages come from `CFGOVPage`, [which uses `CusterableTagManager` for `tags`](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/v1/models/base.py#L67-L68), which comes from django-modelcluster. Wagtail's `Document` [uses django-taggit's `TaggableManager `](https://github.com/wagtail/wagtail/blob/e2b30c1690fb231abbb2e896e5553119db24cf8b/wagtail/documents/models.py#L38).

I suspect this related object manager being different is part of the issue. Both of these have different implementations of `value_from_object`: [taggit's](https://github.com/jazzband/django-taggit/blob/master/taggit/managers.py#L559-L565) vs [modelcluster](https://github.com/wagtail/django-modelcluster/blob/main/modelcluster/contrib/taggit.py#L97-L105). I'm not familiar enough with django-taggit to know quite the impact of these differences right off, but that's my suspicion. It's a long way to say, what you've got here is probably fine — however frustrating the difference between pages and documents.

cfgov/v1/tests/views/test_reports_view.py
willbarton: I don't think this needs a comment — this follows Django convention as far as I'm aware of it. The only thing is that `all()` should be unnecessary. 
```suggestion
        tag_name_queryset = self.document.tags.values_list(
            "name", flat=True)
```

willbarton (approved):

cfgov/v1/tests/views/test_reports_view.py
niqjohnson: Removed the `.all()` in ef34cdb

cfgov/v1/views/reports.py
niqjohnson: Thanks for going down that rabbit hole, @willbarton. I removed the TODO in ef34cdb.

PR #6905, opened 2022-02-25
Add validation to live streaming fields

### LINKED ISSUE #1446: Streaming selection should require a Go Live Date
Currently, you can select "Streaming?", but not a "Go Live Date", as shown:

screen shot 2016-02-11 at 1 59 13 pm

and successfully save the page, but it will error out when you try to view the page because there's no Go Live Date. We should add validation here.

Additionally, though it doesn't error out, we should require a URL as well, right? @kave

jslay-excella:
Resolves #1446

---

## Changes

* Defaults Live Stream Start Date to Event Start Date if Live Stream Start Date is not defined.
* Requires Live Stream Start Date is not on or after Event End Date if Event End Date is defined.

## How to test this PR

1. Check "Streaming" for all test cases
2. Don't define Live Stream Start Date, save page, ensure Live Stream Start date matches Event Start Date
3. Define event end date and start date after event end date, ensure validation error

## Screenshots
image

## Checklist


chosak:
> Requires a Live Video ID if "Streaming" is checked

What about cases where we want to post about an upcoming event where we don't yet have the YouTube livestream URL? I imagine we might schedule and post about events well in advance of setting up the streaming. Currently if you create a new event page with streaming enabled but don't provide a video ID, it looks like this:

image

It seems reasonable to me that we might want to support this user experience where we let people know there will be some streaming link in the future, even if internally we haven't set it yet.

> If no live start date is defined, should it just take the start date of the event itself?

I'm less sure about this. Pinging @csebianlander @ceceliatroyan who may have more experience with creating these pages and so might be more familiar with the typical use cases.

jslay-excella:
> > Requires a Live Video ID if "Streaming" is checked
> 
> What about cases where we want to post about an upcoming event where we don't yet have the YouTube livestream URL? I imagine we might schedule and post about events well in advance of setting up the streaming. Currently if you create a new event page with streaming enabled but don't provide a video ID, it looks like this:
> 
> image
> 
> It seems reasonable to me that we might want to support this user experience where we let people know there will be some streaming link in the future, even if internally we haven't set it yet.
> 
> > If no live start date is defined, should it just take the start date of the event itself?
> 
> I'm less sure about this. Pinging @csebianlander @ceceliatroyan who may have more experience with creating these pages and so might be more familiar with the typical use cases.

Good points! I am not familiar with how this is used entirely myself. I am more than welcome to feedback from some editors. Just wanted to get some changes in and a PR rolling to start trying to resolve the issue.

csebianlander:
@chosak is right re: how we might use these pages to preview a livestream without having a link ready yet -- in my experience we don't tend to get the YouTube links until pretty close up to the event, just because that's not always happening in parallel with other event planning activities.

The other two validation checks described here seem OK to me though, and

> If no live start date is defined, should it just take the start date of the event itself?

I'd say yes to this, assuming such a pre-population can be overriden if it turns out not to be quite right later on? (Livestreams often start ahead of the actual "event start" time, but the exact logistics and timing may not be known to us until day-of, and a placeholder of the datetime used for start time seems very reasonable to me)

jslay-excella:
Pushed changes. Updated room comment for PR. Ready for full review and merge on approval. 

@csebianlander @chosak

csebianlander (approved):

PR #6906, opened 2022-02-25
Add "content owner" field and export capability

csebianlander:
As part of the M&M sprint, we determined that it would be useful for Wagtail to have a "content owner" field so that we can track which Bureau division owns content, for easier reporting and maintenance in the future (see https://github.local/CFPB/el-camino/issues/559). Right now this kind of process is difficult and entirely manual.

This PR adds a "Content Owner" field to all Wagtail pages. The field is not used in any way for rendering the page; it exists solely for internal tracking and maintenance purposes. To assist in that tracking, I've also added a column for it in the "Page Metadata" report, to make it easy to export and filter based on that information.

---

## Additions

- A comma-separated field labeled "Content Owners" on all Wagtail pages, to be used similarly to the Authors and Topic Tags fields.
- A column in the "Page Metadata" report with the above field, so that the information can be exported for reporting use.

## How to test this PR

1. ./frontend.sh
2. localhost
3. Edit any Wagtail page and navigate to the "Configuration" tab; find the new "Content Owner" field.
4. Once data has been added to that field on any pages, navigate to Reports -> Page Metadata
5. Click "Download CSV"
6. In the resulting CSV, note the existence of the "Content Owner(s)" field and any entered data

## Screenshots

Screen Shot 2022-02-25 at 7 33 36 AM

## Checklist

#### Browser testing

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Safari on iOS
- [ ] Chrome on Android

#### Accessibility

- [ ] Screen reader friendly

#### Other

- [ ] Is useable without CSS
- [ ] Is useable without JS


csebianlander:
@schbetsy tagged you here for review since @chosak pointed out to me that the "author byline" work might have some functional overlap here; I don't think it really will but feel free to chime in with thoughts!

cfgov/v1/views/reports.py
```diff
@@ -41,7 +42,8 @@ class PageMetadataReportView(PageReportView):
         ("language", "Language"),
         ("search_description", "Search description"),
         ("tags.names", "Tags"),
-        ("categories.all", "Categories")],
+        ("categories.all", "Categories"),
+        ("content_owners.names", "Content Owner(s)")],
```
chosak, line 46: I think for database query optimization purposes (and consistency) it would be good to add `"content_owners"` to the `prefetch_related()` call below on line 64. This tells Django to do a single query to prefetch the list of content owners for all of the pages, as opposed to having to do one query per page. [Some background reading](https://scoutapm.com/blog/django-and-the-n1-queries-problem) if you're interested.

cfgov/v1/models/base.py
```diff
@@ -50,6 +50,14 @@ class Meta:
         verbose_name_plural = _("Tags")
 
 
+class CFGOVOwnedPages(TaggedItemBase):
```
chosak, line 53: Inheriting from `TaggedItemBase` means that this field will reuse the same tag model as page categories and authors; this means that autocomplete will populate from those things, where ideally we'd want this to autocomplete only from other content owners. Could we consider creating a [custom tag model](https://docs.wagtail.org/en/stable/reference/pages/model_recipes.html#custom-tag-models) for this?

csebianlander: Up for it but not certain whether it's something feasible to integrate easily into this PR vs. modifying in a later branch?

chosak: It'd be better to do it here (if you're up for it) because changing the tag model means moving to a different database table, so it'll avoid having to do some kind of data migration later.

csebianlander:
@chosak My latest commit attempts to replicate the code in Wagtail's recipes page for a custom tag model. It works properly up until a point: the new field doesn't seem to pull in autocomplete results from either of the other fields. However, right now if you try to add owner(s) to any given page and save or publish, you'll get an IntegrityError that looks something like this:

localhost_8000_admin_pages_319_edit_

I've tried a few things but frankly am at a bit of a loss; I'm not familiar enough with the codebase to understand what to try next. Thoughts/suggestions?

cfgov/v1/models/base.py
```diff
@@ -49,6 +49,18 @@ class Meta:
         verbose_name = _("Tag")
         verbose_name_plural = _("Tags")
 
+class CFGOVContentOwner(TagBase):
+    content_object = ParentalKey('CFGOVPage')
+
```
chosak, line 54:
```suggestion
```

@csebianlander I think this line is the problem -- compare against [the example in the Wagtail docs](https://docs.wagtail.org/en/stable/reference/pages/model_recipes.html#custom-tag-models) and you'll see that specifying the `content_object` is needed not on the tag class itself but on the "through model". I tried deleting this and recreating the migration and seem to be able to set the content owners okay, with autocomplete working as expected:

image

csebianlander: Ahh! I'd added that in at some point in attempting to debug and forgot to remove it to try to maintain the structure of the original code. This works great! Will re-commit and turn this back into a proper PR once I get lint and the other tests to pass locally.

csebianlander:
@chosak This should be good to merge upon your final review!

chosak (approved):

PR #6917, opened 2022-03-03
Add 'before_bulk_action' hook to disable delete

jslay-excella:
Adds Wagtail hook to prevent bulk delete of objects. 

## How to test this PR

1. Create a few pages
2. Attempt to delete multiple pages using bulk actions

## Screenshots

image
image


jslay-excella:
Added some cheap tests just to have *some* coverage. However, these tests do not call the hook, but the function directly. These tests, ideally, would call the actual Wagtail hook and let wagtail run the function.

jslay-excella:
Added proper hook tests via wagtail recommended documentation.

willbarton (approved):

PR #6923, opened 2022-03-04
Stub Census API in functional tests

niqjohnson:
Our functional tests for the [rural and underserved areas tool](https://www.consumerfinance.gov/rural-or-underserved-tool/) rely on a couple of Census APIs to work. That's caused problems recently: the Census APIs are occasionally slow, which causes these functional tests to fail on pull requests that have nothing to do with the rural and underserved areas tool. We want to mock those requests instead so we're testing the functionality of the page instead of the speed of the Census API

---

## Additions

- Fixture for Census API response data
- Cypress intercepts for the Census API calls

## Changes

- Minor changes to test descriptions to make them more accurate

## How to test this PR

1.  Start Cypress with `yarn run cypress open --config blockHosts="*.census.gov"`. That config option will block requests to any census.gov URL from the Cypress test environment so you can be sure the intercepts are working.
2. Run the tests for the rural and underserved areas tool (integration tests > consumer-tools > rural-or-underserved-tool.js).
3. Verify that the tests pass and that all the calls to the Census API are intercepted and stubbed (see screenshot below)

## Screenshots

ruat-tests

## Notes and todos

- The tool needs to make four different calls to two different Census APIs to get all the data it needs, and the API responses come back as JSONP, none of which Cypress handles super well. So I arranged things in the least ugly way I could think of, but I'm happy to rework that if anyone has a better way in mind.

## Checklist


test/cypress/integration/consumer-tools/rural-or-underserved-tool/rural-or-underserved-tool.js
```diff
@@ -6,9 +6,10 @@ describe( 'Rural or Underserved Tool', () => {
 
   describe( 'Address Search', () => {
 
-    it( 'Should render a table for a series of addresses searched', () => {
+    it( 'Should render a table for a single address', () => {
       page.open();
       page.searchAddress( '1600 Pennsylvania Avenue Washington DC' );
+      page.interceptCensusAPIRequests();
```
anselmbradford, line 12: All the `.as(…)` calls could be waited for in the tests where the API call is expected. E.g. something like:

```
page.open();
page.interceptCensusAPIRequests();
page.searchAddress( '1600 Pennsylvania Avenue Washington DC' );
// wait for API calls…
cy.wait('@censusGeocoding')
cy.wait('@censusTigerweb1')
cy.wait('@censusTigerweb2')
cy.wait('@censusTigerweb3')
// Or whatever order the calls happen in.
```

niqjohnson: @anselmbradford, I was considering waiting like that, but I wasn't sure about our general approach to using `cy.wait`. There are some waits scattered here and there throughout our tests, so it doesn't seem like we've had a consistent way of waiting before. In this case, since [Cypress says stubbed responses usually return in 20ms or so](https://docs.cypress.io/guides/guides/network-requests#Stub-Responses), I didn't put in explicit waiting; all four of those responses should come back well within the standard wait time. But I'm happy to add them in if we want to move to explicit waiting like that in our tests.

anselmbradford: As I understand it waiting for a request call is fine, and would test that the requests are happening at the right times. Waiting for an arbitrary amount of time is an anti-pattern, but that's not what this is. Generally if I have been intercepting a request and giving it a name, I would expect to wait for it somewhere else (otherwise, why does it have a name?)

niqjohnson: @anselmbradford, I added the waits in 1153b19, although I put them in the helper instead of the main test to keep all the intercept stuff together (speaking of things we do differently among all the tests, we're all over the place on what goes into a helper or not).

Even if we aren't waiting for intercepted requests, I still like giving them an alias with `.as()` to get the nice little pills showing up in the Cypress UI's test body output, like in the screenshot above.

anselmbradford: > Even if we aren't waiting for intercepted requests, I still like giving them an alias with .as() to get the nice little pills showing up in the Cypress UI's test body output, like in the screenshot above.

Ah good point!

anselmbradford (approved):

PR #6927, opened 2022-03-07
Migrate redirects to wagtail

csebianlander:
We can and do manage static redirects within Wagtail, and the usage of batch import/export tools within Wagtail allow this to be an exclusively content management-owned process. However, the existence of a _significant_ quantity of legacy redirects within the redirects.conf file makes it difficult to reliably ascertain whether redirects have been set up properly, and make it harder to reconcile redirect chains and other conflicts or issues with potential SEO and engagement implications.

Prior to this PR, I have already integrated every permanent, non-wildcard-based redirect in the redirects.conf file into Wagtail via batch import. When this PR is merged and deployed, all the redirects removed will still be active via their Wagtail counterpart.

---

## Removals

- All redirects that can be replicated in Wagtail. Note that some of the redirects in the code exceed Wagtail's 255-character limit; these have been omitted from Wagtail and retained in a new section in the redirects.conf file.

## How to test this PR

1. localhost
2. Sync local database with latest production database
3. Test redirects listed as removed from redirects.conf; they should still be functional

## Notes and todos

- Once this PR is merged, I'm going to take on further work to consolidate/clean up the redirects from within Wagtail, including redirect chains. This should not require any further PRs unless wildcard-based redirects are preferable to what exists.

## Checklist


willbarton:
This is awesome, @csebianlander!

cfgov/apache/conf.d/redirects.conf
```diff
@@ -37,13 +31,8 @@ RedirectMatch 404 \/yourstory\/.*\/[^$]
 # See discussion: https://GHE/eregs/eRegs-Notes/issues/2103#issuecomment-149183
 RedirectMatch  ^/eregulations/(.*)2015-26607_20180101(.*)$ /eregulations/$12017-18284_20180101$2
 
-# Ask CFPB
-Redirect permanent /askcfpb/askcfpb/js/ /static/askcfpb/js/
-Redirect permanent /askcfpb/askcfpb/img/ /static/askcfpb/img/
```
niqjohnson, line 42: It works fine, but it feels weird to have redirects for static asset paths living in the content management system now. No need to change anything here for now, but we might want to follow up to see why this redirect was needed and if there's a better way to do it now than a redirect.

cfgov/apache/conf.d/redirects.conf
```diff
@@ -228,24 +199,8 @@ RedirectMatch  (?i)^/hbd([/]?)$  /about-us/blog/four-years-working-for-you/
...
-Redirect permanent /gửi-tiền       /sending-money/vi
-Redirect permanent /perang-padala  /sending-money/tl
-Redirect permanent /transfe-lajan  /sending-money/ht
-Redirect permanent /transfè-lajan  /sending-money/ht
-
-Redirect permanent /blog/prepaid  /blog/prepaid-products-new-disclosures-to-help-you-compare-options
 RedirectMatch  (?i)^/HMDA  /data-research/hmda/
 RedirectMatch  permanent  (?i)^/yourstories  /your-story/?utm_source=print\&utm_medium=flyer\&utm_term=2015\&utm_campaign=TaxInsert
-Redirect permanent /regcomments  /notice-and-comment
-Redirect permanent /debtsurvey.html  /debtsurvey/
```
niqjohnson, line 244: I randomly spot checked this one, and `/debtsurvey/` 404s, so we can probably remove this redirect from Wagtail as a follow-up.

cfgov/apache/conf.d/redirects.conf
```diff
@@ -261,272 +216,17 @@ RedirectMatch permanent (?i)^/retirement/claiming-social-security/?$ /retirement
...
-Redirect permanent /mortgage-closing /owning-a-home/mortgage-closing/?utm_source=mortgageclosing&utm_medium=redirect&utm_campaign=OAHredirects
-Redirect permanent /owning-a-home/check-rates /owning-a-home/explore-rates/?utm_source=checkrates&utm_medium=redirect&utm_campaign=OAHredirects
 RedirectMatch permanent ^/owning-a-home/monthly-payment-worksheet[/]*$ /owning-a-home/resources/monthly_payment_worksheet.pdf
 
-## Redirecting old files stored directly on the server to standard document library locations
-Redirect permanent /owning-a-home/resources/adjustable_rate_loan_estimate.pdf /f/documents/cfpb_buying-a-house_adjustable-rate-loan-estimate.pdf
-Redirect permanent /owning-a-home/resources/checklist_mortgage_closing.pdf /f/documents/cfpb_buying-a-house_mortgage-closing_checklist.pdf
-Redirect permanent /owning-a-home/resources/monthly_payment_worksheet.pdf /f/documents/cfpb_buying-a-house_monthly-payment_worksheet.pdf
-Redirect permanent /owning-a-home/resources/mortgage_closing_forms.pdf /f/documents/cfpb_buying-a-house_closing-forms_guide.pdf
```
niqjohnson, line 468: There might be some redirect chains going on here, or it might have been something weird in my local environment, but none of these `/owning-a-home/resources/` redirects were working for me. They all ended up redirecting to `/static/owning-a-home/resources/[whatever]`. Not sure if that should block this PR or not since it's hard to tell what's environment issues and what's actual redirect issues.

cfgov/apache/conf.d/redirects.conf
```diff
@@ -1001,14 +488,9 @@ RedirectMatch permanent (?i)^/practitioner-resources/servicemembers/planning/?$
 RedirectMatch permanent (?i)^/practitioner-resources/servicemembers/planning/creativesavingsstrategies/?$ /consumer-tools/military-financial-lifecycle/
 RedirectMatch permanent (?i)^/practitioner-resources/servicemembers/protecting/?$ /consumer-tools/military-financial-lifecycle/
 
-# 9/30/20 IA work, practitioner-resources line moved here to allow the above redirects to maintain effect
-Redirect permanent /practitioner-resources/ /consumer-tools/educator-tools/
-
 # Common misspellings
 RedirectMatch (?i)(.*)morgage(.*) $1mortgage$2
 
-Redirect /assets https://files.consumerfinance.gov/a/assets
```
niqjohnson, line 1010: Same as the `/static/askcfpb/` comment above: it feels weird to have a redirect meant for static assets living in the CMS, but it seems to work, so probably not a reason to hold up this PR.

niqjohnson (commented):
It's so great getting all of these one-off redirects out of Apache and into Wagtail!

I looked over the diff of all the removed redirects and spot checked a handful locally. I left a couple of line comments about a few I saw that either weren't working as I'd expected or seemed odd to have in the CMS. I don't know that any of them are critical enough to hold up this PR, though. @chosak, what do you think?

Also, a couple of other things that struck me while looking this over:

- It's a bummer that Wagtail forces you to put a full, absolute URL in for the redirect target instead of the relative path like you can in Apache. That's mostly an issue in testing, though, as it unexpectedly jumped me from `localhost` to `www.consumerfinance.gov` all the time, but that doesn't seem like an issue in production. Also, when you're working on cleaning up all the Wagtail redirects, @csebianlander, you can always convert those "Redirect to any URL" redirects to "Redirect to a page" redirects if it makes sense.
- For the next round, I think we can get even more redirects out of Apache and into Wagtail. We have a bunch of `RedirectMatch` directives in a form like `RedirectMatch permanent  ^/independent-research-at-the-cfpb([/]?)  /blog/independent-research-at-the-cfpb/`. My regex is novice at best, but I think all those are doing are redirecting URLs with and without the trailing slash to the same place. I think we handle that trailing slash globally elsewhere, so I think all of those could be moved to simple Wagtail redirects, too. That sounds like a job for a subsequent PR to keep this one manageable, though.

cfgov/apache/conf.d/redirects.conf
csebianlander: Done!

cfgov/apache/conf.d/redirects.conf
csebianlander: Super weird. As far as I can tell the Wagtail-built redirects are set up properly (i.e. the "to" field URL resolves as it should). I can always cross-check these after the PR merges.

cfgov/apache/conf.d/redirects.conf
csebianlander: Agreed!

cfgov/apache/conf.d/redirects.conf
csebianlander: Agreed!

csebianlander:
@niqjohnson Totally agreed on your thoughts; re: the trailing slash, Wagtail's redirect functionality already seems to handle that just fine. Redirects in Wagtail _also_ appear to be case-insensitive, so I've already removed some redirects that were set up specifically to handle things like OWNING-A-HOME vs owning-a-home, etc.

Very much on board with setting up follow-up issues/items regarding some of those redirects that probably don't make sense to ultimately maintain via Wagtail since they're more about site/server assets than about content maintenance.

chosak:
One of the benefits of having redirects in the web server application (Apache) as opposed to the CMS (Wagtail) is that HTTP requests don't have to enter Python at all. A redirect handled by Apache returns right away without impacting the server hardly at all. A redirect handled by Wagtail not only has to invoke Python, load all of the Python code, run various Django middleware on the request, and then make a database query across the network to AWS to see if there is a matching redirect or not. I don't have a good quantitative comparison of how much overhead Python/Django/Wagtail/DB adds. Besides performance, Apache also gives us some control over redirect behavior that Wagtail doesn't currently. Relative redirects is one; wildcards and case-insensitivity is another.

For these reasons, I tend to think that redirects should be handled in Apache instead of in Wagtail where possible - certainly for long-lived ones which we how we want to keep forever-ish. @willbarton do you have any thoughts on the above?

@csebianlander you mention

> the existence of a significant quantity of legacy redirects within the redirects.conf file makes it difficult to reliably ascertain whether redirects have been set up properly, and make it harder to reconcile redirect chains and other conflicts or issues with potential SEO and engagement implication

I'd be curious how moving the redirects to Wagtail assists with this specifically - is it because when running locally in development using Django runserver you can't easily test the redirects if they are in Apache? Or because the Wagtail redirects interface provides an easier way to search/filter redirects?

csebianlander:
@chosak: both, but honestly I'm ultimately agnostic on where we store redirects. My opinion is just that we shouldn't be storing so many of the same types of redirects in _both_ locations as we do now. Also, things I feel are also true though a bit less strongly: a baseline expectation for content managers should not be that they are development-savvy enough to be able to access and modify the redirects.conf file; and the expectation should not be that a developer's time is required to set up any redirect on CF.gov.

willing to be swayed on both of those points to some degree, though, or at least the first one: I could certainly write documentation right now that our other content-ops team members could follow easily in terms of setting up static redirects by adding line(s) to the .conf file. I suppose since it's a single file, they wouldn't even need the typical dev-workflow setup -- just a github.com account and knowledge of how to edit a raw file directly via github and submit the edit as a PR.

my goal here is simply the consolidation of data; where it's consolidated matters less than that, when we want to pull it, we don't need to be doing so from multiple locations that store the information in slightly different ways.

willbarton:
> For these reasons, I tend to think that redirects should be handled in Apache instead of in Wagtail where possible - certainly for long-lived ones which we how we want to keep forever-ish. @willbarton do you have any thoughts on the above?

@chosak I've started writing a response a few times now, and I don't think I feel strongly either way, but I think it boils down to this: what do we want to optimize for? Do we optimize for consolidation and ease of management of our redirects, or do we optimize for on-server performance?

I think I come down on the side of optimizing for management of redirects. I think this approach is fine for now. And the Wagtail redirect model is [straight-forward enough](https://github.com/wagtail/wagtail/blob/main/wagtail/contrib/redirects/models.py#L10-L57) that if we change our mind in the future, programmatically extracting the redirects shouldn't be problematic.

cfgov/apache/conf.d/redirects.conf
```diff
@@ -80,12 +57,6 @@ deposit-accounts-resources/.*|\
 mortgage-resources/.*\
 ) /compliance/compliance-resources/$1
 
-Redirect permanent /policy-compliance/notice-opportunities-comment/ /rules-policy/notice-opportunities-comment/
-Redirect permanent /policy-compliance/amicus/ /compliance/amicus/
-Redirect permanent /policy-compliance/enforcement/ /enforcement/
-Redirect permanent /policy-compliance/rulemaking/ /rules-policy/
-Redirect permanent /policy-compliance/guidance/ /compliance/
```
schbetsy, line 87: I believe the redirects on line 83-87 were written to catch all of the sub-pages in each of these verticals. For instance, [`policy-compliance/guidance/amicus/`](https://www.consumerfinance.gov/policy-compliance/guidance/amicus/) redirects to `compliance/amicus/` because of this line.

Does you have a way of finding and accounting for all of the sub-pages that are affected by the pattern-matching in this file? (Sorry if I'm behind and you already addressed this.)

schbetsy (commented):
One question about pattern matching, and one request: if we do end up changing how we manage redirects, It would be good to update the "Redirects on cf.gov" page of the Platform wiki.

jslay-excella:
I am all for moving redirects into Wagtail. Having it run through application code to redirect is some additional compute, yes, but how often are we hitting these redirects? We should be writing our code in a manner where it isn't causing redirects (pointing to the correct location to begin with), so that the amount of redirects that are happening overall are minimized. The only static redirects that should really be going on, would be old permalinks and the like on external sites. 

I know from a management aspect, having it in Apache makes it a bit of a nightmare, especially when considering a move to something like nginx. We should maintain the flexibility to be able to use whatever wsgi is best at the time, and not be beholden to one or another. This is a big step forward in that direction.

csebianlander:
Having now worked with @chosak to successfully pull triggered redirects from our server logs, my preference is that this get merged ASAP so we can track and maintain actually used redirects via Wagtail. If we run into major issues with 404s caused by the removal of something here that is somehow not reflected in Wagtail.

If in the future our preference is to move redirects entirely out of Wagtail, it will not represent any significant level of effort; for now, because redirects are primarily owned by the Content Operations team, having them in Wagtail represents the cleanest workflow for keeping them consistently maintained.

Pinging @niqjohnson, @chosak for review/approval/merging once you've had another chance to review.

niqjohnson:
@csebianlander, did you see @schbetsy's [question about redirect directives that handle sub-pages, too](https://github.com/cfpb/consumerfinance.gov/pull/6927#discussion_r827072633https://github.com/cfpb/consumerfinance.gov/pull/6927#discussion_r827072633)? I *think* Wagtail only handles redirects from exactly one URL to exactly one other URL, but I'm not totally sure. I was hampered in trying to confirm that by Wagtail redirects apparently not being testable locally—it looks like any redirects you add will only apply to WWW?

redirect-sites

At any rate, probably worth confirming that (1) Wagtail indeed doesn't handle redirect of child pages if there's a redirect in place for their parent page and, if so, (2) we haven't removed any of those redirects from Apache. Offhand, it looks like a bunch of the redirects in the [redirects associated with the v1 project](https://github.com/cfpb/consumerfinance.gov/pull/6927/files#diff-5a86cd3a4885cbc3a1e2daeb615c616bc4f865b7eddd31a9a250f11fcabe1d4eL478) section were meant to handle child pages, but those are so old they might not even be relevant anymore. Not sure if there are others like those, though.

cfgov/apache/conf.d/redirects.conf
csebianlander: I've added these back in as part of the latest commit. As far as I can tell Wagtail has no functionality for this. I'm not too concerned if there are other stray ones that intend to do this that we've missed, though -- as mentioned elsewhere, using server logs I'm planning to regularly check for redirect usage and remove (or re-add) redirects as needed accordingly. If any affected by this show up in that process, I'll incorporate them accordingly.

csebianlander:
Wagtail definitely doesn't check child pages. For now, I've re-added the redirects that @schbetsy flagged into the .conf file. From here we'll monitor real-lfie usage and adjust accordingly. If it turns out that pattern-matching is a really highly desired feature since it gets a lot of use, that's a point to consider in favor of switching back to .conf as a way to maintain this list in the future.

For now, though, I think this should be all set to go once the checks are complete.

chosak (approved):

PR #6932, opened 2022-03-08
Add "Ignore topic tags" field to Related Posts

csebianlander:
As part of an intake request from the Office of Public Affairs, we may need to change the usage of the "Related posts" About Us sidebar in order to allow _all_ most recent blogs and newsroom items to appear within it.

Currently, our "Related posts" module can only show items that share either one or all of the same topic tags as the page itself. This PR adds a "Ignore topic tags" field to the module that instead ignores tags altogether when determining which pages to show, meaning that the X most recent pages of that type are shown no matter which tags are or are not shared between source page and related page.

---

## Additions

- "Ignore topic tags" field on Related posts module

## Changes

- Edits to allow test scripts to run properly

## How to test this PR

1. locahost
2. frontend.sh
3. On localhost, within Wagtail, modify the Sidebar tab of the "About Us" page by checking the box for "Ignore topic tags" under Related posts module
4. Publish page
5. View local copy of About Us page and compare related items listed to the listings on /blog and /newsroom

## Screenshots
Screen Shot 2022-03-08 at 2 16 14 PM

## Checklist


cfgov/v1/atomic_elements/organisms.py
```diff
@@ -271,6 +271,14 @@ class RelatedPosts(blocks.StructBlock):
...
     )
 
+    ignore_tags = blocks.BooleanBlock(
+        required=False,
+        default=False,
+        label="Ignore topic tags",
+        help_text=('If checked, related items shown will be pulled from the '
+                   'full set of posts, without filtering by topic tag.')
+    )
+
```
niqjohnson, line 281: I know this PR is still a draft, and I'm totally snooping, but because I was just poking around with related posts, this is on the top of my mind. I wonder if we should combine the option to ignore topic tags with the option to match all topic tags to make a single field that controls how topic tags are matched. Maybe a `<select>` with three options, like so (we can refine the labels, but this is the general idea):

- Related posts can match any topic tag assigned to the page
- Related posts must match all topic tags assigned to the page
- Related posts ignore topic tags assigned to the page

We use a similar kind of field in a few other spots in Wagtail (like h[ow to filter and sort topics in filter controls](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/v1/atomic_elements/organisms.py#L764-L778)), so it wouldn't be a new pattern. It would also make it clear that checking both the "match all topic tags" and "ignore topic tags" options are mutually exclusive. And, as an added bonus, it would be a good spot to explain that related posts are pulled in based on the topic tags assigned to the page the module is on, which is the reason I was poking around in the code in the first place (to [update the documentation to explain how the related posts are chosen](https://github.cfpb.gov/CFPB/hubcap/wiki/Related-Posts-(Sidebar)#notes)).

csebianlander: Snooping is great, no worries!

I like this idea and am going to modify the PR to try having it work accordingly. As two separate checkboxes the fields don't really work/cooperate intuitively, so this would definitely be a better user experiences.

csebianlander: Latest commit modifies both BooleanBlocks to instead use a ChoiceBlock for both fields.

The only issue I see with this, that I am not experienced enough to know how to handle, is that this fundamentally changes how the and_filtering value is set and therefore surely would require some sort of database modification to make sure pages that are using that value right now don't break?

chosak: Yes, this will require a data migration (for which we have some possibly-out-of-date documentation [here](https://cfpb.github.io/consumerfinance.gov/migrations/#data-migrations)) to modify any existing pages that have the existing field set.

I agree that this is a nicer change that adding an additional field. @csebianlander writing data migration can be a little tricky - I looked through the existing ones in the repo and [this](https://github.com/cfpb/consumerfinance.gov/blob/54dddc061b5f843a575a6887bb412a4033de8365/cfgov/v1/migrations/0243_fcm_body_plain_text.py) might be a close example to what we would need for this, if you want to take a stab at it.

csebianlander:
Latest commit includes my attempt to write a data migration script. It's even a bit trickier than I'd anticipated because 1) the old field is boolean, while the new field is a series of potential string values, and 2) I do not know the right syntax to call in and save the fields properly. The checks fail because python spits a syntax error back out, and so far in digging I can't figure out what the right syntax should be.

@chosak assistance would be appreciated -- I'll keep digging in the meantime!

chosak:
@csebianlander trying this out I wonder if (as @niqjohnson originally suggested above) the user experience might be more straightforward to have a required dropdown with three options. The current logic, where the default/unselected is "any tag", but then you can optionally select "no tags" or "all tags", feels unexpected. As a user I might expect that the default/unselected is "no tags", and then you could optionally select "any tags" or "all tags". Here's a mockup of what the three option version might look like:

Default:
image

Expanded:
image

I prototyped a commit and migration that seems to work for this approach (although you could also modify it to stick with your existing setup as well if you prefer), see https://github.com/cfpb/consumerfinance.gov/commit/1fe8f60fb079953abf33d1a3abd627aca0d9f2b5.

For testing I was using pages 3701 and 10504 as examples of existing any/all filtering, respectively.

This commit doesn't include the necessary changes to [these tests](https://github.com/cfpb/consumerfinance.gov/blob/157c5e9f68ccbab27e175da5dced0b83f03666e1/cfgov/v1/tests/atomic_elements/organisms/test_related_posts.py#L185-L224) that would need to happen. It would be good to add another test specifically for the "ignore" case; we could collaborate on that if you like.

niqjohnson:
@chosak, that three-item dropdown was what I had in mind. It's the same approach we take for other choice blocks, like when you choose an image width, where the field is required, all the possible choices are in there, and one is selected as a default:

image-widths

Feels like matching that approach works well here, too.

csebianlander:
@chosak - Merged your code and merged with latest updates from main; I can try to set up the test later today and you can review the code updates. At that point I think I'll shift this into a proper PR for merging once all looks good.

chosak (approved):
@csebianlander it seems like something went awry with the merge, so I've rebased this, and then fixed up the linter and the migration since they were recreated in #6954.

This looks good to me - I added the new test I was thinking of in the latest commit - you can see it explicitly tests to make sure that if "ignore" is set, all posts are returned.

With those changes, this LGTM - lmk if you want me to merge!

csebianlander:
Merge away when ready! Thanks so much for the help!

PR #6935, opened 2022-03-08
Border above mega-menu button not properly showing (mobile)

brandon-rusk:
## Changes

-Resolved border not showing properly on mobile mega-menu button

## How to test this PR

1. Open consumerfinance.gov on mobile and note the border above the mega-menu button

## Screenshots

Android:

image

iOS:

image

Resolves platform issue 4331

### DIFF
--- a/cfgov/unprocessed/css/organisms/header.less
+++ b/cfgov/unprocessed/css/organisms/header.less
@@ -116,7 +116,7 @@
         // Hide Global Header Call to Action.
         // Adjust Mobile Megamenu margin
         .respond-to-max( @bp-sm-max, {
-            margin-top: -6px;
+            margin-top: -5.5px;
             > .wrapper {
                 > .m-global-header-cta {
                     display: none;


brandon-rusk:
> Instead of moving the margin to half-pixels, could we add `border-top: 1px solid @gray-40;` above this line? --> https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov%2Funprocessed%2Fcss%2Forganisms%2Fmega-menu.less#L510 ?
> 
> Any issues with that?

I already tried that exact thing (except with `0.5px`) but the issue I am having is there is a `0.5px` discrepancy between Android and iOS in how they render. I have no idea why, but that's why I was not doing `1px`.

If I change it to `margin: -5px`, on Android the bars no longer connect, while it looks good on iOS. The current implementation I have solves it on both, but the line is a little thinner on iOS. I am open to suggestions

anselmbradford:
Another idea, could we stack the global eyebrow on top of the menu, by adding:
```
position: relative;
z-index: 100
```

Here https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov%2Funprocessed%2Fcss%2Fmolecules%2Fglobal-eyebrow.less#L86

That already has the border, but is just laying behind the menu. We'd just have to use a little caution of how high it is layered as to not go over modals. There is a modal on https://www.consumerfinance.gov/data-research/consumer-complaints/search/ at the "Things to know before you use this database" link.

wpears:
Ans' comment about z-index gave me the heebie jeebies so I found the root cause of this weirdness and fixed it in https://github.com/cfpb/consumerfinance.gov/pull/6940

PR #6943, opened 2022-03-10
Trim copy task hack, put in place to solve former relative reference

wpears:
Background: https://github.com/cfpb/consumerfinance.gov/pull/4119

The relative reference problem has since been removed from the design-system, so this hack is no longer needed.

Testing:
 - collectstatic should run without erroring


anselmbradford (approved):
🚀 🚀 🚀 

I ran `python cfgov/manage.py collectstatic` with no reported issues. Is this the way to test?

PR #6946, opened 2022-03-11
File a Complaint: Remove "File a Complaint" link from /complaint

mistergone:
The `/complaint/` page has a link at the top to itself. This causes some users to click through, expecting to see a different page, and becoming confused.

This PR adds a conditional in jinja to suppress the Complaint link at the top of the page **only** when viewing the `/complaint/` url.

## Additions
- Add conditional to `global-header-cta.html`

## How to test this PR
1. Pull it down and get it running.
2. Make sure the Complaint link at the top of the page doesn't appear on `/complaint/` but does appear on other pages.

## Screenshots
Screen Shot 2022-03-16 at 11 57 45 AM

## Checklist

### DIFF
--- a/cfgov/jinja2/v1/_includes/molecules/global-header-cta.html
+++ b/cfgov/jinja2/v1/_includes/molecules/global-header-cta.html
@@ -15,12 +15,14 @@
 
 
 {% macro render( language='en' ) %}
-    {% set complaint_link = '/es/enviar-una-queja/' if language == 'es' else '/complaint/' %}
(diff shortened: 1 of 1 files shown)


wpears (changes requested):
Can this be expanded to also suppress the link on `/es/enviar-una-queja/`?

mistergone:
> Can this be expanded to also suppress the link on `/es/enviar-una-queja/`?

Good catch! I've added a commit to address this.

wpears (approved):

PR #6948, opened 2022-03-14
Reduce model saves in data migrations

chosak:
Our [`migrate_page_types_and_fields`](https://github.com/cfpb/consumerfinance.gov/blob/54dddc061b5f843a575a6887bb412a4033de8365/cfgov/v1/util/migrations.py#L217) helper function for Wagtail page data migrations (documented [here](https://cfpb.github.io/consumerfinance.gov/migrations/#wagtail-specific-considerations)) takes a "mapper" function that gets applied to StreamField data to (potentially) alter it. Unfortunately currently this helper saves every page model instance and revision that it gets run on, *regardless of whether or not it actually changes any data*. This results in extra superfluous model saves, which slows down migrations and could introduce unexpected side effects.

This commit modifies one of our helper functions so that we only save the page or revision if it has actually made any changes to the data.

Users of our helper function specify the "path" to the block they want to alter, and the mapper that (potentially) modifies that block data. Consider for example a nested StreamField block like this, where the "bar" sub-field has one of a number of choices defined in code.

```py
{
    "type": "myblock",
    "value": {
        "foo": 1234,
        "bar": "choice_a",
    },
}
```

We want to write a data migration that removes all use of the specific choice "choice_a" and replace it with "choice_b". But we don't want to touch pages that have some other value, say, "choice_c". Under the current approach, those pages (and their revisions) are saved back to the database, even though their values haven't changed.

This commit attempts to determine whether a migration has actually modified a StreamField block, and only commit the page back to the database if so.

## How to test this PR

See new unit tests.

## Notes and todos

This code assumes that Python's [`copy.deepcopy`](https://docs.python.org/3/library/copy.html#copy.deepcopy) method is sufficient to copy the StreamField block value dict for comparison purposes. I believe this should be a safe assumption based on how StreamField data is stored.

## Checklist


willbarton (approved):
This is a really good catch!

PR #6949, opened 2022-03-14
Modify your-story index

csebianlander:
The "Everyone Has a Story" page and related content is being sunset. This PR modifies the index.html under /your-story to remove its sidebar content related to the to-be-sunset content. In order to keep the page visually balanced, it also moves the About Us content into the sidebar in place of the removed content.

---

## Removals

- "Everyone Has a Story" sidebar content

## Changes

- "About us" section moved to sidebar

## How to test this PR

1. frontend.sh
2. localhost
3. navigate to /your-story
4. Compare to live version (https://www.consumerfinance.gov/your-story)


chosak (commented):
Can the [Leah_thumbnail.png](https://github.com/cfpb/consumerfinance.gov/blob/2ace1532d9b24595db6b436f008f02cc7e03ae19/cfgov/legacy/static/nemo/_/img/Leah_thumbnail.png) and [Higinio_video_thumbnail.png](https://github.com/cfpb/consumerfinance.gov/blob/2ace1532d9b24595db6b436f008f02cc7e03ae19/cfgov/legacy/static/nemo/_/img/Higinio_video_thumbnail.png) images be deleted as part of this PR?

Also, unrelated to this PR, I notice when running Lighthouse against this page that it fails the accessibility check related to heading elements being in order because "Have a complaint about a financial product or service?" and "Have a money question?" are both `h3`s without a parent `h2`. Should these be changed to `h3 class="h2"` to avoid that?

csebianlander:
Re: the images, I think so but we should check to confirm they aren't still being used somewhere on CF.gov.

The code change to fix accessibility seems right, but am I misunderstanding -- shouldn't it be `h2 class="h3"` instead?

chosak:
@csebianlander yes, you're right! My brain was thinking that but I typed it the other way.

Re: the images, there are no other references in the codebase to these:

```sh
% git grep Leah_                               
cfgov/jinja2/v1/your-story/index.html:                <img src="{{ static( 'nemo/_/img/Leah_thumbnail.png') }}"
% git grep Higinio_
cfgov/jinja2/v1/your-story/index.html:                    <img src="{{ static( 'nemo/_/img/Higinio_video_thumbnail.png') }}"
```

I also checked the crawler and it turns out we have other copies of these in the database, as they appear on [Everyone has a story](https://www.consumerfinance.gov/consumer-tools/everyone-has-a-story/). But both of those are configured in Wagtail as the post preview images, and don't rely on the versions currently committed to this repo.

csebianlander:
Updated the PR to remove those images and update the `<h3>`s to `<h2>`s with the `h3` class. Hopefully good to go!

chosak (approved):

csebianlander:
@chosak I'm not authorized to, so would you mind merging this PR when you're able? Thanks!

PR #6950, opened 2022-03-14
Squash migrations

jslay-excella:
Squash migrations using the official method.


jslay-excella:
People added more migrations in the middle of a squash. This PR is pointless now.

PR #6952, opened 2022-03-14
Switch to bare jest

wpears:
This removes our wrapper around `jest` and moves config that was trapped in the gulp task to `jest.config.js`
Some idioms we use to test are changed (since we no longer are interacting with our wrapper, but the cli directly) which are reflected in the doc updates.

You'll also note some extraneous handlebars stuff has been excised! That's just for fun and as a gentle palette cleanser.. (I only removed it here because I became aware of it as the reason for passing `--no-cache` to jest. We don't need to do that anymore (speeding up jest) and moreover we can remove the code supporting the handlebars transformation).

Testing:
 - `yarn run jest` should run all unit tests
 - `yarn run jest Notification-spec.js` should run tests for the notification module
 - `yarn run jest organisms` should run tests for *all* organisms
 - `yarn run test` should run linting, then unit tests
 - `yarn run gulp build` should still run without complaint


willbarton:
Jesting bear

docs/javascript-unit-tests.md
```diff
@@ -39,34 +39,41 @@ it would be a good idea to peruse their docs before diving in here.
...
 
-To run a single test file, use the `--specs` flag to specify a file path:
+To run a single test file, pass the name (or path) of the spec:
 
 ```bash
-gulp test --specs=js/organisms/Footer-spec.js
+yarn run jest Notification-spec.js
+# Equivalent to:
+yarn run jest test/unit_tests/js/molecules/Notification-spec.js
+# The name argument would techniclly would match all Notification-spec.js files
```
contolini, line 48:
```suggestion
# The name argument would technically would match all Notification-spec.js files
```

contolini (approved):
🎉 You could also remove the extraneous `run` from every yarn command (e.g. `yarn jest` instead of `yarn run jest`) but nbd.

PR #6955, opened 2022-03-15
Editorconfig

jslay-excella:
[Brings in `.editorconfig` support](https://editorconfig.org/).

Also brings in a typo fix for migration docs (unrelated, but needed fixing), and was also used to test `.editorconfig`.


willbarton:
Could we add some brief documentation too, with maybe links to plugins for popular editors we have around here (at least VSCode and vim)?

wpears (approved):

PR #6960, opened 2022-03-16
Enforce direct document links to files.consumerfinance.gov

chosak:
Currently pages on consumerfinance.gov can link to Wagtail-managed documents in 2 different ways:

1. A document ID link, e.g. https://www.consumerfinance.gov/documents/10450/cfpb_msem_cobranding_instructions_2022.pdf - these documents/ID/filename URLs are rendered by Wagtail when e.g. you add a document to a rich text field.
2. A direct S3 URL link, e.g. https://files.consumerfinance.gov/f/documents/cfpb_msem_cobranding_instructions_2022.pdf - these are often entered directly into page text or where we don't use Wagtail rich text to link to documents.

Having two types of links makes it harder to track where documents are linked. This commit is intended to eliminate all links of the form `documents/ID/filename` in favor of always linking to S3 directly.

As it happens, Wagtail now supports this functionality out of the box. In Wagtail 2.7, the [`WAGTAILDOCS_SERVE_METHOD`](https://docs.wagtail.org/en/stable/reference/settings.html#wagtaildocs-serve-method) setting was introduced which allows configuring document links so they always point directly to the remote storage (in our case, S3).

It turns out that, way back when we were on Wagtail 1.6, we created a custom document serve view v1.views.documents.DocumentServeView that implemented some custom logic around document serving (essentially, when you hit a documents/ID/filename, it always redirects to S3. With the newer Wagtail setting, this view can be removed.

Fixes internal http://github.local/CFGOV/platform/issues/4364.

## How to test this PR

These steps assume you have a local production dump and a copy of (at least one of) the production documents locally in your cfgov/f/documents subfolder. You can also test by manually uploading a document to Wagtail and creating a page that links to it.

When running locally with the default setup, (`./runserver.sh`), document links on [a page like this](http://localhost:8000/language/tl/) point as expected to a localhost link like http://localhost:8000/f/documents/cfpb_how-to-avoid-foreclosure_tl.pdf (without an ID). But the existing link like http://localhost:8000/documents/7454/cfpb_how-to-avoid-foreclosure_tl.pdf (with the ID) continues to work, and redirects as expected to the link without the ID.

You can also run locally with an S3 configuration using a command like:

```
AWS_STORAGE_BUCKET_NAME=files.consumerfinance.gov \
AWS_ACCESS_KEY_ID=fake \
AWS_SECRET_ACCESS_KEY=fake \
AWS_S3_CUSTOM_DOMAIN=files.consumerfinance.gov \
AWS_S3_URL=https://files.consumerfinance.gov \
S3_ENABLED=True \
./runserver.sh
```

Visiting [a local page like this](http://localhost:8000/language/tl/) now has document links pointing directly to https://files.consumerfinance.gov/f/documents/cfpb_how-to-avoid-foreclosure_tl.pdf. And the existing link like http://localhost:8000/documents/7454/cfpb_how-to-avoid-foreclosure_tl.pdf redirects as desired to the same place.

## Notes and todos

As part of this change I updated our data.json URLs in the /documents/ID/filename form, as documented in our internal wiki at http://github.local/CFGOV/platform/wiki/Data.gov-and-the-data.json-file.

## Checklist


willbarton (approved):
Awesome!

PR #6961, opened 2022-03-17
Add fallback SVG Icon

jslay-excella:
Adds a fallback icon when an icon name that is provided, is not valid (not found, invalid SVG).

## How to test this PR

1. Create a page with an icon that doesn't exist.
2. Ensure the missing icon renders out the fallback icon.

## Screenshots
Incoming...


willbarton (commented):
f-strings, walruses, and type hints, oh my.

This looks great! I think it's worth getting someone to weigh in on whether `error` makes sense as the fallback icon to show to users. @wpears, do you have a feel for who that should be? Glancing through our [iconography](https://cfpb.github.io/design-system/foundation/iconography), I'm not sure there's any others that make any more sense, and missing images have traditionally had some form of an `x` to them. Before I approve, I'd just like some other eyes on that decision.

Code looks good to me though!

PR #6963, opened 2022-03-17
We should be making a 3.8 virtualenv, this does that

wpears:


willbarton (approved):
This is fine as is.

But I wonder if this script has outlived its usefulness. We [document a process using pyenv-virtualenv now](https://cfpb.github.io/consumerfinance.gov/installation/#set-up-a-local-python-environment-optional). virtualenvwrapper (where `mkvirtualenv` comes from) is a fairly antiquated way of making virtualenvs.

willbarton (approved):

PR #6964, opened 2022-03-17
Simple chart refactor

wpears:
Moves some bits of similar code into their own files and trims down some bits of complication/branching.

Testing:
 - Pull branch
 - Get latest wagtail data
 - Go to http://localhost:8000/admin/pages/15874/edit/preview/
 - Charts should look/function identically to the same page on content


brandon-rusk (approved):
Looks good to me, thanks for doing this refactoring!

PR #6968, opened 2022-03-21
Remove modernizr & pull old ie support

wpears:
With https://github.com/cfpb/design-system/pull/1347

From that PR:
>We ship css to all our users that is only targeting browsers that are sub-0.5% usage. This is wasteful and blocking for people on slow connections and mobile phones with slower processing speeds (which are a much greater proportion of our users that ie8, ie9, and ie10).

This matches the results of that PR, removing old ie classes. As well, this removes the machinery that both produces the old ie classes and the js/no-js + modernizr-specific classes on the html element.

Now the behavior can be described as follows:
 - The `html` tag gets a `no-js` class
 - If JavaScript is enabled and supports `window.Map` (modern browsers + ie11), replace this class with `js`
 - Else (if js is disabled or <ie11) serve the `no-js` experience

Testing:
 - Pull and build
 - Ensure pages load as normal
 - disable js in the dev-console
 - Ensure pages load as normal and that behavior is identical to www


anselmbradford (approved):
This looks good to me looking over the code, but I don't have cfgov running currently. Maybe @contolini can test it live.

PR #6974, opened 2022-03-23
Add images report

csebianlander:
Adds a report with details of all the images managed in Wagtail. 

---

## Additions

- An images report using [Wagtail's standard reports interface](https://docs.wagtail.org/en/stable/extending/adding_reports.html)

## How to test this PR

1. Load the latest dump of production data
2. Go to `/admin/reports/images/`, where you should see a list of all the images managed in Wagtail
3. Hit the download CSV button on that list to make sure the CSV download works and all the data looks good in the CSV

## Checklist

### DIFF
--- a/cfgov/v1/templates/v1/images_report.html
+++ b/cfgov/v1/templates/v1/images_report.html
@@ -15,13 +15,13 @@
             <thead>
                 <tr>
                     <th>
-                        ID
-                    </th>
-                    <th>
                         Title
                     </th>
                     <th>
                         File
+                    </th>
+                    <th>
+                        Size (bytes)
                     </th>
                     <th>
                         Collection
@@ -41,15 +41,15 @@
                 {% for image in object_list %}
                     <tr>
                         <td>
-                            {{ image.id }}
-                        </td>
-                        <td>
                             {{ image.title }}
                         </td>
                         <td>
-                            <a href="https://www.consumerfinance.gov{{ image.url }}">
-                                {{ image.filename }}
+                            <a href="{{ image.url }}">
+                                {{ image.url }}
                             </a>
+                        </td>
+                        <td>
+                            {{ image.file_size }}
                         </td>
                         <td>
                             {{ image.collection }}
--- a/cfgov/v1/templates/v1/images_report.html
+++ b/cfgov/v1/templates/v1/images_report.html
@@ -1,0 +1,73 @@
+{% extends 'wagtailadmin/reports/base_report.html' %}
+{% load i18n wagtailadmin_tags %}
+
+{% block actions %}
+    {% if view.list_export %}
+        <div class="actionbutton">
+            <a href="{{ view.csv_export_url }}" class="button bicolor button--icon">{% icon name="download" wrapped=1 %}{% trans 'Download CSV' %}</a>
+        </div>
+    {% endif %}
+{% endblock %}
+
+{% block results %}
+    {% if object_list %}
+        <table class="listing">
+            <thead>
+                <tr>
+                    <th>
+                        ID
+                    </th>
+                    <th>
+                        Title
+                    </th>
+                    <th>
+                        File
+                    </th>
+                    <th>
+                        Collection
+                    </th>
+                    <th>
+                        Tags
+                    </th>
+                    <th>
+                        Uploaded on
+                    </th>
+                    <th>
+                        Uploaded by
+                    </th>
+                </tr>
+            </thead>
+            <tbody>
+                {% for image in object_list %}
+                    <tr>
+                        <td>
+                            {{ image.id }}
+                        </td>
+                        <td>
+                            {{ image.title }}
+                        </td>
+                        <td>
+                            <a href="https://www.consumerfinance.gov{{ image.url }}">
+                                {{ image.filename }}
+                            </a>
+                        </td>
+                        <td>
+                            {{ image.collection }}
+                        </td>
+                        <td>
+                            {{ image.tags.names|join:", " }}
+                        </td>
+                        <td>
+                            {{ image.created_at }}
+                        </td>
+                        <td>
+                            {{ image.uploaded_by_user }}
+                        </td>
+                    </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+    {% else %}
+        <p>There are no images in the image library.</p>
+    {% endif %}
+{% endblock %}
--- a/cfgov/v1/views/reports.py
+++ b/cfgov/v1/views/reports.py
@@ -2,7 +2,7 @@ from datetime import date
 
 from wagtail.admin.views.reports import PageReportView, ReportView
 from wagtail.documents.models import Document
-from wagtail.images.models import Image
+from wagtail.images import get_image_model as Image
 
 from v1.models import CFGOVPage
 
@@ -18,9 +18,14 @@ def process_tags(queryset):
 
 
 def construct_absolute_url(url):
-    """Turn a relativey URL into an absolute URL"""
+    """Turn a relative URL into an absolute URL"""
     return "https://www.consumerfinance.gov" + url
 
+def construct_files_url(url):
+    """Convert Wagtail URL path to files.cf.gov absolute URL."""
+    path = str(url)
+    prefix = "https://files.consumerfinance.gov/f/documents/"
+    return prefix + path.split("/")[1]
 
 class PageMetadataReportView(PageReportView):
     header_icon = "doc-empty-inverse"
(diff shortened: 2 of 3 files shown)


csebianlander:
I'm keeping this as WIP because of a few things that I need other eyes on:
- @niqjohnson - can you look at this and see what edit(s) need to be made to the unit test?
- @chosak - can you review how this PR renders files.CF.gov URLs and let me know how that might be affected by [6960](https://github.com/cfpb/consumerfinance.gov/pull/6960)?

chosak:
> can you review how this PR renders files.CF.gov URLs and let me know how that might be affected by 6960

I would propose just using `image.file.url` here, implemented in [this commit](https://github.com/cfpb/consumerfinance.gov/commit/50191ff1f6123af2c819bd9b2a0d5ded74a15d1e). On localhost this will point to local files, on production it will point properly to S3. You can confirm this behavior locally by running with a command like

```
AWS_STORAGE_BUCKET_NAME=files.consumerfinance.gov \
AWS_ACCESS_KEY_ID=fake \
AWS_SECRET_ACCESS_KEY=fake \
AWS_S3_CUSTOM_DOMAIN=files.consumerfinance.gov \
AWS_S3_URL=https://files.consumerfinance.gov \
S3_ENABLED=True \
./runserver.sh
```

which will give you files.cf.gov URLs when running locally (but won't actually allow you to upload there).

cfgov/v1/templates/v1/images_report.html
```diff
@@ -0,0 +1,73 @@
+{% extends 'wagtailadmin/reports/base_report.html' %}
+{% load i18n wagtailadmin_tags %}
+
+{% block actions %}
+    {% if view.list_export %}
+        <div class="actionbutton">
+            <a href="{{ view.csv_export_url }}" class="button bicolor button--icon">{% icon name="download" wrapped=1 %}{% trans 'Download CSV' %}</a>
+        </div>
+    {% endif %}
+{% endblock %}
```
chosak, line 10: By default Wagtail offers both XLSX and CSV downloads for reports. Digging through history I see we originally removed the XLSX option from the page metadata report in #6060 (see internal platform#3881 for background). This same template code was copied for the document report in #6900, and again here.

1. Do we need/want to eliminate the XLSX option for the document and image reports?
2. If so, would it make sense to implement something like a `v1/reports/report_no_xlsx_base.html` base template that all of these report templates inherit from, to avoid the duplication?

niqjohnson: > If so, would it make sense to implement something like a `v1/reports/report_no_xlsx_base.html` base template that all of these report templates inherit from, to avoid the duplication?

@chosak, yeah, it's probably a good idea at this point to make a no-Excel base template and have all the other custom reports inherit from it.

We eliminated the Excel option originally because the page metadata report was so big it would crash the Excel export in production. I'm not sure if Wagtail has improved how it exports since then or what size is big enough to cause issues, though. In general the Excel export doesn't offer much over the CSV, so I'd be fine making a base template with just the CSV export and using that base template for all of our custom reports.

cfgov/v1/templates/v1/images_report.html
```diff
@@ -0,0 +1,73 @@
...
+            <thead>
+                <tr>
+                    <th>
+                        Title
+                    </th>
+                    <th>
+                        File
+                    </th>
+                    <th>
+                        Size (bytes)
```
niqjohnson, line 24: It might be worth it to do a little processing and spit out file sizes in KBs for easier readability instead.

cfgov/v1/templates/v1/images_report.html
```diff
@@ -0,0 +1,73 @@
...
+            <tbody>
+                {% for image in object_list %}
+                    <tr>
+                        <td>
+                            {{ image.title }}
+                        </td>
+                        <td>
+                            <a href="{{ image.url }}">
+                                {{ image.url }}
+                            </a>
```
niqjohnson, line 49: This might not make sense for images, but in the documents report I [combined the document's filename and URL into a single column](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/v1/templates/v1/documents_report.html#L50-L52). That's mostly because Wagtail annoyingly clips instead of scrolls tables that are too big horizontally, but also because I didn't get much extra information from a separate URL field in the Wagtail view for documents. The URL field is there in the documents CSV, where it serves more of a purpose. Not sure if that's a good pattern for the image report or not, though.

cfgov/v1/views/reports.py
```diff
@@ -106,3 +112,44 @@ def get_filename(self):
...
+    custom_field_preprocess = {
+        "tags.names": {"csv": process_tags},
+        "file": {"csv": construct_files_url},
+    }
+
+    template_name = "v1/images_report.html"
+
+    def get_filename(self):
+        """Get a better filename than the default 'spreadsheet-export'."""
+        return f"all-cfgov-images-{date.today()}"
```
niqjohnson, line 149: This might be outside the scope of this PR, but I wonder if we can reduce some duplication in all the `get_filename`s. Maybe instead of a custom name for each report we can just use the report's title in the filename, like `{self.title}-{date.today()}` (we'd probably want to lowercase and hyphenate the title in the filename, though). We could define that as a function up above and use it in all reports.

cfgov/v1/views/reports.py
```diff
@@ -106,3 +112,44 @@ def get_filename(self):
...
+
+    def get_queryset(self):
+        raw_set = Image().objects.all().order_by("-created_at").prefetch_related("tags")
+        for image in raw_set:
+            image.url = construct_files_url(image.file)
+        return raw_set
```
niqjohnson, line 155: I think you'll need a blank line here at the end of the file to satisfy the linters.

cfgov/v1/wagtail_hooks.py
```diff
@@ -37,7 +37,7 @@
     video_player_test_cases,
 )
 from v1.util import util
-from v1.views.reports import DocumentsReportView, PageMetadataReportView
+from v1.views.reports import DocumentsReportView, ImagesReportView, PageMetadataReportView
```
niqjohnson, line 40: I think the linter will complain about a long line length here. You can have black auto-format things before commits to make the linter happy:

>Install pre-commit via pip or pipx, either globally or into your virtrualenv (recommended).
>
>`pip install -U pre-commit`
>
>Once it is installed, from the root of consumerfinance.gov repo, run the following command to install the pre-commit hooks
>
>`pre-commit install`
>
>This will allow pre-commit to run black, flake8, and isort, on commits where files being changed are relevant.

niqjohnson (commented):
Added a couple of comments. Other than the linting stuff, it's all optional possible improvements. I'm going to defer to @chosak on the best approach to the [unit tests](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/v1/tests/views/test_reports_view.py). Those started off as pretty simple tests for the page metadata report, but now that we have three custom reports, we can probably make some improvements to how they're structured. What do you think, @chosak?

cfgov/v1/templates/v1/images_report.html
csebianlander: Agree that I don't see much need for an XLSX option. I can work on building a `v1/reports/report_no_xlsx_base.html` and adding it to this commit (including corresponding edits to other pages as needed).

cfgov/v1/views/reports.py
csebianlander: I don't think this is outside the scope -- I can try to implement this within this PR.

cfgov/v1/templates/v1/images_report.html
csebianlander: FYI the page metadata report inherits from a different base report, so the change to do this that I just committed only affects the documents report and this new image report.

cfgov/v1/templates/v1/images_report.html
csebianlander: I'd actually rather keep them in bytes at least for now, just to understand the usefulness of the field in the export and data review before doing too much extra work.

cfgov/v1/views/reports.py
```diff
@@ -101,8 +106,48 @@ class DocumentsReportView(ReportView):
...
+    }
+
+    template_name = "v1/images_report.html"
+
+    def get_filename(self):
+        return generate_filename("images")
+
+    def get_queryset(self):
+        return (
+            Image()
```
chosak, line 149: See above; someone reading this might interpret it as creating a new instance of the image class, which would be confusing.

```suggestion
            get_image_model()
```

csebianlander:
@chosak test fail corrected - was just a simple naming-convention change that needed to be reflected in the tests. This should be ready to merge, once all checks are complete.

cfgov/v1/templates/v1/images_report.html
```diff
@@ -0,0 +1,58 @@
...
+            <tbody>
+                {% for image in object_list %}
+                    <tr>
+                        <td>
+                            <a href="{{ image.file.url }}">
+                                {{ image.title }}
+                            </a>
+                        </td>
+                        <td>
+                            {{ image.kb }}
```
chosak, line 37:
```suggestion
                            {{ image.file_size | filesizeformat }}
```

PR #6978, opened 2022-03-26
Replace gulp with esbuild

wpears:
This PR replaces our existing build tool and task runner, gulp, with esbuild and plain yarn scripts.

Testing:
 - Pull branch
 - `yarn`
 - `rm -rf cfgov/static_built`
 - `yarn build`

The output of this build should be essentially identical to the existing build process, meaning you should be able to browse `localhost:8000` and visit all pages and Wagtail without problems.

This also removes a host of dependencies used by the previous build system, including webpack. I attempted to remove now-unnecessary deps where possible, but may have been too aggressive in some cases, do correct me as needed!

Another interesting change is removing individual app-level config (since this was usually webpack + browserslist). Although useful in theory, in practice these were almost identical across apps. For the usecase of using the app-level webpack.config.js file to run a script on build, I decided to just make the idiom calling that script directly from `build.js` (the only time this is used is by regs3k's service worker registration).


esbuild/copy.js
```diff
@@ -0,0 +1,47 @@
...
+  const staticFiles = files.filter( v => !v.match( /.js$|.less$|.css$/ ) );
+
+  const inDirs = [ ...new Set( staticFiles.map( v => dirname( v ) ) ) ];
+  const outDirs = [
+    ...inDirs.map( v => v.replace( unprocessed, baseConfig.outdir ) ),
+    // Create specific icon directory
+    `${ baseConfig.outdir }/icons`,
+    // Hande prebuilt lightbox dep
+    ...[ '', '/images', '/js', '/css' ]
+      .map( v => `${ baseConfig.outdir }/lightbox2${ v }` )
```
contolini, line 17: Per #6983 you no longer have to worry about all the `lightbox2` assets in this file.

contolini:
```
main $  time yarn gulp

real	2m47.594s

pr/6978 $  time yarn build

real	0m10.495s
```

But I use those 2.5 minutes to make tea. Now how am I going to stay hydrated throughout the day? This PR is so effective it's going to cause me physical harm.

Colin-Seifer (approved):

cl-esb.sh
```diff
@@ -0,0 +1,36 @@
...
+
+#js for specific pages, based on url
+"$r/about-us/careers/current-openings/index.js"
+"$r/consumer-tools/debt-collection/index.js"
+"$r/data-research/prepaid-accounts/search-agreements/index.js"
+"$r/owning-a-home/mortgage-estimate/index.js"
+"$r/owning-a-home/index.js"
+"$r/external-site/index.js"
+
+#on-demand: components included on a page via Wagtatil
```
anselmbradford, line 24: Tiny typo `Wagtatil`

wpears: trimmed that entire script :)

package.json
```diff
@@ -34,60 +32,30 @@
     "@cfpb/cfpb-tables": "0.14.1",
     "@cfpb/cfpb-typography": "0.14.1",
     "@cfpb/rental-assistance-finder": "0.4.0",
+    "@frappe/esbuild-plugin-postcss2": "0.1.3",
```
anselmbradford, line 35: Can this go in devDependencies?

package.json
```diff
@@ -15,12 +15,10 @@
   "engines": {
     "node": ">=16.x"
   },
-  "browserslist": "> 0.5%, last 2 versions, Firefox ESR, not dead, ie >= 11",
   "dependencies": {
-    "@babel/core": "7.16.0",
+    "@babel/core": "7.17.8",
```
anselmbradford, line 19: How is babel used? Is it still needed here?

package.json
anselmbradford: Oh right we need all the build things in `dependencies` because of how we build things outside of the local environment, right?

wpears: Yeah, the dep/devdep binary is kinda odd for a user-facing application.

package.json
wpears: Babel is still used by jest, I haven't fully explored possible options to trimming it, but I'd like to (probably in a separate PR).

anselmbradford:
We should probably have a default message for any unsupported apps in IE11 before merging. E.g. CCDB

Screen Shot 2022-04-11 at 2 38 06 PM

contolini (approved):
A PR of this size will likely cause hiccups and ripples in some systems but I've already started using it locally on simple-chart work because IT'S SO FAST.

PR #6980, opened 2022-03-28
Remove broken cypress-in-docker

wpears:
Our specific instantiation of cypress-in-docker was killed in https://github.com/cfpb/consumerfinance.gov/pull/6571. However, we never cleaned up this compose file or the docs. This does that.

A future effort to run cypress in docker should likely target an environment matching github actions, but I'd hesitate to add that until we see a need (and since no one minded that this has been broken since August, I'd reckon that cypress in CI + on one's own machine is good enough).


chosak:
If we are removing this, I would if we could (please) also remove [this part](https://github.com/cfpb/consumerfinance.gov/blob/74e6cdbb031fc05fcf94a1cbbcc0765cef453398/.env_SAMPLE#L185-L191) of .env_SAMPLE, the last remaining remnant of the check against "Darwin". As far as I know the `CFGOV_HOSTNAME ` variable only needed to be defined in .env for Cypress testing purposes.

We do use that same variable in [docker-stack.yml](https://github.com/cfpb/consumerfinance.gov/blob/beb9e8b05d9baa102c714f304757158d7d19ddca/docker-stack.yml#L94), but this is set in [Jenkinsfile](https://github.com/cfpb/consumerfinance.gov/blob/5a77a383db7b7bfb97bb266f2f330cbfbcab9345/Jenkinsfile#L62). Unless there is some other use case I'm not aware of?

willbarton (approved):

PR #6982, opened 2022-03-29
Bump black to 22.3.0

jslay-excella:
Update `.pre-commit-config.yaml` to use black `22.3.0`


jslay-excella:
We might want to update our path filters on GHA to detect when `.pre-commit-config.yaml` changes or `tox.ini` changes, but out of scope for this PR.

wpears (approved):

PR #6984, opened 2022-03-30
Update workflow action versions

anselmbradford:
~~I updated the python version to 3.9.x across all actions to match what was set in functional-tests.yml. Not sure if this version discrepancy was intentional though.~~

## Changes

- Updates actions dependencies to v3 (which uses node 16 by default).
- Downgrade "functional-tests.yml" actions' python versions to 3.8 to match version set in other actions.

## How to test this PR

1. I think we need to merge it and see how it goes.


willbarton:
For Python, we should be using 3.8 everywhere for now.

anselmbradford:
> For Python, we should be using 3.8 everywhere for now.

:+1: Thanks @willbarton, I downgraded the version set in functional-tests.yml to match the other actions. Let me know if that's wrong.

PR #6985, opened 2022-03-30
Removed WellWithAskSearch

jslay-excella:
* Removes `WellWithAskSearch`
* Removes `WellWithAskSearch` from `FullWidthText`


willbarton (approved):

niqjohnson (changes requested):
I think we can also get rid of the reference to it in https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/jinja2/v1/_includes/organisms/full-width-text.html#L44.

This block was added in #5230; I didn't see anything else in there that looks like it could be removed, but it might be worth a second look.

jslay-excella:
Done.

niqjohnson (approved):
Looks good now. Normal wells are still working as intended, and the well-with-search block is no longer available in the full-width text streamfield.

PR #6991, opened 2022-04-04
Adjust + enhance simple-chart filters

wpears:
Simple-chart filters supported a use case of "date filters" with various machinery to support it. This isn't being used anywhere and we've largely settled on the existing buttons + slider for date range picking. Thus, this removes that functionality and associated logic.

As a *special bonus*, I've also made filters work on all series in a chart, which was a long-standing TODO.

As an *extra special bonus*, this PR also makes "View all" optional & opt-in for filters, which allows filters to be used as toggles.


wpears:
Helpful wagtail settings for testing the above:
Screen Shot 2022-04-06 at 5 03 56 PM

PR #6992, opened 2022-04-05
Ensure cache tags exist before invalidating them

willbarton:
This change fixes an issue with our cache tag invalidation on publish of a `AbstractFilterPage` if that page does not belong to a filterable list. In that case, the cache tag invalidation will be attempted but will fail because the list of cache tags to invalidate will be blank.

This fix simply adds a check to ensure the list of cache tags to purge is populated before invalidating. I've modified the tests to cover the behavior we were seeing.

## Checklist


jslay-excella (approved):
lgtm

PR #6994, opened 2022-04-05
Add static html list to raf template for no-js visitors

wpears:
Adds static html to the raf-tool which is displayed when javascript is not supported or enabled.

Testing:
 - Pull branch
 - Visit http://localhost:8000/coronavirus/mortgage-and-housing-assistance/renter-protections/find-help-with-rent-and-utilities/
 - Things should work as normal
 - Open devtools and use cmd + shift +p to open the `run` panel, type `Disable Javascript` and run that command when it populates
 - Refresh the page
 - The static html list should be rendered


chosak:
Where does this data come from? Is it ever updated? Can we include some documentation on how to regenerate this static HTML, if/when that is necessary?

wpears:
>Where does this data come from? Is it ever updated? Can we include some documentation on how to regenerate this static HTML, if/when that is necessary?

This html is copied directly out of the page of the live tool. The underlying data is updated on a ~weekly cadence but the diffs are generally small. My thinking is that this is a better experience than serving absolutely nothing to `no-js` users, so simply shipping a static html list is fine (eg, <1% of users are on `no-js`, <1% of data is changed, users sent down the wrong path would be ~.01%, which seems better than offering no help to them cause they don't support js).

However, do check out https://github.com/cfpb/consumerfinance.gov/pull/6995 which serves a simple html list directly from S3. One can imagine (if updating this `no-js` list becomes a priority for some reason) a Lambda function which watches for updates on the underlying data (https://files.consumerfinance.gov/a/assets/raf/raf.json) and renders the simple html list in #6995 from it. I'm probably leaning towards that solution rather than this one, both for performance and maintenance reasons.

marteki:
Hi, product owner here! My dev setup is blown to bits. Can you share screenshots of this at desktop and mobile sizes?

wpears:
Desktop:
Screen Shot 2022-04-06 at 12 08 58 PM
Mobile:
Screen Shot 2022-04-06 at 12 09 14 PM

I think #6995 is actually heavily preferred for performance reasons (for people on old/slow browsers or who have js off, performance is especially important) but this one is prettier.

marteki:
Prettiness can be improved on the other one with CSS, right?

If so, I would say that we should do the other PR, and iterate on it soon to meet the CFPB Design System guidelines on appearance.

wpears:
Closing in favor of #6995

PR #6996, opened 2022-04-06
Switch tdp svg marker to data: uri

wpears:
Currently, this svg is the only one in our codebase directly referenced in a `less` file. This presents some complications for ongoing frontend build upgrades. Switching this marker to use a data uri obviates those complications with no effect on TDP.

Testing:
 - Pull and build
 - Complete a survey at http://localhost:8000/consumer-tools/educator-tools/youth-financial-education/survey/3-5/
 - The svg marker should be visible:
 
Screen Shot 2022-04-06 at 11 07 10 AM


contolini (approved):
🚗

anselmbradford:
We have a less plugin for including SVGs inline in the design system ([here](https://github.com/cfpb/design-system/blob/00948c3ac7e3425bb6c72c49ffd1d09d849223dd/packages/cfpb-icons/src/icons-svg-inline.js) and [used here](https://github.com/cfpb/design-system/blob/780915fd14a81af30d51524d92b075692fdafd6a/packages/cfpb-icons/src/cfpb-icons.less#L25-L35)). You could consider whether it'd make any sense to use that from the imported cfpb-icons in cfgov to keep the icon available as a stand-alone SVG in the source.

PR #6997, opened 2022-04-06
Fix .gov URL regex

niqjohnson:
External links that point to .gov domains are exempted from going through the external site interstitial, but a missing dot in the regex that finds .gov URLs meant that any URL with "gov" anywhere in it was being treated as a .gov URL.

---

## Additions

- A dot to the regex that was meant to look for .gov URLs but was actually looking for any URL with "gov" anywhere in it
- A couple of unit tests to make sure .gov URLs and URLs that happen to have "gov" somewhere in them are handled correctly

## Changes

- Adjusted the expected HTML output of links in the unit tests to account for the fact that .gov URLs don't have the `data-pretty-href` attribute

## How to test this PR

1. On `main`, try adding two links to a page, one that points to an actual .gov URL (like https://www.fdic.gov) and one that points to a URL that happens to have "gov" somewhere in it (like https://www.realgovsite.lol)
2. Note that neither of those URLs go through the standard external site interstitial
3. Switch to this branch
4. Note that now the link that points to a URL with "gov" somewhere in it goes through the interstitial while the real .gov URL continues to bypass the interstitial
5. Confirm that existing external links continue to work as they do in production (`/consumer-tools/educator-tools/your-money-your-goals/online-resources/` is a good page to test with that has lots of external links on it)

## Checklist


cfgov/core/tests/test_utils.py
```diff
@@ -154,6 +159,14 @@ def check_external_link(
 
         self.assertEqual(add_link_markup(tag, path), str(expected_tag))
 
+    def test_dot_gov_urls(self):
+        url = "https://www.federalreserve.gov"
+        self.check_external_link(url, expected_href=url)
+
+    def test_urls_with_gov_in_them(self):
+        url = "https://www.realgovsite.lol"
```
willbarton, line 167: BRB, buying this domain.

willbarton (approved):

PR #7000, opened 2022-04-07
Fix: Update Elasticsearch when pages are moved

chosak:
There is currently a bug related to the moving of pages: when a filterable page is moved within the Wagtail page tree, it is deleted from its old location but not reindexed at its new location.

This causes problems e.g. when an event page is moved from the list of future events to the list of past events; it fails to show up in its new location because it is not being reindexed.

This happens because the django-elasticsearch-dsl index is configured to [index models that are of type AbstractFilterPage](https://github.com/cfpb/consumerfinance.gov/blob/bc3e2164b6758e4cca7af76af8985ae05e6ffdd1/cfgov/v1/documents.py#L99), but Wagtail page signals can be called with instances of the base Page class, for example [when a page is moved](https://github.com/wagtail/wagtail/blob/ed6365ac0ff97505acdc8761ff72cd717f82c149/wagtail/core/models/__init__.py#L1496-L1503). We therefore need to convert pages from the base Page type to the more specific instance type before trying to reindex them.

Thanks to @ceceliatroyan for raising this issue.

## Checklist

### DIFF
--- a/cfgov/search/elasticsearch_helpers.py
+++ b/cfgov/search/elasticsearch_helpers.py
@@ -15,12 +15,10 @@ from wagtail.core.signals import (
     pre_page_move,
 )
 
-from django_elasticsearch_dsl.registries import registry
 from django_elasticsearch_dsl.signals import BaseSignalProcessor
 from elasticsearch_dsl import analyzer, token_filter, tokenizer
 
 from search.models import Synonym
-from v1.models.learn_page import AbstractFilterPage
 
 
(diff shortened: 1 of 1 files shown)


csebianlander (approved):

wpears (approved):

chosak:
@wpears oops, actually, I wasn't ready to merge this yet. There is one other minor addition that I was in the middle of working on. I'll open a subsequent PR.

PR #7001, opened 2022-04-08
Add browserslist config to esbuild

anselmbradford:
Autoprefixer currently uses browserslist with the `default` browser set. 
To view this:
In `styles.js`, change `plugins: [ autoprefixer ]` to `plugins: [ autoprefixer.info() ]` and run `yarn styles` and see browserslist output.

This PR adds back the browserslist string to `package.json` so that both esbuild and autoprefixer are using the same browser set. 

We could do `"browserslist": "defaults",` in `package.json` if we want to, but I think as written it's a little clearer.

## Additions

- Adds `esbuild-plugin-browserslist` for feeding browserslist string in esbuild.
- Adds browserslist string to `package.json`, without IE11.

## Removals

- Unreferenced `postcss-unmq` and `require-dir`

## How to test this PR

1. In the root directory, run `npx browserslist` to see the browser targets.
2. Change browserslist string to `"browserslist": "ie >= 11",` in `package.json`
3. Re-run `npx browserslist` and see that it only outputs `ie 11`.
4. Run `export DEBUG=esbuild-plugin-browserslist`
5 Run `yarn scripts` and see it fails and esbuild info shows IE11 was targeted (which isn't enough for esbuild).
6. In `styles.js`, change `plugins: [ autoprefixer ]` to `plugins: [ autoprefixer.info() ]` 
7. Run `yarn styles` and scroll up to see browserslist output targets is only IE11.
8. Revert browserslist string in `package.json` to `> 0.5%, last 2 versions, Firefox ESR, not dead`
9. Re-run `yarn scripts` and see the targets are updated and it passes.
10. Re-run `yarn styles` and see the targets are updated.


docs/development-tips.md
```diff
@@ -34,6 +34,11 @@ Feel free to [install a plugin](https://editorconfig.org/#download) for your edi
 
 Some (but not all) JavaScript and CSS files can be rebuilt automatically when they are changed by using `yarn watch`.
 
+## Outputting browser support metrics
+
+Within the root directory, run `npx browserslist` to output the set of browser
+targets given to `esbuild` (for JS) and `autoprefixer` (CSS) transpiling.
+
```
contolini, line 41: > We could do "browserslist": "defaults", in package.json if we want to, but I think as written it's a little clearer.

Can you add this note (and maybe a link to the browserslist docs) to this section for folks who are unfamiliar with the browserslist syntax.

anselmbradford: https://github.com/cfpb/consumerfinance.gov/pull/7001/commits/6e095a678f267a0ee60774362d0793caa839f74b

wpears (commented):
I actually think autoprefixer and esbuild **should not** use the same browserslist since our CSS support and JS support targets differ!

In https://github.com/cfpb/consumerfinance.gov/pull/6978/commits/ce6ea22f2109178c7315e0f73a2fa844c0f5c7d6 I set the esbuild target to include the lowest version of the listed browsers with >0.2% usage (arbitrary cutoff). I know you wanted to use real analytics data & to automate the generation of this list, but I think that work might be in conflict with this PR. Thoughts?

anselmbradford:
> I actually think autoprefixer and esbuild should not use the same browserslist since our CSS support and JS support targets differ!

I fully agree that esbuild and autoprefixer **should have** the ability to have different targets. I also think the apps **should have** this ability as well, lest we can't update the whole site targets because of some edge-case app stakeholder hold-up. 

If we want everything to have the same targets, I think that looks like removing the target in esbuild and having the browserslist string in `package.json`, where esbuild will scoop it up (with the plugin in this PR). 

If we want to have differing targets, then any that differ from what is in base `package.json` defines its own targets (e.g. esbuild uses the base `package.json` values and autoprefixer sets its own target). I haven't checked how this would work with apps (can we add a browserslist file to their directory and will esbuild scoop that up?).

But back to the question of esbuild/autoprefixer targets differing. We do kinda have four domains: un-transpiled JS (no-js state), transpiled JS (js state), un-prefixed CSS, and prefixed CSS.

Screen Shot 2022-04-11 at 2 22 47 PM

The major thing on the CSS-side of the house is dropping gulp-bless, which will majorly bust the layout on IE 10<, which at this point we're saying we don't care due to such low usage. In this PR the targets are `> 0.5%, last 2 versions, Firefox ESR, not dead`, which actually picks up IE11 (you can see with `npx browserslist`). For whatever reason esbuild doesn't care about this, but autoprefixer gets the IE11 prefixing and all is as it is now from a CSS-perspective in IE11. If we explicitly exclude IE with `> 0.5%, last 2 versions, Firefox ESR, not dead, not ie 11`, the layout begins to break in certain spots, but I don't think it will be completely destroyed like it will be in earlier IE without gulp-bless.

Screen Shot 2022-04-11 at 2 29 04 PM

So similar to the argument of our supported browser set being evergreen, maybe at this point it makes sense to have the targets the same for esbuild and autoprefixer? (Again though, with the **possibility** they could differ, if required).

> In https://github.com/cfpb/consumerfinance.gov/commit/ce6ea22f2109178c7315e0f73a2fa844c0f5c7d6 I set the esbuild > target to include the lowest version of the listed browsers with >0.2% usage (arbitrary cutoff). I know you wanted to use 
> real analytics data & to automate the generation of this list, but I think that work might be in conflict with this PR. 
> Thoughts?

So my thoughts on this are something like this: **wouldn't it be nice if we could quickly and clearly answer what our browser support set is, what our visitor browser set is, and how to quickly and cleanly move our support to meet our visitors over time?**

As far as live analytics data, what I have in mind is not necessarily a direct feed into cfgov, but rather a dependency that can become outdated so we know it needs updating. What that might look like is https://github.com/cfpb/cfpb-analytics/blob/main/packages/browserslist-config/browserslist-stats.json gets updated on a regular interval and https://www.npmjs.com/package/@cfpb/browserslist-config gets bumped and released at the same cadence. 

Then in cfgov, the browserslist set in `package.json` could be the direct metrics (with syntax similar to [here](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/apps/analytics-gtm/browserslist#L9)). 

The site wouldn't be constantly shifting around in its build targets automatically, potentially mysteriously breaking stuff, but instead on a regular cadence we could [bump @cfpb/browserslist-config](https://github.com/cfpb/consumerfinance.gov/blob/main/package.json#L23) in a PR for review and over time move our targets to a more modern set.

We could also incorporate [eslint-plugin-compat](https://github.com/amilajack/eslint-plugin-compat) and [stylelint-no-unsupported-browser-features](https://github.com/ismay/stylelint-no-unsupported-browser-features) to be able to lint the codebase against the PR that bumps the browserslist and flag any issues that way.

_Side note: I appreciate this discussion! I feel like we're narrowing in on a lean-mean-great-config-machine!_

PR #7002, opened 2022-04-11
Fix: Update Elasticsearch when pages are moved

chosak:
Followup to #7000. Properly handle cases where the django-elasticsearch-dsl signals are invoked from something that is
not a Wagtail page object. Also add tests.

## Checklist


contolini (approved):
TY

PR #7003, opened 2022-04-11
Add a note to the ccdb no-js state

wpears:
ccdb currently renders an empty page when a `no-js` user visits, see https://github.com/cfpb/consumerfinance.gov/pull/6978#issuecomment-1095417416

This adds a note telling them that is expected and links them back to the data download.
Screen Shot 2022-04-11 at 4 47 04 PM

Testing:
 - Pull and build styles
 - open a tab and disable js (cmd+opt+j, cmd+shift+p, type Disable javascript)
 - visit http://localhost:8000/data-research/consumer-complaints/search/ and the notification should render, with a link back to the data download.


cfgov/jinja2/v1/ccdb-complaint/ccdb-search.html
```diff
@@ -18,7 +18,19 @@
 {%- endblock %}
 
 {% block content_main %}
-  <div id="ccdb-ui-root"></div>
+  <div id="ccdb-ui-root">
+    <div style="max-width:75%;margin: 6em auto;" class="js-is-hidden">
```
anselmbradford, line 22: You could also do `<div class="js-is-hidden wrapper block block__padded-top block__padded-bottom">` to avoid the inline styles. It'll align with the text, not the menu selection bar, which is how we usually align page content.

anselmbradford, line 22: You could also do `<div class="js-is-hidden wrapper block block__padded-top block__padded-bottom">` to avoid the inline styles. It'll align with the text, not the menu selection bar, which is how we usually align page content.

wpears: I tried that first but it seemed too tight visually and there was white space below the footer which seemed somewhat broken
Screen Shot 2022-04-12 at 8 20 44 AM

wpears: The wrapper class makes sense though, added in bcf823dcd09b4918328d7ab1d689ce8830ba9c72

anselmbradford: Weird I don't see that gap at the bottom. Do you see what causes that? 

You could also do it all in classes with `js-is-hidden wrapper u-mt60 u-mb60`, though yes not quite as big a gap (but the same as above "About Us"

Screen Shot 2022-04-12 at 11 18 50 AM

wpears: Fixed in 34d1b2035b746331c25062a493158a5bb8ea3e48

cfgov/jinja2/v1/ccdb-complaint/ccdb-search.html
```diff
@@ -18,7 +18,19 @@
...
+    <div class="js-is-hidden wrapper u-mt60 u-mb60">
+      <div style="padding: 2em 1em;" class="m-notification m-notification__error m-notification__visible">
+        {{ svg_icon('error-round') }}
+        <div class="m-notification_content">
+          <div class="h4 m-notification_message">
+            This interactive tool is not supported in your web browser. Please use a newer browser or
+            <a style="text-decoration:underline" href="/data-research/consumer-complaints/#download-the-data">download the data</a>.
+          </div>
+        </div>
+      </div>
```
anselmbradford, line 31: Oh shouldn't we be using the existing notification template?https://github.com/cfpb/consumerfinance.gov/blob/03ce969697f3090e30d1d2a282e6980637a415f0/cfgov/jinja2/v1/_includes/molecules/notification.html

Usage example is at https://github.com/cfpb/consumerfinance.gov/blob/03ce969697f3090e30d1d2a282e6980637a415f0/cfgov/jinja2/v1/_includes/organisms/filterable-list.html#L43-L52

wpears: The structure of that template is somewhat rigid and in this context prevented me from rendering the inline link. I didn't want to update that template to make all passed messages as `| safe` (because maybe some aren't).

anselmbradford: FWIW I recall something about that being by design, as we don't have notifications with inline links in the DS https://cfpb.github.io/design-system/components/notifications#information-base-notification-1 and the recommendation is to reword the notification to provide all links as a list at the end. I think it was about notification content coming from Wagtail and we didn't want a rich text block for the content.

But I also appreciate that this has already gone through the stakeholder and is very edge case.

anselmbradford: Or
Screen Shot 2022-04-14 at 10 56 32 AM

wpears: Screen Shot 2022-04-14 at 12 32 26 PM

wpears: (in 3834131)

anselmbradford (approved):

PR #7008, opened 2022-04-14
HOLD: Deprecate Planning for Retirement feedback form

chosak:
This change removes the feedback form at the bottom of the Planning for Retirement / Before you Claim page located in English and Spanish at:

https://www.consumerfinance.gov/consumer-tools/retirement/before-you-claim/
https://www.consumerfinance.gov/consumer-tools/retirement/before-you-claim/es/

At the bottom of the current page there is a form that asks "Was this page helpful in thinking about when to start collecting benefits?" with Yes/No buttons. Clicking one of these buttons sends an event to Google Analytics.

This commit removes this form entirely.

## How to test this PR

To test, go through the page in both English and Spanish locally:

http://localhost:8000/consumer-tools/retirement/before-you-claim/
http://localhost:8000/consumer-tools/retirement/before-you-claim/es/

Confirm that the page continues to work and there are no browser console errors.

## Screenshots

|Before|After|
|-|-|
|image|image|

## Checklist


anselmbradford (approved):
This looks good to me. One thing to consider running by design/UX is whether we need the 60px gap between the end of the tool and the legal disclaimer set here https://github.com/cfpb/consumerfinance.gov/blob/5920330a7fc65cffd92663e2c3a5c3bf461ac94e/cfgov/unprocessed/apps/retirement/css/claiming.less#L488

Maybe the extra space is nice to make it visually separated from the "app" content, but also page real estate is precious, so seems like it could go either way.

chosak:
Thanks @anselmbradford.

@huetingj @sonnakim, any thoughts on what Ans mentions? Here's how it looks without the space:

|Current|Without the space|
|-|-|
image|image|

PR #7010, opened 2022-04-15
Resolve 'getFiles' to absolute paths and adjust string replacement logic

wpears:
In some Unix environments, the current way files were resolved was not correct.

This updated logic will work in all cases.


esbuild/utils.js
```diff
@@ -31,7 +30,6 @@ async function getFiles( dir ) {
     return dirent.isDirectory() ? getFiles( res ) : res;
   } ) );
   return files.flat().filter( v => v )
```
anselmbradford, line 32: Need `;` at the end?

anselmbradford (approved):

PR #7018, opened 2022-04-18
Better regex for targeting hidden files (to ignore)

wpears:
If returning absolute paths, the current regex attempting to target hidden files won't work. This fixes that.

Also causes makes the regex case-insensitive, which it should be.

### DIFF
--- a/esbuild/copy.js
+++ b/esbuild/copy.js
@@ -7,10 +7,12 @@ module.exports = async function( baseConfig ) {
   const resolvedBase = resolve( unprocessed );
   const files = await getFiles( resolvedBase );
 
-  const staticFiles = files.filter( v => !v.match( /^\.|\.js$|\.less$|\.css$/ ) );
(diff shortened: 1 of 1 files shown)


Colin-Seifer (approved):

esbuild/copy.js
```diff
@@ -7,10 +7,12 @@ module.exports = async function( baseConfig ) {
   const resolvedBase = resolve( unprocessed );
   const files = await getFiles( resolvedBase );
 
-  const staticFiles = files.filter( v => !v.match( /^\.|\.js$|\.less$|\.css$/ ) );
+  const staticFiles = files.filter( v => !v.match( /\/\.[-.\w]*$|\.js$|\.less$|\.css$/i ) );
```
anselmbradford, line 10: For future maintenance, maybe add a comment above this line with a link to a regex example showing what this is matching and not matching, like is done here https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/js/modules/util/validators.js#L35

PR #7022, opened 2022-04-19
Remove cloud print

wpears:
Removes the non-functional google cloud print library (sunset December 2020 😱) and associated code.

To test, you'll need to pull up the offer page with a functional offer (this can be done by googling "Understanding your financial aid offer"...oddly).

Printing should still work and should redirect to a "save as pdf" page by default on android.


anselmbradford (approved):
This LGTM code-wise, but I can't test. @mistergone should prob have final say.

chosak:
It's possible to test "Understanding your financial aid offer" using some test data; see https://github.com/cfpb/consumerfinance.gov/pull/7019#issue-1207359893 for instructions.

PR #7023, opened 2022-04-19
Add new Research Hub page type

schbetsy:
Adds a new page type to highlight research reports owned by the Office of Research. We will be staging and publishing an instance of this page type soon.

## Additions

- Adds the Research Hub page type, which filters by the Data Point and Industry and Markets tags
- A migration to reflect this addition

## Changes

- Minor corrections to documentation

## Screenshots

coming soon

## Checklist


cfgov/v1/models/sublanding_filterable_page.py
```diff
@@ -66,6 +66,11 @@ class SublandingFilterablePage(FilterableListMixin, CFGOVPage):
     ]
 
 
+class ResearchHubPage(CategoryFilterableMixin, SublandingFilterablePage):
+    template = "sublanding-page/index.html"
```
chosak, line 70:
```suggestion
```

Is this needed, since it's already inherited from the parent `SublandingFilterPage`?

schbetsy: When I remove that line and try to load the page, I see `django.template.exceptions.TemplateDoesNotExist: v1/research_hub_page.html`

cfgov/v1/tests/util/test_ref.py
```diff
@@ -8,20 +8,6 @@
 )
 
 
-class TestCategories(TestCase):
-    def test_no_duplicate_slugs(self):
```
schbetsy, line 12: Here's my commit message for this change:

| Now, a page exists (Research Hub Page) that uses the same filterable page categories that are already used by another page (Research Report Page). This is intentional -- the Research Hub page is supposed to filter a subset of Research Reports. So, this test is no longer enforcing desired behavior. Deleting.

If I'm mistaken and something else is going on here, please let me know.

cfgov/v1/models/sublanding_filterable_page.py
chosak: Ah, you are right. Thanks for checking my assumption. Wagtail doesn't inherit the template from the parent but will instead try loading a template named for the specific page class.

Another way of avoiding the duplicate hardcoding might be to do:

```suggestion
    template = SublandingFilterablePage.template
```

cfgov/v1/tests/util/test_ref.py
chosak: It appears I am the one who added this test in https://github.com/cfpb/consumerfinance.gov/pull/2740 ("New unit test that verifies that all page categories have unique keys") but I can't remember a specific reason why this was desirable.

chosak (approved):
Works for me.

This is a case where it would be nice if Wagtail supported proxy page models, so we didn't need to create a separate database table just to change a tiny bit of logic for the page type. Former CFPBer Kurt filed https://github.com/wagtail/wagtail/issues/2480 about this back in April 2016!

PR #7031, opened 2022-04-22
Add supported browsers script to query user agent for browser JS support

anselmbradford:
`window.Promise` is not a comprehensive cutoff check for browser support of ES features. This PR adds [`browserslist-useragent-regexp`](https://github.com/browserslist/browserslist-useragent-regexp) to generate a regex script based on the current browsersList, which is used to set a global variable for checking if the visiting browser is in the set or not.

## Additions

- `browserslist-useragent-regexp` dependency.
- `yarn generateBrowserSupportScript` script to generate `cfgov/cfgov/v1/_includes/templates/supported_browsers.html`

## Changes

- Include `cfgov/cfgov/v1/_includes/templates/supported_browsers.html` in base template to set a `window. cfpbIsBrowserSupported` variable, which then controls js/no-js logic.
- Generate `supported_browsers.html` when building JS.
- Alphabetize yarn scripts.

## How to test this PR

1. Run `yarn build`.
2. Run `npx browserslist`.
2. Visit the site in a browser that is in the generated list.
3. Inspect the page and see that the `html` element has a `js` class, or check `window.cfpbIsBrowserSupported` in dev console.
4. Visit the site in a browser that is NOT in the generated list.
5. Inspect the page and see that the `html` element has a `no-js` class.
6. Go to `package.json` and edit the `browserslist` string, such as changing `> 0.5%` to `> 80.5%` and run `yarn build` and see that `supported_browsers.html` is updated.

## Notes and todos

- Open to better self-documenting ways of generating the browsers regex JS script that doesn't involve manually minified JS.


anselmbradford:
Closing in favor of https://github.com/cfpb/consumerfinance.gov/pull/7037

@wpears and I discussed this, and decided the infrastructure for this is not worth it. By setting the target to `es6` and making the cutoff support of Promises, we can likely can create a browserslist from https://github.com/Fyrd/caniuse/blob/main/features-json/promises.json. Possibly by copying the `stats` portion of `promises.json` into `browserslist-stats.json` in the cfgov repo root, add a browserslist entry into package.json with `> 0% in my stats` and running `npx browserslist`.

PR #7032, opened 2022-04-22
Use unused error argument in error message

anselmbradford:
## Changes

- Use unused `error` argument in error message.

## How to test this PR

1. The mega menu should show up in legacy browsers (e.g. Chrome 65).


contolini (approved):
Impressive find!

PR #7038, opened 2022-04-26
remove deprecated less compression. esbuild is already doing this

wpears:
Closes https://github.local/CFGOV/platform/issues/4402


anselmbradford (approved):
Do we have a way of turning off compression temporarily? Sometimes I look at the generated styles before it's minified.

wpears:
>Do we have a way of turning off compression temporarily? Sometimes I look at the generated styles before it's minified.

You can just add `minify: false` to the esbuild config

PR #7042, opened 2022-04-28
Simple-chart radio buttons

wpears:
Closes https://github.local/CFPB/super-regular/issues/380

Changes filters so that, when there are five or less filter groups a set of radio buttons is created (instead of a select element + options).

Also adjusts the return values of `makeSelectFilterDOM` and `makeRadioFilterDOM` to return a small object through which events can be wired without having to query the DOM directly.

Also trimmed out some extraneous header creation logic.

Screen Shot 2022-04-28 at 5 09 34 PM

### 
Screen Shot 2022-04-28 at 5 21 33 PM
Testing params:


ielerol:
I like this a lot better than the dropdown, especially with the extra title removed.

I have some comments that are specific to the seasonally adjusted/nonadjusted use case here. First, I'd like the title for the control overall to be "Select data adjustment." Then for the radio button options themselves, I think it makes sense for the first (and default) view to be the seasonally adjusted data, and the label for that selection should be "Seasonally adjusted". 

If it's not feasible to customize the title to be different from the filter options that way, maybe for now we keep it simple and just call it "Select data"?

wpears:
The title of the radio & select are already customizable:
Screen Shot 2022-05-02 at 10 02 32 AM

That's driven by the `label` property of the filter (set to "Adjusted or Unadjusted data" in my initial screenshot). I'm actually going to touch up the logic a bit here to make it slightly more flexible but this already supported for your text suggestion.

The ordering of the radios is more interesting, because that's purely based on the the order it encounters each class/bucket of data when it scans through the rows. However! we're already applying a transform on our data, we can just adjust label of the data and its order in the transform to match your request.

wpears:
Screen Shot 2022-05-02 at 10 24 26 AM

brandon-rusk (approved):

PR #7047, opened 2022-05-02
Simple-chart label positioning

brandon-rusk:
## Changes

- Changed the positioning of the projected value label when there is a projected value present in the chart

## How to test this PR

1. Pull PR
2. `yarn build`
3. Go to http://localhost:8000/admin/pages/4481/edit/
4. Add Simple-Chart block, add projected months value
5. Preview and see new label positioning

## Screenshots

With projected months:
image

With projected months:
image

Without projected months:
image

### DIFF
--- a/cfgov/unprocessed/js/routes/on-demand/simple-chart/simple-chart.js
+++ b/cfgov/unprocessed/js/routes/on-demand/simple-chart/simple-chart.js
@@ -118,7 +118,10 @@ function makeChartOptions( data, dataAttributes ) {
     };
   }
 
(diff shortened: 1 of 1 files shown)


ielerol (approved):
I shared the screenshots with @huetingj and we talked about the spacing, and we are good to go with the spacing as it currently exists. While the space between the project values label and the top of the chart is a bit tighter than normal, in this case it actually potentially helps the readability by keeping that info close to the relevant portion of the chart, and separating it cleanly from the overall chart legend.

wpears (approved):

PR #7048, opened 2022-05-03
Move vars.less to main.less

wpears:
Tiny first step towards consolidation of our somewhat sprawling less code


anselmbradford:
We could also possibly remove this entirely and use the colors directly, or only have it a variable in `inset.less`. The intention of its location was to evaluate it for moving into the contents of https://github.com/cfpb/design-system/blob/main/packages/cfpb-core/src/vars.less, but using the color variables directly is probably clearer. It possibly was set up this way to support theming.

anselmbradford (approved):

PR #7049, opened 2022-05-03
Remove references to font cdn

wpears:
Now that we're always using self-hosted fonts, we don't need these.


anselmbradford:
Are you planning to do a companion PR to the DS? Would be good to reference here

wpears:
Yeah, I was considering it . Not strictly necessary, but there's some logic I can remove and it's probably worth removing. I'll have some time tomorrow to do so.

wpears:
Design-system PR: https://github.com/cfpb/design-system/pull/1370

This is actually non-blocking for this PR, since the default for `use-font-cdn` [has already been set to false in the DS](https://github.com/cfpb/design-system/blob/main/packages/cfpb-typography/src/cfpb-typography.less#L24)

anselmbradford (approved):

PR #7050, opened 2022-05-04
remove broken admin simple-chart-css reference

wpears:
https://github.com/cfpb/consumerfinance.gov/pull/7030 added some simple-chart admin css, but did so in two ways, [the first way](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/v1/wagtail_hooks.py#L133) works fine, and the second way (removed in this PR) is both broken and unnecessary. Thus, I'm removing it!


brandon-rusk (approved):
Thanks for catching this!

PR #7052, opened 2022-05-04
Added Helm CronJob Template with clearsession job

jslay-excella:
* Adds `cronjob.yaml` as a cronJob template to have as a simple feature for adding cronJobs
  * Define jobs in `values.yaml` under `cronJobs`. 
* Adds `clearsessions` job that runs weekly on production, daily on local.

To make a new job, its just adding a new item to the `cronJobs` array:

```
- name: "clearSessions"
  schedule: "@daily"
  command:
    - "django-admin"
  args:
    - "clearsessions"
  restartPolicy: OnFailure
```

Further improvements that could be made:
  * Add option to define image for the cronjob instead of just using the overall image for cfgov

PR #7053, opened 2022-05-05
Q2 dep update

jslay-excella:
Updates Python dependencies and removes unneeded (either defined by other package, or not needed).


willbarton (changes requested):
Can you say more about how you determined some of these are not needed?

Looking at `requirements/libraries.txt`, there look like a lot that, AFAIK, we do need.

jslay-excella:
OK, I went back through, re-adding back in explicitly imported libraries like we discussed this morning. I think this is ready now. 

Also, what version of ES are we using in Prod? (Send through chat pls).

requirements/libraries.txt
```diff
@@ -2,49 +2,43 @@
...
+boto3==1.22.7
 dj-database-url==0.5.0
-djangorestframework==3.11.2
-django-cors-headers==3.9.0
+django-cors-headers==3.11.0
 django-csp==3.7
-django-elasticsearch-dsl==7.1.4
-django-extensions==2.1.3
-django-flags==5.0.8
+django-debug-toolbar==3.4.0
```
willbarton, line 9: We do need this in `libraries.txt`? It *should* only be used for local dev, AFAIK.

willbarton (commented):
Finally dropping the independent `mock` package is heroic.

This looks good, I have one comment on debug-toolbar.

requirements/libraries.txt
jslay-excella: Local dev. Good catch.

wpears (approved):

willbarton (approved):

willbarton:
Oh, this is going into #7052! We should get this merged, then that.

jslay-excella:
Local dev. My bad.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #7054, opened 2022-05-05
Filing Instruction Guide Section Re-Design

cooldragontattoo:
The current Filing Instruction Guide application is an extension of the (now shelved) reports_as_webpage project.  Therefore, the current design system for the application, in both the code and wagtail admin, favors the context of reports and not FIG content.  This re-design overhauls the models.py and jinja templates to cater specifically to FIG content.  Note that customizations to the javascript/css--while required--have not been made yet.   

---

## Additions
- fig section molecule
- fig content organism
- fig sub section organism
- fig sub section 3 organism
- separate `macros.html` for `index.html` template 

## Removals

- CSS classes were removed from `sidenav_toc.html` file
- from `models.py`:
   - all references to reports
   - removed parent/child model paradigm 

## Changes

- re-configured the templates to accept re-tooled context 

## How to test this PR

1.  Pull and run locally.

## Screenshots

| Before | After |
| -- | -- |
| Screen Shot 2022-05-05 at 9 38 27 AM | Screen Shot 2022-05-05 at 9 40 40 AM | 

## Checklist


wpears (changes requested):
I think these changes might be a step back in terms of the model structure of the FIG. I think the existing hierarchical model structure makes sense for nested sections/sub-sections/third level sections and I think it also makes sense to keep the model additions encapsulated in `cfgov/filing_instruction_guide/models.py` instead of adding code to `organisms.py` and `molecules.py` that really only is going to be used on one page.

One thing I'll note: the editing experience in Wagtail could certainly be improved by adjusting the editor CSS, eg, adding a css file named `fig-admin.css` to `templates/wagtailadmin/css/` and including it in the `editor_css` group of css files in `v1/wagtail_hooks.py`

Something like the following expands the blocks:

```
#tab-filing-instruction-guide-content .object .multiple, #tab-filing-instruction-guide-content .object .multiple fieldset {
  width: 100%;
  max-width: 100%;
}
```

wpears:
I did some noodling on https://github.com/cfpb/consumerfinance.gov/tree/fig-nested-blocks, this retains the hierarchical approach but converts to using blocks as is done here (which I think is a good idea).

Lemme know what you think!

cfgov/filing_instruction_guide/models.py
```diff
@@ -2,142 +2,94 @@
...
-        on_delete=models.CASCADE,
-        related_name="report_sections_level_three",
-    )
-
-
+# function passed to context processor 
+def parse_section_builder(request, self):
+
+    # creates section dictionary object
+    def create_section_object(section, type_of_block):
```
schbetsy, line 20: I think we can re-work the code to avoid having to do this, breaking up a wagtail block into its component parts and putting them into a new data structure. This does get the contents onto the page correctly, but I think there's a more straightforward method that uses built-in Wagtail methods and reduces the chance for error.

schbetsy:
Closed in favor of #7065.

PR #7055, opened 2022-05-05
Organism css

wpears:
In [base.html](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/jinja2/v1/_layouts/base.html#L314-L324), we currently add javascript based on whether a block adds it as the `js` class variable its inner `Media` class. This extends that behavior to `css` class variables, which add stylesheets to the head of a rendered document if any blocks on the page need them.

Testing:
 - pull and build
 - visit http://localhost:8000/administrative-adjudication-proceedings/administrative-adjudication-docket/3d-resorts-bluegrass/
 - should look identical to prod
 - the `on-demand/sidebar-contact-info.css` file should be requested

A follow-on PR will actually methodically move the rest of the organisms and molecules to `on-demand`


cfgov/v1/tests/models/test_base.py
```diff
@@ -505,17 +505,17 @@ def test_live_with_another_live_page(self):
...
 
 
-class TestCFGOVPageMediaProperty(TestCase):
-    """Tests how the page.media property pulls in child block JS."""
+class TestCFGOVPageMediaJSProperty(TestCase):
+    """Tests how the page.media_js property pulls in child block JS."""
 
     def test_empty_page_has_no_media(self):
-        return self.assertEqual(CFGOVPage().media, [])
+        return self.assertEqual(CFGOVPage().media_js, [])
```
willbarton, line 512: These don't need a `return` — the assertion will raise an `AssertionError` if fails, and the test runner catches that exception to fail the test. These look like they predate your change. I'm not sure how many more of these there are in this file? It looks like this is all of them, though.

```suggestion
        self.assertEqual(CFGOVPage().media_js, [])
```

cfgov/v1/tests/models/test_base.py
```diff
@@ -505,17 +505,17 @@ def test_live_with_another_live_page(self):
...
 
     def test_empty_page_has_no_page_js(self):
         return self.assertEqual(CFGOVPage().page_js, [])
```
willbarton, line 515:
```suggestion
        self.assertEqual(CFGOVPage().page_js, [])
```

cfgov/v1/tests/models/test_base.py
```diff
@@ -505,17 +505,17 @@ def test_live_with_another_live_page(self):
...
 
     def test_empty_page_has_no_streamfield_js(self):
-        return self.assertEqual(CFGOVPage().streamfield_js, [])
+        return self.assertEqual(CFGOVPage().streamfield_media("js"), [])
```
willbarton, line 518:
```suggestion
        self.assertEqual(CFGOVPage().streamfield_media("js"), [])
```

cfgov/v1/tests/models/test_base.py
```diff
@@ -542,9 +542,51 @@ def test_doesnt_pull_in_media_for_nonexistent_child_blocks(self):
...
         # shouldn't add any additional files because of the FullWithText.
-        self.assertEqual(page.media, ["secondary-navigation.js"])
+        self.assertEqual(page.media_js, ["secondary-navigation.js"])
+
+
+class TestCFGOVPageMediaCSSProperty(TestCase):
+    """Tests how the page.media_css property pulls in child block CSS."""
+
+    def test_empty_page_has_no_media(self):
+        return self.assertEqual(CFGOVPage().media_css, [])
```
willbarton, line 554:
```suggestion
        self.assertEqual(CFGOVPage().media_css, [])
```

cfgov/v1/tests/models/test_base.py
```diff
@@ -542,9 +542,51 @@ def test_doesnt_pull_in_media_for_nonexistent_child_blocks(self):
...
+
+    def test_empty_page_has_no_streamfield_css(self):
+        return self.assertEqual(CFGOVPage().streamfield_media("css"), [])
```
willbarton, line 557:
```suggestion
        self.assertEqual(CFGOVPage().streamfield_media("css"), [])
```

willbarton (commented):
As an approach mirroring how we handle organism Javascript, this makes sense to me! A few trivial things in the tests though.

willbarton (approved):

Issue #666, opened 2015-06-18
Check for consistent indentation and whitespace


jimmynotjim:
@anselmbradford [pointed out L:313](https://github.com/cfpb/cfgov-refresh/pull/660#discussion_r32529145) in `_includes/post-macros.html` but I'm sure there's others. 

There are also a few instances of inconsistent whitespace [here](https://github.com/cfpb/cfgov-refresh/pull/660#discussion_r32544044) and [here](https://github.com/cfpb/cfgov-refresh/pull/660#discussion_r32544104)

It's probably a good time for a quick sweep to make sure all files are indented correctly and excess whitespace is removed.

jimmynotjim:
Adam and I were discussing [Prettier](https://prettier.io/) today. It's opinionated, and may not match all our personal preferences, but being able to offload nonsense like this from each of our mental weights would be 💯. Has anyone else used it and have some first person feedback or recommendations?

TL;DR; is that you write the code however you want and on save Prettier reformats it to match _its_ standards. You can't customize from what I've seen, but I'm ok with that if I never have to note an indentation mixup in a PR ever again.

anselmbradford:
Pretty sure we could use stylelint with the autofix flag to fix these too, and since we already use that elsewhere and it's customizable I'd prob favor it over prettier.

jimmynotjim:
The benefit of Prettier is that it runs in your IDE of choice, so no one has to remember to run it, and the fixes are limited to formatting, it won't try and fix other non-stylistic linting issues. If we did use Prettier, we would still want to use Linters to catch larger issues like improper conditions or use of returns, etc. In the case of the Atom plugin, it comes with [ESLint integration as well](https://github.com/prettier/prettier-atom).

anselmbradford:
Seems like it could be on a per-dev basis to use Prettier or not since it's local in the IDE, but the project should prob still have a matching stylelint config overall.

Scotchester:
It looks to me like Prettier could be useful for our JS and CSS, _if_ we can get people to install the plugin in their editor. It all hinges on that. (It doesn't appear to support plain HTML or Jinja templates, from my initial look at it, but I'm OK with that, since my strongest opinions about formatting impacting code readability are reserved for HTML.)

Conceivably, we could cook up a script to run Prettier in Travis before merges, as a backstop for folks not having run Prettier before submitting the PR.

ascott1:
[...]

PR #7057, opened 2022-05-05
Add chart legend title and flexible margin

brandon-rusk:
## Changes

- Added a new title to the chart legend and adaptable margins based on whether the projected months label is present or not

## Screenshots

No projected months label:
image

With projected months label:
image


ielerol (commented):
The legend title size and spacing looks good to me!

wpears (changes requested):
Cool, I think the labelling works here. Two suggestions though, first, it seems a little more idiomatic to define the `title` styles completely in the object, as in the following:
```
  {
      text: '(Click to show/hide data)',
      style: {
        fontStyle: 'italic',
        fontWeight: 'normal',
        fontSize: '14px',
        color: '#666'
      }
    }
 ```
 Second, I don't think it makes sense to include this label on charts where we only have one series, because we *specifically disable* the ability to show/hide the data in those cases ([in this code block](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/js/routes/on-demand/simple-chart/simple-chart.js#L111-L120)).

What I think you should do, then, is instead of adding this title to `common-styles`, add an `else` statement to the above check on the series length and set `defaultObj.legend.title` to the above object in that `else` block.

wpears (approved):
👍

PR #7060, opened 2022-05-11
Add Bandit to check for common Python security issues

willbarton:
This PR adds [Bandit](https://bandit.readthedocs.io/) to our tox `lint` env and to our pre-commit configuration, and adds configuration for it, and adds Bandit to our docs. The configuration is set to ignore all tests, and to ignore `markup_safe` and `mark_safe`.

I've also fixed the issues Bandit immediately flagged when running through the entire codebase:

- Hardcoding temp directories -> use `tempfile.gettempdir()`
- Use of `subprocess` -> use `# nosec` to flag them as acceptable
- Bandit seems to detect any variable with `password` or `token` in the name as a potential password. Marked with `# nosec`
- Bandit flags the use of `random` to ensure you’re not using it for security/cryptography. Marked with `# nosec`.
- Uses of `assert` in TDP. `assert` is not intended for use in enforcing interface constraints and can be removed when compiling to optimized bytecode with `python -o`. I’ve replaced the `assert`s with `if` statements that `raise AssertionError` to maintain the existing behavior. This should probably be refactored, as I don’t readily see any exceptions catching these `AssertionError`s, which mean they can generate 500s for users. But I didn’t do a close reading of the code around them.
- Bandit flags a use of MD5 in TDP. For now, I’ve added `# nosec`, but when we get to Python 3.9+ we can use I’ve set `usedforsecurity=False` in that call (Bandit ignores that argument below 3.9), but it might be worth checking out more thoroughly. I've added a `TODO` comment there.

## Checklist


Colin-Seifer (approved):

chosak:
@willbarton I'm getting a local test failure on my Mac now that I believe might be due to this commit. Specifically, on Mac, is there a guarantee that calling [`tempfile.gettempdir()` here](https://github.com/cfpb/consumerfinance.gov/commit/fee41b78652bdb00f3f9aced13de76db0a8811ee#diff-20e3532f5c2a96c9845af1cd7eaf2e19eaa0f0c8845b20f8bf5dae11f0bd700cR12) will return a path under /tmp, required by a test [here](https://github.com/cfpb/consumerfinance.gov/blob/51908c330cabfc44857c014623d83f5c49c67e87/cfgov/ask_cfpb/tests/models/test_pages.py#L133)?

According to the Python docs here, `gettempdir()` will use the value of the environment variable `TMPDIR` is set, which it is for me:

```sh
 % echo $TMPDIR
/var/folders/0f/qgk6y5z97yqd38_yz4595p940000gq/T/
```

Do you have this environment variable set? Do the tests pass locally for you?

willbarton:
Interesting! I do have `TMPDIR` set to `/tmp`... and it looks like that comes from the `.zshrc` I've carried around for many, many years.

I'll get a fix in for this.

PR #7062, opened 2022-05-11
Replace axios with fetch in rate-checker and rout

wpears:
In the past, we've used a number of `fetch`-alikes for making requests in older browsers. Due to our support changes, this is no longer needed and we can just use native `fetch`.

This PR makes that change for the rural and underserved tool and the owning-a-home ratechecker. (unfortunately the jsonp code can't be replaced in rout because the census service doesn't set CORS headers... oddly enough).

Testing:
 - Pull, `./frontend.sh`
 - `yarn jest` should pass
 - http://localhost:8000/rural-or-underserved-tool/ should function as expected
 - http://localhost:8000/owning-a-home/explore-rates/ should function as expected (and set the mortgage amount to ~1M to see jumbo loan effects as well).


anselmbradford:
FYI if it matters for anything, the `fetch` set is slightly more restrictive than the `promises` set
https://caniuse.com/promises
https://caniuse.com/fetch

wpears:
That definitely matters! I'll update our checks to use `fetch` instead of `Promise`, both are well beyond our practical visitor numbers, but it's good to match usage to support

anselmbradford (approved):
LGTM :+1:

PR #7065, opened 2022-05-12
New models for Filing Instruction Guide content page

schbetsy:
Previously, the filing instructions guide used the data model that we had originally designed for the Reports as Webpages project. We borrowed that code to get the nice design and layout with the Table of Contents for long, detailed content pages. However, the pages used a structured clusterable data model instead of a streamfield, and wouldn't allow wagtail content blocks (e.g. table, well) in the content. This PR switches the data models to a streamfield while trying to retain the sidenav TOC functionality.

If merged, will supersede #7054.

## Changes

- Totally gut and simplify the models in the `filing_instruction_guide` app to use a StreamField paradigm
  - This allows us to use simplified logic for generating templates and assigning section numbers to the sections in the page content
 - Rather than assigning section numbers at page load time, assign them at page save time (via a custom Wagtail form) instead

## How to test this PR

1. Pull the branch and run it
2. Create a FIG Content Page
3. Add sections, subsections, and level-three subsections to the page
4. Publish or preview the page
5. See that the sections are numbered in a logical way in both the side-nav and the page content. 
6. Check that the anchor links in the side-nav work to get you to the correct content section. Check that you can click a section header to anchor to that section.

## Notes and todos

Todo:
- The TOC is supposed to become an accordion at smaller screen widths, but that's currently busted.
- I'd like to find a better way to display the section ID field in the Wagtail Admin to make it clear it's programmatically edited, not a field for the content editor to edit.

Note: Eventually, we'll make this page type (currently called FIG Content Page) into a more generic page type (probably Table of Contents Page) so it can be used for a variety of projects. For now, the code and models are still changing too much, and the styling isn't ready.

## Checklist


cfgov/filing_instruction_guide/wagtail_hooks.py
```diff
@@ -0,0 +1,12 @@
...
+
+from filing_instruction_guide.models import FIGContentPage
+
+
+def assign_content_section_ids(request, page):
+    if page.specific_class == FIGContentPage:
+        page.assign_section_ids()
+
+
+hooks.register("after_create_page", assign_content_section_ids)
```
chosak, line 11: Per [the Wagtail documentation](https://docs.wagtail.org/en/v2.15.5/reference/hooks.html#after-create-page):

> If you set attributes on a Page object, you should also call save_revision(), since the edit and index view pick up their data from the revisions table rather than the actual saved page record.

In practice this means that when you save a new page as draft and then go to edit it, the section IDs don't show up in the edit view. They only seem to show up in the edit view once you've published once.

cfgov/filing_instruction_guide/wagtail_hooks.py
```diff
@@ -0,0 +1,12 @@
...
+hooks.register("after_edit_page", assign_content_section_ids)
```
chosak, line 12: As an alternative to using Wagtail hooks here, you could take a look at instead using a custom Page model form, which, per the Wagtail documentation [here](https://docs.wagtail.org/en/v2.15.5/advanced_topics/customisation/page_editing_interface.html#customising-generated-forms) can be used "to automatically generate field content". In the code example linked there, the form's `save()` method sets page fields based on the content of other page fields; it seems like this could also be used to modify a field's content as you are doing here.

On a custom form you can also use `clean()` to validate page content. We use a custom Page model form to perform some extra validation checks on JobListingPages [here](https://github.com/cfpb/consumerfinance.gov/blob/f02731f3a44ec3e9ab030e75dc0b4ffa655c2e6b/cfgov/jobmanager/models/pages.py#L53-L74), which I believe is our only current use of a custom `WagtailAdminPageForm`.

Under the hooks-based approach, if a user creates a new page and then previews it without saving, the hooks never get called, and so the section numbers don't appear. Similarly, if a page is published and then a user makes an section-impacting edit and then previews without saving, the outdated previously-saved numbers appear instead of the new ones.

Additionally, the page is first saved based on what happens in the admin and then saved again in the hook. What I like about the custom form approach is that it allows custom page data validation/customization logic to run with the rest of the code happening in the admin, which means that only a single save is needed (or no saves, when previewing).

If the current implementation meets your needs, that's great, but I thought this might be another option to explore.

cfgov/jinja2/v1/_includes/organisms/fig-section.html
```diff
@@ -0,0 +1,25 @@
+<div>
+    {% if block.block_type == 'Fig_Section' %}
+        <h2 class="report-header">
+	    <a id="{{ value.section_id }}"
```
chosak, line 4: Could these IDs be generated dynamically, as part of the rendering, or is it important that they live in the page data? If they don't have to actually persist in the database there'd be no need to do the page-editing complications above. (I do note that the previous implementation you change here worked more like that, so forgive me if this suggestion isn't helpful.)

One approach to generate at render time (even with the StreamField-based approach) would be to create some kind of Jinja tag that returned an appropriate ID each time it was called, so if you had a template (or multiple nested templates) with calls like this:

```html
{{ get_section_number('section') }}
{{ get_section_number('subsection') }}
{{ get_section_number('section') }}
{{ get_section_number('subsection') }}
```

it'd render like this:

```
1.
1.1.
2.
2.1.
```

We have an example of doing something like that [here](https://github.com/cfpb/consumerfinance.gov/blob/54dddc061b5f843a575a6887bb412a4033de8365/cfgov/v1/jinja2tags/__init__.py#L99); this works by storing the current state as you render onto the request object.

schbetsy: I hadn't considered that kind of approach at all, super interesting! I'm still not sure which approach is better (generating section IDs at save time vs page load time), but here was my reasoning for assigning them at save time:
1. A content editor might benefit from seeing the section IDs in the Wagtail admin, to help them orient when editing the content. (Just speculation -- I haven't tested this)
2. The section IDs in the table of contents need to match the section IDs in the main page content, since the anchor links are based on section ID. I figured if we calculate the IDs at page load time, we'd be doing so twice -- once for TOC, and again for main content, which could be more error-prone. (Again, speculation)

But I don't love the solution I've made either. It has section ID as editable field when the user isn't supposed to edit it, which is annoying. It also has the issue you noted above where the section IDs don't get assigned before previewing, though I'll look into the solution you suggested.

cfgov/filing_instruction_guide/wagtail_hooks.py
schbetsy: Thanks for suggesting this! Since I never use the preview workflow, I hadn't noticed that it doesn't update the numbering 😬  . I'll look into the `clean()` form strategy.

schbetsy: I just pushed an update using the custom form strategy. That works much better and is more straight forward, good suggestion. Thanks for taking the time to explain it @chosak!

cfgov/filing_instruction_guide/models.py
```diff
@@ -169,16 +69,47 @@ class FIGContentPage(CFGOVPage):
...
-
     def get_context(self, request, *args, **kwargs):
         context = super().get_context(request, *args, **kwargs)
-        context.update(
-            {
-                "report_sections": get_report_sections(request, self),
-                "report_appendices": get_report_appendices(request, self),
-            }
-        )
+        context.update({"get_sections": self.get_toc_headers(request)})
```
wpears, line 114: Can you rename this to just `sections`? Feels like `get_sections` implies it's a function that needs to be called

cfgov/jinja2/v1/_includes/organisms/sidenav-toc.html
```diff
@@ -29,49 +29,23 @@ <h3>Table of contents</h3>
...
-              {%- for child in item.children -%}
-                  <li class="m-list_item">
-                      {{ nav_link.render(child.title, "#" + child.id, false) }}
-                  </li>
-              {%- endfor %}
-              </ul>
-          {%- endif -%}
-          </li>
-      {%- endfor %}
+       {%- for item in get_sections %}
```
wpears, line 32: Same note on `get_sections` -> `sections`

cfgov/filing_instruction_guide/models.py
```diff
@@ -169,16 +69,47 @@ class FIGContentPage(CFGOVPage):
...
+    def assign_section_ids(self):
+        ind = sub_ind = sub3_ind = 0
+        for section in self.content:
+            id = ""
+            sec_type = section.block_type
+            if sec_type == "Fig_Section":
+                ind += 1
+                sub_ind = 0
+                sub3_ind = 0
+                id = f"{ind}."
```
wpears, line 99: I think it would be better to remove the trailing period here (and on lines 103 and 106). I don't think the trailing periods add much clarity and we'd also hew more closely to https://www.consumerfinance.gov/rules-policy/regulations/ etc

wpears (changes requested):
Few minor comments

cfgov/filing_instruction_guide/models.py
schbetsy: Good call. Done in 94e6733bc591a957308339e34173f68bc206de3f.

cfgov/jinja2/v1/_includes/organisms/sidenav-toc.html
schbetsy: :+1:

cfgov/filing_instruction_guide/models.py
schbetsy: Interesting. We included the trailing period because it was included in the original design by @nataliafitzgerald (which you can see in GHE/super-regular/issues/73). Since they'd be easy to remove later, maybe we could consult a designer and make that improvement later?

schbetsy: Thank you @wpears for reviewing!

wpears (approved):

PR #6044, opened 2020-09-28
Archiving: Add non-navigable archive breadcrumb

virginiacc:
Adds a non-navigable `Archive` breadcrumb to archived pages.

## Additions

- Adds `Archive` breadcrumb if page has been marked as archived

## How to test this PR

1. Check out branch
2. In admin, mark page as archived
3. Navigate to page and note `Archive` breadcrumb

## Screenshots

Screen Shot 2020-09-27 at 5 23 18 PM

PR #7075, opened 2022-05-20
Fix "filter_children" feature of filterable lists

chosak:
The FilterableList organism has a configuration option named "filter_children" with help text "If checked this list will only filter its child pages."

Prior to the 2021 migration of filterable lists to Elasticsearch (specifically #6044), this selection would mean that a filterable
list page would only return pages that were direct children of itself. After the migration, grandchildren and other descendants were included.

This change reverts back to the previous behavior, meaning filterable lists work in two modes:

1. Filter direct children only (the default)
2. Filter every filterable page on the site (enabled by unchecking that checkbox when creating a filterable list)

We don't currently have any pages that need to filter an entire subtree. There are currently 3 pages on the site that use this second approach:

https://www.consumerfinance.gov/activity-log/
https://www.consumerfinance.gov/about-us/newsroom/
https://www.consumerfinance.gov/data-research/research-hub/

This fixes internal platform#4147.

## How to test this PR

To test, verify that these three pages still work as expected locally:

http://localhost:8000/data-research/research-hub/

And other pages that have this checkbox checked also work properly:

Also verify that event pages work properly:

http://localhost:8000/about-us/events/
http://localhost:8000/about-us/events/archive-past-events/

Note that this change does include a modification to the Elasticsearch index so you will need to rebuild your index (e.g. via refresh-data.sh) before testing locally.

## Notes and todos

Additionally this will need a production reindex when this launches.

### DIFF
--- a/cfgov/core/testutils/test_cases.py
+++ b/cfgov/core/testutils/test_cases.py
@@ -1,0 +1,56 @@
+from django.test import TestCase
+
+from wagtail.core.models import Page, Site
+
+
+class WagtailPageTreeTestCase(TestCase):
+    """Mixin that creates a Wagtail page tree for tests.
+
+    Test cases that inherit from this should define a get_page_tree()
+    classmethod which returns a tree of pages as a nested list:
+
+    [
+        ParentPage,
+        [
+            ChildPage,
+            ChildPage,
+            [
+                GrandchildPage,
+                GrandchildPage,
+            ],
+        ],
+    ]
+
+    Pages returned from that method get added as children of the Wagtail
+    default site root for the duration of the test case.
+
+    The class variable `page_tree` can also be used by tests to reference
+    pages that were added to the tree.
+    """
+
+    @classmethod
+    def setUpTestData(cls):
+        site = Site.objects.get(is_default_site=True)
+
+        cls.page_tree = cls._build_page_tree(
+            site.root_page, cls.get_page_tree()
+        )
+
+    @classmethod
+    def _build_page_tree(cls, root, page_tree):
+        tree = []
+        parent = root
+
+        for node in page_tree:
+            if isinstance(node, Page):
+                root.add_child(instance=node)
+                tree.append(node)
+                parent = node
+            else:
+                tree.extend(cls._build_page_tree(parent, node))
+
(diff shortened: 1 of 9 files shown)


cfgov/core/testutils/test_cases.py
```diff
@@ -0,0 +1,56 @@
...
+            ],
+        ],
+    ]
+
+    Pages returned from that method get added as children of the Wagtail
+    default site root for the duration of the test case.
+
+    The class variable `page_tree` can also be used by tests to reference
+    pages that were added to the tree.
+    """
```
willbarton, line 29: This is awesome to have!

cfgov/v1/models/filterable_list_mixins.py
```diff
@@ -41,29 +42,36 @@ def get_search_class():
...
+        # Site. If the page doesn't live under a Site (for example, it is
+        # in the Trash), use the default Site.
+        if not children_only:
+            site = self.get_site()
 
-    def get_filterable_search(self):
-        """Return a FilterablePagesDocumentSearch object"""
-        site = self.get_site()
+            if not site:
+                site = Site.objects.get(is_default_site=True)
```
willbarton, line 67: It's kind of a shame that `get_site()` doesn't provide an optional way to fetch the default site.

willbarton (approved):
This looks good to me!

cfgov/v1/models/filterable_list_mixins.py
```diff
@@ -41,29 +42,36 @@ def get_search_class():
...
-        if filterable_list_block is None:
-            return "/"
-
-        if filterable_list_block.value["filter_children"]:
-            return self.get_url()
+    def get_filterable_search(self):
+        # By default, filterable pages only search their direct children.
+        # But this can be overriden by a setting on a FilterableList block
+        # added to the page's content StreamField.
+        if filter_block := self.get_filterable_list_wagtail_block():
```
willbarton, line 55: Walrus clapping

csebianlander (approved):
All of this (and the explanation) looks very sensible to me from a content-editor perspective. Thanks for resolving this!

PR #7078, opened 2022-05-24
Fix deployable zipfile extraction

willbarton:
This PR fixes our deployable zipfile extraction. Previously our zipfile extraction supported multiple Python versions (intended for 2 and 3), and used an internal pip `RequirementSet` to ensure the wheel files for the target Python were installed. [pip recently moved the function we used](https://github.com/pypa/pip/commit/d673aa14284788ea12a789b34846353b7cb3d46f), which makes it harder to call that function in isolation.

This PR removes the use of pip's internals and multiple Python version support entirely. It also unpins pip in our test requirements, as we install the latest pip when deploying and should test against that.

## Checklist

### DIFF
--- a/cfgov/deployable_zipfile/create.py
+++ b/cfgov/deployable_zipfile/create.py
@@ -62,7 +62,6 @@ def create_zipfile(
     requirements_file,
     zipfile_basename,
     extra_static,
-    extra_python,
 ):
     with temp_directory() as temp_dir:
         # Bootstrap wheels get stored in the zip under /bootstrap_wheels/.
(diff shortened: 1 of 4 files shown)


cfgov/deployable_zipfile/create.py
```diff
@@ -73,8 +72,6 @@ def create_zipfile(
         wheel_dir = os.path.join(temp_dir, "wheels")
 
         wheel_executables = [sys.executable]
```
chosak, line 74: The `for` loop is no longer needed, and this whole code block (lines 75-82) can be simplified to

```py
save_wheels(sys.executable, wheel_dir, "-r%s" % requirements_file)
```

willbarton: Removed in [84918c2](https://github.com/cfpb/consumerfinance.gov/pull/7078/commits/84918c2b666993d4b1edc9a43959dd8288041a36).

PR #7081, opened 2022-05-25
Deprecate page archiving functionality

chosak:
This commit deprecates the CFGOVPage archiving functionality that was implemented in 2020 but never launched. This removes the common `is_archived` and `archived_at` fields on all pages, as well as the related Django management commands `archive_pages` and `restore_archive_pages`. It also removes all related frontend components related to the display of archived pages.

See internal platform#4416 for some discussion on this change.

## How to test this PR

There's not much to test since this is removal of functionality. But you can confirm there are no archiving options when editing a page in Wagtail. Besides that the only potential impact might be the changes to the frontend templates, but again those should just be removal of components that are unused, so a FEWD code review would be helpful there.

## Checklist


willbarton (approved):
This looks thorough to me!

PR #7046, opened 2022-05-02
Deprecate feedback forms

chosak:
This PR deprecates the following feedback forms on the website, each previously implemented as a separate PR:

1. The Wagtail feedback block (#7007)
2. Planning for retirement feedback form (#7008)
3. Buying a House feedback form (#7017)
4. Paying for College feedback form (#7019)

## How to test this PR

See linked PRs above.

## Notes and todos

This PR does not clean up the backend feedback storage and related code such as the feedback export tool; that code will be cleaned up in a future PR.

## Checklist

- [x] PR has an informative and human-readable title
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code follows the standards laid out in the [CFPB development guidelines](https://github.com/cfpb/development)

PR #7082, opened 2022-05-26
Deprecate conference registration code and "form blocks" path

chosak:
With the removal of website feedback forms in #7046, we were left with only a single use of our custom [forms in Wagtail ](https://cfpb.github.io/consumerfinance.gov/forms-in-wagtail/) approach: the conference registration form. This code has not been used since 2018; more recent use cases have used Qualtrics instead.

The existing forms approach supports POST requests to any Wagtail page that may be routed to an AbstractFormBlock-derived StreamField block that performs some action. The relevant code is here, and reading this docstring is an indication of how fragile this code has been:

https://github.com/cfpb/consumerfinance.gov/blob/e948e1f589e44999e4744911a73d06e5ff5905cc/cfgov/v1/models/base.py#L365-L377

This means that POSTs to any page on the site can theoretically be valid, and we have to have logic to process and approve or reject these. In the past this has caused various bugs and also opens up a potential vulnerability surface. It would be better if we didn't support this behavior at all in favor of specific POST-able endpoints. (Removing it also makes it more feasible to someday serve Wagtail URLs as static content.)

Given the above, this PR:

1. Removes all conference registration related code, including the database table used to store registration data.
2. Removes the framework that supports POSTing to CFGOVPages and related form handling code.

This is a conceptually large change that removes quite a bit of unused code (29 files deleted).

## Notes and todos

We have some (internal) code around the conference registration database table because of its potential to store PII; this code will also need to be cleaned up.

## Checklist


willbarton (approved):
So nice to see the complexity for handling `POST` coming out of `CFGOVPage`.

And some of the circular dependencies between packages and their blocks too!

higs4281 (approved):
Cathartic.

PR #7084, opened 2022-05-27
Helm Postgres Updates

jslay-excella:
* Updates `postgresql` child chart to `11.4.0`
* Adds `cfgov.postgresEnv` template helper for postgres environment variables. 
  * Now leverages `.Subcharts` function to allow `postgresql` chart to provide the values to us.


willbarton (changes requested):
I'd like to better understand why this breakout is necessary. It does away with any advantage [dj_database_url](https://github.com/jazzband/dj-database-url) gives us for human-friendliness in local development, while leaving the dependency in place, and it's not clear to me what we gain in return.

This change will also break all of our deployments without corresponding changes to our Ansible roles and variables.

jslay-excella:
Superceded by #7123

